PR #116, opened 2023-09-29
Mirrors branch deniallugo-pla-539-refactor-should_stop_execution from…

Deniallugo:
… private at a4d5ca6625210471d9de66d61e7c9a41a336afb8

# What ❔

## Checklist

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -2,4 +2,3 @@
 .github/release-please/** @RomanBrodetski @perekopskiy @Deniallugo @popzxc
 **/CHANGELOG.md @RomanBrodetski @perekopskiy @Deniallugo @popzxc
 CODEOWNERS @RomanBrodetski @perekopskiy @Deniallugo @popzxc
-.github/workflows/** @matter-labs/devops
--- a/core/lib/zksync_core/src/api_server/execution_sandbox/error.rs
+++ b/core/lib/zksync_core/src/api_server/execution_sandbox/error.rs
@@ -61,6 +61,10 @@ impl From<Halt> for SandboxExecutionError {
             Halt::FailedToAppendTransactionToL2Block(reason) => {
                 SandboxExecutionError::Revert(reason, vec![])
             }
+            Halt::TracerCustom(reason) => SandboxExecutionError::Revert(reason, vec![]),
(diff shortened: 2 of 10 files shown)


CODEOWNERS
StanislavBreadless, line 1: Are you sure it should be in the scope of this PR?

core/lib/zksync_core/src/api_server/execution_sandbox/error.rs
```diff
@@ -61,6 +61,10 @@ impl From<Halt> for SandboxExecutionError {
             Halt::FailedToAppendTransactionToL2Block(reason) => {
                 SandboxExecutionError::Revert(reason, vec![])
             }
+            Halt::TracerCustom(reason) => SandboxExecutionError::Revert(reason, vec![]),
+            Halt::ValidationOutOfGas => Self::UnexpectedVMBehavior(
```
StanislavBreadless, line 65: `Self::UnexpectedVMBehavior` may not be appropriate here, since it would say that the VM should've never returned such error & that the user should contact the team

CODEOWNERS
Deniallugo: it's a bug from the mirroring.

core/lib/zksync_core/src/api_server/execution_sandbox/error.rs
Deniallugo: are you suggesting Revert?

core/lib/zksync_core/src/api_server/execution_sandbox/error.rs
```diff
@@ -62,8 +62,9 @@ impl From<Halt> for SandboxExecutionError {
                 SandboxExecutionError::Revert(reason, vec![])
             }
             Halt::TracerCustom(reason) => SandboxExecutionError::Revert(reason, vec![]),
-            Halt::ValidationOutOfGas => Self::UnexpectedVMBehavior(
+            Halt::ValidationOutOfGas => Self::Revert(
```
StanislavBreadless, line 65: Maybe better to reuse `Self::AccountValidationFailed`?

StanislavBreadless (approved):

PR #120, opened 2023-09-29
feat: Rewrite server binary to use `vise` metrics

slowli:
# What ❔

Makes the `zksync_core` crate use `vise` metrics.

## Why ❔

This completes transition of the server codebase to `vise` metrics.

## Checklist


saitgn7:
Good

core/lib/zksync_core/src/metrics.rs
```diff
@@ -0,0 +1,186 @@
...
+impl fmt::Display for BlockStage {
+    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Sealed => formatter.write_str("sealed"),
+            Self::Tree => formatter.write_str("tree"),
+            Self::MetadataCalculated => formatter.write_str("metadata_calculated"),
+            Self::MerkleProofCalculated => formatter.write_str("merkle_proof_calculated"),
+            Self::L1 { l1_stage, tx_type } => {
+                let l1_stage = match l1_stage {
+                    BlockL1Stage::Saved => "save", // not "saved" for backward compatibility
```
slowli, line 79: It's actually spelled in 2 different ways: `eth_sender` uses "save", and house keeper uses "saved".

core/lib/zksync_core/src/api_server/execution_sandbox/vm_metrics.rs
```diff
@@ -1,104 +1,189 @@
...
+    Initialization,
+    ValidateInSandbox,
+    Validation,
+    Execution,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, EncodeLabelValue, EncodeLabelSet)]
+#[metrics(label = "stage", rename_all = "snake_case")]
+pub(in crate::api_server) enum SubmitTxStage {
+    #[metrics(name = "1_validate")]
```
slowli, line 86: Manual specs could be avoided if we don't need numeric prefixes.

core/lib/zksync_core/src/api_server/web3/metrics.rs
```diff
@@ -0,0 +1,197 @@
...
+    fn from(transport: &ApiTransport) -> Self {
+        match transport {
+            ApiTransport::Http(_) => Self::Http,
+            ApiTransport::WebSocket(_) => Self::Ws,
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, EncodeLabelValue)]
+#[metrics(rename_all = "snake_case")] // FIXME: was PascalCase
```
slowli, line 34: `rename_all` currently doesn't support PascalCase, and IMO it would make sense to rename these label values for consistency with other label values.

core/lib/zksync_core/src/api_server/execution_sandbox/validate.rs
```diff
@@ -72,36 +73,39 @@ impl TxSharedArgs {
...
             );
             span.exit();
             result
-        }).await.unwrap();
+        })
+        .await
+        .unwrap();
 
-        metrics::histogram!("server.api.validation_sandbox", stage_started_at.elapsed(), "stage" => "validate_in_sandbox");
+        stage_latency.observe(); // FIXME: renamed from `server.api.validation_sandbox`
```
slowli, line 108: So, instead of a separate metric, this is now a part of the `api_web3_sandbox` metric with the `stage="validate_in_sandbox"` label.

core/lib/zksync_core/src/api_server/execution_sandbox/vm_metrics.rs
```diff
@@ -1,104 +1,189 @@
...
+    pub estimate_gas_binary_search_iterations: Histogram<usize>,
+}
+
+#[vise::register]
+pub(in crate::api_server) static SANDBOX_METRICS: vise::Global<SandboxMetrics> =
+    vise::Global::new();
+
+#[derive(Debug, Metrics)]
+#[metrics(prefix = "api_execution")]
+pub(super) struct ExecutionMetrics {
```
slowli, line 117: These metrics look like they should be a part of `SandboxMetrics`, but they have a differing prefix (`api_execution_*` vs `api_web3_*`).

core/lib/zksync_core/src/state_keeper/metrics.rs
```diff
@@ -0,0 +1,378 @@
...
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, EncodeLabelSet)]
+struct TxAggregationLabels {
+    criterion: &'static str,
+    seal_resolution: Option<SealResolutionLabel>,
+}
+
+#[derive(Debug, Metrics)]
+#[metrics(prefix = "server_tx_aggregation")]
+pub(super) struct TxAggregationMetrics {
```
slowli, line 133: This should probably be a part of `StateKeeperMetrics`; it's only a separate type because it has a differing prefix.

core/lib/zksync_core/src/state_keeper/metrics.rs
```diff
@@ -0,0 +1,378 @@
...
+    FinishBatch,
+}
+
+const GAS_PER_NANOSECOND_BUCKETS: Buckets = Buckets::values(&[
+    0.01, 0.03, 0.1, 0.3, 0.5, 0.75, 1., 1.5, 3., 5., 10., 20., 50.,
+]);
+
+/// Executor-related state keeper metrics.
+#[derive(Debug, Metrics)]
+#[metrics(prefix = "state_keeper")]
```
slowli, line 362: Note the discrepancy between this prefix and the prefix for `StateKeeperMetrics` (`server_state_keeper`). IMO, it would make sense to use the same prefix in both cases.

slowli (commented):
General notes:

- I've tried to convert all metrics in the crate, but have stopped at prover-related metrics in the witness generator and prover-related house keeper modules. Will they be moved from the main workspace any time soon?
- Current metric naming is not very consistent (e.g., latencies may have `_time` or `_latency` suffix, or none at all). I've went for backward compatibility, which sometimes meant splitting metric types where logically a single type would look better; not specifying metric units etc.
- I haven't added metric docs for most metrics, only for those I have a good understanding of.
- PTAL at `FIXME`s to ack / not ack metric renaming or removal.

core/lib/zksync_core/src/api_server/execution_sandbox/vm_metrics.rs
popzxc: I don't think it's that bad; can be migrated on per-case basis later.

popzxc (approved):
I didn't check everything thoroughly, but conceptually LGTM.

popzxc (dismissed):

popzxc (approved):

PR #121, opened 2023-09-29
feat(vm): Improve tracer trait

Deniallugo:
# What ❔

Changing the way, how tracers are working in VM. Try to make less calls and add more static dispatching

Tracer Api changes: 

1. Method before decoding now always empty (zk_evm don't call it) 
2. Method before cycle doesn't exist anymore
3. Method after_cycle renamed to finish_cycle and unified with should stop execution
4. Method save_results was removed 
5. All methods from ExecutionProcessing trait moved to VmTracer

Vm Changes: 
Refund tracer now dispatching statically and not dynamically 

## Checklist


mm-zk:
@Deniallugo - can you add more info in the PR description? 
which methods you are removing, what are you changing in the API

core/lib/vm/src/tracers/default_tracers.rs
```diff
@@ -222,37 +206,64 @@ impl<S: WriteStorage, H: HistoryMode> DefaultExecutionTracer<S, H> {
...
+                }
+            }
+            VmExecutionMode::Batch => {}
+        };
+        if self.validation_run_out_of_gas() {
+            return TracerExecutionStatus::Stop(TracerExecutionStopReason::Abort(
+                Halt::ValidationOutOfGas,
+            ));
+        }
+        TracerExecutionStatus::Continue
```
RomanBrodetski, line 229: this is a little hard to follow - we do `match`, don't return only in the `Batch` variabt, and then do an `if`. I think it would be cleaner with

```
match self.execution_mode {
            VmExecutionMode::OneTx if self.tx_has_been_processed() => 
                      TracerExecutionStatus::Stop(TracerExecutionStopReason::Finish)
            }
            VmExecutionMode::Bootloader if self.ret_from_the_bootloader == Some(RetOpcode::Ok) => 
                    TracerExecutionStatus::Stop(TracerExecutionStopReason::Finish);
            }
            VmExecutionMode::Batch if self.validation_run_out_of_gas() => 
TracerExecutionStatus::Stop(TracerExecutionStopReason::Abort(
                Halt::ValidationOutOfGas,
        _ => 
        TracerExecutionStatus::Continue
            
        };
```

core/lib/vm/src/tracers/traits.rs
```diff
@@ -11,16 +11,14 @@ use crate::types::outputs::VmExecutionResultAndLogs;
...
 
 /// Run tracer for collecting data during the vm execution cycles
-pub trait ExecutionProcessing<S: WriteStorage, H: HistoryMode>:
-    DynTracer<S, H> + ExecutionEndTracer<H>
-{
+pub trait VmTracer<S: WriteStorage, H: HistoryMode>: DynTracer<S, H> {
     fn initialize_tracer(&mut self, _state: &mut ZkSyncVmState<S, H>) {}
-    fn before_cycle(&mut self, _state: &mut ZkSyncVmState<S, H>) {}
-    fn after_cycle(
+    fn finish_cycle(
```
mm-zk, line 16: can you add a comment - when is this method executed? after each op code?

core/lib/vm/src/tracers/traits.rs
```diff
@@ -29,6 +27,8 @@ pub trait ExecutionProcessing<S: WriteStorage, H: HistoryMode>:
         _stop_reason: VmExecutionStopReason,
     ) {
     }
+
+    fn save_results(&mut self, _result: &VmExecutionResultAndLogs) {}
```
mm-zk, line 31: Same here - AFAIK you mentioned that you are planning to remove this one very soon, right? or does it stay?

core/lib/vm/src/tracers/refunds.rs
```diff
@@ -332,12 +332,3 @@ pub(crate) fn pubdata_published<S: WriteStorage, H: HistoryMode>(
         + l2_l1_long_messages_bytes
         + published_bytecode_bytes
 }
-
-impl<S: WriteStorage, H: HistoryMode> VmTracer<S, H> for RefundsTracer {
-    fn save_results(&mut self, result: &mut VmExecutionResultAndLogs) {
-        result.refunds = Refunds {
```
mm-zk, line 338: when is the result.refunds set now?

core/lib/vm/src/tracers/default_tracers.rs
```diff
@@ -41,6 +40,7 @@ pub(crate) struct DefaultExecutionTracer<S, H: HistoryMode> {
     in_account_validation: bool,
     final_batch_info_requested: bool,
     pub(crate) result_tracer: ResultTracer,
+    pub(crate) refund_tracer: Option<RefundsTracer>,
```
mm-zk, line 43: can you add more comment here (and explain why the RefundsTracer is needed here at all - and it cannot be simply in the list of custom tracers)

RomanBrodetski:
@Deniallugo can you please explain why the `RefundTracer` needs to be treated specially?

core/lib/vm/src/implementation/execution.rs
```diff
@@ -68,15 +73,21 @@ impl<S: WriteStorage, H: HistoryMode> Vm<S, H> {
 
         let result = tx_tracer.result_tracer.into_result();
 
-        let mut result = VmExecutionResultAndLogs {
+        let refunds = if let Some(refund_tracer) = tx_tracer.refund_tracer {
```
mm-zk, line 76: nit:
refund_tracer.map(get_refund).unwrap_or_default()

core/lib/vm/src/tracers/default_tracers.rs
montekki: The two variants can be squashed into one i think with `|`

core/lib/vm/src/tracers/traits.rs
Deniallugo: i've added comments when it's called

core/lib/vm/src/tracers/refunds.rs
Deniallugo: Now in default tracer

core/lib/vm/src/tracers/traits.rs
Deniallugo: Yeah, and i've removed it completely..

core/lib/vm/src/tracers/default_tracers.rs
Deniallugo: We have to always check `self.validation_run_out_of_gas`. Independent from execution_mode. 
Thanks for the other suggestions :) 


> The two variants can be squashed into one i think with |

Only without if..

core/lib/vm/src/tracers/traits.rs
```diff
@@ -88,10 +81,24 @@ pub enum TracerExecutionStatus {
...
+            (Self::Stop(TracerExecutionStopReason::Abort(reason)), _) => {
+                Self::Stop(TracerExecutionStopReason::Abort(reason.clone()))
+            }
+            (_, Self::Stop(TracerExecutionStopReason::Abort(reason))) => {
+                Self::Stop(TracerExecutionStopReason::Abort(reason.clone()))
+            }
+            (Self::Stop(TracerExecutionStopReason::Finish), _)
+            | (_, Self::Stop(TracerExecutionStopReason::Finish)) => {
+                Self::Stop(TracerExecutionStopReason::Finish)
+            }
```
montekki, line 101: same style for `|` would be nice

montekki (approved):

core/lib/vm/src/implementation/execution.rs
```diff
@@ -18,16 +16,16 @@ use crate::VmExecutionStopReason;
...
-        mut tracers: Vec<Box<dyn VmTracer<S, H>>>,
+        tracers: Vec<Box<dyn VmTracer<S, H>>>,
         execution_mode: VmExecutionMode,
     ) -> VmExecutionResultAndLogs {
+        let mut enable_refund_tracer = false;
         if let VmExecutionMode::OneTx = execution_mode {
-            // For correct results we have to include refunds tracer to the desired tracers
-            tracers.push(RefundsTracer::new(self.batch_env.clone()).into_boxed());
-            // Move the pointer to the next transaction
             self.bootloader_state.move_tx_to_execute_pointer();
```
perekopskiy, line 24: nit: "// Move the pointer to the next transaction" can be left

core/lib/vm/src/tracers/traits.rs
```diff
@@ -88,10 +81,22 @@ pub enum TracerExecutionStatus {
...
+        match (self, other) {
+            (Self::Continue, Self::Continue) => Self::Continue,
+            (Self::Stop(TracerExecutionStopReason::Abort(reason)), _)
+            | (_, Self::Stop(TracerExecutionStopReason::Abort(reason))) => {
+                Self::Stop(TracerExecutionStopReason::Abort(reason.clone()))
+            }
+            (Self::Stop(TracerExecutionStopReason::Finish), _)
+            | (_, Self::Stop(TracerExecutionStopReason::Finish)) => {
+                Self::Stop(TracerExecutionStopReason::Finish)
+            }
```
perekopskiy, line 99: nit: it would be cleaner if the comment and arms (and their order) were consistent.

perekopskiy (commented):
Looks really nice! Left a couple of nits

perekopskiy (approved):

montekki (approved):

core/lib/vm/src/tracers/traits.rs
```diff
@@ -88,10 +81,22 @@ pub enum TracerExecutionStatus {
...
-/// Stop the vm execution if the tracer conditions are met
-pub trait ExecutionEndTracer<H: HistoryMode> {
-    // Returns whether the vm execution should stop.
-    fn should_stop_execution(&self) -> TracerExecutionStatus {
-        TracerExecutionStatus::Continue
+impl TracerExecutionStatus {
+    /// Chose the stricter ExecutionStatus
+    /// If both statuses are Continue, then the result is Continue
+    /// If one of the statuses is Abort, then the result is Abort
+    /// If one of the statuses is Finish, then the result is Finish
```
slowli, line 88: Nit: This part of the comment looks outdated.

core/lib/vm/src/tracers/traits.rs
```diff
@@ -88,10 +81,22 @@ pub enum TracerExecutionStatus {
     Stop(TracerExecutionStopReason),
```
slowli, line 81: Would using a `Result<(), TracerExecutionStopReason>` instead of `TracerExecutionStatus` be non-idiomatic? Or is it the case that other statuses besides `Continue` and `Stop` could be added in the future?

core/lib/vm/src/tracers/traits.rs
```diff
@@ -88,10 +81,22 @@ pub enum TracerExecutionStatus {
...
-pub trait ExecutionEndTracer<H: HistoryMode> {
-    // Returns whether the vm execution should stop.
-    fn should_stop_execution(&self) -> TracerExecutionStatus {
-        TracerExecutionStatus::Continue
+impl TracerExecutionStatus {
+    /// Chose the stricter ExecutionStatus
+    /// If both statuses are Continue, then the result is Continue
+    /// If one of the statuses is Abort, then the result is Abort
+    /// If one of the statuses is Finish, then the result is Finish
+    pub fn stricter(&self, other: &Self) -> Self {
```
slowli, line 89: Nit: Couldn't `stricter` take `self` and `other` by value? IMO, this would be more idiomatic and would make cloning `reason` unnecessary.

Deniallugo: I've measured a performance and seems because it's kinda big structure the reference works better.

core/lib/vm/src/tracers/traits.rs
Deniallugo: I don't feel this right.. 
Stop is not an `Error` and `Continue` is not Ok..

core/lib/vm/src/tracers/traits.rs
Deniallugo: No..
We try to find both `Continue`, then If one of the statues abort than abort, then if one of them finish and neiter of them are abort, then it's finish..

slowli: Ah, sorry; I've misread the comment. To be really nitpicky, `Abort` and `Finish` are stop *reasons* (judging by the type name), not *statuses*.

core/lib/vm/src/tracers/default_tracers.rs
RomanBrodetski: Why? We dobn't check it now since we return

PR #128, opened 2023-10-02
feat(boojum): Adding `zk status prover` command

mm-zk:
# What ❔

* Adding zk status prover command to see the current state of FRI prover
* mostly focused on local deployments (might require some changes before we run it on boojnet)

## Why ❔

* to better see what's going on in the prover pipeline

## Checklist

```shell
zk status prover
```

image

### DIFF
--- a/infrastructure/zk/src/status.ts
+++ b/infrastructure/zk/src/status.ts
@@ -8,18 +8,18 @@ import { assert } from 'console';
 let pool: Pool | null = null;
 
 const GETTER_ABI = [
-    'function getTotalBlocksCommitted() returns (uint256)', 
-    'function getTotalBlocksVerified() returns (uint256)',
+    'function getTotalBlocksCommitted() view returns (uint256)', 
+    'function getTotalBlocksVerified() view returns (uint256)',
 ];
 
 const VERIFIER_ABI = [
-    'function verificationKeyHash() returns (bytes32)',
+    'function verificationKeyHash() view returns (bytes32)',
 ];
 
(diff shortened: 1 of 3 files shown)


infrastructure/zk/src/status.ts
```diff
@@ -0,0 +1,210 @@
+import { Command } from 'commander';
+
+import { Pool } from 'pg';
+import { ethers } from 'ethers';
+import { assert } from 'console';
+
+// Postgress connection pool - must be intialized later - as the ENV variables are set later.
+let pool: any = null;
```
ly0va, line 8: To still get type hints, you can do something like
```suggestion
let pool: Pool | null = null;
```

infrastructure/zk/src/status.ts
```diff
@@ -0,0 +1,210 @@
...
+
+const GETTER_ABI = [
```
ly0va, line 10: More of a nit/suggestion, but if you want to hardcode the ABIs, ethers supports the following format

```
const GETTER_ABI = [
    'function getTotalBlocksCommitted() returns (uint256)', 
    'function getTotalBlocksVerified() returns (uint256)'
]
```

ly0va (approved):

infrastructure/zk/src/status.ts
mm-zk: oh - that's super nice. Thanks !

infrastructure/zk/src/status.ts
mm-zk: done.

PR #142, opened 2023-10-02
fix(prover): Use SNARK wrapper VK for picking prover jobs

ly0va:
# What ❔

Prover and Witgen now use snark wrapper VKs to pick their jobs.

## Why ❔

`prover_protocol_versions` table had SNARK wrapper VKs, but prover and witgens picked jobs based on FRI scheduler VKs.

Since if one changes the other also does, we can use the SNARK one instead, since it's also used in `eth_sender`.


core/lib/zksync_core/src/lib.rs
```diff
@@ -122,6 +123,26 @@ pub async fn genesis_init(
...
+            None,
+            Default::default(),
+            None,
+            contracts_config.verifier_addr,
+            zksync_contracts::verifier_contract(),
+        )
+        .await?;
+
+    assert_eq!(
+        vk, first_l1_verifier_config.recursion_scheduler_level_vk_hash,
```
StanislavBreadless, line 142: Probably typo
```suggestion
        vk_hash, first_l1_verifier_config.recursion_scheduler_level_vk_hash,
```

prover/vk_setup_data_generator_server_fri/src/commitment_utils.rs
```diff
@@ -24,6 +25,8 @@ pub struct VkCommitments {
 
 fn circuit_commitments() -> anyhow::Result<L1VerifierConfig> {
     let commitments = generate_commitments().context("generate_commitments()")?;
+    let snark_wrapper_vk = std::env::var("CONTRACTS_RECURSION_SCHEDULER_LEVEL_VK_HASH")
```
perekopskiy, line 28: There is a JSON with snark wrapper vk (vk_setup_data_generator_server_fri/data/snark_verification_scheduler_key.json), I think you should calculate hash based on this key like it's done for other hashes and not just take it from env.

ly0va: Unfortunately, the VK encodings that we use before hashing in Verifier.sol and in zkevm_test_harness are different (not only the order of fields is different, but also Verifier adds an extra field).

Unless we want to do a large copy-paste the hashes would be different. I think we should leave this for now and add actually computing the vk commitment later.

perekopskiy: This is really unfortunate, I'm ok if we merge it to `boojum-integration` as is to unblock testing, but we should fix it before merging to main.

prover/vk_setup_data_generator_server_fri/src/commitment_utils.rs
```diff
@@ -33,14 +34,26 @@ fn circuit_commitments() -> anyhow::Result<L1VerifierConfig> {
...
+        // You can actually compute the SNARK-wrapper VK from the FRI VK, but this is not yet
+        // implemented in the `zkevm_test_harness`, so instead we're loading it from a separate
+        // file.
+        recursion_scheduler_level_vk_hash: H256::from_str(&snark_wrapper_vk)
+            .context("invalid SNARK wrapper VK")?,
     })
 }
 
+pub get_snark_vk_hash() -> anyhow::Result<H256> {
+    let snark_vk = get_snark_vk()?;
```
mm-zk: does this compile?

ly0va: Already removed, was a merge artifact or something

prover/vk_setup_data_generator_server_fri/data/snark_verification_scheduler_key.json
```diff
@@ -21,46 +21,46 @@
     },
```
mm-zk, line 21: please revert these changes.

core/lib/zksync_core/src/lib.rs
```diff
@@ -122,6 +124,26 @@ pub async fn genesis_init(
     )
     .context("Failed to restore operator address from private key")?;
 
+    // We don't need to worry about backward-compatibility with the old verifier here,
+    // since this is only run during genesis.
+    let eth_client = QueryClient::new(eth_client_url)?;
```
mm-zk, line 129: I'm introducing a flag in https://github.com/matter-labs/zksync-era/pull/158 - so you could depend on it - this way we'd still be backwards compatible.

mm-zk (approved):

PR #145, opened 2023-10-03
feat: make chain id to be U256

shahar4:


core/lib/basic_types/src/lib.rs
```diff
@@ -76,6 +79,76 @@ impl TryFrom<U256> for AccountTreeId {
...
+                // try to parse as hex
+                s.parse::<U256>().map_err(de::Error::custom)?
+            }
+        };
+        assert!(
+            u256 <= L2ChainId::MAX.into(),
+            "Too big chain ID. MAX: {}",
+            L2ChainId::MAX
+        );
+        Ok(L2ChainId(u256))
```
Deniallugo, line 107: Please, use Error instead of panics

core/lib/state/src/in_memory.rs
```diff
@@ -22,7 +22,7 @@ impl InMemoryStorage {
     /// Constructs a storage that contains system smart contracts.
     pub fn with_system_contracts(bytecode_hasher: impl Fn(&[u8]) -> H256) -> Self {
         Self::with_system_contracts_and_chain_id(
-            L2ChainId(IN_MEMORY_STORAGE_DEFAULT_NETWORK_ID),
+            L2ChainId::from(u64::from(IN_MEMORY_STORAGE_DEFAULT_NETWORK_ID)),
```
Deniallugo, line 25: should we just use u64 for `IN_MEMORY_STORAGE_DEFAULT_NETWORK_ID`?

core/lib/types/src/l2/mod.rs
```diff
@@ -299,7 +299,7 @@ fn signature_to_vrs(signature: &[u8], tx_type: u32) -> (Option<U64>, Option<U256
 
     if let Ok(sig) = signature {
         (
-            Some(U64::from(sig.v())),
+            Some(U256::from(sig.v())),
```
Deniallugo, line 302: why? It's U64 in ethereum

AnastasiiaVashchuk: because the idea is to change everything to U256 in the server for ChainId. If chainId is U256 then v will also be U256

PR #151, opened 2023-10-04
feat(prover): Witness change for snark wrapper, updated keys

mm-zk:
# What ❔

* Changed witness for snark wrapper - to pass the previous AUX and META fields
* updated snark_wrapper_verification key to the latest version
* updated the vk_hash to match the newly generated JSON, and made sure that it is being loaded at genesis.

## Why ❔

* to make boojnet work locally


ly0va:
We can merge this, but I'll need to update https://github.com/matter-labs/zksync-era/pull/142 after

StanislavBreadless (approved):

PR #153, opened 2023-10-04
ci: Adds call of build-docker-from-tag.yml from release-please.

yorik:
# What ❔
Adds call of build-docker-from-tag.yml from release-please.
Allows manually call of build-docker-from-tag.yml.

## Why ❔

To create releases due to issue in github which doesn't allow automatic workflow triggers from workflow.

## Checklist

### DIFF
--- a/.github/workflows/build-docker-from-tag.yml
+++ b/.github/workflows/build-docker-from-tag.yml
@@ -1,5 +1,7 @@
 name: Build Image from tag
 on:
+  workflow_call:
+  workflow_dispatch:
   push:
     tags:
       - core-v**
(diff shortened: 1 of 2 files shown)


.github/workflows/build-docker-from-tag.yml
```diff
@@ -1,5 +1,17 @@
 name: Build Image from tag
 on:
+  workflow_call:
```
Deniallugo, line 3: Should we allow users to override the tag?

yorik: If you call that manually the tag will be empty, so user have to supply the tag.

artmakh (approved):

PR #154, opened 2023-10-04
fix(API): return correct v value for Legacy tx

AnastasiiaVashchuk:


core/lib/dal/src/models/storage_transaction.rs
```diff
@@ -336,8 +337,14 @@ impl From<StorageApiTransaction> for api::Transaction {
 impl<'r> FromRow<'r, PgRow> for StorageApiTransaction {
     fn from_row(db_row: &'r PgRow) -> Result<Self, Error> {
         let row_signature: Option<Vec<u8>> = db_row.get("signature");
-        let signature = row_signature
-            .and_then(|signature| PackedEthSignature::deserialize_packed(&signature).ok());
+        let transaction_type = db_row
+            .try_get::<Option<i32>, &str>("tx_format")
```
Deniallugo, line 341: It's not for this PR. 
@perekopskiy should we mark tx_format as not null? seems it supposed to be always exist

core/lib/dal/src/models/storage_transaction.rs
```diff
@@ -336,8 +337,14 @@ impl From<StorageApiTransaction> for api::Transaction {
...
+            .unwrap_or_default();
+
+        let (mut v, mut s, mut r) = (None, None, None);
+        if let (Some(signature), Some(tx_type)) = (&row_signature, transaction_type) {
+            (v, s, r) = signature_to_vrs(signature, tx_type as u32);
+        }
```
Deniallugo, line 347: I think it will be a bit cleaner. 

```suggestion
        let (v,r,s) = if let (Some(signature), Some(tx_type)) = (&row_signature, transaction_type) {
             signature_to_vrs(signature, tx_type as u32)
        } else {
             (None, None, None)
        }
```

Deniallugo: And even with this suggestion, you messed up params. 
you have (v,s,r), but function returns (v,r,s)

core/tests/ts-integration/tests/api/web3.test.ts
```diff
@@ -808,6 +808,18 @@ describe('web3 API compatibility tests', () => {
...
+    test('Should check v value of the legacy transaction', async () => {
+        const LEGACY_TX_TYPE = 0;
+        let tx = await alice.sendTransaction({
+            type: LEGACY_TX_TYPE,
+            to: alice.address
+        });
+        let receipt = await tx.wait();
+
+        let block_with_transactions = await alice.provider.send('eth_getBlockByHash', [receipt.blockHash, true]);
+        expect(block_with_transactions.transactions[0].v >= 27);
```
Deniallugo, line 820: Can we somehow check the signature? like it was in the initial problem.

core/lib/dal/src/models/storage_transaction.rs
perekopskiy: Agree, I would create a ticket to check if there are transactions with tx_format IS NULL, if not then add a constraint, otherwise investigate

AnastasiiaVashchuk: https://linear.app/matterlabs/issue/PLA-612/mark-tx-format-as-not-null-in-db

core/tests/ts-integration/tests/api/web3.test.ts
StanislavBreadless: Not sure what you mean by check the signature, the problem was the at `v` was in the range of [0,1]

StanislavBreadless (approved):

core/tests/ts-integration/tests/api/web3.test.ts
AnastasiiaVashchuk: not sure

AnastasiiaVashchuk: I have run locally the script, that Chainlink uses. It does not return null, as it did before.

Deniallugo: can we integrate this script into ci?

AnastasiiaVashchuk: its Go

Deniallugo: so, I mean, a few commits above you messed up a bit with v,r,s positions and this test won't catch it. 

We have a very specific problem our signature was incorrect. That's why I want to check the signature correctness instead of checking some internal components, that we changed. 

Even more, it's an integration test. So you have to check the whole behavior as a black box, without knowing the exact issue. In this test, you are checking the exact params you changed. But it's better to check the desired behaviour and desired behaviour we can describe in two ways: 
1. The return tx is the exact tx we sent to the server 
2. The signature of this is correct

perekopskiy, line 820: You shouldn't rely on the fact that the transaction is the first in the block. You can just use `provider.getTransaction`. Nit: variables name should be camelCase -- `blockWithTransactions`, etc

core/lib/dal/src/models/storage_transaction.rs
```diff
@@ -503,6 +503,12 @@ pub fn web3_transaction_select_sql() -> &'static str {
 pub fn extract_web3_transaction(db_row: PgRow, chain_id: L2ChainId) -> api::Transaction {
     let mut storage_api_tx = StorageApiTransaction::from_row(&db_row).unwrap();
     storage_api_tx.inner_api_transaction.chain_id = chain_id.as_u64();
+    if storage_api_tx.inner_api_transaction.transaction_type == Some(U64::from(0)) {
+        storage_api_tx.inner_api_transaction.v = storage_api_tx
```
Deniallugo, line 507: why it's not in from row implementation?

Deniallugo (approved):

core/tests/ts-integration/tests/api/web3.test.ts
```diff
@@ -808,6 +808,76 @@ describe('web3 API compatibility tests', () => {
         ).toBeRejected();
     });
 
+    test('Should check transaction signature', async () => {
+        const CHAIN_ID = +process.env.CHAIN_ETH_ZKSYNC_NETWORK_ID!;
```
perekopskiy, line 812: nit: you can get chain id from API

core/tests/ts-integration/tests/api/web3.test.ts
```diff
@@ -808,6 +808,76 @@ describe('web3 API compatibility tests', () => {
...
+
+        txFromApi = await alice.provider.getTransaction(tx_handle.hash);
+
+        signerAddr = ethers.utils.recoverAddress(ethers.utils.keccak256(serializedEip1559TxReq), {
+            r: txFromApi.r!,
+            s: txFromApi.s!,
+            v: txFromApi.v!
+        });
+        expect(signerAddr).toEqual(alice.address);
+        expect(Math.abs(txFromApi.v!) <= 1).toEqual(true);
```
perekopskiy, line 878: nit: `Math.abs` is not needed here

core/tests/ts-integration/tests/api/web3.test.ts
```diff
@@ -808,6 +808,76 @@ describe('web3 API compatibility tests', () => {
         ).toBeRejected();
     });
 
+    test('Should check transaction signature', async () => {
```
perekopskiy, line 811: Please add comments explaining what the test actually does and checks

core/tests/ts-integration/tests/api/web3.test.ts
AnastasiiaVashchuk: imo, its better to use env var for tests

perekopskiy (approved):

PR #156, opened 2023-10-04
feat(basic_witness_producer_input): Add Basic Witness Producer Input component

EmilLuta:
# What ❔

This PR introduces a new component that will be ran alongside all components in the service. The component extracts all information needed for running Basic Witness Generator in a binary serialized Input and uploads it to Object Store (GCS for now).

## Why ❔

Part of work decoupling Prover Subsystems from Server Core. Today, Witness Generators need Server Core Database's access to generate witnesses. More information can be found [here](https://www.notion.so/matterlabs/Decouple-Witness-Generator-from-Core-Database-313ece3293d74ee8aea1d1fca823df15) for the Witness Generator split and [here](https://www.notion.so/matterlabs/Segregating-prover-and-server-to-use-API-for-communication-6cdcb7a508e84420bdfcb153af0e2524) for the Prover Subsystems and Server core split.

## Checklist


core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
```diff
@@ -0,0 +1,418 @@
...
+
+                let current_l2_block_info =
+                    rt_handle.block_on(read_l2_block_info(&mut connection, miniblock + 1));
+                let prev_l2_block_info =
+                    rt_handle.block_on(read_l2_block_info(&mut connection, miniblock));
+                let l2_block_env = L2BlockEnv {
+                    number: current_l2_block_info.l2_block_number,
+                    timestamp: current_l2_block_info.l2_block_timestamp,
+                    prev_block_hash: prev_l2_block_info.l2_block_hash,
+                    max_virtual_blocks_to_create: 1,
```
Deniallugo, line 273: it's not correct you have to get the value from info. it's only true for api

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
```diff
@@ -0,0 +1,418 @@
...
+        .await
+        .unwrap();
+        let connection_pool = self.connection_pool.clone();
+        tokio::task::spawn_blocking(move || {
+            let rt_handle = tokio::runtime::Handle::current();
+            let mut connection = rt_handle
+                .block_on(connection_pool.access_storage())
+                .unwrap();
+            let pg_storage =
+                PostgresStorage::new(rt_handle.clone(), connection, MiniblockNumber(2_u32), true)
```
Deniallugo, line 233: miniblock number? are you sure? you are working on l1 batch level

core/lib/dal/migrations/20231016052911_add_basic_witness_input_producer_jobs_table.up.sql
```diff
@@ -0,0 +1,23 @@
...
+    attempts              SMALLINT  NOT NULL DEFAULT 0,
+    status                TEXT      NOT NULL,
+    picked_by             TEXT,
+    input_blob_url        TEXT,
+    error                 TEXT,
+    created_at            TIMESTAMP NOT NULL,
+    updated_at            TIMESTAMP NOT NULL,
+    processing_started_at TIMESTAMP,
+    time_taken            TIME,
+    protocol_version      INT       REFERENCES protocol_versions (id)
```
RomanBrodetski, line 13: doesn't l1_batches table have the protocol version? Is it different from this one? If not, why denormalize?

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
```diff
@@ -0,0 +1,307 @@
...
+
+fn execute_tx<S: ReadStorage>(tx: &Transaction, vm: &mut VmInstance<S, HistoryEnabled>) {
+    vm.make_snapshot();
+    if vm
+        .inspect_transaction_with_bytecode_compression(vec![], tx.clone(), true)
+        .is_ok()
+    {
+        vm.pop_snapshot_no_rollback();
+        return;
+    }
```
RomanBrodetski, line 79: oh this is a little hard to follow (I can imagine it was confusing to write as well) - let's add some comments maybe?

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
```diff
@@ -0,0 +1,307 @@
...
+        // BEWARE, HERE BE DRAGONS.
+        // This usage here is a race condition waiting to happen.
+        // Context -- The VM run in BasicWitnessInputProducer MUST be identical to run from StateKeeper.
+        // Race condition scenario:
+        // 1. StateKeeper runs with a config A, which has value a for `validation_computational_gas_limit`
+        // 2. StateKeeper runs the batch and saves data to database.
+        // 3. Deployment is made and changes to config B, with value b for `validation_computational_gas_limit`.
+        // 4. BasicWitnessInputProducer loads config B
+        // 5. BasicWitnessInputProducer runs the L1Batch that was saved with config A using config B.
+        // In this scenario, the outputs will be different, which will make Witness Generation crash.
```
RomanBrodetski, line 108: is there a way to fail even earlier? That is - when we generate the basic witness input. do we have `validation_computational_gas_limit` saved per-block anywhere? Shall we add it to the DB? If we do that, we can read that value instead of relying on the config cc @Deniallugo

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    async fn process_job_impl(
+        job: BasicWitnessInputProducerJob,
+        started_at: Instant,
+        connection_pool: ConnectionPool,
+        validation_computational_gas_limit: u32,
+        l2_chain_id: L2ChainId,
+    ) -> anyhow::Result<WitnessBlockState> {
+        let l1_batch_number = job.l1_batch_number;
```
RomanBrodetski, line 132: nit: I personally don't see much value in this local variable. `job.l1_batch_number` is perfectly readable, and arguably carries more semantical meaning (you **know** what kind of l1 batch number the author refers to - the one from the `job`. I would remove this line

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
```diff
@@ -0,0 +1,307 @@
...
+            fee_account_addr,
+            validation_computational_gas_limit,
+            l2_chain_id,
+        )
+        .await
+        .unwrap();
+
+        drop(connection);
+
+        let connection_pool = connection_pool.clone();
```
RomanBrodetski, line 172: why do we need to do this?

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
```diff
@@ -0,0 +1,307 @@
...
+            let miniblock_and_transactions = rt_handle.block_on(
+                connection
+                    .transactions_dal()
+                    .get_miniblock_with_transactions_for_l1_batch(l1_batch_number),
+            );
+            tracing::info!("Started execution of l1_batch: {l1_batch_number:?}");
+            for (miniblock, txs) in miniblock_and_transactions {
+                tracing::debug!("Started execution of miniblock: {miniblock:?}");
+                for tx in txs {
+                    execute_tx(&tx, &mut vm);
```
RomanBrodetski, line 202: nit: I would add transaction logging at `debug` level as well - will be useful when we'll just start debugging

RomanBrodetski (commented):
good progress - just left some nits

core/lib/dal/src/blocks_dal.rs
```diff
@@ -1395,6 +1396,52 @@ impl BlocksDal<'_, '_> {
         .await?;
         Ok(())
     }
+
+    pub async fn get_last_miniblock_for_l1_batch(
```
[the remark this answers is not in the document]
RomanBrodetski: I don't think it's harder to read.

`let (_, last_miniblock) = blocks_dal.get_miniblock_range_for_l1_batch`

core/lib/dal/src/basic_witness_input_producer_dal.rs
```diff
@@ -0,0 +1,145 @@
...
+        )
+        .instrument("create_basic_witness_input_producer_job")
+        .report_latency()
+        .execute(self.storage.conn())
+        .await
+        .expect("failed to create basic witness input producer job");
+    }
+
+    pub async fn get_next_basic_witness_input_producer_job(&mut self) -> Option<L1BatchNumber> {
+        let max_attempts = 5;
```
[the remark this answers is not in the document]
RomanBrodetski: yes, looked at the links - so there is already an inconsistency - why do we "unstuck" some jobs in house keeper, while add an `OR` statement for others?

So yeah - feel free to keep as is.

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
```diff
@@ -0,0 +1,203 @@
...
+    async fn process_job_impl(
+        job: BasicWitnessInputProducerJob,
+        started_at: Instant,
+        connection_pool: ConnectionPool,
+        validation_computational_gas_limit: u32,
+        l2_chain_id: L2ChainId,
+    ) -> anyhow::Result<WitnessBlockState> {
+        // This spawn_blocking is needed given PostgresStorage's interface is a mix of blocking and non-blocking
+        // The interface may be either refactored or broken down in a async and sync interface.
+        tokio::task::spawn_blocking(move || {
```
akash-chandrakar, line 72: `tokio::task::spawn_blocking` are primarily suitable for long-running computations or blocking I/O.
For other jobs witness-gen/circuit-synth it fine as they are long running cpu-intensive, I am assuming the computation performed here are also long running?

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
```diff
@@ -0,0 +1,203 @@
...
+            .connection_pool
+            .access_storage()
+            .await
+            .expect("couldn't get a connection from the pool");
+        let l1_batch_to_process = connection
+            .basic_witness_input_producer_dal()
+            .get_next_basic_witness_input_producer_job()
+            .await;
+        match l1_batch_to_process {
+            Some(number) => Ok(Some((
```
akash-chandrakar, line 143: nit: we can return None early using `?`.

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    async fn save_result(
+        &self,
+        job_id: Self::JobId,
+        started_at: Instant,
+        artifacts: Self::JobArtifacts,
+    ) -> anyhow::Result<()> {
+        let upload_started_at = Instant::now();
+        let _object_path = self.object_store.put(job_id, &artifacts).await?;
```
akash-chandrakar, line 191: shall we persist the `_object_path` retrieved from gcs into db?

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
```diff
@@ -0,0 +1,203 @@
...
+        metrics::histogram!(
+            "basic_witness_input_producer.upload_input_time",
+            upload_started_at.elapsed(),
+        );
+        let mut connection = self.connection_pool.access_storage().await?;
+        connection
```
akash-chandrakar, line 197: to make sure I understand correctly, the tree would continue to insert into the `proof_generation_details`, i.e how would the saved blob be shared to prover?
I am guessing we would have another PR that would include this blob in the `ProofGenerationDataRequest`?

popzxc (commented):
I also have some general questions, will ask them separately.

core/lib/zksync_core/src/lib.rs
```diff
@@ -617,6 +625,19 @@ pub async fn initialize_components(
     .await
     .context("add_witness_generator_to_task_futures()")?;
 
+    if components.contains(&Component::BasicWitnessInputProducer) {
```
popzxc, line 628: Uh, and one more. Seems like this component will only run one task at a time, utilizing a single thread (VM is single threaded IIRC). Wouldn't we want to have better pod utilization, running N tasks in parallel?

This is assuming that, like other components, we plan to run these producers as stand-alone components and maybe horizontally scale them.

core/lib/zksync_core/src/basic_witness_input_producer/vm_interactions.rs
```diff
@@ -0,0 +1,130 @@
...
+        fee_account_addr,
+        validation_computational_gas_limit,
+        l2_chain_id,
+    )
+    .await
+    .unwrap();
+
+    let pg_storage = PostgresStorage::new(rt_handle.clone(), connection, miniblock_number, true)
+        .with_caches(PostgresStorageCaches::new(
+            128 * 1_024 * 1_024, // 128MB -- picked as the default value used in EN
```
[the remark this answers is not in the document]
RomanBrodetski: it's only valuable if we want to cache it between  l1 batches. Is it the case here?

core/bin/zksync_server/src/main.rs
```diff
@@ -28,7 +28,7 @@ struct Cli {
     /// Comma-separated list of components to launch.
     #[arg(
         long,
-        default_value = "api,tree,eth,data_fetcher,state_keeper,witness_generator,housekeeper"
+        default_value = "api,tree,eth,data_fetcher,state_keeper,witness_generator,housekeeper,basic_witness_input_producer"
```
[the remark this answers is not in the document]
RomanBrodetski: if we run `witness_generator`, we shall run `basic_witness_input_producer` as well, so I think yeah

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
RomanBrodetski: yes, if for nothing else than consistency with other components.

core/lib/zksync_core/src/basic_witness_input_producer/vm_interactions.rs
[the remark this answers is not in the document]
popzxc: ...but then the cache must be created externally, and not for each VM instance. Plus we must expect that the batches will be processed in order, otherwise cache will be mostly invalidated (e.g. caches won't really work if there will be multiple producers running).

core/lib/dal/migrations/20231016052911_add_basic_witness_input_producer_jobs_table.up.sql
```diff
@@ -0,0 +1,22 @@
...
+    updated_at            TIMESTAMP NOT NULL,
+    processing_started_at TIMESTAMP,
+    time_taken            TIME
+);
+
+CREATE INDEX IF NOT EXISTS idx_basic_witness_input_producer_jobs_status
+    ON basic_witness_input_producer_jobs (status);
+
+CREATE INDEX IF NOT EXISTS idx_basic_witness_input_producer_jobs_attempts_processing_status
+    ON basic_witness_input_producer_jobs (processing_started_at, attempts, status);
```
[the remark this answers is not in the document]
RomanBrodetski: this is a strange index indeed - I think the `attempts` and `status` fields will never be used, as they will only be indexed/sorted **within** one value of `processing_started_at`. But `processing_started_at` is obviously a high-cardinality field

core/lib/dal/src/blocks_dal.rs
```diff
@@ -1395,6 +1396,32 @@ impl BlocksDal<'_, '_> {
         .await?;
         Ok(())
     }
+
+    pub async fn get_fee_address_for_l1_batch(
+        &mut self,
+        l1_batch_number: &L1BatchNumber,
```
perekopskiy, line 1402: nit: I think receiving number by reference doesn't make sense there, it's cheaper to copy number than reference it

core/lib/dal/src/blocks_dal.rs
```diff
@@ -1395,6 +1396,32 @@ impl BlocksDal<'_, '_> {
...
+            l1_batch_number.0 as u32
+        )
+        .fetch_optional(self.storage.conn())
+        .await?
+        .map(|row| be_bytes_to_safe_address(&row.fee_account_address).unwrap()))
+    }
+
+    pub async fn get_virtual_blocks_for_miniblock(
+        &mut self,
+        miniblock_number: &MiniblockNumber,
```
perekopskiy, line 1415: nit: same

core/lib/zksync_core/src/lib.rs
RomanBrodetski: @popzxc but then we can probably just have pods with a single CPU for this?

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
```diff
@@ -0,0 +1,203 @@
...
+        // Race condition scenario:
+        // 1. StateKeeper runs with a config A, which has value a for `validation_computational_gas_limit`
+        // 2. StateKeeper runs the batch and saves data to database.
+        // 3. Deployment is made and changes to config B, with value b for `validation_computational_gas_limit`.
+        // 4. BasicWitnessInputProducer loads config B
+        // 5. BasicWitnessInputProducer runs the L1Batch that was saved with config A using config B.
+        // In this scenario, the outputs will be different, which will make Witness Generation crash.
+        // This has been discussed between @evl and @deniallugo as a terrible solution, but best so far.
+        let validation_computational_gas_limit =
+            state_keeper_config.validation_computational_gas_limit;
```
perekopskiy, line 49: `validation_computational_gas_limit` in state keeper is used to reject some transactions. In this component you only re-execute non-rejected transactions, so I think it makes sense to set this limit to `<TYPE>::MAX`. We do so in EN IIRC. This also resolves race condition

core/lib/zksync_core/src/lib.rs
popzxc: @RomanBrodetski true. I don't know what's more cost-effective, have N single CPU nodes or have N/M M-CPU nodes. If we can get a single CPU with performance similar to the node running state keeper, I guess it would be sufficient.

RomanBrodetski: it's just so much simpler to have it single-threaded.. We pay for nodes, not for pods, so if it's cheaper to have a noce with 4 CPUs we can just run 4 pods 1 CPU each there (IIUC)

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
akash-chandrakar: I see, if we intend to not populate the db with blob_url, then the  blob_url column isn't very useful since it would be always empty.
For debugging purposes it very handy to have the `blob_url` present in the db.

core/lib/dal/src/basic_witness_input_producer_dal.rs
```diff
@@ -0,0 +1,152 @@
...
+        )
+        .instrument("create_basic_witness_input_producer_job")
+        .report_latency()
+        .execute(self.storage.conn())
+        .await
+        .expect("failed to create basic witness input producer job");
+    }
+
+    pub async fn get_next_basic_witness_input_producer_job(&mut self) -> Option<L1BatchNumber> {
+        let max_attempts = 10;
```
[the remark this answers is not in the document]
akash-chandrakar: with the config we won't need to manually re-queue the jobs in the `db` if we have exceeded processing it within its retry attempts, with constant we might require updating the db.

core/lib/zksync_core/src/lib.rs
popzxc: Sure

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
popzxc: `spawn_blocking` is mandatory here, and it's addressed in one of my comments.

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
```diff
@@ -0,0 +1,203 @@
...
+
+    async fn process_job_impl(
+        job: BasicWitnessInputProducerJob,
+        started_at: Instant,
+        connection_pool: ConnectionPool,
+        validation_computational_gas_limit: u32,
+        l2_chain_id: L2ChainId,
+    ) -> anyhow::Result<WitnessBlockState> {
+        // This spawn_blocking is needed given PostgresStorage's interface is a mix of blocking and non-blocking
+        // The interface may be either refactored or broken down in a async and sync interface.
```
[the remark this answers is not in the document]
popzxc: Yup, I think the comment needs to either be rephrased or removed.

core/lib/dal/migrations/20231016052911_add_basic_witness_input_producer_jobs_table.up.sql
```diff
@@ -0,0 +1,22 @@
...
+
+CREATE INDEX IF NOT EXISTS idx_basic_witness_input_producer_jobs_time
```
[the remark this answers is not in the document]
popzxc: Wouldn't logs suffice for troubleshooting? I'm not a fan of assumptions that we're going to debug prod via accessing prod db (and designing the system for that).

core/lib/zksync_core/src/lib.rs
```diff
@@ -617,6 +625,19 @@ pub async fn initialize_components(
...
+        add_basic_witness_input_producer(
+            &mut task_futures,
+            &connection_pool,
```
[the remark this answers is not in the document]
popzxc: > We'll need at least 2 connections. 

Aren't they used sequentially (e.g. never at the same time)? It's not a unique connection, but rather a pool with a single connection. Could it be that you just need to `drop` a connection between instantiating the `PostgresStorage`?

core/lib/zksync_core/src/metadata_calculator/updater.rs
```diff
@@ -200,10 +202,22 @@ impl TreeUpdater {
                     .witness_generator_dal()
                     .save_witness_inputs(l1_batch_number, object_key, protocol_version_id)
                     .await;
-                storage
+                let mut transaction = storage
+                    .start_transaction()
+                    .await
+                    .expect("could not get get transaction from storage");
```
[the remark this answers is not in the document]
popzxc: > We try to get a transaction and if it errors, it means that acquiring transaction failed. Am I missing something obvious?

I think at least in the blockchain context when you hear "get a transaction from storage" you first think about getting an actual tx that can be executed by the VM. With DBs I usually heard the notion "start a DB transaction", so here I'd expect the message to be e.g. "Couldn't start a DB transaction". IMHO, it'll be less ambiguous.

core/lib/zksync_core/src/lib.rs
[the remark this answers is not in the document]
popzxc: This is why I mentioned `drop`. If you will `drop` the `PostgresStorage` (or `VmInstance`), it'll drop the connection and it will return to the pool.
One connection still should suffice.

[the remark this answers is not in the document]
popzxc: You can fetch all this information before instantiating the VM, as suggested in [this comment chain](https://github.com/matter-labs/zksync-era/pull/156#discussion_r1363393559).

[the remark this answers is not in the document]
popzxc: I mean, the suboptimal code in one place should not be considered a justification to write suboptimal code elsewhere.
Even given your example, if the linked ticket will be fixed, there will be no more justification to use more connections than required in this new component 😅

core/lib/zksync_core/src/basic_witness_input_producer/vm_interactions.rs
```diff
@@ -0,0 +1,101 @@
...
+        .unwrap()
+        .unwrap_or_else(|| {
+            panic!(
+                "l1_batch_number {:?} must have fee_address_account",
+                l1_batch_number
+            )
+        });
+    let (system_env, l1_batch_env) = rt_handle
+        .block_on(load_l1_batch_params(
+            // rt_handle.clone(),
```
Deniallugo, line 55: commented code

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+                "Started execution of miniblock: {:?}",
+                miniblock_execution_data.number
+            );
+            for tx in &miniblock_execution_data.txs {
+                tracing::debug!("Started execution of tx: {tx:?}");
+                execute_tx(tx, &mut vm);
+                tracing::debug!("Finished execution of tx: {tx:?}");
+            }
+            let next_miniblock_execution_data = if index + 1 < miniblocks_execution_data.len() {
+                Some(&miniblocks_execution_data[index + 1])
```
Deniallugo, line 83: you can just finish the loop here and do not set the new miniblock.

Deniallugo (commented):
Looks good. 
I'd prefer to have a bit more generic solution. 
But it's a good start. 

Please solve the previous comments

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
Deniallugo: yes, i understand it. 
But when you return None, you call the function, which do literally noting if it's none. 
So for some you can call this function and for none just do nothing and remove some clunky code from set_new_miniblock

core/lib/dal/src/basic_witness_input_producer_dal.rs
```diff
@@ -0,0 +1,152 @@
...
+use zksync_types::L1BatchNumber;
+
+#[derive(Debug)]
+pub struct BasicWitnessInputProducerDal<'a, 'c> {
+    pub(crate) storage: &'a mut StorageProcessor<'c>,
+}
+
+/// Status of a job that the producer will work on.
+#[derive(Debug, strum::Display, strum::EnumString, strum::AsRefStr)]
+pub enum BasicWitnessInputProducerJobStatus {
```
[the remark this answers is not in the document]
Deniallugo: Just for the follow up task. 
It's kinda popular behaviour for all prover related jobs. 
Seems we have to generalize it fully, together with save failed jobs and so on. 

Once we spoke with @RomanBrodetski about unifying all witness related tables to the one big table

.gitignore
```diff
@@ -70,3 +70,6 @@ artifacts-zk/
 cache-zk/
 zksolc
 verified_sources
+
+# RustRover IDE artifacts
+rustc-ice-*
```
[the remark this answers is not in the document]
Deniallugo: i'm slowly fighting with it :)

core/lib/zksync_core/src/basic_witness_input_producer/vm_interactions.rs
```diff
@@ -0,0 +1,98 @@
...
+    miniblock_execution_data: &MiniblockExecutionData,
+) {
+    let miniblock_state = L2BlockEnv {
+        number: miniblock_execution_data.number.0,
+        timestamp: miniblock_execution_data.timestamp,
+        prev_block_hash: miniblock_execution_data.prev_block_hash,
+        max_virtual_blocks_to_create: miniblock_execution_data.virtual_blocks,
+    };
+    vm.start_new_l2_block(miniblock_state);
+}
```
Deniallugo, line 98: NIT: 
i'd prefer to have a function `get_l2_block_env` in `MiniblockExecutionData` and remove this function

core/lib/zksync_core/src/lib.rs
```diff
@@ -823,6 +849,27 @@ async fn run_tree(
     Ok(())
 }
 
+async fn add_basic_witness_input_producer_to_task_futures(
```
perekopskiy, line 852: nit: Witness generator is not spawned when `ZKSYNC_LOCAL_SETUP=true` is passed to env. I think it makes sense to have the same behavior for Producer.

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
```diff
@@ -0,0 +1,185 @@
...
+                miniblock_execution_data.number
+            );
+            for tx in &miniblock_execution_data.txs {
+                tracing::debug!("Started execution of tx: {tx:?}");
+                execute_tx(tx, &mut vm);
+                tracing::debug!("Finished execution of tx: {tx:?}");
+            }
+            if index + 1 < miniblocks_execution_data.len() {
+                vm.start_new_l2_block((&miniblocks_execution_data[index + 1]).into());
+            };
```
popzxc, line 83: Nit: Probably may be implemented slightly more elegantly:

```rust
let mut next_miniblocks = miniblocks_execution_data.iter().skip(1);
for miniblock_execution_data in miniblock_execution_data.iter() {
    // ...tx reexec

    if let Some(next_miniblock) = next_miniblocks.next() {
        vm.start_new_l2_block((&miniblocks_execution_data[index + 1]).into());
    }

    // ...
}
```

core/lib/dal/migrations/20231016052911_add_basic_witness_input_producer_jobs_table.up.sql
```diff
@@ -0,0 +1,19 @@
...
+    updated_at            TIMESTAMP NOT NULL,
+    processing_started_at TIMESTAMP,
+    time_taken            TIME
+);
+
+CREATE INDEX IF NOT EXISTS idx_basic_witness_input_producer_jobs_status
+    ON basic_witness_input_producer_jobs (status);
+
+CREATE INDEX IF NOT EXISTS idx_basic_witness_input_producer_jobs_status_processing_attempts
+    ON basic_witness_input_producer_jobs (status, processing_started_at, attempts);
```
perekopskiy, line 19: The second index starts from `status`, so the first index is useless

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
```diff
@@ -0,0 +1,185 @@
...
+use zksync_types::block::MiniblockExecutionMode;
+
+/// Component that extracts all data (from DB) necessary to run a Basic Witness Generator.
+/// This component will upload Witness Inputs to the object store.
+/// This allows Witness Generator workflow (that needs only Basic Witness Generator Inputs)
+/// to be run only using the object store information, having no other external dependency.
+#[derive(Debug)]
+pub struct BasicWitnessInputProducer {
+    connection_pool: ConnectionPool,
+    validation_computational_gas_limit: u32,
```
perekopskiy, line 26: I think we shouldn't keep `validation_computational_gas_limit` as a field given it's always equal to `u32::MAX`. Also, please add a comment everywhere `u32::MAX` is used

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
```diff
@@ -0,0 +1,185 @@
...
+    async fn process_job(
+        &self,
+        job: Self::Job,
+        started_at: Instant,
+    ) -> JoinHandle<anyhow::Result<Self::JobArtifacts>> {
+        let validation_computational_gas_limit = self.validation_computational_gas_limit;
+        let l2_chain_id = self.l2_chain_id;
+        let connection_pool = self.connection_pool.clone();
+        tokio::task::spawn_blocking(move || {
+            let rt_handle = Handle::current();
```
popzxc, line 153: Nit: For extra explicitness, you may consider making `rt_handle` a field of the structure and obtain it in the constructor. This way, the struct defintion would already tell you that its implementation has some blocking functionality.

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
```diff
@@ -0,0 +1,185 @@
...
+
+    async fn save_result(
+        &self,
+        job_id: Self::JobId,
+        started_at: Instant,
+        artifacts: Self::JobArtifacts,
+    ) -> anyhow::Result<()> {
+        let upload_started_at = Instant::now();
+        let object_path = self.object_store.put(job_id, &artifacts).await?;
+        metrics::histogram!(
```
popzxc, line 173: Please don't forget to migrate the module to `vise`. I failed to find a task in linear for that, probably it makes sense to create one, e.g. as a sub-task in [the umbrella issue](https://linear.app/matterlabs/issue/PLA-166/witness-generation-remove-dependency-from-db).

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
```diff
@@ -0,0 +1,185 @@
...
+            "basic_witness_input_producer.upload_input_time",
+            upload_started_at.elapsed(),
+        );
+        let mut connection = self.connection_pool.access_storage().await?;
+        connection
+            .basic_witness_input_producer_dal()
+            .mark_job_as_successful(job_id, started_at, object_path)
+            .await
+            .unwrap();
```
popzxc, line 182: Nit: No need to `unwrap` here.

core/lib/zksync_core/src/basic_witness_input_producer/vm_interactions.rs
```diff
@@ -0,0 +1,83 @@
...
+) {
+    let prev_l1_batch_number = l1_batch_number - 1;
+    let (_, miniblock_number) = rt_handle
+        .block_on(
+            connection
+                .blocks_dal()
+                .get_miniblock_range_of_l1_batch(prev_l1_batch_number),
+        )
+        .unwrap()
+        .unwrap_or_else(|| {
```
popzxc, line 31: Nit: `process_job_impl` returns `anyhow::Result`, so all the unwraps here can be replaced with corresponding errors.

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
```diff
@@ -0,0 +1,185 @@
...
+
+    fn process_job_impl(
+        rt_handle: Handle,
+        l1_batch_number: L1BatchNumber,
+        started_at: Instant,
+        connection_pool: ConnectionPool,
+        validation_computational_gas_limit: u32,
+        l2_chain_id: L2ChainId,
+    ) -> anyhow::Result<WitnessBlockState> {
+        let mut connection = rt_handle.block_on(connection_pool.access_storage())?;
```
popzxc, line 54: Here and in other places: when returning `anyhow::Result`, please always add [`context`](https://docs.rs/anyhow/latest/anyhow/trait.Context.html). You may look for examples of usage in this crate.

Without `context`, debugging anyhow errors often becomes really confusing.

core/lib/zksync_core/src/basic_witness_input_producer/vm_interactions.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    let pg_storage = PostgresStorage::new(rt_handle.clone(), connection, miniblock_number, true);
+    let storage_view = StorageView::new(pg_storage).to_rc_ptr();
+    let vm = VmInstance::new(l1_batch_env, system_env, storage_view.clone());
+
+    (vm, storage_view)
+}
+
+pub(super) fn execute_tx<S: ReadStorage>(tx: &Transaction, vm: &mut VmInstance<S, HistoryEnabled>) {
+    // attempt to run with bytecode compression
```
popzxc, line 69: Micronit/bikeshedding: Even though it isn't thoroughly followed in the codebase _sigh_, I still nag people to write comments as proper sentences: starting with a capital and ending with a period.

core/lib/zksync_core/src/basic_witness_input_producer/vm_interactions.rs
```diff
@@ -0,0 +1,83 @@
...
+        .is_ok()
+    {
+        vm.pop_snapshot_no_rollback();
+        return;
+    }
+
+    // attempt to run without bytecode compression, if it failed with compression
+    vm.rollback_to_the_latest_snapshot();
+    vm.inspect_transaction_with_bytecode_compression(vec![], tx.clone(), false)
+        .expect("Compression can't fail if we don't apply it");
```
popzxc, line 82: Similarly, panic can be avoided here.

core/lib/zksync_core/src/state_keeper/io/common.rs
```diff
@@ -67,16 +68,16 @@ pub(crate) fn poll_iters(delay_interval: Duration, max_wait: Duration) -> usize
...
     storage: &mut StorageProcessor<'_>,
     current_l1_batch_number: L1BatchNumber,
     fee_account: Address,
     validation_computational_gas_limit: u32,
     chain_id: L2ChainId,
-) -> Option<PendingBatchData> {
-    // If pending miniblock doesn't exist, it means that there is no unsynced state (i.e. no transaction
-    // were executed after the last sealed batch).
+) -> Option<(SystemEnv, L1BatchEnv)> {
+    // If miniblock doesn't exist (for instance if it's pending),
```
popzxc, line 78: Micronit/bikeshedding: the line splits feel a bit random here (e.g. the lines are unnecessarily short and the length is inconsistent).

popzxc (approved):
Looks much better!

Some context on why I suggest to propagate errors instead of panics: our end goal is to be able to run with `panic=abort`, which isn't really possible right now due to panics being so widespread. At the same time, we cannot really implement a strongly-typed error hierarchy right away, thus `anyhow`. But `anyhow` becomes really messy without `context`, so... A bit annoying (and surely inconsistent at the moment), but hopefully it'll get better over time.

akash-chandrakar (approved):

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
popzxc: Well, we don't really have to use `rt_handle` all the time -- but I see your reasoning and indeed probably it's not the best idea. Let's keep it as-is.

core/lib/dal/src/basic_witness_input_producer_dal.rs
```diff
@@ -0,0 +1,168 @@
...
+        .map(|job| L1BatchNumber(job.l1_batch_number as u32));
+
+        Ok(l1_batch_number)
+    }
+
+    pub async fn mark_job_as_successful(
+        &mut self,
+        l1_batch_number: L1BatchNumber,
+        started_at: Instant,
+        object_path: String,
```
slowli, line 107: Nit: You can use `&str` instead of `String` here; a string is not consumed by the method.

core/lib/dal/src/basic_witness_input_producer_dal.rs
```diff
@@ -0,0 +1,168 @@
...
+    /// otherwise it will stay in this state as final state.
+    #[strum(serialize = "failed")]
+    Failed,
+}
+
+impl BasicWitnessInputProducerDal<'_, '_> {
+    pub async fn create_basic_witness_input_producer_job(
+        &mut self,
+        l1_batch_number: L1BatchNumber,
+    ) -> anyhow::Result<()> {
```
slowli, line 51: Is using `anyhow::Result` as opposed to `Result<_, sqlx::Error>` a conscious decision here? AFAICT, the existing DALs use `anyhow::Result` only if errors are generated in DAL logic. (I'd argue that some of these cases could be replaced with [`sqlx::Error::Decode`](https://docs.rs/sqlx/latest/sqlx/enum.Error.html#variant.Decode) / `ColumnDecode`.

core/lib/dal/src/transactions_dal.rs
```diff
@@ -968,34 +968,64 @@ impl TransactionsDal<'_, '_> {
...
+    }
+
+    /// Returns miniblocks with their transactions to be used in VM execution.
+    /// `miniblock_execution_mode` parameter describes what miniblocks are collected:
+    ///    - Reexecution -- miniblocks state_keeper needs to reexecute on restart.
+    ///      These are the transactions that are included to some miniblock,
+    ///      but not included to L1 batch. The order of the transactions is the same as it was
+    ///      during the previous execution.
+    ///    - L1Batch -- miniblocks that are included in the given L1 batch.
+    pub async fn get_miniblocks_to_execute_for(
```
slowli, line 1019: Nit: This method could propagate DB errors as well, although it could require some effort to fix its uses.

core/lib/types/src/block.rs
```diff
@@ -82,16 +82,25 @@ pub struct MiniblockHeader {
...
     pub number: MiniblockNumber,
     pub timestamp: u64,
     pub prev_block_hash: H256,
     pub virtual_blocks: u32,
     pub txs: Vec<Transaction>,
 }
 
+/// Used to determine what kind of query to apply when collection MiniblockExecutionData.
+///   - Reexecute will get all miniblocks with transactions that need reexecution.
+///   - L1Batch will get all miniblocks with transactions belonging to an already executed L1 batch.
```
slowli, line 97: Nit: You can place these comments directly on enum variants.

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
```diff
@@ -0,0 +1,185 @@
...
+        );
+
+        tracing::info!("Started execution of l1_batch: {l1_batch_number:?}");
+        for (index, miniblock_execution_data) in miniblocks_execution_data.iter().enumerate() {
+            tracing::debug!(
+                "Started execution of miniblock: {:?}",
+                miniblock_execution_data.number
+            );
+            for tx in &miniblock_execution_data.txs {
+                tracing::debug!("Started execution of tx: {tx:?}");
```
slowli, line 77: Nit: This could spam logs (of course, could be configured away, but still); I'd suggest to use the `TRACE` level here.

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
slowli, line 83: Nit: extra semicolon.

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
```diff
@@ -0,0 +1,185 @@
...
+                miniblock_execution_data.number
+            );
+        }
+        vm.finish_batch();
+        tracing::info!("Finished execution of l1_batch: {l1_batch_number:?}");
+
+        metrics::histogram!(
+            "basic_witness_input_producer.input_producer_time",
+            started_at.elapsed(),
+        );
```
slowli, line 96: Nit: Is it too much to ask to use `vise` metrics here. I know witness-related server components don't use them, but the remaining parts of the server do.

core/lib/zksync_core/src/basic_witness_input_producer/vm_interactions.rs
```diff
@@ -0,0 +1,83 @@
...
+
+pub(super) fn create_vm(
+    rt_handle: Handle,
+    l1_batch_number: L1BatchNumber,
+    mut connection: StorageProcessor<'_>,
+    validation_computational_gas_limit: u32,
+    l2_chain_id: L2ChainId,
+) -> (
+    VmInstance<PostgresStorage, HistoryEnabled>,
+    Rc<RefCell<StorageView<PostgresStorage>>>,
```
slowli, line 21: Nit: It's better to use `StoragePtr<_>` alias here for future compatibility.

core/lib/zksync_core/src/metadata_calculator/updater.rs
```diff
@@ -201,10 +203,23 @@ impl TreeUpdater {
...
                     .save_witness_inputs(l1_batch_number, object_key, protocol_version_id)
                     .await;
-                storage
+                let mut transaction = storage
+                    .start_transaction()
+                    .await
+                    .expect("could not start DB storage");
+                transaction
+                    .basic_witness_input_producer_dal()
+                    .create_basic_witness_input_producer_job(l1_batch_number)
```
slowli, line 212: Just to understand better: Is it necessary to place job creation in `MetadataCalculator` (as opposed to, say, the state keeper)? Does it depend on the `MetadataCalculator` output? In general, this looks like the distributed orchestration anti-pattern; `MetadataCalculator` orchestrates other components via job queues, which doesn't look like its scope of responsibilities. Could it make sense for the new component to create jobs for itself instead? (By checking new L1 batches without created jobs – seems like a straightforward SQL query.) In this case, it would be closer to choreography pattern (the component reacts to events – encoded as L1 batch entities – emitted by another component, namely the state keeper).

core/lib/dal/src/basic_witness_input_producer_dal.rs
```diff
@@ -0,0 +1,168 @@
...
+#[derive(Debug)]
+pub struct BasicWitnessInputProducerDal<'a, 'c> {
+    pub(crate) storage: &'a mut StorageProcessor<'c>,
+}
+
+/// The amount of attempts to process a job before giving up.
+const JOB_MAX_ATTEMPT: i16 = 10;
+
+/// Time to wait for job to be processed
+const JOB_PROCESSING_TIMEOUT: PgInterval = pg_interval_from_duration(Duration::from_secs(10 * 60));
```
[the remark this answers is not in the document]
popzxc: There is no `Duration::from_minutes` :harold:

core/lib/dal/src/transactions_dal.rs
```diff
@@ -968,44 +969,78 @@ impl TransactionsDal<'_, '_> {
...
+        &mut self,
```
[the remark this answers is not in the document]
popzxc: I also initially thought about having two methods that will share the impl. I don't have a strong opinion tho, fine with either option.

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+        let (mut vm, storage_view) =
+            create_vm(rt_handle.clone(), l1_batch_number, connection, l2_chain_id)
+                .context("failed to create vm for BasicWitnessInputProducer")?;
+
+        tracing::info!("Started execution of l1_batch: {l1_batch_number:?}");
+        let mut next_miniblocks_data = miniblocks_execution_data.iter().skip(1);
+        for miniblock_execution_data in miniblocks_execution_data.iter() {
+            tracing::debug!(
+                "Started execution of miniblock: {:?}",
+                miniblock_execution_data.number
```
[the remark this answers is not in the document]
popzxc: (genuine question) Do we need to? AFAIK we're currently working on reducing the amount of data we log; and all the info on miniblocks is already exposed from the state keeper logs if it's needed.

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+            );
+            for tx in &miniblock_execution_data.txs {
+                tracing::trace!("Started execution of tx: {tx:?}");
+                execute_tx(tx, &mut vm)
+                    .context("failed to execute transaction in BasicWitnessInputProducer")?;
+                tracing::trace!("Finished execution of tx: {tx:?}");
+            }
+            if let Some(next_miniblock_data) = next_miniblocks_data.next() {
+                vm.start_new_l2_block(L2BlockEnv::from_miniblock_data(next_miniblock_data));
```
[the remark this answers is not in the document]
popzxc: We're walking in circles now 😅  It was [my suggestion](https://github.com/matter-labs/zksync-era/pull/156#discussion_r1368217091) to use iterator here.

I'd argue that using indexing is not idiomatic in Rust.

core/lib/zksync_core/src/state_keeper/io/common.rs
popzxc: I guess I'm not a folk, since I read it as "If miniblock doesn't exist (for instance if it's pending)," -> pause (why, I've got no new information yet?) -> "it means that there is no unsynced state" -> pause (I forgot what was on the previous line) -> "(i.e. no transactions were executed after the last sealed batch).". 
Personally I use the following logic: if there is no logical separation between parts of the sentence, do NOT break on commas/periods/etc, break on space to not create a pause; if parts are not related much, do the opposite. E.g. "inject pauses", if that makes sense.

Though not a strong opinion, feel free to keep it as is.

popzxc (commented):
No new findings. I do agree with some of comments other folks left, so will wait with an approval a bit.

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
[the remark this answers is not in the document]
RomanBrodetski: Screenshot 2023-10-24 at 09 13 29

Pasting here for convinience.

Definitely not a big deal and also have no idea what's more idiomatic - but comparing the two versions for me indexes definitely look easier to follow (that mutable variable disgusts me haha also I'm looking for additional usages since the variable definition is not in the same line as usage/rewrite).

Will ask @slowli what he things - more because of curiousity than anything else.

@EmilLuta sorry for back and forth

core/lib/zksync_core/src/metadata_calculator/updater.rs
RomanBrodetski: you are rasiging a good point. We used to create the witness job in metadata calc because witness job does depend on its output (merkle proofs). Interesting point on creating the jobs "for itself" - I like it. @EmilLuta I'd say feel free to keep as and let's discuss our approach to queues as a whole (would be interesting to see who creates jobs for whom right now)

core/lib/dal/src/basic_witness_input_producer_dal.rs
```diff
@@ -0,0 +1,168 @@
...
+    InProgress,
+    /// The final (happy case) state we expect all jobs to end up. After the run is complete,
+    /// the job uploaded it's inputs, it lands in successful.
+    #[strum(serialize = "successful")]
+    Successful,
+    /// The job failed for reasons. It will be marked as such and the error persisted in DB.
+    /// If it failed less than MAX_ATTEMPTs, the job will be retried,
+    /// otherwise it will stay in this state as final state.
+    #[strum(serialize = "failed")]
+    Failed,
```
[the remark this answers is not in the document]
RomanBrodetski: 1. It's good to have the status field typed as a [Postgres Enum](https://www.postgresql.org/docs/current/datatype-enum.html) - as opposed to just a string. This should not be done in this PR, and should probably be tackled together with other queues

2. if we don't have such an enum on the DB level (or if it gets out of sync with the one in rust), we should be careful when we convert DB rows to rust structs that has this enum as one of its fields. The process will panic in runtime if you use `sqlx` to read a DB row with a status value that doesn't have a matching enum variant (using `query_as!` macro). We had such an issue in one of our tools in production. You don't do something like this in this PR so no action is needed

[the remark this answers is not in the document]
Deniallugo: I'd be very careful with postgres enums, it's really hard to maintain them and change the list of enums..

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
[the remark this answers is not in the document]
RomanBrodetski: definitely doesn't feel more functional for me - we maintain a mutable variable which is introduce outside of the scope loop. That said, none of the two are functional - and it cannot be achieved here due to the VM interface

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+        tracing::info!("Finished execution of l1_batch: {l1_batch_number:?}");
+
+        BASIC_WITNESS_INPUT_PRODUCER_METRICS
+            .process_batch_time
+            .observe(started_at.elapsed());
+        tracing::info!(
+            "BasicWitnessInputProducer took {:?} for L1BatchNumber {}",
+            started_at.elapsed(),
+            l1_batch_number.0
+        );
```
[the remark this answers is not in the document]
RomanBrodetski: when we see that the component starts to work slower, we'll immediately see if it's due to increased number of ellmnts it processes or if the logic performance itself degraded

core/lib/zksync_core/src/metadata_calculator/updater.rs
```diff
@@ -201,10 +203,23 @@ impl TreeUpdater {
...
+                    .await
+                    .expect("failed to create basic_witness_input_producer job");
+                transaction
                     .proof_generation_dal()
                     .insert_proof_generation_details(l1_batch_number, object_key)
                     .await;
+                transaction
+                    .commit()
+                    .await
+                    .expect("could not commit transaction");
```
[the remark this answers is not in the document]
RomanBrodetski: > transaction is not necessary -- may save a roundtrip to DB, but that's a really weak argument

probably not - when you do DB queries, it executes all of them as you do them, and when you commit transaction it just sends a `COMMIT ` statement. That is, transactions are managed on the DB side so to say

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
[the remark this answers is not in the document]
popzxc: I don't quite agree. This is just a manual implementation of [`Itertools::zip_longest`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.zip_longest). I wanted to suggest using it, but I thought that the API there isn't really good (introduces a separate type as an item, plus using it would require a slightly odd pattern matching given that we know beforehand which iterator is longer).
The mutable variable is an iterator that is meant to be consumed (and which is created under the hood in every `for` loop in fact); the produced items are still immutable.

This _is_ a functional approach, just the Rust dialect of it.

core/lib/zksync_core/src/state_keeper/io/common.rs
```diff
@@ -134,12 +133,31 @@ pub(crate) async fn load_pending_batch(
...
+        chain_id,
+    )
+    .await?;
 
     let pending_miniblocks = storage
         .transactions_dal()
         .get_miniblocks_to_reexecute()
-        .await;
+        .await
+        .unwrap();
```
popzxc, line 160: Sneaky unwrap be sneaky.

popzxc (dismissed):
LGTM except for unwrap.

Well done. This PR was tough -- normally it gets easier later, once you learn what you'll be nitpicked on 😅

Deniallugo (dismissed):

core/lib/zksync_core/src/state_keeper/io/common.rs
popzxc: Oh right. I noticed the `?` before and didn't understand that it's `Try` on `Option`, not on `Result` 🤦 
Fine to keep it (for now).

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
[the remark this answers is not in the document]
slowli: I was going to suggest using `slice::windows()` 🙃 I don't quite like in the current approach that the "next item" iterator is iterated manually, rather than together with the "current item" iterator. It could be possible to redefined the "next item" iterator as something like

```rust
let next_miniblocks_data = 
    miniblocks_execution_data.iter().skip(1).map(Some).chain([None]);
```

...and then zip it with the "current item" iterator, right?

core/lib/dal/src/transactions_dal.rs
```diff
@@ -972,40 +973,74 @@ impl TransactionsDal<'_, '_> {
...
+    pub async fn get_miniblocks_to_reexecute(
+        &mut self,
+    ) -> anyhow::Result<Vec<MiniblockExecutionData>> {
+        let transactions = sqlx::query_as!(
             StorageTransaction,
             "SELECT * FROM transactions \
-            WHERE miniblock_number IS NOT NULL AND l1_batch_number IS NULL \
-            ORDER BY miniblock_number, index_in_block",
+                    WHERE miniblock_number IS NOT NULL AND l1_batch_number IS NULL \
+                    ORDER BY miniblock_number, index_in_block",
```
slowli, line 983: Bikeshedding: Is extra indentation a conscious decision? I liked the previously used one, and it's more in line with other multiline SQL queries (including new ones in `BasicWitnessInputProducerDal`).

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+        let attempts = self
+            .connection_pool
+            .access_storage()
+            .await
+            .unwrap()
+            .basic_witness_input_producer_dal()
+            .mark_job_as_failed(job_id, started_at, error)
+            .await
+            .expect("errored whilst marking job as failed");
+        match attempts {
```
slowli, line 133: Nit: Using `if let Some(_)` / `else` is arguably more idiomatic with `Option`s than `match`ing.

core/lib/zksync_core/src/metadata_calculator/updater.rs
```diff
@@ -174,6 +174,8 @@ impl TreeUpdater {
             // That is, if we run multiple tree instances, we'll get metadata correspondence
             // right away without having to implement dedicated code.
 
+            // If there is an `object_key`, then the witness is being processed.
+            // If not, it must've been skipped (witness/proving is done at a very small percentage on testnet).
```
slowli, line 178: Nit: I don't quite understand this comment. `object_key` is obtained from `process_l1_batch()` above, and it's `Some` for full trees and `None` for lightweight ones. How does prover sampling factor into this?

slowli (dismissed):
Other than small nits, looks fine 👍

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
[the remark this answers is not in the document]
popzxc: I'm not against it -- the construction feels a bit harder to parse mentally IMHO, but the loop will look nicer indeed.

core/lib/zksync_core/src/basic_witness_input_producer/mod.rs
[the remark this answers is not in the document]
RomanBrodetski: > We'd actually see this in StateKeeper, I reckon this component is not the place to troubleshoot "some" VM behaviours.

But maybe it gets slower for some other reason - not VM-related, but this componen-specific - exactly what we are looking for. 

Feel free to merge as is and we add later if needed

RomanBrodetski (dismissed):
👍

spartucus:
Since the current `witness_generator` use PG instead of blob, as below code: https://github.com/matter-labs/zksync-era/blob/1e431a65d17c3fdd3b873e31b3193ac72a0842f9/prover/witness_generator/src/basic_circuits.rs#L671

This Basic Witness Producer component does not work any more, am I understand correctly?

@EmilLuta @shahar4

RomanBrodetski:
> This Basic Witness Producer component does not work any more, am I understand correctly?

@spartucus the component works, but its output is indeed not used. This migration is planned for the coming months - prover subsystem will not have access to any of the core tables. It will use this component output instead

spartucus:
@RomanBrodetski Thanks for your kind reply. 

>  prover subsystem will not have access to any of the core tables. It will use this component output instead

The `BasicWitnessInputProducer`'s output is `WitnessBlockState`, is it enough for the prover? Or you will add more data in the output?

For the current code implementation, my question is, given that `BasicWitnessInputProducer` creates a `VmInstance`,  which executes the transaction and finish the batch, why can't its output be used as a witness? `WitnessGenerator` on the other hand, uses `zkevm_test_harness::external_calls::run_with_fixed_params` to generate witnesses, any consideration on this?

PR #158, opened 2023-10-04
feat(boojum): added flag to control prover selection at genesis

mm-zk:
# What ❔

* Added CONTRACTS_PROVER_AT_GENESIS flag to control which prover should be used during genesis
* Updated zk status command to compare the keys in the database and in the contracts

## Why ❔

* without this flag, we were pushing the old keys to the contracts, which resulted in failed proofs.


ly0va (approved):
:+1:

PR #162, opened 2023-10-05
feat(storage): save enum indices in RocksDB

perekopskiy:
# What ❔

Enumeration indices now are saved along with values in the same column family. Indices are added gradually for old DB entries. The number of keys processed each L1 batch is configurable.

## Why ❔

Enumeration indices in storage are necessary for boojum upgrade.

## Checklist


core/bin/external_node/src/config/mod.rs
```diff
@@ -282,6 +285,10 @@ impl OptionalENConfig {
         10
     }
 
+    const fn default_enum_index_migration_chunks() -> u16 {
+        1
```
perekopskiy, line 289: The default value for EN is not yet calibrated and will be updated after tests on staging

core/lib/dal/src/storage_logs_dedup_dal.rs
```diff
@@ -133,4 +133,27 @@ impl StorageLogsDedupDal<'_, '_> {
...
+            .iter()
+            .enumerate()
+            .map(|(index, key)| (key, index))
+            .collect();
+        let hashed_keys: Vec<_> = hashed_keys.iter().map(H256::as_bytes).collect();
+        let mut indices: Vec<(H256, u64)> = sqlx::query!(
+            "SELECT hashed_key, index FROM initial_writes \
+            WHERE hashed_key = ANY($1)",
+            &hashed_keys as &[&[u8]]
+        )
```
RomanBrodetski, line 148:
```suggestion
        )
        .instrument("enum_indices_for_keys")
        .report_latency()
```

Would probably be good to instrument - not 100% sure this will compile (didn't test) - please have a look. We can let the CI figure it out as well :)

core/lib/state/src/rocksdb/mod.rs
```diff
@@ -15,7 +15,7 @@
 //! | Contracts    | address (20 bytes)     | `Vec<u8>`               | Contract contents                    |
 //! | Factory deps | hash (32 bytes)        | `Vec<u8>`               | Bytecodes for new contracts that a certain contract may deploy. |
```
RomanBrodetski, line 16: Does this stay intact?

core/lib/state/src/rocksdb/mod.rs
```diff
@@ -131,23 +173,111 @@ impl RocksdbStorage {
...
+
+    async fn save_missing_enum_indices(&self, conn: &mut StorageProcessor<'_>) {
+        if let (Some(start_from), true) = (
+            self.enum_migration_start_from(),
+            self.enum_index_migration_chunks > 0,
+        ) {
+            let started_at = Instant::now();
+            let chunk_from = u16::from_be_bytes(start_from);
+            let chunk_to = chunk_from.saturating_add(self.enum_index_migration_chunks - 1);
+            tracing::info!("RocksDB enum index migration is not finished, processing chunks {chunk_from}..={chunk_to}");
```
RomanBrodetski, line 190: will we immediately see the total number of chunks? It's always good to see the progress, let's report is somehow. Grafana might be an overkill.

Or maybe you do report it and I missed it

core/lib/state/src/rocksdb/mod.rs
```diff
@@ -108,7 +126,33 @@ impl RocksdbStorage {
...
+            assert_eq!(keys_with_unknown_indices.len(), enum_indices.len());
+            self.pending_patch.state = logs_with_known_indices
+                .into_iter()
+                .chain(
+                    logs_with_unknown_indices
+                        .into_iter()
+                        .zip(enum_indices)
+                        .map(|((key, value), index)| (key, (value, index))),
+                )
+                .collect();
```
slowli, line 155: Nit: It could make sense to extract this to a separate method for readability.

core/lib/state/src/rocksdb/mod.rs
```diff
@@ -131,23 +175,111 @@ impl RocksdbStorage {
...
+                for ((key, value), index) in keys.into_iter().zip(values).zip(indices) {
+                    write_batch.put_cf(
+                        StateKeeperColumnFamily::State,
+                        key.as_bytes(),
+                        &Self::serialize_state_value(value, index),
+                    );
+                }
+            }
+
+            if chunk_to == u16::MAX {
```
Deniallugo, line 219: Why? Are we going to process only 65536 batches? (it's a default value for EN node now)

core/lib/state/src/rocksdb/mod.rs
```diff
@@ -131,23 +173,111 @@ impl RocksdbStorage {
...
+            .and_then(|value| {
+                if value.len() == 40 {
+                    Some(u64::from_be_bytes(value[32..].try_into().unwrap()))
+                } else {
+                    None
+                }
+            })
     }
 
     /// Processes storage `logs` produced by transactions.
```
RomanBrodetski, line 268: I think the comment needs to be updated

core/lib/zksync_core/src/state_keeper/batch_executor/mod.rs
```diff
@@ -84,6 +84,7 @@ pub struct MainBatchExecutorBuilder {
     save_call_traces: bool,
     max_allowed_tx_gas_limit: U256,
     upload_witness_inputs_to_gcs: bool,
+    enum_index_migration_chunks: u16,
```
RomanBrodetski, line 87: nit: this reads as the **number of chunks** , not the **chunk_size**. I would rename it to `chunk_size`

core/lib/state/src/rocksdb/mod.rs
```diff
@@ -131,23 +173,111 @@ impl RocksdbStorage {
...
+            let chunk_from = u16::from_be_bytes(start_from);
+            let chunk_to = chunk_from.saturating_add(self.enum_index_migration_chunks - 1);
+            tracing::info!("RocksDB enum index migration is not finished, processing chunks {chunk_from}..={chunk_to}");
+
+            let mut write_batch = self.db.new_write_batch();
+            let mut number_of_keys_migrated = 0;
+            for chunk in chunk_from..=chunk_to {
+                let (keys, values): (Vec<_>, Vec<_>) = self
+                    .db
+                    .prefix_iterator_cf(StateKeeperColumnFamily::State, &chunk.to_be_bytes())
```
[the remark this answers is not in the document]
perekopskiy: Agree, it will be simpler

core/lib/state/src/rocksdb/mod.rs
perekopskiy: Total number of chunks is 65536 by design. Anyway, I'm going to change approach according to https://github.com/matter-labs/zksync-era/pull/162#discussion_r1347307942, and will leave progress reported in logs

core/lib/zksync_core/src/state_keeper/batch_executor/mod.rs
perekopskiy: I actually meant number of chunks, anyway approach will be changed

core/lib/state/src/rocksdb/mod.rs
perekopskiy: Not sure what you mean by batch here. If L1 batches then no, if chunks then yes because 65536 chunks cover all keys. Anyway, approach will be changed according to https://github.com/matter-labs/zksync-era/pull/162#discussion_r1347307942

core/lib/dal/src/storage_logs_dedup_dal.rs
```diff
@@ -133,4 +133,27 @@ impl StorageLogsDedupDal<'_, '_> {
         .map(|row| H256::from_slice(&row.hashed_key))
         .collect()
     }
+
+    pub async fn enum_indices_for_keys(&mut self, hashed_keys: &[H256]) -> Vec<u64> {
```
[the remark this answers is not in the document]
perekopskiy: If it returned `HashMap` then I'd need to add some logic in places where `enum_indices_for_keys` is called. Currently it just calls this method and zips the vector it returns, so I want to it as is

core/lib/state/src/in_memory.rs
```diff
@@ -61,22 +64,38 @@ impl InMemoryStorage {
                 ]
             })
             .chain(system_context_init_log)
-            .filter_map(|log| (log.kind == StorageLogKind::Write).then_some((log.key, log.value)))
+            .sorted_by_key(|log| log.key)
+            .enumerate()
+            .filter_map(|(idx, log)| {
+                (log.kind == StorageLogKind::Write)
+                    .then_some((log.key, (log.value, idx as u64 + 1)))
```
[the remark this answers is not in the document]
perekopskiy: There are no guarantee. Agree with BTreeMap, btw this is how I did it `boojum-integration` branch 🙂

core/lib/state/src/rocksdb/mod.rs
```diff
@@ -131,23 +152,157 @@ impl RocksdbStorage {
...
+        tracing::info!(
+            "RocksDB enum index migration is not finished, current progress {}%, starting from key {start_from:?}",
+            progress_percentage.with_prec(5)
+        );
+
+        let mut write_batch = self.db.new_write_batch();
+        let (keys, values): (Vec<_>, Vec<_>) = self
+            .db
+            .from_iterator_cf(StateKeeperColumnFamily::State, start_from.as_bytes())
+            .take(self.enum_index_migration_chunk_size)
```
slowli, line 213: Wouldn't you want to place `take()` after `filter_map` so that it counts the number of actually migrated keys?

core/lib/state/src/rocksdb/mod.rs
```diff
@@ -131,23 +152,157 @@ impl RocksdbStorage {
...
+                &Self::serialize_state_value(value, index),
+            );
+        }
+
+        let next_key = keys
+            .last()
+            .and_then(|last_key| h256_to_u256(*last_key).checked_add(U256::one()))
+            .map(u256_to_h256);
+        match next_key {
+            None => {
```
slowli, line 239: - Nit: `if let Some(_) = next_key` is considered more idiomatic than `match`.
- If you replace `filter_map` and `take` as suggested in the comment above, you would be able to detect the last chunk by the number of keys in it (no need to perform an extra iteration to determine that the next chunk is empty).

core/lib/state/src/rocksdb/mod.rs
```diff
@@ -131,23 +152,157 @@ impl RocksdbStorage {
...
+                write_batch.put_cf(
+                    StateKeeperColumnFamily::State,
+                    Self::ENUM_INDEX_MIGRATION_FINISHED,
+                    &[1u8; 1],
```
slowli, line 243: Nit: This is just `&[1_u8]`.

core/lib/state/src/rocksdb/mod.rs
```diff
@@ -131,23 +152,157 @@ impl RocksdbStorage {
...
+            self.enum_migration_start_from(),
+            self.enum_index_migration_chunk_size > 0,
+        ) else {
+            return;
+        };
+
+        let started_at = Instant::now();
+        let progress_percentage = BigDecimal::from(100)
+            * u256_to_big_decimal(h256_to_u256(start_from))
+            / u256_to_big_decimal(U256::MAX);
```
slowli, line 203: Bikeshedding: IMO, using `BigDecimal` just for this purpose is an overkill. IMO, it would be sufficient to output the `start_from` key (probably formatted as `{start_from:0>64x}`, i.e. 0-padded hex). Alternatively, perform arithmetic in `U512`.

core/lib/state/src/rocksdb/mod.rs
```diff
@@ -131,23 +152,157 @@ impl RocksdbStorage {
...
+
+        let mut write_batch = self.db.new_write_batch();
+        let (keys, values): (Vec<_>, Vec<_>) = self
+            .db
+            .from_iterator_cf(StateKeeperColumnFamily::State, start_from.as_bytes())
+            .take(self.enum_index_migration_chunk_size)
+            .filter_map(|(key, value)| {
+                (key.as_ref() != Self::BLOCK_NUMBER_KEY
+                    && key.as_ref() != Self::ENUM_INDEX_MIGRATION_PROGRESS_KEY
+                    && key.as_ref() != Self::ENUM_INDEX_MIGRATION_FINISHED
```
slowli, line 217: Nit: These checks can be extracted into a method (`is_special_key() -> bool`) and placed close to the constants in order to simplify maintainability.

core/lib/state/src/rocksdb/mod.rs
```diff
@@ -131,23 +152,157 @@ impl RocksdbStorage {
...
+                    && value.len() != 40)
+                    .then(|| (H256::from_slice(&key), H256::from_slice(&value[..32])))
```
slowli, line 219: I'd suggest to create a type for values and encoding / decoding methods for it in order to improve readability:

```rust
#[derive(Debug, Clone, Copy)]
struct StateValue {
    value: H256,
    enum_index: Option<u64>,
}

impl StateValue {
    fn deserialize(bytes: &[u8]) -> Self { /* ... */ }
    fn serialize(&self) -> Vec<u8> { /* ... */ }
}
```

core/lib/state/src/rocksdb/mod.rs
```diff
@@ -131,23 +152,157 @@ impl RocksdbStorage {
...
+                );
+            }
+        }
+        self.db
+            .write(write_batch)
+            .expect("failed to save state data into rocksdb");
+        tracing::info!(
+            "RocksDB enum index migration chunk took {:?}, migrated {} keys",
+            started_at.elapsed(),
+            keys.len()
```
slowli, line 265: Do we want to meter these values (e.g., latency as a histogram and migrated keys as a counter)?

core/lib/state/src/rocksdb/mod.rs
```diff
@@ -131,23 +152,157 @@ impl RocksdbStorage {
...
+    /// Returns storage logs to apply.
+    fn process_transaction_logs(
+        &self,
+        updates: HashMap<StorageKey, H256>,
+    ) -> HashMap<StorageKey, (StorageValue, Option<u64>)> {
+        updates
+            .into_iter()
+            .filter_map(|(key, value)| {
+                if self.read_value_inner(&key).is_some() {
+                    Some((key, (value, self.read_enum_index(&key))))
```
slowli, line 300: Two reads of the same key instead of one (the second read is hot, but still). See also a comment about `StateValue` above.

core/lib/state/src/rocksdb/mod.rs
```diff
@@ -274,11 +437,39 @@ impl RocksdbStorage {
...
+            )
+            .expect("failed to read `ENUM_INDEX_MIGRATION_FINISHED`")
+            .is_some();
+        (!finished).then(|| {
+            self.db
+                .get_cf(
+                    StateKeeperColumnFamily::State,
+                    Self::ENUM_INDEX_MIGRATION_PROGRESS_KEY,
+                )
+                .expect("failed to read `ENUM_INDEX_MIGRATION_FINISHED`")
```
slowli, line 469: Nit: Copy-paste artifact (should be `ENUM_INDEX_MIGRATION_PROGRESS_KEY`).

core/lib/storage/src/db.rs
```diff
@@ -329,6 +329,25 @@ impl<CF: NamedColumnFamily> RocksDB<CF> {
...
+        let cf = self.column_family(cf);
+        self.inner
+            .db
+            .iterator_cf(cf, IteratorMode::From(key_from, Direction::Forward))
+            .map(Result::unwrap)
+            .fuse()
+        // ^ The rocksdb docs say that a raw iterator (which is used by the returned ordinary iterator)
+        // can become invalid "when it reaches the end of its defined range, or when it encounters an error."
+        // We panic on RocksDB errors elsewhere and fuse it to prevent polling after the end of the range.
+        // Thus, `unwrap()` should be safe.
```
slowli, line 349: Bikeshedding: instead of copy-pasting, I'd refer to the above comment: "`unwrap()` is safe for the same reasons as in `prefix_iterator_cf()`".

core/lib/state/src/rocksdb/mod.rs
```diff
@@ -60,17 +64,34 @@ impl NamedColumnFamily for StateKeeperColumnFamily {
 pub struct RocksdbStorage {
     db: RocksDB<StateKeeperColumnFamily>,
     pending_patch: InMemoryStorage,
+    enum_index_migration_chunk_size: usize,
 }
 
 impl RocksdbStorage {
     const BLOCK_NUMBER_KEY: &'static [u8] = b"block_number";
+    const ENUM_INDEX_MIGRATION_PROGRESS_KEY: &'static [u8] = b"enum_index_migration_progress";
```
slowli, line 72: Nit: IMO, `ENUM_INDEX_MIGRATION_CURSOR` is a more straightforward naming.

core/lib/state/src/rocksdb/mod.rs
```diff
@@ -60,17 +64,34 @@ impl NamedColumnFamily for StateKeeperColumnFamily {
...
         let db = RocksDB::new(path, true);
         Self {
             db,
             pending_patch: InMemoryStorage::default(),
+            enum_index_migration_chunk_size: 0,
+        }
+    }
+
+    /// Creates a new storage with the provided RocksDB `path` and enum index migration settings.
+    pub fn new_with_enum_index_migration_chunk_size(
```
slowli, line 86: Why not have a separate method?

```rust
pub fn enable_enum_index_migration(&mut self, chunk_size: usize) {
    self.enum_index_migration_chunk_size = chunk_size;
}
```

Do we want to make this API intentionally awkward to use? 🙃

core/lib/dal/src/storage_logs_dedup_dal.rs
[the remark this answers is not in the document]
slowli: This logic is as simple as an indexing operation (keys are available in both cases `enum_indices_for_keys()` is called, AFAICT). I understand this is in the bikeshedding territory, but IMO making DAL smarter than it should be (i.e., assuming that all requested keys exist in this case) is an anti-pattern.

core/lib/state/src/rocksdb/mod.rs
```diff
@@ -8,18 +8,22 @@
...
-//! | Contracts    | address (20 bytes)     | `Vec<u8>`               | Contract contents                    |
-//! | Factory deps | hash (32 bytes)        | `Vec<u8>`               | Bytecodes for new contracts that a certain contract may deploy. |
-
-use std::{collections::HashMap, mem, path::Path, time::Instant};
+//! | Column       | Key                             | Value                           | Description                               |
+//! | ------------ | ------------------------------- | ------------------------------- | ----------------------------------------- |
+//! | State        | 'block_number'                  | serialized block number         | Last processed L1 batch number (u32)      |
+//! | State        | 'enum_index_migration_progress' | serialized next key to process  | Hashed key migration should continue from |
+//! | State        | 'enum_index_migration_finished' | 1 byte flag                     | Denotes if the migration if completed     |
+//! | State        | hashed `StorageKey`             | 32 bytes value ++ 8 bytes index | State for the given key                   |
```
slowli, line 16: For specificity, you should specify index endianness. Also, why is it big-endian anyway? We don't need to sort by the index (which is where big-endian encoding would work and little-endian wouldn't). Not that it matters much, but most modern CPU architectures use little-endian. Also, if we care about space usage that much, we could use variable-length encoding (e.g., [LEB128](https://en.wikipedia.org/wiki/LEB128) used in Protobuf; it's used in the Merkle tree crate).

slowli (commented):
Ideally, this functionality should be tested. One approach that comes to mind is syncing `RocksdbStorage` from Postgres (you can test enum indices at this point) and then manually removing enum indices for all keys in the state CF (by directly using `RocksDB` operations), so that they can be synced again.

core/lib/state/src/rocksdb/mod.rs
perekopskiy: I don't think we care that much about space usage at the moment. I wouldn't use LEB128 for indices now, probably in the future we can use it for both values and indices. Are there any benefits in using little-endianness without variable-length encoding?

core/lib/state/src/rocksdb/mod.rs
perekopskiy: I think logging them is enough

core/lib/state/src/rocksdb/mod.rs
RomanBrodetski: FWIW both little and big endianness in used in the VM  code - so whatever we chose here we won't be 100% consistent. I'd keep as is

perekopskiy: Well, keys and values are now stored in state keeper's RockDB in big-endian format, so I'd choose big-, unless @slowli has a strong reason to use little-

slowli: No particularly strong opinion; I'm fine with leaving it as is. Regarding values, they *probably* should be interpreted as 32-byte arrays (`H256` in our terms) rather than unsigned integers with specific endianness (`U256` in our terms) because we don't really rely on any arithmetic operations performed on them on the storage level 🙃 (This is outside the scope of this PR, I'm pretty sure.) For keys, using big-endian encoding makes sense because it orders them in the natural order.

core/lib/dal/src/storage_logs_dal.rs
```diff
@@ -295,18 +297,24 @@ impl StorageLogsDal<'_, '_> {
...
 
         let stage_start = Instant::now();
         let prev_values_for_updated_keys = self
             .get_storage_values(&modified_keys, last_miniblock)
-            .await;
+            .await
+            .into_iter()
+            .map(|(key, value)| {
+                let value = value.unwrap(); // We already filtered out keys that weren't touched.
+                let index = l1_batch_and_index_by_key.get(&key).unwrap().1;
```
slowli, line 315: Nit: You can use index access instead of `get(_).unwrap()`.

core/lib/state/src/rocksdb/mod.rs
```diff
@@ -55,22 +60,85 @@ impl NamedColumnFamily for StateKeeperColumnFamily {
...
+        } else {
+            Self {
+                value: H256::from_slice(&bytes[..32]),
+                enum_index: Some(u64::from_be_bytes(bytes[32..40].try_into().unwrap())),
+            }
+        }
+    }
+
+    pub fn serialize(&self) -> Vec<u8> {
+        self.value
```
slowli, line 85: Bikeshedding: IMO, this is less readable than an approach with `extend_from_slice`, but that's just a personal opinion.

```rust
let mut buffer = Vec::with_capacity(40);
buffer.extend_from_slice(self.value.as_bytes());
if let Some(index) = self.enum_index {
    buffer.extend_from_slice(&index.to_be_bytes());
}
buffer
```

core/lib/state/src/rocksdb/mod.rs
```diff
@@ -131,23 +199,162 @@ impl RocksdbStorage {
...
+                    (!new_value.is_zero()).then_some((
+                        key,
+                        StateValue {
+                            value: new_value,
+                            enum_index: None,
+                        },
+                    ))
+                }
+            })
+            .collect()
```
slowli, line 357: Nit: AFAICT, you don't need to collect the iterator here; instead, you could return `impl Iterator` directly (it would have `&self` lifetime, but this looks OK).

core/lib/state/src/rocksdb/mod.rs
```diff
@@ -131,23 +199,162 @@ impl RocksdbStorage {
...
+        tracing::info!(
+            "RocksDB enum index migration is not finished, starting from key {start_from:0>64x}"
+        );
+
+        let mut write_batch = self.db.new_write_batch();
+        let (keys, values): (Vec<_>, Vec<_>) = self
+            .db
+            .from_iterator_cf(StateKeeperColumnFamily::State, start_from.as_bytes())
+            .filter_map(|(key, value)| {
+                (!Self::is_special_key(&key) && value.len() != 40)
```
slowli, line 256: You could deserialize a `StateValue` from `value` bytes after the `is_special_key` check to not duplicate a part of deserialization logic.

core/lib/state/src/rocksdb/mod.rs
```diff
@@ -131,23 +199,162 @@ impl RocksdbStorage {
...
+
+        for (key, value) in keys.iter().zip(values) {
+            let index = enum_indices[key];
+            write_batch.put_cf(
+                StateKeeperColumnFamily::State,
+                key.as_bytes(),
+                &StateValue {
+                    value,
+                    enum_index: Some(index),
+                }
```
slowli, line 275: Nit: Having a constructor for `StateValue` could improve readability in such multi-line statements: `StateValue::new(value, index).serialize()` is arguably easier to read.

core/lib/state/src/rocksdb/mod.rs
slowli: With the new method, you can *replace* this one with

```rust
let mut storage = RocksdbStorage::new(path);
storage.enable_enum_index_migration(chunk_size);
// Do something with `storage`
```

It's more composable this way, and the methods adhere to the single responsibility principle. Sorry for not being clear.

core/lib/state/src/rocksdb/mod.rs
```diff
@@ -8,18 +8,23 @@
...
-//! | State        | hashed `StorageKey`    | 32 bytes value          | State for the given key              |
-//! | Contracts    | address (20 bytes)     | `Vec<u8>`               | Contract contents                    |
-//! | Factory deps | hash (32 bytes)        | `Vec<u8>`               | Bytecodes for new contracts that a certain contract may deploy. |
-
-use std::{collections::HashMap, mem, path::Path, time::Instant};
+//! | Column       | Key                             | Value                           | Description                               |
+//! | ------------ | ------------------------------- | ------------------------------- | ----------------------------------------- |
+//! | State        | 'block_number'                  | serialized block number         | Last processed L1 batch number (u32)      |
+//! | State        | 'enum_index_migration_cursor'   | serialized hashed key           | Hashed key migration should continue from |
+//! | State        | 'enum_index_migration_finished' | 1 byte flag                     | Denotes if the migration if completed     |
```
slowli, line 15: Rly smol nit: Technically, the *existence* of this key denotes that a migration is completed, and the value doesn't matter. (One could think that 0 value means that the migration is not completed, but this is not the case.) On that note, IMO it would be easier to combine cursor with the flag (e.g., use an empty cursor bytes to denote that the migration is complete), but that's just a personal preference.

slowli (commented):
Nice test! 👍 Other than some nits, looks good.

core/lib/state/src/rocksdb/mod.rs
RomanBrodetski: I would also combine but yeah it's very minor

core/lib/state/src/rocksdb/mod.rs
perekopskiy: Nah, it was clear, I just forgot to remove this one

core/lib/state/src/rocksdb/mod.rs
perekopskiy: Agree, it's more readable indeed

slowli (approved):
Other than minor nits, looks good. Please update the PR description; it describes an outdated implementation.

core/lib/dal/src/storage_logs_dedup_dal.rs
```diff
@@ -133,4 +134,21 @@ impl StorageLogsDedupDal<'_, '_> {
         .map(|row| H256::from_slice(&row.hashed_key))
         .collect()
     }
+
+    pub async fn enum_indices_for_keys(&mut self, hashed_keys: &[H256]) -> HashMap<H256, u64> {
```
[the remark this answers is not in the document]
perekopskiy: Hmm, so `get_l1_batches_for_initial_writes/get_l1_batches_and_indices_for_initial_writes/enum_indices_for_keys` do the same job and are different only in the return type (L1BatchNumber / (L1BatchNumber, u32 /*index*/) / u32).

Also, all the places where we use them don't look very performance-critical, i.e. loading `index` and `l1_batch_number` when we need only one of them should be negligible. I think I can leave only `get_l1_batches_and_indices_for_initial_writes`

slowli (approved):

PR #163, opened 2023-10-05
feat(merkle tree): Snapshot recovery for Merkle tree

slowli:
# What ❔

Allows to recover a Merkle tree from a snapshot (collection of tree entries ordered by ascending key). The recovery procedure is fault-tolerant (may be paused and restarted).

## Why ❔

This is one of components for recovering a node from a snapshot.

## Checklist


slowli:
### Recovery timings

Preliminary order of timings with 10M recovered entries (50 chunks x 200k entries / chunk), performed locally on my MacBook using the new `recovery` example in the tree crate:

- Tree traversal (creating new nodes) is 0.18s / chunk.
- Computing hashes is 1.3s / chunk.
- Persisting to RocksDB is 0.2s / chunk.
- Total recovery time is 77.5s.

(I've initially messed up with key distribution for the generated snapshot, so the previous values were incorrect.) Overall, these values are on par with the main tree operation (I've tested it locally as well for this branch and the `main` branch to see if there are perf regressions – doesn't look like that). The only weird thing is that the tree crate readme contains significantly smaller timing for computing hashes (0.7s) for about the same amount of hashed data. Since the tree in `main` still has 1.3s timing for computing hashes, if this is a regression, it's not related to this PR.

### Leaf versions are hard

I've spent quite a bit of time implementing "correct" node version recovery so that after the recovery the tree is identical to the pruned tree at the recovered version, not only in terms of (key, value, enum index) triples, but in terms of versions of each tree node as well. A *slight* problem that I've realized too late is that the tree is the only place we can get these versions from. I've initially thought that a leaf version is equal to the L1 batch number it was last modified in, but this is actually not the case. A leaf may be moved afterwards (caused by a new leaf being inserted adjacent to the leaf), which would increase its version as well. So either we need to get leaf versions from the tree, or we'd need to give up on having version-accurate tree after recovery.

core/lib/merkle_tree/src/recovery.rs
```diff
@@ -0,0 +1,244 @@
+//! Merkle tree recovery logic.
```
popzxc, line 1: Maybe a short overview of the recovery algorithm here would be helpful for the future folks.

popzxc (approved):
I didn't check whether the algorithm is correct, I think a great number of tests you added will do a better job than me here.
The code itself looks good.

popzxc (approved):
:heart:

PR #168, opened 2023-10-06
feat(vm): update pubdata fees calculation

koloz193:
# What ❔

Updated how we charge for pubdata given the new compression strategy.

## Why ❔

Given that state diffs value sizes are now dynamic due to compression, we need to update how we calculate how many bytes to charge the end user for.

## Checklist


core/lib/types/src/storage_writes_deduplicator.rs
```diff
@@ -113,40 +123,65 @@ impl StorageWritesDeduplicator {
...
+                    }
                     updates.push(UpdateItem {
                         key,
                         update_type: UpdateType::Remove(value),
                         is_write_initial,
                     });
                 }
                 (true, Some(new_value)) => {
+                    let mut value_size =
+                        compress_with_best_strategy(log.log_query.read_value, new_value).len();
```
StanislavBreadless, line 140: since the value (in the end) is changed from initial_value -> new_value, it is more correct to:

- Substract from the previous `total_size` the size of the previously modified value there
- Add to the `total_size` the new size of the value

And the `value_size` here should be counted via comparison between `initial_value` and `new_value` and not `log.log_query.read_value` and `new_value`

core/lib/types/src/storage_writes_deduplicator.rs
```diff
@@ -113,40 +123,65 @@ impl StorageWritesDeduplicator {
...
+                        .unwrap_or_else(|| panic!("key {:?} doesn't exist on update", key));
                     updates.push(UpdateItem {
                         key,
-                        update_type: UpdateType::Update(old_value),
+                        update_type: UpdateType::Update(*old_value),
                         is_write_initial,
-                    })
+                    });
+                    if value_size > old_value.size {
+                        value_size = value_size + old_value.size - 1;
```
StanislavBreadless, line 152: sorry, why do you reduce `value_size` here? This reduced value will then go to the `ModifiedSlot`

Maybe I misunderstand something, but it seems that a generally easier logic would be:

```
*total_size -= old_value.size;
*total_size += value_size;
```

and that's it

core/lib/types/src/storage_writes_deduplicator.rs
```diff
@@ -113,40 +123,65 @@ impl StorageWritesDeduplicator {
...
+                        ModifiedSlot {
+                            value: new_value,
+                            tx_index: log.log_query.tx_number_in_block,
+                            size: value_size,
+                        },
+                    );
                 }
                 (false, Some(new_value)) => {
+                    let value_size =
+                        compress_with_best_strategy(log.log_query.read_value, new_value).len();
```
StanislavBreadless, line 170: Same here, need to compare against initial value.

core/lib/vm/src/old_vm/oracles/storage.rs
```diff
@@ -321,18 +325,24 @@ impl<S: WriteStorage, H: HistoryMode> VmStorageOracle for StorageOracle<S, H> {
...
+//   - The second time we publish it, we will use the 4/5 byte representation of this 8-byte instead of the 32
+//     bytes of the entire key.
+// For value compression, we use a metadata byte which holds the length of the value and the operation from the
+// previous state to the new state, and the compressed value. The maxiumum for this is 33 bytes.
+// Total bytes for initial writes then becomes 65 bytes and repeated writes becomes 38 bytes.
+fn get_pubdata_price_bytes(query: &LogQuery, is_initial: bool) -> u32 {
     // TODO (SMA-1702): take into account the content of the log query, i.e. values that contain mostly zeroes
     // should cost less.
+    let compressed_value_size =
+        compress_with_best_strategy(query.read_value, query.written_value).len() as u32;
```
StanislavBreadless, line 341: you should compare not `.read_value` and `.written_value`, but between the value that was at the very start of the batch in this slot and the one that is being written to this slot

for instance, if we incremented by 255 (which is add with 1 byte) and then we incremented once more by 255 (the diff between read_value and written_value will be add with 1 byte also). The `value_update_price` will see that the user has already paid for 1 byte, while actually the user needs to pay for the additional one byte created (since the final diff requires 2 bytes). 

If it is not too hard, I would also suggest creating a test for such case

core/lib/types/src/storage_writes_deduplicator.rs
koloz193: the change above covers this but it was because we were only calculating the diff and need to account for the extra metadata byte counted twice. with your comment above this is no longer needed

core/lib/vm/src/old_vm/oracles/storage.rs
koloz193: instead of that, could we check to see if the `written_value` > `read_value` and charge for the difference? if it's `<=` then we set the cost be `0`

core/lib/config/src/constants/ethereum.rs
```diff
@@ -12,8 +12,8 @@ pub const GUARANTEED_PUBDATA_PER_L1_BATCH: u64 = 4000;
 
 /// The maximum number of pubdata per L1 batch. This limit is due to the fact that the Ethereum
 /// nodes do not accept transactions that have more than 128kb of pubdata.
-/// The 18kb margin is left in case of any inpreciseness of the pubdata calculation.
-pub const MAX_PUBDATA_PER_L1_BATCH: u64 = 120000;
+/// The 128kb margin is left in case of any inpreciseness of the pubdata calculation.
+pub const MAX_PUBDATA_PER_L1_BATCH: u64 = 110000;
```
StanislavBreadless, line 16: 1. Here the margin left is 18kb and not 128kb 
2. Why do you think such change is needed? We should generally be able to send transactions with size up to 128kb. Some margin is needed, but why that large?

core/lib/types/src/storage_writes_deduplicator.rs
```diff
@@ -3,14 +3,18 @@ use std::collections::HashMap;
...
 use crate::{AccountTreeId, StorageKey, StorageLogQuery, StorageLogQueryType, U256};
 
-#[derive(Debug, Clone, Copy, PartialEq)]
+#[derive(Debug, Clone, Copy, PartialEq, Default)]
 pub struct ModifiedSlot {
     /// Value of the slot after modification.
     pub value: U256,
     /// Index (in L1 batch) of the transaction that lastly modified the slot.
     pub tx_index: u16,
+    /// Size of update in bytes
```
StanislavBreadless, line 16: I would also emphasize that it is the size that it will take up in pubdata (it might not be obvious for a person without context)

core/lib/types/src/storage_writes_deduplicator.rs
```diff
@@ -101,10 +105,16 @@ impl StorageWritesDeduplicator {
             };
 
             let is_write_initial = log.log_type == StorageLogQueryType::InitialWrite;
-            let field_to_change = if is_write_initial {
-                &mut self.metrics.initial_storage_writes
+            let (field_to_change, total_size) = if is_write_initial {
+                (
+                    &mut self.metrics.initial_storage_writes,
+                    &mut self.metrics.total_writes_size,
```
StanislavBreadless, line 111: In both `if` and `else` clauses you return the same `&mut self.metrics.total_writes_size`, maybe we could do it without it and simplify the code

core/lib/types/src/storage_writes_deduplicator.rs
```diff
@@ -113,40 +123,59 @@ impl StorageWritesDeduplicator {
                         panic!("tried removing key: {:?} before insertion", key)
                     });
                     *field_to_change -= 1;
+                    *total_size -= value.size;
+                    if is_write_initial {
+                        *total_size -= BYTES_PER_DERIVED_KEY as usize;
```
StanislavBreadless, line 128: I don't have any strong opinion about it, but I think it might make code a bit leaner if instead of `metrics.total_size` we stored `metrics.total_updated_values_size` and in order to obtain the full length we could do:

`total_updated_values_size + BYTES_PER_DERIVED_KEY * initial_storage_writes + BYTES_PER_ENUMERATION_INDEX  * repeated_storage_writes`

This way there wouldn't be as many repeated `if`s as now

core/lib/types/src/storage_writes_deduplicator.rs
```diff
@@ -161,23 +190,49 @@ impl StorageWritesDeduplicator {
 
     fn rollback(&mut self, updates: Vec<UpdateItem>) {
         for item in updates.into_iter().rev() {
-            let field_to_change = if item.is_write_initial {
-                &mut self.metrics.initial_storage_writes
+            let (field_to_change, total_size) = if item.is_write_initial {
+                (
+                    &mut self.metrics.initial_storage_writes,
+                    &mut self.metrics.total_writes_size,
```
StanislavBreadless, line 196: Same here, no in both clausess `&mut self.metrics.total_writes_size,` is returned

core/multivm_deps/vm_virtual_blocks/src/old_vm/oracles/storage.rs
```diff
@@ -46,6 +48,8 @@ pub struct StorageOracle<S: WriteStorage, H: HistoryMode> {
     // to cover this slot.
     // `paid_changes` history is necessary
     pub(crate) paid_changes: HistoryRecorder<HashMap<StorageKey, u32>, HistoryEnabled>,
+
+    initial_values: HistoryRecorder<HashMap<StorageKey, U256>, HistoryEnabled>,
```
StanislavBreadless, line 52: This is `multivm_deps` folder, while you should modify the `vm` folder (where the latest VM resides)

core/multivm_deps/vm_virtual_blocks/src/old_vm/oracles/storage.rs
```diff
@@ -131,6 +137,25 @@ impl<S: WriteStorage, H: HistoryMode> StorageOracle<S, H> {
             .unwrap_or_default()
     }
 
+    fn get_initial_value(&self, storage_key: &StorageKey) -> Option<U256> {
+        self.initial_values.inner().get(storage_key).copied()
+    }
+
+    fn set_initial_value(&mut self, storage_key: &StorageKey, value: U256, timestamp: Timestamp) {
+        match self.get_initial_value(storage_key) {
```
StanislavBreadless, line 145: `if self.get_initial_value(storage_key).is_none()` is a bit shorter

Also, in the `vm` crate there is already a mechanism to store initial values (only recently added to `boojum-integration`), however, it only adds the initial values only during writes, so when using the `vm` crate you should do similarly to how you do in your PR (set initial values even during reads)

core/multivm_deps/vm_virtual_blocks/src/old_vm/oracles/storage.rs
```diff
@@ -260,7 +301,10 @@ impl<S: WriteStorage, H: HistoryMode> VmStorageOracle for StorageOracle<S, H> {
         _monotonic_cycle_counter: u32,
         partial_query: &LogQuery,
     ) -> RefundType {
-        let price_to_pay = self.value_update_price(partial_query);
+        let storage_key = storage_key_of_log(partial_query);
```
StanislavBreadless, line 304: unused field

core/multivm_deps/vm_virtual_blocks/src/old_vm/oracles/storage.rs
```diff
@@ -144,16 +169,26 @@ impl<S: WriteStorage, H: HistoryMode> StorageOracle<S, H> {
...
     // TODO (SMA-1701): update VM to accept gas instead of pubdata.
-    fn value_update_price(&self, query: &LogQuery) -> u32 {
+    fn value_update_price(&mut self, query: &LogQuery) -> u32 {
         let storage_key = storage_key_of_log(query);
 
         let base_cost = self.base_price_for_write(query);
 
+        let initial_value = self
+            .get_initial_value(&storage_key)
+            .unwrap_or(query.read_value);
```
StanislavBreadless, line 188: If the initial value is not present, you should read it from storage. In `zk_evm` the `read_value` provided is always zero. The correct `read_value` is actually set in the `vm` crate: https://github.com/matter-labs/zksync-era/blob/040374900656c854a7b9de32e5dbaf47c1c47889/core/lib/vm/src/oracles/storage.rs#L121

When the `estimate_refunds_for_write` function is called, the `read_value` is not inserted anywhere

core/multivm_deps/vm_virtual_blocks/src/old_vm/oracles/storage.rs
koloz193: it looks like rebasing main changed the modified file

core/lib/types/src/storage_writes_deduplicator.rs
```diff
@@ -107,46 +110,57 @@ impl StorageWritesDeduplicator {
...
+                    *total_size -= value.size;
                     updates.push(UpdateItem {
                         key,
                         update_type: UpdateType::Remove(value),
                         is_write_initial,
                     });
                 }
                 (true, Some(new_value)) => {
+                    let value_size = compress_with_best_strategy(initial_value, new_value).len();
                     let old_value = self
```
StanislavBreadless, line 130: Why there was a need to rewrite this part? I mean, 

```
                    self.modified_key_values.insert(
                        key,
                        ModifiedSlot {
                            value: new_value,
                            tx_index: log.log_query.tx_number_in_block,
                            size: value_size,
                        },
                    );
```

would already return you the old value

core/lib/types/src/storage_writes_deduplicator.rs
```diff
@@ -167,17 +181,29 @@ impl StorageWritesDeduplicator {
...
 
+            let total_size = &mut self.metrics.total_updated_values_size;
+
             match item.update_type {
                 UpdateType::Insert => {
-                    self.modified_key_values.remove(&item.key);
+                    let value = self
+                        .modified_key_values
+                        .remove(&item.key)
+                        .unwrap_or_default();
```
StanislavBreadless, line 191: better to have `unwrap_or_else(|| panic!` like in the function above, since this mapping is never supposed to be empty

core/lib/types/src/storage_writes_deduplicator.rs
```diff
@@ -167,17 +181,29 @@ impl StorageWritesDeduplicator {
...
                     *field_to_change -= 1;
+                    *total_size -= value.size;
                 }
                 UpdateType::Update(value) => {
-                    self.modified_key_values.insert(item.key, value);
+                    let old_value = self
+                        .modified_key_values
+                        .insert(item.key, value)
+                        .unwrap_or_default();
```
StanislavBreadless, line 199: same here regarding `unwrap_or_else`

core/lib/vm/src/oracles/storage.rs
```diff
@@ -88,12 +90,30 @@ impl<S: WriteStorage, H: HistoryMode> StorageOracle<S, H> {
             || *key == storage_key_for_eth_balance(&BOOTLOADER_ADDRESS)
     }
 
+    fn get_initial_value(&self, storage_key: &StorageKey) -> Option<U256> {
+        self.initial_values.inner().get(storage_key).copied()
+    }
+
+    fn set_initial_value(&mut self, storage_key: &StorageKey, value: U256, timestamp: Timestamp) {
+        if self.get_initial_value(storage_key).is_none() {
```
StanislavBreadless, line 98: You can use `insert` function for `HistoryRecorder<HashMap<T>>`

For an example you can search the following snippet:

```
if !self.initial_values.inner().contains_key(&key) {
            self.initial_values
                .insert(key, current_value, query.timestamp);
        }
```

and replace the snippet with `set_initial_value`

core/lib/vm/src/oracles/storage.rs
```diff
@@ -88,12 +90,30 @@ impl<S: WriteStorage, H: HistoryMode> StorageOracle<S, H> {
...
+        }
+    }
+
     pub fn read_value(&mut self, mut query: LogQuery) -> LogQuery {
         let key = triplet_to_storage_key(query.shard_id, query.address, query.key);
         let current_value = self.storage.read_from_storage(&key);
 
         query.read_value = current_value;
 
+        self.set_initial_value(&key, current_value, Timestamp(0));
```
StanislavBreadless, line 115: the timestamp will later be used to rollback the state of the struct, so need to use `query.timestamp`

core/lib/types/src/storage_writes_deduplicator.rs
koloz193: i think it was left over from an intermediate change

core/lib/vm/src/oracles/storage.rs
```diff
@@ -201,7 +215,11 @@ impl<S: WriteStorage, H: HistoryMode> StorageOracle<S, H> {
     fn base_price_for_write_query(&self, query: &LogQuery) -> u32 {
```
StanislavBreadless, line 215: (It is not for this line, but a nit for lines that are not in this PR)

Since you've already added the `set_initial_value` function, you can replace the following snippet:

```
if !self.initial_values.inner().contains_key(&key) {
            self.initial_values
                .insert(key, current_value, query.timestamp);
        }
```

with it.

core/lib/vm/src/oracles/storage.rs
```diff
@@ -313,6 +337,12 @@ impl<S: WriteStorage, H: HistoryMode> VmStorageOracle for StorageOracle<S, H> {
             let to_pay_by_user = self.base_price_for_write_query(&query);
             let prepaid = self.prepaid_for_write(&storage_key);
 
+            let initial_value = self
```
StanislavBreadless, line 340: Inside `self.write_value(query` the initial value is also updated, please remove it from here

StanislavBreadless (approved):
👍 , only small nits are left

PR #173, opened 2023-10-09
fix(db): drop constraint prover_jobs_fri_l1_batch_number_fkey

perekopskiy:
# What ❔

`prover_jobs_fri_l1_batch_number_fkey` is dropped.

## Why ❔

`prover_jobs_fri` is filled only in prover DB, while `l1_batches` -- only in core DB, so we can't have the constraint.

## Checklist


core/lib/dal/migrations/20231009073918_drop-prover_jobs_fri_l1_batch_number_fkey.down.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE prover_jobs_fri ADD CONSTRAINT prover_jobs_fri_l1_batch_number_fkey
+    FOREIGN KEY (l1_batch_number) REFERENCES l1_batches (number);
```
ly0va, line 2: The original constraint also has `ON DELETE CASCADE` but I don't think it's a big deal.

ly0va (approved):

mm-zk (approved):

PR #186, opened 2023-10-10
chore: Add stage2 upgrade data

AnastasiiaVashchuk:


etc/upgrades/1696936385-refunds_enhancement/common.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "refunds_enhancement",
```
perekopskiy, line 2:
```suggestion
  "name": "refunds-enhancement",
```
for consistency with other upgrade names. Also, please replace `_` with `-` in folder name

PR #187, opened 2023-10-11
feat(boojum): FRI prover changes from boojum-integration

ly0va:
# What ❔

`prover/` directory changes from boojum-integration branch

## Why ❔

To make the diff smaller

## Checklist


prover/Cargo.toml
```diff
@@ -5,9 +5,9 @@ members = [
     "prover_fri_types",
 
     # binaries
-    "prover",
+    # "prover",
     "circuit_synthesizer",
-    "setup_key_generator_and_server",
+    # "setup_key_generator_and_server",
```
ly0va, line 10: These are commented out because they do not compile and (AFAIK) are not used.

mm-zk: ack. Then we should remove them in the next PR.

mm-zk (approved):

PR #188, opened 2023-10-11
feat(zk tool): Added yarn & directory checks

mm-zk:
# What ❔

* zk will now print a warning if you run it from outside of $ZKSYNC_HOME directory
* it will also warn you if your yarn has a wrong version

## Why ❔

* we use ZKSYNC_HOME to figure out which binaries to use, so if you are working on couple versions of era in multiple different directories, and forget to update ZKSYNC_HOME, then you might be pushing wrong binaries without knowing.
* if your version of yarn is incorrect, you're going to get a lot of strange typescript compilation errors along the way.

These are only printed as warnings, they don't prevent you from running the tool.

image


bin/zk
```diff
@@ -1,5 +1,37 @@
...
+    if [[ "$CURRENT_DIR_ABS" != "$ZKSYNC_HOME_ABS"* ]]; then
+        echo -e "\e[31mWarning: You are not in a subdirectory of ZKSYNC_HOME ($ZKSYNC_HOME_ABS).\e[0m"
+        return 1
+    fi
+    return 0
+}
+
+# Currently many parts of our zk typescript are checked & verified with yarn v1.22.19 - and might fail with newer versions of yarn.
+check_yarn_version() {
+    desired_version="1.22.19"
```
Deniallugo, line 24: is it possible to use only major an minor version? without patch?

bin/zk
```diff
@@ -1,5 +1,37 @@
 #!/usr/bin/env bash
 
+# checks that the current directory that user is in, is inside the $ZKSYNC_HOME.
```
popzxc, line 3: Maybe it's better to move these checks to the `zk` itself?
I think introducing the split of logic between the `zk` itself and shell script invoking it may be confusing.

mm-zk: I've originally started putting it in zk itself (as writing typescript is a lot easier than shell ;-)

But the problem is, that zk is called 'too late'. For example, if you're using a wrong version of yarn, you'll get all the errors - and you'll not see the message.

similar for directory - all the 'yarn compile' work will run - and then you'll have to re-do it again.

popzxc: Makes sense. This then will make a good comment in the code!

popzxc:
@mm-zk could you please update the PR (especially @Deniallugo 's suggestion -- I believe right now the check is too strict).

popzxc (approved):

PR #192, opened 2023-10-11
feat(fri-prover): In witness - panic if protocol version is not available

mm-zk:
# What ❔

* When running witness in continuous mode, panic when you cannot load the matching protocol version.

## Why ❔

* Witness is running with a given commitment key. At startup, it checks if there is a matching protocol version for this key, and later uses this protocol version to fetch its task.
* We are running witness in 2 modes: either 'batch' (where it runs up to X tasks and quits) or continuous. Batch is mostly used in production, while continuous is used in local deployments.
* And if in the local deployment, the database was not updated - and protocol version is missing - then the job would get stuck forever (as it will try to pickup job for non-existing protocol). So instead, let's panic explicitly.

### DIFF
--- a/prover/witness_generator/src/main.rs
+++ b/prover/witness_generator/src/main.rs
@@ -71,12 +71,11 @@ async fn main() -> anyhow::Result<()> {
     let started_at = Instant::now();
     let use_push_gateway = opt.batch_size.is_some();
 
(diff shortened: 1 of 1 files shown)


ly0va (approved):
LGTM

PR #193, opened 2023-10-11
fix(prover): Fix statistic query

Deniallugo:
# What ❔

Fix extremly expensive query

## Why ❔
It could overload our database

## Checklist

### DIFF
--- a/core/lib/dal/src/prover_dal.rs
+++ b/core/lib/dal/src/prover_dal.rs
@@ -314,6 +314,7 @@ impl ProverDal<'_, '_> {
                 r#"
                 SELECT COUNT(*) as "count!", circuit_type as "circuit_type!", status as "status!"
                 FROM prover_jobs
+                WHERE status <> 'skipped' and status <> 'successful' 
                 GROUP BY circuit_type, status
                 "#
             )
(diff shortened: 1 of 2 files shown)


RomanBrodetski (approved):

core/lib/dal/src/prover_dal.rs
```diff
@@ -314,6 +314,7 @@ impl ProverDal<'_, '_> {
                 r#"
                 SELECT COUNT(*) as "count!", circuit_type as "circuit_type!", status as "status!"
                 FROM prover_jobs
+                WHERE status <> 'skipped' and status <> 'successful'
```
perekopskiy, line 317: I'm not sure it will improve things much. Instead, I suggest adding `WHERE l1_batch_number >= $last_proven_batch`

Deniallugo: I don't think it's better and it will take a bit more time to implement. 
we don't have this value at the start and now we can. 
Seems they are eventually close to each other. Let's keep this fix

RomanBrodetski: I'm also not sure if this will help, but Serghei thought it will. I'm also thinking `WHERE l1_batch_number >= $last_proven_batch` is better though - even if we need an additional query to determine it

PR #194, opened 2023-10-11
feat(types): introduce state diff record type and compression

koloz193:
# What ❔

Introduces the `StateDiffRecord` type and advanced compression for this type.

## Why ❔

In boojum, all storage writes will be replaced with `StateDiffRecord` types. In order to cut down on the number of bytes published to L1 we want to compress this data as much as possible.

## Checklist


core/lib/types/src/storage/writes/compression.rs
```diff
@@ -0,0 +1,434 @@
...
+// output size, what type of operation was performed, and value/extended compression.
+trait CompressionMode: 'static {
+    /// Id of the operation being performed.
+    fn operation_id(&self) -> usize;
+    /// Number of bytes the compressed value requires.
+    fn output_size(&self) -> Option<usize>;
+    /// Compress the value.
+    fn compress_value_only(&self) -> Option<Vec<u8>>;
+    /// Concatenation of the metadata byte (5 bits for len and 3 bits for operation type) and the compressed value.
+    fn compress_extended(&self) -> Option<Vec<u8>>;
```
RomanBrodetski, line 13: not that important but perhaps would be cleaner if we added the metadata byte on the caller side. 

That is, we'd have a `find_best_compressor` method that would return an instance of `CompressionMode`, and the caller would invoke `.operation_id()` on it to detemine the metadata byte. Also, we wouldn't need to duplicate the somewhat convoluted `((self.output_size().unwrap() << 3)` expression

core/lib/types/src/storage/writes/mod.rs
```diff
@@ -0,0 +1,499 @@
...
+    pub key: U256,
+    /// derived_key == Blake2s(bytes32(address), key)
+    pub derived_key: [u8; 32],
+    /// index in tree of state diff
+    pub enumeration_index: u64,
+    /// previous value
+    pub initial_value: U256,
+    /// updated value
+    pub final_value: U256,
+}
```
RomanBrodetski, line 48: hmm so I think we have some structrs that represent this, let's think if we can reuse (`LogQuery` )

koloz193: i think it would be possible but it might not be the best idea given that most fields would not be used, we'd need to introduce some fields for this particular usage, and logically a StateDiff is the overall storage update (can be a combination of `LogQuery`)

core/lib/types/src/storage/writes/compression.rs
RomanBrodetski: I see you resolved it - did you consider doing this? I don't insist, just wanted to make sure you didn't overloop this comment

core/lib/types/src/storage/writes/compression.rs
```diff
@@ -0,0 +1,446 @@
...
+        .filter(|e| e.output_size().is_some())
+        .min_by(|x, y| {
+            let x_size = x.output_size();
+            let y_size = y.output_size();
+
+            x_size.cmp(&y_size)
+        });
+
+    match compressor {
+        None => CompressionByteNone::new(new_value)
```
RomanBrodetski, line 255: in theory you can get rid of this logic if you add `CompressionByteNone` to the `default` list. You'd need to do an `unwrtap` there though since you don't have information that `CompressionByteNone` never returns `None` on the type level - so ultimately up to you

core/lib/types/src/storage/writes/mod.rs
RomanBrodetski: will quickly ping @perekopskiy here - what do you think, is a new struct justified here? I personally don't have a strong opionoin on that

core/lib/types/src/storage/writes/compression.rs
koloz193: yup, I kept the original code but within each I abstracted out metadata byte creation to a function. I agree that overall it looks cleaner and provides 1 place to change rather than `N`

core/lib/types/src/storage/writes/compression.rs
koloz193: i did make an attempt to do `unwrap_or_default` with no compression as the default but it started to reach the bounds of my rust knowledge as there were complaints from the compiler about typing (`CompressionByteNone` vs `dyn CompressionMode`)

core/lib/types/src/storage/writes/compression.rs
```diff
@@ -0,0 +1,446 @@
...
+            .compress_extended()
+            .unwrap(),
+        Some(compressor) => compressor.compress_extended().unwrap(),
+    }
```
perekopskiy, line 259:
```suggestion
    compressors
        .iter()
        .filter_map(|e| e.compress_extended())
        .min_by_key(|bytes| bytes.len())
        .unwrap_or_else(|| {
            CompressionByteNone::new(new_value)
                .compress_extended()
                .unwrap()
        })
```

core/lib/types/src/storage/writes/compression.rs
```diff
@@ -0,0 +1,446 @@
...
+
+        let diff = buffer[(32 - size)..].to_vec();
+
+        Some(diff)
+    }
+
+    fn compress_extended(&self) -> Option<Vec<u8>> {
+        match self.compress_value_only() {
+            None => None,
+            Some(compressed_val) => {
```
perekopskiy, line 59: Nit: you can use `Option::map` instead of match

core/lib/types/src/storage/writes/mod.rs
perekopskiy: I support having a new struct

core/lib/types/src/storage/writes/mod.rs
```diff
@@ -0,0 +1,499 @@
...
+        comp_state_diff
+    }
+}
+
+/// Compresses a vector of state diff records according to the following:
+/// num_initial writes (u32) || compressed initial writes || compressed repeated writes
+pub fn compress_state_diffs(mut state_diffs: Vec<StateDiffRecord>) -> Vec<u8> {
+    let mut res = vec![];
+
+    state_diffs.sort_by_key(|rec| (rec.address, rec.key));
```
mm-zk, line 140: AFAIK this sorting order is very important, as it has to match something in the circuits, right?
(might be worth mentioning it)

core/lib/types/src/storage/writes/mod.rs
```diff
@@ -0,0 +1,499 @@
...
+    }
+
+    for state_diff in repeated_writes {
+        res.extend(state_diff.compress());
+    }
+
+    prepend_header(res)
+}
+
+fn prepend_header(compressed_state_diffs: Vec<u8>) -> Vec<u8> {
```
mm-zk, line 158: comment

core/lib/types/src/storage/writes/mod.rs
```diff
@@ -0,0 +1,499 @@
...
+    }
+
+    /// compression follows the following algo:
+    /// 1. if repeated write:
+    ///      entry <- enumeration_index || value
+    /// 2. if initial write:
+    ///      entry <- blake2(bytes32(address), key) || value
+    /// size:
+    ///      initial:  64 bytes
+    ///      repeated: 40 bytes
```
mm-zk, line 115: is it 40 bytes or 36 bytes now?

(as we use U32::max ==> 4 bytes)

core/lib/types/src/storage/writes/mod.rs
```diff
@@ -0,0 +1,499 @@
...
+    ///      before:  156 bytes for each
+    pub fn compress(&self) -> Vec<u8> {
+        let mut comp_state_diff = match self.enumeration_index {
+            0 => self.derived_key.to_vec(),
+            enumeration_index if enumeration_index < u32::MAX.into() => {
+                (self.enumeration_index as u32).to_be_bytes().to_vec()
```
mm-zk, line 121: I wonder if we should check that this matches BYTES_PER_ENUMERATION_INDEX (some debug_assert?)

core/lib/types/src/storage/writes/mod.rs
```diff
@@ -0,0 +1,499 @@
...
+    /// 2. if initial write:
+    ///      entry <- blake2(bytes32(address), key) || value
+    /// size:
+    ///      initial:  64 bytes
+    ///      repeated: 40 bytes
+    ///      before:  156 bytes for each
+    pub fn compress(&self) -> Vec<u8> {
+        let mut comp_state_diff = match self.enumeration_index {
+            0 => self.derived_key.to_vec(),
+            enumeration_index if enumeration_index < u32::MAX.into() => {
```
perekopskiy, line 120: Shouldn't it be `<=` here?

core/lib/types/src/storage/writes/mod.rs
```diff
@@ -0,0 +1,499 @@
...
+
+use crate::H256;
+use serde::{Deserialize, Serialize};
+use zksync_basic_types::{Address, U256};
+
+use self::compression::compress_with_best_strategy;
+
+mod compression;
+
+const COMPRESSION_VERSION_NUMBER: u8 = 1;
```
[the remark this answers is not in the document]
koloz193: agreed, going to move this into the compression file so its closer to where the changes that require it to be incremented are

core/lib/types/src/storage/writes/compression.rs
koloz193: neat suggestion, will add it in

core/lib/types/src/storage/writes/mod.rs
```diff
@@ -0,0 +1,499 @@
...
+const BYTES_PER_ENUMERATION_INDEX: u8 = 4;
+const STATE_DIFF_RECORD_SIZE: usize = 156;
+
+// 2 * 136 - the size that allows for two keccak rounds.
+pub const PADDED_ENCODED_STORAGE_DIFF_LEN_BYTES: usize = 272;
+
+/// In vm there are two types of writes Initial and Repeated. After the first write to the leaf,
```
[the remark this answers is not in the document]
koloz193: should be key, the leaf number just determines the enum index

core/lib/types/src/storage/writes/mod.rs
koloz193: its 38, we still need to add in automated switch over from 4 byte enum index to 5 byte enum index once we get closer to that value

core/lib/types/src/storage/writes/mod.rs
```diff
@@ -0,0 +1,509 @@
...
+    }
+
+    /// compression follows the following algo:
+    /// 1. if repeated write:
+    ///      entry <- enumeration_index || compressed value
+    /// 2. if initial write:
+    ///      entry <- blake2(bytes32(address), key) || compressed value
+    /// size:
+    ///      initial:  max of 65 bytes
+    ///      repeated: max of 38 bytes
```
mm-zk, line 121: why 38? 

4 for enumeration index, + 33 for max compression?

core/lib/types/src/storage/writes/compression.rs
```diff
@@ -0,0 +1,409 @@
...
+/// If none are found then use the full 32 byte new value with the metadata byte being `0x00`
+pub fn compress_with_best_strategy(prev_value: U256, new_value: U256) -> Vec<u8> {
+    let compressors = default_passes(prev_value, new_value);
+
+    compressors
+        .iter()
+        .filter_map(|e| e.compress_extended())
+        .min_by_key(|bytes| bytes.len())
+        .unwrap_or_else(|| {
+            CompressionByteNone::new(new_value)
```
mm-zk, line 220: we could probably add CompressionByteNone as the last entry to the default_passes. Then you don't need to treat it as a special case.

core/lib/types/src/storage/writes/compression.rs
```diff
@@ -0,0 +1,409 @@
+use super::*;
```
mm-zk, line 1: do we need this? 
(I don't like the '*' inclusions, as it can be quite confusing on where the types are coming from)

mm-zk (dismissed):

core/lib/types/src/storage/writes/compression.rs
```diff
@@ -0,0 +1,409 @@
...
+
+        let mut buffer = [0u8; 32];
+        diff.to_big_endian(&mut buffer);
+
+        let diff = buffer[(32 - size)..].to_vec();
+
+        Some(diff)
+    }
+
+    fn compress_extended(&self) -> Option<Vec<u8>> {
```
perekopskiy, line 81: No need to overwrite `compress_extended` here

core/lib/types/src/storage/writes/compression.rs
perekopskiy: +1 for this

RomanBrodetski: was also thinking of this - the downside would be that we'd need to unwrap as `CompressionByteNone` returns `Some`

perekopskiy: but it's already unwrapped

core/lib/types/src/storage/writes/mod.rs
koloz193: at some point in the future we need to update to 5 byte enumeration index. this way we won't have to update this again

core/lib/types/src/storage/writes/compression.rs
koloz193: not in this scenario, `min_by_key` returns an `Option<Vec<u8>>`. I personally like the separation, `CompressionByteNone` feels less like a compression strategy and more of a default as there's no gain savings

core/lib/types/src/storage/writes/compression.rs
perekopskiy: @koloz193 will you apply this?

mm-zk (approved):

core/lib/types/src/storage/writes/compression.rs
koloz193: yup!

koloz193: just pushed an update with it

perekopskiy: hmm, I don't see it

koloz193: sorry `zk fmt` issues, should be reflected now

perekopskiy (approved):

PR #203, opened 2023-10-12
feat: Use 1.4.0 in vm crate

StanislavBreadless:
# What ❔

Use v1.4.0 in the `vm` crate

## Why ❔

To have a more correct version of `vm` for the boojum integration.

Note, that I did not change `types` crate much. It means that a lot of default behavior is taken from the old version of VM, which is still okay, since:
- Most of the types are shared between these zk_evm versions by `zk_evm_abstractions`.
- The logic is equivalent. Most of the difference is inside the `prover` folder.  

## Checklist


mm-zk (approved):

mm-zk:
Please update the PR description

PR #209, opened 2023-10-12
feat(merkle tree): Expose Merkle tree API

slowli:
# What ❔

Expose API allowing to query Merkle proofs for a list of hashed keys at a specific tree version (= L1 batch number).

## Why ❔

- This is one of components for recovering a node from a snapshot; it will be used to authenticate snapshot chunks.
- In the future, it could be used for `eth_getProof`.

## Checklist


core/lib/zksync_core/src/metadata_calculator/api.rs
```diff
@@ -0,0 +1,215 @@
+//! Primitive Merkle tree API used internally to fetch proofs.
+
+use anyhow::Context as _;
+use async_trait::async_trait;
+use axum::{
```
slowli, line 5: I've implemented HTTP API using `axum` since it seems reasonably lightweight. We could try using something more performant (gRPC with `prost` / `tonic`?), but since it's not used in the codebase and this is intended to be internal API, it looks like an overkill.

core/lib/zksync_core/src/metadata_calculator/api.rs
```diff
@@ -0,0 +1,215 @@
...
+        Ok(response.entries)
+    }
+}
+
+impl AsyncTreeReader {
+    async fn get_proofs(
+        State(mut this): State<Self>,
+        Json(request): Json<TreeProofsRequest>,
+    ) -> Result<Json<TreeProofsResponse>, TreeApiError> {
+        // FIXME: metrics
```
slowli, line 142: Waiting for #120 to be merged.

core/lib/zksync_core/src/metadata_calculator/mod.rs
```diff
@@ -123,6 +125,27 @@ impl MetadataCalculator {
         self.health_updater.subscribe()
     }
 
+    /// Runs the Merkle tree API server on the specified address.
+    pub fn run_api_server(
```
popzxc, line 129: Thought: it is a bit odd to see `MetadataCalculator` to be responsible for creating an HTTP server. Probably, I would've expected some way to get a DB snapshot (or a Merkle Tree instantiated over a snapshot) from the MetadataCalculator, and see _an externally created_ server utilize this API of MetadataCalculator. I.e. metadata calculator just owns the tree, but not much more.

That being said, I don't necessarily propose to rewrite the code now, just thought that it's worth mentioning.

popzxc (approved):
Generally looks good, though I'm not in context regarding the use cases of specialized server for merkle proofs. Trusting you on this one :)

popzxc (approved):

PR #213, opened 2023-10-12
refactor: made unit tests use separate real dbs

pompon0:
# What ❔
Made unit tests use separate real dbs created in a temporary postgres container.

## Why ❔
It makes the tests more realistic and greatly simplifies the dal crate.


core/lib/dal/src/connection/mod.rs
```diff
@@ -56,46 +55,80 @@ impl ConnectionPoolBuilder {
...
+    let db_name = db_url.path()[1..].to_string();
+    let db_copy_name = format!("{PREFIX}{}", rand::thread_rng().gen::<u64>());
+    db_url.set_path("");
+    let mut conn = loop {
+        if let Ok(conn) = sqlx::PgConnection::connect(db_url.as_ref()).await {
+            break conn;
+        } else {
+            tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+        }
+    };
```
slowli, line 108: Nit: It could make sense to limit the number of attempts and panic with the last error message when they are exhausted. Then again, depends on whether we assume that test DB URL is always valid.

slowli (commented):
DB tests look significantly nicer than before 👍

core/lib/zksync_core/src/lib.rs
```diff
@@ -1213,6 +1213,7 @@ async fn circuit_breakers_for_components(
 
 #[tokio::test]
 async fn test_house_keeper_components_get_added() {
+    zksync_dal::set_database_url_to_test_db().await;
```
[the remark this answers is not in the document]
pompon0: done

core/lib/dal/src/connection/mod.rs
```diff
@@ -56,46 +55,80 @@ impl ConnectionPoolBuilder {
...
+    db_url.set_path(&db_copy_name);
+    db_url
+}
+
+#[derive(Clone)]
+pub struct ConnectionPool(pub(crate) PgPool);
+
+impl fmt::Debug for ConnectionPool {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "connection pool")
```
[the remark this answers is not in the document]
pompon0: done

core/lib/dal/src/connection/mod.rs
```diff
@@ -56,46 +55,80 @@ impl ConnectionPoolBuilder {
...
-    Real(PgPool),
-    Test(TestPool),
+const PREFIX: &str = "test-";
+
+pub(super) async fn create_test_db() -> url::Url {
+    use rand::Rng as _;
+    use sqlx::{Connection as _, Executor as _};
+    let db_url = crate::get_test_database_url().unwrap();
+    let mut db_url = url::Url::parse(&db_url).unwrap();
+    let db_name = db_url.path()[1..].to_string();
```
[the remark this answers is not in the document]
pompon0: done

core/lib/dal/src/connection/mod.rs
```diff
@@ -56,46 +55,80 @@ impl ConnectionPoolBuilder {
...
-pub enum ConnectionPool {
-    Real(PgPool),
-    Test(TestPool),
+const PREFIX: &str = "test-";
+
+pub(super) async fn create_test_db() -> url::Url {
+    use rand::Rng as _;
+    use sqlx::{Connection as _, Executor as _};
+    let db_url = crate::get_test_database_url().unwrap();
+    let mut db_url = url::Url::parse(&db_url).unwrap();
```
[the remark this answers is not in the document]
pompon0: done

core/lib/dal/src/connection/mod.rs
pompon0: done

.github/workflows/vm-perf-comparison.yml
```diff
@@ -38,21 +38,28 @@ jobs:
...
-          docker-compose -f docker-compose-runner.yml up -d zk
+          docker compose -f docker-compose-runner.yml up -d zk
 
       - name: run benchmarks on base branch
         shell: bash
         run: |
           ci_run zk
           ci_run zk compiler system-contracts
-          ci_run cargo bench --package vm-benchmark --bench iai | tee base-iai
+          # ci_run cargo bench --package vm-benchmark --bench iai | tee base-iai
```
popzxc, line 48: Commented line

.github/workflows/vm-perf-comparison.yml
```diff
@@ -38,21 +38,28 @@ jobs:
 
       - name: init
         run: |
-          docker-compose -f docker-compose-runner.yml up -d zk
+          docker compose -f docker-compose-runner.yml up -d zk
```
popzxc, line 41: Out of curiosity: is this required for the needs of the PR?

core/lib/dal/src/connection/mod.rs
```diff
@@ -56,46 +55,93 @@ impl ConnectionPoolBuilder {
...
-        ConnectionPool::Real(pool)
+        Ok(ConnectionPool(pool))
     }
 }
 
-#[derive(Debug, Clone)]
-pub enum ConnectionPool {
-    Real(PgPool),
-    Test(TestPool),
+pub(super) async fn create_test_db() -> anyhow::Result<url::Url> {
```
popzxc, line 95: Could you please add a doc-comment here explaining how "the magic" works? I think it would help someone in X months trying to understand how DB tests actually work.

docker-compose-runner.yml
```diff
@@ -34,3 +34,10 @@ services:
       - "5432:5432"
     environment:
       - POSTGRES_HOST_AUTH_METHOD=trust
+  postgres_tmp:
```
popzxc, line 37: Maybe something like `postgres_for_unit_tests` would be better? Esp with a comment like "An instance of postgres configured to execute Rust unit-tests, tuned for performance".

core/lib/dal/src/connection/mod.rs
popzxc: ...also I think it's a good place to document the whole flow (e.g. mention that `fsync=off` is required for performance and that we create a copy of the database using a fully-migrated template).

popzxc (approved):
A few nits, but it's clearly an improvement! Well done.

.github/workflows/vm-perf-comparison.yml
pompon0: I suppose it is not, it is required in other places (there is no docker-compose within the zk container) so I've changed that here for consistency.

core/lib/dal/src/connection/mod.rs
pompon0: done

docker-compose-runner.yml
pompon0: done

.github/workflows/vm-perf-comparison.yml
hatemosphere: it's definitely not "consistency", we have "docker-compose" in dozens of other places. if you need to call compose from compose that is itself running in docker, there's definitely something monumentally wrong going on and we need to fix this

hatemosphere (changes requested):

.github/workflows/vm-perf-comparison.yml
pompon0: We have the zk container from which various operations using zk are performed.
When calling "zk test rust" I need to recreate a postgres container that the database-interacting unit tests will use. Ideologically the lifetime of this container should be limited to "zk test rust" execution, therefore it would be better to run a nested docker instead, but I suppose that is not supported.

> there's definitely something monumentally wrong going on

what is wrong with managing containers from other containers? AFAIK kubernetes does exactly that.

.github/workflows/vm-perf-comparison.yml
pompon0: reverted

hatemosphere (approved):

popzxc (approved):

PR #219, opened 2023-10-13
feat: Add new commitments

perekopskiy:
# What ❔

- `events_queue_commitment` and `bootloader_initial_content_commitment` are calculated in full tree and saved into DB.
- nightly rust is used
- lints found by clippy are fixed

## Why ❔

Preparation for boojum upgrade

## Checklist


core/lib/types/src/block.rs
```diff
@@ -62,6 +62,12 @@ pub struct L1BatchHeader {
     pub base_system_contracts_hashes: BaseSystemContractsHashes,
     /// Version of protocol used for the L1 batch.
     pub protocol_version: Option<ProtocolVersionId>,
+    /// The commitment to the final events queue state after the batch is committed.
+    /// Practically, it is a commitment to all events that happened on L2 during the batch execution.
```
mm-zk, line 66: love the comments - thank you.

perekopskiy: I copied them from `boojum-integration` branch 😅

core/lib/zksync_core/Cargo.toml
```diff
@@ -82,6 +82,8 @@ actix-web = "4.0.0-beta.8"
 
 tracing = "0.1.26"
 
+zkevm_test_harness = { git = "https://github.com/matter-labs/era-zkevm_test_harness.git", branch = "v1.4.0" }
```
Deniallugo, line 85: we can't afford this dependency

core/lib/dal/src/blocks_dal.rs
```diff
@@ -871,7 +918,8 @@ impl BlocksDal<'_, '_> {
                     l2_l1_compressed_messages, l2_l1_merkle_root, l1_gas_price, l2_fair_gas_price, \
                     rollup_last_leaf_index, zkporter_is_available, bootloader_code_hash, \
                     default_aa_code_hash, base_fee_per_gas, aux_data_hash, pass_through_data_hash, \
-                    meta_parameters_hash, protocol_version \
+                    meta_parameters_hash, protocol_version, \
```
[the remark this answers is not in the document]
perekopskiy: It will be an even bigger mess if we have a table with only these two new fields. I'd like to separate all commitment fields after the upgrade.

core/lib/zksync_core/src/metadata_calculator/updater.rs
```diff
@@ -133,17 +137,63 @@ impl TreeUpdater {
...
                 &metadata.initial_writes,
             )
             .await;
-            let metadata = MetadataCalculator::build_l1_batch_metadata(metadata, &header);
-            prepare_results_latency.observe();
+            check_consistency_latency.observe();
+
+            let (events_queue_commitment, bootloader_initial_content_commitment) =
+                if self.mode == MerkleTreeMode::Full {
+                    let events_queue_commitment_latency =
```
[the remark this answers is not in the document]
perekopskiy: I don't think we need it, we calculate commitments for all protocol versions, they just won't be sent to L1 for old ones.

core/lib/zksync_core/src/metadata_calculator/updater.rs
```diff
@@ -5,18 +5,22 @@ use tokio::sync::watch;
 
 use std::{ops, time::Instant};
 
+use zkevm_test_harness::witness::utils::{
+    events_queue_commitment_fixed, initial_heap_content_commitment_fixed,
+};
```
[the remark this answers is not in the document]
perekopskiy: Why move into VM? It's not something VM internal

core/lib/zksync_core/Cargo.toml
perekopskiy: Do you have an alternative?

core/lib/zksync_core/src/metadata_calculator/helpers.rs
```diff
@@ -231,11 +231,12 @@ impl L1BatchWithLogs {
...
         let hashed_keys_for_zero_values: Vec<_> = touched_slots
             .iter()
-            .filter_map(|(key, value)| {
+            .filter(|(_, value)| {
                 // Only zero values are worth checking for initial writes; non-zero values are always
                 // written per deduplication rules.
-                value.is_zero().then(|| key.hashed_key())
+                value.is_zero()
             })
+            .map(|(key, _value)| key.hashed_key())
```
StanislavBreadless, line 239: Small nit

```suggestion
            .map(|(key, _)| key.hashed_key())
```

core/lib/zksync_core/src/metadata_calculator/updater.rs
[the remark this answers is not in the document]
StanislavBreadless: It is maybe VM internal in the sense that if we have a future upgrade, the commitment may change, though maybe similar to how eth_sender's functionality to ABI-encode the `commitBatches` transaction is not part of the VM, this one shouldn't be either

core/lib/zksync_core/src/metadata_calculator/updater.rs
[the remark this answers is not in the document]
StanislavBreadless: What will happen if we have a future upgrade where the `events_queue_commitment_fixed` function changes? Won't `metadata_calculator` calculate the false commitment? Though maybe it is outside of the scope of this PR

core/lib/zksync_core/src/state_keeper/io/seal_logic.rs
```diff
@@ -130,12 +130,15 @@ impl UpdatesManager {
 
         let initial_bootloader_contents =
             finished_batch.final_bootloader_memory.unwrap_or_default();
+
+        let events_queue = Vec::new(); // TODO: put actual value when new VM is merged.
```
StanislavBreadless, line 134: Also, if we want to reduce the diff for the final PR we can integrate returning of the events_queue with the old VM (it will change the interface of the VM, but it won't change its behaviour, so should be fine)

core/lib/zksync_core/src/metadata_calculator/updater.rs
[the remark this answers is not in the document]
perekopskiy: The plan is to eventually move this part to "commitment generation" component and we can make it dependent on protocol version

core/lib/zksync_core/src/metadata_calculator/updater.rs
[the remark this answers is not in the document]
perekopskiy: Discussed with @Deniallugo, we decided to wrap this commitment calc with a separate crate and make it dependent on protocol version

core/lib/zksync_core/src/state_keeper/io/seal_logic.rs
perekopskiy: I think it's not worth it. When we merge new VM version we will just glue old VM result types into new type, with `events_queue: Vec::new()` the same way it's done currently in `boojum-integration` and it won't require change to old vm versions' implementation. If I do it now then I need to change at least the current latest VM version impl.

StanislavBreadless: makes sense

core/lib/zksync_core/src/metadata_calculator/updater.rs
```diff
@@ -146,40 +143,42 @@ impl TreeUpdater {
...
+                );
+                bootloader_commitment_latency.observe();
+
+                (
+                    events_queue_commitment,
+                    bootloader_initial_content_commitment,
+                )
+            } else {
+                (None, None)
+            };
```
Deniallugo, line 181: does it make sense to move the whole logic to the commitment_utils?

core/lib/zksync_core/src/state_keeper/io/seal_logic.rs
StanislavBreadless: @perekopskiy Maybe you could introduce `events_queue: Vec::new()` in all the VM types? Again, as I understand one of the main purposes of this PR is to reduce the diff of `main` compared to the final `boojum-integration` branch. I mean just provide empty vector for the sake of the needed returned type

core/lib/zksync_core/src/metadata_calculator/updater.rs
perekopskiy: hmm, I don't think it's a good idea for `commitment_utils` to depend on DAL. I'd rather extract this part to a separate function

Deniallugo: sorry, forget about it..

core/lib/zksync_core/src/state_keeper/io/seal_logic.rs
perekopskiy: Yes, the main purpose is to reduce the diff. But I don't want to introduce code that won't be needed when the complete boojum integration is merged. We don't need this field in old VM types but only in a new one, which is not in main yet. Also, it will be confusing to have a field that will be always empty

docker/server-v2/Dockerfile
```diff
@@ -1,10 +1,22 @@
 # Will work locally only after prior contracts build
+# syntax=docker/dockerfile:experimental
+FROM debian:bookworm-slim as builder
 
-FROM rust:1.72-bookworm as builder
+WORKDIR /usr/src/zksync
+COPY . .
```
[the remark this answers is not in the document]
perekopskiy: Good catch! I incorrectly applied diff from `boojum-integration` branch

Deniallugo (approved):

PR #225, opened 2023-10-15
chore: update pull-request template

hyunchel:
# What ❔

This PR updates pull-request template to use uniform heading levels.

## Why ❔

Heading level 1 on "What ❔" seems to be a mistake. It should have heading level 2 like the rest of headings.


.markdownlintignore
```diff
@@ -1,2 +1,5 @@
 # Ignore submodule
 bellman-cuda
+
+# Ignore Github configuration markdowns
+.github
```
popzxc, line 5: I think it would be better to ignore the particular line rather than ignore a whole directory. An example could be found in the [configuration section of the markdownlint project](https://github.com/DavidAnson/markdownlint/blob/main/README.md#configuration).

yarn.lock
```diff
@@ -405,6 +405,20 @@
     patch-package "^6.2.2"
     postinstall-postinstall "^2.1.0"
 
+"@ethereumjs/rlp@^4.0.1":
```
popzxc, line 408: I guess this file was modified by mistake, please revert changes here.

popzxc (commented):
Thanks! The change makes sense, but I have a couple of comments.
Also, sorry for the long review -- we're getting used to developing in the open.

.markdownlintignore
hyunchel: I agree with your approach. The problem with this approach, however, is that any comments in `.github` file are treated as a raw Github template. If we were to add such comment that ignores linting on `.github/pull_request_template.md`, Github PR will be opened with that comment in the template. 
Instead of having users manually delete them, or having those lines be included in PRs, I decided to mark the entire `.github` directory as an exception.

hyunchel:
Thanks for the review! I did my first attempt to resolve your comments. Please let me know if any changes are still required after my comments - I will gladly fix them.

popzxc (approved):
Thank you for contribution!

popzxc (approved):

PR #236, opened 2023-10-16
feat(reorg_detector): compare miniblock hashes for reorg detection

montekki:
# What ❔

This PR adds logic to compare the L2 miniblock hashes to `reorg_detector` logic.

## Why ❔

Currently `reorg_detector` only considers the L1 batches and their root hashes which
in case of a reorg on the miniblock level leaves such reorg undetected and nodes stuck.


core/lib/zksync_core/src/reorg_detector/mod.rs
```diff
@@ -109,25 +140,53 @@ impl ReorgDetector {
...
         // quicker than the main node. So, we allow us to be at most one batch ahead of the main node.
         // If the gap is bigger, it's certainly a reorg.
         // Allowing the gap is safe: if reorg has happened, it'll be detected anyway in the future iterations.
-        Ok(main_node_l1_batch_root_hash.is_none() && en_ahead_for <= Some(1))
+        Ok(
+            sealed_miniblock_number.0 as u64 >= last_main_node_miniblock.as_u64()
+                && main_node_miniblock_hash == local_miniblock_hash
+                && main_node_l1_batch_root_hash.is_none()
+                && en_ahead_for <= Some(1),
+        )
```
perekopskiy, line 189: I think we can leave the only check here: `Ok(main_node_miniblock_hash.is_none() && en_ahead_for_miniblocks <= Some(1))`.

core/lib/zksync_core/src/reorg_detector/mod.rs
```diff
@@ -156,9 +225,13 @@ impl ReorgDetector {
...
+            if self.root_hashes_match(sealed_l1_batch_number).await?
+                && self.miniblock_hashes_match(sealed_miniblock_number).await?
+            {
                 EN_METRICS.last_correct_batch[&CheckerComponent::ReorgDetector]
                     .set(sealed_l1_batch_number.0.into());
+                EN_METRICS.last_correct_miniblock[&CheckerComponent::ReorgDetector]
+                    .set(sealed_miniblock_number.0.into());
                 tokio::time::sleep(SLEEP_INTERVAL).await;
             } else {
                 tracing::warn!(
```
perekopskiy, line 237: I'd prefer to log more precise messages, i.e log whether L1 batch or miniblock hash is diverging

core/lib/zksync_core/src/reorg_detector/mod.rs
montekki: Can you elaborate a bit? Bc to me this is not obvious bc why would the miniblock hash be none in this case for example?

perekopskiy: I mean EN can only be legally ahead for 1 miniblock, at least that's my understanding, @Deniallugo please confirm. Previously we used a weaker rule "EN can only be legally ahead for 1 L1 batch" but we can make it stricter as reorg detector is now aware of miniblocks.

montekki: No, it can be legally ahead 1 batch, which what the current code suggests.

montekki: I think that you can generally speaking be legally ahead for the arbitrary amounts of miniblocks in the case main node was reverted.

perekopskiy: AFAIU meaning of "legally" in this case assumes no reverts

montekki: how can that happen without any rollbacks then even theoretically since EN cannot produce any of its own blocks?

perekopskiy: Good question, one thing is DB replicas can have different lags. So we can fetch some miniblock from one API server, process it, and then reorg detector will query another API server which is lagging. Most of the time lag is 0, but we don't want ENs to panic because of slight differences between API responses

montekki: Ok, so essentially there is no guarantee that this lag will be only 1 block, right?

perekopskiy: Yes, leaving 1 batch lag allowed is more safe

montekki: I suspect that at the end of the day simply being ahead whatever amounts of mibiblocks/batches is not a reorg scenario. We probably should be triggering the reorg only in the event when we see wrong hashes at the same height.

perekopskiy: Maybe, lagging behind too much should be catched somewhere IMO, but I'm not sure that reorg detector is a right place for it

montekki: behind is completely another question, yes, talking only about ahead here

perekopskiy: oh, sorry, I meant ahead

montekki: right, this makes sense then. I've updated code to trigger reorgs only if we have detected mismatches in block hashes at the same height.

perekopskiy (approved):

PR #243, opened 2023-10-17
refactor(en): Refactor sync action queue in EN

slowli:
# What ❔

Refactors action queue used by EN to be more idiomatic. Covers sync logic with tests.

## Why ❔

The idea is to reuse EN code other than the fetcher with the new sync mechanism that will be brought by consensus.

## Checklist


popzxc (dismissed):
Nice! Also thought that sealing logic may belong to IO, but never managed to implement it.
Like how it looks now.

popzxc (approved):

PR #244, opened 2023-10-17
feat(snapshot creator): Syncing from snapshots

tomg10:
# What ❔
Draft of snapshot creator with new API for fetching snapshots metadata and very simple snapshot creator binary. I did not want to make this PR too big, so there are some TODO items
## Why ❔
Snapshot-based node syncing allows for a newly created node to initialize without having a recent full backup of the Postgres storage or creating a Merkle tree from scratch. 

## Checklist


core/bin/snapshot_creator/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
```
mm-zk, line 1: drive-by comment:

If we are adding a new crate/binary - please also create a README.md file describing what this tool is for, with some examples on how to run it.

core/lib/dal/src/snapshots_dal.rs
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    #[db_test(dal_crate)]
+    async fn adding_files(pool: ConnectionPool) {
+        let mut conn = pool.access_storage().await.unwrap();
+        let mut dal = conn.snapshots_dal();
+        let l1_batch_number = L1BatchNumber(100);
+        dal.add_snapshot(
+            l1_batch_number,
+            &["test_file1.bin".to_string(), "test_file2.bin".to_string()],
```
slowli, line 96: Maybe a stupid question: What are files supposed to be on the DB level and on the API level? Are they the same? (I guess a key in a publicly accessible GCS bucket – if so, is it assumed that the GCS bucket URL is known by clients?)

core/lib/object_store/src/objects.rs
```diff
@@ -63,6 +65,30 @@ macro_rules! serialize_using_bincode {
...
+/// Derives [`StoredObject::serialize()`] and [`StoredObject::deserialize()`] using
+/// the `json` (de)serializer. Should be used in `impl StoredObject` blocks.
+
+impl StoredObject for StorageLogsSnapshot {
+    const BUCKET: Bucket = Bucket::StorageSnapshot;
+    type Key<'a> = StorageLogsSnapshotKey;
+
+    fn encode_key(key: Self::Key<'_>) -> String {
+        format!(
+            "storage_logs_{}_part{}.json",
```
slowli, line 77: Nit: L1 batch number and part should be left-padded with zeros to guarantee natural ordering of keys.

core/lib/object_store/src/objects.rs
```diff
@@ -63,6 +65,30 @@ macro_rules! serialize_using_bincode {
...
+            key.l1_batch_number, key.chunk_id
+        )
+    }
+
+    //TODO use better language agnostic serialization format like protobuf
```
slowli, line 82: Intuitively, using JSON is a non-starter because of chunk size (order of 1 GiB); a binary format would be more compact and could support streaming deserialization etc. (esp. if it's Protobuf). We'll likely require some compression as well (although it may be sufficient to enable on the HTTP level).

core/lib/zksync_core/src/api_server/snapshots/mod.rs
```diff
@@ -0,0 +1,79 @@
+use std::{net::SocketAddr, time::Duration};
+
+use actix_cors::Cors;
+use actix_web::{
+    dev::Server,
+    {web, App, HttpResponse, HttpServer},
+};
```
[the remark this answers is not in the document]
popzxc: Please do not add new methods to the `zks` namespace. This namespace is "production" one, so we should only add methods there when we're sure that we want it to be a part of public API.

Regarding actix vs axum -- yes, I think we should stick to axum.

core/bin/snapshot_creator/src/main.rs
```diff
@@ -0,0 +1,73 @@
...
+    let fake_batch_id = L1BatchNumber(
+        conn.snapshots_dal()
+            .get_snapshots()
+            .await
+            .unwrap()
+            .snapshots
+            .len() as u32
+            + 1,
+    );
+    println!("Creating snapshot for block {}", fake_batch_id.0);
```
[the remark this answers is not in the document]
RomanBrodetski: right, and we do need JSON log format so that it is properly parsed by datadog

core/lib/zksync_core/src/api_server/snapshots/mod.rs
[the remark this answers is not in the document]
RomanBrodetski: Yeah we need to decide whether it's a new namespace or new server. I was initially thinking we need a server as this endpoint is very distinct from others and might be even served from separate machines. Although moving towards decentralization and people running nodes as single processes, a new namespace makes more sense. 

Sp yeah now leaning towards a new namespace.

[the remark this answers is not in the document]
popzxc: `dev_` would make a nice name for unstable/experimental methods btw 😉

core/lib/dal/src/snapshots_dal.rs
RomanBrodetski: >  if so, is it assumed that the GCS bucket URL is known by clients?

Interesting - in other places where we use GCS we store file names (just like @tomg10 did here). But this one is different as it deals with publicly accessible files. I was thinking we'd store file names (or even full urls) here and will concatenate them with the url base in the API and return such links to the users.

core/lib/zksync_core/src/api_server/snapshots/mod.rs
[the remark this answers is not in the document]
RomanBrodetski: I was thinking `snapshot_`, and the namespace itself is unstable yet... As opposed to moving from dev later on. But for smaller features/single endpoints `dev` would work

etc/system-contracts
RomanBrodetski, line 1: @tomg10 be very careful with those - it's imperative we don't merge changes to smart contracts by mistake

RomanBrodetski (commented):
Looks like a really good start!!

core/lib/zksync_core/src/api_server/snapshots/mod.rs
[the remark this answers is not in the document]
popzxc: Introducing `dev_` right now will make moving existing stuff from `zks_` easier, so we're doing two things in one go 😅

core/lib/dal/migrations/20231026192041_applied-snapshot-status.down.sql
```diff
@@ -0,0 +1,3 @@
+-- Add down migration script here
+
+DROP TABLE IF EXISTS applied_snapshot_status;
```
RomanBrodetski, line 3: nit: please set up your editor so that it edds newline in the end of files on save

core/lib/dal/migrations/20231026192041_applied-snapshot-status.up.sql
```diff
@@ -0,0 +1,9 @@
+-- Add up migration script here
+
+CREATE TABLE applied_snapshot_status
+(
+    l1_batch_number         BIGINT NOT NULL PRIMARY KEY,
+    miniblock_number        BIGINT NOT NULL,
+    is_finished             BOOL NOT NULL,
+    last_finished_chunk_id   INT
```
RomanBrodetski, line 8: I wonder if we really need two tables `snapshots` and `applied_snapshot_status`, this way we would:
* enforce consistency: currently nothing prevents you from having a `applied_snapshot_status` with no matching `snapshots` instance. You could have NOT NULL Foregn key for that, but I don't see a reason for split.

core/lib/dal/migrations/20231013163109_create_snapshots_table.up.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE snapshots
+(
+    l1_batch_number  BIGINT    NOT NULL PRIMARY KEY,
+    miniblock_number BIGINT    NOT NULL,
```
RomanBrodetski, line 4: Let's discuss if we need to savae both l1 batch number and miniblock number in the DB. This is denormalization because you can get the l1 batch number -> miniblock number from `miniblocks` table. 

Although `miniblocks` might be empty until we apply this snapshot. What do we use the `miniblock_number` field for?

RomanBrodetski (commented):
Started doing the review but then saw that some of the previous comments are not resolved yet. Will continue when they are resolved.

core/lib/dal/migrations/20231013163109_create_snapshots_table.up.sql
tomg10: I added it mostly out of convenience, its useful to pass miniblock number in snapshot metadata, but it definitely does not need to be stored in DB, corrected

core/lib/dal/migrations/20231026192041_applied-snapshot-status.up.sql
tomg10: applied_snapshot_status is a table that is used only on EN side, where it keeps progress of applying the snapshot to EN instance. Is exists only to allow recovery process resilient i.e. that it could be resumed if it's interrupted at any point (though that part is very much still WIP). 
snapshots table keeps track of created snapshots and is used in snapshots API.
I decided to have different tables as applying snapshot needs completely different set of columns than the table used for snapshot management from the creator point of view.

core/lib/dal/src/storage_logs_snapshots_dal.rs
```diff
@@ -0,0 +1,141 @@
...
+pub struct StorageLogsSnapshotsDal<'a, 'c> {
+    pub(crate) storage: &'a mut StorageProcessor<'c>,
+}
+
+impl StorageLogsSnapshotsDal<'_, '_> {
+    // not yet used by snapshot_creator
+    #[cfg_attr(not(test), allow(unused))]
+    async fn get_storage_logs_count(&mut self, l1_batch_number: i32) -> Result<u64, sqlx::Error> {
+        let count = sqlx::query!(
+            "SELECT count(*) FROM initial_writes WHERE l1_batch_number >= $1",
```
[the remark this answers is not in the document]
tomg10: Why do you think it would need to check all rows? I see there exists an index on (l1_batch_number) in initial_writes table

core/lib/dal/src/storage_logs_snapshots_dal.rs
```diff
@@ -0,0 +1,92 @@
...
+        .unwrap_or_default();
+        Ok(MiniblockNumber(miniblock_number as u32))
+    }
+    pub async fn get_storage_logs_count(
+        &mut self,
+        l1_batch_number: L1BatchNumber,
+    ) -> Result<u64, sqlx::Error> {
+        let count = sqlx::query!(
+            "SELECT count(*) FROM initial_writes WHERE l1_batch_number >= $1",
+            l1_batch_number.0 as i32
```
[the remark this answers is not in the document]
tomg10: it should, nice catch!

core/lib/dal/src/storage_logs_snapshots_dal.rs
```diff
@@ -0,0 +1,92 @@
...
+}
+
+impl StorageLogsSnapshotsDal<'_, '_> {
+    // not yet used by snapshot_creator
+    pub async fn get_last_miniblock_number(
+        &mut self,
+        l1_batch_number: L1BatchNumber,
+    ) -> Result<MiniblockNumber, sqlx::Error> {
+        let miniblock_number: i64 = sqlx::query!(
+            "select MAX(number) from miniblocks where l1_batch_number <= $1",
```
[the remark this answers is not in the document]
tomg10: corrected

core/lib/dal/migrations/20231026192041_applied-snapshot-status.up.sql
RomanBrodetski: ah I understand better now - `snapshots` is when you are creating snapshots, while `applied_...` is for when you recover from snapshots. So there is no real 1-1 relationship

core/lib/zksync_core/src/api_server/snapshots/mod.rs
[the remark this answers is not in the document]
tomg10: I added it to snapshot namespace to not to have to bother with migrating to another namespace later. @popzxc Can you tell a bit more what do you mean by "Introducing dev_ right now will make moving existing stuff from zks_" ?

core/lib/dal/src/storage_logs_snapshots_dal.rs
```diff
@@ -0,0 +1,141 @@
...
+    // not yet used by snapshot_creator
+    #[cfg_attr(not(test), allow(unused))]
+    async fn get_storage_logs_count(&mut self, l1_batch_number: i32) -> Result<u64, sqlx::Error> {
+        let count = sqlx::query!(
+            "SELECT count(*) FROM initial_writes WHERE l1_batch_number >= $1",
+            l1_batch_number as i32
+        )
+        .fetch_one(self.storage.conn())
+        .await
+        .unwrap()
```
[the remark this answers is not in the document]
tomg10: what's the easiest way to do it @slowli ?

core/lib/object_store/src/objects.rs
tomg10: I changed naming to 
`format!(
            "snapshot_l1_batch_{}_part_{:0<3}.json",
            key.l1_batch_number, key.chunk_id
        )`
in the newest version both storage logs and factory deps can be in the same file, though currently only storage_logs are chunked a I assumed there are orders of magnitudes less factory deps than there are storage logs

core/lib/dal/migrations/20231026192041_applied-snapshot-status.down.sql
tomg10: done

core/lib/object_store/src/objects.rs
tomg10: I have used HTTP compression a lot and I'm a big fan of it, but here I am a bit concerned about storage size on GS as a single uncompressed snapshot would be more than 100GB and I would expect them to compress really well.

core/bin/snapshot_creator/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+use zksync_dal::ConnectionPool;
+use zksync_object_store::{ObjectStore, ObjectStoreFactory};
+use zksync_types::snapshots::{StorageLogsSnapshot, StorageLogsSnapshotKey};
+use zksync_utils::ceil_div;
+
+async fn run(blob_store: Box<dyn ObjectStore>, pool: ConnectionPool) {
+    // TODO metrics
+    let mut conn = pool.access_storage().await.unwrap();
+
+    let batch_id = conn
```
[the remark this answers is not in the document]
tomg10: corrected

core/bin/snapshot_creator/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+    let mut builder = vlog::ObservabilityBuilder::new().with_log_format(log_format);
+    if let Some(sentry_url) = sentry_url {
+        builder = builder
+            .with_sentry_url(&sentry_url)
+            .context("Invalid Sentry URL")?
+            .with_sentry_environment(environment);
+    }
+    let _guard = builder.build();
+
+    let blob_store = ObjectStoreFactory::prover_from_env()
```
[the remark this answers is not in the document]
tomg10: corrected

tomg10:
It's still WIP, though it works relatively well already:
2023-10-30T12:50:25.782786Z  INFO cross_external_nodes_checker::checker: No divergences found

One thing that I'm not super happy with is that with current implementation, I additionally fetch L1BatchWithMetadata for last L1 batch from main node.
Oh, and the query to fetch logs looks extremely ugly now as tree depends on providing keys in increasing order and to sort numbers encoded in little endian, you have to sort "from the end" like, so I ended with a query like:

ORDER BY ARRAY(
     SELECT get_byte(storage_logs.hashed_key, 32 - generate_series)
     FROM generate_series(1, 32)
 )
 But that almost certainly can be fixed by modifying tree recovery code

core/lib/zksync_core/src/api_server/snapshots/mod.rs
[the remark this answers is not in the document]
popzxc: I meant that we can do two things at the same time. `dev_` namespace, if implemented now, could've been used soon to migrate some of the methods we have in `zks_` but don't want to. I'm not opposed to having `snapshot_` namespace though.

core/lib/object_store/src/objects.rs
popzxc: Protobuf seems fine for consistency, but I'm not an expert w.r.t. best practices; probably we need to discuss how to do it properly with folks who have enough experience here.

RomanBrodetski: > a single uncompressed snapshot would be more than 100GB

you mean one chunk??

RomanBrodetski: We'll be compressing chunks individually won't we

core/lib/zksync_core/src/sync_layer/snapshots.rs
```diff
@@ -0,0 +1,285 @@
...
+        .expect("Block must exist");
+    sync_protocol_version(storage, client, sync_block.protocol_version).await;
+    storage
+        .blocks_dal()
+        .insert_miniblock(&MiniblockHeader {
+            number: sync_block.number,
+            timestamp: sync_block.timestamp,
+            hash: sync_block.hash.unwrap(),
+            l1_tx_count: 0,
+            l2_tx_count: 0,
```
RomanBrodetski, line 62: why zeros?

core/lib/zksync_core/src/sync_layer/snapshots.rs
```diff
@@ -0,0 +1,285 @@
...
+    tracing::info!("Found snapshot with data up to l1_batch {}, created at {}, storage_logs are divided into {} chunk(s)", l1_batch_number, metadata.generated_at, snapshot.chunks.len());
+
+    if applied_snapshot_status.is_none() {
+        applied_snapshot_status = Some(AppliedSnapshotStatus {
+            l1_batch_number: l1_batch_number,
+            is_finished: false,
+            last_finished_chunk_id: None,
+        });
+    };
+    let mut applied_snapshot_status = applied_snapshot_status.unwrap();
```
RomanBrodetski, line 210: do we somehow validate that `snapshot_response` is the same as it was when we failed?

RomanBrodetski: that is - main node could have created a n ewer snapshot

core/lib/zksync_core/src/sync_layer/snapshots.rs
```diff
@@ -0,0 +1,285 @@
...
+    storage
+        .applied_snapshot_status_dal()
+        .set_applied_snapshot_status(&applied_snapshot_status)
+        .await
+        .unwrap();
+
+    let recovered_version = l1_batch_number.0 as u64;
+    let db = RocksDB::new(Path::new(&merkle_tree_db_path));
+    let db: &mut dyn PruneDatabase = &mut (RocksDBWrapper::from(db));
+    let mut recovery = MerkleTreeRecovery::new(db, recovered_version);
```
RomanBrodetski, line 221: we have to be very careful with when we recover the tree and when we recover the DB. I think you have all the building blocks here, but we need to think it through better: what if the tree is run on a separate machine? can we still make it work?

RomanBrodetski: perhaps worth speaking in voice - let's do later today or on Thursday

slowli:
@tomg10 

> But that almost certainly can be fixed by modifying tree recovery code

Eh, I'm not entirely sure about this. Sure, a Merkle tree technically can be recovered with keys inserted in a random order (since we don't care about exact node version equality for the original and the recovered tree); the recovery will lose some nice properties (e.g., straightforward cursor for fault tolerance), but it seems manageable. OTOH, Merkle range proofs will only work for tree chunks if chunks, well, cover ranges of the tree keys, which IIUC won't work if we change key ordering, e.g. to non-reversed `hashed_key`s.

@RomanBrodetski @popzxc What's this reverse ordering of `hashed_key` bytes for the tree is caused by, btw? I guess we can't change it at this point, but was there a specific reason for it, or is it just a quirk?

tomg10:
I believe It's caused by the endianness, if a number is 0x0A0B0C, then in little-endian it will be represented in memory(and stored in postgress) as [0x0C, 0x0B, 0x0A], least signifficant number is stored at the smallest address

core/lib/zksync_core/src/sync_layer/snapshots.rs
```diff
@@ -0,0 +1,285 @@
...
+            .set_applied_snapshot_status(&applied_snapshot_status)
+            .await
+            .unwrap();
+    }
+
+    tracing::info!("Processing chunks finished, finalizing merkle tree");
+
+    let tree = recovery.finalize();
+    let started_at = Instant::now();
+    tree.verify_consistency(recovered_version).unwrap();
```
slowli, line 281: I think consistency verification should be opt-in or not performed at all. If local tests are any indication, it can take more time than the tree recovery.

core/lib/zksync_core/src/api_server/web3/namespaces/snapshots.rs
```diff
@@ -0,0 +1,78 @@
...
+    pub async fn get_snapshot_by_l1_batch_number_impl(
+        &self,
+        l1_batch_number: L1BatchNumber,
+    ) -> Result<Option<Snapshot>, Web3Error> {
+        let mut storage_processor = self.state.connection_pool.access_storage().await.unwrap();
+        let mut snapshots_dal = storage_processor.snapshots_dal();
+        let snapshot_files = snapshots_dal
+            .get_snapshot_files(l1_batch_number)
+            .await
+            .unwrap();
```
slowli, line 38: Errors should be propagated here using [`internal_error()`](https://github.com/matter-labs/zksync-era/blob/40328a3a195b3544d50dec723eeff3b6737cd3e8/core/lib/zksync_core/src/api_server/web3/backend_jsonrpc/error.rs#L39).

core/lib/dal/src/storage_logs_snapshots_dal.rs
```diff
@@ -0,0 +1,117 @@
...
+use zksync_types::{AccountTreeId, Address, L1BatchNumber, MiniblockNumber, StorageKey, H256};
+
+#[derive(Debug)]
+pub struct SnapshotChunksDal<'a, 'c> {
+    pub(crate) storage: &'a mut StorageProcessor<'c>,
+}
+
+impl SnapshotChunksDal<'_, '_> {
+    // not yet used by snapshot_creator
+    pub async fn get_last_miniblock_number(
```
slowli, line 12: Nit: There's already `get_miniblock_range_of_l1_batch()` in `BlocksDal`. I'd suggest to reuse it, although technically speaking it does some unnecessary work.

core/bin/snapshot_creator/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+zksync_core = { path = "../../lib/zksync_core" }
+zksync_object_store = { path = "../../lib/object_store" }
+vlog = { path = "../../lib/vlog" }
+
+anyhow = "1.0"
+clap = { version = "4.2.4", features = ["derive"] }
+tokio = { version = "1", features = ["full"] }
+tracing = "0.1"
+futures = "0.3"
+serde = { version = "1.0.175", features = ["derive"] }
```
slowli, line 27: These dependencies are unused, along with `clap`.

core/lib/dal/src/snapshots_dal.rs
```diff
@@ -0,0 +1,135 @@
...
+                .collect();
+        Ok(AllSnapshots { snapshots: records })
+    }
+
+    pub async fn get_snapshot_metadata(
+        &mut self,
+        l1_batch_number: L1BatchNumber,
+    ) -> Result<Option<SnapshotMetadata>, sqlx::Error> {
+        let record: Option<SnapshotMetadata> = sqlx::query!(
+            "SELECT l1_batch_number, created_at FROM snapshots WHERE l1_batch_number = $1",
```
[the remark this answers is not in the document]
tomg10: When I was adding this, I was mostly thinking about a mechanism to be added later that would be removing too old snapshots

core/bin/snapshot_creator/src/main.rs
```diff
@@ -0,0 +1,115 @@
...
+            chunk_id,
+        };
+        blob_store.put(key, &result).await.unwrap();
+        let output_file = blob_store.get_full_path::<SnapshotChunk>(key);
+        output_files.push(output_file.clone());
+        tracing::info!(
+            "Finished storing chunk {}/{} in {}",
+            key.chunk_id + 1,
+            chunks_count,
+            output_file
```
slowli, line 73: Nit: simple variables (i.e., not expressions) can be inlined into logging macros:

```rust
tracing::info!(
    "Finished storing chunk {}/{chunks_count} in {output_file}",
    key.chunk_id + 1
);
```

core/bin/snapshot_creator/src/main.rs
```diff
@@ -0,0 +1,115 @@
+use anyhow::Context as _;
+
+use zksync_dal::connection::DbVariant;
+use zksync_dal::ConnectionPool;
+use zksync_object_store::{ObjectStore, ObjectStoreFactory};
+use zksync_types::snapshots::{SnapshotChunk, SnapshotFactoryDependency, SnapshotStorageKey};
+use zksync_utils::ceil_div;
+
+async fn run(blob_store: Box<dyn ObjectStore>, pool: ConnectionPool) {
```
slowli, line 9: It makes sense propagating errors here using `anyhow::Result` + `anyhow::Context` to add contextual info.

core/bin/snapshot_creator/src/main.rs
```diff
@@ -0,0 +1,115 @@
...
+            key.chunk_id + 1,
+            chunks_count,
+            output_file
+        );
+    }
+
+    conn.snapshots_dal()
+        .add_snapshot(l1_batch_number, &output_files)
+        .await
+        .unwrap();
```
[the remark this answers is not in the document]
tomg10: that was my intention. I did not plan to make creator too resilient now (maybe I would add some retries, but that's all).

.githooks/pre-commit
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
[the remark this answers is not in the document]
tomg10: this script relies on feature `sh` does not need to have, it fails on my ubuntu. It uses "function" keyword that only `bash` has. Having said that, I forgot about this change and accidentally committed it and it really belongs in a separate PR

core/bin/snapshot_creator/src/main.rs
```diff
@@ -0,0 +1,115 @@
...
+            .storage_logs_snapshots_dal()
+            .get_storage_logs_chunk(l1_batch_number, chunk_id, chunk_size)
+            .await
+            .unwrap();
+        let mut factory_deps: Vec<SnapshotFactoryDependency> = vec![];
+        if chunk_id == 0 {
+            factory_deps = conn
+                .storage_logs_snapshots_dal()
+                .get_all_factory_deps(miniblock_number)
+                .await;
```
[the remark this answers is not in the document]
tomg10: do you suggest introducing separate StoredObject type for factory dependencies only? I wanted to avoid having separate "file type" for every type of object we need to store as it would introduce lots of additional complexity, but I agree it looks 'hacky' in the current form. It's also like that as I'm still not sure how big would the dependencies be on prod, it still may be required to introduce chunking of them as well.

core/bin/snapshot_creator/src/main.rs
```diff
@@ -0,0 +1,115 @@
...
+        blob_store.put(key, &result).await.unwrap();
+        let output_file = blob_store.get_full_path::<SnapshotChunk>(key);
+        output_files.push(output_file.clone());
+        tracing::info!(
+            "Finished storing chunk {}/{} in {}",
+            key.chunk_id + 1,
+            chunks_count,
+            output_file
+        );
+    }
```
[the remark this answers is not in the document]
tomg10: +1

core/lib/object_store/src/objects.rs
tomg10: I mean 100GB for a whole snapshot, but it will quickly add up to terabytes if we are to create them regularly. I do have quite a bit of experience with using them from my previous company, we used them on almost all our APIs that required high bandwith

tomg10: yes, I expected compression to be on the individual file level

popzxc:
> What's this reverse ordering of hashed_key bytes for the tree is caused by, btw? I guess we can't change it at this point, but was there a specific reason for it, or is it just a quirk?

@slowli Most likely, because the crypto team implemented it this way; and there most likely because it was more efficient proving-wise. There is also a chance that it just happened to be written this way, without a particularly strong reason, though. I agree that it's unlikely that we can change it now.

core/bin/snapshot_creator/src/main.rs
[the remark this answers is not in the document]
RomanBrodetski: Got it.  @tomg10 thinking more about this - I think we decided to include data from initial_srites to storage_logs directly - that is, in the snapshot chunks. So you shouldnt need a separate file type, you'll just have more fields in each record (i think it's only enumeration_index).

core/lib/dal/src/transactions_web3_dal.rs
```diff
@@ -32,7 +32,7 @@ impl TransactionsWeb3Dal<'_, '_> {
                 WITH sl AS (
                     SELECT * FROM storage_logs
                     WHERE storage_logs.address = $1 AND storage_logs.tx_hash = $2
-                    ORDER BY storage_logs.miniblock_number DESC, storage_logs.operation_number DESC
+                    ORDER BY storage_logs.miniblock_number DESC, storage_logs.hashed_key DESC
```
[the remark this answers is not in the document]
tomg10: Currently, the order in which storage_logs get inserted into DB is not deterministic, EN may have storage logs with different operation_numbers than main node so this query may return different values for EN and main node before this change. Btw, @popzxc  @slowli isn't the non-determinism a bug, i.e. don't we make an assumption somewhere that the enumeration_numbers are the same everywhere or is this a feature?

core/lib/zksync_core/src/sync_layer/snapshots.rs
tomg10: I changed so it's not performed at all

core/bin/snapshot_creator/Cargo.toml
tomg10: corrected

core/lib/zksync_core/src/sync_layer/snapshots.rs
```diff
@@ -0,0 +1,496 @@
...
+    pub state_keeper_db_path: String,
+    pub l2_erc20_bridge_addr: Address,
+    pub chain_id: L2ChainId,
+    pub main_node_url: String,
+    pub enum_index_migration_chunk_size: usize,
+}
+pub struct SnapshotApplier<'a, 'b, 'd> {
+    connection_pool: &'a ConnectionPool,
+    client: &'a dyn MainNodeClient,
+    recovery: MerkleTreeRecovery<'b, RocksDBWrapper>,
```
slowli, line 37: Lifetime here can be replaced in favor of `'static`. I probably need to get rid of this lifetime anyway.

core/lib/zksync_core/src/sync_layer/snapshots.rs
```diff
@@ -0,0 +1,496 @@
...
+        connection_pool: &'a ConnectionPool,
+        client: &'a dyn MainNodeClient,
+        merkle_tree_db_path: &String,
+        state_keeper_config: StateKeeperConfig,
+        metadata_calculator_config: MetadataCalculatorConfig<'d>,
+    ) -> anyhow::Result<SnapshotApplier<'a, 'b, 'd>, SnapshotApplierError> {
+        let mut storage = connection_pool
+            .access_storage_tagged("snapshots_applier")
+            .await
+            .unwrap();
```
slowli, line 66: Gentle reminder to propagate errors (adding context where appropriate) instead of panicking.

core/lib/zksync_core/src/sync_layer/snapshots.rs
```diff
@@ -0,0 +1,496 @@
...
+        let mut applied_snapshot_status = storage
+            .applied_snapshot_status_dal()
+            .get_applied_snapshot_status()
+            .await
+            .unwrap();
+
+        if !storage.blocks_dal().is_genesis_needed().await.unwrap()
+            && applied_snapshot_status.is_none()
+        {
+            return Err(Canceled(
```
slowli, line 78: Bikeshedding: Personally, importing enum variants makes code significantly more difficult to read. There are exceptions (short functions / methods mostly consisting of a `match`), but they only prove the rule.

core/lib/zksync_core/src/sync_layer/snapshots.rs
```diff
@@ -0,0 +1,496 @@
...
+            blob_store,
+            applied_snapshot_status,
+            snapshot,
+            state_keeper_config,
+            metadata_calculator_config,
+        })
+    }
+
+    async fn build_state_keeper(config: StateKeeperConfig) -> ZkSyncStateKeeper {
+        let (_, stop_receiver) = watch::channel(false);
```
slowli, line 137: I don't understand how this is supposed to work. Since the stop sender is immediately dropped, the components initialized with `stop_receiver` should immediately stop. Is this their intended behavior? Why are these components (e.g., fetcher) are started and then waited for anyway? Components are supposed to be spawned as Tokio tasks and run until the program termination; i.e., if a component is waited for, it is a logical problem on its own.

core/lib/zksync_core/src/sync_layer/fetcher.rs
```diff
@@ -26,6 +26,11 @@ pub struct MainNodeFetcherCursor {
     pub(super) l1_batch: L1BatchNumber,
 }
 
+struct FetcherIterationResult {
```
slowli, line 29: Nit: Please use `#[derive(Debug)]` for all types (even private ones); it simplifies debugging, logging etc.

core/lib/zksync_core/src/sync_layer/fetcher.rs
```diff
@@ -94,15 +99,15 @@ pub struct MainNodeFetcher {
...
+    pub async fn run(mut self, stop_after_one_batch: bool) -> anyhow::Result<()> {
         tracing::info!(
             "Starting the fetcher routine. Initial miniblock: {}, initial l1 batch: {}",
             self.cursor.miniblock,
             self.cursor.l1_batch
         );
         // Run the main routine and reconnect upon the network errors.
         loop {
-            match self.run_inner().await {
+            match self.run_inner(stop_after_one_batch).await {
```
slowli, line 110: Really stupid question: What is the purpose of stopping after one batch? It could be theoretically achieved without changing the `MainNodeFetcher` interface by sending a stop signal after observing an L1 batch getting processed.

core/lib/dal/src/storage_logs_dal.rs
```diff
@@ -68,6 +69,46 @@ impl StorageLogsDal<'_, '_> {
...
+                hashed_key = log.key.hashed_key(),
+                address = log.key.address(),
+                key = log.key.key(),
+                value = log.value
+            );
+            writeln_str!(
+                &mut buffer,
+                r"{}|\\x{:x}|{miniblock_number}|{now}|{now}",
+                log.enumeration_index,
+                H256::zero()
```
slowli, line 105: Ah yes, there's a FIXME in the snapshot syncing spec regarding `tx_hash`. I'm actually not entirely sure that fake `tx_hash`es will work as expected; there are at least some DB queries that rely on `tx_hash` values being correct (one of which, `get_transaction_receipt()`, you've touched in this PR).

core/lib/dal/src/storage_logs_dal.rs
```diff
@@ -68,6 +69,46 @@ impl StorageLogsDal<'_, '_> {
         copy.finish().await.unwrap();
     }
 
+    pub async fn insert_storage_logs_from_snapshot(
+        &mut self,
+        miniblock_number: MiniblockNumber,
+        snapshot_storage_logs: &[SnapshotStorageLog],
+    ) {
```
slowli, line 76: Please propagate `sqlx` errors here.

RomanBrodetski:
@tomg10 please `resolve` the comments that you already taken care of - will make it a little easier to review 🙏

core/lib/zksync_core/src/sync_layer/snapshots.rs
tomg10: By adding stop_after_one_batch my intention was to modify those components to allow running them only until certain point in L1 history. This way, after the restore process there exists exactly one L1 batch in db that looks exactly like it would if it was inserted during normal state_keeper operation. We also don't need to worry if snapshot applier for instance 'forgot' to populate one table. Most changes in the codebase won't have to take into consideration that the snapshot applier exists. I suggest a huddle tomorrow to talk more about this change

core/tests/ts-integration/src/context-owner.ts
```diff
@@ -379,6 +383,8 @@ export class TestContextOwner {
 
             await this.collectFunds();
 
+            await this.cleanUpDockerAndPostgres();
```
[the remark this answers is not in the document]
tomg10: In general this is a good remark, but this particular cleanup is designed in a way if this fails that the resources will be just cleaned up in the the next setupContext (info about instances that need cleaning is stored in a file)

core/lib/dal/src/transactions_web3_dal.rs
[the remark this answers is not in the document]
tomg10: sorry, I mixed terms. I meant operation_number, you guessed what I meant correctly and answered my question. About "If so, can we take any of them?" Who do you think can know the answer to this (there can be multiple such logs, currently it's UB which of the contracts is returned)?

core/lib/zksync_core/src/sync_layer/fetcher.rs
tomg10: I don't like the idea of running those components 'for a while until certain conditions are met' that much, we can talk more about it using voice chat

RomanBrodetski:
@tomg10 you'll be opening other PRs right - shall we close this?

PR #251, opened 2023-10-18
refactor(vm): move vm types to interface level

Deniallugo:
# What ❔

Use common types for vm_latest and vm_with_virtual blocks

## Why ❔

We are slightly moving forward to interface unification and now these types allows us to make less transformation

## Checklist


perekopskiy (dismissed):

core/lib/multivm/src/interface/mod.rs
```diff
@@ -0,0 +1,12 @@
+pub mod types;
+
+pub use types::{
```
popzxc, line 3: Nit: IMHO, it's better to expose `errors`/`inputs`/`outputs` rather than all the types in one pack.

popzxc (approved):

core/lib/multivm/src/interface/mod.rs
Deniallugo: Agree, it was done like this before, don't want to make a lot of changes

core/lib/multivm/src/versions/vm_latest/types/l1_batch.rs
```diff
@@ -0,0 +1,41 @@
...
+    ]
+    .into_iter()
+    .collect();
+
+    if let Some(prev_block_hash) = l1_batch.previous_batch_hash {
+        base_params.insert(PREV_BLOCK_HASH_SLOT, h256_to_u256(prev_block_hash));
+        base_params.insert(SHOULD_SET_NEW_BLOCK_SLOT, U256::from(1u32));
+    }
+    base_params.into_iter().collect()
+}
```
montekki, line 41: This is a bit heavy with first creating a vec then creating a `HashMap` and then mutating it back to the `vec`. you could do something like

```rust
let (prev_block_hash, should_set_new_block_slot) = l1_batch.previous_batch_hash.map(|h| (h256_to_u256(prev_block_hash), U256::from(1u32)).unwrap_or_default();
```

and then just plug these values into a `vec![]` initialization and just return this `vec![]` since the default value for `U256` is 0 and for prev block hash slot it is default already.

core/lib/multivm/src/versions/vm_virtual_blocks/types/l1_batch_env.rs
```diff
@@ -0,0 +1,41 @@
...
+    ]
+    .into_iter()
+    .collect();
+
+    if let Some(prev_block_hash) = l1_batch_env.previous_batch_hash {
+        base_params.insert(PREV_BLOCK_HASH_SLOT, h256_to_u256(prev_block_hash));
+        base_params.insert(SHOULD_SET_NEW_BLOCK_SLOT, U256::from(1u32));
+    }
+    base_params.into_iter().collect()
+}
```
montekki, line 41: Same here

montekki (approved):

PR #259, opened 2023-10-18
feat(core): adding pubdata to statekeeper and merkle tree

koloz193:
# What ❔

Setting the ground for bigger boojum changes. Adding boojum related fields to the data model and types. Extending state tree to have a concept of `StateDiffs`. 

## Why ❔

We want to reduce the size of the overall boojum merge, this sets the basis for more extensive changes. These changes are all meant to be non-intrusive.

## Checklist


core/lib/dal/src/models/storage_block.rs
```diff
@@ -43,6 +43,8 @@ pub struct StorageL1BatchHeader {
     pub bootloader_code_hash: Option<Vec<u8>>,
     pub default_aa_code_hash: Option<Vec<u8>>,
     pub protocol_version: Option<i32>,
+    pub system_logs: Vec<Vec<u8>>,
```
mm-zk, line 46: add comments:
mention how it differs from l2_to_l1_logs. 
(and that it might not be set for older batches, right)?

core/lib/dal/src/models/storage_block.rs
```diff
@@ -43,6 +43,8 @@ pub struct StorageL1BatchHeader {
...
+    pub compressed_state_diffs: Option<Vec<u8>>,
```
mm-zk, line 47: mention that it might not be set for older batches, right? 
(are there any other reasons why this might be missing?)

core/lib/merkle_tree/src/domain.rs
```diff
@@ -29,6 +30,8 @@ pub struct TreeMetadata {
     pub repeated_writes: Vec<RepeatedStorageWrite>,
     /// Witness information. As with `repeated_writes`, no-op updates will be omitted from Merkle paths.
     pub witness: Option<PrepareBasicCircuitsJob>,
+    /// State diffs performed in the processed L1 batch in the order of provided `StorageLog`s.
```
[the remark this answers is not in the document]
mm-zk: do we have any guarantees that we do only a single state diff per address? 
(I'm wondering if we shouldn't do a Map instead of Vec)

mm-zk (changes requested):
(I'll add some more changes after my meetings ;-)

mm-zk (changes requested):
(I'll add some more changes after my meetings ;-)

core/lib/dal/src/models/storage_block.rs
koloz193: nope, every batch will have 7 or 8 system logs (the 8th being present if there is a network update)

core/lib/dal/src/models/storage_block.rs
```diff
@@ -260,6 +285,9 @@ impl TryInto<L1BatchMetadata> for StorageL1Batch {
                         .ok_or(StorageL1BatchConvertError::Incomplete)?,
                 ),
             },
+            state_diffs_compressed: self
```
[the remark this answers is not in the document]
koloz193: good catch, we could add a default to the column or change this parsing to use a default value. I lean towards the latter as in boojum we should always have this field

core/lib/types/src/block.rs
```diff
@@ -60,6 +60,8 @@ pub struct L1BatchHeader {
     /// The L2 gas price that the operator agrees on.
     pub l2_fair_gas_price: u64,
     pub base_system_contracts_hashes: BaseSystemContractsHashes,
+    /// System logs are those emitted as part of the Vm excecution.
```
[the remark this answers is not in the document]
koloz193: the l2_to_l1 logs field will still hold user logs but none of the system logs (given that one of the system logs holds the merkle root for all user logs in a given batch)

core/lib/types/src/l2_to_l1_log.rs
```diff
@@ -38,6 +38,17 @@ impl L2ToL1Log {
         self.serialize_commitment(&mut buffer);
         buffer
     }
+
+    pub fn packed_encoding(&self) -> Vec<u8> {
```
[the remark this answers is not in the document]
koloz193: adding a test, also removing in favor of to_bytes and having the called make it into a vector. `packed_encoding` was only used for pubdata construction and parsed out within `L1Messenger`

core/lib/zksync_core/Cargo.toml
```diff
@@ -83,6 +83,8 @@ actix-web = "4.0.0-beta.8"
 
 tracing = "0.1.26"
```
mm-zk, line 85: do we need this?

mm-zk (changes requested):

core/lib/types/src/fee.rs
```diff
@@ -3,8 +3,7 @@ use zksync_utils::ceil_div;
 
 use crate::U256;
 
-#[derive(Debug, Default, Clone, Copy, Serialize, Deserialize)]
-#[serde(rename_all = "camelCase", tag = "result")]
+#[derive(Debug, Default, Clone, Copy)]
```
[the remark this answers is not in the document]
koloz193: yes, it was part of the boojum-integration branch. maybe @StanislavBreadless has insight on if it needs to be removed or not

core/lib/types/src/event.rs
```diff
@@ -163,6 +234,83 @@ pub fn extract_long_l2_to_l1_messages(all_generated_events: &[VmEvent]) -> Vec<V
...
+                    ethabi::ParamType::Uint(16),
+                    ethabi::ParamType::Address,
+                    ethabi::ParamType::FixedBytes(32),
+                    ethabi::ParamType::FixedBytes(32),
+                ])],
+                &event.value,
+            )
+            .expect("Failed to decode L2ToL1LogSent message")
+            .first()
+            .unwrap()
```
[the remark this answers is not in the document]
koloz193: we could but given that we already filter for the sender to be the `L1Messenger` and the topic to be `L2ToL1LogSent` we should never be in the scenario where we fail to decode the log unless an update is made to the structure without the proper sequencer side changes.

core/lib/merkle_tree/src/domain.rs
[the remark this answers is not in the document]
koloz193: they're guaranteed to be unique at the derived key (hash of address, key concatenation) given these values come from applying logs to the tree

core/lib/zksync_core/Cargo.toml
koloz193: yes there are still files within `zksync_core` that rely on it

core/lib/dal/migrations/20231018205126_l1_batches_boojum_fields.up.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE l1_batches
+    ADD COLUMN IF NOT EXISTS system_logs BYTEA[]
+    NOT NULL DEFAULT '{}';
+
+ALTER TABLE l1_batches
+    ADD COLUMN IF NOT EXISTS compressed_state_diffs BYTEA;
```
slowli, line 6: @RomanBrodetski A question I've wanted to ask for some time: Is it not feasible to have a catch-all `jsonb` column in `l1_batches` (and maybe other tables) so that we don't need to perform migrations each time we add a new piece of data for the corresponding entity?

core/lib/merkle_tree/src/domain.rs
```diff
@@ -180,15 +189,20 @@ impl ZkSyncTree {
         let starting_leaf_count = self.tree.latest_root().leaf_count();
         let starting_root_hash = self.tree.latest_root_hash();
 
+        let instructions_with_hashed_keys = instructions
```
slowli, line 192: Bikeshedding: It would be moderately easy to have `Self::transform_logs()` return `(Key, _)` instead of `(StorageKey, _)` (the original `StorageKey` can be easily associated with entries given that `transform_logs` doesn't skip or reorder input logs. OTOH, this would make it inconsistent with `Self::filter_write_logs()`, so I'm not insisting.

core/lib/merkle_tree/src/domain.rs
```diff
@@ -280,18 +295,31 @@ impl ZkSyncTree {
...
                     initial_writes.push(InitialStorageWrite {
                         index: leaf_index,
-                        key,
+                        key: key.hashed_key_u256(),
                         value,
                     });
+                    state_diffs.push(StateDiffRecord {
+                        address: *key.address(),
+                        key: h256_to_u256(*key.key()),
+                        derived_key: StorageKey::raw_hashed_key(key.address(), key.key()),
```
slowli, line 318: Just to understand: Is replacing these 3 fields with `StorageKey` incorrect from the domain perspective? AFAICT, a `StorageKey` would contain data sufficient to restore these fields.

core/lib/types/src/event.rs
```diff
@@ -51,6 +53,61 @@ static L1_MESSAGE_EVENT_SIGNATURE: Lazy<H256> = Lazy::new(|| {
...
+    sender: Address,
+    key: U256,
+    value: U256,
+}
+
+impl L1MessengerL2ToL1Log {
+    pub fn packed_encoding(&self) -> Vec<u8> {
+        let mut res = vec![];
+        res.extend_from_slice(&self.l2_shard_id.to_be_bytes());
+        res.extend_from_slice(&(self.is_service as u8).to_be_bytes());
```
slowli, line 78: Bikeshedding: `u8::to_be_bytes()` looks weird; you could just `push()` a `u8` value.

core/lib/zksync_core/src/state_keeper/batch_executor/tests/tester.rs
```diff
@@ -98,6 +98,8 @@ impl Tester {
             .access_storage_tagged("state_keeper")
             .await
             .unwrap();
+
+        secondary_storage.enable_enum_index_migration(100);
```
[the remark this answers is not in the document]
mm-zk: good point. Porbably this whole file here should be under cfg(test) - but that's separate issue / separate PR

mm-zk:
from my perspective LGTM. Please update PR description (and maybe change it to real PR and not draft ;-)

core/lib/dal/migrations/20231018205126_l1_batches_boojum_fields.up.sql
koloz193: makes sense esp for fields not used relationally

core/lib/merkle_tree/src/domain.rs
koloz193: Not incorrect, the basis for this was from test harness. It definitely could be updated but I'm not sure there's enough benefit to doing it now

core/lib/merkle_tree/src/domain.rs
```diff
@@ -302,13 +330,22 @@ impl ZkSyncTree {
...
+                            final_value: h256_to_u256(value),
+                        });
                     }
                     // Else we have a no-op update that must be omitted from `repeated_writes`.
                 }
                 TreeLogEntry::Read { .. } | TreeLogEntry::ReadMissingKey => {}
             }
         }
-        (initial_writes, repeated_writes)
+        state_diffs.sort_by_key(|rec| (rec.address, rec.key));
```
popzxc, line 347: Nit: Is stable sort required here? Would `sort_unstable_by_key` suffice?

core/lib/state/Cargo.toml
```diff
@@ -22,6 +22,8 @@ tokio = { version = "1", features = ["rt"] }
 tracing = "0.1"
 itertools = "0.10.3"
 
+metrics = "0.21"
```
popzxc, line 25: If I didn't miss anything, looks like the new metrics use `vise` library. I guess this dependency can be removed?

core/lib/types/src/commitment.rs
```diff
@@ -209,28 +253,54 @@ struct L1BatchAuxiliaryOutput {
     initial_writes: Vec<InitialStorageWrite>,
     #[allow(dead_code)]
     repeated_writes: Vec<RepeatedStorageWrite>,
+
     l2_l1_logs_compressed: Vec<u8>,
     l2_l1_logs_linear_hash: H256,
     l2_l1_logs_merkle_root: H256,
+
+    // Once cut over to boojum, these fields are no longer required as their values
```
popzxc, line 261: Would it make sense to create a task to get rid of these fields then? Otherwise we're risking having them here forever.

core/lib/types/src/event.rs
```diff
@@ -51,6 +53,61 @@ static L1_MESSAGE_EVENT_SIGNATURE: Lazy<H256> = Lazy::new(|| {
     )
 });
 
+// struct L2ToL1Log {
```
popzxc, line 56: Would it make sense to add a bit more context to this comment? E.g. prepend with "Solidity structure counterpart:" and move the structure into a markdown code block.

core/lib/zksync_core/src/state_keeper/batch_executor/tests/tester.rs
popzxc, line 102: Nit: A comment on why is this line is required would be nice to have.

core/lib/zksync_core/Cargo.toml
```diff
@@ -83,6 +83,8 @@ actix-web = "4.0.0-beta.8"
...
 
+zkevm_test_harness = { git = "https://github.com/matter-labs/era-zkevm_test_harness.git", branch = "v1.4.0" }
```
perekopskiy, line 86: This import is not used, please remove it

core/lib/types/src/fee.rs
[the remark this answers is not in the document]
perekopskiy: Unless there is an explanation why we should change (de)serialization, it shouldn't be changed

core/lib/types/src/event.rs
```diff
@@ -163,6 +220,87 @@ pub fn extract_long_l2_to_l1_messages(all_generated_events: &[VmEvent]) -> Vec<V
...
+            let tuple = ethabi::decode(
+                params,
+                &event.value,
+            )
+            .expect("Failed to decode L2ToL1LogSent message")
+            .first()
+            .unwrap()
+            .clone();
+            let Token::Tuple(tokens) = tuple else {
+                unreachable!("Tuple was expected, got: {}", tuple);
```
perekopskiy, line 257: nit: it should be `panic`, not `unreachable`

core/lib/types/src/event.rs
```diff
@@ -163,6 +220,87 @@ pub fn extract_long_l2_to_l1_messages(all_generated_events: &[VmEvent]) -> Vec<V
...
+            };
+            let [
+                Token::Uint(shard_id),
+                Token::Bool(is_service),
+                Token::Uint(tx_number_in_block),
+                Token::Address(sender),
+                Token::FixedBytes(key_bytes),
+                Token::FixedBytes(value_bytes),
+            ] = tokens.as_slice() else {
+                unreachable!("Invalid tuple types");
```
perekopskiy, line 267: same here

core/lib/types/src/event.rs
```diff
@@ -163,6 +220,87 @@ pub fn extract_long_l2_to_l1_messages(all_generated_events: &[VmEvent]) -> Vec<V
...
+                key: U256::from_big_endian(key_bytes),
+                value: U256::from_big_endian(value_bytes),
+            }
+        })
+        .collect()
+}
+
+// Extracts all the bytecode publication requests
+// that were emitted by the L1Messenger contract
+pub fn extract_bytecode_publication_requests_from_l1_messenger(
```
perekopskiy, line 283: Please cover it with test(s)

perekopskiy (changes requested):

core/lib/dal/src/models/storage_block.rs
```diff
@@ -151,6 +166,15 @@ pub struct StorageL1Batch {
     pub l1_gas_price: i64,
     pub l2_fair_gas_price: i64,
 
+    // These fields are not used, but are present for compatibility reasons
+    pub gas_per_pubdata_byte_in_block: Option<i32>,
+    pub gas_per_pubdata_limit: i64,
+
+    pub skip_proof: bool,
```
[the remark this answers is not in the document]
koloz193: they were part of the larger boojum integration and already part of the db. Happy to remove them though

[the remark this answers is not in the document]
perekopskiy: I don't know why they were added in that PR, we don't use these fields as a part of `StorageBlock` struct. So adding them only bloats SQL queries and `sqlx-data.json`.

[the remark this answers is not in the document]
koloz193: ok i can remove, it seems they are unused in the `boojum-integration` branch as well

core/lib/dal/src/blocks_dal.rs
```diff
@@ -183,7 +184,8 @@ impl BlocksDal<'_, '_> {
                 l2_l1_compressed_messages, l2_l1_merkle_root, l1_gas_price, l2_fair_gas_price, \
                 rollup_last_leaf_index, zkporter_is_available, bootloader_code_hash, \
                 default_aa_code_hash, base_fee_per_gas, aux_data_hash, pass_through_data_hash, \
-                meta_parameters_hash, protocol_version, \
+                meta_parameters_hash, protocol_version, system_logs, compressed_state_diffs, skip_proof, 
+                gas_per_pubdata_byte_in_block, gas_per_pubdata_limit, \
```
perekopskiy, line 188: No need to select `skip_proof, gas_per_pubdata_byte_in_block, gas_per_pubdata_limit`. Please fix it for all touched queries

core/lib/state/src/lib.rs
```diff
@@ -55,6 +55,9 @@ pub trait ReadStorage: fmt::Debug {
         let code_key = get_known_code_key(bytecode_hash);
         self.read_value(&code_key) != H256::zero()
     }
+
+    /// Retrieves the enumeration index for a given `key`.
+    fn get_enumeration_index(&mut self, _key: &StorageKey) -> Option<u64>;
```
perekopskiy, line 60: nit: 
```suggestion
    fn get_enumeration_index(&mut self, key: &StorageKey) -> Option<u64>;
```

core/lib/state/src/rocksdb/mod.rs
```diff
@@ -183,8 +184,6 @@ impl RocksdbStorage {
         tracing::info!(
             "Secondary storage for L1 batch #{latest_l1_batch_number} initialized, size is {estimated_size}"
         );
-
-        self.save_missing_enum_indices(conn).await;
```
perekopskiy, line 187: This shouldn't be touched

core/lib/types/src/commitment.rs
koloz193: i think so, @perekopskiy @RomanBrodetski what do you think and should i make this a platform or protocol task?

core/lib/merkle_tree/src/domain.rs
koloz193: the typing was changed to leverage the the additional address field, in addition to key, for constructing the `StateDiffRecord`

core/lib/types/src/commitment.rs
perekopskiy: It will be relevant to l1_batches table split/clean up, feel free to create platform task

perekopskiy (dismissed):

core/lib/dal/src/blocks_dal.rs
```diff
@@ -974,16 +984,16 @@ impl BlocksDal<'_, '_> {
...
             FROM l1_batches \
             LEFT JOIN commitments ON commitments.l1_batch_number = l1_batches.number \
             JOIN protocol_versions ON protocol_versions.id = l1_batches.protocol_version \
             WHERE eth_commit_tx_id IS NULL \
                 AND number != 0 \
                 AND protocol_versions.bootloader_code_hash = $1 AND protocol_versions.default_account_code_hash = $2 \
                 AND commitment IS NOT NULL \
-                AND (protocol_versions.id = $3 OR protocol_versions.upgrade_tx_hash IS NULL) \
+                AND (protocol_versions.id = $3 OR protocol_versions.upgrade_tx_hash IS NULL)
             ORDER BY number LIMIT $4",
```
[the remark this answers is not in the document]
koloz193: i need to resolve the merge conflict so i will add that in since it touches the same file

perekopskiy (approved):

PR #265, opened 2023-10-19
fix(db): Fix write stalls in RocksDB (again)

slowli:
# What ❔

RocksDB write stalls are still happening, this time for a different reason. Previously, they were caused by too many immutable memtables, this time – by too many level-0 SST files. This PR:

- Tunes RocksDB options some more (the main tuning point is [optimizing level-style compaction](https://docs.rs/rocksdb/latest/rocksdb/struct.Options.html#method.optimize_level_style_compaction)).
- Increases the number of retries on stall and introduces exponential backoff.
- Introduces a dozen of RocksDB metrics that should help monitoring RocksDB health.

## Why ❔

Having write stalls leads to panics and is obviously bad.

## Checklist

### DIFF
--- a/core/lib/zksync_core/src/metadata_calculator/helpers.rs
+++ b/core/lib/zksync_core/src/metadata_calculator/helpers.rs
@@ -18,7 +18,7 @@ use zksync_merkle_tree::{
     domain::{TreeMetadata, ZkSyncTree, ZkSyncTreeReader},
     Key, MerkleTreeColumnFamily, NoVersionError, TreeEntryWithProof,
 };
-use zksync_storage::RocksDB;
+use zksync_storage::{RocksDB, RocksDBOptions};
 use zksync_types::{block::L1BatchHeader, L1BatchNumber, StorageLog, H256};
 
 use super::metrics::{LoadChangesStage, TreeUpdateStage, METRICS};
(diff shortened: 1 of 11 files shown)


slowli:
On the stage env (if I read RocksDB logs correctly), the latest write stalls were caused by compaction that ran *just a bit* too long (~2s) to be covered by retries. Since now the interval before the last retry is ~1.9s, it could fix the problem on its own. One thing that we could try is to check whether the writes are stopped immediately after DB initialization, and if they are, wait until writes become unstuck (maybe, with an upper cap like 10s), folding this time into RocksDB initialization. A problem with this is approach – if I read RocksDB logs correctly – is that writes are stopped ~0.5s after the DB is initialized, not immediately (not sure why).

core/lib/zksync_core/src/metadata_calculator/helpers.rs
```diff
@@ -439,13 +450,23 @@ mod tests {
...
         }
     }
 
+    async fn create_tree(temp_dir: &TempDir) -> AsyncTree {
+        AsyncTree::new(
+            temp_dir.path().to_owned(),
+            MerkleTreeMode::Full,
+            500,
+            0,
+            16 << 20, // 16 MiB
```
AnastasiiaVashchuk, line 465: why `16 << 20`?

AnastasiiaVashchuk (dismissed):

perekopskiy (approved):

AnastasiiaVashchuk (approved):

PR #268, opened 2023-10-19
test(en): Test sync actions produced by main node fetcher

slowli:
# What ❔

- Tests actions produced by the main node fetcher. As a side effect, adds a way to launch a JSON-RPC HTTP server in-process and adds a primitive test for it as well.
- Refactors the main node fetcher slightly, extracting the EN–main node client interface to a trait.

## Why ❔

- Creates reference action queue contents that may be tested against when an alternative fetcher based on the consensus gossip network is implemented.
- Test coverage good.

## Checklist


popzxc (approved):
Nice.

PR #269, opened 2023-10-19
feat(core): added stats to track pubdata needed for compression

koloz193:
# What ❔

Updated code responsible for determining seal criteria based on pubdata size to take into account new compression strategy. 

Code does NOT affect the flow or current values.

## Why ❔

With our new compression strategy, we can no longer assume the size of initial/repeated writes. Calculating based on actual value updates will allow us to fit more writes into a batch.

## Checklist


core/lib/constants/src/ethereum.rs
```diff
@@ -10,7 +10,9 @@ pub static ETHEREUM_ADDRESS: Address = Address::zero();
 /// to be published inside the body of transaction (i.e. excluding of factory deps).
 pub const GUARANTEED_PUBDATA_PER_L1_BATCH: u64 = 4000;
 
-/// The maximum number of pubdata per L1 batch.
+/// The maximum number of pubdata per L1 batch. This limit is due to the fact that the Ethereum
+/// nodes do not accept transactions that have more than 128kb of pubdata.
+/// The 18kb margin is left in case of any inpreciseness of the pubdata calculation.
```
perekopskiy, line 15: isn't margin equal to 128 - 120 = 8?

core/lib/types/src/storage_writes_deduplicator.rs
```diff
@@ -11,6 +12,8 @@ pub struct ModifiedSlot {
     pub value: U256,
     /// Index (in L1 batch) of the transaction that lastly modified the slot.
     pub tx_index: u16,
+    /// Size of pubdata update in bytes
+    pub size: usize,
```
perekopskiy, line 16: nit: I'd prefer to change field name to `compressed_size` as it is more explicit

perekopskiy (commented):
> Code does NOT affect the flow or current values, we'll need to update [size fn](https://github.com/matter-labs/zksync-era/blob/b30f9f81b42771179c352dc137d5f0bf57e0fb02/core/lib/types/src/tx/tx_execution_info.rs#L80) in tx_execution_info for this to be in use.

We need and already can make `size fn` logic depending on protocol version. If you wish it can be a separate PR but anyway it needs to be done.

core/lib/types/src/tx/tx_execution_info.rs
```diff
@@ -35,13 +35,15 @@ impl TxExecutionStatus {
 pub struct DeduplicatedWritesMetrics {
     pub initial_storage_writes: usize,
     pub repeated_storage_writes: usize,
+    pub total_updated_values_size: usize,
```
mm-zk, line 38: comment

core/lib/types/src/storage_writes_deduplicator.rs
```diff
@@ -357,41 +392,47 @@ mod tests {
...
                     tx_index: 0,
+                    compressed_size: 2,
                 },
             ),
             (
                 new_storage_key(3, 4),
                 ModifiedSlot {
                     value: 7u32.into(),
                     tx_index: 0,
+                    compressed_size: 2,
```
mm-zk, line 435: can we check metrics values inside these tests?

mm-zk (approved):

koloz193:
> > Code does NOT affect the flow or current values, we'll need to update [size fn](https://github.com/matter-labs/zksync-era/blob/b30f9f81b42771179c352dc137d5f0bf57e0fb02/core/lib/types/src/tx/tx_execution_info.rs#L80) in tx_execution_info for this to be in use.
> 
> We need and already can make `size fn` logic depending on protocol version. If you wish it can be a separate PR but anyway it needs to be done.

I will follow up with a separate PR covering this

core/lib/types/src/tx/tx_execution_info.rs
```diff
@@ -33,21 +35,35 @@ impl TxExecutionStatus {
...
+            ProtocolVersionId::Version17 => {
+                self.total_updated_values_size
+                    + (BYTES_PER_DERIVED_KEY as usize) * self.initial_storage_writes
+                    + (BYTES_PER_ENUMERATION_INDEX as usize) * self.repeated_storage_writes
+            }
+            _ => {
+                self.initial_storage_writes * InitialStorageWrite::SERIALIZED_SIZE
+                    + self.repeated_storage_writes * RepeatedStorageWrite::SERIALIZED_SIZE
+            }
+        }
```
perekopskiy, line 66: It's better to do `if protocol_version >= 17 then ... else ...`, so for versions > 17 we apply new impl

core/lib/zksync_core/src/gas_tracker/mod.rs
```diff
@@ -103,9 +109,18 @@ pub(crate) fn commit_gas_count_for_l1_batch(
...
         .sum();
-    let additional_calldata_bytes = metadata.initial_writes_compressed.len() as u32
-        + metadata.repeated_writes_compressed.len() as u32
-        + metadata.l2_l1_messages_compressed.len() as u32
+
+    // Protocol Version 17 introduces boojum, and along with boojum an update to how storage writes
+    // are communicated/compressed.
+    let additional_calldata_bytes = match header.protocol_version.unwrap() {
+        ProtocolVersionId::Version17 => {
+            panic!("add in the size of the compressed state diffs from metadata")
```
perekopskiy, line 117: you panic here because we don't have `compressed_state_diffs` merged into main, right?

perekopskiy: anyway, the whole expression looks messy, please store the result of the match in a separate variable

core/lib/zksync_core/src/state_keeper/seal_criteria/conditional_sealer.rs
```diff
@@ -27,7 +28,14 @@ impl ConditionalSealer {
...
             const TX_COUNT: usize = 1;
 
-            let resolution = sealer.should_seal(config, MOCK_BLOCK_TIMESTAMP, TX_COUNT, data, data);
+            let resolution = sealer.should_seal(
+                config,
+                MOCK_BLOCK_TIMESTAMP,
+                TX_COUNT,
+                data,
+                data,
+                ProtocolVersionId::latest(),
```
perekopskiy, line 37: This is incorrect to use `latest()` here. I'd make `find_unexecutable_reason` accept version id as argument. It's called only in a `tx_sender` where it's okay to assume that the current protocol version is equal to the protocol versions of the latest sealed miniblock

core/lib/zksync_core/src/gas_tracker/mod.rs
koloz193: correct, i can update it. i tried out the match with addition and thought it was a clever way of chaining the two but dont feel strongly about keeping it that way

core/lib/zksync_core/src/state_keeper/seal_criteria/conditional_sealer.rs
perekopskiy: I meant not just moving `ProtocolVersionId::latest` stack up but also replacing it with "protocol versions of the latest sealed miniblock" that you can load from DB

core/lib/zksync_core/src/state_keeper/seal_criteria/mod.rs
```diff
@@ -93,7 +93,7 @@ impl SealData {
         let execution_metrics = ExecutionMetrics::from_tx_metrics(tx_metrics);
         let writes_metrics = DeduplicatedWritesMetrics::from_tx_metrics(tx_metrics);
         let gas_count = gas_count_from_tx_and_metrics(&transaction, &execution_metrics)
-            + gas_count_from_writes(&writes_metrics);
+            + gas_count_from_writes(&writes_metrics, ProtocolVersionId::latest());
```
perekopskiy, line 96: Here the approach should be the same as for `find_unexecutable_reason`

core/lib/zksync_core/src/state_keeper/seal_criteria/conditional_sealer.rs
perekopskiy: @koloz193

core/lib/zksync_core/src/api_server/tx_sender/mod.rs
```diff
@@ -906,8 +908,24 @@ impl<G: L1GasPriceProvider> TxSender<G> {
...
+            .access_storage_tagged("api")
+            .await
+            .unwrap()
+            .blocks_dal()
+            .get_last_sealed_miniblock_header()
+            .await
+            .unwrap()
+            .unwrap_or_default()
+            .protocol_version
+            .unwrap_or_default();
```
perekopskiy, line 923: nit, you can unwrap in both places safely:
1. Unwrapping header is safe because genesis header must be present before running API.
2. Protocol versions are not set only for some historical blocks, so it's safe to unwrap `protocol_version` of **latest** block

Then you won't need `MiniblockHeader` to implement `Default`.

core/lib/types/src/fee.rs
```diff
@@ -22,6 +22,9 @@ pub struct TransactionExecutionMetrics {
     pub total_log_queries: usize,
     pub cycles_used: u32,
     pub computational_gas_used: u32,
+    // Field holding the total number of bytes used in storage writes. Unused for now
+    // but required for new pubdata compression
+    pub total_updated_values_size: usize,
```
RomanBrodetski, line 27: nit: having a comment is sufficient but in theory we could name it  `updated_values_bytes_size`. Feel free to discard.

core/lib/types/src/storage_writes_deduplicator.rs
```diff
@@ -167,17 +180,33 @@ impl StorageWritesDeduplicator {
...
+                        .unwrap_or_else(|| {
+                            panic!("tried removing key: {:?} before insertion", item.key)
+                        });
+                    *total_size += value.compressed_size;
+                    *total_size -= old_value.compressed_size;
                 }
                 UpdateType::Remove(value) => {
                     self.modified_key_values.insert(item.key, value);
                     *field_to_change += 1;
+                    *total_size += value.compressed_size;
```
RomanBrodetski, line 209: ah oh this method is complex, and we adding more mutable/imperative statements.. I don't see a quick improvement here - just something to keep in mind in the future

core/lib/zksync_core/src/state_keeper/seal_criteria/mod.rs
```diff
@@ -111,6 +112,7 @@ pub(super) trait SealCriterion: fmt::Debug + Send + 'static {
         tx_count: usize,
         block_data: &SealData,
         tx_data: &SealData,
+        protocol_version: ProtocolVersionId,
```
RomanBrodetski, line 115: Having to pass `protovol_version ` everywhere is a little awkward. I feel like it's worth to discuss a little if there are other approaches. If we are not mindful about this, we might en d up having a `protocol_version` expanding to every method in our codebase. 

@popzxc @slowli - any quick thoughts on this?

RomanBrodetski (commented):
I really like how we develop this feature in small PRs. I can follow them - unlike having a long-lived branch

core/lib/types/src/fee.rs
koloz193: okay, i think thats a fair point. if another update is needed before merging in then i will include this as well

core/lib/zksync_core/src/state_keeper/seal_criteria/mod.rs
koloz193: could we instead have it as part of the `StateKeeperConfig`, initially i tried to keep it out of there because i thought it was protocol version agnostic but seeing `bootloader_hash` and `default_aa_hash`, i'm thinking it has some reliance on protocol version. wdyt?

perekopskiy: `StateKeeperConfig` is static, meaning it's loaded from env at the start and isn't changed, whilst protocol version can be changed. `bootloader_hash` and `default_aa_hash` are artifacts, hopefully, they are not used anywhere and should be removed.

As for Roman's concern: I generally agree, maybe making seal criteria depend on protocol version is not the best solution. But we can remove this dependency after the upgrade, so it doesn't look that bad

popzxc: Agree with Misha (though I'd mark it with a TODO to actually remove it one day).

core/lib/zksync_core/src/gas_tracker/mod.rs
```diff
@@ -103,8 +109,20 @@ pub(crate) fn commit_gas_count_for_l1_batch(
...
         .map(|factory_dep| factory_dep.len() as u32)
         .sum();
-    let additional_calldata_bytes = metadata.initial_writes_compressed.len() as u32
-        + metadata.repeated_writes_compressed.len() as u32
+
+    // Protocol Version 17 introduces boojum, and along with boojum an update to how storage writes
+    // are communicated/compressed.
+    let state_diff_size = match header.protocol_version.unwrap() {
+        ProtocolVersionId::Version17 => {
+            panic!("add in the size of the compressed state diffs from metadata")
```
popzxc, line 117: (I didn't fully understand the previous discussion on this line, so sorry in advance): do I get it right that this panic is going to be fixed soon? If so, would it make sense to replace `panic!` with `todo!` to make the intention more explicit?

perekopskiy: Also, this panic makes upgrade test fail in CI. I don't like this test to rely on the fact that server works with an incremented version, but I also don't see how it can be changed.. I think the easiest solution is to wait for https://github.com/matter-labs/zksync-era/pull/259 to be merged and remove this panic

core/lib/types/src/storage_writes_deduplicator.rs
```diff
@@ -107,26 +110,31 @@ impl StorageWritesDeduplicator {
                 &mut self.metrics.repeated_storage_writes
             };
 
+            let total_size = &mut self.metrics.total_updated_values_size;
```
slowli, line 113: Maybe a stupid question: Could `metrics` be computed on demand based on `self.modified_key_values` instead of being contained in `StorageWritesDeduplicator`? IMO, if it's possible (and doesn't have significant computational overhead), it would be preferable because such computation would be stateless (doesn't need to take rollbacks into account etc.).

core/lib/zksync_core/src/state_keeper/seal_criteria/mod.rs
slowli: In general, we could have the set of criteria used for a certain input be dependent on the protocol version (e.g., by extending the `SealCriterion` interface). In this way, dependency on the protocol version would be contained in `ConditionalSealer`. I'm not sure whether this would make things better or worse, given that it seems like this dependency on the protocol version isn't supposed to be permanent.

core/lib/types/src/block.rs
```diff
@@ -65,7 +65,7 @@ pub struct L1BatchHeader {
 }
 
 /// Holder for the miniblock metadata that is not available from transactions themselves.
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[derive(Default, Debug, Clone, PartialEq, Serialize, Deserialize)]
```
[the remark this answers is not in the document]
koloz193: yea, we changed from `unwrap_or_default` to `unwrap` in the place it should have been used. will update

core/lib/zksync_core/src/api_server/tx_sender/mod.rs
```diff
@@ -906,8 +908,24 @@ impl<G: L1GasPriceProvider> TxSender<G> {
...
+            .access_storage_tagged("api")
+            .await
+            .unwrap()
+            .blocks_dal()
+            .get_last_sealed_miniblock_header()
+            .await
+            .unwrap()
+            .unwrap()
+            .protocol_version
+            .unwrap();
```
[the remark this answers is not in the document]
koloz193: good call out, both calling functions can be refactored a bit to passthrough a StorageProcessor

popzxc (dismissed):
Disclaimer: I only checked the code, not the domain logic.
Strongly recommend getting approval from someone with enough knowledge to validate the correctness 😅

core/lib/types/src/protocol_version.rs
```diff
@@ -70,6 +73,7 @@ impl ProtocolVersionId {
             ProtocolVersionId::Version15 => VmVersion::VmVirtualBlocks,
             ProtocolVersionId::Version16 => VmVersion::VmVirtualBlocksRefundsEnhancement,
             ProtocolVersionId::Version17 => VmVersion::VmVirtualBlocksRefundsEnhancement,
+            ProtocolVersionId::Version18 => VmVersion::VmVirtualBlocksRefundsEnhancement,
```
StanislavBreadless, line 76: nit: shall we have `todo` here too?

StanislavBreadless (approved):

core/lib/types/src/protocol_version.rs
koloz193: we could, originally @perekopskiy said we should use what v17 was using but if its going to be changed with the overall boojum integration merge i think todo is safe

perekopskiy: we can't use todo because of upgrade test..

koloz193:
closing this pr in favor of returning the data from the vm

PR #270, opened 2023-10-19
fix(en): gracefully shutdown en waiting for reorg detector

montekki:
This PR changes the shutdown order of the external node to a more graceful one.

# Why?

Currently sutting down the EN because of reorgs and inconsistency checks may be raceful and a shutting down consistency checker may bring down the reorg detector with it which would prevent the reorg detector from doing it's job in the legitimate cases of reorgs where the node actually needs to be rolled back to some previous state.

# What happens currently?

The concerns have been focused around the logic of consistency checker that used to panic when an inconsistency was found. That is no longer the case (`anyhow::bail!` used to be `panic`):

https://github.com/matter-labs/zksync-era/blob/feb8a6c7053cc5e0202088f6a1f7644316e1ad05/core/lib/zksync_core/src/consistency_checker/mod.rs#L181-L183

however, this error would still unconditionally bring the node down because currently the handler for `ConsistencyChecker` is a part of `wait_tasks` that would resolve if any of the tasks has resolved:

https://github.com/matter-labs/zksync-era/blob/feb8a6c7053cc5e0202088f6a1f7644316e1ad05/core/bin/external_node/src/main.rs#L401-L404

The proposed logic works as follows:

The handle to reorg detector is fused and on the node shutdown, after all the components are gracefully shutdown we manually check the result of the reorg detector. This logic handles both situations:

1. node is shutting down
2. reorg detector has found a reorg and is performing a rollback and shutdown.


core/bin/external_node/src/main.rs
```diff
@@ -177,12 +177,7 @@ async fn init_tasks(
     let tree_handle =
         task::spawn(metadata_calculator.run(tree_pool, prover_tree_pool, tree_stop_receiver));
 
-    let consistency_checker_handle = if !config.optional.experimental_multivm_support {
```
popzxc, line 180: Is this variable used anywhere now? If not, probably it makes sense to remove it altogether (I think it's undocumented, so no changes in docs/announcements required).

popzxc (dismissed):

core/bin/external_node/src/main.rs
montekki: No, will remove

popzxc (approved):

Deniallugo (approved):

Deniallugo (dismissed):

core/bin/external_node/src/main.rs
```diff
@@ -387,46 +379,77 @@ async fn main() -> anyhow::Result<()> {
...
+                            .rollback_db(last_correct_batch, BlockReverterFlags::all())
+                            .await;
+                        tracing::info!("Rollback successfully completed, the node has to restart to continue working");
+                    }
+                } else {
+                    tracing::error!("Reorg detector actor failed");
+                }
+                break
             }
+            else => break
```
slowli, line 450: Stupid question: What does `else` do here? [Tokio docs say](https://docs.rs/tokio/latest/tokio/macro.select.html) that it's used with conditional branches, but there are no conditional branches in this `select!`.

core/bin/external_node/src/main.rs
```diff
@@ -387,46 +379,77 @@ async fn main() -> anyhow::Result<()> {
...
+        particular_crypto_alerts,
+        graceful_shutdown,
+        tasks_allowed_to_finish,
+    );
+
+    tokio::pin!(wait_for_tasks);
+
+    loop {
+        tokio::select! {
+            _ = &mut wait_for_tasks => {
```
slowli, line 410: What happens if the same `select!`ed future is polled twice during multiple loop iterations? Will it panic? Will it resolve immediately? IIRC, [`futures::select!`](https://docs.rs/futures/latest/futures/macro.select.html) doesn't evaluate completed futures, but I'm not sure whether this is the case for `tokio::select!` (Tokio docs make it seem like it's not the case).

core/bin/external_node/src/main.rs
montekki: Yep should be removed.

core/bin/external_node/src/main.rs
montekki: For the concrete case you are pointing at here, all will be fine, tokio will be driving this function to completion, also you can do much more complicated stuff with it. https://tokio.rs/tokio/tutorial/select#loops

Deniallugo (approved):

core/bin/external_node/src/main.rs
slowli: I'll take your word for it. Just pointing out that the doc you're referencing confirms that polling `&mut _` future multiple times in a loop results in a panic:

> This example uses some new syntax. The first branch includes `, if !done`. This is a branch precondition. Before explaining how it works, let's look at what happens if the precondition is omitted. Leaving out `, if !done` and running the example results in the following output:
>> thread 'main' panicked at '`async fn` resumed after completion', src/main.rs:1:55
>> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

montekki: hmmmm

RomanBrodetski:
Taking a step back - why do we need such an (arguably) complicated logic? It's partly because we need to run the reorg detector in parallel. We have a similar case with Circuit Breaker on the main node. Is there any potential to extract some abstraction?

RomanBrodetski:
I also raise this because this is evidently error-prone - otherwise we wouldn't be doing this change

montekki:
@RomanBrodetski actually probably i can simplify the logic with `join` instead of a `select` inside a `loop` + clean shutdown, that will probably work.

core/bin/external_node/src/main.rs
```diff
@@ -402,46 +397,59 @@ async fn main() -> anyhow::Result<()> {
...
 
-    let reorg_detector = ReorgDetector::new(&main_node_url, connection_pool.clone());
-    let reorg_detector_handle = tokio::spawn(reorg_detector.run());
+    let reorg_detector = ReorgDetector::new(&main_node_url, connection_pool.clone(), stop_receiver);
+    let mut reorg_detector_handle = tokio::spawn(reorg_detector.run()).fuse();
 
     let particular_crypto_alerts = None;
     let graceful_shutdown = None::<futures::future::Ready<()>>;
     let tasks_allowed_to_finish = false;
+    let mut reord_detector_last_correct_batch = None;
```
popzxc, line 411: Nit: reorg.

core/bin/external_node/src/main.rs
```diff
@@ -402,46 +397,59 @@ async fn main() -> anyhow::Result<()> {
...
-                    None,
-                    connection_pool,
-                    L1ExecutedBatchesRevert::Allowed,
-                );
-                reverter
-                    .rollback_db(last_correct_batch, BlockReverterFlags::all())
-                    .await;
-                tracing::info!("Rollback successfully completed, the node has to restart to continue working");
-                return Ok(());
+            reord_detector_last_correct_batch = last_correct_batch;
```
popzxc, line 420: Nit: Formatting is off on this line (sadly, `rustfmt` ain't no friend of macros and generally in a pretty poor state rn).

montekki: right right.

popzxc:
@montekki I think it's a good solution for now. A proper task management entity is in the plans too.

etc/env/ext-node.toml
```diff
@@ -32,10 +32,6 @@ api_namespaces = ["eth", "web3", "net", "pubsub", "zks", "en", "debug"]
 bootloader_hash="0x0100038581be3d0e201b3cc45d151ef5cc59eb3a0f146ad44f0f72abf00b594c"
 default_aa_hash="0x0100038dc66b69be75ec31653c64cb931678299b9b659472772b2550b703f41c"
 
-# Whether to try using MultiVM
-# WARN: Only to be used locally until the feature is fully tested.
-experimental_multivm_support = false
-
```
RomanBrodetski, line 38: was this removed on purpose?

Deniallugo (approved):

hatemosphere (approved):

PR #273, opened 2023-10-20
feat(hyperchains): Hyperchain command improvements (including docker related)

githubdoramon:
# What ❔

- Fixes some flow errors on the hyerpchain initializer
- Changes the hyperchain command name to a "zk stack" umbrella command
- Addes docker steup command - builds server and generates a docker compose file

## Why ❔

- Improves the experience of starting hyperchains

## Checklist


infrastructure/zk/src/docker.ts
```diff
@@ -138,6 +143,10 @@ export async function build(image: string, cmd: Command) {
     await dockerCommand('build', image, cmd.customTag);
 }
 
+export async function customBuildForHyperchain(image: string, imageName: string) {
+    await dockerCommand('build', image, '', false, imageName);
+}
```
popzxc, line 148: Are we OK with the fact that we can build the hyperchain image via `zk`, but cannot push?
Right now it's hard for me to comprehend how will we introduce a generic solution on pushing images, since we have a significant number of performance optimizations implemented to make prover images work fast.

What's the plan here?

infrastructure/zk/src/hyperchain_wizard.ts
```diff
@@ -567,6 +575,67 @@ export function getTokens(network: string): L1Token[] {
...
+
+    await docker.customBuildForHyperchain('server-v2', envName);
+
+    await warning(`\n\n\n\n\nDocker image for server created: Server image: ${envName}/server-v2:latest\n`);
+
+    let hasProver = false;
+
+    if (process.env.ETH_SENDER_SENDER_PROOF_SENDING_MODE !== 'SkipEveryProof') {
+        hasProver = true;
+        // TODO: Hyperchain is using prover, so we must include Boojum images - wait for Boojum merge
```
popzxc, line 607: As mentioned above, using real provers requires a different config, so it probably also must be handled.
Also, same w.r.t. `TODO`s and task numbers.

infrastructure/zk/src/hyperchain_wizard.ts
```diff
@@ -567,6 +575,67 @@ export function getTokens(network: string): L1Token[] {
...
+        hasProver
+    };
+
+    const templateFileName = 'docker-compose-hyperchain-template';
+
+    const templateString = fs.existsSync(templateFileName) && fs.readFileSync(templateFileName).toString().trim();
+
+    var template = Handlebars.compile(templateString);
+
+    var result = template(composeArgs);
```
popzxc, line 623: `var` is bad. Why not `const`?

infrastructure/zk/src/hyperchain_wizard.ts
```diff
@@ -567,6 +575,67 @@ export function getTokens(network: string): L1Token[] {
...
+
+    const composeArgs = {
+        envFilePath: `./etc/env/${envName}.env`,
+        serverImageName: envName,
+        hasProver
+    };
+
+    const templateFileName = 'docker-compose-hyperchain-template';
+
+    const templateString = fs.existsSync(templateFileName) && fs.readFileSync(templateFileName).toString().trim();
```
popzxc, line 619: The spacing feels a bit odd here, why an empty line between seemingly related lines?

infrastructure/zk/src/hyperchain_wizard.ts
githubdoramon: bad copy/pasta. Will change

infrastructure/zk/src/hyperchain_wizard.ts
githubdoramon: No good reason at all. Will change

docker-compose-hyperchain-template
```diff
@@ -0,0 +1,158 @@
...
+      dockerfile: ./docker/prover-fri-gateway/Dockerfile
+    env_file:
+      - ./common.env
+    env:
+      FRI_PROVER_GATEWAY_API_URL: http://zkstack-core:3075
+    ports: # assumes default ports in .env
+      - "3312:3312" # prometheus metrics
+    volumes:
+      - prover_artifacts:/etc_prover_artifacts
+      - server_artifacts:/etc_server_artifacts
```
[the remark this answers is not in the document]
githubdoramon: The initial idea is to allow people to run boojum locally on CPU indeed

infrastructure/zk/package.json
```diff
@@ -26,6 +26,7 @@
         "@types/node": "^14.6.1",
         "@types/node-fetch": "^2.5.7",
         "@types/tabtab": "^3.0.1",
+        "handlebars": "^4.7.8",
```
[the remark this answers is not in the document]
githubdoramon: Good point.

infrastructure/zk/src/docker.ts
```diff
@@ -25,7 +25,8 @@ async function dockerCommand(
     command: 'push' | 'build',
     image: string,
     customTag?: string,
-    publishPublic: boolean = false
+    publishPublic: boolean = false,
+    hyperchainName?: string
```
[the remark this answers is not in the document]
githubdoramon: way better approach indeed

infrastructure/zk/src/docker.ts
githubdoramon: I think it is totally ok. Generic pushing would be too complex for now. If someone knows how to build, they most likely can figure out pushing the tagged image themselves

popzxc: But then wouldn't it make more sense for them to also build the images themselves? E.g. they may want to have their own tagging logic.
My main concern is that this functionality doesn't seem to be fitting; e.g. it does *half* the job, and if you still have to do the other half youself, then why bother? Overall, all it does is passes some flags to the `docker build` command; might have been simply covered in docs. This way it will probably be less confusing, e.g. "you build -> you push" instead of "zk does some magic -> you have to reverse-engineer this magic -> you push".

githubdoramon: But the build part is more than just firing the command, it does have some specific logic for some components. But you do make a good point on making it more customizable. Changed a little bit, let me know what you think

githubdoramon: Developer is only forced to use it If they are customizing their hyperchain. So they need to build the image.

popzxc: No objections, just pointing out that we may improve things on this front to be friendlier to the developers.

etc/hyperchains/docker-compose-hyperchain-template
```diff
@@ -0,0 +1,158 @@
...
+# ~16+ CPU cores
+# ~384+ GB RAM
+# This assumes the following:
+# - This Docker Compose is running from https://github.com/matter-labs/zksync-era/tree/boojum-integration
+# - Hyperchain Wizard was already ran and common.env file is generated in this dir
+# - DB is available/ready/initialized and common.env is pointing to it
+# - Contracts are already built and setup key is downloaded (most of the things should be done in Hyperchain Wizard)
+# reference: https://github.com/matter-labs/zksync-era/blob/main/.github/workflows/build-core-template.yml#L67-L72
+# - Data keys are already downloaded
+# from: https://console.cloud.google.com/storage/browser/matterlabs-zksync-v2-infra-blob-store/prover_setup_data/2d33a27?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&orgonly=true&project=matterlabs-infra&supportedpurview=organizationId&prefix=&forceOnObjectsSortingFiltering=false
```
popzxc, line 11: Is this URL correct? I tried using it in incognito mode and it asks me to log in. Shouldn't it be a public one?

infrastructure/zk/src/docker.ts
```diff
@@ -138,6 +140,10 @@ export async function build(image: string, cmd: Command) {
     await dockerCommand('build', image, cmd.customTag);
 }
 
+export async function customBuildForHyperchain(image: string, imageName: string) {
+    await dockerCommand('build', image, '', false, imageName);
```
popzxc, line 144: Nit: `image` and `imageName` look confusing here, probably should be `image` and `dockerOrg` (or at least `hyperchainName`).

popzxc (approved):

etc/hyperchains/docker-compose-hyperchain-template
hatemosphere: @githubdoramon i'd rather remove everything from here

etc/hyperchains/docker-compose-hyperchain-template
```diff
@@ -0,0 +1,158 @@
...
+# to: ./prover_setup-data
```
hatemosphere, line 12: should this be if'ed by "cpu prover" condition"?

infrastructure/zk/src/docker.ts
githubdoramon: ops. Totally

etc/hyperchains/docker-compose-hyperchain-template
githubdoramon: @hatemosphere ?

githubdoramon: ah, damn, didn't see your comment before pushing

popzxc (approved):

PR #274, opened 2023-10-20
fix(witness-generator): Witness generator oracle with cached storage refunds

perekopskiy:
# What ❔

During VM run in state keeper we save all storage refunds that are returned by oracle. Later in basic witness generation these refunds are reused.
Also, `zk_evm` is updated so that it now passes correct timestamps for `estimate_refunds_for_write` queries. Correct timestamps are necessary for VM storage oracle to rollback `returned_refunds`

## Why ❔

With new VM, pubdata tracer mutates oracle affecting storage refunds. As this tracer is only run in state keeper and not in witness generator we must save refunds from state keeper and use them in wit-gen.

## Checklist


Deniallugo (approved):

Deniallugo (dismissed):

core/lib/dal/src/blocks_dal.rs
```diff
@@ -241,6 +241,30 @@ impl BlocksDal<'_, '_> {
...
 
+    pub async fn get_storage_refunds(
+        &mut self,
+        number: L1BatchNumber,
+    ) -> anyhow::Result<Option<Vec<u32>>> {
+        let Some(row) = sqlx::query!(
+            "SELECT storage_refunds FROM l1_batches WHERE number = $1",
+            number.0 as i64
+        )
+        .instrument("get_initial_bootloader_heap")
```
slowli, line 252: Wrong name; should be something like `get_storage_refunds`.

Deniallugo (approved):

PR #279, opened 2023-10-20
fix: Change no pending batches 404 error into a success response

igamigo:
# What ❔

When no pending batches were found on the server API, the response was a 404 which was not handled on the client (gateway) side and was treated as an HTTP client error, causing a misleading log line:

```2023-10-19T18:30:46.726413Z ERROR zksync_prover_fri_gateway::api_data_fetcher: HTTP request failed due to error: HTTP status client error (404 Not Found) for url (http://127.0.0.1:3320/proof_generation_data)```

When testing out the gpu proving pipeline for the ZK stack effort, the error was hit soon after starting all services.
This PR changes this so in info trace is printed stating that there are currently no pending batches to be proven.  

## Why ❔

There is some semantic dissonance on receiving an HTTP client error and a log line when calling the API, since it's possible and reasonable that there are no pending batches at some points.

## Checklist


popzxc (approved):
Thanks!

mm-zk (approved):

PR #287, opened 2023-10-23
chore: updated era-zkevm_opcode_defs dependency

pompon0:
# What ❔

Updated the era-zkevm_opcode_defs dependency and minimal set of other dependencies which required update to a compatible version. This unlocked updated sha2 and sha3 dependencies to 0.10.8.
Prover integration test run for extra safety: https://github.com/matter-labs/zksync-era-workflows/actions/runs/6612607366

## Why ❔

To include https://github.com/matter-labs/era-zkevm_opcode_defs/pull/16 which in turn allows us to depend on up-to-date versions of sha2 and sha3.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,13 +9,13 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "617a8268e3537fe1d8c9ead925fca49ef6400927ee7bc26750e90ecee14ce4b8"
 dependencies = [
  "bitflags 1.3.2",
- "bytes 1.4.0",
+ "bytes 1.5.0",
  "futures-core",
  "futures-sink",
  "memchr",
  "pin-project-lite",
  "tokio",
- "tokio-util 0.7.8",
+ "tokio-util 0.7.9",
  "tracing",
 ]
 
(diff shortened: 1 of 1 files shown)


Cargo.lock
```diff
@@ -1193,12 +1177,10 @@ dependencies = [
 [[package]]
 name = "circuit_definitions"
 version = "0.1.0"
-source = "git+https://github.com/matter-labs/era-zkevm_test_harness.git?branch=v1.4.0#1934433af48520ac70dc3080b0d1feba877a07c5"
+source = "git+https://github.com/matter-labs/era-zkevm_test_harness.git?branch=v1.4.0#43aeb53d7d9c909508a98f9fc140edff0e9d2357"
```
popzxc, line 1180: Changes like this are extremely dangerous. There is no notion of semver and stuff in many crypto repos, so simply changing a git revision may be a breaking change requiring an update.

I would be extremely scared if I had to merge this PR, chances of it breaking everything are very high.

pompon0: So what's the right approach of upgrading stuff around here?

popzxc: Well, I think you won't like the answer. We normally update deps 1 by 1 when strictly necessary 😓

popzxc: (...and sometimes still break stuff)

pompon0: I've consulted @akash-chandrakar . I've manually bumped the minimal set of packages and run the prover integration test against it.

popzxc (approved):

PR #289, opened 2023-10-23
Update CHANGELOG.md

nikcode9:
it was just kept blank added it.


popzxc:
Thank you for the contribution; though we generate the changelog automatically based on PR titles.
We're currently in transition to developing fully in the open, so some integrations don't work as they should yet.
This issue is known and will be fixed at some point.

PR #292, opened 2023-10-23
fix(db): Fix write stalls in RocksDB (for real this time)

slowli:
# What ❔

[A previous fix](https://github.com/matter-labs/zksync-era/pull/265) didn't really work judging by Merkle tree behavior on the stage env. This PR makes the initialization timeout configurable (and increases the default value from 10s to 30s; 30s is approximately equal to the compaction duration) and slightly increases the number of retries on stalled writes.

## Why ❔

Having write stalls leads to panics and is obviously bad.

## Checklist

### DIFF
--- a/core/bin/external_node/src/config/mod.rs
+++ b/core/bin/external_node/src/config/mod.rs
@@ -194,10 +194,9 @@ pub struct OptionalENConfig {
     /// large value (order of 512 MiB) is helpful for large DBs that experience write stalls.
     #[serde(default = "OptionalENConfig::default_merkle_tree_memtable_capacity_mb")]
     merkle_tree_memtable_capacity_mb: usize,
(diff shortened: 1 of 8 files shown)


slowli:
Couple of other options we might consider:

- Increase [the number of level-0 SST files for stopped writes](https://docs.rs/rocksdb/latest/rocksdb/struct.Options.html#method.set_level_zero_stop_writes_trigger). I feel that it doesn't solve the problem, just postpones it.
- Add a subcommand to `rocksdb_util` (`compact`?) that would just create a RocksDB instance and wait indefinitely until its writes are not stopped. This could work well with the main node, but AFAICT, `rocksdb_util` is currently not shipped with the EN Docker image and making it work there could require some work.

RomanBrodetski (dismissed):
Very good, let's keep trying!

popzxc (approved):

PR #293, opened 2023-10-23
feat: Port boojum eth-sender changes

perekopskiy:
# What ❔

- `MetadataCalculatorModeConfig::Full` changed so it optionally saves artifacts for witness to GCS, EN runs metadata calculator in full mode, so it has all commitments
- Circuit breaker for facet selectors removed -- it won't be possible to perform an upgrade with it running.
- Eth-sender, block-reverter, consistency-checker use L1 contract function that corresponds to the protocol version of L1 batch it's used for. 

## Why ❔

Preparation for boojum upgrade

## Checklist


core/lib/types/src/commitment.rs
```diff
@@ -131,36 +131,40 @@ impl L1BatchWithMetadata {
...
+                ),
+                Token::Array(
+                    self.factory_deps
+                        .iter()
+                        .map(|bytecode| Token::Bytes(bytecode.to_vec()))
+                        .collect(),
+                ),
+            ])
+        } else {
+            unimplemented!()
```
perekopskiy, line 166: I'll implement it after https://github.com/matter-labs/zksync-era/pull/259 is merged

core/lib/contracts/src/lib.rs
```diff
@@ -393,3 +393,314 @@ impl BaseSystemContracts {
         }
     }
 }
+
+pub static COMMIT_BLOCKS_FUNCTION: Lazy<Function> = Lazy::new(|| {
```
Deniallugo, line 397: Could we read it from file?

StanislavBreadless, line 397: Maybe call it `PRE_BOOJUM_COMMIT_BLOCKS_FUNCTION`?

core/lib/zksync_core/src/consistency_checker/mod.rs
```diff
@@ -97,11 +98,14 @@ impl ConsistencyChecker {
             Some(1.into()),
             "Main node gave us a failed commit tx"
         );
-
-        let commitments = self
-            .contract
-            .function("commitBlocks")
-            .unwrap()
+        let commit_function =
+            if block_metadata.header.protocol_version.unwrap() <= ProtocolVersionId::Version16 {
```
StanislavBreadless, line 102: `is_pre_boojum`?

core/lib/types/src/commitment.rs
```diff
@@ -131,36 +131,40 @@ impl L1BatchWithMetadata {
...
-                    .collect(),
-            ),
-            Token::Array(
-                self.factory_deps
-                    .iter()
-                    .map(|bytecode| Token::Bytes(bytecode.to_vec()))
-                    .collect(),
-            ),
-        ])
+        if self.header.protocol_version.unwrap().is_pre_boojum() {
```
Deniallugo, line 134: Just thinking out of loud. 
Should we make commitment generic over the version and make different implementation depends on it? 

And it touches all aspect of working with contracts. Probably a trait that has methods `encode_commtiment`, `generate_proof_calldata` e.t.c.

We always have kinda the same interface   with some internal differences and seems we can reflect them in more generic way. 

Now we randomly have `is_pre_boojum` checks

core/lib/zksync_core/src/consistency_checker/mod.rs
Deniallugo, line 102: why you are not using `is_pre_boojum` ?

StanislavBreadless (approved):

Deniallugo (commented):
Let's discuss the idea with generics over the encoding data for sending it to L1. 


Seems pretty useful for future changes

core/lib/contracts/src/lib.rs
perekopskiy: `BLOCKS` in name implicitly states it's pre-boojum as for post-boojum we have `BATCHES`, but I agree that `PRE_BOOJUM` prefix would look better. I will change it to `PRE_BOOJUM_COMMIT_FUNCTION`

perekopskiy: Not from the file we used to read from as in `contracts` submodule there will be only new contracts. I can move old ABI from here to a separate JSON file but I'm not sure where is a good place for it

Deniallugo: not yet exists, i hope once we will have it

core/lib/types/src/commitment.rs
perekopskiy: I agree, but the logic of creating calldata is convoluted: selector is chosen in eth_tx_aggregator, abi encoding happens in `aggregated_operations.rs`, `commitment.rs`. It will be quite difficult to refactor it, so I suggest we come back to it after boojum upgrade

akash-chandrakar:
@perekopskiy I think I didn't follow this comment:

> MetadataCalculatorModeConfig::Full changed so it optionally saves artifacts for witness to GCS

If I am reading it right it means **the witness-inputs to GCS are not always present**,  if that's the case do you think it will prevent us from generating proof for batches where the GCS witness input data is not present.

perekopskiy:
@akash-chandrakar I meant, that made saving to GCS configurable. It must be enabled for main server but disabled for EN

Deniallugo (approved):
I'm not fully satisfied with the current state. 
I still want to improve upgrade on ETH-Senders side. Now it's a bit painfull. 

And we have to think more about how to develop features such as boojum in a nicer way.

PR #301, opened 2023-10-24
fix(db): Fix root cause of RocksDB misbehavior

slowli:
# What ❔

Fixes (hopefully) the root cause of recent RocksDB misbehavior. After recent refactoring, it improperly cancels background compactions / flushes each time a `RocksDB` instance is dropped, which leads to compactions / flushes being constantly interrupted. (After refactoring, `RocksDB` instances are `Clone`eable and act essentially as `Arc` wrappers; a new instance is created and dropped, in particular, [when processing](https://github.com/matter-labs/zksync-era/blob/ec41692063909742005acd656694a8f0838fcf33/core/lib/zksync_core/src/metadata_calculator/updater.rs#L385) each chunk of L1 batches.) A proper solution, implemented by this PR, is to cancel background work when dropping an internal, non-`Arc`'d DB instance.

## Why ❔

That's a bug that makes RocksDB behave *very* unstably, as witnessed by recent issues with it.

## Checklist

### DIFF
--- a/core/lib/storage/src/db.rs
+++ b/core/lib/storage/src/db.rs
@@ -448,13 +448,21 @@ impl<CF: NamedColumnFamily> RocksDB<CF> {
 
         let raw_batch_bytes = raw_batch.data().to_vec();
         let mut retries = self.stalled_writes_retries.intervals();
+        let mut stalled_write_reported = false;
+        let started_at = Instant::now();
         loop {
             match self.write_inner(raw_batch) {
-                Ok(()) => return Ok(()),
+                Ok(()) => {
+                    if stalled_write_reported {
+                        METRICS.observe_stalled_write_duration(CF::DB_NAME, started_at.elapsed());
+                    }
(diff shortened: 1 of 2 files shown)


slowli:
In hindsight, the issue was obvious; recent RocksDB logs are littered with "Shutdown: canceling all background work" logs, which I couldn't attribute previously.

popzxc (dismissed):
~~I hoped that the PR would be titled *RocksDB Wars: Episode IV -- The Stalled Writes Strike Back*~~

Nice and certainly makes sense. If this works out, would it make sense to roll back the new code added in the previous 3 PRs? Less code is generally better than more code.

slowli:
@popzxc We have write stalls reported for the state keeper cache, so I think we should leave this logic as-is for now. New RocksDB metrics also make sense, IMO; for one, they'll allow to estimate whether this fix works.

RomanBrodetski (approved):

PR #321, opened 2023-10-26
feat(house_keeper): Remove GCS Blob Cleaner

EmilLuta:
# What ❔

This feature is replaced with a 30 days TTL policy in GCS. The policy is already applied, this work cleans unnecessary component on core side.

## Why ❔

Unnecessary work on infra resources. Additionally, the implementation closely ties the codebase to GCS.

## Checklist


popzxc (approved):

popzxc:
Probably we should document somewhere that we expect GCS TTL to be configured, but I'm not sure where exactly.

akash-chandrakar (dismissed):

popzxc:
I agree that it doesn't have to be done in this PR.
I also agree that infra-agnostic docs will be good to have -- we aim to be fully infra-agnostic via zkStack in the future; so knowing that there is no embedded blob cleaning functionality may be useful to some folks (imagine someone running their own object store on bare metal).

akash-chandrakar (dismissed):

Deniallugo (dismissed):

core/lib/dal/src/witness_generator_dal.rs
```diff
@@ -750,7 +750,7 @@ impl WitnessGeneratorDal<'_, '_> {
             let job_ids = sqlx::query!(
```
RomanBrodetski, line 750: wait, do we need this method at all?

core/lib/dal/src/witness_generator_dal.rs
```diff
@@ -844,57 +810,6 @@ impl WitnessGeneratorDal<'_, '_> {
         }
     }
 
-    pub async fn mark_leaf_aggregation_gcs_blobs_as_cleaned(&mut self, l1_batch_numbers: Vec<i64>) {
-        {
-            sqlx::query!(
-                r#"
-                UPDATE leaf_aggregation_witness_jobs
-                SET is_blob_cleaned=TRUE
```
popzxc, line 852: A few questions that are probably out of the scope for this PR.
1. Do we need a migration where we drop this column?
2. Also, do we want to keep all the records in this table forever, or maybe we want to drop old entries after some time?

popzxc (approved):

RomanBrodetski (dismissed):

RomanBrodetski (dismissed):

montekki (approved):

popzxc (approved):

PR #351, opened 2023-10-30
chore(vm): apply new interface for old vms

Deniallugo:
# What ❔

Apply new vm interface for all old vms
Apply new vm interface for multivm

Introduce a few macroses for simple dispatching between vms and tracers 

## Why ❔

Part of the refactoring and applying the new interface everywhere 

## Checklist


core/lib/multivm/src/versions/vm_latest/tracers/refunds.rs
```diff
@@ -50,9 +51,10 @@ pub(crate) struct RefundsTracer {
     gas_spent_on_bytecodes_and_long_messages: u32,
     l1_batch: L1BatchEnv,
     pubdata_published: u32,
+    _phantom: PhantomData<S>,
 }
 
-impl RefundsTracer {
+impl<S> RefundsTracer<S> {
```
slowli, line 57: Is this type argument required just for `dispatch_tracers!` to work out of the box in `impl Tracer for Tracer for DefaultExecutionTracer`? If so, IMO, that's a weird reason to add a type param to a type; `dispatch_tracers!` could be improved to support more complex expressions, or could be removed altogether (see the other comment).

core/lib/multivm/src/interface/traits/vm.rs
```diff
@@ -53,13 +53,14 @@ use crate::interface::types::outputs::{
...
+use crate::interface::{FinishedL1Batch, VmMemoryMetrics};
 use crate::vm_latest::HistoryEnabled;
 use crate::HistoryMode;
-use zksync_state::{StoragePtr, WriteStorage};
+use zksync_state::StoragePtr;
 use zksync_types::Transaction;
 use zksync_utils::bytecode::CompressedBytecodeInfo;
 
-pub trait VmInterface<S: WriteStorage, H: HistoryMode> {
+pub trait VmInterface<S, H: HistoryMode> {
```
slowli, line 63: Not sure whether this is applicable in the current case, but it usually makes sense to split such large traits into smaller parts with well-defined boundaries. E.g., in this case methods that don't depend on a `TracerDispatcher` could be extracted to a separate trait that would be used as a super trait for `VmInterface`.

slowli (commented):
Some surface-level drive-by comments; hopefully, at least marginally helpful.

core/lib/multivm/src/versions/vm_latest/tracers/dispatcher_macros.rs
```diff
@@ -0,0 +1,24 @@
+#[macro_export]
+macro_rules! dispatch_tracers {
```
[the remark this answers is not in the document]
Deniallugo: It was a little painful to use several tracers at once.
and this boilerplate became huge. I think now it's a lot easier

core/lib/multivm/src/versions/vm_latest/tracers/refunds.rs
Deniallugo: From the one side yes, but even though without this requirements it was an ugly trait specification for calls with memory..

core/lib/multivm/src/interface/traits/vm.rs
Deniallugo: it's already splitted to two traits.
I'm not sure i want to introduce more of them..

popzxc (dismissed):

core/lib/multivm/src/versions/vm_refunds_enhancement/tracers/dispatcher.rs
```diff
@@ -13,6 +13,12 @@ pub struct TracerDispatcher<S: WriteStorage, H: HistoryMode> {
     tracers: Vec<TracerPointer<S, H>>,
 }
 
+impl<S: WriteStorage, H: HistoryMode> TracerDispatcher<S, H> {
+    pub fn new(tracers: Vec<TracerPointer<S, H>>) -> Self {
```
slowli, line 17: Off-topic: Maybe, it's personal, but using `_Pointer` suffix for a boxed trait object looks *bizarre* to me (yeah, I understand that `Box` is a smart pointer). `Boxed_` / `Box_` would look much more like a conventional name; cf. [`BoxFuture`](https://docs.rs/futures/latest/futures/future/type.BoxFuture.html) from the `futures` crate.

core/lib/multivm/src/versions/vm_latest/tracers/default_tracers.rs
```diff
@@ -44,27 +44,125 @@ pub(crate) struct DefaultExecutionTracer<S: WriteStorage, H: HistoryMode> {
...
+    }
+}
+
 impl<S: WriteStorage, H: HistoryMode> Debug for DefaultExecutionTracer<S, H> {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         f.debug_struct("DefaultExecutionTracer").finish()
     }
 }
 
+macro_rules! dispatch_tracers {
```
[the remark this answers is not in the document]
Deniallugo: Actually, that exactly what we want to avoid.. 

The initial idea was, that we have all tracers including the service one such as result together in one vector. But it caused serious performance degradation ~5% for each tracer.. That's why we've started  to move service tracers into default tracer.

core/lib/multivm/src/versions/vm_latest/tracers/refunds.rs
Deniallugo:
```
        <ResultTracer as DynTracer<S, SimpleMemory<H>>>::after_decoding(
            &mut self.result_tracer,
            state,
            data,
            memory,
        );

```
That's how it was before

core/lib/multivm/src/versions/vm_refunds_enhancement/tracers/dispatcher.rs
Deniallugo: So, actually here i want to keep a bit flexibility and play with different smart pointers. E.g. Rc<RefCell>. Initially it was boxed, as you suggest.

core/lib/multivm/src/versions/vm_latest/tracers/default_tracers.rs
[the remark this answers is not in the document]
slowli: Is the performance regression caused by dynamic dispatch? Didn't think it could have that significant impact. Dynamic dispatch could be replaced with a static dispatch via an enum, but it'd look no better than the current solution. Alternatively, one could try emulating `impl for<T: VmTracer<S, H>> FnMut(&mut T)`, but that also looks quite awkward.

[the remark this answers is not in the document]
Deniallugo: but it made =( 

So I don't think the current macros is bad :) and it seem to meet our needs.

core/lib/multivm/src/versions/vm_latest/tracers/refunds.rs
slowli: You can probably avoid specifying `SimpleMemory<H>` (i.e., use `DynTracer<S, _>`) in this particular case, which would simplify syntax a little. In general, I still think that using full method syntax is preferable to adding type params for a type in such cases; type params are appropriate if the type itself doesn't work without them, not if some method invocations require using full method syntax 🙃

Deniallugo: It actually simplifies the trait implementation.

core/lib/multivm/src/versions/vm_m6/vm.rs
```diff
@@ -0,0 +1,285 @@
...
+                let bytecode_hash = hash_bytecode(bytecode);
+                let is_known = !deps_hashes.insert(bytecode_hash)
+                    || self
+                        .vm
+                        .state
+                        .storage
+                        .storage
+                        .get_ptr()
+                        .borrow_mut()
+                        .is_bytecode_exists(&bytecode_hash);
```
slowli, line 197: Nit: This can get similar treatment to `is_bytecode_known` in VM 1.3.2.

slowli (dismissed):
I don't feel comfortable saying that these changes are correct, but they look good on my surface level of understanding.

montekki (dismissed):

montekki (approved):

PR #367, opened 2023-11-01
chore: fix file path

penghuarong:
# What ❔

- fix file path

## Why ❔

- fix file path

## Checklist


popzxc:
There is no error there, `base` corresponds to the templates of the configs, while `dev` is generated during `zk init` and corresponds to the real config. 

However, this section is outdated already and should be removed completely -- we will handle this a bit later. Thanks for flagging.

PR #371, opened 2023-11-01
feat(en): Implement gossip fetcher

slowli:
# What ❔

...i.e., a fetcher component that would use gossip network instead of JSON-RPC API.

## Why ❔

This can be used by external nodes to sync with the main node.

## Checklist


core/lib/zksync_core/src/sync_layer/gossip/conversions.rs
```diff
@@ -0,0 +1,88 @@
...
+};
+use zksync_types::block::ConsensusBlockFields;
+use zksync_types::{
+    api::en::SyncBlock, Address, L1BatchNumber, MiniblockNumber, ProtocolVersionId, H256,
+};
+
+use crate::sync_layer::fetcher::FetchedBlock;
+
+// FIXME: should use Protobuf
+#[derive(Debug, Serialize, Deserialize)]
```
slowli, line 17: IMO, using JSON serialization is fine until the logic is actually integrated into EN.

core/lib/zksync_core/src/sync_layer/gossip/conversions.rs
```diff
@@ -0,0 +1,88 @@
...
+    pub(super) fn from_gossip_block(block: &FinalBlock) -> anyhow::Result<Self> {
+        let number = u32::try_from(block.header.number.0)
+            .context("Integer overflow converting block number")?;
+        let payload: BlockPayload = serde_json::from_slice(&block.payload.0)
+            .context("Failed deserializing block payload")?;
+
+        Ok(Self {
+            number: MiniblockNumber(number),
+            l1_batch_number: payload.l1_batch_number,
+            protocol_version: ProtocolVersionId::latest(), // FIXME
```
slowli, line 74: I'm not sure where to get this field from. Consensus messages have their own protocol versions, but I'm not 100% sure whether they should be the same (if so, it needs to be changed on the consensus side and probably shouldn't be made a constant, since protocol versions are semi-frequently updated).

core/lib/zksync_core/Cargo.toml
```diff
@@ -39,6 +39,12 @@ zksync_health_check = { path = "../health_check" }
 vlog = { path = "../vlog" }
 
 multivm = { path = "../multivm" }
+# Consensus dependenices
+zksync_concurrency = { package = "concurrency", version = "0.1.0", git = "https://github.com/matter-labs/era-consensus.git", rev = "a0905d4896a12d60296e24130dee6a0c0b730c32" }
+zksync_consensus_schema = { package = "schema", version = "0.1.0", git = "https://github.com/matter-labs/era-consensus.git", rev = "a0905d4896a12d60296e24130dee6a0c0b730c32" }
+zksync_consensus_roles = { package = "roles", version = "0.1.0", git = "https://github.com/matter-labs/era-consensus.git", rev = "a0905d4896a12d60296e24130dee6a0c0b730c32" }
+zksync_consensus_storage = { package = "storage", version = "0.1.0", git = "https://github.com/matter-labs/era-consensus.git", rev = "a0905d4896a12d60296e24130dee6a0c0b730c32" }
+zksync_consensus_executor = { package = "executor", version = "0.1.0", git = "https://github.com/matter-labs/era-consensus.git", rev = "a0905d4896a12d60296e24130dee6a0c0b730c32" }
```
slowli, line 47: When are we changing package names in the consensus repo, btw?

core/lib/zksync_core/src/sync_layer/fetcher.rs
```diff
@@ -60,11 +102,78 @@ impl MainNodeFetcherCursor {
...
+                timestamp: block.timestamp,
+                // `block.virtual_blocks` can be `None` only for old VM versions where it's not used, so it's fine to provide any number.
+                virtual_blocks: block.virtual_blocks,
+            });
+            FETCHER_METRICS.miniblock.set(block.number.0.into());
+        }
+
+        // This detection is somewhat shaky. For now, the only empty miniblocks are fictive ones (i.e.,
+        // the last miniblock in an L1 batch); all other miniblocks must contain at least 1 transaction.
+        let last_in_batch = block.transactions.is_empty();
```
slowli, line 155: Not entirely sure whether this observation is something that we can rely upon, but it allows not to delay queueing commands in the gossip fetcher. (Otherwise, we'd need to delay queueing each empty L2 block to see whether the next L2 block has the same L1 batch number.)

popzxc: Attempt to not rely on this fact was the main reason I've added a corresponding field to `SyncBlock`. Technically, nothing prevents us from creating an empty miniblock (or even an empty batch), so I didn't feel comfortable with having such an assumption (esp given that EN is expected to be resilient).

slowli: Got it. Am I correct in that while not all empty miniblocks may be fictive, each fictive miniblock is certainly empty (has no transactions)? Otherwise, the gossip fetcher will be even more complex and slow (it'll have to delay *all* miniblocks, while with the logic I've described above it would only delay empty ones, which are in practice always fictive rn.)

popzxc: > each fictive miniblock is certainly empty (has no transactions)

This is not guaranteed on the protocol level. In fact, the protocol (e.g. bootloader and everything below it) doesn't even know anything about fictive miniblocks, they only exist to "park" some internal emitted VM events that don't belong to any transaction. This is purely a state keeper concept.

core/lib/zksync_core/src/sync_layer/gossip/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+
+/// Starts fetching L2 blocks using peer-to-peer gossip network.
+pub async fn start_gossip_fetcher(
+    pool: ConnectionPool,
+    actions: ActionQueueSender,
+    executor_config: ExecutorConfig,
+    node_key: node::SecretKey,
+    mut stop_receiver: watch::Receiver<bool>,
+) -> anyhow::Result<()> {
+    let result = scope::run!(&ctx::root(), |ctx, s| {
```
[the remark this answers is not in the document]
slowli: AFAIU (cc @pompon0 for expertise):

- The concurrency framework doesn't provide a runtime; rather, it runs on Tokio.
- It should be fully safe and sound to use regardless of the environment (in particular, "adjacent" runtimes).
- We don't have a choice given that consensus codebase extensively relies on the concurrency framework.

core/lib/zksync_core/src/sync_layer/fetcher.rs
popzxc: However, we can discuss how much trust do we have in this invariant (we _may_ make it a protocol invariant if we really need, though initially it was but a kludge to fix our explorer). At the very least, I think we can live with this assumption while in the PoC phase (to get something running quicker).

core/lib/zksync_core/src/sync_layer/gossip/mod.rs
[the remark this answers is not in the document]
pompon0: yes, it is ok.

core/lib/zksync_core/Cargo.toml
pompon0: https://github.com/matter-labs/era-consensus/pull/25

core/lib/zksync_core/src/sync_layer/gossip/conversions.rs
```diff
@@ -0,0 +1,91 @@
...
+    Ok(FinalBlock {
+        header,
+        payload,
+        justification,
+    })
+}
+
+impl FetchedBlock {
+    /// **Important.** `last_in_batch` is always set to `false` by this method; it must be properly
+    /// set by the outside logic (currently implemented by `CursorWithCachedBlock`).
```
pompon0, line 66: can we make last_in_batch an additional argument to from_gossip_block?

core/lib/types/src/block.rs
```diff
@@ -84,6 +84,16 @@ pub struct MiniblockHeader {
...
 }
 
+/// Consensus-related L2 block (= miniblock) fields.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ConsensusBlockFields {
+    /// Hash of the previous consensus block.
+    pub prev_block_hash: H256,
+    /// Protobuf serialization of quorum certificate for the block.
+    pub commit_qc_bytes: Bytes,
```
pompon0, line 94: can we make it a dedicated type? Either a wrapper of bytes, or import CommitQC struct from consensus crate directly?

core/lib/zksync_core/src/sync_layer/gossip/conversions.rs
pompon0: The current idea is that zksync-era protocol version is supposed to imply the ProtocolVersion in era-consensus (i.e. we can make it the same, just not all protocol changes will affect the consensus itself). You can extract the protocol version from commit_qc (it contains the ReplicaCommit message which contains the protocol_version field). It will be eventually set by the main node logic, currently it is just a constant.

core/lib/zksync_core/src/sync_layer/gossip/mod.rs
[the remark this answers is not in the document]
pompon0: More specifically, panics will be propagated in case an awaited task is executed on a runtime which gets shut down.

core/lib/zksync_core/src/sync_layer/gossip/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+mod storage;
+#[cfg(test)]
+mod tests;
+mod utils;
+
+use self::{buffered::Buffered, storage::PostgresBlockStorage};
+use super::{fetcher::FetcherCursor, sync_action::ActionQueueSender};
+
+/// Starts fetching L2 blocks using peer-to-peer gossip network.
+pub async fn start_gossip_fetcher(
```
pompon0, line 25: 'run_gossip_fetcher'?

core/lib/zksync_core/src/sync_layer/gossip/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+            if stop_receiver.changed().await.is_err() {
+                tracing::warn!(
+                    "Stop signal sender for gossip fetcher was dropped without sending a signal"
+                );
+            }
+            s.cancel();
+            tracing::info!("Stop signal received, gossip fetcher is shutting down");
+            Ok(())
+        });
+        start_gossip_fetcher_inner(ctx, pool, actions, executor_config, node_key)
```
pompon0, line 43: a more idiomatic way of expressing that would be:
```
scope::run({
  s.spawn_bg(start_gossip_fetcher_inner(...));
  if stop_receiver.changed().await.is_err() { ... }
  tracing::info!("signal received ...");
  Ok(())
})
```

core/lib/zksync_core/src/sync_layer/gossip/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+    })
+    .await;
+
+    result.or_else(|err| {
+        if err.root_cause().is::<ctx::Canceled>() {
```
pompon0, line 48: can we avoid inspecting the error? I.e. either catch cancellation internally or make the scope::run! return Canceled rather than anyhow (or some other strongly typed error).

core/lib/zksync_core/src/sync_layer/gossip/storage.rs
```diff
@@ -0,0 +1,351 @@
...
+                .map_err(StorageError::Database)?;
+            self.block_sender.send_if_modified(|number| {
+                if *number != sealed_miniblock_number {
+                    *number = sealed_miniblock_number;
+                    true
+                } else {
+                    false
+                }
+            });
+            if ctx.sleep(POLL_INTERVAL).await.is_err() {
```
pompon0, line 94: nit: to indicate that the expecte error is cancellation, you can write it as: `if let Err(ctx::Canceled) = ctx.sleep() { ... }`

core/lib/zksync_core/src/sync_layer/gossip/storage.rs
```diff
@@ -0,0 +1,351 @@
...
+        &self,
+        ctx: &ctx::Ctx,
+        number: BlockNumber,
+    ) -> StorageResult<Option<FinalBlock>> {
+        let get_block = async {
+            let number = u32::try_from(number.0).context("block number is too large")?;
+            let mut storage = self.storage().await?;
+            Self::block(&mut storage, MiniblockNumber(number)).await
+        };
+        tokio::select! {
```
pompon0, line 187: I think that at this point we should make `Ctx::wait` public (https://github.com/matter-labs/era-consensus/blob/ccc3ee909d6da397f94fc43898ae0df20cd5f94f/node/libs/concurrency/src/ctx/mod.rs#L196C4-L196C29) with a disclaimer that it should only be used on cancel-safe context-unaware futures.

core/lib/zksync_core/src/sync_layer/gossip/storage.rs
```diff
@@ -0,0 +1,351 @@
...
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use rand::{thread_rng, Rng};
+
+    use zksync_concurrency::scope;
+    use zksync_types::L2ChainId;
```
pompon0, line 229: I know, that zksync-era isn't very consistent on that, but can we move towards having tests in a separate file?

core/lib/zksync_core/src/sync_layer/gossip/conversions.rs
slowli: I've created a task for this in Linear. Pragmatically, I'd prefer to leave a FIXME here because otherwise we'd need to mock retrieval of the past protocol versions in tests.

slowli:
I'd prefer to implement and test syncing from a non-genesis miniblock in a follow-up PR. AFAICT at this point, this would comprise the following items:

- Load the first block with consensus fields from Postgres when a node starts and comparing it against the reference block. If there are no blocks with consensus fields, the node will persist it from the reference.
- Use this first block in `BlockStore::first_block()` rather than block 0.

Most of PR should be dedicated to testing this functionality.

core/lib/zksync_core/src/sync_layer/gossip/storage/mod.rs
```diff
@@ -0,0 +1,225 @@
...
+#[async_trait]
+impl ContiguousBlockStore for PostgresBlockStorage {
+    async fn schedule_next_block(&self, ctx: &ctx::Ctx, block: &FinalBlock) -> StorageResult<()> {
+        // last_in_batch` is always set to `false` by this call; it is properly set by `CursorWithCachedBlock`.
+        let fetched_block =
+            FetchedBlock::from_gossip_block(block, false).map_err(StorageError::Database)?;
+        let actions = sync::lock(ctx, &self.cursor).await?.advance(fetched_block);
+        ctx.wait(async {
+            for actions_chunk in actions {
+                self.actions.push_actions(actions_chunk).await;
```
pompon0, line 219: is push_actions cancel safe? Is executing this loop partially cancel safe?

pompon0 (approved):

core/lib/zksync_core/src/sync_layer/gossip/storage/mod.rs
slowli: Executing a loop partially should be cancel-safe (each `actions_chunk` corresponds to a separate L2 block, so it is OK if a further chunk is lost). There is a caveat: it ceases to be cancel-safe if we assume that the surrounding struct (`PostgresBlockStorage`) can be used after cancellation – which we currently don't and likely don't ever plan to, but that's non-local reasoning. If we assume this possibility, than executing a loop partially is **not** cancel-safe since it'll lead to a mismatch between the state of the action queue and the cursor (i.e., `actions` and `cursor` fields).

`push_actions` is somewhat similar – if we assume that the action queue consumer stops together with cancellation and doesn't resume, we should be fine. In the general case, losing some actions in the queue will likely lead to panics or maybe even to the corrupted node state (not sure whether it's possible to trigger).

Given all this, would you advise not to use `ctx.wait()` here?

pompon0: Yes, I'd suggest to await unconditionally if cancellation would break the state, as long as push_actions is known to eventually complete.

pompon0: otherwise annotate the whole schedule_next_block method that it invalidates the PostgresBlockStorage object if cancelled.

core/lib/zksync_core/src/sync_layer/gossip/buffered/mod.rs
```diff
@@ -0,0 +1,320 @@
...
+            blocks: BTreeMap::new(),
+        }
+    }
+
+    fn head_block(&self) -> Option<FinalBlock> {
+        self.blocks.values().next_back().cloned()
+    }
+
+    #[tracing::instrument(level = "trace", skip(self))]
+    fn set_store_block(&mut self, store_block_number: BlockNumber) {
```
RomanBrodetski, line 64: do I understand it right that we should invoke this method when the corresponding block is persisted - and as soon as it happens, we don't need it in buffer anymore? Either way, can you please elaborate on when exactly we need to call this method? (I'll keep reading the PR and maybe understand myself)

core/lib/zksync_core/src/sync_layer/fetcher.rs
```diff
@@ -60,11 +104,75 @@ impl MainNodeFetcherCursor {
...
+    pub(super) fn advance(&mut self, block: FetchedBlock) -> Vec<SyncAction> {
+        assert_eq!(block.number, self.next_miniblock);
+
+        let mut new_actions = Vec::new();
+        if block.l1_batch_number != self.l1_batch {
+            assert_eq!(
+                block.l1_batch_number,
+                self.l1_batch.next(),
+                "Unexpected batch number in the next received miniblock"
+            );
```
montekki, line 122: A DQ: do we trust the sender to always correctly send this block or anyone can connect to us via gossip, send random block l1 batch number and DoS the node?

RomanBrodetski (dismissed):
hey @slowli I read the PR and the code looks good, but I would def benefit from some mid-level data flow description. Maybe just a couple of sentences comparing the current approach with the one you introduce

core/lib/zksync_core/src/sync_layer/gossip/buffered/mod.rs
```diff
@@ -0,0 +1,320 @@
...
+        assert!(
+            store_block_number > self.store_block_number,
+            "`ContiguousBlockStore` invariant broken: unexpected new head block number"
+        );
```
montekki, line 68: Same qestion here and further down the code: is DoS possible?

core/lib/zksync_core/src/sync_layer/fetcher.rs
slowli: A very good question! The L1 batch number is covered by the block quorum certificate (proof of authenticity, jointly signed by the validators). So a node cannot fake this number w/o the quorum certificate becoming incorrect, which [will be detected](https://github.com/matter-labs/era-consensus/blob/9f58e688e70f277550b14cb2a9bc343c874bb0eb/node/actors/sync_blocks/src/peers/mod.rs#L466-L469) on the consensus side, and the block will be rejected. If the L1 batch number is incorrect with a correct quorum certificate, we have a compromised supermajority of validators, or there's something seriously wrong with the node (e.g., its deserialization logic is malfunctioning), so panicking seems like a natural step at this point.

core/lib/zksync_core/src/sync_layer/gossip/buffered/mod.rs
slowli: No, this assertion just encapsulates the `ContiguousBlockStore` invariant (blocks are saved in order of their numbers; this is true for the state keeper now and in the observable future – even if we somehow manage to parallelize transaction processing within a block, we'd still probably execute and persist blocks in order). The external factors like node peers cannot influence this invariant; if it's broken, it means that there's a bug in the node implementation.

core/lib/zksync_core/src/sync_layer/fetcher.rs
montekki: > we have a compromised supermajority of validators

So a compromised supermajority of validators will bring down all honest nodes which is theoretically exactly what they would want?

slowli: If we have a compromised supermajority of validators, we have *far* more problems besides crashing non-validating nodes. E.g., they can issue different versions of the L2 blockchain to different nodes, stop processing some or all transactions altogether etc.

core/lib/zksync_core/src/sync_layer/gossip/buffered/mod.rs
slowli: I've added a couple of doc comments explaining the data flow.

core/lib/zksync_core/src/sync_layer/fetcher.rs
montekki: My understanding of PoS (if we are aiming for one) is:

>  they can issue different versions of the L2 blockchain to different nodes

That is an equivocation and a proper PoS should have slashing mechanisms for that

> stop processing some or all transactions altogether etc.

That is a liveness check and dead validators should also be slashed.

Slashing would be impossible if all honest nodes are crashing, but I won't push this offtopic any further.

montekki (approved):

PR #390, opened 2023-11-02
feat(state-keeper): Disable some seal criteria for boojum

perekopskiy:
# What ❔

Disables `InitialWritesCriterion`, `RepeatedWritesCriterion`, `L2ToL1LogsCriterion` for post-boojum blocks by setting limit to `usize::max`.

## Why ❔

Some circuits are no longer single-instance in boojum

## Checklist

### DIFF
--- a/core/lib/zksync_core/src/api_server/tx_sender/mod.rs
+++ b/core/lib/zksync_core/src/api_server/tx_sender/mod.rs
@@ -907,7 +907,13 @@ impl<G: L1GasPriceProvider> TxSender<G> {
         };
 
         let seal_data = SealData::for_transaction(transaction, tx_metrics);
-        if let Some(reason) = ConditionalSealer::find_unexecutable_reason(sk_config, &seal_data) {
+        // Using `ProtocolVersionId::latest()` for a short period we might end up in a scenario where the StateKeeper is still pre-boojum
+        // but the API assumes we are post boojum. In this situation we will determine a tx as being executable but the StateKeeper will
+        // still reject them as it's not.
+        let protocol_version = ProtocolVersionId::latest();
+        if let Some(reason) =
+            ConditionalSealer::find_unexecutable_reason(sk_config, &seal_data, protocol_version)
+        {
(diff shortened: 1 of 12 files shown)


koloz193 (approved):

core/lib/zksync_core/src/api_server/tx_sender/mod.rs
```diff
@@ -907,7 +907,13 @@ impl<G: L1GasPriceProvider> TxSender<G> {
         };
 
         let seal_data = SealData::for_transaction(transaction, tx_metrics);
-        if let Some(reason) = ConditionalSealer::find_unexecutable_reason(sk_config, &seal_data) {
+        // Using `ProtocolVersionId::latest()` for a short period we might end up in a scenario where the StateKeeper is still pre-boojum
+        // but the API assumes we are post boojum. In this situation we will determine a tx as being executable but the StateKeeper will
+        // still reject them as it's not.
+        let protocol_version = ProtocolVersionId::latest();
```
Deniallugo, line 913: Seems, that we have to have a place somewhere with the current version of the system..

Deniallugo (approved):

core/lib/zksync_core/src/api_server/tx_sender/mod.rs
perekopskiy: Well, this place is `SELECT protocol_version FROM miniblocks ORDER BY number DESC LIMIT 1`. But apparently, it slows down API.. It's not a big deal in this case, so let's just leave latest

Deniallugo: we can cache it

Deniallugo (approved):

core/lib/zksync_core/src/api_server/tx_sender/mod.rs
perekopskiy: Yes, Zach and me even started doing that but it requires too many changes for such small thing

PR #112, opened 2023-09-29
feat!: boojum integration

StanislavBreadless:
# What

Needed server changes for the Boojum integration. 

Note, that the server integration is not yet final and changes are anticipated

## Checklist

- [ ] PR title corresponds to the body of PR (we generate changelog entries fom PRs).
- [ ] Tests for the changes have been added / updated.
- [ ] Documentation comments have been added / updated.
- [ ] Code has been formatted via `zk fmt` and `zk lint`.

PR #415, opened 2023-11-06
chore: Fixes to the boojum integration PR

StanislavBreadless:
# What ❔

Fixes comments by @Deniallugo and @mm-zk 

## Checklist


core/lib/multivm/src/versions/vm_latest/bootloader_state/utils.rs
```diff
@@ -116,10 +116,19 @@ pub(crate) fn apply_pubdata_to_memory(
...
+    // - One slot is for the selector of the call to the L1Messenger.
+    // - The other slot is for the 0x20 offset for the calldata.
     let l1_messenger_pubdata_start_slot = OPERATOR_PROVIDED_L1_MESSENGER_PUBDATA_OFFSET + 2;
 
-    pubdata_information
-        .build_pubdata()
+    let pubdata = pubdata_information.build_pubdata();
+
+    assert!(
+        pubdata.len() / 32 <= OPERATOR_PROVIDED_L1_MESSENGER_PUBDATA_SLOTS - 2,
```
mm-zk, line 127: nit: this assert is a little bit incorrect - if pubdata.len is not divisible by 32.

mm-zk (approved):

core/lib/multivm/src/versions/vm_latest/bootloader_state/utils.rs
StanislavBreadless: it is enforced by `build_pubdata` that it is divisible by 32

core/lib/multivm/src/interface/types/outputs/execution_state.rs
```diff
@@ -12,7 +12,9 @@ pub struct CurrentExecutionState {
     pub used_contract_hashes: Vec<U256>,
     /// L2 to L1 logs produced by the VM.
     pub system_logs: Vec<L2ToL1Log>,
-    /// L2 to L1 logs produced by the L1Messeger
+    /// L2 to L1 logs produced by the L1Messeger.
+    /// For pre-boojum VMs, there was no distinction between user logs and system
+    /// logs and so all the outputted logs were treated as user_l2_to_l1_logs.
     pub user_l2_to_l1_logs: Vec<L2ToL1Log>,
```
Deniallugo, line 18: Why it's still just l2 to l1 logs and not UserL2ToL1Logs, seems you've introduced it earlier

core/lib/multivm/src/versions/vm_latest/implementation/execution.rs
```diff
@@ -49,6 +47,7 @@ impl<S: WriteStorage, H: HistoryMode> Vm<S, H> {
             tracers,
             self.storage.clone(),
             refund_tracers,
+            Some(PubdataTracer::new(self.batch_env.clone(), execution_mode)),
```
Deniallugo, line 50: why do we need some if it's always enabled?

core/lib/types/src/eth_sender.rs
```diff
@@ -12,20 +12,6 @@ pub struct EthTx {
     pub predicted_gas_cost: u64,
 }
 
-impl std::fmt::Debug for EthTx {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        // Do not print raw_tx
-        f.debug_struct("EthTx")
-            .field("id", &self.id)
-            .field("nonce", &self.nonce)
-            .field("contract_address", &self.contract_address)
```
Deniallugo, line 21: why so? do we really need raw_tx bytes? it could be 128kb long

core/lib/multivm/src/interface/types/outputs/execution_state.rs
StanislavBreadless: missed it here, thanks

core/lib/multivm/src/versions/vm_latest/implementation/execution.rs
Deniallugo: what is the custom implementation?

core/lib/types/src/eth_sender.rs
StanislavBreadless: @Deniallugo It is just rolling back to how it is done in the `main` branch. https://github.com/matter-labs/zksync-era/pull/112#discussion_r1377461695

Deniallugo: Ok!

ly0va: I'd rather not have raw tx bytes in the output, main reason why I implemented the custom Debug function for this. They are not helpful for debugging at all, and simply take space in the terminal and obstruct actually useful logs.

core/lib/multivm/src/versions/vm_latest/implementation/execution.rs
StanislavBreadless: For instance, the bootloader tester written by @mm-zk: https://github.com/matter-labs/era-system-contracts/tree/main/bootloader/test_infra

StanislavBreadless: I don't think that @mm-zk 's tooling would be broken if the pubdata tracer is always put there, though still I thought that our general approach is to preserve choices for the users of the VM

Deniallugo: ok... just less checks is a bit better performance wise :)

core/lib/types/src/eth_sender.rs
StanislavBreadless: ok

Deniallugo (approved):

core/lib/multivm/src/versions/vm_latest/implementation/execution.rs
mm-zk: I don't have strong opinions here (as long as this tracer doesn't prevent test bootloader from running)

PR #416, opened 2023-11-06
fix: some fixes for boojum

perekopskiy:
# What ❔

Fix comments opened by me in #112 

## Checklist

### DIFF
--- a/core/lib/dal/src/blocks_dal.rs
+++ b/core/lib/dal/src/blocks_dal.rs
@@ -1001,6 +1001,7 @@ impl BlocksDal<'_, '_> {
         default_aa_hash: H256,
         protocol_version_id: ProtocolVersionId,
     ) -> anyhow::Result<Vec<L1BatchWithMetadata>> {
+        let is_pre_boojum = protocol_version_id.is_pre_boojum();
         let raw_batches = sqlx::query_as!(
             StorageL1Batch,
             "SELECT number, l1_batches.timestamp, is_finished, l1_tx_count, l2_tx_count, fee_account_address, \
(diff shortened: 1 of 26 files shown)


core/lib/dal/src/blocks_dal.rs
```diff
@@ -1021,12 +1022,14 @@ impl BlocksDal<'_, '_> {
                 AND protocol_versions.bootloader_code_hash = $1 AND protocol_versions.default_account_code_hash = $2 \
                 AND commitment IS NOT NULL \
                 AND (protocol_versions.id = $3 OR protocol_versions.upgrade_tx_hash IS NULL) \
-                AND events_queue_commitment IS NOT NULL \
-                AND bootloader_initial_content_commitment IS NOT NULL
-            ORDER BY number LIMIT $4",
+                AND (events_queue_commitment IS NOT NULL AND bootloader_initial_content_commitment IS NOT NULL
+                    OR $4 = FALSE
+                )
+            ORDER BY number LIMIT $5",
```
Deniallugo, line 1028: TBH, i think it's better separate it to two methods...

perekopskiy: Ok, I agree that DAL method whose behavior depends on protocol version is too much

StanislavBreadless (approved):

core/lib/dal/src/blocks_dal.rs
```diff
@@ -994,6 +994,52 @@ impl BlocksDal<'_, '_> {
         })
     }
 
+    pub async fn legacy_get_ready_for_commit_l1_batches(
```
Deniallugo, line 997: is it a good name? i think pre boojum would be better

ly0va (approved):

PR #440, opened 2023-11-07
feat(hyperchain): Adding prover related commands to zk stack

githubdoramon:
## What ❔

Add commands to zk stack to setup a prover (for now only CPU, it blocks the GPU option), and adds the necessary components to the compose file

## Why ❔

"Without a prover, a hyperchain is nothing more than a even worst OP rollup" - Canales, Ramon

## Checklist


etc/hyperchains/docker-compose-hyperchain-template
```diff
@@ -44,101 +47,88 @@ services:
...
-      - server_artifacts:/etc_server_artifacts
+      - artifacts:{{artifactsPath}}
+      - {{proverSetupArtifacts}}:{{proverSetupArtifacts}}
   zkstack-witness-generator-basic-circuits:
     networks:
       - zkstack
-    build:
-      context: .
-      dockerfile: ./docker/witness-generator/Dockerfile
+    image: matterlabs/witness-generator:latest
```
hatemosphere, line 66: where are these "latest" tags coming from? in this repo we always had "2.0-latest" instead

hatemosphere: also why are we building some and some are getting pulled from the registry? we need to stay consistent, otherwise we can end up with running multiple versions of code. the same can happen if using only latest from registry, so ideally we want to either pull all witn specific pinned tag (latest semver release deployed and tested to mainnet) or build all from code and dockerfiles

hatemosphere (changes requested):

etc/hyperchains/docker-compose-hyperchain-template
```diff
@@ -3,24 +3,27 @@ networks:
   zkstack:
     driver: bridge
 volumes:
```
akash-chandrakar, line 5: is it possible to validate this docker-compose template in `CI` to ensure vars configured inside works ?

githubdoramon: This is the idea indeed. @hatemosphere will add some CI pipeline to spin it up with  "zk stack demo skip-env-setup"

etc/hyperchains/docker-compose-hyperchain-template
githubdoramon: Server is built (because is the "easiest" one to change without major protocol changes (and is part of the genesis setup anyways). All prover-related ones are just being pulled for now.

hatemosphere: decided to fetch and use latest semver in separate/next PR

etc/hyperchains/docker-compose-hyperchain-template
hatemosphere: we have a plan to run basic tests indeed, however let's not do it in this PR

hatemosphere (approved):

PR #444, opened 2023-11-08
feat(dal): Do not load config from env in DAL crate

popzxc:
## What ❔

- DAL crate no longer loads `DATABASE_URL` and siblings, as well as `DATABASE_POOL_SIZE` implicitly. Now they have to always be passed externally.
- `DbVariant` is removed: DAL no longer assumes that there are different databases.
- Some unused code was removed along the path.

## Why ❔

DAL was glued to the env configs. Now it isn't.

## Checklist


core/bin/external_node/src/config/mod.rs
```diff
@@ -375,11 +375,35 @@ impl RequiredENConfig {
...
 
+/// Configuration for Postgres database.
+/// While also mandatory, it historically used different naming scheme for corresponding
+/// environment variables.
+/// Thus it is kept separately for backward compatibility and ease of deserialization.
+#[derive(Debug, Clone, Deserialize, PartialEq)]
+pub struct PostgresConfig {
+    pub database_url: String,
+    pub max_connections: u32,
+}
```
popzxc, line 386: In theory, we could turn `DATABASE_xxx` vars into `EN_DATABASE_xxx` so that the whole naming scheme is consistent, however
a) surely it's out of scope for this PR
b) with many partners already having their nodes set up and running, I don't really know if it's worth the hustle.

Right now, I think it's OK to ignore this topic completely until someone complains.

core/bin/zksync_server/src/main.rs
```diff
@@ -150,6 +125,40 @@ async fn main() -> anyhow::Result<()> {
         object_store_config: ObjectStoreConfig::from_env().ok(),
     };
 
+    let postgres_config = configs.postgres_config.clone().context("PostgresConfig")?;
+
+    if opt.genesis || is_genesis_needed(&postgres_config).await {
+        let network = NetworkConfig::from_env().context("NetworkConfig")?;
```
popzxc, line 131: Again: both here and in `zksync_core/src/lib.rs` a lot of code was ugly and is ugly.
I have opted out of refactoring it, since it's going to be removed in subsequent PRs.

core/lib/config/src/configs/database.rs
```diff
@@ -111,9 +111,6 @@ impl MerkleTreeConfig {
 /// Database configuration.
 #[derive(Debug, Clone, PartialEq, Deserialize)]
 pub struct DBConfig {
-    /// Statement timeout in seconds for Postgres connections. Applies only to the replica
-    /// connection pool used by the API servers.
-    pub statement_timeout_sec: Option<u64>,
```
popzxc, line 116: Funnily, it was the only env variable related to the Postgres in the DBConfig.

core/lib/dal/src/connection/mod.rs
```diff
@@ -4,42 +4,31 @@ use sqlx::{
...
+use crate::{metrics::CONNECTION_METRICS, StorageProcessor};
 
-#[derive(Debug, Clone, Copy)]
-pub enum DbVariant {
-    Master,
-    Replica,
-    Prover,
-    TestTmp,
+/// Obtains the test database URL from the environment variable.
+fn get_test_database_url() -> anyhow::Result<String> {
```
popzxc, line 16: This is the only place where we use env variables in the DAL crate now.
I thought that since we require a special setup for tests anyway, it's OK to say that the code running tests is responsible for making sure that this env variable is set. I think it's a fair price for nicer interface of creating test pools.

core/lib/zksync_core/src/data_fetchers/token_list/mock.rs
```diff
@@ -69,7 +68,10 @@ struct TokenGenesisListItem {
 }
 
 fn get_genesis_token_list(network: &str) -> Vec<TokenGenesisListItem> {
-    let mut file_path = parse_env::<PathBuf>("ZKSYNC_HOME");
+    let mut file_path: PathBuf = std::env::var("ZKSYNC_HOME")
```
popzxc, line 71: `std::env::var("ZKSYNC_HOME")` will also be gone, but later.

core/bin/external_node/src/config/mod.rs
```diff
@@ -375,11 +375,35 @@ impl RequiredENConfig {
...
+
+impl PostgresConfig {
+    pub fn from_env() -> anyhow::Result<Self> {
+        Ok(Self {
+            database_url: env::var("DATABASE_URL")
```
Deniallugo, line 391: But here you use convention from the main database

popzxc: image

core/lib/config/src/configs/database.rs
```diff
@@ -143,12 +140,59 @@ impl DBConfig {
...
+    /// connection pool used by the API servers.
+    pub statement_timeout_sec: Option<u64>,
+}
+
+impl PostgresConfig {
+    /// Returns a copy of the master database URL as a `Result` to simplify error propagation.
+    pub fn master_url(&self) -> anyhow::Result<String> {
+        self.master_url
+            .clone()
+            .ok_or_else(|| anyhow::anyhow!("Master DB URL is absent"))
```
slowli, line 171: Nit: You can use the `anyhow::Context` trait to convert `Option`s with a similar result.

core/lib/dal/src/connection/mod.rs
```diff
@@ -51,41 +40,28 @@ impl ConnectionPoolBuilder {
...
-            .unwrap_or_else(|| parse_env("DATABASE_POOL_SIZE"));
-
-        let options = PgPoolOptions::new().max_connections(max_connections);
-        let mut connect_options: PgConnectOptions = database_url
+        let options = PgPoolOptions::new().max_connections(self.max_size);
+        let mut connect_options: PgConnectOptions = self
+            .database_url
             .parse()
-            .with_context(|| format!("Failed parsing {:?} database URL", self.db))?;
+            .with_context(|| format!("Failed parsing database URL: {}", self.database_url))?;
```
slowli, line 47: `database_url` can contain sensitive information (user / password for the database); that's why it wasn't included into the error messages previously. For the same reason, it may make sense to exclude it from `Debug` implementation of `ConnectionPoolBuilder`.

core/bin/external_node/src/main.rs
```diff
@@ -348,10 +349,13 @@ async fn main() -> anyhow::Result<()> {
         .main_node_url()
         .context("Main node URL is incorrect")?;
 
-    let connection_pool = ConnectionPool::builder(DbVariant::Master)
-        .build()
-        .await
-        .context("failed to build a connection_pool")?;
+    let connection_pool = ConnectionPool::builder(
+        config.postgres.database_url.clone(),
+        config.postgres.max_connections,
```
pompon0, line 354: how about taking the whole config as an argument?

core/bin/external_node/src/config/mod.rs
Deniallugo: Yeah, but our partners use `EN_DATABASE`

core/bin/external_node/src/main.rs
popzxc: This function is invoked using different configs for server/EN, and for the server, we have multiple different URLs.
I didn't find a way to introduce a reusable struct here that would be better than just passing two variables.

pompon0: nvm, there are different PostgresConfig types

core/lib/config/src/configs/database.rs
```diff
@@ -143,12 +140,59 @@ impl DBConfig {
...
+    }
+}
+
+/// Collection of different database URLs and general PostgreSQL options.
+/// All the entries are optional, since some components may only require a subset of them,
+/// and any component may have overrides.
+#[derive(Debug, Clone)]
+pub struct PostgresConfig {
+    /// URL for the main (sequencer) database.
+    pub master_url: Option<String>,
```
pompon0, line 154: how about url::Url?

core/bin/external_node/src/config/mod.rs
popzxc: Could you please show where this variable is actually used?
Before this PR DAL itself has been loading the variable without any prefixes, and our [config samples](https://github.com/matter-labs/zksync-era/blob/main/docs/external-node/prepared_configs/mainnet-config.env) in docs also do not have prefixes for the database.

This was done this way _exactly_ to be backward compatible.

core/lib/zksync_core/src/lib.rs
```diff
@@ -187,8 +192,16 @@ pub async fn genesis_init(
     Ok(())
 }
 
-pub async fn is_genesis_needed() -> bool {
-    let mut storage = StorageProcessor::establish_connection(true).await.unwrap();
+pub async fn is_genesis_needed(postgres_config: &PostgresConfig) -> bool {
+    let db_url = postgres_config
+        .master_url
+        .clone()
+        .expect("Master DB URL is missing");
```
pompon0, line 199: .master_url().unwrap()

pompon0 (dismissed):

core/lib/config/src/configs/database.rs
popzxc: I think it will result in [double parsing](https://github.com/matter-labs/zksync-era/blob/main/core/lib/dal/src/connection/mod.rs#L74) without clear benefits. On the other hand, we'll have to add a new dependency here for that.

prover/witness_generator/src/main.rs
```diff
@@ -80,14 +80,21 @@ async fn main() -> anyhow::Result<()> {
...
+    .build()
+    .await
+    .context("failed to build a connection_pool")?;
+    let prover_connection_pool = ConnectionPool::builder(
+        postgres_config.prover_url()?,
+        postgres_config.max_connections()?,
+    )
+    .build()
+    .await
+    .context("failed to build a prover_connection_pool")?;
```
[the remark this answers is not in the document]
popzxc: Yup. Well, at least `rustfmt` seems to be happy with it.

core/lib/dal/src/connection/mod.rs
```diff
@@ -4,42 +4,31 @@ use sqlx::{
...
+/// Obtains the test database URL from the environment variable.
+fn get_test_database_url() -> anyhow::Result<String> {
+    env::var("TEST_DATABASE_URL").context(
+        "TEST_DATABASE_URL must be set. Normally, this is done by the 'zk' tool. \
+        Make sure that you are running the tests with 'zk test rust' command or equivalent.",
+    )
 }
 
 /// Builder for [`ConnectionPool`]s.
 #[derive(Debug)]
```
slowli, line 24: Gentle reminder to exclude `database_url` from the `Debug` implementation.

core/lib/zksync_core/src/lib.rs
```diff
@@ -1097,7 +1126,11 @@ async fn add_house_keeper_to_task_futures(
         .house_keeper_config
         .clone()
         .ok_or_else(|| anyhow::anyhow!("house_keeper_config"))?;
-    let connection_pool = ConnectionPool::singleton(DbVariant::Replica)
+    let postgres_config = configs
+        .postgres_config
+        .clone()
+        .ok_or_else(|| anyhow::anyhow!("postgres_config"))?;
```
slowli, line 1132: Nit: Can be simplified to `.context("postgres_config")`.

AnastasiiaVashchuk (dismissed):

perekopskiy (dismissed):

slowli (dismissed):

Deniallugo (approved):

PR #448, opened 2023-11-08
feat(job-processor): report attempts metrics

perekopskiy:
## What ❔

Every component implementing JobProcessor now exposes two metrics:

`job_attempts` -- heatmaps in grafana will be added displaying this metric.
`max_attempts_reached` -- alert will fire when this metric is increased (alert rules not added yet).

Also, `attempts` columns in DB were increased in both `get_next*` methods and `requeue_stuck_jobs`, so it was basically `attempts += 2` on retry. I removed `attempts` update in `requeue_stuck_jobs`.

## Why ❔

Better monitoring and alerting

## Checklist


core/lib/queued_job_processor/src/lib.rs
```diff
@@ -9,10 +9,26 @@ use tokio::time::sleep;
 
 use zksync_utils::panic_extractor::try_extract_panic_message;
 
+use vise::{Buckets, Counter, Histogram, LabeledFamily, Metrics};
+
+const BYTE_SIZE_BUCKETS: Buckets = Buckets::exponential(1.0..=64.0, 2.0);
```
slowli, line 14: Nit: `ATTEMPT_BUCKETS`? The current name looks like a copy-paste artifact.

core/lib/queued_job_processor/src/lib.rs
```diff
@@ -127,6 +147,10 @@ pub trait JobProcessor: Sync + Send {
             job_id,
             error_message
         );
+
+        if attempts == Some(self.max_attempts()) {
+            METRICS.max_attempts_reached[&(Self::SERVICE_NAME, format!("{job_id:?}"))].inc();
```
slowli, line 152: Stupid question: Can't having `job_id` as a label lead to a huge number of time series for different metrics? Or do we assume that running out of attempts is rare enough? I *think* having job ID logged with a `WARN` / `ERROR` level and/or represented as [a metric exemplar](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#exemplars) would be more appropriate. (I'm not that familiar with exemplars though, and `vise` currently doesn't support them, although they [are supported](https://docs.rs/prometheus-client/latest/prometheus_client/metrics/exemplar/index.html) by the underlying library.)

core/lib/queued_job_processor/src/lib.rs
```diff
@@ -138,4 +162,8 @@ pub trait JobProcessor: Sync + Send {
         started_at: Instant,
         artifacts: Self::JobArtifacts,
     ) -> anyhow::Result<()>;
+
+    fn max_attempts(&self) -> u32;
+
+    async fn get_job_attempts(&self, job_id: &Self::JobId) -> anyhow::Result<Option<u32>>;
```
slowli, line 168: What does an error returned from this method mean? ("Number of attempts is not available at the time"?) Is it significantly different from `Ok(None)`? E.g., `wait_for_task` returns early if `get_job_attempts` fails – should it do that, given that the number of attempts is not mandatory anyway?

prover/proof_fri_compressor/src/compressor.rs
```diff
@@ -202,4 +205,16 @@ impl JobProcessor for ProofCompressor {
             .await;
         Ok(())
     }
+
+    fn max_attempts(&self) -> u32 {
+        self.max_attempts
+    }
+
+    async fn get_job_attempts(&self, job_id: &L1BatchNumber) -> anyhow::Result<Option<u32>> {
+        let mut prover_storage = self.pool.access_storage().await.unwrap();
```
slowli, line 214: Is `unwrap()` intentional here and in other implementations given that an error could be propagated?

core/lib/dal/src/fri_proof_compressor_dal.rs
```diff
@@ -89,8 +89,22 @@ impl FriProofCompressorDal<'_, '_> {
...
         .unwrap()
-        .map(|row| L1BatchNumber(row.l1_batch_number as u32));
-        result
+        .map(|row| L1BatchNumber(row.l1_batch_number as u32))
+    }
+
+    pub async fn get_proof_compression_job_attempts(
+        &mut self,
+        l1_batch_number: L1BatchNumber,
+    ) -> Option<u32> {
```
slowli, line 98: Maybe, it makes sense to propagate `sqlx` errors right away for new methods (as in `get_basic_witness_input_producer_job_attempts`)?

core/lib/queued_job_processor/src/lib.rs
perekopskiy: This should be rare enough. For production envs I think this is completely ok, but for staging it could be a problem in theory. 
The reason why I want to keep it is that having job id as a label allows us to display it in the alert message which is pretty useful. I'd go with label and change it later if it ever becomes an issue

core/lib/queued_job_processor/src/lib.rs
perekopskiy: This is for DB errors. `Ok(None)` is for cases when attempts are not applicable/should be ignored -- currently it's used in old prover stack + contract verifier. For old prover components I didn't want to add DAL methods given that we are going to remove these components in a few weeks. For contract verifier it's always 1 attempt.

Now I'm thinking that probably it makes sense to remove `Option` from the interface and just return `1` for old prover components, contract verifier.

prover/proof_fri_compressor/src/compressor.rs
perekopskiy: No, makes sense to propagate it

core/lib/dal/src/fri_proof_compressor_dal.rs
perekopskiy: Agree

akash-chandrakar (approved):

slowli (approved):

PR #455, opened 2023-11-09
feat(core): adds a get proof endpoint in zks namespace

montekki:
## What ❔

This PR adds a new JSON-RPC endpoint `zks_getProof` that allows users to request proofs
of data in the tree that can later be verified against storarage roots at a given point
in batch history.


core/lib/zksync_core/src/api_server/web3/namespaces/zks.rs
```diff
@@ -620,4 +622,30 @@ impl<G: L1GasPriceProvider> ZksNamespace<G> {
...
+            .collect();
+
+        let keys = self
+            .state
+            .tree_api
+            .as_ref()
+            .ok_or(Web3Error::GetProofsUnavailable)?
+            .get_proofs(l1_batch_number, hashed_keys)
+            .await
+            .map_err(|err| internal_error(METHOD_NAME, err))?;
```
slowli, line 647: A potential issue here is that if the tree API is load-balanced via K8s services (makes sense), a query to a certain upstream server may fail because it doesn't have an L1 batch locally while other servers may have it. Not sure how best to deal with it. A tree API server provides its latest processed L1 batch [via a separate endpoint](https://github.com/matter-labs/zksync-era/blob/664ce33622af220a24360f7f11a52a14141c3fdc/core/lib/zksync_core/src/api_server/tree/mod.rs#L92), so we could theoretically use app-level load-balancing aware of the current server states (encapsulated in a `TreeApiClient` implementation). Obviously, this could be dealt with later.

core/lib/config/src/configs/api.rs
```diff
@@ -87,6 +87,8 @@ pub struct Web3JsonRpcConfig {
     /// The value is per active connection.
     /// Note: For HTTP, rate limiting is expected to be configured on the infra level.
     pub websocket_requests_per_minute_limit: Option<u32>,
+    /// Api of the tree API, currently used to proxy `getProof` calls to the tree
```
RomanBrodetski, line 90:
```suggestion
     /// Tree API url, currently used to proxy `getProof` calls to the tree
```

core/lib/web3_decl/src/error.rs
```diff
@@ -37,4 +37,6 @@ pub enum Web3Error {
     InvalidFilterBlockHash,
     #[error("Query returned more than {0} results. Try smaller range of blocks")]
     TooManyLogs(usize),
+    #[error("getProofs API is not available")]
+    GetProofsUnavailable,
```
RomanBrodetski, line 41: Perhaps `TreeApiUnavailable`?

core/lib/zksync_core/src/api_server/web3/namespaces/zks.rs
```diff
@@ -620,4 +622,30 @@ impl<G: L1GasPriceProvider> ZksNamespace<G> {
     ) -> Result<Vec<Log>, Web3Error> {
         self.state.translate_get_logs(filter).await
     }
+
+    #[tracing::instrument(skip_all)]
+    pub async fn get_proofs_impl(
+        &self,
+        address: Address,
+        keys: Vec<H256>,
+        l1_batch_number: L1BatchNumber,
```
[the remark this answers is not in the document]
RomanBrodetski: agreed - I think we use BlockNumber/BlockId everywhere?

core/lib/zksync_core/src/api_server/web3/namespaces/zks.rs
RomanBrodetski: yes, can be dealt with later - but indeed need to think about it, especially if we support BlockNumber/BlockId. Would be silly to get latest block number in the DB and then hit a server with no such l1 batch and return error :)

RomanBrodetski (commented):
just some nits

core/lib/zksync_core/src/api_server/web3/namespaces/zks.rs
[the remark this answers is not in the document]
montekki: Why, in `zks` none of the methods use those, it is always either `L1BatchNumber` or `MiniblockNumber`:

https://github.com/matter-labs/zksync-era/blob/4e3cf93305067b5d5fd2b43b1102ef70e70b269d/core/lib/web3_decl/src/namespaces/zks.rs#L78

https://github.com/matter-labs/zksync-era/blob/4e3cf93305067b5d5fd2b43b1102ef70e70b269d/core/lib/web3_decl/src/namespaces/zks.rs#L96-L97

https://github.com/matter-labs/zksync-era/blob/4e3cf93305067b5d5fd2b43b1102ef70e70b269d/core/lib/web3_decl/src/namespaces/zks.rs#L59-L61

slowli (approved):

PR #458, opened 2023-11-09
feat: protobuf-generated json configs for the main node (BFT-371)

pompon0:
example json config:
```
{
  "httpPort": 3050,
  "httpUrl": "http://localhost:3050",
  "wsPort": 3051,
  "wsUrl": "ws://localhost:3051",
  "reqEntitiesLimit": 10000,
  "filtersLimit": 10000,
  "subscriptionsLimit": 10000,
  "pubsubPollingInterval": "200",
  "threadsPerServer": 1,
  "maxNonceAhead": 50,
  "gasPriceScaleFactor": 1.2,
  "accountPks": {
    "keys": [
      "AQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQE=",
      "AgICAgICAgICAgICAgICAgICAgICAgICAgICAgICAgI="
    ]
  },
  "estimateGasScaleFactor": 1.2,
  "estimateGasAcceptableOverestimation": 1000,
  "maxTxSize": "1000000"
}
```


core/lib/config/Cargo.toml
```diff
@@ -12,7 +12,13 @@ categories = ["cryptography"]
 [dependencies]
 zksync_basic_types = { path = "../../lib/basic_types" }
 zksync_contracts = { path = "../../lib/contracts" }
+zksync_protobuf.workspace = true
```
popzxc, line 15: I'm afraid this kills the idea of detaching config definitions from the serialization format. Chances are, we'll move from depending on env variables to be always present to depending on protobuf configs to be always present, which is not really better.
I'd assume that we need a proxy crate like `zksync_protobuf_config` that will only be used in binary crates, so that library code has no access to some kind of "default" config provider.

Cargo.toml
```diff
@@ -57,3 +57,6 @@ exclude = []
 [profile.perf]
 inherits = "release"
 debug = true
+
+[workspace.dependencies]
+zksync_protobuf = { version = "0.1.0", git = "https://github.com/matter-labs/era-consensus.git", rev = "9b300371512a4e443d086ecd19c1877719c15646" }
```
popzxc, line 62: I'm afraid this will break the prover workspace, we'll need to fix that first :( 
(prover workspace was the reason why I had to rollback all my attempts to switch to workspace deps in the past

pompon0: ack, removed workspace.dependencies

core/lib/config/Cargo.toml
pompon0: I've created a separate crate and defined a trait ProtoRepr (reverse to ProtoFmt) since rust does not allow to implement an external trait for external type. The trait and auxiliary encode/decode methods are currently defined locally but will be moved to zksync_protobuf crate later.

core/lib/protobuf_config/src/proto/api.proto
```diff
@@ -0,0 +1,62 @@
+syntax = "proto3";
```
Deniallugo, line 1: wdyt about using proto2 and having required fields?

yorik: Please do not use required fields. It's always a disaster in long term

pompon0: We have restricted out code generator so that proto binary encoding is canonical (so that we can sign proto binary messages). For simplicity, only proto3 syntax is supported and all fields are required to be either optional or repeated. Canonical encoding is obviously not required for configs, however it would take some extra work to make code generator distinguish these use cases.

pompon0: and yes, having required fields in protos is considered an antipattern, because required fields can never be deprecated.

core/lib/protobuf_config/src/proto/utils.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto3";
+
+package zksync.config;
+
+message Prometheus {
+  optional uint32 listener_port = 1; // required
+  optional string pushgateway_url = 2; // required
+  optional uint64 push_interval_ms = 3;
```
yorik, line 8: nit: I'd use seconds here, we definitely don't want to push more often than 30s.

core/lib/protobuf_config/src/proto/fri_prover_group.proto
```diff
@@ -0,0 +1,24 @@
...
+
+package zksync.config;
+
+message CircuitIdRoundTuple {
+  optional uint32 circuit_id = 1; // required; u8
+  optional uint32 aggregation_round = 2; // required; u8
+}
+
+message FriProverGroup {
+  repeated CircuitIdRoundTuple group_0 = 1;
```
yorik, line 11: Maybe repeated FriProverGroup or map over CircuitIdRoundTuple would be better.
AFAIK number of groups could change.

core/lib/protobuf_config/src/proto/eth_watch.proto
```diff
@@ -0,0 +1,8 @@
+syntax = "proto3";
+
+package zksync.config;
+
+message ETHWatch {
+  optional uint64 confirmations_for_eth_event = 1; // optional
+  optional uint64 eth_node_poll_interval = 2; // required; ms
```
yorik, line 7: `_ms` suffix is missing

core/lib/protobuf_config/src/proto/chain.proto
yorik, line 1: There are different suffix styles for durations across all files:
* `_seconds`
* `_in_secs`
* `_sec`
* `_s`
* no suffix

I think it would be better to have single style, I personally prefer `_s`/`_ms`.

core/lib/protobuf_config/src/proto/utils.proto
pompon0: https://github.com/matter-labs/zksync-era/blob/ec6e923ddb5063d391c59e04dad28f9d2c15dcd1/core/lib/config/src/configs/utils.rs#L12 specifies this in ms. I converted it 1-1 so that migration from env vars is straightforward. 

We can improve the config format later. For all duration params I'd suggest to use a Duration message:
https://github.com/matter-labs/era-consensus/blob/dbdc5312fa60d5f1716a6ac791218353c0cc451e/node/libs/protobuf/src/proto/std.proto#L23
or (if that's too verbose) a string parseable to time::Duration (like "10s", or "500ms"), so that the unit is not specified in the field name.

core/lib/protobuf_config/src/proto/eth_watch.proto
pompon0: https://github.com/matter-labs/zksync-era/blob/ec6e923ddb5063d391c59e04dad28f9d2c15dcd1/core/lib/config/src/configs/eth_watch.rs#L13

core/lib/protobuf_config/src/proto/utils.proto
yorik: SGTM

yorik (dismissed):
Approve for proto files

perekopskiy (approved):

perekopskiy (approved):

perekopskiy (approved):

Issue #451, opened 2023-11-09
`debug_trace*` requires `tracerConfig` to be set 


alexqrid:
### 🐛 Bug Report
Hi team!
Seems that there is a bug either in docs or in implementation.
In [docs](https://era.zksync.io/docs/api/api.html#debug-traceblockbynumber) `tracerConfig` described as optional argument, but when I actually query the public rpc or my NaaS provider nodes I receive the error `missing field `tracerConfig` at line 1 column 24`

```sh
$ curl --location 'https://mainnet.era.zksync.io' \
--header 'Content-Type: application/json' \
--data '{"jsonrpc": "2.0",
	"id":1,
	"method":"debug_traceBlockByNumber",
	"params":["0x11a8bf6",{"tracer": "callTracer"}]
}'
# Response
{
    "jsonrpc": "2.0",
    "error": {
        "code": -32602,
        "message": "Invalid params",
        "data": "missing field `tracerConfig` at line 1 column 24"
    },
    "id": 1
}
```

However setting the `tracerConfig` fixes the problem, but it's wrong documented behaviour.

```sh
$ curl --location 'https://mainnet.era.zksync.io' \
--header 'Content-Type: application/json' \
--data '{"jsonrpc": "2.0",
	"id":1,
	"method":"debug_traceBlockByNumber",
	"params":["0x11a8bf6",{"tracer": "callTracer", "tracerConfig": { "onlyTopCall": false }}]
}'

{
    "jsonrpc": "2.0",
    "result": [
        {
            "result": {
                "type": "Call",
                "from": "0x0000000000000000000000000000000000000000",
                "to": "0x0000000000000000000000000000000000008001",
                "gas": "0x8b14b",
                "gasUsed": "0x51b39",
                "value": "0x0",
                "output": "0x",
                "input": "0x095ea7b3000000000000000000000000be7d1fd1f6748bbdefc4fbacafbb11c6fc506d1d0000000000000000000000000000000000000000000000000000000001312d00",
                "error": null,
                "revertReason": null,
                "calls": ...
}
```

RedaOps:
Hi!

This happens because you still include the `{"tracer": "callTracer"}` param, but it also expects a `tracerConfig` in that JSON object in the second parameter. For example, this request will not fail:

```
curl --location 'https://mainnet.era.zksync.io' \
--header 'Content-Type: application/json' \
--data '{"jsonrpc": "2.0",
        "id":1,
[...]

PR #468, opened 2023-11-12
fix(api): `debug_trace*` no longer throws error if the `TracerConfig` object is incomplete

RedaOps:
## What ❔

* Derived `Default` for `CallTracerConfig`
* Allow default values for `tracer_config` field inside `TracerConfig` 
* Integration tests for this exact issue to confirm it is solved.
* Running `zk test i api` will now also run the test suite for the debug namespace. (this is optional and not required in any way for the bug fix, so I can remove it)

## Why ❔

Check out the discussion in #451 
The following `debug` namespace rpc methods would throw an error if you would include a `TracerConfig` as an argument but only specified the `tracer` and not also the `tracer_config` (see issue mentioned above for full description of the issue):

* `debug_traceBlockByNumber`
*  `debug_traceBlockByHash`
*  `debug_traceCall`
*  `debug_traceTransaction`

Some ETH nodes I tested do not throw an error in this case, so the zkSync node should behave the same way after this fix.

## Checklist


RedaOps:
If the issue described in #451 is not actually a bug and is expected behaviour, you can close this PR.

Thanks!

alexqrid:
@RedaOps Thanks for the explanation! Yes, the key problem is that the behaviour of the api is not consistent with `Ethereum JSON-RPC`, so your PR makes sense

alexqrid (approved):

alexqrid:
cc @slowli

core/lib/types/src/api/mod.rs
```diff
@@ -637,7 +637,7 @@ pub struct CallTracerConfig {
 #[serde(rename_all = "camelCase")]
 pub struct TracerConfig {
     pub tracer: SupportedTracers,
-    pub tracer_config: CallTracerConfig,
+    pub tracer_config: Option<CallTracerConfig>,
```
slowli, line 640: I'd suggest to annotate this field with `#[serde(default)]` and derive `Default` with `CallTracerConfig`. If I understand correctly, this will achieve the same result without needing to change `debug` namespace implementation.

RedaOps: Thanks! Will look into it.

slowli (approved):

PR #472, opened 2023-11-13
refactor(merkle tree): Remove lifetime in `MerkleTree`

slowli:
## What ❔

Removes the lifetime parameter in `MerkleTree`, `MiniMerkleTree` and related types.

## Why ❔

The lifetime is there because of `HashTree` / `HashEmptySubtree` traits, which only have stateless `'static` implementations now and for the foreseeable future. Thus, these lifetimes can be replaced with `'static`.

## Checklist

### DIFF
--- a/core/lib/merkle_tree/src/lib.rs
+++ b/core/lib/merkle_tree/src/lib.rs
@@ -78,7 +78,7 @@ pub use crate::{
     },
 };
 
-use crate::{storage::Storage, types::Root};
+use crate::{hasher::HasherWithStats, storage::Storage, types::Root};
 use zksync_crypto::hasher::blake2::Blake2Hasher;
 
 /// Binary Merkle tree implemented using AR16MT from Diem [Jellyfish Merkle tree] white paper.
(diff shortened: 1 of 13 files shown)


core/lib/merkle_tree/src/lib.rs
```diff
@@ -123,12 +123,12 @@ use zksync_crypto::hasher::blake2::Blake2Hasher;
 ///
 /// [Jellyfish Merkle tree]: https://developers.diem.com/papers/jellyfish-merkle-tree/2021-01-14.pdf
 #[derive(Debug)]
-pub struct MerkleTree<'a, DB> {
+pub struct MerkleTree<DB> {
     db: DB,
-    hasher: &'a dyn HashTree,
+    hasher: &'static dyn HashTree,
```
popzxc, line 128: At this point is there any advantage of keeping a reference to `dyn HashTree` vs having `MerkleTree` generic over `HashTree` and storing it directly?

popzxc (dismissed):

core/lib/merkle_tree/src/lib.rs
slowli: Dynamic dispatch slightly reduces compilation times and produced executable size at the cost of theoretical performance degradation (I haven't observed any noticeable performance degradation when performing benchmarks locally.) Since we have a single hash function when using Merkle tree, a parametric implementation should produce that much overhead in practice, so I think switching to it makes sense.

popzxc (approved):

PR #482, opened 2023-11-13
fix(core)!: Actually generate an Ethereum-compatible signature.

thomasknauth:
Add the missing '\x19Ethereum ...' prefix and message length before hashing and signing.

## What ❔

Reconcile comment and implementation. The `sign_message` function now generates an Ethereum-compatible signature for the input `message`.

## Why ❔

At a minimum, the comment and implementation disagreed (comment said "this generates an Ethereum-compatible signature, which it didn't actually do; the implementation just hashed `message` w/o adding the prefix and message length). I opted to fix the implementation (instead of adjusting the comment), as this seemed the more apt way forward to me.

## Checklist

### DIFF
--- a/core/lib/eth_signer/src/pk_signer.rs
+++ b/core/lib/eth_signer/src/pk_signer.rs
@@ -36,7 +36,12 @@ impl EthereumSigner for PrivateKeySigner {
     /// The sign method calculates an Ethereum specific signature with:
     /// sign(keccak256("\x19Ethereum Signed Message:\n" + len(message) + message))).
     async fn sign_message(&self, message: &[u8]) -> Result<PackedEthSignature, SignerError> {
-        let signature = PackedEthSignature::sign(&self.private_key, message)
+        let prefix = format!("\x19Ethereum Signed Message:\n{}", message.len());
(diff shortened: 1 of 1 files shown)


RomanBrodetski:
hmm so how do we currently use this `sign_message`? From the first glance, this change should break somethign (or something should have not worked before)

thomasknauth:
I agree, it would be a breaking change. As far as I can tell, it has no users. I stumbled upon it while prototyping something. Some of the methods in `pk_signer`, e.g., `sign_transaction` are used in the SDK though. Not sure if any 3rd parties are using this API.

popzxc (commented):
This is a leftover from v1/lite. 
Looks like this functionality was implemented correctly there, but after some refactoring it no longer works in v2/era: adding the prefix was handled by the `PackedEthSignature` itself.

That said, I remember a lot of pain with this prefix, since in v1 we had apps/wallets where some expected message to be prefixed, and some didn't expect any prefix... I don't remember all the context though.

Given that this functionality isn't used anywhere, I believe that the right action item is to remove the `sign_message` method completely.

popzxc:
Fixed by https://github.com/matter-labs/zksync-era/pull/1294

PR #490, opened 2023-11-14
feat: added consensus types and consensus column to miniblocks table

pompon0:
To execute consensus on miniblocks we need to store extra data per miniblock.


core/lib/zksync_core/src/consensus/mod.rs
```diff
@@ -0,0 +1,124 @@
...
+
+mod payload;
+mod proto;
+
+pub(crate) use payload::Payload;
+
+struct PostgresState(zksync_dal::ConnectionPool);
+
+#[async_trait::async_trait]
+pub trait BlockChainState {
```
slowli, line 15: Is it possible to remove this trait from this PR? It seems unrelated to adding the `consensus` field for Postgres, doesn't look necessary, and it raises some questions. The main question is about the data flow: can we afford non-atomic update of consensus block fields for an already-persisted L2 block? Right now, my PR (#371) isn't accounting for this possibility (note that saving consensus fields is built into miniblock persistence logic and is atomic with saving the miniblock header itself), and in general it raises some concerns: What if a node crashes after persisting a block, but before adding consensus fields for it? What if another node requests a block in such a state? It's not that these issues are insurmountable, but IMO, atomicity makes things quite a bit easier.

core/lib/zksync_core/src/consensus/proto/mod.proto
```diff
@@ -0,0 +1,21 @@
...
+  // Default derive(serde::Serialize) encoding of the zksync_types::Transaction.
+  // TODO(gprusak): it is neither efficient, nor suitable for version control.
+  //     replace with a more robust encoding.
+  optional string json = 1; // required
+}
+
+message Payload {
+  optional bytes hash = 1; // required; H256
+  optional uint32 l1_batch_number = 2; // required
+  optional uint64 timestamp = 3; // required; ms since UNIX epoch?
```
slowli, line 15: As defined now, `SyncBlock.timestamp` is the number of *seconds* since epoch. This imprecise timestamp works because miniblock timestamps must be distinct as per EVM restrictions.

core/lib/zksync_core/src/consensus/mod.rs
pompon0: yeah, it's unrelated. Removing.

core/lib/zksync_core/src/consensus/payload.rs
```diff
@@ -0,0 +1,92 @@
+use anyhow::Context as _;
+use zksync_consensus_roles::validator;
+use zksync_protobuf::{required, ProtoFmt};
+use zksync_types::api::en::SyncBlock;
+use zksync_types::{Address, L1BatchNumber, Transaction, H256};
+
+pub(crate) struct Payload {
```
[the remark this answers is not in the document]
pompon0: `types` crate is imported by virtually everything. I don't think that's the desired level of visibility (here Payload can be even crate-private). ConsensusBlockFields is there only because it is embedded in Block. Otherwise it would be better suited in the `dal` crate.

core/lib/zksync_core/src/consensus/proto/mod.proto
pompon0: thx, where it is defined though?

slowli: It is equivalent to the `timestamp` field in the `miniblocks` Postgres table and this to [`MiniblockHeader.timestamp`](https://github.com/matter-labs/zksync-era/blob/f4313a4e5a67f616d2dfa8f364c47cb73cef1ec7/core/lib/types/src/block.rs#L74). It maps to multiple fields in the state keeper / VM types (e.g., [`L2BlockEnv.timestamp`](https://github.com/matter-labs/zksync-era/blob/f4313a4e5a67f616d2dfa8f364c47cb73cef1ec7/core/lib/multivm/src/interface/types/inputs/l2_block.rs#L6) and [`MiniblockParams.timestamp`](https://github.com/matter-labs/zksync-era/blob/f4313a4e5a67f616d2dfa8f364c47cb73cef1ec7/core/lib/zksync_core/src/state_keeper/io/mod.rs#L53)). AFAICT, the fact that it's seconds since epoch is implicit / tribal knowledge; see e.g. [`sleep_past`](https://github.com/matter-labs/zksync-era/blob/f4313a4e5a67f616d2dfa8f364c47cb73cef1ec7/core/lib/zksync_core/src/state_keeper/io/mempool.rs#L357) implementation. Also, there's a fact that EVM block timestamps are measured in seconds. I think that it can be quite confusing, and optimally timestamps should have a newtype wrapper.

core/lib/zksync_core/src/consensus/payload.rs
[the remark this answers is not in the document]
slowli: I'm not 100% sure about future use cases, but I could imagine a scenario in which other crates would want to take a look at the (consensus) block payload. Alternatively, we may want to split `zksync_core` into multiple crates in the longer term. But without concrete evidence, these are just hypotheses, so it's fine to leave this definition as-is.

core/lib/zksync_core/src/consensus/proto/mod.proto
```diff
@@ -0,0 +1,21 @@
+syntax = "proto3";
+
+package zksync.zksync_core.consensus;
```
slowli, line 3: Bikeshedding: Duplicating `zksync` in the package path doesn't look particularly good. Maybe, it makes sense to shorten it to `zksync.core.consensus` or use something like `zksync.server.consensus`?

core/lib/types/src/block.rs
```diff
@@ -85,6 +88,43 @@ pub struct MiniblockHeader {
...
+    fn serialize<S: serde::Serializer>(&self, s: S) -> Result<S::Ok, S::Error> {
+        zksync_protobuf::serde::serialize(self, s)
+    }
+}
+
+impl<'de> Deserialize<'de> for ConsensusBlockFields {
+    fn deserialize<D: serde::Deserializer<'de>>(d: D) -> Result<Self, D::Error> {
+        zksync_protobuf::serde::deserialize(d)
+    }
+}
```
slowli, line 126: Couldn't you use new macros in `zksync_protobuf::serde` to shorten this boilerplate? Or do you prefer explicitness?

slowli (commented):
Other than compilation problems (IIUC, related to selecting the `tokio` version), looks good.

core/lib/zksync_core/src/consensus/proto/mod.proto
pompon0: I want the crates to have disjoint proto namespaces, zksync.core.consensus for zksync_core should be ok.

core/lib/types/src/block.rs
pompon0: there is no new macros there now.

RomanBrodetski (approved):

PR #498, opened 2023-11-15
feat: Snapshot Creator

tomg10:
## What ❔
Snapshot creator is small command line tool for creating a snapshot of zkSync node for EN node to be able to initialize
to a certain L1 Batch.

Snapshots do not contain full transactions history, but rather a minimal subset of information needed to bootstrap EN
node.

## Checklist

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #
 # Pre-commit hook verifying that inappropriate code will not be committed.
 
--- a/core/bin/snapshots_creator/README.md
+++ b/core/bin/snapshots_creator/README.md
@@ -58,7 +58,7 @@ pub struct SnapshotStorageLog {
 
 ```rust
 pub struct SnapshotFactoryDependencies {
-		pub factory_deps: Vec<SnapshotFactoryDependency>
+  pub factory_deps: Vec<SnapshotFactoryDependency>
 }
 
 pub struct SnapshotFactoryDependency {
--- a/core/bin/snapshots_creator/README.md
+++ b/core/bin/snapshots_creator/README.md
@@ -1,0 +1,68 @@
+# Snapshots Creator
+
+Snapshot creator is small command line tool for creating a snapshot of zkSync node for EN node to be able to initialize
+to a certain L1 Batch.
+
+Snapshots do not contain full transactions history, but rather a minimal subset of information needed to bootstrap EN
+node.
+
+Usage (local development):\
+First run `zk env dev` \
+then the creator can be run using:  
+`cargo run --bin snapshots_creator --release`
+
+Snapshot contents can be stored based on blob_store config either in local filesystem or GS.
+
+## Snapshots format
+
+Each snapshot consists of three types of objects: header, storage logs chunks and factory deps:
+
+- Snapshot Header (currently returned by snapshots namespace of JSON-RPC API)
+
+```rust
+pub struct SnapshotHeader {
+    pub l1_batch_number: L1BatchNumber,
+    pub miniblock_number: MiniblockNumber,
+    // ordered by chunk_id
+    pub storage_logs_chunks: Vec<SnapshotStorageLogsChunkMetadata>,
+    pub factory_deps_filepath: String,
+    pub last_l1_batch_with_metadata: L1BatchWithMetadata,
+    pub generated_at: DateTime<Utc>,
+}
+
+pub struct SnapshotStorageLogsChunkMetadata {
+    pub chunk_id: u64,
+    // can be either a gs or filesystem path
+    pub filepath: String,
+}
+```
+
+- Snapshot Storage logs chunks (most likely to be stored in gzipped protobuf files, but this part is still WIP) :
+
+```rust
+pub struct SnapshotStorageLogsChunk {
+    // sorted by hashed_keys interpreted as little-endian numbers
+    pub storage_logs: Vec<SnapshotStorageLog>,
+}
+
+// "most recent" for each key together with info when the key was first used
+pub struct SnapshotStorageLog {
+    pub key: StorageKey,
+    pub value: StorageValue,
+    pub l1_batch_number_of_initial_write: L1BatchNumber,
+    pub enumeration_index: u64,
+}
+```
+
+- Factory dependencies (most likely to be stored as protobufs in the very near future)
+
+```rust
+pub struct SnapshotFactoryDependencies {
+		pub factory_deps: Vec<SnapshotFactoryDependency>
+}
+
+pub struct SnapshotFactoryDependency {
+    pub bytecode_hash: H256,
+    pub bytecode: Vec<u8>,
+}
+```
--- a/core/bin/snapshots_creator/src/main.rs
+++ b/core/bin/snapshots_creator/src/main.rs
@@ -1,0 +1,251 @@
+use anyhow::Context as _;
+use prometheus_exporter::PrometheusExporterConfig;
+use tokio::sync::watch;
+use tokio::sync::watch::Receiver;
+use vise::{Gauge, Metrics};
+use zksync_config::configs::PrometheusConfig;
+use zksync_config::PostgresConfig;
+
+use zksync_dal::ConnectionPool;
+use zksync_env_config::object_store::SnapshotsObjectStoreConfig;
+use zksync_env_config::FromEnv;
+use zksync_object_store::{ObjectStore, ObjectStoreFactory};
+use zksync_types::snapshots::{
+    SnapshotFactoryDependencies, SnapshotStorageLogsChunk, SnapshotStorageLogsStorageKey,
+};
+use zksync_types::zkevm_test_harness::zk_evm::zkevm_opcode_defs::decoding::AllowedPcOrImm;
+use zksync_types::{L1BatchNumber, MiniblockNumber};
+use zksync_utils::ceil_div;
+use zksync_utils::time::seconds_since_epoch;
+
+#[derive(Debug, Metrics)]
+#[metrics(prefix = "snapshots_creator")]
+struct SnapshotsCreatorMetrics {
+    pub storage_logs_chunks_count: Gauge<u64>,
+
+    pub snapshot_generation_duration: Gauge<u64>,
+
+    pub snapshot_l1_batch: Gauge<u64>,
+
+    pub snapshot_generation_timestamp: Gauge<u64>,
+}
+#[vise::register]
+pub(crate) static METRICS: vise::Global<SnapshotsCreatorMetrics> = vise::Global::new();
+
+async fn maybe_enable_prometheus_metrics(stop_receiver: Receiver<bool>) -> anyhow::Result<()> {
+    let prometheus_config = PrometheusConfig::from_env().ok();
+    if let Some(prometheus_config) = prometheus_config {
+        let exporter_config = PrometheusExporterConfig::push(
+            prometheus_config.gateway_endpoint(),
+            prometheus_config.push_interval(),
+        );
+
+        tracing::info!("Starting prometheus exporter with config {prometheus_config:?}");
+        tokio::spawn(exporter_config.run(stop_receiver));
+    } else {
+        tracing::info!("Starting without prometheus exporter");
+    }
+    Ok(())
+}
+async fn process_storage_logs_single_chunk(
+    blob_store: &dyn ObjectStore,
+    pool: &ConnectionPool,
+    l1_batch_number: L1BatchNumber,
+    chunk_id: u64,
+    chunk_size: u64,
+    chunks_count: u64,
+) -> anyhow::Result<String> {
+    let mut conn = pool.access_storage().await?;
+    let logs = conn
+        .snapshots_creator_dal()
+        .get_storage_logs_chunk(l1_batch_number, chunk_id, chunk_size)
+        .await
+        .context("Error fetching storage logs count")?;
+    let storage_logs_chunk = SnapshotStorageLogsChunk { storage_logs: logs };
+    let key = SnapshotStorageLogsStorageKey {
+        l1_batch_number,
+        chunk_id,
+    };
+    let filename = blob_store
+        .put(key, &storage_logs_chunk)
+        .await
+        .context("Error storing storage logs chunk in blob store")?;
+
+    let output_filepath_prefix = blob_store.get_storage_prefix::<SnapshotStorageLogsChunk>();
+    let output_filepath = format!("{output_filepath_prefix}/{filename}");
+
+    tracing::info!(
+        "Finished storing storage logs chunk {}/{chunks_count}, output stored in {output_filepath}",
+        chunk_id + 1,
+    );
+    Ok(output_filepath)
+}
+
+async fn process_factory_deps(
+    blob_store: &dyn ObjectStore,
+    pool: &ConnectionPool,
+    miniblock_number: MiniblockNumber,
+    l1_batch_number: L1BatchNumber,
+) -> anyhow::Result<String> {
+    tracing::info!("Processing factory dependencies");
+    let mut conn = pool.access_storage().await?;
+    let factory_deps = conn
+        .snapshots_creator_dal()
+        .get_all_factory_deps(miniblock_number)
+        .await;
+    let factory_deps = SnapshotFactoryDependencies { factory_deps };
+    let filename = blob_store
+        .put(l1_batch_number, &factory_deps)
+        .await
+        .context("Error storing factory deps in blob store")?;
+    let output_filepath_prefix = blob_store.get_storage_prefix::<SnapshotFactoryDependencies>();
+    let output_filepath = format!("{output_filepath_prefix}/{filename}");
+    tracing::info!(
+        "Finished processing factory dependencies, output stored in {}",
+        output_filepath
+    );
+    Ok(output_filepath)
+}
+
+async fn run(blob_store: Box<dyn ObjectStore>, pool: ConnectionPool) -> anyhow::Result<()> {
+    let mut conn = pool.access_storage().await?;
+    let start_time = seconds_since_epoch();
+
+    let l1_batch_number = conn.blocks_dal().get_sealed_l1_batch_number().await? - 1; // we subtract 1 so that after restore, EN node has at least one l1 batch to fetch
+
+    if !conn
+        .snapshots_dal()
+        .get_snapshot_metadata(l1_batch_number)
+        .await?
+        .is_none()
+    {
+        tracing::info!(
+            "Snapshot for L1 batch number {} already exists, exiting",
+            l1_batch_number
+        );
+        return Ok(());
+    }
+
+    let miniblock_number = conn
+        .blocks_dal()
+        .get_miniblock_range_of_l1_batch(l1_batch_number)
+        .await?
+        .unwrap()
+        .1;
+    let storage_logs_count = conn
+        .snapshots_creator_dal()
+        .get_storage_logs_count(l1_batch_number)
+        .await?;
+
+    drop(conn);
+
+    //TODO load this from config
+    let chunk_size = 1_000_000;
+    let chunks_count = ceil_div(storage_logs_count, chunk_size);
+
+    tracing::info!(
+        "Creating snapshot for storage logs up to miniblock {}, l1_batch {}",
+        miniblock_number,
+        l1_batch_number.0
+    );
+    tracing::info!(
+        "{} chunks of max size {} will be generated",
+        chunks_count,
+        chunk_size
+    );
+
+    let factory_deps_output_file =
+        process_factory_deps(&*blob_store, &pool, miniblock_number, l1_batch_number).await?;
+
+    let mut storage_logs_output_files = vec![];
+
+    for chunk_id in 0..chunks_count {
+        tracing::info!(
+            "Processing storage logs chunk {}/{chunks_count}",
+            chunk_id + 1
+        );
+        let output_file = process_storage_logs_single_chunk(
+            &*blob_store,
+            &pool,
+            l1_batch_number,
+            chunk_id,
+            chunk_size,
+            chunks_count,
+        )
+        .await?;
+        storage_logs_output_files.push(output_file.clone());
+    }
+
+    let mut conn = pool.access_storage().await?;
+
+    conn.snapshots_dal()
+        .add_snapshot(
+            l1_batch_number,
+            &storage_logs_output_files,
+            factory_deps_output_file,
+        )
+        .await?;
+
+    METRICS.snapshot_l1_batch.set(l1_batch_number.0.as_u64());
+    METRICS.storage_logs_chunks_count.set(chunks_count);
+    METRICS
+        .snapshot_generation_timestamp
+        .set(seconds_since_epoch());
+    METRICS
+        .snapshot_generation_duration
+        .set(seconds_since_epoch() - start_time);
+
+    tracing::info!(
+        r#"Run metrics:
+snapshot_generation_duration: {}sec
+snapshot_l1_batch: {},
+snapshot_generation_timestamp: {}
+storage_logs_chunks_count: {}
+  "#,
+        METRICS.snapshot_generation_duration.get(),
+        METRICS.snapshot_l1_batch.get(),
+        METRICS.snapshot_generation_timestamp.get(),
+        METRICS.storage_logs_chunks_count.get()
+    );
+
+    Ok(())
+}
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    let (stop_sender, stop_receiver) = watch::channel(false);
+
+    tracing::info!("Starting snapshots creator");
+    #[allow(deprecated)] // TODO (QIT-21): Use centralized configuration approach.
+    let log_format = vlog::log_format_from_env();
+    #[allow(deprecated)] // TODO (QIT-21): Use centralized configuration approach.
+    let sentry_url = vlog::sentry_url_from_env();
+    #[allow(deprecated)] // TODO (QIT-21): Use centralized configuration approach.
+    let environment = vlog::environment_from_env();
+
+    maybe_enable_prometheus_metrics(stop_receiver).await?;
+    let mut builder = vlog::ObservabilityBuilder::new().with_log_format(log_format);
+    if let Some(sentry_url) = sentry_url {
+        builder = builder
+            .with_sentry_url(&sentry_url)
+            .context("Invalid Sentry URL")?
+            .with_sentry_environment(environment);
+    }
+    let _guard = builder.build();
+
+    let object_store_config =
+        SnapshotsObjectStoreConfig::from_env().context("SnapshotsObjectStoreConfig::from_env()")?;
+    let blob_store = ObjectStoreFactory::new(object_store_config.0)
+        .create_store()
+        .await;
+
+    let postgres_config = PostgresConfig::from_env().context("PostgresConfig")?;
+    let pool = ConnectionPool::singleton(postgres_config.replica_url()?)
+        .build()
+        .await?;
+
+    run(blob_store, pool).await?;
+    tracing::info!("Finished running snapshot creator!");
+    stop_sender.send(true).ok();
+    Ok(())
+}
--- a/core/lib/dal/src/snapshots_creator_dal.rs
+++ b/core/lib/dal/src/snapshots_creator_dal.rs
@@ -1,0 +1,110 @@
+use crate::instrument::InstrumentExt;
+use crate::StorageProcessor;
+use zksync_types::snapshots::{SnapshotFactoryDependency, SnapshotStorageLog};
+use zksync_types::{AccountTreeId, Address, L1BatchNumber, MiniblockNumber, StorageKey, H256};
+
+#[derive(Debug)]
+pub struct SnapshotsCreatorDal<'a, 'c> {
+    pub(crate) storage: &'a mut StorageProcessor<'c>,
+}
+
+impl SnapshotsCreatorDal<'_, '_> {
+    pub async fn get_storage_logs_count(
+        &mut self,
+        l1_batch_number: L1BatchNumber,
+    ) -> Result<u64, sqlx::Error> {
+        let count = sqlx::query!(
+            "SELECT count(*) FROM initial_writes WHERE l1_batch_number <= $1",
+            l1_batch_number.0 as i32
+        )
+        .fetch_one(self.storage.conn())
+        .await
+        .unwrap()
+        .count
+        .unwrap();
+        Ok(count as u64)
+    }
+
+    pub async fn get_storage_logs_chunk(
+        &mut self,
+        l1_batch_number: L1BatchNumber,
+        chunk_id: u64,
+        chunk_size: u64,
+    ) -> Result<Vec<SnapshotStorageLog>, sqlx::Error> {
+        let miniblock_number: i64 = sqlx::query!(
+            "select MAX(number) from miniblocks where l1_batch_number = $1",
+            l1_batch_number.0 as i64
+        )
+        .fetch_one(self.storage.conn())
+        .await?
+        .max
+        .unwrap_or_default();
+
+        let storage_logs = sqlx::query!(
+            r#"
+            SELECT storage_logs.key,
+                   storage_logs.value,
+                   storage_logs.address,
+                   storage_logs.miniblock_number,
+                   initial_writes.l1_batch_number,
+                   initial_writes.index
+            FROM (SELECT hashed_key,
+                         max(ARRAY [miniblock_number, operation_number]::int[]) AS op
+                  FROM storage_logs
+                  WHERE miniblock_number <= $1
+                  GROUP BY hashed_key
+                  ORDER BY hashed_key) AS keys
+                     INNER JOIN storage_logs ON keys.hashed_key = storage_logs.hashed_key
+                AND storage_logs.miniblock_number = keys.op[1]
+                AND storage_logs.operation_number = keys.op[2]
+                     INNER JOIN initial_writes ON keys.hashed_key = initial_writes.hashed_key
+            WHERE miniblock_number <= $1
+            ORDER BY ARRAY(
+                 SELECT get_byte(storage_logs.hashed_key, 32 - generate_series)
+                 FROM generate_series(1, 32)
+             )
+            LIMIT $2 OFFSET $3;
+             "#,
+            miniblock_number,
+            chunk_size as i64,
+            (chunk_size * chunk_id) as i64
+        )
+        .instrument("get_storage_logs_chunk")
+        .report_latency()
+        .fetch_all(self.storage.conn())
+        .await?
+        .iter()
+        .map(|row| SnapshotStorageLog {
+            key: StorageKey::new(
+                AccountTreeId::new(Address::from_slice(&row.address)),
+                H256::from_slice(&row.key),
+            ),
+            value: H256::from_slice(&row.value),
+            l1_batch_number_of_initial_write: L1BatchNumber(row.l1_batch_number as u32),
+            enumeration_index: row.index as u64,
+        })
+        .collect();
+        Ok(storage_logs)
+    }
+
+    pub async fn get_all_factory_deps(
+        &mut self,
+        miniblock_number: MiniblockNumber,
+    ) -> Vec<SnapshotFactoryDependency> {
+        sqlx::query!(
+            "SELECT bytecode, bytecode_hash FROM factory_deps WHERE miniblock_number <= $1",
+            miniblock_number.0 as i64,
+        )
+        .instrument("get_all_factory_deps")
+        .report_latency()
+        .fetch_all(self.storage.conn())
+        .await
+        .unwrap()
+        .into_iter()
+        .map(|row| SnapshotFactoryDependency {
+            bytecode_hash: H256::from_slice(&row.bytecode_hash),
+            bytecode: row.bytecode,
+        })
+        .collect()
+    }
+}
--- a/core/lib/dal/src/snapshots_dal.rs
+++ b/core/lib/dal/src/snapshots_dal.rs
@@ -1,0 +1,146 @@
+use crate::StorageProcessor;
+use sqlx::types::chrono::{DateTime, Utc};
+use zksync_types::snapshots::{AllSnapshots, SnapshotMetadata};
+use zksync_types::L1BatchNumber;
+
+#[derive(Debug)]
+pub struct SnapshotsDal<'a, 'c> {
+    pub(crate) storage: &'a mut StorageProcessor<'c>,
+}
+
+impl SnapshotsDal<'_, '_> {
+    pub async fn add_snapshot(
+        &mut self,
+        l1_batch_number: L1BatchNumber,
+        storage_logs_filepaths: &[String],
+        factory_deps_filepaths: String,
+    ) -> Result<(), sqlx::Error> {
+        sqlx::query!(
+            "INSERT INTO snapshots (l1_batch_number, created_at, storage_logs_filepaths, factory_deps_filepath) \
+             VALUES ($1, now(), $2, $3)",
+            l1_batch_number.0 as i32,
+            storage_logs_filepaths,
+            factory_deps_filepaths,
+        )
+        .execute(self.storage.conn())
+        .await?;
+        Ok(())
+    }
+
+    pub async fn get_all_snapshots(&mut self) -> Result<AllSnapshots, sqlx::Error> {
+        let records: Vec<SnapshotMetadata> = sqlx::query!(
+            "SELECT l1_batch_number, created_at, factory_deps_filepath FROM snapshots"
+        )
+        .fetch_all(self.storage.conn())
+        .await?
+        .into_iter()
+        .map(|r| SnapshotMetadata {
+            l1_batch_number: L1BatchNumber(r.l1_batch_number as u32),
+            generated_at: DateTime::<Utc>::from_naive_utc_and_offset(r.created_at, Utc),
+            factory_deps_filepath: r.factory_deps_filepath,
+        })
+        .collect();
+        Ok(AllSnapshots { snapshots: records })
+    }
+
+    pub async fn get_snapshot_metadata(
+        &mut self,
+        l1_batch_number: L1BatchNumber,
+    ) -> Result<Option<SnapshotMetadata>, sqlx::Error> {
+        let record: Option<SnapshotMetadata> = sqlx::query!(
+            "SELECT l1_batch_number, created_at, factory_deps_filepath FROM snapshots WHERE l1_batch_number = $1",
+            l1_batch_number.0 as i32
+        )
+        .fetch_optional(self.storage.conn())
+        .await?
+        .map(|r| SnapshotMetadata {
+            l1_batch_number: L1BatchNumber(r.l1_batch_number as u32),
+            generated_at: DateTime::<Utc>::from_naive_utc_and_offset(r.created_at, Utc),
+            factory_deps_filepath: r.factory_deps_filepath,
+        });
+        Ok(record)
+    }
+
+    pub async fn get_snapshot_files(
+        &mut self,
+        l1_batch_number: L1BatchNumber,
+    ) -> Result<Option<Vec<String>>, sqlx::Error> {
+        let record = sqlx::query!(
+            "SELECT storage_logs_filepaths \
+            FROM snapshots WHERE l1_batch_number = $1",
+            l1_batch_number.0 as i32
+        )
+        .fetch_optional(self.storage.conn())
+        .await?;
+
+        Ok(record.map(|r| r.storage_logs_filepaths))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::ConnectionPool;
+    use zksync_types::L1BatchNumber;
+
+    #[tokio::test]
+    async fn adding_snapshot() {
+        let pool = ConnectionPool::test_pool().await;
+        let mut conn = pool.access_storage().await.unwrap();
+        let mut dal = conn.snapshots_dal();
+        let l1_batch_number = L1BatchNumber(100);
+        dal.add_snapshot(
+            l1_batch_number,
+            &[],
+            "gs:///bucket/factory_deps.bin".to_string(),
+        )
+        .await
+        .expect("Failed to add snapshot");
+
+        let snapshots = dal
+            .get_all_snapshots()
+            .await
+            .expect("Failed to retrieve snapshots");
+        assert_eq!(1, snapshots.snapshots.len());
+        assert_eq!(
+            snapshots.snapshots[0].l1_batch_number,
+            l1_batch_number as L1BatchNumber
+        );
+
+        let snapshot_metadata = dal
+            .get_snapshot_metadata(l1_batch_number)
+            .await
+            .expect("Failed to retrieve snapshot")
+            .unwrap();
+        assert_eq!(
+            snapshot_metadata.l1_batch_number,
+            l1_batch_number as L1BatchNumber
+        );
+    }
+
+    #[tokio::test]
+    async fn adding_files() {
+        let pool = ConnectionPool::test_pool().await;
+        let mut conn = pool.access_storage().await.unwrap();
+        let mut dal = conn.snapshots_dal();
+        let l1_batch_number = L1BatchNumber(100);
+        dal.add_snapshot(
+            l1_batch_number,
+            &[
+                "gs:///bucket/test_file1.bin".to_string(),
+                "gs:///bucket/test_file2.bin".to_string(),
+            ],
+            "gs:///bucket/factory_deps.bin".to_string(),
+        )
+        .await
+        .expect("Failed to add snapshot");
+
+        let files = dal
+            .get_snapshot_files(l1_batch_number)
+            .await
+            .expect("Failed to retrieve snapshot");
+        assert!(files.is_some());
+        let files = files.unwrap();
+        assert!(files.contains(&"gs:///bucket/test_file1.bin".to_string()));
+        assert!(files.contains(&"gs:///bucket/test_file2.bin".to_string()));
+    }
+}
--- a/core/lib/object_store/src/gcs.rs
+++ b/core/lib/object_store/src/gcs.rs
@@ -205,6 +205,14 @@ impl ObjectStore for GoogleCloudStorage {
     async fn remove_raw(&self, bucket: Bucket, key: &str) -> Result<(), ObjectStoreError> {
         self.remove_inner(bucket.as_str(), key).await
     }
+
+    fn get_storage_prefix_raw(&self, bucket: Bucket) -> String {
+        format!(
+            "https://storage.googleapis.com/{}/{}",
+            self.bucket_prefix.clone(),
+            bucket.as_str()
+        )
+    }
 }
 
 #[cfg(test)]
--- a/core/lib/object_store/src/mock.rs
+++ b/core/lib/object_store/src/mock.rs
@@ -45,4 +45,8 @@ impl ObjectStore for MockStore {
         bucket_map.remove(key);
         Ok(())
     }
+
+    fn get_storage_prefix_raw(&self, bucket: Bucket) -> String {
+        format!("{bucket}")
+    }
 }
--- a/core/lib/object_store/src/objects.rs
+++ b/core/lib/object_store/src/objects.rs
@@ -1,8 +1,12 @@
 //! Stored objects.
 
+use flate2::{read::GzDecoder, write::GzEncoder, Compression};
+use std::io::Read;
 use zksync_types::aggregated_operations::L1BatchProofForL1;
+use zksync_types::snapshots::{SnapshotFactoryDependencies, SnapshotStorageLogsStorageKey};
 use zksync_types::{
     proofs::{AggregationRound, PrepareBasicCircuitsJob},
+    snapshots::SnapshotStorageLogsChunk,
     storage::witness_block_state::WitnessBlockState,
     zkevm_test_harness::{
         abstract_zksync_circuit::concrete_circuits::ZkSyncCircuit,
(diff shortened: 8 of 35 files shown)


core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,251 @@
...
+
+    conn.snapshots_dal()
+        .add_snapshot(
+            l1_batch_number,
+            &storage_logs_output_files,
+            factory_deps_output_file,
+        )
+        .await?;
+
+    METRICS.snapshot_l1_batch.set(l1_batch_number.0.as_u64());
```
slowli, line 189: Why do you set this metric here, and not at the start of `run()`?

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,251 @@
...
+    METRICS.storage_logs_chunks_count.set(chunks_count);
+    METRICS
+        .snapshot_generation_timestamp
+        .set(seconds_since_epoch());
```
slowli, line 193: Is this really a useful metric? IMO, other metrics you've defined make sense, but I struggle to understand what will this one allow to measure.

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,251 @@
...
+    conn.snapshots_dal()
+        .add_snapshot(
+            l1_batch_number,
+            &storage_logs_output_files,
+            factory_deps_output_file,
+        )
+        .await?;
+
+    METRICS.snapshot_l1_batch.set(l1_batch_number.0.as_u64());
+    METRICS.storage_logs_chunks_count.set(chunks_count);
```
slowli, line 190: It makes sense to set this metric after processing each chunk in order to track snapshot creation progress.

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,251 @@
...
+    METRICS
+        .snapshot_generation_timestamp
+        .set(seconds_since_epoch());
+    METRICS
+        .snapshot_generation_duration
+        .set(seconds_since_epoch() - start_time);
```
slowli, line 196: IMO, it would make more sense to track processing of a single chunk, so that this metric is more "dynamic" and allows to measure the health / speed of snapshot creation while it's in progress. Optionally, you can split it by stages (loading from Postgres, saving to GCS for now; saving to Postgres when fault-tolerance is implemented). It would be a `Histogram<Duration>`; see other metric definitions for the examples of usage, such as [in `MetadataCalculator`](https://github.com/matter-labs/zksync-era/blob/38bb4823c7b5e0e651d9f531feede66c24afd19f/core/lib/zksync_core/src/metadata_calculator/metrics.rs#L114).

core/lib/object_store/src/mock.rs
```diff
@@ -45,4 +45,8 @@ impl ObjectStore for MockStore {
         bucket_map.remove(key);
         Ok(())
     }
+
+    fn get_storage_prefix_raw(&self, bucket: Bucket) -> String {
+        format!("{bucket}")
```
slowli, line 50: Nit: You can use `bucket.to_string()` here.

core/lib/object_store/src/objects.rs
```diff
@@ -63,6 +67,62 @@ macro_rules! serialize_using_bincode {
...
+    }
+}
+
+impl StoredObject for SnapshotStorageLogsChunk {
+    const BUCKET: Bucket = Bucket::StorageSnapshot;
+    type Key<'a> = SnapshotStorageLogsStorageKey;
+
+    fn encode_key(key: Self::Key<'_>) -> String {
+        format!(
+            "snapshot_l1_batch_{}_storage_logs_part_{:0<3}.json.gzip",
```
slowli, line 104: Shouldn't it be `{:0>3}` (i.e., the formatter chunk is *right*-aligned)? Could you add a unit test to cover this?

core/lib/zksync_core/src/api_server/web3/namespaces/snapshots.rs
```diff
@@ -0,0 +1,79 @@
...
+            state: self.state.clone(),
+        }
+    }
+}
+impl<G: L1GasPriceProvider> SnapshotsNamespace<G> {
+    pub fn new(state: RpcState<G>) -> Self {
+        Self { state }
+    }
+    pub async fn get_all_snapshots_impl(&self) -> Result<AllSnapshots, Web3Error> {
+        let mut storage_processor = self.state.connection_pool.access_storage().await.unwrap();
```
slowli, line 24: - Please use `access_storage_tagged("api")` (see a comment above).
- Please propagate errors using `internal_error`.

core/lib/zksync_core/src/api_server/web3/namespaces/snapshots.rs
```diff
@@ -0,0 +1,79 @@
...
+            .unwrap();
+        if snapshot_files.is_none() {
+            Ok(None)
+        } else {
+            let snapshot_metadata = snapshots_dal
+                .get_snapshot_metadata(l1_batch_number)
+                .await
+                .unwrap()
+                .unwrap();
+            let snapshot_files = snapshot_files.as_ref().unwrap();
```
slowli, line 47: To be more idiomatic, you can match `snapshot_files`:

```rust
let Some(snapshot_files) = snapshot_files else {
    return Ok(None);
};
// work with files
```

core/lib/types/src/snapshots.rs
```diff
@@ -0,0 +1,80 @@
...
+pub struct SnapshotMetadata {
+    pub l1_batch_number: L1BatchNumber,
+    pub generated_at: DateTime<Utc>,
+    pub factory_deps_filepath: String,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct SnapshotHeader {
+    pub l1_batch_number: L1BatchNumber,
```
slowli, line 24: Nit: This is duplicated in `last_l1_batch_with_metadata.header.number`.

core/lib/types/src/snapshots.rs
```diff
@@ -0,0 +1,80 @@
...
+    pub miniblock_number: MiniblockNumber,
+    pub storage_logs_chunks: Vec<SnapshotStorageLogsChunkMetadata>,
+    pub factory_deps_filepath: String,
+    pub last_l1_batch_with_metadata: L1BatchWithMetadata,
```
slowli, line 28: `L1BatchWithMetadata` has a lot of unnecessary info, particularly in `metadata` and `factory_deps` fields. Do we really want to pass it in its entirety, or could it make sense to create a struct specifically for snapshot recovery? cc @RomanBrodetski

core/lib/object_store/src/objects.rs
```diff
@@ -63,6 +67,62 @@ macro_rules! serialize_using_bincode {
     };
 }
 
+/// Derives [`StoredObject::serialize()`] and [`StoredObject::deserialize()`] using
+/// the `json` (de)serializer. Should be used in `impl StoredObject` blocks.
```
slowli, line 71: ???

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,251 @@
...
+    tracing::info!(
+        "Starting to generate {} chunks of max size {}",
+        chunks_count,
+        chunk_size
+    );
+
+    let factory_deps_output_file =
+        process_factory_deps(&*blob_store, &pool, miniblock_number, l1_batch_number).await?;
+
+    let mut storage_logs_output_files = vec![];
```
slowli, line 160: Nit: You can specify the correct capacity for the created `Vec`: `Vec::with_capacity(chunks_count)`.

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,251 @@
...
+    }
+    Ok(())
+}
+async fn process_storage_logs_single_chunk(
+    blob_store: &dyn ObjectStore,
+    pool: &ConnectionPool,
+    l1_batch_number: L1BatchNumber,
+    chunk_id: u64,
+    chunk_size: u64,
+    chunks_count: u64,
```
slowli, line 56: Nit: Is this arg just used for logging? If so, I'd move logging to the caller site, so that passing this arg is not necessary.

slowli (changes requested):

core/bin/snapshots_creator/src/main.rs
tomg10: I wanted it so that I could later define an alert on "last snapthot age"

core/bin/snapshots_creator/src/main.rs
tomg10: I've never used such metrics like that, but it sounds like a nice idea!

core/lib/object_store/src/objects.rs
tomg10: some leftovery, which my IDE tried to parse as documentation and hid behind "Not availablee", thanks for spotting!

core/lib/types/src/snapshots.rs
tomg10: I assumed that it doesn't cost us much to jut have those few additional fields while it lets us reuse lots of existing code and that it just wouldn't be worth it.

core/bin/snapshots_creator/src/main.rs
tomg10: Does it have benefits other than better performance? I find the code a bit less readable with the suggested change

core/bin/snapshots_creator/src/main.rs
tomg10: yeah, it's used only for that. I moved this logging back and forth at least twice yesterday. I totally agree that it's not ideal to have an argument just for that and it's a bit ankward, but I find the 'main logic' code a bit more readable when the logging is moved inside the method, so in the end I left it there

core/lib/object_store/src/objects.rs
tomg10: it definitely should! I will add a test for that

core/lib/dal/src/snapshots_creator_dal.rs
```diff
@@ -0,0 +1,107 @@
...
+    pub(crate) storage: &'a mut StorageProcessor<'c>,
+}
+
+impl SnapshotsCreatorDal<'_, '_> {
+    pub async fn get_storage_logs_count(
+        &mut self,
+        l1_batch_number: L1BatchNumber,
+    ) -> Result<u64, sqlx::Error> {
+        let count = sqlx::query!(
+            "SELECT count(*) FROM initial_writes WHERE l1_batch_number <= $1",
```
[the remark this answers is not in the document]
tomg10: yes, I see now, big thanks for showing me this!

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,251 @@
...
+#[derive(Debug, Metrics)]
+#[metrics(prefix = "snapshots_creator")]
+struct SnapshotsCreatorMetrics {
+    pub storage_logs_chunks_count: Gauge<u64>,
+
+    pub snapshot_generation_duration: Gauge<u64>,
+
+    pub snapshot_l1_batch: Gauge<u64>,
+
+    pub snapshot_generation_timestamp: Gauge<u64>,
```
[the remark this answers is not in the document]
tomg10: btw, is there any particular reason, why lint doesn't catch this "pub items in private struc" ?

core/lib/dal/src/snapshots_creator_dal.rs
```diff
@@ -0,0 +1,107 @@
...
+            miniblock_number.0 as i64,
+        )
+        .instrument("get_all_factory_deps")
+        .report_latency()
+        .fetch_all(self.storage.conn())
+        .await
+        .unwrap()
+        .into_iter()
+        .map(|row| SnapshotFactoryDependency {
+            bytecode_hash: H256::from_slice(&row.bytecode_hash),
```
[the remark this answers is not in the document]
tomg10: It could be calculated during the recovery, but it did not add a big overhead and simplified recovery so decided to just pass it around.

core/lib/dal/src/snapshots_dal.rs
```diff
@@ -0,0 +1,146 @@
...
+        .await?
+        .map(|r| SnapshotMetadata {
+            l1_batch_number: L1BatchNumber(r.l1_batch_number as u32),
+            generated_at: DateTime::<Utc>::from_naive_utc_and_offset(r.created_at, Utc),
+            factory_deps_filepath: r.factory_deps_filepath,
+        });
+        Ok(record)
+    }
+
+    pub async fn get_snapshot_files(
```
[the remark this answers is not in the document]
tomg10: The only reason was not to pass files in get_all_snapshots, but I've taken this decision from the original API design without much thought, you are absolutely right that it shouldn't be that many files to justify this split.

core/bin/snapshots_creator/src/main.rs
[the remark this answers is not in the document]
slowli: It doesn't do that by default (lots of clippy lints are off by default, and we don't switch them on), but there's a lint that does [just that](https://doc.rust-lang.org/stable/nightly-rustc/rustc_lint/builtin/static.UNREACHABLE_PUB.html). You can add these and other lints on the crate level (see, e.g. [the Merkle tree crate](https://github.com/matter-labs/zksync-era/blob/ecd68ca64c18e8733d79e1e3a81a6fdb87e5b5e5/core/lib/merkle_tree/src/lib.rs#L35-L42), which doesn't have this specific lint turned on 🙃), although eventually we'd probably want to use a tool like [`cargo-cranky`](https://crates.io/crates/cargo-cranky) to unify lints throughout the codebase.

core/lib/dal/src/snapshots_creator_dal.rs
```diff
@@ -0,0 +1,107 @@
...
+            l1_batch_number.0 as i64
+        )
+        .fetch_one(self.storage.conn())
+        .await?
+        .max
+        .unwrap_or_default();
+
+        let storage_logs = sqlx::query!(
+            r#"
+            SELECT storage_logs.key,
```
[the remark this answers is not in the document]
tomg10: I'm not convinced yet, the raw strings have a few very nice features for me that ordinary string literals with line escaped do not, I can also just-copy paste the query into query console in my IDE not having to remove the line escapes manually, I can also easily format the query there, I almost never do the query formatting manually.

core/lib/dal/src/snapshots_dal.rs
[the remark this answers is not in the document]
tomg10: corrected

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,251 @@
...
+        .snapshot_generation_duration
+        .set(seconds_since_epoch() - start_time);
+
+    tracing::info!(
+        r#"Run metrics:
+snapshot_generation_duration: {}sec
+snapshot_l1_batch: {},
+snapshot_generation_timestamp: {}
+storage_logs_chunks_count: {}
+  "#,
```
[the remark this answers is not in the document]
tomg10: corrected

core/lib/types/src/snapshots.rs
tomg10: I know, but I feel that it's very natural that snapshot header contains l1_batch_number that is it's primary_key, that header is just an embedded object related to the snapshot header that accidentally happens to have this field

core/lib/object_store/src/mock.rs
tomg10: corrected

core/bin/snapshots_creator/src/main.rs
tomg10: I was thinking a quite a lot about it, but so far I don't yet see, how our operations could benefit from it. I would mostly look on total snapshot duration to spot regressions. I don't imagine looking at it during normal operation and if some particular snapshot generation fails / is very slow, the logs would be the first place to diagnose what has happened and should contain all needed info about speed of snapshot creation. Why do you think it would be useful to have those details directly in metrics?

core/lib/dal/src/snapshots_dal.rs
```diff
@@ -0,0 +1,146 @@
...
+pub struct SnapshotsDal<'a, 'c> {
+    pub(crate) storage: &'a mut StorageProcessor<'c>,
+}
+
+impl SnapshotsDal<'_, '_> {
+    pub async fn add_snapshot(
+        &mut self,
+        l1_batch_number: L1BatchNumber,
+        storage_logs_filepaths: &[String],
+        factory_deps_filepaths: String,
```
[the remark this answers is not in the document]
tomg10: corrected

core/bin/snapshots_creator/src/main.rs
slowli: Performance is the main benefit. TBH, I don't see how `Vec::with_capacity(chunks_count)` is less readable than `vec![]`, but it's personal.

core/bin/snapshots_creator/src/main.rs
slowli: It would allow to track that a particular snapshot generation is slow (or more generally, will allow to track its progress). I may be missing something, but how would it be tracked otherwise using metrics? I want to have this info on a Grafana dashboard to be able to assess snapshot generation health just by looking at; additionally, metrics could be used for altering. As another use case: measure the impact of integrations (GCS / Postgres split) depending on their configuration or code changes.

core/lib/dal/src/snapshots_creator_dal.rs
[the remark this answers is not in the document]
slowli: I understand the benefits of raw string literals, but IMO they do not outweigh the drawbacks (formatting-dependent compilation and query searching). E.g., you can work with a query in the IDE before "saving" it in the final place in `sqlx::query!` (yeah, it's a one-way workflow). There's also a question of code consistency; we want to have a common style for DB queries, and while the old ones not always satisfy this style (the style itself was introduced relatively recently and isn't precisely formalized), most newer ones do. @RomanBrodetski @popzxc WDYT?

core/lib/types/src/snapshots.rs
slowli: The problem is not the cost, but rather compatibility that we'll probably need to care about (e.g., we want to change `L1BatchWithMetadata` and it now suddenly breaks snapshot recovery). Also, a cognitive overhead – if we use a dedicated data struct for snapshot generation, we *by design* know which data we use and why it's included (so it's easier to maintain), while with `L1BatchWithMetadata` it would be easier to end up with spaghetti code relying on `L1BatchWithMetadata` intrinsics.

RomanBrodetski (commented):
looks good, just left some minor comments

core/lib/types/src/snapshots.rs
```diff
@@ -0,0 +1,81 @@
...
+    pub bytecode: Vec<u8>,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct AppliedSnapshotStatus {
+    pub l1_batch_number: L1BatchNumber,
+    pub is_finished: bool,
+    pub last_finished_chunk_id: Option<u64>,
+}
```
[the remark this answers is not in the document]
tomg10: not really, I missed this when splitting that bigger PR; corrected

core/bin/snapshots_creator/README.md
```diff
@@ -0,0 +1,68 @@
...
+Snapshot contents can be stored based on blob_store config either in local filesystem or GS.
+
+## Snapshots format
+
+Each snapshot consists of three types of objects: header, storage logs chunks and factory deps:
+
+- Snapshot Header (currently returned by snapshots namespace of JSON-RPC API)
+
+```rust
+pub struct SnapshotHeader {
```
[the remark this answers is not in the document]
tomg10: corrected

core/bin/snapshots_creator/src/main.rs
tomg10: I wasn't aware it was possible to extract 'last update time' from prometheus, thanks! I removed this metric

core/bin/snapshots_creator/src/main.rs
tomg10: I added some additional fine-grained metrics

core/lib/dal/src/snapshots_creator_dal.rs
```diff
@@ -0,0 +1,112 @@
...
+    ) -> Result<u64, sqlx::Error> {
+        let count = sqlx::query!(
+            r#"
+            SELECT index
+            FROM initial_writes
+            WHERE l1_batch_number <= $1
+            ORDER BY l1_batch_number DESC , index DESC 
+            LIMIT 1;
+            "#,
+            l1_batch_number.0 as i32
```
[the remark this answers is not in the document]
tomg10: done

core/lib/dal/src/snapshots_creator_dal.rs
```diff
@@ -0,0 +1,107 @@
...
+        .count
+        .unwrap();
+        Ok(count as u64)
+    }
+
+    pub async fn get_storage_logs_chunk(
+        &mut self,
+        l1_batch_number: L1BatchNumber,
+        chunk_id: u64,
+        chunk_size: u64,
```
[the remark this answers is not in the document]
tomg10: corrected

core/lib/dal/src/snapshots_creator_dal.rs
```diff
@@ -0,0 +1,107 @@
...
+    ) -> Result<Vec<SnapshotStorageLog>, sqlx::Error> {
```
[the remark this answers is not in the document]
tomg10: corrected

core/lib/types/src/snapshots.rs
tomg10: I will work on this in snapshots applier PR

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,258 @@
...
+        .unwrap()
+        .1;
+    let storage_logs_count = conn
+        .snapshots_creator_dal()
+        .get_storage_logs_count(l1_batch_number)
+        .await?;
+
+    drop(conn);
+
+    //TODO load this from config
```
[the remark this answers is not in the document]
tomg10: corrected

core/lib/dal/src/snapshots_creator_dal.rs
```diff
@@ -0,0 +1,107 @@
...
+        let miniblock_number: i64 = sqlx::query!(
+            "select MAX(number) from miniblocks where l1_batch_number = $1",
+            l1_batch_number.0 as i64
+        )
+        .fetch_one(self.storage.conn())
+        .await?
+        .max
+        .unwrap_or_default();
```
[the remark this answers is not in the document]
tomg10: corrected, though this was definitely a "nit" for me

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,273 @@
...
+    l1_batch_number: L1BatchNumber,
+) -> anyhow::Result<String> {
+    let latency = METRICS.factory_deps_processing_durations.start();
+    tracing::info!("Processing factory dependencies");
+    let mut conn = pool.access_storage_tagged("snapshots_creator").await?;
+    let factory_deps = conn
+        .snapshots_creator_dal()
+        .get_all_factory_deps(miniblock_number)
+        .await?;
+    let factory_deps = SnapshotFactoryDependencies { factory_deps };
```
slowli, line 107: You can drop a connection here so that it's not retained when not required (over persisting a potentially bulky file in GCS). Ditto for `process_storage_logs_single_chunk`.

slowli (commented):
Looks much better than before 👍  I've got mostly nits, but there are a couple of *real* questions / suggestions as well.

core/lib/zksync_core/src/api_server/web3/namespaces/snapshots.rs
```diff
@@ -0,0 +1,89 @@
...
+                    chunk_id: chunk_id as u64,
+                    filepath: filepath.clone(),
+                })
+                .collect();
+            let l1_batch_with_metadata = storage_processor
+                .blocks_dal()
+                .get_l1_batch_metadata(l1_batch_number)
+                .await
+                .map_err(|err| internal_error(method_name, err))?
+                .unwrap();
```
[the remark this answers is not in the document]
tomg10: I think I will just remove the unwrap and replace it with correct error handling, the code would be cleaner this way

core/lib/dal/src/snapshots_creator_dal.rs
```diff
@@ -0,0 +1,104 @@
...
+pub struct SnapshotsCreatorDal<'a, 'c> {
+    pub(crate) storage: &'a mut StorageProcessor<'c>,
+}
+
+impl SnapshotsCreatorDal<'_, '_> {
+    pub async fn get_storage_logs_count(
+        &mut self,
+        l1_batch_number: L1BatchNumber,
+    ) -> sqlx::Result<u64> {
+        let count = sqlx::query!(
```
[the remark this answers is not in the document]
tomg10: We've never added support for instrumentation for it. I don't feel strong enough with my rust to add it now.

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,273 @@
...
+    Ok(output_filepath)
+}
+
+async fn run(blob_store: Box<dyn ObjectStore>, pool: ConnectionPool) -> anyhow::Result<()> {
+    let config = SnapshotsCreatorConfig::from_env().context("SnapshotsCreatorConfig::from_env")?;
+
+    let mut conn = pool.access_storage_tagged("snapshots_creator").await?;
+    let start_time = seconds_since_epoch();
+
+    let l1_batch_number = conn.blocks_dal().get_sealed_l1_batch_number().await? - 1; // we subtract 1 so that after restore, EN node has at least one l1 batch to fetch
```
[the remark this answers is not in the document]
tomg10: corrected

core/bin/snapshots_creator/src/main.rs
tomg10: corrected

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,273 @@
...
+    #[metrics(buckets = Buckets::LATENCIES)]
+    factory_deps_processing_durations: Histogram<Duration>,
+}
+#[vise::register]
+pub(crate) static METRICS: vise::Global<SnapshotsCreatorMetrics> = vise::Global::new();
+
+async fn maybe_enable_prometheus_metrics(stop_receiver: Receiver<bool>) -> anyhow::Result<()> {
+    let prometheus_config = PrometheusConfig::from_env().ok();
+    if let Some(prometheus_config) = prometheus_config {
+        let exporter_config = PrometheusExporterConfig::push(
```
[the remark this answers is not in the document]
tomg10: I've started a discussion on this topic on devops channel

core/lib/dal/src/snapshots_creator_dal.rs
```diff
@@ -0,0 +1,104 @@
...
+            ORDER BY l1_batch_number DESC , index DESC 
+            LIMIT 1;
+            "#,
+            l1_batch_number.0 as i32
+        )
+        .instrument("get_storage_logs_count")
+        .report_latency()
+        .fetch_one(self.storage.conn())
+        .await
+        .unwrap()
```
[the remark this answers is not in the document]
tomg10: corrected

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,273 @@
...
+        return Ok(());
+    }
+
+    let last_miniblock_number_in_batch = conn
+        .blocks_dal()
+        .get_miniblock_range_of_l1_batch(l1_batch_number)
+        .await?
+        .unwrap()
+        .1;
+    let storage_logs_chunks_count = conn
```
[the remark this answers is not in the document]
tomg10: I totally agree, I'm not sure what I did here :D, corrected

core/bin/snapshots_creator/README.md
```diff
@@ -0,0 +1,24 @@
...
+First run `zk env dev` \
+then the creator can be run using:  
+`zk run snapshots_creator`
+
+Snapshot contents can be stored based on blob_store config either in local filesystem or GS.
+
+## Snapshots format
+
+Each snapshot consists of three types of objects (see
+[snapshots.rs](https://github.com/matter-labs/zksync-era/core/lib/types/src/snapshots.rs)) : header, storage logs chunks
```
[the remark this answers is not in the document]
tomg10: URL path has also some nice features and is consistent with rest of our repo, are 'relative' paths clickable on github (they're not in my IDE)?

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,273 @@
...
+use zksync_utils::time::seconds_since_epoch;
+
+#[derive(Debug, Metrics)]
+#[metrics(prefix = "snapshots_creator")]
+struct SnapshotsCreatorMetrics {
+    storage_logs_chunks_count: Gauge<u64>,
+
+    storage_logs_chunks_left_to_process: Gauge<u64>,
+
+    snapshot_generation_duration: Gauge<u64>,
```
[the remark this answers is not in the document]
tomg10: It's very interesting to see completely different mapping of metrics to prometheus 'primitives' than what I'm used to. Cronjobs are very rarely run more often than every hour so for me using a histogram feels 'hacky' here. In such scenarios there is only one job run at the same time and we want only a single number to track. That being said, as those two primitives have comparable features, I totally buy the consistency argument (also because we don't and probably won't have many cronjobs)

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,273 @@
...
+
+    snapshot_l1_batch: Gauge<u64>,
+
+    #[metrics(buckets = Buckets::LATENCIES)]
+    storage_logs_processing_durations: Histogram<Duration>,
```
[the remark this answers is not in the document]
tomg10: corrected

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,273 @@
...
+struct SnapshotsCreatorMetrics {
+    storage_logs_chunks_count: Gauge<u64>,
+
+    storage_logs_chunks_left_to_process: Gauge<u64>,
+
+    snapshot_generation_duration: Gauge<u64>,
+
+    snapshot_l1_batch: Gauge<u64>,
+
+    #[metrics(buckets = Buckets::LATENCIES)]
```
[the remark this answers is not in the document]
tomg10: corrected

core/lib/zksync_core/src/api_server/web3/namespaces/snapshots.rs
```diff
@@ -0,0 +1,89 @@
...
+        }
+    }
+}
+impl<G: L1GasPriceProvider> SnapshotsNamespace<G> {
+    pub fn new(state: RpcState<G>) -> Self {
+        Self { state }
+    }
+    pub async fn get_all_snapshots_impl(&self) -> Result<AllSnapshots, Web3Error> {
+        let method_name = "get_all_snapshots";
+        let mut storage_processor = self
```
slowli, line 26: Nit: We probably want to measure method latency as with other Web3 methods using `API_METRICS`.

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,306 @@
...
+        .get_snapshot_metadata(l1_batch_number)
+        .await?
+        .is_some()
+    {
+        tracing::info!("Snapshot for L1 batch number {l1_batch_number} already exists, exiting",);
+        return Ok(());
+    }
+    drop(master_conn);
+
+    // snapshots always
```
slowli, line 163: Unfinished comment? I sometimes do this as well 🙃

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,306 @@
...
+
+use crate::chunking::get_chunk_hashed_keys_range;
+use zksync_dal::ConnectionPool;
+use zksync_env_config::object_store::SnapshotsObjectStoreConfig;
+use zksync_env_config::FromEnv;
+use zksync_object_store::{ObjectStore, ObjectStoreFactory};
+use zksync_types::snapshots::{
+    SnapshotFactoryDependencies, SnapshotStorageLogsChunk, SnapshotStorageLogsStorageKey,
+};
+use zksync_types::zkevm_test_harness::zk_evm::zkevm_opcode_defs::decoding::AllowedPcOrImm;
```
slowli, line 26: ??? That's a weird way to convert `u32` to `u64` 🤔 You should use `Into::into()` instead.

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,306 @@
...
+use tokio::sync::watch::Receiver;
+use vise::Unit;
+use vise::{Buckets, Gauge, Histogram, Metrics};
+use zksync_config::configs::PrometheusConfig;
+use zksync_config::{PostgresConfig, SnapshotsCreatorConfig};
+
+use crate::chunking::get_chunk_hashed_keys_range;
+use zksync_dal::ConnectionPool;
+use zksync_env_config::object_store::SnapshotsObjectStoreConfig;
+use zksync_env_config::FromEnv;
```
slowli, line 21: Bikeshedding: I understand that we don't enforce or even have a consistent import style, but IMO we should *at least* ensure at most one import entry for each crate. Not blocking obviously, so feel free to ignore.

core/lib/object_store/src/gcs.rs
```diff
@@ -205,6 +205,14 @@ impl ObjectStore for GoogleCloudStorage {
     async fn remove_raw(&self, bucket: Bucket, key: &str) -> Result<(), ObjectStoreError> {
         self.remove_inner(bucket.as_str(), key).await
     }
+
+    fn get_storage_prefix_raw(&self, bucket: Bucket) -> String {
```
slowli, line 209: Bikeshedding: `get_` prefix [isn't usually](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter) used for getters in Rust (at least, in the standard library); so this method can be called `storage_prefix_raw`. (`get_raw` is different IMO, because it performs a non-trivial action, and there would be nothing left if "get" is removed.)

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,306 @@
...
+    let distinct_storage_logs_keys_count = conn
+        .snapshots_creator_dal()
+        .get_distinct_storage_logs_keys_count(l1_batch_number)
+        .await?;
+
+    drop(conn);
+
+    let chunk_size = config.storage_logs_chunk_size;
+    // we force at least 10 chunks to avoid situations where only one chunk is created in tests
+    let chunks_count = max(10, ceil_div(distinct_storage_logs_keys_count, chunk_size));
```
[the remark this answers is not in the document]
tomg10: It does not negatively influence the creator. If there are 10 small chunks, even some of them empty that's okay.

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,306 @@
...
+        output_filepath
+    );
+    Ok(output_filepath)
+}
+
+async fn run(
+    blob_store: Box<dyn ObjectStore>,
+    replica_pool: ConnectionPool,
+    master_pool: ConnectionPool,
+) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
tomg10: I don't think it's needed even with fault tolerance

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,306 @@
...
+        .create_store()
+        .await;
+
+    let postgres_config = PostgresConfig::from_env().context("PostgresConfig")?;
+    let creator_config =
+        SnapshotsCreatorConfig::from_env().context("SnapshotsCreatorConfig::from_env")?;
+
+    let replica_pool = ConnectionPool::builder(
+        postgres_config.replica_url()?,
+        creator_config.concurrent_queries_count,
```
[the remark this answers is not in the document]
tomg10: IMO it's the config is nicer if it's normalized and I really want to have an explicit snapshot config parameter for this

core/bin/snapshots_creator/src/main.rs
tomg10: yeaaah, I'm not sure what I did here, I think my IDE suggested it and I did not see where it comes from :D

core/bin/snapshots_creator/src/chunking.rs
```diff
@@ -0,0 +1,21 @@
+use std::cmp::min;
+
+pub fn get_chunk_hashed_keys_range(chunk_id: u64, chunks_count: u64) -> ([u8; 2], [u8; 2]) {
```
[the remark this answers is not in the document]
tomg10: I wanted to make the bucket sizes as similar as possible, for your code, the last bucket will be smaller than others, which is undesired for instance for metrics

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,306 @@
...
+    };
+    let filename = blob_store
+        .put(key, &storage_logs_chunk)
+        .await
+        .context("Error storing storage logs chunk in blob store")?;
+
+    let output_filepath_prefix = blob_store.get_storage_prefix::<SnapshotStorageLogsChunk>();
+    let output_filepath = format!("{output_filepath_prefix}/{filename}");
+
+    let elapsed_ms = latency.observe().as_millis();
```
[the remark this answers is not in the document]
tomg10: corrected

core/bin/snapshots_creator/src/main.rs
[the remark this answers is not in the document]
RomanBrodetski: hmm don't agree with this - while it's technically it doesn't impact correctness, it would look strange at least and incorrect at most if we are generating 10 chunks for a very small dataset. And this can very well happen in the real world - not for Era mainnet, but for some of the hyperchains

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,306 @@
...
+    )
+    .await?;
+
+    let mut storage_logs_output_files = vec![];
+
+    METRICS
+        .storage_logs_chunks_left_to_process
+        .set(chunks_count);
+    let mut tasks =
+        FuturesUnordered::<Pin<Box<dyn Future<Output = anyhow::Result<String>>>>>::new();
```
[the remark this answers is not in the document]
RomanBrodetski: hmm interesting - I don't have enough rust experience to reason about what's more idiomatic, but from a quick glance, the current solution seems way too imperative, I would think it's possible to use a higher-level abstraction. Again, from a quick glance `for_each_concurrent` seems perfect, but a semaphore solution is also simpler than this

For me it's a little bit like using a while loop and manipulating some counter directly instead of a higher-order function like `map`.

> which I can monitor progress and from which I can pull finished tasks, that's why I also don't want to use join_all as I want to know exactly when tasks finish.

Not sure how exactly this applies here - can you please elaborate?

core/bin/snapshots_creator/src/main.rs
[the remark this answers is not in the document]
RomanBrodetski: yeah I agree that a natural value would be the `max_connections`, but I'd also prefer to have a separate value here.

RomanBrodetski (commented):
Great work!

core/bin/snapshots_creator/src/main.rs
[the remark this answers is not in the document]
tomg10: I agree that's certainly some sort of a tradeoff, but so far it's the best way I see of making sure that we really test generating at least 2 chunks (we don't know, how many storage logs are at the moment the integration test is run)

core/lib/object_store/src/gcs.rs
tomg10: good to know, corrected

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,306 @@
...
+
+    let output_filepath_prefix = blob_store.get_storage_prefix::<SnapshotStorageLogsChunk>();
+    let output_filepath = format!("{output_filepath_prefix}/{filename}");
+
+    let elapsed_ms = latency.observe().as_millis();
+    tracing::info!(
+        "Finished storage logs chunk {}/{chunks_count}, step took {elapsed_ms}ms, output stored in {output_filepath}",
+        chunk_id + 1
+    );
+    drop(conn);
```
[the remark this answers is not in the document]
tomg10: corrected

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,306 @@
...
+use anyhow::Context as _;
+use futures::stream::FuturesUnordered;
+use futures::StreamExt;
+use prometheus_exporter::PrometheusExporterConfig;
+use std::cmp::max;
+use std::future::Future;
+use std::pin::Pin;
+use std::time::Duration;
+use tokio::sync::watch;
+use tokio::sync::watch::Receiver;
```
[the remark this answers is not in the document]
tomg10: as discussed, we will fix such issues automatically

core/lib/dal/src/snapshots_creator_dal.rs
```diff
@@ -0,0 +1,100 @@
...
+                AND storage_logs.miniblock_number = keys.op[1]
+                AND storage_logs.operation_number = keys.op[2]
+                     INNER JOIN initial_writes ON keys.hashed_key = initial_writes.hashed_key;
+             "#,
+            miniblock_number.0 as i64,
+            min_hashed_key,
+            max_hashed_key,
+        )
+        .instrument("get_storage_logs_chunk")
+        .report_latency()
```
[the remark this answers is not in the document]
tomg10: corrected

core/lib/zksync_core/src/api_server/web3/namespaces/snapshots.rs
tomg10: fixed

core/lib/dal/src/snapshots_creator_dal.rs
```diff
@@ -0,0 +1,100 @@
...
+        .fetch_all(self.storage.conn())
+        .await?
+        .iter()
+        .map(|row| SnapshotStorageLog {
+            key: StorageKey::new(
+                AccountTreeId::new(Address::from_slice(row.address.as_ref().unwrap())),
```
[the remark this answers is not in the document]
tomg10: I'm not sure whether the the remedy is not worse than the disease here, I would argue that we would rather optimize for query readability than deserialization readability (but I see that we use it everywhere, so I fixed it)

core/bin/snapshots_creator/src/main.rs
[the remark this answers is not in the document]
tomg10: I meant that I do not see any nice functional version of this code where we can track progress, but I also did not take `map` with side effects as a consideration. I usually try to avoid using it with functions that are not 'pure'. Also, the part that uses the counter is not thread-safe, so we would need a second semaphore (or rather a mutex). TLDR; I think the devil is in the details and that imperative version is the lesser evil, but would very like to be proven that this is not the case ;). @slowli can you show me a snippet or even pseudocode how you imagine the code using semaphore(s) would be structured?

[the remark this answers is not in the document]
slowli: The core code is as follows:

```rust
let semaphore = Semaphore::new(config.concurrent_queries_count as usize);
let tasks = (0..chunks_count).map(|chunk_id| {
    process_storage_logs_single_chunk(
        &semaphore,
        &*blob_store,
        &replica_pool,
        last_miniblock_number_in_batch,
        l1_batch_number,
        chunk_id,
        chunks_count,
    )
});
let storage_logs_output_files = futures::future::try_join_all(tasks).await?;
```

...where `process_storage_logs_single_chunk` is modified to acquire a `semaphore` permit:

```rust
async fn process_storage_logs_single_chunk(
    semaphore: &Semaphore,
    // other args...
) -> anyhow::Result<String> {
    let _permit = semaphore.acquire().await?;
    // ...
}
```

> Also, the part that uses the counter is not thread-safe

Why would we need a counter in the first place? Theoretically, we could use an atomic, but if it's only required for metrics and logging, then you should use [`Gauge::dec_by()`](https://matter-labs.github.io/vise/vise/struct.Gauge.html#method.dec_by) (which uses an atomic internally).

A variant with `Stream`s:

```rust
let tasks = futures::stream::iter(0..chunks_count).map(|chunk_id| {
    process_storage_logs_single_chunk(
        &*blob_store,
        &replica_pool,
        last_miniblock_number_in_batch,
        l1_batch_number,
        chunk_id,
        chunks_count,
    )
});
let storage_logs_output_files: Vec<_> = tasks
    .buffer_unordered(config.concurrent_queries_count as usize)
    .try_collect()
    .await?;
```

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,304 @@
...
+            last_chunk_id += 1;
+        }
+        if let Some(result) = tasks.next().await {
+            tracing::info!(
+                "Completed chunk {}/{}, {} chunks are still in progress",
+                last_chunk_id - tasks.len() as u64,
+                chunks_count,
+                tasks.len()
+            );
+            storage_logs_output_files.push(result.context("Chunk task failed")?);
```
slowli, line 223: Just to clarify: Is it intended that `storage_logs_output_files` can be ordered in any way (i.e., not by `chunk_id`)?

core/bin/snapshots_creator/src/chunking.rs
```diff
@@ -0,0 +1,29 @@
...
+    let mut start_bytes = (chunk_start as u16).to_be_bytes().to_vec();
+    let mut end_bytes = (chunk_end as u16).to_be_bytes().to_vec();
+
+    start_bytes.resize(32, 0);
+    end_bytes.resize(32, 0);
+
+    std::ops::Range {
+        start: H256::from_slice(&start_bytes),
+        end: H256::from_slice(&end_bytes),
+    }
```
slowli, line 28: Ranges are idiomatically constructed as `start..end`; i.e.

```rust
H256::from_slice(&start_bytes)..H256::from_slice(&end_bytes)
```

core/lib/zksync_core/src/api_server/web3/namespaces/snapshots.rs
```diff
@@ -0,0 +1,96 @@
...
+use crate::api_server::web3::metrics::API_METRICS;
+use crate::api_server::web3::state::RpcState;
+use crate::l1_gas_price::L1GasPriceProvider;
+use zksync_types::snapshots::{AllSnapshots, SnapshotHeader, SnapshotStorageLogsChunkMetadata};
+use zksync_types::L1BatchNumber;
+use zksync_web3_decl::error::Web3Error;
+
+#[derive(Debug)]
+pub struct SnapshotsNamespace<G> {
+    pub state: RpcState<G>,
```
slowli, line 11: Nit: This field may be private.

core/lib/zksync_core/src/api_server/web3/namespaces/snapshots.rs
```diff
@@ -0,0 +1,96 @@
...
+        Self {
+            state: self.state.clone(),
+        }
+    }
+}
+impl<G: L1GasPriceProvider> SnapshotsNamespace<G> {
+    pub fn new(state: RpcState<G>) -> Self {
+        Self { state }
+    }
+    pub async fn get_all_snapshots_impl(&self) -> Result<AllSnapshots, Web3Error> {
```
slowli, line 25: Nit: There should be newlines between methods / functions for readability. `rustfmt` doesn't seem to have a granular enough control over this, unfortunately (there's [`blank_lines_lower_bound`](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#blank_lines_lower_bound), but it applies to *all* items, which is an overkill).

core/tests/ts-integration/tests/api/snapshots-creator.test.ts
```diff
@@ -0,0 +1,84 @@
+import { TestMaster } from '../../src/index';
+import fs from 'fs';
+import * as zlib from 'zlib';
+import { snapshots_creator } from 'zk/build/run/run';
+describe('Snapshots API tests', () => {
```
slowli, line 5: Nit: A blank line between `import`s and main code would improve readability.

core/tests/ts-integration/tests/api/snapshots-creator.test.ts
```diff
@@ -0,0 +1,84 @@
...
+
+            readStream.pipe(gunzip);
+        });
+    }
+    async function createAndValidateSnapshot() {
+        let existingBatchNumbers = (await getAllSnapshots()).snapshotsL1BatchNumbers as number[];
+        await runCreator();
+        let newBatchNumbers = (await getAllSnapshots()).snapshotsL1BatchNumbers as number[];
+        let addedSnapshots = newBatchNumbers.filter((x) => existingBatchNumbers.indexOf(x) === -1);
+        expect(addedSnapshots.length).toEqual(1);
```
slowli, line 53: Dumb question: Does this assume that there's some activity regarding the node (e.g., other tests) run since the last snapshot was created? Otherwise, the node may be at the same snapshot as previously, which would lead to this check failing.

core/tests/ts-integration/tests/api/snapshots-creator.test.ts
```diff
@@ -0,0 +1,84 @@
...
+        let fullSnapshot = await getSnapshot(l1BatchNumber);
+        let miniblockNumber = fullSnapshot.miniblockNumber;
+
+        expect(fullSnapshot.l1BatchNumber).toEqual(l1BatchNumber);
+        for (let chunkMetadata of fullSnapshot.storageLogsChunks) {
+            console.log(`Verifying ${chunkMetadata.filepath}`);
+            let path = `${process.env.ZKSYNC_HOME}/${chunkMetadata.filepath}`;
+
+            let output = JSON.parse(await decompressGzip(path));
+            expect(output['storageLogs'].length > 0);
```
slowli, line 65: Dumb question: Doesn't this check contradict the fact the discussion about minimum chunk amount, in which you've said that some chunks may end up empty?

core/tests/ts-integration/tests/api/snapshots-creator.test.ts
```diff
@@ -0,0 +1,84 @@
...
+
+            gunzip.on('data', (chunk) => (data += chunk.toString()));
+            gunzip.on('end', () => resolve(data));
+            gunzip.on('error', reject);
+
+            readStream.pipe(gunzip);
+        });
+    }
+    async function createAndValidateSnapshot() {
+        let existingBatchNumbers = (await getAllSnapshots()).snapshotsL1BatchNumbers as number[];
```
slowli, line 49: Nit: Most of these `let`s could be replaced with `const`s since JS / TS const-ness is shallow, rather than structural as in Rust; if a variable itself is not reassigned, it can be a `const`; and at least some code styles for JS / TS consider `const`s preferable. This is obviously not a huge issue, just FYI.

slowli (commented):
Regarding merging: In my mind, any nits / off-topic comments do not block a merge; they can be dealt with in a follow-up PR (even by the reviewer if they so desire). There's a couple of "substantive" comments remaining in this case, though.

core/bin/snapshots_creator/src/chunking.rs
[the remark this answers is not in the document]
tomg10: And it makes sense to use your logic

core/tests/ts-integration/tests/api/snapshots-creator.test.ts
tomg10: Yes, this test implicitly expects that or that it will be run exactly once since server start. I plan to work on this in the future (a solution is to start a fresh zk sync node for such test). In general, unfortunately there are a few such caveats built-in into ts-integration tests

core/bin/snapshots_creator/src/main.rs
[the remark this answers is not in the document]
tomg10: Yeah, the 'root' issue is the non-reproductible test setup. I've been also thinking about chunk number as a configuration parameter, but I think the lesser evil is just to have that hardcoded "10" for now

core/bin/snapshots_creator/src/main.rs
tomg10: definitely not! nice catch

core/lib/zksync_core/src/api_server/web3/namespaces/snapshots.rs
tomg10: fixed

core/tests/ts-integration/tests/api/snapshots-creator.test.ts
tomg10: fixed

core/tests/ts-integration/tests/api/snapshots-creator.test.ts
tomg10: this test implicitly depends on the fact that there are some storage logs (created during context setup), This assert it's a bit different, if it fails, that means that either one of those assumptions about the setup is no longer valid (and the test should be corrected as it may no longer check anything and fail silently) or that there is something very wrong with snapshots generation

core/tests/ts-integration/tests/api/snapshots-creator.test.ts
tomg10: I got too used to IDE nudging me about such issues (I must either get WebStorm again or actively start checking such issues); corrected

core/lib/zksync_core/src/api_server/web3/namespaces/snapshots.rs
tomg10: fixed

core/bin/snapshots_creator/src/main.rs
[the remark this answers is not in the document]
tomg10: corrected

core/bin/snapshots_creator/src/main.rs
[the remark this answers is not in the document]
tomg10: corrected

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,291 @@
+mod chunking;
```
slowli, line 1: Nit: Please place `mod` declarations below imports. AFAIU, there are not rules to enforce `mod` code style, but in the existing code they are mostly placed in a single block below all `use`s, which IMO is reasonable.

core/lib/zksync_core/src/api_server/web3/mod.rs
```diff
@@ -119,6 +122,7 @@ impl Namespace {
         Namespace::Zks,
         Namespace::En,
         Namespace::Pubsub,
+        Namespace::Snapshots,
```
slowli, line 125: Just to understand better: Do we *want* to enable this namespace for every node that doesn't configure namespaces manually (see e.g. [relevant EN config logic](https://github.com/matter-labs/zksync-era/blob/7c992323a31baafb882f6ebb448359e0635269b2/core/bin/external_node/src/config/mod.rs#L338-L342))? IMO, it should be an opt-in; e.g., for ENs exposing this namespace makes little sense. Perhaps, we should rename `NON_DEBUG` into something like `DEFAULT` to clearly signal its purpose. cc @RomanBrodetski

core/lib/zksync_core/src/api_server/web3/namespaces/snapshots.rs
```diff
@@ -0,0 +1,102 @@
...
+                .map_err(|err| internal_error(method_name, err))?
+                .unwrap();
+            let miniblock_number = storage_processor
+                .blocks_dal()
+                .get_miniblock_range_of_l1_batch(l1_batch_number)
+                .await
+                .map_err(|err| internal_error(method_name, err))?
+                .unwrap()
+                .1;
+            method_latency.observe();
```
slowli, line 89: Nit: You can avoid duplicating "finalization" code by extracting `Option<SnapshotHeader>` into a variable:

```rust
let snapshot_header = if let Some(snapshot_metadata) = snapshot_metadata {
    // ...
   Some(SnapshotHeader { /* ... */ })
} else {
    None
};
method_latency.observe();
Ok(snapshot_header)
```

core/tests/ts-integration/tests/api/snapshots-creator.test.ts
slowli: Aren't `output['storageLogs']` logs for a specific chunk? If so, it could be the case that we do have storage logs (e.g., ones in from the genesis miniblock), but some chunks don't have storage logs (IIRC, there's about 20 logs in the genesis miniblock). This isn't a blocking issue, just want to find out whether I understand test logic correctly.

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,291 @@
...
+    let latency = METRICS.snapshot_generation_duration.start();
+
+    let config = SnapshotsCreatorConfig::from_env().context("SnapshotsCreatorConfig::from_env")?;
+
+    let mut conn = replica_pool
+        .access_storage_tagged("snapshots_creator")
+        .await?;
+
+    // we subtract 1 so that after restore, EN node has at least one l1 batch to fetch
+    let l1_batch_number = conn.blocks_dal().get_sealed_l1_batch_number().await? - 1;
```
slowli, line 147: This may theoretically panic if there's only the genesis L1 batch in the storage (e.g., in tests).

core/bin/snapshots_creator/src/main.rs
```diff
@@ -0,0 +1,291 @@
...
+    let mut master_conn = master_pool
+        .access_storage_tagged("snapshots_creator")
+        .await?;
+    if master_conn
+        .snapshots_dal()
+        .get_snapshot_metadata(l1_batch_number)
+        .await?
+        .is_some()
+    {
+        tracing::info!("Snapshot for L1 batch number {l1_batch_number} already exists, exiting",);
```
slowli, line 158: Nit: Unnecessary comma at the end of the macro.

slowli (approved):
Awesome job! Keep in mind that in order to deploy these changes, we would need to migrate DBs on the corresponding envs – IIUC, in the current setup this would be semi-necessary because of the enabled-by-default `snapshots` Web3 namespace the methods in which access new tables. (Thus, I'd suggest to switch it off by default; see the relevant line comment.)

core/lib/zksync_core/src/api_server/web3/mod.rs
RomanBrodetski: hmmm actually we might want to expose it - and jsut respond that we don't have any snapshots generated. In the future nothing prevents EN operators to generate snapshots as well

slowli (dismissed):
Looks mergeable; feel free to close the remaining conversations if you don't plan to fix them in this PR. Wouldn't recommend merging on the Friday evening, just in case 🙃 (We have automated deployments to staging, so if something goes awfully wrong, the merge could break this env.)

.githooks/pre-commit
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
yorik, line 1: Any reason for that?
I think it would be better to use default shell in the system, unless we need some bashisms.

artmakh (dismissed):

.githooks/pre-commit
tomg10: when I started developing this PR it had one, but I see it was already corrected, I'll revert this to use sh

slowli (approved):

PR #499, opened 2023-11-16
refactor(eth_client): Use `BlockId` for `block_id` instead of String in `eth_client`

RedaOps:
## What ❔

The `block` function within `eth_client` now uses the type `BlockId` instead of `String` for its `block_id` parameter.

## Why ❔

TODO comment:

https://github.com/matter-labs/zksync-era/blob/cb873bd0da6b421160ce96b8d578f1351861f376/core/lib/eth_client/src/clients/http/query.rs#L289-L308

This PR fixes the TODO.

The `web3` crate being used inside the project was updated to version `0.19.0` inside `core/lib/basic_types/Cargo.toml` in commit 829ef5085f938ddda1f2a695930c6b4308e1643a.

The `web3` crate now supports `BlockNumber::Finalized`.
Source: https://docs.rs/web3/latest/web3/types/enum.BlockNumber.html
Source: https://github.com/tomusdrw/rust-web3/releases/tag/v0.19.0

## Checklist


popzxc (approved):
Thanks!

PR #504, opened 2023-11-16
feat(core): adds a get proof endpoint in zks namespace to http endpoint on main node

montekki:
## What ❔

Support for `zks_getProof` was accedentally only implemented for WS apis in #455, this PR deploys them also to HTTP endpoints.


core/lib/web3_decl/src/namespaces/zks.rs
```diff
@@ -110,4 +110,12 @@ pub trait ZksNamespace {
 
     #[method(name = "getLogsWithVirtualBlocks")]
     async fn get_logs_with_virtual_blocks(&self, filter: Filter) -> RpcResult<Vec<Log>>;
+
+    #[method(name = "zks_getProof")]
```
slowli, line 114: Shouldn't the name be simply `getProof`? Other method names don't have the `zks_` prefix.

slowli (commented):
Akchually 🤓, the PR title isn't technically true: [we use](https://github.com/matter-labs/zksync-era/blob/87bd22693ec59b9346b92e96537e46c49515b378/core/bin/external_node/src/main.rs#L239) the `jsonrpc` backend both for HTTP and WS API servers for the external node; `jsonrpsee` is only used for HTTP on the main node.

slowli (approved):

PR #510, opened 2023-11-17
feat: Add a setup Makefile target to install all dependencies

jrchatruc:
## What ❔

Running `make deps` now installs all the dependencies required to run the stack. For now supports both `macOS` and `ubuntu`/`debian`.

## Checklist

### DIFF
--- a/install_deps.sh
+++ b/install_deps.sh
@@ -1,0 +1,59 @@
+#!/usr/bin/env bash
+
+set -e
+
+ZKSYNC_HOME="$(pwd)"
+OS=$(uname -s)
+
+export ZKSYNC_HOME=$ZKSYNC_HOME
+export PATH="$ZKSYNC_HOME/bin:$PATH"
+
+if [ "$OS" = "Darwin" ]
+then
+	brew install axel openssl postgresql tmux node@18
+	brew link node@18 --overwrite
+	curl -L https://github.com/matter-labs/zksolc-bin/releases/download/v1.3.16/zksolc-macosx-arm64-v1.3.16 --output zksolc
+	chmod a+x zksolc
+	curl -L https://github.com/ethereum/solidity/releases/download/v0.8.19/solc-macos --output solc
+	chmod a+x solc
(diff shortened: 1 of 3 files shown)


install_deps.sh
```diff
@@ -0,0 +1,53 @@
...
+ZKSYNC_HOME="$(pwd)"
+OS=$(uname -s)
+
+export ZKSYNC_HOME=$ZKSYNC_HOME
+export PATH="$ZKSYNC_HOME/bin:$PATH"
+
+if [ "$OS" = "Darwin" ]
+then
+	brew install axel openssl postgresql tmux node@18
+	brew link node@18 --overwrite
```
popzxc, line 14: Will it work fine if somebody, for example, already uses `nvm`?

install_deps.sh
```diff
@@ -0,0 +1,53 @@
...
+	curl -fsSL https://get.docker.com | sh
+	curl -SL https://github.com/docker/compose/releases/download/v2.23.0/docker-compose-linux-x86_64 -o /usr/local/bin/docker-compose
+	curl -L https://github.com/matter-labs/zksolc-bin/releases/download/v1.3.16/zksolc-linux-amd64-musl-v1.3.16 --output zksolc
+	curl -L https://github.com/ethereum/solidity/releases/download/v0.8.19/solc-static-linux --output solc
+	chmod a+x solc
+	chmod a+x /usr/local/bin/docker-compose
+	chmod a+x zksolc
+	mkdir -p $(HOME)/.config
+	mv solc $(HOME)/.config
+	mv zksolc $(HOME)/.config
```
popzxc, line 35: I would say that `~/.config` isn't really a directory where I would expect binaries to reside.

popzxc (commented):
Honestly, I'm not sure if we want to have this, at least in the current form.
The script seems to be, well, opinionated in how it installs stuff and where it puts it.
The consequences of running the script may be not really predictable for the person who will run it, resulting in some issues that will not be easy to debug.

perekopskiy:
Closing as stale

PR #511, opened 2023-11-17
feat: explicit toolchain in clippy (zk lint rust) execution

tomg10:
## What ❔

Provide explicit toolchain version when executing clippy so that linting locally has the same effect as when run on CI

## Why ❔

It's too easy now to use `zk lint rust `so that it outputs errors that are different than those in CI (especially that the setup-dev.md does not ask to install specific rust version, and there are multiple versions of rust used in different dockerfiles in our repo), this change makes this command more stable

## Checklist

### DIFF
--- a/infrastructure/zk/src/lint.ts
+++ b/infrastructure/zk/src/lint.ts
@@ -39,7 +39,7 @@ async function lintSystemContracts(check: boolean = false) {
 function clippyRustToolchain(): string {
     //There are no specific reasons to run this particular version other than because clippy is not very stable
     // and even the same clippy sem-ver version can result in different output based on compiler version
-    return 'nightly-2023-08-21';
+    return '1.74.0';
 }
 async function clippy() {
     process.chdir(process.env.ZKSYNC_HOME!);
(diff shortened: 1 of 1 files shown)


infrastructure/zk/src/lint.ts
```diff
@@ -36,14 +36,19 @@ async function lintSystemContracts(check: boolean = false) {
     await utils.spawn(`yarn --silent --cwd etc/system-contracts lint:${check ? 'check' : 'fix'}`);
 }
 
+function clippyRustToolchain(): string {
+    //There are no specific reasons to run this particular version other than because clippy is not very stable
+    // and even the same clippy sem-ver version can result in different output based on compiler version
+    return 'nightly-2023-08-21';
+}
```
montekki, line 43: Fixing this here may break a lot of stuff because our nightly builds are going to be built with the actual nightly compilers each new day (or at least that is the idea). I would advice just not to override rust in your builds and default to what the repo specifies.

Deniallugo:
@tomg10 I've updated nightly rust https://github.com/matter-labs/zksync-era/pull/530 we can close it now

PR #525, opened 2023-11-20
test(api): Test coverage for WebSocket server

slowli:
## What ❔

Cover basic WS-specific JSON-RPC functionality (subscriptions, `getFilterChanges`) with tests.

## Why ❔

Will allow to set the expectations the current server behavior, so that it could be evolved in the future w/o unexpected / breaking changes (e.g., when implementing `jsonrpsee`-based subscriptions).

## Checklist


slowli:
@montekki I haven't managed to find a way to reproduce duplicating or dropping event notifications so far. The notification logic, while not always straightforward, seems correct. If you have an idea how to reproduce the bug, please share.

core/lib/zksync_core/src/api_server/web3/tests/ws.rs
```diff
@@ -0,0 +1,694 @@
...
+        drop(storage);
+        let new_events: Vec<_> = new_events.iter().collect();
+
+        let lower_bound_logs = client.get_filter_changes(lower_bound_filter_id).await?;
+        let FilterChanges::Logs(lower_bound_logs) = lower_bound_logs else {
+            panic!("Unexpected getFilterChanges output: {:?}", lower_bound_logs);
+        };
+        assert_logs_match(&lower_bound_logs, &new_events);
+
+        // FIXME: is this expected?
```
montekki, line 652: A fixme should either be figured out before merging, or probably a new tracking issue has to be created and referenced here.

core/lib/zksync_core/src/api_server/web3/tests/ws.rs
```diff
@@ -0,0 +1,694 @@
...
+        let FilterChanges::Logs(topics_logs) = topics_logs else {
+            panic!("Unexpected getFilterChanges output: {:?}", topics_logs);
+        };
+        assert_logs_match(&topics_logs, &[events[1], events[3]]);
+
+        let new_all_logs = client.get_filter_changes(all_logs_filter_id).await?;
+        let FilterChanges::Logs(new_all_logs) = new_all_logs else {
+            panic!("Unexpected getFilterChanges output: {:?}", new_all_logs);
+        };
+        assert_eq!(new_all_logs, all_logs); // FIXME: is this expected?
```
montekki, line 583: Same here

core/lib/zksync_core/src/api_server/web3/tests/ws.rs
```diff
@@ -0,0 +1,694 @@
...
+#[async_trait]
+impl WsTest for LogSubscriptionsWithManyBlocks {
+    async fn test(&self, client: &WsClient, pool: &ConnectionPool) -> anyhow::Result<()> {
+        let Subscriptions {
+            mut all_logs_subscription,
+            mut address_subscription,
+            ..
+        } = Subscriptions::new(client).await?;
+        // Wait a little until subscriptions are fully registered.
+        tokio::time::sleep(POLL_INTERVAL).await;
```
montekki, line 405: Sometimes in this code "Wait a little" waits for POLL_INTERVAL and sometimes double that; judging by the logic in `notify_*` are you absolutely sure that at least the former will not cause flakiness in the CI?

core/lib/zksync_core/src/api_server/web3/tests/ws.rs
slowli: Yep, it should be discussed whether this `getFilterChanges` behavior is expected (IMO, it's not, but I don't have experience with `getFilterChanges` in Ethereum), and creating a task to fix it if it's not.

core/lib/zksync_core/src/api_server/web3/tests/ws.rs
slowli: I've removed timeouts in tests in favor of listening to events emitted by subscription logic. It required significant refactoring of that logic, but IMO, it improved it a bit 🙃

core/lib/zksync_core/src/api_server/web3/tests/ws.rs
montekki: > Polling method for a filter, which returns an array of logs which occurred since last poll.

Yeah we should create a task

montekki: I think that the bug was introduced in #526 

https://github.com/matter-labs/zksync-era/blob/925da43c3b85b4818cb910f286b0f69ae118d692/core/lib/zksync_core/src/api_server/web3/namespaces/eth.rs#L862-L868

RomanBrodetski (commented):
Was only able to have a quick look and it looks good. @slowli please let me know if you want me to have a close look at any particular part!

core/lib/zksync_core/src/api_server/web3/tests/ws.rs
slowli: This behavior most certainly wasn't introduced in #526. The initial version of this PR with the tests predates #526 being merged; #526 just reproduces the behavior that already existed (just slightly refactors it). This is what the FIXME is about – investigating whether the existing behavior is correct, and fixing it if it's not.

slowli:
@RomanBrodetski Please take a look at FIXMEs in tests. They are related to the `getFilterChanges` implementation, which (as we've briefly discussed elsewhere) may return the same updates multiple times. (Since the method it uses is modifying the filter cursor, and the HTTP / WS transport cannot be assumed to be reliable, we have neither "at most once" nor "at least once" semantics for update delivery.)

montekki:
@slowli Why again are `getFilterChanges` tests in the websocket testing? I believe that these methods are not websocket-specific and should instead go into main http endpoint testing. Only `eth_subscribe` is websocket-specific. If you agree I propose to move these tests into #546 and try to trigger all the corner cases there?

slowli:
@montekki 

> I believe that these methods are not websocket-specific and should instead go into main http endpoint testing

They are indeed not WS-specific, although in practice it's impossible to use `getFilterChanges` HTTP endpoint for our testnet / mainnet deployments because of K8s load balancing (subsequent queries may arrive at different pods that don't have the referenced filter installed).

As for your proposal: Could it make sense to move tests within this PR (just make them test the HTTP server)? If this PR is merged before #546, then *that* PR could update the tests to reflect corrected server behavior.

montekki:
> As for your proposal: Could it make sense to move tests within this PR (just make them test the HTTP server)? If this PR is merged before https://github.com/matter-labs/zksync-era/pull/546, then that PR could update the tests to reflect corrected server behavior.

Yeah good we could do two things I think
1. either just move the abovementioned fix PR in here and expand the tests a bit
2. or merge this one with referencing the fix PR in the `FIXME`s

Whatever you prefer

montekki (approved):

montekki (approved):

PR #526, opened 2023-11-20
chore: Add metrics on types of filters in API

Artemka374:
## What ❔

Add metrics on types of filters, especially:
* Count of the current amount of filters of specific type
* Frequency of requests to specific filter type
* Lifetime of filters of specific type
* Request count for a specific filter

## Why ❔

We receive a lot of requests in our WebSocket API to getFilterChanges method.

We want to understand types of filters and distribution of load among them.

## Checklist


core/lib/zksync_core/src/api_server/web3/metrics.rs
```diff
@@ -195,3 +195,61 @@ pub(super) struct PubSubMetrics {
...
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, EncodeLabelValue, EncodeLabelSet)]
+#[metrics(label = "filter_type", rename_all = "snake_case")]
+pub(super) enum FilterType {
+    Events,
+    Blocks,
+    PendingTransactions,
+}
+
+impl From<TypedFilter> for FilterType {
```
slowli, line 207: It make sense to implement `From<&TypedFilter>` instead; no data from the filter is required to construct `FilterType`.

core/lib/zksync_core/src/api_server/web3/metrics.rs
```diff
@@ -195,3 +195,61 @@ pub(super) struct PubSubMetrics {
...
+        match value {
+            TypedFilter::Events(_, _) => FilterType::Events,
+            TypedFilter::Blocks(_) => FilterType::Blocks,
+            TypedFilter::PendingTransactions(_) => FilterType::PendingTransactions,
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct InstalledFilter {
```
slowli, line 218: Bikeshedding: It doesn't look right to me to place this type in the `metrics` module. While all its additional fields *for now* are related to metrics, it isn't necessarily the case.

core/lib/zksync_core/src/api_server/web3/metrics.rs
```diff
@@ -195,3 +195,61 @@ pub(super) struct PubSubMetrics {
...
+    pub filter: TypedFilter,
+    guard: Box<GaugeGuard>,
+    start: Instant,
```
slowli, line 221: Bikeshedding: Please use more informative field names, such as `created_at` in this case.

core/lib/zksync_core/src/api_server/web3/metrics.rs
```diff
@@ -195,3 +195,61 @@ pub(super) struct PubSubMetrics {
...
+    // FIXME: Change the type
+    pub last_request: u64,
+}
+
+impl InstalledFilter {
+    pub fn new(filter: TypedFilter, guard: GaugeGuard) -> Self {
```
slowli, line 227: Can't `guard` be constructed inside `new()` to make this method more robust?

core/lib/zksync_core/src/api_server/web3/metrics.rs
```diff
@@ -195,3 +195,61 @@ pub(super) struct PubSubMetrics {
...
+    fn drop(&mut self) {
+        FILTER_METRICS.filter_lifetime[&FilterType::from(self.filter.clone())]
+            .observe(self.start.elapsed());
+    }
+}
+
+#[derive(Debug, Metrics)]
+#[metrics(prefix = "api_filter")]
+pub(super) struct FilterMetrics {
+    pub metrics_count: Family<FilterType, Gauge>,
```
slowli, line 247: Nit: You can add doc comments for metrics to make their purpose clearer; they are exported to Prometheus and can be displayed in Grafana tooltips etc.

core/lib/zksync_core/src/api_server/web3/metrics.rs
```diff
@@ -195,3 +195,61 @@ pub(super) struct PubSubMetrics {
...
+    #[metrics(buckets = Buckets::LATENCIES)]
```
slowli, line 248: For new metrics, you should specify `unit` if possible (e.g., `unit = Unit::Seconds` here and for `filter_lifetime`). The old metrics don't have `unit` set because this would change their names (by convention, unit is added as a suffix to the metric name, e.g. `api_filter_request_frequency_seconds`), so it would break existing Grafana dashboards.

core/lib/zksync_core/src/api_server/web3/mod.rs
```diff
@@ -63,6 +65,17 @@ use self::state::{Filters, InternalApiConfig, RpcState, SealedMiniblockNumber};
 /// Timeout for graceful shutdown logic within API servers.
 const GRACEFUL_SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(5);
 
+/// Represents all kinds of `Filter`.
+#[derive(Debug, Clone)]
+pub enum TypedFilter {
```
slowli, line 70: Is `pub` visibility required here? In general, it makes sense to restrict visibility to private, then to `pub(super)` / `pub(crate)`, and only if this doesn't work, make something public. Same goes for other types you define in this PR.

core/lib/zksync_core/src/api_server/web3/state.rs
```diff
@@ -578,13 +582,29 @@ impl Filters {
 
     /// Retrieves filter from the state.
     pub fn get(&self, index: U256) -> Option<&TypedFilter> {
-        self.state.get(&index)
+        self.state.get(&index).map(|installed_filter| {
```
slowli, line 585: Nit: You can use `?` to short-circuit this logic for readability:

```rust
let installed_filter = self.state.get(&index)?;
// ...
Some(&installed_filter.filter)
```

Also, it shouldn't compile without using `get_mut()` and changing the signature to something like `fn get_and_update(&mut self, index: U256)`. Alternatively (I'd prefer this option), you can move all mutation logic to `update()`.

core/lib/zksync_core/src/api_server/web3/state.rs
```diff
@@ -578,13 +582,29 @@ impl Filters {
...
 
     /// Updates filter in the state.
     pub fn update(&mut self, index: U256, new_filter: TypedFilter) -> bool {
         if let Some(typed_filter) = self.state.get_mut(&index) {
-            *typed_filter = new_filter;
+            let filter_type = FilterType::from(new_filter.clone());
+            let guard = FILTER_METRICS.metrics_count[&filter_type].inc_guard(1);
+
+            // FIXME: is previous value dropped?
+            *typed_filter = InstalledFilter::new(new_filter, guard);
```
slowli, line 607: You shouldn't drop the previous value it'll lead to incorrect metrics, because semantically the filter is *updated*; it's still in place after the update. Instead, you should just update the `filter` field of the filter.

core/lib/zksync_core/src/api_server/web3/mod.rs
```diff
@@ -7,12 +7,14 @@ use serde::Deserialize;
...
 
+use chrono::NaiveDateTime;
 use std::{net::SocketAddr, sync::Arc, time::Duration};
 use tokio::task::JoinHandle;
 
 use zksync_dal::{ConnectionPool, StorageProcessor};
 use zksync_health_check::{HealthStatus, HealthUpdater, ReactiveHealthCheck};
 use zksync_types::{api, MiniblockNumber};
+use zksync_web3_decl::types::Filter;
 use zksync_web3_decl::{
```
slowli, line 18: Bikeshedding: Although we don't enforce any particular `use` grouping for now, IMO, it makes sense to at least have a single `use` group for external crates in order to improve readability.

Artemka374: That was more because I haven't formatted the code properly

core/lib/zksync_core/src/api_server/web3/state.rs
```diff
@@ -543,10 +546,53 @@ impl<E> RpcState<E> {
...
 pub struct Filters {
-    state: HashMap<U256, TypedFilter>,
+    state: HashMap<U256, InstalledFilter>,
     max_cap: usize,
 }
 
+#[derive(Debug, Clone)]
+struct InstalledFilter {
+    pub filter: TypedFilter,
+    #[allow(dead_code)]
```
slowli, line 556: This lint is usually resolved by naming a field with a `_` prefix, e.g. `_guard`.

core/lib/zksync_core/src/api_server/web3/state.rs
```diff
@@ -543,10 +546,53 @@ impl<E> RpcState<E> {
...
+}
+
+impl InstalledFilter {
+    pub fn new(filter: TypedFilter) -> Self {
+        let guard = FILTER_METRICS.metrics_count[&FilterType::from(&filter)].inc_guard(1);
+        Self {
+            filter,
+            guard: Arc::new(guard),
+            created_at: Instant::now(),
+            last_request: chrono::Utc::now().naive_utc().timestamp() as u64,
```
slowli, line 570: Why not use `Instant` type for this field as well?

core/lib/zksync_core/src/api_server/web3/state.rs
```diff
@@ -543,10 +546,53 @@ impl<E> RpcState<E> {
...
+            request_count: 0,
+        }
+    }
+
+    pub fn update_stats(&mut self) {
+        let previous_request_timestamp = self.last_request;
+        let now = chrono::Utc::now().naive_utc().timestamp() as u64;
+
+        self.last_request = now;
+        self.request_count.add_assign(1);
```
slowli, line 580: `self.request_count += 1` does the same thing and is more conventional.

core/lib/zksync_core/src/api_server/web3/state.rs
```diff
@@ -543,10 +546,53 @@ impl<E> RpcState<E> {
...
+
+        let filter_type = FilterType::from(&self.filter);
+        FILTER_METRICS.request_frequency[&filter_type]
+            .observe(Duration::from_secs(now - previous_request_timestamp));
+    }
+}
+
+impl Drop for InstalledFilter {
+    fn drop(&mut self) {
+        FILTER_METRICS.filter_count[&FilterType::from(&self.filter)].observe(self.request_count);
```
slowli, line 590: Nit: you can extract type into a local variable to not create it twice (the compiler may optimize it away, but it's easier to read in any case).

core/lib/zksync_core/src/api_server/web3/state.rs
```diff
@@ -543,10 +546,53 @@ impl<E> RpcState<E> {
...
-    state: HashMap<U256, TypedFilter>,
+    state: HashMap<U256, InstalledFilter>,
     max_cap: usize,
 }
 
+#[derive(Debug, Clone)]
+struct InstalledFilter {
+    pub filter: TypedFilter,
+    #[allow(dead_code)]
+    guard: Arc<GaugeGuard>,
```
slowli, line 557: An `Arc` is a code smell, necessary only because of the `Clone` implementation. You can remove `derive(Clone)` for `InstalledFilter` and `Filters` since it's actually unused (which is natural – `Filters` are enclosed into an `Arc<RwLock<_>>`).

core/lib/zksync_core/src/api_server/web3/state.rs
```diff
@@ -578,13 +625,24 @@ impl Filters {
...
-        self.state.get(&index)
+        let installed_filter = self.state.get(&index)?;
+
+        Some(&installed_filter.filter)
+    }
+
+    pub fn update_stats(&mut self, index: U256) -> bool {
+        if let Some(installed_filter) = self.state.get_mut(&index) {
+            installed_filter.update_stats();
+            true
```
slowli, line 636: Nit: You don't seem to use the returned value anywhere, so it can be removed. Ditto for `update()`.

core/lib/zksync_core/src/api_server/web3/state.rs
```diff
@@ -578,13 +625,24 @@ impl Filters {
 
     /// Retrieves filter from the state.
     pub fn get(&self, index: U256) -> Option<&TypedFilter> {
-        self.state.get(&index)
+        let installed_filter = self.state.get(&index)?;
+
+        Some(&installed_filter.filter)
+    }
+
+    pub fn update_stats(&mut self, index: U256) -> bool {
```
slowli, line 633: I was thinking that it may be possible to pack updating stats and the filter together, but it's actually not the case (and even if it was possible, it probably makes sense to measure intervals between at the start of `getFilterChanges`, not at the end). So combining this logic with `get()` may make sense. A read lock on `Filters` can be acquire faster theoretically, but since in practice we never hold locks, the effect on performance should be minimal.

core/lib/zksync_core/src/api_server/web3/metrics.rs
```diff
@@ -195,3 +194,40 @@ pub(super) struct PubSubMetrics {
...
+        match value {
+            TypedFilter::Events(_, _) => FilterType::Events,
+            TypedFilter::Blocks(_) => FilterType::Blocks,
+            TypedFilter::PendingTransactions(_) => FilterType::PendingTransactions,
+        }
+    }
+}
+
+#[derive(Debug, Metrics)]
+#[metrics(prefix = "api_filter")]
```
slowli, line 217: Bikeshedding: `api_web3_filter` looks like a more appropriate prefix; most existing Web3-related metrics have the `api_web3` prefix.

core/lib/zksync_core/src/api_server/web3/metrics.rs
```diff
@@ -195,3 +194,40 @@ pub(super) struct PubSubMetrics {
...
+    /// Number of currently active filters grouped by the filter type
+    pub metrics_count: Family<FilterType, Gauge>,
+    /// Time in seconds between consecutive requests to the filter grouped by the filter type
+    #[metrics(buckets = Buckets::LATENCIES, unit = Unit::Seconds)]
+    pub request_frequency: Family<FilterType, Histogram<Duration>>,
+    /// Lifetime of a filter in seconds grouped by the filter type
+    #[metrics(buckets = Buckets::LATENCIES, unit = Unit::Seconds)]
+    pub filter_lifetime: Family<FilterType, Histogram<Duration>>,
+    /// Number of requests to the filter grouped by the filter type
+    #[metrics(buckets = Buckets::linear(0.0..=10.0, 1.0))]
```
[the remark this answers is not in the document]
Artemka374: That is true, I completely don't understand in which range it should be for now

core/lib/zksync_core/src/api_server/web3/state.rs
Artemka374: true, I was more focused on making new structures work and not changing old ones to not break anything

core/lib/zksync_core/src/api_server/web3/state.rs
Artemka374: Regarding that, do we need to have  `RwLock`  there at all in case we combine them?

Artemka374: I'm thinking maybe we can just use `Mutex`, since we are not doing just read anywhere

core/lib/zksync_core/src/api_server/web3/metrics.rs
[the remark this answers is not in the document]
RomanBrodetski: I'd do exponential until like 2^20

RomanBrodetski (commented):
Looks good! Let's wait for @slowli for the final review

core/lib/zksync_core/src/api_server/web3/namespaces/eth.rs
```diff
@@ -242,22 +243,34 @@ impl<G: L1GasPriceProvider> EthNamespace<G> {
...
+
+        let mut typed_filter = TypedFilter::Events(filter, from_block);
+
+        let logs = self.filter_changes(&mut typed_filter).await?;
+
+        self.state
+            .installed_filters
+            .lock()
+            .await
+            .update(idx, typed_filter);
```
slowli, line 272: Eh, isn't the previous behavior (the filter is **not** updated by the call) the correct one? Maybe, I wasn't clear in the previous comment here – I wanted you to add a comment why the filter shouldn't be updated, not to add logic for updating the filter.

core/lib/zksync_core/src/api_server/web3/state.rs
```diff
@@ -24,10 +25,11 @@ use zksync_types::{
 };
 use zksync_web3_decl::{
     error::Web3Error,
-    types::{Filter, Log, TypedFilter},
+    types::{Filter, Log},
 };
 
-use super::metrics::API_METRICS;
+use super::metrics::{FilterType, API_METRICS, FILTER_METRICS};
+use crate::api_server::web3::TypedFilter;
```
slowli, line 32: Nit: You can combine this `use` with the `use` clause below.

core/lib/zksync_core/src/api_server/web3/state.rs
```diff
@@ -577,17 +622,18 @@ impl Filters {
...
 
     /// Retrieves filter from the state.
-    pub fn get(&self, index: U256) -> Option<&TypedFilter> {
-        self.state.get(&index)
+    pub fn get_and_update_stats(&mut self, index: U256) -> Option<&TypedFilter> {
+        let installed_filter = self.state.get_mut(&index)?;
+
+        installed_filter.update_stats();
+
+        Some(&installed_filter.filter)
```
slowli, line 630: Bikeshedding: IMO, inserting an empty line after each code line like here can harm readability. Of course, it's very hard (impossible?) to quantify the empty line rules, so feel free to ignore this comment. (IMO, an "empty line per ~5–10 statements" is a good ratio, but it's a personal preference.)

core/lib/zksync_core/src/api_server/web3/state.rs
```diff
@@ -577,17 +622,18 @@ impl Filters {
     }
 
     /// Retrieves filter from the state.
-    pub fn get(&self, index: U256) -> Option<&TypedFilter> {
-        self.state.get(&index)
+    pub fn get_and_update_stats(&mut self, index: U256) -> Option<&TypedFilter> {
```
slowli, line 625: Nit: It may make sense to clone `TypedFilter` right here (so, return `Option<TypedFilter>`); otherwise, a shared reference will borrow `self` mutably (this is determined by `&mut self`), which sort of defeats its purpose.

slowli (commented):
Nice work! The only real question is changes in `get_filter_logs` (see the line comment).

core/lib/zksync_core/src/api_server/web3/namespaces/eth.rs
Artemka374: Uh, yes, I guess did not understand you correctly

core/lib/zksync_core/src/api_server/web3/state.rs
Artemka374: I am usually more preferring to divide logical blocks by empty lines or in places where it becomes hard to understand what is going on (in this place it is more about logical blocks)

slowli (approved):

PR #533, opened 2023-11-22
feat: zk fmt sqlx-queries

tomg10:
## What ❔

Added option to format queries using `zk fmt sqlx-queries` 
It uses https://github.com/sql-formatter-org/sql-formatter as a formatter

## Checklist

### DIFF
--- a/core/lib/dal/src/accounts_dal.rs
+++ b/core/lib/dal/src/accounts_dal.rs
@@ -42,15 +42,21 @@ impl AccountsDal<'_, '_> {
             })
             .collect();
         let rows = sqlx::query!(
-            r#"
-                SELECT storage.value as "value!",
-                    tokens.l1_address as "l1_address!", tokens.l2_address as "l2_address!",
-                    tokens.symbol as "symbol!", tokens.name as "name!", tokens.decimals as "decimals!", tokens.usd_price as "usd_price?"
-                    FROM storage
-                INNER JOIN tokens ON
-                    storage.address = tokens.l2_address OR (storage.address = $2 AND tokens.l2_address = $3)
-                WHERE storage.hashed_key = ANY($1) AND storage.value != $4
-            "#,
+            "SELECT STORAGE.value AS \"value!\", \
+                    tokens.l1_address AS \"l1_address!\", \
+                    tokens.l2_address AS \"l2_address!\", \
+                    tokens.symbol AS \"symbol!\", \
+                    tokens.name AS \"name!\", \
+                    tokens.decimals AS \"decimals!\", \
+                    tokens.usd_price AS \"usd_price?\" \
+               FROM STORAGE \
+              INNER JOIN tokens ON STORAGE.address = tokens.l2_address \
+                 OR ( \
+                    STORAGE.address = $2 \
+                AND tokens.l2_address = $3 \
+                    ) \
+              WHERE STORAGE.hashed_key = ANY ($1) \
+                AND STORAGE.value != $4",
             &hashed_keys,
             L2_ETH_TOKEN_ADDRESS.as_bytes(),
             ETHEREUM_ADDRESS.as_bytes(),
--- a/core/lib/dal/src/accounts_dal.rs
+++ b/core/lib/dal/src/accounts_dal.rs
@@ -42,21 +42,21 @@ impl AccountsDal<'_, '_> {
             })
             .collect();
         let rows = sqlx::query!(
-            "SELECT STORAGE.value AS \"value!\", \
+            "SELECT storage.value AS \"value!\", \
                     tokens.l1_address AS \"l1_address!\", \
                     tokens.l2_address AS \"l2_address!\", \
                     tokens.symbol AS \"symbol!\", \
                     tokens.name AS \"name!\", \
                     tokens.decimals AS \"decimals!\", \
                     tokens.usd_price AS \"usd_price?\" \
-               FROM STORAGE \
-              INNER JOIN tokens ON STORAGE.address = tokens.l2_address \
+               FROM storage \
+              INNER JOIN tokens ON storage.address = tokens.l2_address \
                  OR ( \
-                    STORAGE.address = $2 \
+                    storage.address = $2 \
                 AND tokens.l2_address = $3 \
                     ) \
-              WHERE STORAGE.hashed_key = ANY ($1) \
-                AND STORAGE.value != $4",
+              WHERE storage.hashed_key = ANY ($1) \
+                AND storage.value != $4",
             &hashed_keys,
             L2_ETH_TOKEN_ADDRESS.as_bytes(),
             ETHEREUM_ADDRESS.as_bytes(),
--- a/core/lib/dal/src/accounts_dal.rs
+++ b/core/lib/dal/src/accounts_dal.rs
@@ -42,21 +42,24 @@ impl AccountsDal<'_, '_> {
             })
             .collect();
         let rows = sqlx::query!(
-            "SELECT storage.value AS \"value!\", \
-                    tokens.l1_address AS \"l1_address!\", \
-                    tokens.l2_address AS \"l2_address!\", \
-                    tokens.symbol AS \"symbol!\", \
-                    tokens.name AS \"name!\", \
-                    tokens.decimals AS \"decimals!\", \
-                    tokens.usd_price AS \"usd_price?\" \
-               FROM storage \
-              INNER JOIN tokens ON storage.address = tokens.l2_address \
+            "SELECT \
+                 storage.value AS \"value!\", \
+                 tokens.l1_address AS \"l1_address!\", \
+                 tokens.l2_address AS \"l2_address!\", \
+                 tokens.symbol AS \"symbol!\", \
+                 tokens.name AS \"name!\", \
+                 tokens.decimals AS \"decimals!\", \
+                 tokens.usd_price AS \"usd_price?\" \
+             FROM \
+                 storage \
+                 INNER JOIN tokens ON storage.address = tokens.l2_address \
                  OR ( \
-                    storage.address = $2 \
-                AND tokens.l2_address = $3 \
-                    ) \
-              WHERE storage.hashed_key = ANY ($1) \
-                AND storage.value != $4",
+                     storage.address = $2 \
+                     AND tokens.l2_address = $3 \
+                 ) \
+             WHERE \
+                 storage.hashed_key = ANY ($1) \
+                 AND storage.value != $4",
             &hashed_keys,
             L2_ETH_TOKEN_ADDRESS.as_bytes(),
             ETHEREUM_ADDRESS.as_bytes(),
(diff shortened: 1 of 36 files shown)


core/lib/dal/src/accounts_dal.rs
```diff
@@ -42,15 +42,21 @@ impl AccountsDal<'_, '_> {
...
+                    tokens.decimals AS \"decimals!\", \
+                    tokens.usd_price AS \"usd_price?\" \
+               FROM storage \
+              INNER JOIN tokens ON storage.address = tokens.l2_address \
+                 OR ( \
+                    storage.address = $2 \
+                AND tokens.l2_address = $3 \
+                    ) \
+              WHERE storage.hashed_key = ANY ($1) \
+                AND storage.value != $4",
```
montekki, line 59: I suggest using [raw strings](https://tourofrust.com/63_en.html) you can get rid of the backslashes here.

RomanBrodetski: @montekki see discussion here: https://github.com/matter-labs/zksync-era/pull/498#discussion_r1395441556

infrastructure/zk/src/format_sql.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    const rawQuery = extractQueryFromRustString(query);
+    const formattedQuery = formatQuery(rawQuery);
+    const reconstructedRustString = embedTextInsideRustString(formattedQuery);
+
+    return addIndent(reconstructedRustString, baseIndent) + (endedWithAComma ? ',' : '') + '\n';
+}
+
+async function formatFile(filePath: string) {
+    let content = fs.readFileSync(filePath, { encoding: 'utf-8' }).split('\n');
```
slowli, line 89: Nit: I understand that we shouldn't care that much about infra code (as opposed to the server code), but shouldn't this function either use an async version of I/O calls, or be declared sync itself?

infrastructure/zk/src/format_sql.ts
```diff
@@ -0,0 +1,147 @@
...
+
+function formatQuery(query: string) {
+    let formattedQuery = query;
+    try {
+        formattedQuery = format(query, {
+            language: 'postgresql',
+            tabWidth: 4,
+            keywordCase: 'upper',
+            expressionWidth: 80,
+            indentStyle: 'tabularRight'
```
slowli, line 13: Bikeshedding: IMO, the standard ident style is more readable; `tabularRight` produces fewer lines, but because of right-aligning some keywords, it has some unexpected (personal opinion) spacing, such as one at the start of the first line.

infrastructure/zk/src/fmt.ts
```diff
@@ -61,6 +62,13 @@ export const command = new Command('fmt')
                 case 'system-contracts':
                     await prettierSystemContracts(cmd.check);
                     break;
+                case 'sqlx-queries':
+                    if (cmd.check) {
+                        console.warn('format-sqlx-queries does not support --check, skipping!');
```
slowli, line 67: Wouldn't it be possible to support checking by parsing all `sqlx` queries and checking for each of them that the formatted version is exactly equal to the original version?

slowli (commented):
Awesome idea! 👍 As usual, I do have some bikeshedding topics.

infrastructure/zk/src/format_sql.ts
tomg10: I changed the code so that it is really async

infrastructure/zk/src/fmt.ts
tomg10: corrected

infrastructure/zk/src/format_sql.ts
tomg10: As this is something that is very personal, I will make an experiment and let the team vote on slack to choose the formatting style

infrastructure/zk/src/format_sql.ts
```diff
@@ -0,0 +1,161 @@
...
+    for (const keyword of keywordToLower) {
+        // we replace keyword but only if it's not part of a longer word
+        formattedQuery = formattedQuery.replace(RegExp(`\\b${keyword}\\b`, 'g'), keyword.toLowerCase());
+    }
+
+    const formattedLines = formattedQuery.split('\n');
+
+    const minIndent = Math.min(...formattedLines.map((line) => line.search(/\S/)));
+    formattedQuery = formattedQuery
+        .split('\n')
```
slowli, line 30: Nit: You can use `formattedLines` here instead of `formattedQuery.split('\n')`.

infrastructure/zk/src/format_sql.ts
```diff
@@ -0,0 +1,161 @@
...
+        query = query.slice(0, query.length - 1);
+    }
+    //removing quotes
+    if (!query.startsWith('r#')) {
+        query = query.slice(1, query.length - 1);
+    } else {
+        query = query.slice(3, query.length - 2);
+    }
+
+    //getting rid of all "/" characters, both from escapes and line breaks
```
slowli, line 49: Nit: it's "\" characters, not "/". Also, theoretically we could have an escaped `\` chars in the string, although I don't think they are used anywhere rn.

infrastructure/zk/src/format_sql.ts
```diff
@@ -0,0 +1,161 @@
...
+
+function embedTextInsideRustString(query: string) {
+    query = query.replace(/"/g, '\\"');
+    const formattedLines = query.split('\n');
+    for (let i = 0; i < formattedLines.length; i++) {
+        let currentLine = '';
+        currentLine += i == 0 ? '"' : ' ';
+        currentLine += formattedLines[i];
+        currentLine += i != formattedLines.length - 1 ? ' \\' : '';
+        currentLine += i == formattedLines.length - 1 ? '"' : '';
```
slowli, line 63: Nit: Please use strict comparisons (`===` / `!==`) everywhere; non-strict comparisons are considered harmful in JS / TS.

infrastructure/zk/src/format_sql.ts
```diff
@@ -0,0 +1,161 @@
...
+function addIndent(query: string, indent: number) {
+    return query
+        .split('\n')
+        .map((line) => ' '.repeat(indent) + line)
+        .join('\n');
+}
+
+function formatRustStringQuery(query: string) {
+    const baseIndent = query.search(/\S/);
+    let endedWithAComma = query.trim().endsWith(',');
```
slowli, line 79: - Nit: it seems `endedWithAComma` can be declared as a `const`.
- Bikeshedding: Subjectively, articles (a / the) are usually skipped in identifiers (i.e., `endedWithComma`).
- IMO, it would make sense to remove the comma here so that `extractQueryFromRustString` doesn't need to handle it as well.

infrastructure/zk/src/format_sql.ts
```diff
@@ -0,0 +1,161 @@
...
+        } else {
+            modifiedFile += line + '\n';
+        }
+    }
+    modifiedFile = modifiedFile.slice(0, modifiedFile.length - 1);
+
+    if (!check) {
+        await fs.promises.writeFile(filePath, modifiedFile);
+    } else {
+        if (content !== modifiedFile) {
```
slowli, line 142: Nit: Can't this be collapsed with the preceding `else` into `else if`?

infrastructure/zk/src/format_sql.ts
```diff
@@ -0,0 +1,161 @@
...
+            const rawStringQueryEnded = (line.endsWith('"#,') || line.endsWith('"#')) && builtQuery;
+            const regularStringQueryEnded = (line.endsWith('",') || line.endsWith('"')) && builtQuery;
+            builtQuery += line + '\n';
+            const lineEndIsNotEscape = !line.endsWith('\\"') && !line.endsWith('\\",');
+            if (
+                (isRawString && rawStringQueryEnded) ||
+                (!isRawString && regularStringQueryEnded && lineEndIsNotEscape)
+            ) {
+                isInsideQuery = false;
+                let formattedQuery = formatRustStringQuery(builtQuery);
```
slowli, line 130: Nit: This var looks redundant (and if it stays, should be a `const`).

infrastructure/zk/src/format_sql.ts
```diff
@@ -0,0 +1,161 @@
...
+async function formatFile(filePath: string, check: boolean) {
+    let content = await fs.promises.readFile(filePath, { encoding: 'utf-8' });
+    let linesToQuery = null;
+    let isInsideQuery = false;
+    let isRawString = false;
+    let builtQuery = '';
+
+    let modifiedFile = '';
+    for (const line of content.split('\n')) {
+        if (line.endsWith('sqlx::query!(')) {
```
slowli, line 97: Are single-line queries intentionally excluded from formatting? They could theoretically be formatted into multiple lines by the formatter. I'm fine with tackling these queries separately.

infrastructure/zk/src/format_sql.ts
```diff
@@ -0,0 +1,161 @@
...
+        if (content !== modifiedFile) {
+            console.warn(`Sqlx query format issues found in ${filePath}.`);
+        }
+    }
+    return content === modifiedFile;
+}
+
+export async function formatSqlxQueries(check: boolean) {
+    process.chdir(`${process.env.ZKSYNC_HOME}`);
+    let { stdout: filesRaw } = await utils.exec('find core/lib/dal -type f -name "*.rs"');
```
slowli, line 151: Nit: This may be OS-specific; it would be safer to use something like [`fs.readdir`](https://nodejs.org/dist/latest-v20.x/docs/api/fs.html#fspromisesreaddirpath-options). OTOH, we probably already have quite a lot *NIX-specific logic.

infrastructure/zk/src/format_sql.ts
tomg10: corrected

infrastructure/zk/src/format_sql.ts
tomg10: corrected

infrastructure/zk/src/format_sql.ts
tomg10: corrected

infrastructure/zk/src/format_sql.ts
tomg10: corrected

infrastructure/zk/src/format_sql.ts
RomanBrodetski: @tomg10 I agree it's not a big deal but I'm a bit suprized - sounds like an almost perfect usecase for `else if`

slowli (approved):
LGTM, but I'd love other opinions as to whether this formatting makes sense. As some potential alternatives (given that we have a tool to unify SQL queries):

- We could use raw literal strings without indentation. IMO, they look ugly.
- We could use raw literal strings *with* indentation. Doesn't satisfy the "query is not recompiled regardless of surrounding formatting" criterion.
- We could extract queries into files (i.e., use [`query_file`](https://docs.rs/sqlx/0.5.13/sqlx/macro.query_file.html) etc.). IMO, extracting queries makes it more difficult to keep the query context when reviewing code.

FWIW, I've tried updating `sqlx` to the latest version (0.7), and it still doesn't transform whitespace in query literals; so, we wouldn't avoid the formatting problem by updating.

tomg10:
@slowli I will add this formatting to agenda to our next weekly platform team meeting

slowli (approved):

PR #535, opened 2023-11-22
feat: Add metric to CallTracer for calculating maximum depth of the calls

Artemka374:
## What ❔

Add metric to CallTracer for calculating maximum depth of the calls

## Why ❔

We need to know what our limits are.

## Checklist


core/lib/multivm/src/tracers/call_tracer/mod.rs
```diff
@@ -38,4 +52,16 @@ impl CallTracer {
         let cell = self.result.as_ref();
         cell.set(result).unwrap();
     }
+
+    fn push_call_and_update_stats(&mut self, call: FarcallAndNearCallCount) {
+        self.stack.push(call);
+        self.max_stack_depth = self.max_stack_depth.max(self.stack.len());
```
Deniallugo, line 58: we have to include near calls as well

Artemka374: but near calls are updated in the other method, when the actual increasing if value occurs

Artemka374: Ah, okay, I understand your point

Deniallugo: I mean, that near call is also part of the actual depth.
We just don't add recursion here because it can cause stack overflow later

Artemka374: Got it, then my previous thought was a bit wrong :)

core/lib/multivm/src/tracers/call_tracer/mod.rs
```diff
@@ -38,4 +52,24 @@ impl CallTracer {
         let cell = self.result.as_ref();
         cell.set(result).unwrap();
     }
+
+    fn push_call_and_update_stats(&mut self, call: FarcallAndNearCallCount) {
+        self.stack.push(call.clone());
+
+        self.max_stack_depth = self
+            .max_stack_depth
+            .max(self.stack.len() + call.near_calls_after);
```
slowli, line 61: Nit: You can extract `call.near_calls_after` into a local var to avoid cloning `call` above. Cloning may be optimized away by the compiler, but I wouldn't rely on it.

core/lib/multivm/src/tracers/call_tracer/mod.rs
```diff
@@ -18,11 +23,20 @@ struct FarcallAndNearCallCount {
     near_calls_after: usize,
 }
 
+impl Drop for CallTracer {
+    fn drop(&mut self) {
+        CALL_METRICS.call_stack_depth.observe(self.max_stack_depth);
+        CALL_METRICS.max_near_calls.observe(self.max_near_calls);
+    }
+}
```
slowli, line 31: Stupid question: Is it correct that a single tracer is always used for exactly one transaction? Otherwise, we'd get bogus zero values for metrics and/or "glue" some values together.

core/lib/multivm/src/tracers/call_tracer/mod.rs
```diff
@@ -38,4 +52,24 @@ impl CallTracer {
...
+    fn push_call_and_update_stats(&mut self, call: FarcallAndNearCallCount) {
+        self.stack.push(call.clone());
+
+        self.max_stack_depth = self
+            .max_stack_depth
+            .max(self.stack.len() + call.near_calls_after);
+        self.max_near_calls = self.max_near_calls.max(call.near_calls_after);
+    }
+
+    fn increase_last_and_update_stats(&mut self, value: usize) {
```
slowli, line 65: - Nit: AFAICT, this is always called with `value == 1`. Could this be called with another `value`? If not, it may make sense to remove the arg in order to simplify the code.
- Bikeshedding: Maybe name this method something like `increase_near_call_count`? The current name looks a bit esoteric.

core/lib/multivm/src/tracers/call_tracer/metrics.rs
```diff
@@ -0,0 +1,15 @@
+use vise::{Buckets, Histogram, Metrics};
+
+#[derive(Debug, Metrics)]
+#[metrics(prefix = "call_tracer")]
```
slowli, line 4: Bikeshedding: It may make sense to include the high-level component name (`vm`?) first – `vm_call_tracer`.

core/lib/multivm/src/tracers/call_tracer/metrics.rs
```diff
@@ -0,0 +1,15 @@
...
+pub struct CallMetrics {
+    /// Maximum call stack depth during the execution of the transaction.
+    #[metrics(buckets = Buckets::linear(1.0..=64.0, 1.0))]
```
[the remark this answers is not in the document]
Artemka374: TBH no, I meant linear, just wanted to understand which range is better for that metric :) But I guess we can do that exponential

core/lib/multivm/src/tracers/call_tracer/mod.rs
Artemka374: As far as I understood - yes, but maybe I am wrong

slowli (dismissed):
Looks good. I'm not that familiar with VM, so I suggest to wait for @Deniallugo's approval.

core/lib/multivm/src/tracers/call_tracer/mod.rs
```diff
@@ -38,4 +52,25 @@ impl CallTracer {
...
         cell.set(result).unwrap();
     }
+
+    fn push_call_and_update_stats(&mut self, call: FarcallAndNearCallCount) {
+        let near_calls_after = call.near_calls_after;
+        self.stack.push(call);
+
+        self.max_stack_depth = self
+            .max_stack_depth
+            .max(self.stack.len() + near_calls_after);
```
Deniallugo, line 62: It's definitely not correct, because you don't take into account that all stack calls have this near_call_after.

core/lib/multivm/src/tracers/call_tracer/vm_latest/mod.rs
```diff
@@ -185,9 +183,11 @@ impl CallTracer {
             return;
         };
 
+        self.call_depth_on_prefix.pop();
```
Deniallugo, line 186: seems you have to pop only for far calls

Artemka374: Generally yes, but since we doing pop for stack and after that it can push something to stack again, where it will recalculate new value for prefix, it should work okay

Artemka374: So it will anyway be pushed again and length of call stack will be the same with prefix vec length all the time

Deniallugo: Let's write a test :)

core/lib/multivm/src/tracers/call_tracer/vm_latest/mod.rs
```diff
@@ -187,7 +182,7 @@ impl CallTracer {
 
         if current_call.near_calls_after > 0 {
             current_call.near_calls_after -= 1;
-            self.stack.push(current_call);
+            self.push_call_and_update_stats(current_call.farcall, current_call.near_calls_after);
```
Deniallugo, line 185: I don't think you have to update number of calls here. For Each push, you are adding number of near calls to the depth_prefix as a result for near call with length 3 you will add 6 to the depth.

PR #537, opened 2023-11-23
feat(en): Support arbitrary genesis block for external nodes

slowli:
## What ❔

Support non-zero genesis block specified in executor configuration. Check whether this block exists on initialization; validate its correspondence if it does, and persist consensus fields if it doesn't.

## Why ❔

This is necessary to support gossip-based syncing in practice; we likely won't back-sign all blocks in all envs.

## Checklist

### DIFF
--- a/core/lib/zksync_core/src/sync_layer/fetcher.rs
+++ b/core/lib/zksync_core/src/sync_layer/fetcher.rs
@@ -6,7 +6,7 @@ use std::time::Duration;
 use zksync_dal::StorageProcessor;
 use zksync_types::{
     api::en::SyncBlock, block::ConsensusBlockFields, Address, L1BatchNumber, MiniblockNumber,
-    ProtocolVersionId, H256,
+    ProtocolVersionId,
 };
 use zksync_web3_decl::jsonrpsee::core::Error as RpcError;
 
@@ -29,7 +29,6 @@ pub(super) struct FetchedBlock {
     pub last_in_batch: bool,
     pub protocol_version: ProtocolVersionId,
     pub timestamp: u64,
-    pub hash: H256,
     pub l1_gas_price: u64,
     pub l2_fair_gas_price: u64,
     pub virtual_blocks: u32,
@@ -38,15 +37,14 @@ pub(super) struct FetchedBlock {
     pub consensus: Option<ConsensusBlockFields>,
 }
 
-impl FetchedBlock {
-    fn from_sync_block(block: SyncBlock) -> Self {
+impl From<SyncBlock> for FetchedBlock {
+    fn from(block: SyncBlock) -> Self {
         Self {
             number: block.number,
             l1_batch_number: block.l1_batch_number,
             last_in_batch: block.last_in_batch,
             protocol_version: block.protocol_version,
             timestamp: block.timestamp,
-            hash: block.hash.unwrap_or_default(),
             l1_gas_price: block.l1_gas_price,
             l2_fair_gas_price: block.l2_fair_gas_price,
             virtual_blocks: block.virtual_blocks.unwrap_or(0),
@@ -64,7 +62,6 @@ impl FetchedBlock {
 pub struct FetcherCursor {
     // Fields are public for testing purposes.
     pub(super) next_miniblock: MiniblockNumber,
-    pub(super) prev_miniblock_hash: H256,
     pub(super) l1_batch: L1BatchNumber,
 }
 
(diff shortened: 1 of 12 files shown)


core/lib/zksync_core/src/sync_layer/gossip/storage/mod.rs
```diff
@@ -116,22 +213,28 @@ impl PostgresBlockStorage {
             .map_err(StorageError::Database)
     }
 
+    async fn sync_block(
+        ctx: &ctx::Ctx,
+        storage: &mut StorageProcessor<'_>,
+        number: MiniblockNumber,
+    ) -> StorageResult<Option<SyncBlock>> {
+        let operator_address = Address::default(); // FIXME: where to get this address from?
```
slowli, line 221: Just to highlight this remnant from the previous PR: I don't know where this address can be obtained from. It's used in case a miniblock is not sealed.

core/lib/zksync_core/src/sync_layer/gossip/storage/mod.rs
```diff
@@ -67,23 +70,117 @@ impl CursorWithCachedBlock {
...
 
 impl PostgresBlockStorage {
     /// Creates a new storage handle. `pool` should have multiple connections to work efficiently.
-    pub fn new(pool: ConnectionPool, actions: ActionQueueSender, cursor: FetcherCursor) -> Self {
+    pub async fn new(
+        ctx: &ctx::Ctx,
+        pool: ConnectionPool,
+        actions: ActionQueueSender,
+        cursor: FetcherCursor,
+        genesis_block: &FinalBlock,
```
slowli, line 86: AFAIU, this approach won't work for initializing the main node storage; we'll have the chicken-and-egg problem. (Where do we get the genesis block params from? Either we'll need to sign blocks retroactively, or we'll just have the block number.) OTOH, I have some doubts that `PostgresBlockStorage` could be used for the main node as-is (of course, most of it could be reused, but it may require some variability depending on the node role).

slowli (commented):
Related to https://github.com/matter-labs/era-consensus/pull/40; while additional validation of `ExecutorConfig` is not strictly required, it allows to avoid some nasty corner cases (e.g., an EN using an incorrect genesis block).

core/lib/zksync_core/src/sync_layer/fetcher.rs
```diff
@@ -38,15 +37,14 @@ pub(super) struct FetchedBlock {
...
-    fn from_sync_block(block: SyncBlock) -> Self {
+impl From<SyncBlock> for FetchedBlock {
+    fn from(block: SyncBlock) -> Self {
         Self {
             number: block.number,
             l1_batch_number: block.l1_batch_number,
             last_in_batch: block.last_in_batch,
             protocol_version: block.protocol_version,
             timestamp: block.timestamp,
-            hash: block.hash.unwrap_or_default(),
```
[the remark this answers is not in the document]
pompon0: IMO local validation of the block would be a good idea, even if what we receive is properly signed.

core/lib/zksync_core/src/sync_layer/gossip/storage/mod.rs
pompon0: @brunoffranca  what was the long term plan on how to deal with operator address? AFAIR it is currently fixed anyway, but eventually will be dynamic.

core/lib/zksync_core/src/sync_layer/gossip/storage/tests.rs
```diff
@@ -125,3 +141,190 @@ async fn processing_new_blocks() {
         .unwrap();
     assert_second_block_actions(&mut actions).await;
 }
+
+#[tokio::test]
+async fn ensuring_consensus_fields_for_genesis_block() {
+    let ctx = &ctx::test_root(&ctx::RealClock);
```
pompon0, line 147: abort_on_panic?

pompon0 (approved):

core/lib/zksync_core/src/sync_layer/fetcher.rs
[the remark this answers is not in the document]
slowli: сc @popzxc @RomanBrodetski WDYT? Should I create a task for implementing block hash validation?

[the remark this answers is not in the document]
RomanBrodetski: It really does seem like a good idea to validate, maybe I don't see the other side of the coin? What's the downside? Performance/Engineering time?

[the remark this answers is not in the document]
slowli: I see no downsides; I've started the implementation already 🙃

core/lib/zksync_core/src/sync_layer/gossip/storage/mod.rs
slowli: JFYI, I've created a separate task in Linear to deal with the operator address.

PR #541, opened 2023-11-24
feat: refactor witness_vector_generator metrics(WIP)

Artemka374:
## What ❔

Add various metrics to the Prover subsystems.
Refactor metrics in prover to use vise.

## Why ❔

We have some metric coverage on the prover subsystem, but it's incomplete.

## Checklist


prover/prover_fri_utils/src/metrics.rs
```diff
@@ -0,0 +1,41 @@
...
+
+impl std::fmt::Display for StageLabel {
+    fn fmt(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        formatter.write_str(match self.0 {
+            AggregationRound::BasicCircuits => "basic_circuits",
+            AggregationRound::LeafAggregation => "leaf_aggregation",
+            AggregationRound::NodeAggregation => "node_aggregation",
+            AggregationRound::Scheduler => "scheduler",
+        })
+    }
```
RomanBrodetski, line 29: I'm surprised we didn't have this mapping. I checked - and indeed we didn't! @EmilLuta just wanted to draw your attention here - so that next time we reuse this mapping and not create new

prover/witness_generator/src/metrics.rs
```diff
@@ -0,0 +1,65 @@
...
+    }
+}
+
+impl std::fmt::Display for StageLabel {
+    fn fmt(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        formatter.write_str(match self.0 {
+            AggregationRound::BasicCircuits => "basic_circuits",
+            AggregationRound::LeafAggregation => "leaf_aggregation",
+            AggregationRound::NodeAggregation => "node_aggregation",
+            AggregationRound::Scheduler => "scheduler",
```
RomanBrodetski, line 24: wait, didn't you have this metrics above? I'm confused now

RomanBrodetski (commented):
Overall looks good, please consider my comments.

Separately, we should probably differentiate between the metrics on server side and those on prover side - I mentioned this in one of the comments above. Maybe we should also differentiate on the metrics level. In practical terms, I think only the `proof-generation-details` is on the server side - so the relevant topics should be prefixed with something like `server.prover_data_handler. ... `. But that's optional

prover/witness_generator/src/metrics.rs
Artemka374: Do you mean this StageLabel type? If yes, then the thing is that we need to use AggregationRound, which is an external type and we use wrappers for it here. Since all these metrics are in different crates, I wasn't really sure how to work with them.

prover/witness_generator/src/metrics.rs
```diff
@@ -0,0 +1,65 @@
...
+    #[metrics(buckets = Buckets::LATENCIES)]
+    pub blob_save_time: Family<StageLabel, Histogram<Duration>>,
+}
+
+#[vise::register]
+pub(crate) static WITNESS_GENERATOR_METRICS: vise::Global<WitnessGeneratorMetrics> =
+    vise::Global::new();
+
+#[derive(Debug, Metrics)]
+#[metrics(prefix = "server_witness_generator_fri")]
```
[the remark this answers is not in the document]
RomanBrodetski: I agree, not, it will not be too difficult - IMO worth it. I raise a similar point in my comment above

core/lib/types/src/proofs.rs
```diff
@@ -98,6 +98,17 @@ impl AggregationRound {
     }
 }
 
+impl std::fmt::Display for AggregationRound {
+    fn fmt(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        formatter.write_str(match self {
+            AggregationRound::BasicCircuits => "basic_circuits",
```
slowli, line 104: Nit: You can use `Self::BasicCircuits` here for brevity. Also, the `types` crate uses [`strum`](https://docs.rs/strum/) to automate similar conversions for some other enums (although I can't say I'm a fan of having this kind of automations that can easily be written by hand).

core/lib/dal/src/fri_prover_dal.rs
```diff
@@ -341,6 +341,30 @@ impl FriProverDal<'_, '_> {
         }
     }
 
+    pub async fn min_unproved_l1_batch_number_for_aggregation_round(
+        &mut self,
+        aggregation_round: AggregationRound,
+    ) -> Option<L1BatchNumber> {
+        {
```
slowli, line 348: This additional scope is not necessary; it's present for some existing methods as a refactoring artifact. Also, it could make sense to use a consistent literal style for queries (e.g., ordinary literals with escaped newlines, like in `get_oldest_not_compressed_batch` – it is resistant to changes in the surrounding indentation and results in a searchable query in `sqlx-data.json`). OTOH, after #533 is merged, it would be possible to format queries automatically.

prover/circuit_synthesizer/src/metrics.rs
```diff
@@ -0,0 +1,15 @@
...
+#[metrics(prefix = "prover_circuit_synthesizer")]
+pub(crate) struct CircuitSynthesizerMetrics {
+    #[metrics(buckets = Buckets::LATENCIES, labels = ["blob_size_in_gb"])]
+    pub blob_sending_time: LabeledFamily<u64, Histogram<Duration>>,
+    #[metrics(buckets = Buckets::LATENCIES, labels = ["circuit_type"])]
+    pub synthesize: LabeledFamily<&'static str, Histogram<Duration>>,
+}
+
+#[vise::register]
+pub(crate) static CIRCUIT_SYNTHESIZER_METRICS: vise::Global<CircuitSynthesizerMetrics> =
```
slowli, line 14: Bikeshedding: If a crate uses a single metrics type, IMO, the corresponding static could be safely called just `METRICS` – there are no other metrics to confuse them with.

prover/prover/src/synthesized_circuit_provider.rs
```diff
@@ -69,11 +70,10 @@ impl RemoteSynthesizer for SynthesizedCircuitProvider {
...
                     assembly_queue.capacity()
                 );
-                metrics::histogram!(
-                    "server.prover.queue_free_slots",
-                    queue_free_slots as f64,
-                    "queue_capacity" => assembly_queue.capacity().to_string()
-                );
+                let label: &'static str =
+                    Box::leak(assembly_queue.capacity().to_string().into_boxed_str());
+                PROVER_METRICS.queue_free_slots[&label].observe(queue_free_slots as f64);
```
slowli, line 75: - There are a couple of remaining `Box:leak()` places; please fix.
- You should define the histogram as `Histogram<usize>` so that the original value type is reflected in the metric definition.

prover/prover_fri_utils/Cargo.toml
```diff
@@ -6,12 +6,14 @@ edition = "2021"
...
 zksync_object_store = { path = "../../core/lib/object_store" }
 zksync_config = { path = "../../core/lib/config" }
 zksync_types = { path = "../../core/lib/types" }
 zksync_prover_fri_types = { path = "../prover_fri_types" }
 zksync_dal = { path = "../../core/lib/dal" }
 
 tracing = "0.1"
-serde = { version = "1.0", features = ["derive"] }
-metrics = "0.21"
+serde = { version = "1.0", features = ["derive"] }
```
slowli, line 19: Nit: Missing newline at the end of the file.

prover/witness_generator/src/main.rs
```diff
@@ -225,11 +229,8 @@ async fn main() -> anyhow::Result<()> {
             round,
             started_at.elapsed()
         );
-        metrics::gauge!(
-            "server.init.latency",
-            started_at.elapsed(),
-            "stage" => format!("fri_witness_generator_{:?}", round)
-        );
+        let label: String = format!("fri_witness_generator_{:?}", opt.round);
```
slowli, line 232: - Why is it changed to `opt.round` here? AFAIU, it's not always equal to `round`.
- Ideally, we'd want to use a wrapper similar to `StageLabel` here. (Or maybe, `StageLabel` itself if it doesn't cost too much to change dashboards etc.)

RomanBrodetski (approved):

PR #546, opened 2023-11-27
fix: Cursor not moving correctly after poll in `get_filter_changes`

Artemka374:
## What ❔

When polling filter changes, add 1 to actual from_block value

## Why ❔

Otherwise, last block that was included in poll will be included to the next one.

## Checklist


core/lib/zksync_core/src/api_server/web3/namespaces/eth.rs
```diff
@@ -861,9 +861,8 @@ impl<G: L1GasPriceProvider> EthNamespace<G> {
                     .map_err(|err| internal_error(METHOD_NAME, err))?;
                 *from_block = logs
                     .last()
-                    .map(|log| MiniblockNumber(log.block_number.unwrap().as_u32()))
+                    .map(|log| MiniblockNumber(log.block_number.unwrap().as_u32() + 1))
                     .unwrap_or(*from_block);
```
slowli, line 865: AFAIU, you could increase the cursor here as well to at least `*from_block + 1`. More precisely, in both cases (whether `logs` are empty or not), the cursor can be increased to `to_block + 1`. The current implementation will query redundant miniblock ranges for events that we know (by performing previous requests) are not there. 

There's a caveat: there seems to be a TOCTOU possibility with the `*from_block != to_block` check, and generally, fetched events may belong to a greater miniblock that was inserted into the DB after `to_block` was fetched. IMO, it would make sense to refactor `get_logs()` so that it takes a concrete range of miniblock numbers – `from_block..=to_block` – as an input. This would make this logic easier to reason about and would probably simplify DAL logic as well.

core/lib/zksync_core/src/api_server/web3/namespaces/eth.rs
```diff
@@ -861,9 +861,8 @@ impl<G: L1GasPriceProvider> EthNamespace<G> {
                     .map_err(|err| internal_error(METHOD_NAME, err))?;
                 *from_block = logs
```
slowli, line 862: Off-topic (?): I don't quite like that in `TypedFilter::Blocks`, the cursor (named `from_block` in the code above) is exclusive (i.e., `get_block_hashes_after(from_block, _)` fetches block hashes starting from `from_block + 1`), but here, the cursor is inclusive. I think this could be the cause of the bug in the first place. IMO, it would make sense to uniformly use either exclusive or inclusive cursor for all `TypedFilter` variants (I'd slightly prefer inclusive, but really any consistent option would do fine).

slowli (commented):
While the current fix works, I wonder whether we can do more comprehensive refactoring of `TypedFilter`-related code (see line comments).

core/lib/zksync_core/src/api_server/web3/namespaces/eth.rs
Artemka374: The thing is that if `from_block` is greater than actual latest block number and the query occurred before adding the block with that number, cursor will move forward anyway and changes from this block won't be returned. If talking about moving it to `to_block + 1` we are actually thinking of doing it that way, but again, if user manually entered `to_block` greater then latest number of block, we should move it to the latest current block number

core/lib/zksync_core/src/api_server/web3/namespaces/eth.rs
```diff
@@ -826,6 +826,10 @@ impl<G: L1GasPriceProvider> EthNamespace<G> {
                     .state
                     .resolve_filter_block_number(filter.to_block)
                     .await?;
+                let latest_block = self
+                    .state
+                    .resolve_filter_block_number(Some(BlockNumber::Latest))
+                    .await?;
```
montekki, line 832: Why is this constantly `BlockNumber::Latest` ? I believe that this at least should look to what user has specified in `to_block` and if that is `None` default to `Latest` or whatever the ethereum spec says?

core/lib/zksync_core/src/api_server/web3/namespaces/eth.rs
montekki: > MO, it would make sense to uniformly use either exclusive or inclusive cursor for all TypedFilter variants (I'd slightly prefer inclusive, but really any consistent option would do fine).

Yep

core/lib/zksync_core/src/api_server/web3/namespaces/eth.rs
Artemka374: Generally I think that it shouldn't be a problem, since we use it only when updating the `from_block` when we are actually comparing what is the current `to_block` and this one

core/lib/zksync_core/src/api_server/web3/namespaces/eth.rs
```diff
@@ -826,6 +826,10 @@ impl<G: L1GasPriceProvider> EthNamespace<G> {
                     .state
                     .resolve_filter_block_number(filter.to_block)
                     .await?;
+                let latest_block = self
```
slowli, line 829: - We have a similar TOCTOU issue here as with `to_block`; `latest_block` may get outdated before events are fetched. If we want a quick fix, I guess the initial PR version would do better (although, as we've discussed, it's not very efficient with DB queries). If we want to avoid TOCTOU, I think we'd need to rework getting events from DB so that it relies on specific block ranges; or we could issue all queries in a transaction. (Although the transaction-based solution is simpler, IMO, in this case simplifying DAL logic would be warranted.)
- As an optimization, we can avoid this query if `filter.to_block` is not set or set to "latest" (which I suspect would be a common occurrence).

core/lib/dal/src/events_web3_dal.rs
```diff
@@ -119,10 +121,10 @@ impl EventsWeb3Dal<'_, '_> {
 
         let mut where_sql = format!("(miniblock_number >= {})", filter.from_block.0 as i64);
 
-        if let Some(to_block) = filter.to_block {
-            let block_sql = web3_block_number_to_sql(to_block);
-            where_sql += &format!(" AND (miniblock_number <= {})", block_sql);
-        }
+        let block_sql =
+            web3_block_number_to_sql(BlockNumber::Number(U64::from(filter.to_block.as_u64())));
+        where_sql += &format!(" AND (miniblock_number <= {})", block_sql);
```
slowli, line 126: Can't you treat `filter.to_block` in the same way as `from_block` above? Conversion to `BlockNumber` looks redundant. Also, maybe off-topic, but it seems reasonable to make from / to numbers query arguments, instead of inlining them as numbers.

core/lib/zksync_core/src/api_server/web3/tests/mod.rs
```diff
@@ -472,20 +471,16 @@ impl HttpTest for LogFilterChangesWithBlockBoundaries {
...
-        let FilterChanges::Logs(upper_bound_logs) = upper_bound_logs else {
+        let FilterChanges::Hashes(upper_bound_logs) = upper_bound_logs else {
             panic!("Unexpected getFilterChanges output: {:?}", upper_bound_logs);
         };
-        assert!(upper_bound_logs
-            .iter()
-            .all(|log| log.block_number.unwrap() < 3.into()));
+        assert!(upper_bound_logs.is_empty());
 
         let lower_bound_logs = client.get_filter_changes(lower_bound_filter_id).await?;
```
slowli, line 485: Please simplify logic a bit here as well. Previously, `start_idx` would need to be defined to filter out old logs, but now it can be set to 0; i.e., you can call

```rust
assert_logs_match(&lower_bound_logs, &new_events);
```

core/lib/zksync_core/src/api_server/web3/tests/mod.rs
```diff
@@ -458,11 +458,10 @@ impl HttpTest for LogFilterChangesWithBlockBoundaries {
         };
         assert_logs_match(&lower_bound_logs, &new_events);
 
-        // FIXME(#546): update test after behavior is fixed
         let new_upper_bound_logs = client.get_filter_changes(upper_bound_filter_id).await?;
-        assert_eq!(new_upper_bound_logs, FilterChanges::Logs(upper_bound_logs));
+        assert_eq!(new_upper_bound_logs, FilterChanges::Hashes(vec![]));
```
slowli, line 462: Nit: you can use

```rust
assert_matches!(_, FilterChanges::Hashes(hashes) if hashes.is_empty());
```

here and in other cases where empty changes are returned.

core/lib/zksync_core/src/api_server/web3/namespaces/eth.rs
```diff
@@ -816,16 +828,26 @@ impl<G: L1GasPriceProvider> EthNamespace<G> {
                 } else {
                     vec![]
                 };
+
+                let mut to_block = self
+                    .state
+                    .resolve_filter_block_number(filter.to_block)
+                    .await?;
+
+                if let Some(BlockNumber::Number(_)) = filter.to_block {
```
slowli, line 837: Nit: You can use `if matches!(filter.to_block, Some(BlockNumber::Number(_)))`, since you don't bind any vars in the matched pattern.

core/lib/zksync_core/src/api_server/web3/namespaces/eth.rs
```diff
@@ -795,7 +800,12 @@ impl<G: L1GasPriceProvider> EthNamespace<G> {
                     )
                     .await
                     .map_err(|err| internal_error(METHOD_NAME, err))?;
-                *from_timestamp = last_timestamp.unwrap_or(*from_timestamp);
+
+                *from_timestamp = match last_timestamp {
```
slowli, line 804: - Nit: You could've left the old logic (i.e., `unwrap_or`); it's functionally equivalent to this one and is more concise.
- Bikeshedding: maybe, it makes sense to call this field `from_timestamp_excluded` (and perhaps changing `TypedFilter` variants so it used named fields) so that it's harder to make a mistake as to the field meaning?

slowli (dismissed):
Nice 👍

slowli (approved):

PR #550, opened 2023-11-27
feat: flag for prover setup running `zk init`

IAvecilla:
## What ❔

This PR skips plonk setup by default and adds a new flag for `init` command named `--prover-setup`

## Why ❔

The `init` command takes considerable time during the plonk setup step. To speed things up, it's beneficial to execute this step only when required. Utilizing the `--prover-setup` flag specifically triggers the plonk step, ensuring that the plonk keys are generated only when running the prover alongside the server.

## Checklist

### DIFF
--- a/infrastructure/zk/src/init.ts
+++ b/infrastructure/zk/src/init.ts
@@ -165,7 +165,7 @@ export interface InitArgs {
 const DEFAULT_ARGS: InitArgs = {
     skipSubmodulesCheckout: false,
     skipEnvSetup: false,
-    skipPlonkStep: false,
+    skipPlonkStep: true,
     governorPrivateKeyArgs: [],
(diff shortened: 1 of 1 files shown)


popzxc (commented):
I like this PR, but at the same time it now will be really easy to forget about the fact that such a thing/subcommand exists at all.

I guess at the bare minimum I would expect it to be covered in docs (e.g. something like section mentioning roughly "If you intend to work with prover-related stuff, you should also run this command, which will download the PLONK setup for you. It will take a while").
Ideally, it would also be nice to find places where PLONK setup is accessed, and if it's missing, the error/panic message should mention something like "If you're running this code locally, be sure to download PLONK setup via `zk` utility".

perekopskiy:
Closing as stale

PR #554, opened 2023-11-28
feat: Implemented 1 validator consensus for the main node

pompon0:
## What ❔

Implemented 1 validator consensus for the main node.
It is NOT enabled yet, corresponding configuration still needs to be provided.
It is responsible for populating consensus column in the Miniblocks table.
Consensus will be running asynchronously for now, not affecting the miniblock production at all.

Fixes BFT-388.

## Why ❔

Reaching a consensus with just 1 validator doesn't provide much value, however we want to start running the consensus code in production (in a "shadow" mode) to make sure it works as expected before we start relying on it.


core/lib/zksync_core/src/consensus/storage/mod.rs
```diff
@@ -0,0 +1,394 @@
...
+    let n = MiniblockNumber(
+        block
+            .header
+            .number
+            .0
+            .try_into()
+            .context("MiniblockNumber")
+            .map_err(StorageError::Database)?,
+    );
+    let mut txn = start_transaction(ctx, storage)
```
slowli, line 140: Just to understand better: Is there a risk here that a non-transactional update would result in a logical error, or is it just a best practice of wrapping related DB calls in a transaction?

core/lib/zksync_core/src/consensus/storage/mod.rs
```diff
@@ -0,0 +1,394 @@
...
+async fn start_transaction<'a, 'b, 'c: 'b>(
+    ctx: &ctx::Ctx,
+    storage: &'c mut StorageProcessor<'a>,
+) -> Result<StorageProcessor<'b>, StorageError> {
+    ctx.wait(storage.start_transaction())
+        .await?
+        .map_err(|err| StorageError::Database(err.into()))
+}
+
+async fn commit<'a>(ctx: &ctx::Ctx, txn: StorageProcessor<'a>) -> StorageResult<()> {
```
slowli, line 63: Nit: `'a` can be elided here and in `fetch_block` (i.e., you can write `txn: StorageProcessor<'_>`).

core/lib/zksync_core/src/consensus/wrap_error.rs
```diff
@@ -0,0 +1,22 @@
+use std::fmt::Display;
+
+pub trait WrapError: Sized {
+    fn wrap<C: Display + Send + Sync + 'static>(self, c: C) -> Self {
+        self.with_wrap(|| c)
+    }
+    fn with_wrap<C: Display + Send + Sync + 'static, F: FnOnce() -> C>(self, f: F) -> Self;
+}
+
+impl WrapError for anyhow::Error {
```
slowli, line 10: Nit: This implementation is used in 2 places (`impl WrapError for StorageError` and a single place in `put_block`), and can be replaced with `.context(_)`. IMO, it could make sense to specialize this extension for `StorageError` / `StorageResult` specifically and maybe rename methods to something like `db_context()` / `with_db_context()`.

core/lib/zksync_core/src/consensus/storage/mod.rs
```diff
@@ -0,0 +1,394 @@
...
+            genesis: genesis.header.number,
+            head: sync::watch::channel(head).0,
+            pool,
+        })
+    }
+}
+
+#[async_trait::async_trait]
+impl WriteBlockStore for SignedBlockStore {
+    /// Verify that `payload` is a correct proposal for the block `block_number`.
```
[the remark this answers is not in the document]
pompon0: yes, this is for single-validator only. multi-validator will require interaction with StateKeeper and will look totally different.

core/lib/zksync_core/src/consensus/wrap_error.rs
pompon0: I'd rather prefer to generalize it, so that WrapError can be implemented for an arbitrary error type which accumulates context in a nontrivial way. Also StorageError ~ Canceled + Internal, a type which IMO is as applicable as anyhow::Error, so I think it would be useful to move StorageError to concurrency crate and rename it.

pompon0: anyhow::Error needs some helper to support WrapError, because (as mentioned above) it doesn't implement anyhow::Context directly. This helper doesn't necessarily have to be WrapError implementation, but it would look quite consistent IMO.

core/lib/zksync_core/src/consensus/storage/mod.rs
```diff
@@ -0,0 +1,394 @@
...
+        .await?
+        .context("sync_block()")
+        .map_err(StorageError::Database)?
+    else {
+        return Ok(None);
+    };
+    Ok(Some(sync_block.try_into().map_err(StorageError::Database)?))
+}
+
+async fn put_block<'a>(
```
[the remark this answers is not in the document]
pompon0: yes, I suppose so. Given that the high level logic is quite different for these modules I haven't paid any attention to deduplicate helper functions so far.

core/lib/zksync_core/src/consensus/storage/mod.rs
```diff
@@ -0,0 +1,394 @@
...
+        .await
+        .wrap("fetch_block()")?
+        .context("head not found")
+        .map_err(StorageError::Database)?;
+    while let Some(next) = fetch_block(ctx, storage, block.header.number.next())
+        .await
+        .wrap("fetch_block()")?
+    {
+        block = next;
+    }
```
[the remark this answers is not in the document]
pompon0: find_head_number takes time linear to the number of miniblocks without consensus field set, which can get quite large, especially initially when we start backfilling starting from a fixed genesis block. Note that we have a polling loop with this query, so it can get quite expensive. Sequential lookups on the other side will have amortized cost.

core/lib/zksync_core/src/consensus/storage/mod.rs
pompon0: it is a read-write transaction and I don't want to rely on external assumptions about the storage access patterns if I don't have to.

core/lib/zksync_core/src/consensus/storage/mod.rs
```diff
@@ -0,0 +1,394 @@
...
+async fn storage<'a>(
+    ctx: &ctx::Ctx,
+    pool: &'a ConnectionPool,
+) -> StorageResult<StorageProcessor<'a>> {
+    ctx.wait(pool.access_storage_tagged("sync_layer"))
+        .await?
+        .map_err(StorageError::Database)
+}
+
+async fn start_transaction<'a, 'b, 'c: 'b>(
```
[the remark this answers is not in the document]
pompon0: done

core/lib/zksync_core/src/consensus/wrap_error.rs
```diff
@@ -0,0 +1,22 @@
...
+    fn with_wrap<C: Display + Send + Sync + 'static, F: FnOnce() -> C>(self, f: F) -> Self {
+        anyhow::Context::with_context(Err::<(), _>(anyhow::Error::from(self)), f)
+            .err()
+            .unwrap()
```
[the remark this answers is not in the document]
slowli: It has inherent [`context()` method](https://docs.rs/anyhow/latest/anyhow/struct.Error.html#method.context).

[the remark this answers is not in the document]
pompon0: thx, somehow I missed it

core/lib/zksync_core/src/consensus/wrap_error.rs
pompon0: my proposal: https://github.com/matter-labs/era-consensus/pull/43

core/lib/zksync_core/src/consensus/storage.rs
```diff
@@ -0,0 +1,423 @@
...
+    ) -> ctx::Result<()> {
+        tracing::error!("SignedBlockStore::verify_payload({block_number:?})");
+        let storage = &mut storage(ctx, &self.pool).await.wrap("storage()")?;
+        let want = storage
+            .fetch_payload(ctx, block_number, self.operator_address)
+            .await
+            .wrap("fetch_payload()")?
+            .context("unknown block")?;
+        if payload != &want.encode() {
+            return Err(anyhow::anyhow!("unexpected payload").into());
```
slowli, line 262: Nit: It could make sense to include `payload` / wanted payload into the error message for debugging purposes, although it would make to parse `payload` instead of encoding `want` in this case.

core/lib/zksync_core/src/consensus/storage.rs
```diff
@@ -0,0 +1,423 @@
...
+            .await
+            .wrap("fetch_block()")
+    }
+
+    async fn missing_block_numbers(
+        &self,
+        _ctx: &ctx::Ctx,
+        _range: ops::Range<validator::BlockNumber>,
+    ) -> ctx::Result<Vec<validator::BlockNumber>> {
+        Ok(vec![]) // The storage never has missing blocks by construction
```
slowli, line 340: Note that I've fixed this place for the EN block store. Although in practice we won't call `missing_block_numbers` with args that would lead to a non-empty response, this is non-local reasoning, so I've found it too shaky.

core/lib/zksync_core/src/consensus/storage.rs
```diff
@@ -0,0 +1,423 @@
...
+        Ok(ctx
+            .wait(storage.0.consensus_dal().put_replica_state(replica_state))
+            .await?
+            .context("put_replica_state()")?)
+    }
+}
+
+#[async_trait::async_trait]
+impl PayloadSource for SignedBlockStore {
+    async fn propose(
```
slowli, line 373: Dumb question: Does the consensus code expect that this method can take an extended period of time to complete? I.e., could this block other logic from executing / completing?

core/lib/zksync_core/src/consensus/testonly.rs
```diff
@@ -0,0 +1,291 @@
...
+                .fetch_block(ctx, validator::BlockNumber(i), self.operator_address)
+                .await?
+                .with_context(|| format!("missing block {i}"))?;
+            block.validate(validators, 1).unwrap();
+        }
+        Ok(())
+    }
+}
+
+async fn run_metadata_calculator(ctx: &ctx::Ctx, pool: ConnectionPool) -> anyhow::Result<()> {
```
slowli, line 221: Nit: IMO, `run_mock_metadata_calculator` would be a better name; otherwise, one could get an idea that a real `MetadataCalculator` should be run here.

core/lib/zksync_core/src/consensus/testonly.rs
```diff
@@ -0,0 +1,291 @@
...
+            s.spawn_bg(
+                ZkSyncStateKeeper::without_sealer(
+                    stop_receiver,
+                    Box::new(io),
+                    Box::new(MockBatchExecutorBuilder),
+                )
+                .run(),
+            );
+            ctx.canceled().await;
+            let _ = stop_sender.send(true);
```
slowli, line 286: Nit: You can use `send_replace` instead of `send` to signal that you allow for no receivers to be present.

core/lib/zksync_core/src/consensus/tests.rs
```diff
@@ -0,0 +1,86 @@
...
+#[tokio::test(flavor = "multi_thread")]
+async fn test_backfill() {
+    zksync_concurrency::testonly::abort_on_panic();
+    let ctx = &ctx::test_root(&ctx::RealClock);
+    let rng = &mut ctx.rng();
+    let pool = ConnectionPool::test_pool().await;
+
+    scope::run!(ctx, |ctx, s| async {
+        // Start state keeper.
+        const OPERATOR_ADDRESS: Address = Address::repeat_byte(17);
```
slowli, line 26: Bikeshedding: IMO, `const`s in the middle of the function body look weird; it's a partial case of [a Clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#/items_after_statements) that we don't have enabled.

core/lib/zksync_core/src/consensus/tests.rs
```diff
@@ -0,0 +1,86 @@
...
+
+        // Prepare genesis block for consensus.
+        const GENESIS_BLOCK: validator::BlockNumber = validator::BlockNumber(5);
+        let genesis_payload = {
+            let mut storage = storage::storage(ctx, &pool).await.unwrap();
+            storage
+                .fetch_payload(ctx, GENESIS_BLOCK, OPERATOR_ADDRESS)
+                .await
+                .unwrap()
+                .unwrap()
```
slowli, line 42: Dumb question: Can't errors be propagated here?

core/lib/zksync_core/src/consensus/tests.rs
slowli, line 1: Good job with testing 👍

core/lib/zksync_core/src/consensus/testonly.rs
```diff
@@ -0,0 +1,291 @@
...
+use zksync_concurrency::{ctx, scope, sync, time};
+use zksync_consensus_roles::validator;
+use zksync_contracts::SystemContractCode;
+use zksync_dal::ConnectionPool;
+use zksync_types::{
+    api, Address, L1BatchNumber, L2ChainId, MiniblockNumber, ProtocolVersionId, H256,
+};
+
+#[derive(Debug)]
+struct MockMainNodeClient;
```
slowli, line 21: Nit: You can probably reuse `MockMainNodeClient` from `sync_layer::tests` here; w/o generating blocks for it, it effectively works in the same way.

slowli (commented):
I think there's an opportunity to share some main / test code with the EN-related consensus logic. I can do this in a follow-up PR after this one is merged.

core/lib/zksync_core/src/consensus/storage.rs
```diff
@@ -0,0 +1,423 @@
...
+        Ok(Some(sync_block.try_into()?))
+    }
+
+    pub async fn put_block(
+        &mut self,
+        ctx: &ctx::Ctx,
+        block: &validator::FinalBlock,
+        operator_address: Address,
+    ) -> ctx::Result<()> {
+        tracing::error!("put_block({:?})", block.header.number);
```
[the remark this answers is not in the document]
pompon0: yeah, that's from debugging, sry

core/lib/zksync_core/src/consensus/storage.rs
```diff
@@ -0,0 +1,423 @@
...
+        }
+
+        // Verify that the payload matches the storage.
+        let want_payload: consensus::Payload = sync_block.try_into()?;
+        if want_payload.encode() != block.payload {
+            let got_payload = consensus::Payload::decode(&block.payload)?;
+            return Err(anyhow::anyhow!(
+                "payload mismatch: got {got_payload:?}, want {want_payload:?}"
+            )
+            .into());
```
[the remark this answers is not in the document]
pompon0: I'm not a big fan of macros and this explicit error-handling pattern feels ok to me. However if it starts to annoy  people at scale, we can implement that. In fact it should be possible to generalize anyhow::bail to any error type which implements `From<anyhow::Error>`.

core/lib/zksync_core/src/consensus/storage.rs
```diff
@@ -0,0 +1,423 @@
...
+            .await?
+            .context("sync_block()")?
+            .context("unknown block")?;
+        let want = &ConsensusBlockFields {
+            parent: block.header.parent,
+            justification: block.justification.clone(),
+        };
+
+        // Early exit if consensus field is already set to the expected value.
+        if Some(want.encode()) == sync_block.consensus {
```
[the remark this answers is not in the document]
pompon0: I was thinking about that and I haven't reached a strong conclusion, I guess that we may want to stay on the safe side for now and forbid overwriting it altogether.

core/lib/zksync_core/src/consensus/storage.rs
```diff
@@ -0,0 +1,423 @@
...
+    async fn put_block(&self, ctx: &ctx::Ctx, block: &validator::FinalBlock) -> ctx::Result<()> {
+        tracing::error!("SignedBlockStore::put_block()");
+        let storage = &mut storage(ctx, &self.pool).await.wrap("storage()")?;
+
+        // Currently main node is the only validator, so it should be the only one creating new
+        // blocks. To ensure that no gaps in the blocks are created we check here that we always
+        // insert a new head.
+        let head = *self.head.borrow();
+        let head = storage
+            .find_head_forward(ctx, head, self.operator_address)
```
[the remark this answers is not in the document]
pompon0: sure, I can add it

core/lib/zksync_core/src/consensus/storage.rs
```diff
@@ -0,0 +1,423 @@
...
+            .await
+            .wrap("put_block()")
+    }
+}
+
+#[async_trait::async_trait]
+impl BlockStore for SignedBlockStore {
+    async fn head_block(&self, ctx: &ctx::Ctx) -> ctx::Result<validator::FinalBlock> {
+        let storage = &mut storage(ctx, &self.pool).await.wrap("storage()")?;
+        let head = *self.head.borrow();
```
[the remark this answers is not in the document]
pompon0: I decided to keep the semantics simple and delegate "head updates" solely to run_background_tasks. This will prevent not noticing a hard-to-debug situation in which one forgets to call run_background_tasks and all methods keep working efficiently, except for subscribe() which works usually-unless-we-stop-calling-other-methods.

core/lib/zksync_core/src/consensus/storage.rs
pompon0: Good point. propose() may be blocking in case there is nothing to propose for a while(no txs in mempool for example). Still it shouldn't just give up proposing right away, but rather wait, timing out only if the consensus moves to another view already. With the current implementation it will block both replica and the leader actors from processing messages. Replica definitely shouldn't be blocked by that. Leader should be fine in most cases with being blocked, but we should analyze that more thoroughly.

core/lib/zksync_core/src/consensus/storage.rs
```diff
@@ -0,0 +1,423 @@
...
+        operator_address: Address,
+    ) -> ctx::Result<Option<consensus::Payload>> {
+        let n = MiniblockNumber(block_number.0.try_into().context("MiniblockNumber")?);
+        let Some(sync_block) = ctx
+            .wait(self.0.sync_dal().sync_block(n, operator_address, true))
+            .await?
+            .context("sync_block()")?
+        else {
+            return Ok(None);
+        };
```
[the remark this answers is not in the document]
pompon0: done

core/lib/zksync_core/src/consensus/storage.rs
pompon0: done

core/lib/zksync_core/src/consensus/testonly.rs
pompon0: done

core/lib/zksync_core/src/consensus/testonly.rs
pompon0: done

core/lib/zksync_core/src/consensus/tests.rs
pompon0: done

core/lib/zksync_core/src/consensus/testonly.rs
pompon0: done

core/lib/zksync_core/src/consensus/storage.rs
pompon0: done

core/lib/zksync_core/src/consensus/mod.rs
```diff
@@ -1,6 +1,62 @@
...
+    pub node_key: node::SecretKey,
+    pub validator_key: validator::SecretKey,
+    pub operator_address: Address,
+}
 
-pub(crate) use self::payload::Payload;
+impl Config {
+    #[allow(dead_code)]
+    pub async fn run(self, ctx: &ctx::Ctx, pool: ConnectionPool) -> anyhow::Result<()> {
+        assert_eq!(
```
slowli, line 33: Nit: Given that we propagate errors here, should this use `anyhow::ensure!` instead?

core/lib/zksync_core/src/consensus/mod.rs
```diff
@@ -1,6 +1,62 @@
...
+mod storage;
+
+#[cfg(test)]
+pub(crate) mod testonly;
+#[cfg(test)]
+mod tests;
+
+pub(crate) use self::{payload::Payload, storage::sync_block_to_consensus_block};
+
+pub struct Config {
```
slowli, line 22: Nit: Please derive `Debug` as we've discussed.

core/lib/zksync_core/src/consensus/tests.rs
```diff
@@ -0,0 +1,88 @@
...
+        // Start state keeper.
+        let (mut sk, sk_runner) = testonly::StateKeeperHandle::new(OPERATOR_ADDRESS);
+        s.spawn_bg(sk_runner.run(ctx, &pool));
+
+        // Populate storage with a bunch of blocks.
+        sk.push_random_blocks(rng, 10).await;
+        sk.sync(ctx, &pool).await.context("sk.sync(<1st phase>)")?;
+
+        // Prepare genesis block for consensus.
+        const GENESIS_BLOCK: validator::BlockNumber = validator::BlockNumber(5);
```
slowli, line 37: Nit: Please move this constant next to `OPERATOR_ADDRESS`, as we've discussed.

core/lib/zksync_core/src/sync_layer/gossip/buffered/mod.rs
```diff
@@ -319,16 +315,32 @@ impl<T: ContiguousBlockStore> BlockStore for Buffered<T> {
...
+        &self,
+        _ctx: &ctx::Ctx,
+        _block_number: BlockNumber,
+        _payload: &Payload,
+    ) -> ctx::Result<()> {
+        // This is storage for non-validator nodes (aka full nodes),
+        // so verify_payload() won't be called.
+        // Still, it probably would be better to either
+        // * move verify_payload() to BlockStore, so that Buffered can just forward the call
+        // * create another separate trait for verify_payload.
```
slowli, line 329: IMO, it would be better to implement it as a separate trait (subtrait of `WriteBlockStore`, maybe) since this functionality indeed seems specific to validator nodes. OTOH, this could perform sanity checks and the hypothetical upstream caller in the consensus library would panic on an error (which would mean that the node cannot interpret the received block, or that the supermajority of validators has gone malicious).

core/lib/zksync_core/Cargo.toml
```diff
@@ -40,11 +40,12 @@ vlog = { path = "../vlog" }
...
-zksync_consensus_executor = { version = "0.1.0", git = "https://github.com/matter-labs/era-consensus.git", rev = "da015d4c94b19962bc11622b6cc256e214256555" }
-zksync_protobuf = { version = "0.1.0", git = "https://github.com/matter-labs/era-consensus.git", rev = "da015d4c94b19962bc11622b6cc256e214256555" }
+zksync_concurrency = { version = "0.1.0", git = "https://github.com/matter-labs/era-consensus.git", rev = "b06ae4b144719d8ffef0067a4b54a266df0dff35" }
+zksync_consensus_roles = { version = "0.1.0", git = "https://github.com/matter-labs/era-consensus.git", rev = "b06ae4b144719d8ffef0067a4b54a266df0dff35" }
+zksync_consensus_storage = { version = "0.1.0", git = "https://github.com/matter-labs/era-consensus.git", rev = "b06ae4b144719d8ffef0067a4b54a266df0dff35" }
+zksync_consensus_executor = { version = "0.1.0", git = "https://github.com/matter-labs/era-consensus.git", rev = "b06ae4b144719d8ffef0067a4b54a266df0dff35" }
+zksync_consensus_bft = { version = "0.1.0", git = "https://github.com/matter-labs/era-consensus.git", rev = "b06ae4b144719d8ffef0067a4b54a266df0dff35" }
```
slowli, line 47: Nit: Could it make sense to re-export the relevant items (i.e., `PayloadSource` for now) from the `executor` crate? From the usability perspective, it'd make sense to not require an additional direct dependency just for a single item.

slowli (approved):

core/lib/dal/src/blocks_dal.rs
```diff
@@ -396,6 +396,30 @@ impl BlocksDal<'_, '_> {
...
+    pub async fn get_last_miniblock_number_with_consensus_fields(
+        &mut self,
+    ) -> anyhow::Result<Option<MiniblockNumber>> {
+        let Some(row) = sqlx::query!("SELECT number FROM miniblocks WHERE consensus IS NOT NULL ORDER BY number DESC LIMIT 1")
+            .fetch_optional(self.storage.conn())
+            .await? else { return Ok(None) };
+        Ok(Some(MiniblockNumber(row.number.try_into()?)))
+    }
+
+    /// Fetches the number of the next miniblock with consensus field
```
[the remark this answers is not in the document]
pompon0: done

core/lib/zksync_core/src/consensus/mod.rs
pompon0: done

core/lib/zksync_core/src/consensus/mod.rs
pompon0: done

core/lib/zksync_core/src/consensus/tests.rs
pompon0: done

core/lib/zksync_core/src/sync_layer/gossip/buffered/mod.rs
pompon0: ack

core/lib/zksync_core/src/sync_layer/tests.rs
```diff
@@ -1,23 +1,21 @@
 //! High-level sync layer tests.
-
```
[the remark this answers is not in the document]
pompon0: reverted

core/lib/zksync_core/Cargo.toml
pompon0: ack

core/lib/zksync_core/src/consensus/storage.rs
```diff
@@ -0,0 +1,425 @@
...
+        Ok(())
+    }
+
+    /// Puts a block into this storage.
+    async fn put_block(&self, ctx: &ctx::Ctx, block: &validator::FinalBlock) -> ctx::Result<()> {
+        let storage = &mut storage(ctx, &self.pool).await.wrap("storage()")?;
+
+        // Currently main node is the only validator, so it should be the only one creating new
+        // blocks. To ensure that no gaps in the blocks are created we check here that we always
+        // insert a new head.
```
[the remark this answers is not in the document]
pompon0: done

brunoffranca (dismissed):

brunoffranca (approved):

brunoffranca (approved):

PR #568, opened 2023-11-30
fix: Sync protocol version between consensus and server blocks

slowli:
## What ❔

Aligns the protocol version for consensus blocks with that of `SyncBlock`s.

## Why ❔

Required for gossip-based block syncing to work correctly.

## Checklist


core/lib/zksync_core/src/sync_layer/gossip/conversions.rs
```diff
@@ -36,11 +44,17 @@ impl FetchedBlock {
         let payload = consensus::Payload::decode(&block.payload)
             .context("Failed deserializing block payload")?;
 
+        let protocol_version = block.justification.message.protocol_version;
+        let protocol_version =
+            u16::try_from(protocol_version.as_u32()).context("Protocol version overflow")?;
```
pompon0, line 49: nit: u16::try_from error message already includes information about the overflow: https://doc.rust-lang.org/src/core/num/error.rs.html#24

pompon0 (dismissed):

core/lib/zksync_core/src/sync_layer/gossip/conversions.rs
pompon0: >  I don't want to lose PR approvals

I feel your motivation here, but I still hate it :P

slowli: Since the PR has got merge conflicts, I'll lose approvals anyway, so I've fixed it 👍

PR #585, opened 2023-12-02
chore: update bytecode_compression.md

leonarddt05:
My suggestions regarding this section are as follows:

    1. contracts use in our chain must be copied into L1 (so that the chain can be reconstructed from L1 if needed). - It should be 'contracts used in our chain'.
       
    2. Looping over the encoded data, derived from _rawCompressedData, the corresponding chunks are pulled from the dictionary and compared to the original byte code, reverting if there is a mismatch. - Better phrasing: 'Looping over the encoded data, derived from _rawCompressedData, the corresponding chunks are retrieved from the dictionary and compared to the original bytecode. If there is a mismatch, the operation reverts'.
       
    3. the latter of which comes from the output of the server’s compression algorithm. - Better phrasing: 'the latter of which comes from the server’s compression algorithm output.'

## Checklist


RomanBrodetski:
for that you'll need a local dev environment set (see https://github.com/matter-labs/zksync-era/blob/main/docs/guides/setup-dev.md). Then just run `zk fmt` and commit the changes! Thanks.

perekopskiy:
Closing as stale

PR #611, opened 2023-12-05
perf(external-node): Use async miniblock sealing in external IO

perekopskiy:
## What ❔

External IO uses async miniblock sealing. 

## Why ❔

Execution of transactions and miniblock sealing (writing data to postgres) happen in parallel so the perfomance is better. 

## Checklist


core/lib/zksync_core/src/sync_layer/external_io.rs
```diff
@@ -459,10 +463,9 @@ impl StateKeeperIO for ExternalIO {
...
         };
 
+        let store_latency = L1_BATCH_METRICS.start_storing_on_en();
         let mut storage = self.pool.access_storage_tagged("sync_layer").await.unwrap();
         let mut transaction = storage.start_transaction().await.unwrap();
-
-        let store_latency = L1_BATCH_METRICS.start_storing_on_en();
         // We don't store the transactions in the database until they're executed to not overcomplicate the state
         // recovery on restart. So we have to store them here.
         for tx in &updates_manager.miniblock.executed_transactions {
```
perekopskiy, line 471: EN currently saves transactions to DB in a sync manner (see lines https://github.com/matter-labs/zksync-era/pull/611/files#diff-31a0b1d8722a196212b88c601a2d540ee4be28a2bf32c534885e2910784c44baR471-R494). Options we have:
- Leave it as is, this part was taking only 40ms during today's high TPS period, 40ms/1000s per miniblock is pretty negligible. I prefer this approach
- Add a flag to `MiniblockSealCommand`, so it will insert transactions to DB if flag is true. Flag will be set to true for EN and false for main node. This a good option if saving will start taking more than 40ms.
- Change logic of `mark_txs_as_executed_in_miniblock`: currently it handles txs replacement with `UPDATE transactions` and we can change that to `INSERT .. ON CONFLICT DO UPDATE`, then we won't need to save transactions separately for EN. I don't really like it because on main node we will always fall into `ON CONFLICT` branch wasting time on trying to insert.

slowli: Are you sure that we're not messing with EN correctness by splitting saving transactions and the miniblock data into separate transactions? The comment above implies that there may be complications with the state recovery on restart if transactions are saved in the DB before their execution; and if the EN crashes after saving transactions and before sealing a miniblock, that's what we'll end up with. IMO, adding a flag to `MiniblockSealCommand` to save transactions would be a safer option, and it (marginally) improves performance as well.

core/lib/zksync_core/src/sync_layer/tests.rs
```diff
@@ -17,6 +17,7 @@ use zksync_types::{
 };
 
 use super::{fetcher::FetcherCursor, sync_action::SyncAction, *};
+use crate::state_keeper::MiniblockSealer;
```
slowli, line 20: Nit: You have the `crate::state_keeper` import "path" below.

core/lib/zksync_core/src/sync_layer/tests.rs
```diff
@@ -156,7 +157,11 @@ impl StateKeeperHandles {
         ensure_genesis(&mut pool.access_storage().await.unwrap()).await;
 
         let sync_state = SyncState::new();
+        let (miniblock_sealer, miniblock_sealer_handle) = MiniblockSealer::new(pool.clone(), 0);
```
slowli, line 160: Do the tests work with non-zero capacity (e.g., 1)? If they do, I'd prefer to use it, so that the tests more closely match real conditions.

core/lib/zksync_core/src/sync_layer/external_io.rs
perekopskiy: I'm almost sure it will work. We do `INSERT .. ON CONFLICT DO NOTHING` and having those transactions in DB earlier shouldn't affect anything.
Though, I thought about it one more time and would prefer to be future-proof performance wise by adding flag to seal command. Will add this flag

PR #619, opened 2023-12-06
chore: remove old witness generator

AnastasiiaVashchuk:


core/bin/zksync_server/src/main.rs
```diff
@@ -156,10 +156,7 @@ async fn main() -> anyhow::Result<()> {
 
     // OneShotWitnessGenerator is the only component that is not expected to run indefinitely
     // if this value is `false`, we expect all components to run indefinitely: we panic if any component returns.
-    let is_only_oneshot_witness_generator_task = matches!(
-        components.as_slice(),
-        [Component::WitnessGenerator(Some(_), _)]
-    );
+    let is_only_oneshot_witness_generator_task = false;
```
AnastasiiaVashchuk, line 159: this var is passed further to the `initialize_components` function as `use_prometheus_push_gateway` argument. This is very unclear to me and I do not understand how they are connected.

core/lib/zksync_core/src/metadata_calculator/updater.rs
```diff
@@ -207,10 +207,6 @@ impl TreeUpdater {
                             .await;
                     }
                 }
-                prover_storage
-                    .witness_generator_dal()
-                    .save_witness_inputs(l1_batch_number, object_key, protocol_version_id)
-                    .await;
```
slowli, line 213: Just to understand the remaining use of `prover_storage` in `MetadataCalculator`: It seems that it's only used to sync protocol versions via `.protocol_versions_dal().save_prover_protocol_version(_)`. Is this use still valid? AFAICT, there's a separate `save_prover_protocol_version()` for the FRI prover, and it's appropriately called from the prover system.

core/bin/zksync_server/src/main.rs
slowli: AFAIU, oneshot witgen is the only configuration that makes the server stop on its own after some time; i.e., it's not supposed to run infinitely as a server. That's why metrics are pushed to the gateway instead of being pulled. IMO, it'd make sense to simplify `initialize_components` and this function to "inline" this value / get rid of unnecessary args.

slowli (commented):
You can remove `zksync_verification_key_generator_and_server` dependency in `zksync_core` manifest as obsolete. Logically, `zksync_prover_utils` and `zksync_queued_job_processor` would be removed as well, but they are used by house keeper and BWIP components for now.

core/lib/zksync_core/src/metadata_calculator/updater.rs
AnastasiiaVashchuk: cc @EmilLuta

slowli (approved):

PR #640, opened 2023-12-08
chore(CI): Speeding up docker builds in CI

mm-zk:
## What ❔

* Removed zk contract compilation that was run before prover docker builds
* Removed docker builds for old prover and old circuit_synthesizer
* Downloading CRS file only for the snark wrapper / compressor job

## Why ❔

* to speed up docker CI that runs on every PR

### DIFF
--- a/docker/proof-fri-compressor/Dockerfile
+++ b/docker/proof-fri-compressor/Dockerfile
@@ -19,7 +19,7 @@ RUN curl https://sh.rustup.rs -sSf | bash -s -- -y && \
 WORKDIR /usr/src/zksync
 COPY . .
 
-RUN cargo build --release --bin proof_fri_compressor
+RUN cargo build --release --bin zksync_proof_fri_compressor
 
 FROM debian:bookworm-slim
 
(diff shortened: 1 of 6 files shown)


docker/proof-fri-compressor/Dockerfile
```diff
@@ -19,7 +19,7 @@ RUN curl https://sh.rustup.rs -sSf | bash -s -- -y && \
 WORKDIR /usr/src/zksync
 COPY . .
 
-RUN cargo build --release
+RUN cargo build --release --bin proof_fri_compressor
```
popzxc, line 22: Probably won't really improve stuff right now, since we [generate a bogus workspace in CI for prover images](https://github.com/matter-labs/zksync-era/blob/b0721f0641861fe879fd79c00dcd614497f6df35/.github/workflows/build-prover-template.yml#L163) that only contains one crate, but still a nice addition for the future.

popzxc (approved):

docker/proof-fri-compressor/Dockerfile
mm-zk: ah, so that's how it works..

as I kept wondering why we don't change the directory to 'prover'..

PR #642, opened 2023-12-08
ci(local-node): replace deprecated node setup

haraldh:
## What ❔

As in commit 33174aa59 and `docker/zk-environment/Dockerfile` use the repo to setup node, instead of the `curl | sh`.

## Why ❔
The old method is deprecated and it saves one minute waiting time instead of the deprecation banner being displayed.

## Checklist


haraldh:
rebased on current main

PR #655, opened 2023-12-11
chore(en): adds version metric collecting

montekki:
## Checklist


core/bin/external_node/src/main.rs
```diff
@@ -117,6 +130,23 @@ async fn init_tasks(
...
     task_handles.push(tokio::spawn(miniblock_sealer.run()));
+    let pool = connection_pool.clone();
+    task_handles.push(tokio::spawn(async move {
+        loop {
+            let protocol_version = pool
+                .access_storage()
+                .await
+                .unwrap()
+                .protocol_versions_dal()
+                .last_version_id()
```
perekopskiy, line 141: For main node `SELECT MAX(id) FROM protocol_versions != SELECT protocol_version FROM l1_batches ORDER BY number DESC LIMIT 1` as upgrades are scheduled so there can be a version in `protocol_version` that will be used in future. Though, for EN we currently don't populate `protocol_versions` in advance, I don't think we should assume it here.
i.e. I propose to change `SELECT MAX(id) FROM protocol_versions` to ` SELECT protocol_version FROM l1_batches ORDER BY number DESC LIMIT 1` here

core/bin/external_node/src/main.rs
```diff
@@ -117,6 +131,23 @@ async fn init_tasks(
...
+    task_handles.push(tokio::spawn(async move {
+        loop {
+            let protocol_version = pool
+                .access_storage()
+                .await
+                .unwrap()
+                .protocol_versions_dal()
+                .last_used_verion_id()
+                .await
+                .unwrap_or(ProtocolVersionId::Version0);
```
perekopskiy, line 144: Would be cleaner if you report metric only if value is some instead or `unwrap_or `

core/bin/external_node/src/main.rs
```diff
@@ -34,9 +35,14 @@ use zksync_dal::{healthcheck::ConnectionPoolHealthCheck, ConnectionPool};
...
 use zksync_state::PostgresStorageCaches;
 use zksync_storage::RocksDB;
+use zksync_types::ProtocolVersionId;
 use zksync_utils::wait_for_tasks::wait_for_tasks;
 
 mod config;
+mod metrics;
+
+const RELEASE_MANIFEST: &str =
+    std::include_str!("../../../../.github/release-please/manifest.json");
```
perekopskiy, line 45: This most probably is not included in the docker image, so should be included to work in the current state. Also doesn't look good to depend on some file in .github folder in the code, but I don't really have any alternatives

montekki: `include_str` is a compile-time thing.

montekki: This is ugly but this is because our versioning in Cargo.toml does not match the actual versions in that manifest, so unless we fix versioning this is going to be this way.

core/bin/external_node/src/main.rs
montekki: The node may be not syncing and have `0` batches and then it would not report anything though

perekopskiy (approved):

PR #656, opened 2023-12-11
fix: Follow up metrics fixes vol.2

Artemka374:
## What ❔

Fix oldest not generated batch query

## Why ❔

For metrics to work correctly

## Checklist

### DIFF
--- a/core/lib/dal/src/proof_generation_dal.rs
+++ b/core/lib/dal/src/proof_generation_dal.rs
@@ -109,7 +109,7 @@ impl ProofGenerationDal<'_, '_> {
         .ok_or(sqlx::Error::RowNotFound)
     }
 
-    pub async fn get_oldest_unprocessed_batch(&mut self) -> Option<L1BatchNumber> {
+    pub async fn get_oldest_unpicked_batch(&mut self) -> Option<L1BatchNumber> {
         let result: Option<L1BatchNumber> = sqlx::query!(
             "SELECT l1_batch_number \
              FROM proof_generation_details \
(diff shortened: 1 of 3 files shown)


core/lib/dal/src/proof_generation_dal.rs
```diff
@@ -129,7 +129,7 @@ impl ProofGenerationDal<'_, '_> {
         let result: Option<L1BatchNumber> = sqlx::query!(
             "SELECT l1_batch_number \
              FROM proof_generation_details \
-             WHERE status <> 'generated' \
+             WHERE status = 'picked_by_prover' \
```
RomanBrodetski, line 132: what if we don't have any rows that are "picked by prover" for some reason? Same applies to the query above.

RomanBrodetski: anyway, currently this and the above methods have confusing names and it's not clear for me what exactly they report. Why does oldest-not-generated-batch has to be `picked_by_prover`? What if the prover is down, and we don't move statuses to `picked_by_prover` at all?

I was thinking the previous query was correct - can you please explain why it's not? Only because `skipped`? If so, we can do `NOT IN`

Artemka374: Yes, agree about names, and yes, it is not correct because of `skipped`. Talking about `picked_by_prover` status, I guess I didn't really thought about it, it will return incorrect values if the scenario you described will occur.

Artemka374: But I don't think that the same thing can be applied to query above

RomanBrodetski (approved):

PR #658, opened 2023-12-11
fix: removing sqlx check from pre-push

tomg10:
Removing this check for now, I will re-add is as an opt-in check, probably with other checks bundled in. It's best not to require built project on every push.


RomanBrodetski:
@tomg10 can you please add some description to the PR?  Why are we removing it?

RomanBrodetski:
> It's best not to require built project on every push.


agreed. Used to do --no-verify but that skips all the checks

montekki (approved):

PR #671, opened 2023-12-12
chore: remove witgen related tables

AnastasiiaVashchuk:


core/lib/zksync_core/src/house_keeper/blocks_state_reporter.rs
```diff
@@ -36,13 +36,6 @@ impl L1BatchMetricsReporter {
                     .unwrap(),
                 BlockStage::MetadataCalculated,
             ),
-            (
-                conn.blocks_dal()
-                    .get_last_l1_batch_number_with_witness_inputs()
-                    .await
-                    .unwrap(),
-                BlockStage::MerkleProofCalculated,
-            ),
```
AnastasiiaVashchuk, line 45: I have checked, and I did not see its usage here(where its supposed to be) - https://grafana.matterlabs.dev/d/0Y8DUff4k/era-general?orgId=1&refresh=1m&from=now-30d&to=now&viewPanel=63
So I thought its safe to remove this.

core/lib/zksync_core/src/eth_sender/aggregator.rs
```diff
@@ -259,10 +259,7 @@ impl Aggregator {
         }
         let proofs = match proof_loading_mode {
             ProofLoadingMode::OldProofFromDb => {
-                prover_storage
-                    .prover_dal()
-                    .get_final_proofs_for_blocks(batch_to_prove, batch_to_prove)
-                    .await
+                unreachable!("OldProofFromDb is not supported anymore")
```
AnastasiiaVashchuk, line 262: https://linear.app/matterlabs/issue/PLA-694/remove-proofloadingmode-from-zksync-era-codebase

core/lib/dal/migrations/20231208134254_drop_old_witness_generation_related_tables.down.sql
```diff
@@ -0,0 +1,83 @@
...
+    time_taken TIME DEFAULT '00:00:00'::TIME WITHOUT TIME ZONE NOT NULL,
+    processing_started_at TIMESTAMP,
+    error VARCHAR,
+    attempts INTEGER DEFAULT 0 NOT NULL,
+    merkel_tree_paths_blob_url TEXT,
+    is_blob_cleaned boolean DEFAULT false NOT NULL,
+    protocol_version INTEGER
+    CONSTRAINT witness_inputs_prover_protocol_version_fkey REFERENCES prover_protocol_versions
+);
+CREATE INDEX IF NOT EXISTS witness_inputs_blob_cleanup_status_index ON witness_inputs (status, is_blob_cleaned);
```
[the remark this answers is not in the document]
RomanBrodetski: not sure, but I think we'll never be able to roll back such migrations. We could even consider leaving it empty or adding some exception so that it's clear that we cannot roll back. Even if we make this migration work, I think most of the others are broken/not compatible

core/lib/zksync_core/src/house_keeper/blocks_state_reporter.rs
slowli: The value of this metric could be viewed by editing PromQL for the dashboard (rn, it explicitly lists stages that should be included, and `merkle_proof_calculated` is not one of them). The metric looks stagnant on testnet / mainnet (e.g., it's 61,729 on the mainnet while the current L1 batch is approx. 344,000), which I guess confirms that it's outdated. I must warn that using `git blame` in this case is not the best option; I've just refactored metrics in the codebase to use the `vise` façade 🙃

slowli (approved):

PR #676, opened 2023-12-13
feat(dal): Make ConnectionPoolBuilder owned

popzxc:
## What ❔

Removes lifetime in the ConnectionPoolBuilder

## Why ❔

To make it easier to store the builder in long-living structures.

## Checklist

### DIFF
--- a/core/lib/dal/src/connection/mod.rs
+++ b/core/lib/dal/src/connection/mod.rs
@@ -19,13 +19,13 @@ fn get_test_database_url() -> anyhow::Result<String> {
 }
 
 /// Builder for [`ConnectionPool`]s.
-pub struct ConnectionPoolBuilder<'a> {
-    database_url: &'a str,
+pub struct ConnectionPoolBuilder {
+    database_url: String,
     max_size: u32,
     statement_timeout: Option<Duration>,
 }
 
-impl<'a> fmt::Debug for ConnectionPoolBuilder<'a> {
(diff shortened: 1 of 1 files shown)


core/lib/dal/src/connection/mod.rs
```diff
@@ -150,17 +150,17 @@ impl ConnectionPool {
     }
 
     /// Initializes a builder for connection pools.
-    pub fn builder(database_url: &str, max_pool_size: u32) -> ConnectionPoolBuilder<'_> {
+    pub fn builder(database_url: &str, max_pool_size: u32) -> ConnectionPoolBuilder {
         ConnectionPoolBuilder {
-            database_url,
+            database_url: database_url.to_string(),
```
slowli, line 155: Micro-nit: I think `to_owned()` is more canonical way to express this conversion (there's [a Clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#/str_to_string) for this).

slowli (approved):

core/lib/dal/src/connection/mod.rs
popzxc: I personally find this lint controversial, unless enforced throughout the workspace.
If/when we will get to the point where we enable it, I'll be more than happy to follow Clippy's suggestions (e.g. I'm sure that *everyone* follows it and the codebase is consistent), but without it, I think it just adds to the potential nitpicking reason pool (which isn't small anyways), making reviews longer without any particular benefit.

PR #679, opened 2023-12-13
feat: applied status snapshots dal

tomg10:
Adding applied snapshots DAL that will be used to track status of EN node applied snapshot.

### DIFF
--- a/core/lib/dal/migrations/20231213192041_applied-snapshot-status.up.sql
+++ b/core/lib/dal/migrations/20231213192041_applied-snapshot-status.up.sql
@@ -1,0 +1,6 @@
+CREATE TABLE applied_snapshot_status
+(
+    l1_batch_number         BIGINT NOT NULL PRIMARY KEY,
+    is_finished             BOOL NOT NULL,
+    last_finished_chunk_id   INT
+)
--- a/core/lib/dal/migrations/20231213192041_applied-snapshot-status.up.sql
+++ b/core/lib/dal/migrations/20231213192041_applied-snapshot-status.up.sql
@@ -1,6 +1,0 @@
-CREATE TABLE applied_snapshot_status
-(
-    l1_batch_number         BIGINT NOT NULL PRIMARY KEY,
-    is_finished             BOOL NOT NULL,
-    last_finished_chunk_id   INT
-)
--- a/core/lib/dal/migrations/20231213192041_snapshot-recovery.up.sql
+++ b/core/lib/dal/migrations/20231213192041_snapshot-recovery.up.sql
@@ -5,7 +5,6 @@ CREATE TABLE snapshot_recovery
     miniblock_number       BIGINT    NOT NULL,
     miniblock_root_hash    BYTEA     NOT NULL,
 
-    is_finished            BOOL      NOT NULL,
     last_finished_chunk_id INT,
     total_chunk_count     INT       NOT NULL,
 
--- a/core/lib/dal/migrations/20231213192041_snapshot-recovery.up.sql
+++ b/core/lib/dal/migrations/20231213192041_snapshot-recovery.up.sql
@@ -1,0 +1,14 @@
+CREATE TABLE snapshot_recovery
+(
+    l1_batch_number        BIGINT    NOT NULL PRIMARY KEY,
+    l1_batch_root_hash     BYTEA     NOT NULL,
+    miniblock_number       BIGINT    NOT NULL,
+    miniblock_root_hash    BYTEA     NOT NULL,
+
+    is_finished            BOOL      NOT NULL,
+    last_finished_chunk_id INT,
+    total_chunk_count     INT       NOT NULL,
+
+    created_at             TIMESTAMP NOT NULL,
+    updated_at             TIMESTAMP NOT NULL
+)
--- a/core/lib/dal/src/applied_snapshot_status_dal.rs
+++ b/core/lib/dal/src/applied_snapshot_status_dal.rs
@@ -1,4 +1,4 @@
-use zksync_types::{snapshots::AppliedSnapshotStatus, L1BatchNumber};
+use zksync_types::{snapshots::AppliedSnapshotStatus, L1BatchNumber, MiniblockNumber};
 
 use crate::StorageProcessor;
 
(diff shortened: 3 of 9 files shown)


core/lib/dal/migrations/20231213192041_applied-snapshot-status.up.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE applied_snapshot_status
```
slowli, line 1: Bikeshedding: Maybe name this `snapshot_recovery`? "Status" sound like something logically small / confined, while I'm not sure that this entity will be this.

core/lib/dal/migrations/20231213192041_applied-snapshot-status.up.sql
```diff
@@ -0,0 +1,6 @@
...
+(
+    l1_batch_number         BIGINT NOT NULL PRIMARY KEY,
+    is_finished             BOOL NOT NULL,
+    last_finished_chunk_id   INT
```
slowli, line 5: We have discussed that we'd want to store snapshot metadata in this table, right? For the purposes of metadata calculator recovery, we'd need:

- Snapshot miniblock number
- Hash of the snapshot L1 batch

For the state keeper recovery, we'd most probably also need:

- Hash of the snapshot miniblock

Also, AFAIU, we have `created_at` / `updated_at` timestamps for all tables for audit purposes.

core/lib/dal/src/applied_snapshot_status_dal.rs
```diff
@@ -0,0 +1,100 @@
...
+#[derive(Debug)]
+pub struct AppliedSnapshotStatusDal<'a, 'c> {
+    pub(crate) storage: &'a mut StorageProcessor<'c>,
+}
+
+impl AppliedSnapshotStatusDal<'_, '_> {
+    pub async fn set_applied_snapshot_status(
+        &mut self,
+        status: &AppliedSnapshotStatus,
+    ) -> Result<(), sqlx::Error> {
```
slowli, line 14: Nit: You can use `sqlx::Result<()>` here; it's used in most other cases.

core/lib/dal/src/applied_snapshot_status_dal.rs
```diff
@@ -0,0 +1,100 @@
...
+                    is_finished            = excluded.is_finished,
+                    last_finished_chunk_id = excluded.last_finished_chunk_id
+            "#,
+            status.l1_batch_number.0 as i64,
+            status.is_finished,
+            status.last_finished_chunk_id.map(|v| v as i32)
+        )
+            .execute(self.storage.conn())
+            .await
+            .unwrap();
```
slowli, line 32: - Nit: Indentation should be 4 spaces smaller here.
- Please propagate an error instead of `unwrap()`ping it.

core/lib/dal/migrations/20231213192041_applied-snapshot-status.up.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE applied_snapshot_status
+(
+    l1_batch_number         BIGINT NOT NULL PRIMARY KEY,
+    is_finished             BOOL NOT NULL,
```
slowli, line 4: In vacuum, I think that a status column modeling an enum would be better, since it'd allow for a richer lifecycle and allow extending it in the future.

core/lib/types/src/snapshots.rs
```diff
@@ -71,3 +71,10 @@ pub struct SnapshotFactoryDependencies {
 pub struct SnapshotFactoryDependency {
     pub bytecode: Vec<u8>,
 }
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+pub struct AppliedSnapshotStatus {
+    pub l1_batch_number: L1BatchNumber,
+    pub is_finished: bool,
+    pub last_finished_chunk_id: Option<u64>,
```
slowli, line 79: I think here these columns would be better replaced with an enum to more accurately model the snapshot lifecycle; something like:

```rust
pub enum SnapshotRecoveryState {
    NotStarted,
    Pending { last_finished_chunk_id: u64 },
    Finished,
}
```

core/lib/dal/migrations/20231213192041_applied-snapshot-status.up.sql
RomanBrodetski: Enum is better than bool, but perhaps in this case we can get away without a status column at all - having both `last_finished_chunk_id` and `last_chunk_id` we can infer this info from the DB - in this case `status` is denormalization. 
Although it's less explicit so I won't insist.

Also, I think we should mark snapshot recovery as done as soon as the postgres recolvery stage is complete. At this point other components might can start - as we discussed, they must not rely on any l1 batches present in the system. Or do you guys already invision a more extensive lifecycle of rows in this table?

slowli: Do you mean inferring the recovery status from `last_finished_chunk_id` and `total_number_of_chunks`? I think this is OK, at least at this point. There's no need to invent complexities where we can avoid them.

core/lib/types/src/snapshots.rs
tomg10: It may be required to introduce more sophisticates progress tracking, but I'd try to delay making this change as long as possible, when we are sure this indeed is needed

core/lib/dal/migrations/20231213192041_applied-snapshot-status.up.sql
tomg10: > Also, I think we should mark snapshot recovery as done as soon as the postgres recolvery stage is complete (...)
I agree with that part.
I also agree that we could infer is_finished from last_finished_chunk_id and total_chunk_count, but I am a big fan of denormalization in such cases (i.e. a single column that contains "status" of the whole row)

core/lib/dal/src/applied_snapshot_status_dal.rs
tomg10: corrected

core/lib/dal/src/applied_snapshot_status_dal.rs
tomg10: corrected

core/lib/dal/migrations/20231213192041_applied-snapshot-status.up.sql
tomg10: I added all those columns

core/lib/dal/migrations/20231213192041_applied-snapshot-status.up.sql
RomanBrodetski: I still think it's better not to denormalize unless we have a good reason to, which I don't see here. 
But it's not a strong opinion - it's also more explicit with such a column, so feel free to do either way

core/lib/dal/src/applied_snapshot_status_dal.rs
```diff
@@ -0,0 +1,142 @@
+use zksync_types::{snapshots::AppliedSnapshotStatus, L1BatchNumber, MiniblockNumber};
+
+use crate::StorageProcessor;
+
+#[derive(Debug)]
+pub struct AppliedSnapshotStatusDal<'a, 'c> {
```
slowli, line 6: Since the table is renamed, it makes sense to rename this to `SnapshotRecoveryDal`, its getter to `snapshot_recovery_dal()`, `AppliedSnapshotStatus` to `SnapshotRecoveryStatus` etc. Technically can be done in a follow-up PR, but it shouldn't be hard 🙃

core/lib/dal/src/applied_snapshot_status_dal.rs
```diff
@@ -0,0 +1,142 @@
...
+    pub(crate) storage: &'a mut StorageProcessor<'c>,
+}
+
+impl AppliedSnapshotStatusDal<'_, '_> {
+    pub async fn set_applied_snapshot_status(
```
slowli, line 11: This can be done later, but I think it makes sense in general to have a more typesafe interface by having separate methods for creating and updating the snapshot:

```rust
pub async fn create_snapshot_recovery_status(
    &mut self,
    status: &SnapshotRecoveryStatus
) -> sqlx::Result<()> {
    // Uses `INSERT`; fails if the `SnapshotRecoveryStatus` exists
}

pub async fn set_last_finished_chunk_id(
    &mut self,
    last_finished_chunk_id: u64,
) -> sqlx::Result<()> {
    // Uses `UPDATE`; fails if the snapshot doesn't exist
}
```

This is how it's done in other DALs (e.g., see L1 batch lifecycle in `BlocksDal`), and IMO it makes code easier to maintain in general. If we don't have full understanding of modifying methods, we can skip adding them for now.

core/lib/types/src/snapshots.rs
```diff
@@ -172,3 +172,14 @@ impl ProtoFmt for SnapshotStorageLogsChunk {
         }
     }
 }
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+pub struct AppliedSnapshotStatus {
+    pub l1_batch_number: L1BatchNumber,
+    pub l1_batch_root_hash: Vec<u8>,
```
slowli, line 179: Please use `H256` for hashes.

core/lib/types/src/snapshots.rs
```diff
@@ -172,3 +172,14 @@ impl ProtoFmt for SnapshotStorageLogsChunk {
         }
     }
 }
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
```
slowli, line 176: Nit: If we don't need `Serialize` and `Deserialize` rn, it makes sense to not derive them; these traits could trivially be derived later if necessary.

core/lib/dal/migrations/20231213192041_snapshot-recovery.up.sql
```diff
@@ -0,0 +1,14 @@
+CREATE TABLE snapshot_recovery
+(
+    l1_batch_number        BIGINT    NOT NULL PRIMARY KEY,
+    l1_batch_root_hash     BYTEA     NOT NULL,
+    miniblock_number       BIGINT    NOT NULL,
+    miniblock_root_hash    BYTEA     NOT NULL,
+
+    is_finished            BOOL      NOT NULL,
```
slowli, line 8: Just to highlight: As we've just had a conversation with @RomanBrodetski regarding (de)normalization, does this field still make sense? 🙃 AFAIU, at least in the initial iteration, `is_finished` will always equal `last_finished_chunk_id == total_chunk_count - 1`.

slowli (commented):
Some relatively minor nits; overall, looks solid 👍

core/lib/dal/src/applied_snapshot_status_dal.rs
tomg10: I agree it may make sense to refactor this in the future, but after this table "stabilizes"

core/lib/types/src/snapshots.rs
tomg10: done

core/lib/types/src/snapshots.rs
tomg10: done

core/lib/dal/migrations/20231213192041_snapshot-recovery.up.sql
tomg10: corrected

core/lib/dal/src/applied_snapshot_status_dal.rs
tomg10: corrected

slowli (approved):
Nice.

PR #686, opened 2023-12-14
feat: Dockerfile and step in CI to build snapshots-creator image

tomg10:


docker/snapshots-creator/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+
+ENV RUSTUP_HOME=/usr/local/rustup \
+    CARGO_HOME=/usr/local/cargo \
+    PATH=/usr/local/cargo/bin:$PATH
+
+RUN curl https://sh.rustup.rs -sSf | bash -s -- -y && \
+    rustup install nightly-2023-08-21 && \
+    rustup default nightly-2023-08-21
+
+RUN cargo build --release --bin snapshots_creator
```
slowli, line 19: Dumb question: Do we prefer a separate image (as opposed to e.g. extending the server image or `cross_external_nodes_checker` image – perhaps, renaming it to something like "utils") to keep the single responsibility principle? As another option, we could use an existing image as a base and only change its entrypoint (this should save on the image build time IIUC).

slowli (approved):

docker/snapshots-creator/Dockerfile
tomg10: We definitely could, in my previous company we were able to have a single image for most our deployments and it worked really great, but It's a bigger topic that I suggest raising on team channel. In practical terms, I'm not sure yet this would save us build time as those images are probably built in parallel, but it would certainly be a productivity boost.

artmakh (approved):

PR #688, opened 2023-12-14
feat(en): Fix operator address assignment for ENs

slowli:
## What ❔

- Moves `fee_account_address` from the `l1_batches` table to the `miniblocks` table using a combined SQL and programmatic migration.
- Replaces uses of `l1_gas_price`, `l2_fair_gas_price` and `base_fee_per_gas` from `l1_batches` to the corresponding `miniblocks` fields. Removes uses of `l1_batches.is_finished`.
- Updates `L1BatchHeader` / `MiniblockHeader` correspondingly.

## Why ❔

Currently, operator address is not persisted for miniblocks, but only for L1 batches. This has several negative implications:

- ENs cannot be used to synchronize other ENs since SyncBlock.operator_address they return may be bogus.
- We have nowhere to get the real operator address for gossip-based syncing as well.

## Checklist


slowli (commented):
The overall migration strategy is as follows:

1. Perform SQL migration changing the DB schema. It should be fast and shouldn't disrupt system operation.
2. Perform programmatic data migration (see `fee_address_migration` module). This migration is run in the background in the state keeper / EN pod. It relies that step 1 is complete (i.e., data schema is changed).
3. (Will be implemented separately) Perform another SQL migration removing obsolete L1 batch columns.
 
DB queries are rewritten in a way that they will continue to work after the obsolete L1 batch columns are removed. (The natural exception is migration-specific queries, but they should never run at this point.) I.e., code deployed on step 2 should continue working both before and after completing step 3.

core/lib/zksync_core/src/state_keeper/io/fee_address_migration.rs
```diff
@@ -0,0 +1,304 @@
...
+    sleep_interval: Duration,
+    stop_receiver: watch::Receiver<bool>,
+) -> anyhow::Result<MigrationOutput> {
+    anyhow::ensure!(chunk_size > 0, "Chunk size must be positive");
+
+    let mut storage = pool.access_storage().await?;
+    // Check whether the last miniblock is migrated. If it is, we can finish early.
+    // This will also protect us from running unsupported `BlocksDal` queries after the 2nd DB migration is run
+    // removing `l1_batches.fee_account_address`.
+    if is_fee_address_migrated(&mut storage, last_miniblock).await? {
```
perekopskiy, line 74: It seems that your approach assumes that migration is run in a one-shot. If the process was terminated and there were some non-migrated miniblocks then they won't be migrated on restart. I mean the following scenario:
You launch server, it migrates pending miniblocks and starts migration for others -> meanwhile some batch is sealed so some blocks are not considered to be pending anymore -> process exits -> on restart  `is_fee_address_migrated(&mut storage, last_miniblock).await` returns true, so migration is skipped.

As a solution, you can create some temp DB table and save there last sealed miniblock number when you're calling `copy_fee_account_address_for_pending_miniblocks` for the first time. Then pass this number to `migrate_miniblocks` instead of `self.io.current_miniblock_number() - 1`. Having a temporary table looks pretty ugly but I don't have other solutions in mind

slowli: Yep, that's a problem. I've spent some effort to make the migration fault-tolerant (there's even a corresponding test), but it evidently wasn't enough 🙃

In terms of alternatives: What about checking the start element of each chunk of miniblocks to see whether it needs to be migrated instead of using binary search? As long as the chunk size is a hard-coded constant, it seems to work, and if the chunk size is large, it'll be reasonably efficient (e.g., if the chunk size is 100,000, we'd need to check approx. 220 addresses on the mainnet, and we could do this in a single DB query). A failure case would be if we deem that the chunk is migrated, but it's really not (or at least not *fully* migrated). IIUC, this can only happen if we assume that the node is downgraded and produces some miniblocks w/o `fee_account_address` after the migration is complete (but in this case, your approach won't work either).

core/lib/dal/migrations/20231212121822_add_miniblocks_fee_account_address.up.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE miniblocks
+    ADD COLUMN fee_account_address BYTEA NOT NULL DEFAULT '\x0000000000000000000000000000000000000000'::bytea;
```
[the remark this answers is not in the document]
slowli: I'll need to check whether there are other RPC methods returning this data, but the obvious one is `zks_getBlockDetails`. Would it be OK for it to return an error in case of non-migrated address (i.e., if `fee_account_address == Address::zero()`)? Another option would be to access `l1_batches.fee_account_address` in this case; it *should* continue working even after this column is removed because this case would never be triggered. I don't really like this non-local / cumbersome reasoning though.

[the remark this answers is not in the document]
perekopskiy: Returning error will result in indexers' halt during the migration, as I understand its time is of order of 2-3 minutes, so doesn't sound like a good thing.
If implementing a fallback to `l1_batches.fee_account_address` is not really difficult, I'd go this way.

core/lib/zksync_core/src/state_keeper/io/fee_address_migration.rs
perekopskiy: Your alternative looks better, I think it's ok to forbid downgrading

core/lib/dal/migrations/20231212121822_add_miniblocks_fee_account_address.up.sql
perekopskiy, line 2: Have you estimated time for this query? It looks like it will take a lot of time and doesn't differ much from executing full migration in SQL

slowli: No, I haven't. What's the best way to do this? Would having nullable `fee_account_address` be better in terms of performance?

perekopskiy: You can test it on stage DB, i.e. just run
```
ALTER TABLE miniblocks
    ADD COLUMN fee_account_address BYTEA NOT NULL DEFAULT '\x0000000000000000000000000000000000000000'::bytea;
```
vs
```
ALTER TABLE miniblocks
    ADD COLUMN fee_account_address BYTEA;
```

and drop the column after the test

slowli: Both `ALTER TABLE` statements execute almost instantly on the stage DB. Could there be differences in how they are handled down the line, or does this mean that the current approach is OK?

perekopskiy: I think it is OK then

perekopskiy (dismissed):
Looks good! Sorry for the long review. Please resolve conflicts and I will reapprove

perekopskiy (approved):

RomanBrodetski (approved):
Looks good! Surprized it's that involved though. I don't see how exactly we could do it simpler. But it's a good change and we need experience in migrating large tables.

RomanBrodetski (approved):

PR #690, opened 2023-12-14
fix: error when running `zk init`

### LINKED ISSUE #689: Error when running `zk init` in `main`
#### 📝 Description

There's an error when you run `zk init` in the main branch.

#### 🔄 Reproduction Steps

**Path 1**

1. Clone the repo.
2. Setup `zk`.
3. Run `zk init`.

**Path 2**
1. Clone the repo.
2. `cd` into `infrastructure/zk`.
3. `yarn install`.
4. `yarn build`.
5. `yarn start init`.

#### 🤔 Expected Behavior

`zk init` should not fail.

#### 😯 Current Behavior

`zk init` is failing.

#### 🖥️ Environment

-

#### 📋 Additional Context

-

#### 📎 Log Output

```
docker: Error response from daemon: create ./volumes: "./volumes" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.
See 'docker run --help'.
Error: Child process exited with code 125
error Command failed with exit code 1.
```

ilitteri:
Resolves #689.

## What ❔

This PR fixes `zk init` cmd. Updating the path to the volume passed to the `docker run --rm -v` cmd in `infrastructure/zk/src/down.ts`.

## Why ❔

The change mentioned above is made because the command `docker run -v` does not accept a relative paths but absolute paths, and `./volumes/...` is being passed on line 7 of the mentioned file.

If you clone this repo and run `zk init` after installing `zk`, you'll see something like this:

```
docker: Error response from daemon: create ./volumes: "./volumes" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.
See 'docker run --help'.
Error: Child process exited with code 125
error Command failed with exit code 1.
```

## Checklist

### DIFF
--- a/infrastructure/zk/src/down.ts
+++ b/infrastructure/zk/src/down.ts
@@ -4,7 +4,7 @@ import * as utils from './utils';
 export async function down() {
     await utils.spawn('docker compose down -v');
     await utils.spawn('docker compose rm -s -f -v');
-    await utils.spawn('docker run --rm -v ./volumes:/volumes postgres:14 bash -c "rm -rf /volumes/*"');
+    await utils.spawn('docker run --rm -v volumes:/volumes postgres:14 bash -c "rm -rf /volumes/*"');
 }
 
 export const command = new Command('down').description('stop development containers').action(down);


ilitteri:
There's a chance that we should update the workflows files too, waiting to see if the CI fails.

infrastructure/zk/src/down.ts
```diff
@@ -4,7 +4,7 @@ import * as utils from './utils';
 export async function down() {
     await utils.spawn('docker compose down -v');
     await utils.spawn('docker compose rm -s -f -v');
-    await utils.spawn('docker run --rm -v ./volumes:/volumes postgres:14 bash -c "rm -rf /volumes/*"');
+    await utils.spawn('docker run --rm -v volumes:/volumes postgres:14 bash -c "rm -rf /volumes/*"');
```
montekki, line 7: `docker --version`? I am pretty sure this works on 24.0.7

ilitteri: Updating Docker solved the issue, nevertheless I think that it would be better to have absolut paths in order to avoid this kind of issues. What do you think?

montekki: it is fishy because it is not an absolute path.

mationorato: Hey! They are two different options to define a relative path, but the second one works for multiple docker versions, so I think it is better to avoid unexpected errors.

IMH I think we should keep the second option that @ilitteri suggested, or raise a proper error otherwise.

montekki: It does not do what it is supposed to do, you can start a network with `zk init`, and then run `zk down` and then `ls ./volumes`, with this patch you will see this:

```
exa ./volumes/
geth  postgres  prysm
```
so it doesn't actually clean the directory although it is supposed to.

PR #691, opened 2023-12-15
feat(snapshot creator): Make snapshot creator fault-tolerant

slowli:
## What ❔

Makes the snapshot creator tolerant to failures. The snapshot creator now checks whether there's a partially created snapshot on start. If there is, the creator continues producing this snapshot instead of starting from scratch.

## Why ❔

Since the snapshot creator can run for an extended amount of time, not having fault tolerance is unmaintainable in the long run.

## Checklist


prover/Cargo.lock
```diff
@@ -7227,7 +7266,7 @@ dependencies = [
 [[package]]
 name = "zksync_protobuf_build"
 version = "0.1.0"
-source = "git+https://github.com/matter-labs/era-consensus.git?rev=da015d4c94b19962bc11622b6cc256e214256555#da015d4c94b19962bc11622b6cc256e214256555"
+source = "git+https://github.com/matter-labs/era-consensus.git?rev=49b1a98f80d0e9f74fdceadece4283e745c71599#49b1a98f80d0e9f74fdceadece4283e745c71599"
```
RomanBrodetski, line 7269: pass-by comment: did you change file on purpose?

slowli: Somewhat. I have both the main and prover workspaces set up in the IDE, so the prover lockfile is kept up to date. We don't have a check that it's up to date in CI, and it often gets outdated in practice.

core/lib/types/src/snapshots.rs
```diff
@@ -3,28 +3,47 @@ use zksync_basic_types::{L1BatchNumber, MiniblockNumber};
...
+/// Information about all snapshots persisted by the node.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct AllSnapshots {
+    /// L1 batch numbers for complete snapshots. Ordered by descending number (i.e., 0th element
+    /// corresponds to the newest snapshot).
     pub snapshots_l1_batch_numbers: Vec<L1BatchNumber>,
+    /// Same as `snapshots_l1_batch_numbers`, but for incomplete snapshots (ones in which some
+    /// storage log chunks may be missing).
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
```
tomg10, line 15: Do I understand correctly that it removes this field completely from response? Is removing some fields, i.e. dynamic schema in any way standard behaviour? I would just expect the field to always be there, even if empty

tomg10: Also, why do we want to have this field in the first place?

core/lib/types/src/snapshots.rs
```diff
@@ -3,28 +3,47 @@ use zksync_basic_types::{L1BatchNumber, MiniblockNumber};
...
-//contains all data not contained in factory_deps/storage_logs files to perform restore process
+/// Snapshot data returned by using JSON-RPC API.
+/// Contains all data not contained in factory deps / storage logs files to perform restore process.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct SnapshotHeader {
     pub l1_batch_number: L1BatchNumber,
     pub miniblock_number: MiniblockNumber,
-    //ordered by chunk ids
+    pub is_complete: bool,
```
tomg10, line 45: As I'm thinking of it longer, can we make the API transparent to the resilience in snapshots creator? i.e. we remove this field and behave as if incomplete snapshots don't exist in RPC API?

core/lib/types/src/snapshots.rs
```diff
@@ -3,28 +3,47 @@ use zksync_basic_types::{L1BatchNumber, MiniblockNumber};
...
-    pub storage_logs_filepaths: Vec<String>,
+    /// Paths to the storage log blobs. Ordered by the chunk ID. If a certain chunk is not produced yet,
+    /// the corresponding path is `None`.
+    pub storage_logs_filepaths: Vec<Option<String>>,
+}
+
+impl SnapshotMetadata {
+    /// Checks whether a snapshot is complete (contains all information to restore from).
+    pub fn is_complete(&self) -> bool {
+        self.storage_logs_filepaths.iter().all(Option::is_some)
```
tomg10, line 34: Can we make this a column in db?

core/lib/dal/src/snapshots_dal.rs
```diff
@@ -14,14 +35,17 @@ impl SnapshotsDal<'_, '_> {
...
-    ) -> Result<(), sqlx::Error> {
+    ) -> sqlx::Result<()> {
         sqlx::query!(
-            "INSERT INTO snapshots (l1_batch_number, storage_logs_filepaths, factory_deps_filepath, created_at, updated_at) \
-             VALUES ($1, $2, $3, NOW(), NOW())",
+            "INSERT INTO snapshots ( \
+                l1_batch_number, storage_logs_filepaths, factory_deps_filepath, \
+                created_at, updated_at \
+            ) \
+            VALUES ($1, array_fill(''::text, ARRAY[$2::integer]), $3, NOW(), NOW())",
```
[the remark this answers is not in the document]
tomg10: But in general I am a big fan of denormalization if it makes the code simpler or makes it easier to write some queries

tomg10 (changes requested):
Looks correct and should not have any corner cases, it may only get stuck (for instance I've seen postgres queries get stuck on very large queries), but It's not very probable and there would be an alert for that. 
I also suggested we make resilience completely transparent on the RPC API level and suggested a few changes regarding snapshots db schema.

core/lib/types/src/snapshots.rs
RomanBrodetski: this is denormalization and might result in inconsistent state

core/lib/dal/src/snapshots_dal.rs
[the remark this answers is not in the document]
RomanBrodetski: > But in general I am a big fan of denormalization if it makes the code simpler or makes it easier to write some queries

Hmm, what do you mean here by denormalization here? like, storing filenames together with ids as opposed to an array with empty strings?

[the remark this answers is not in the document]
tomg10: by denormalization I mean storing filenames together with ids

core/lib/types/src/snapshots.rs
tomg10: I agree, but I think it would make it significantly easier to see the 'status' of the snapshot just by looking at the table

tomg10: currently, you'd have to write a query for that (there might be hundreds of chunks)

tomg10: but I should have definitely marked this comment as a 'nit'

core/lib/zksync_core/src/api_server/web3/tests/snapshots.rs
```diff
@@ -0,0 +1,125 @@
...
+
+use std::collections::HashSet;
+
+use zksync_types::block::{BlockGasCount, L1BatchHeader};
+use zksync_web3_decl::namespaces::SnapshotsNamespaceClient;
+
+use super::*;
+use crate::state_keeper::tests::create_l1_batch_metadata;
+
+async fn seal_l1_batch(
```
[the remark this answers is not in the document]
slowli: IMO, it's a bit premature and in any case would best be solved separately (similar methods are used in several crates, e.g. `zksync_state`).

core/lib/zksync_core/src/api_server/web3/tests/snapshots.rs
```diff
@@ -0,0 +1,125 @@
...
+    let metadata = create_l1_batch_metadata(number.0);
+    storage
+        .blocks_dal()
+        .save_l1_batch_metadata(number, &metadata, H256::zero(), false)
+        .await?;
+    Ok(())
+}
+
+#[derive(Debug)]
+struct SnapshotBasics {
```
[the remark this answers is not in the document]
slowli: What name would you prefer? It's in a test module and implements `HttpTest` crate which IMO describes its purpose just fine. I'm not a fan of enterprise-y `SnapshotBasicsHttpTest` naming schema, but that's not a strong preference – if you think it'll convey the intention better, I'll switch it.

core/lib/zksync_core/src/api_server/web3/tests/snapshots.rs
```diff
@@ -0,0 +1,125 @@
...
+            assert!(self.chunk_ids.contains(&chunk.chunk_id));
+            assert!(chunk.filepath.starts_with("file:///storage_logs/"));
+        }
+        Ok(())
+    }
+}
+
+#[tokio::test]
+async fn snapshot_without_chunks() {
+    test_http_server(SnapshotBasics::new([])).await;
```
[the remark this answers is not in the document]
slowli: Yes, this is correct; `test_http_server` provides reusable setup and teardown, so that it doesn't need to be copied for each HTTP server test.

core/lib/types/src/snapshots.rs
slowli: Yes, I think this makes sense at least initially 👍

core/lib/types/src/snapshots.rs
slowli: As we've discussed, normalization argument seems to win. We still have time to change the DB schema in the nearest future, so if we reconsider this decision, it could be implemented separately.

core/lib/zksync_core/src/api_server/web3/tests/snapshots.rs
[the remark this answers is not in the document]
tomg10: I didn't have any specific name in my mind, but I think that using name `SnapshotBasicsHttpTest` or similar to it would have made it a lot easier for me to understand what this does if this is the first time I see a HttpTest

tomg10 (approved):

core/bin/snapshots_creator/src/creator.rs
```diff
@@ -0,0 +1,335 @@
...
+        config: SnapshotsCreatorConfig,
+        min_chunk_count: u64,
+    ) -> anyhow::Result<()> {
+        let latency = METRICS.snapshot_generation_duration.start();
+
+        let Some(progress) = self
+            .load_or_initialize_snapshot_progress(&config, min_chunk_count)
+            .await?
+        else {
+            return Ok(()); // No snapshot creation is necessary
```
RomanBrodetski, line 270: would it make sense to mention in the comment why it happens - that is, the snapshot for the current l1 batch is already genertaed?

RomanBrodetski (approved):

RomanBrodetski (approved):

PR #693, opened 2023-12-15
feat(api): remove jsonrpc backend

montekki:
## What ❔

The new release of `jsonrpsee` removes all known blockers for removing `jsonrpc` backend.

## Checklist


core/lib/zksync_core/src/api_server/web3/mod.rs
```diff
@@ -730,24 +510,37 @@ impl<G: 'static + Send + Sync + L1GasPriceProvider> ApiBuilder<G> {
...
+            );
+            let server = server_builder
+                .set_batch_request_config(batch_request_config)
+                .set_http_middleware(middleware)
+                .set_id_provider(EthSubscriptionIdProvider::default())
+                .max_response_body_size(response_body_size_limit)
+                .build(addr)
+                .await
+                .with_context(|| format!("Failed building {transport_str} JSON-RPC server"))?;
+            (server.local_addr(), server.start(rpc))
```
montekki, line 538: This `if` `else` is neeeded because otherwise inserting the `LimitMiddleware` will result in `ServerBuilder` types mismatch. So we need to go all the way to actually creating the server to avoid that.

core/lib/zksync_core/src/api_server/web3/pubsub.rs
```diff
@@ -49,18 +55,14 @@ impl<V: Clone> PubSubNotifier<V> {
             .context("get_sealed_miniblock_number()")
     }
 
-    async fn current_subscribers(&self) -> Vec<V> {
-        self.subscribers.read().await.values().cloned().collect()
-    }
```
montekki, line 54: Can no longer do that, it clones and returns the subscription structures and these clones would alter be dropped causing the subscription to close.

core/lib/zksync_core/src/api_server/web3/pubsub.rs
```diff
@@ -238,83 +304,50 @@ impl EthSubscribe {
...
 
-    /// Assigns ID for the subscriber if the connection is open, returns error otherwise.
-    fn assign_id(
-        subscriber: typed::Subscriber<PubSubResult>,
-    ) -> Result<(typed::Sink<PubSubResult>, SubscriptionId), ()> {
-        let id = H128::random();
-        let sub_id = SubscriptionId::String(format!("0x{}", hex::encode(id.0)));
-        let sink = subscriber.assign_id(sub_id.clone())?;
-        Ok((sink, sub_id))
-    }
```
montekki, line 249: Moved to `EthereumSubscriptionIdProvider`.

core/lib/zksync_core/src/api_server/web3/pubsub.rs
```diff
@@ -181,18 +231,34 @@ impl PubSubNotifier<(typed::Sink<PubSubResult>, PubSubFilter)> {
...
+                for (sink, filter) in subscribers.values() {
                     for log in &new_logs {
                         if filter.matches(log) {
-                            if sink.notify(Ok(PubSubResult::Log(log.clone()))).is_err() {
+                            if sink
+                                .send(
+                                    SubscriptionMessage::from_json(&PubSubResult::Log(log.clone()))
+                                        .expect("PubSubResult always serializable to json;qed"),
+                                )
+                                .await
```
slowli, line 245: Just to understand better: What are the factors determining whether we will block here? AFAIU, it *could* be related to the [`message_buffer_capacity` parameter](https://docs.rs/jsonrpsee-server/latest/jsonrpsee_server/struct.ServerBuilder.html#method.set_message_buffer_capacity) (1,024 by default; IIUC it's applied per connection, rather than per subscription). In general, are we OK with one slow client blocking others? I guess one other option would be to drop slow clients (e.g., by using [`send_timeout`](https://docs.rs/jsonrpsee-server/latest/jsonrpsee_server/struct.SubscriptionSink.html#method.send_timeout) instead of `send`), and/or handling all clients concurrently (although I'm not entirely sure how the latter approach would impact performance).

core/lib/zksync_core/src/api_server/web3/pubsub.rs
```diff
@@ -388,3 +408,20 @@ impl EthSubscribe {
...
+impl EthPubSubServer for EthSubscribe {
+    async fn subscribe(
+        &self,
+        pending: PendingSubscriptionSink,
+        sub_type: String,
+        filter: Option<PubSubFilter>,
+    ) -> SubscriptionResult {
+        let self_ = self.clone();
+
+        self.runtime_handle
```
slowli, line 422: Would this logic work w/o `runtime_handle`? At least judging by a comment near the `runtime_handle` field, it seems `jsonrpc`-specific.

core/lib/zksync_core/src/api_server/web3/pubsub.rs
```diff
@@ -3,23 +3,41 @@
...
+};
 
 use super::{
     metrics::{SubscriptionType, PUB_SUB_METRICS},
     namespaces::eth::EVENT_TOPIC_NUMBER_LIMIT,
 };
 
-pub(super) type SubscriptionMap<T> = Arc<RwLock<HashMap<SubscriptionId, T>>>;
+#[derive(Debug, Clone, Copy, Default)]
+#[non_exhaustive]
```
slowli, line 30: Why do you need `#[non_exhaustive]` here? The struct itself can have reduced visibility (e.g., `pub(super)`), in which case `non_exhaustive` particularly doesn't make sense (it is only effective for public data types).

core/lib/zksync_core/src/api_server/web3/backend_jsonrpsee/batch_limiter_middleware.rs
```diff
@@ -0,0 +1,96 @@
...
+static METRICS: vise::Global<LimitMiddlewareMetrics> = vise::Global::new();
+pub struct LimitMiddleware<S> {
+    inner: S,
+    rate_limiter: Option<Arc<RateLimiter<NotKeyed, InMemoryState, DefaultClock, NoOpMiddleware>>>,
+    transport: Transport,
+}
+
+impl<S> LimitMiddleware<S> {
+    pub(crate) fn new(inner: S, requests_per_minute_limit: Option<u32>) -> Self {
+        API_METRICS.ws_open_sessions.inc_by(1);
```
slowli, line 46: Nit: Similarly to subscriptions, you can use `GaugeGuard` here as well to more neatly encapsulate `ws_open_sessions` changes.

core/lib/zksync_core/src/api_server/web3/backend_jsonrpsee/batch_limiter_middleware.rs
```diff
@@ -0,0 +1,96 @@
...
+    /// Size of batch requests.
+    #[metrics(buckets = Buckets::exponential(1.0..=512.0, 2.0))]
+    size: Family<Transport, Histogram<usize>>,
+    /// Number of requests rejected by the limiter.
+    rejected: Family<Transport, Counter>,
+}
+
+#[vise::register]
+static METRICS: vise::Global<LimitMiddlewareMetrics> = vise::Global::new();
+pub struct LimitMiddleware<S> {
```
slowli, line 38: - Nit: Missing an empty line between items.
- Nit: Visibility could be reduced to `pub(crate)`.
- Nit: Could make sense to add a doc comment that an instance of this middleware is allocated per connection, so that it would be easier to reason about it for people not very familiar with `jsonrpsee`.

core/lib/zksync_core/src/api_server/web3/backend_jsonrpsee/batch_limiter_middleware.rs
```diff
@@ -0,0 +1,96 @@
...
+
+impl<S> LimitMiddleware<S> {
+    pub(crate) fn new(inner: S, requests_per_minute_limit: Option<u32>) -> Self {
+        API_METRICS.ws_open_sessions.inc_by(1);
+
+        Self {
+            inner,
+            rate_limiter: requests_per_minute_limit.map(|limit| {
+                Arc::new(RateLimiter::direct(Quota::per_minute(
+                    NonZeroU32::new(limit).expect("requests per minute must be > 0; qed"),
```
slowli, line 52: Nit: Is it actually checked that `requests_per_minute_limit` is positive beforehand? AFAICT, it's not checked in `ApiBuilder`.

core/lib/zksync_core/src/api_server/web3/backend_jsonrpsee/batch_limiter_middleware.rs
```diff
@@ -0,0 +1,96 @@
...
+    size: Family<Transport, Histogram<usize>>,
+    /// Number of requests rejected by the limiter.
+    rejected: Family<Transport, Counter>,
+}
+
+#[vise::register]
+static METRICS: vise::Global<LimitMiddlewareMetrics> = vise::Global::new();
+pub struct LimitMiddleware<S> {
+    inner: S,
+    rate_limiter: Option<Arc<RateLimiter<NotKeyed, InMemoryState, DefaultClock, NoOpMiddleware>>>,
```
slowli, line 40: Nit: Why is `Arc` used here? It seems that the code compiles with it removed.

core/lib/zksync_core/src/api_server/web3/backend_jsonrpsee/batch_limiter_middleware.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+impl<'a, S> RpcServiceT<'a> for LimitMiddleware<S>
+where
+    S: Send + Clone + Sync + RpcServiceT<'a>,
+{
+    type Future = ResponseFuture<S::Future>;
+
+    fn call(&self, request: Request<'a>) -> Self::Future {
+        if let Some(ref rate_limiter) = self.rate_limiter {
```
slowli, line 73: Bikeshedding: IMO, using `ref` / `ref mut` is really old-school; in most cases, it can be avoided by borrowing the matched data (i.e., `.. = &self.rate_limiter` in this case).

core/lib/zksync_core/src/api_server/web3/backend_jsonrpsee/batch_limiter_middleware.rs
slowli, line 1: Future work (?): Ideally, it'd make sense to cover rate limiting logic with a couple of unit tests (e.g., simple rate-limiting case and rate-limiting for batch requests). This is subjective, of course, but part of my confidence that the new WS backend works is that it's somewhat covered with unit tests.

slowli (commented):
Overall, looks quite nice 👍

core/lib/zksync_core/src/api_server/web3/pubsub.rs
montekki: removed

core/lib/zksync_core/src/api_server/web3/pubsub.rs
montekki: removed

core/lib/zksync_core/src/api_server/web3/backend_jsonrpsee/batch_limiter_middleware.rs
montekki: fixed.

core/lib/zksync_core/src/api_server/web3/backend_jsonrpsee/batch_limiter_middleware.rs
montekki: Removed `Arc`

core/lib/zksync_core/src/api_server/web3/backend_jsonrpsee/batch_limiter_middleware.rs
montekki: changed to new school

core/lib/zksync_core/src/api_server/web3/backend_jsonrpsee/batch_limiter_middleware.rs
montekki: Added `GaugeGuard`.

core/lib/zksync_core/src/api_server/web3/backend_jsonrpsee/batch_limiter_middleware.rs
montekki: I've added the test for rate limiting, that is for our own logic implementation. For batches we currenly will rely on [jsonrpsee implementation](https://docs.rs/jsonrpsee-server/0.21.0/jsonrpsee_server/struct.ServerBuilder.html#method.set_batch_request_config) for both http and ws servers and idk if that makes sense to test 3-d party  code

core/lib/zksync_core/src/api_server/web3/pubsub.rs
```diff
@@ -80,16 +94,34 @@ impl PubSubNotifier<typed::Sink<PubSubResult>> {
...
                             break;
                         }
                         PUB_SUB_METRICS.notify[&SubscriptionType::Blocks].inc();
                     }
                 }
                 notify_latency.observe();
+                drop(subscribers);
+                for closed_sub in closed_subscriptions.into_iter() {
+                    if self.subscribers.write().await.remove(&closed_sub).is_some() {
+                        PUB_SUB_METRICS.active_subscribers[&SubscriptionType::Blocks].dec_by(1);
```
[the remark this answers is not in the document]
montekki: Moved to metering with `GaugeGuard`.

core/lib/zksync_core/src/api_server/web3/backend_jsonrpsee/batch_limiter_middleware.rs
montekki: Actually made it `NoneZeroU32` throughout the codebase.

core/lib/zksync_core/src/api_server/web3/pubsub.rs
montekki: I've added a hardcoded time limit to `send_timeout` here, should we move it to configuration or change to lower amount (currently 1 second)?

core/lib/zksync_core/src/api_server/web3/tests/mod.rs
```diff
@@ -123,10 +138,19 @@ async fn spawn_server(
...
         ApiTransportLabel::Http => ApiBuilder::jsonrpsee_backend(api_config, pool).http(0),
-        ApiTransportLabel::Ws => ApiBuilder::jsonrpc_backend(api_config, pool)
-            .ws(0)
-            .with_polling_interval(POLL_INTERVAL)
-            .with_subscriptions_limit(100),
+        ApiTransportLabel::Ws => {
+            let mut builder = ApiBuilder::jsonrpsee_backend(api_config, pool)
+                .ws(0)
+                .with_polling_interval(POLL_INTERVAL)
+                .with_subscriptions_limit(100);
```
slowli, line 145: BTW, is `subscriptions_limit` used anywhere after refactoring? It seems like it's not, and the corresponding config param doesn't really do anything (although I have doubts it worked as intended previously). I wonder how it *should* work; unlike with filters, there's no obvious "least recently used" indicator with subscriptions, so there's no simple policy to evict "old" subscriptions when newer ones arrive.

core/lib/zksync_core/src/api_server/web3/backend_jsonrpsee/batch_limiter_middleware.rs
```diff
@@ -0,0 +1,93 @@
...
+static METRICS: vise::Global<LimitMiddlewareMetrics> = vise::Global::new();
+
+/// A rate-limiting middleware.
+///
+/// `jsonrpsee` will allocate the instance of this struct once per session.
+pub(crate) struct LimitMiddleware<S> {
+    inner: S,
+    rate_limiter: Option<RateLimiter<NotKeyed, InMemoryState, DefaultClock, NoOpMiddleware>>,
+    transport: Transport,
+    __guard: GaugeGuard,
```
slowli, line 48: Nit: One underscore is enough (it indicates that `guard` is not read, just used for its `Drop` implementation).

core/lib/zksync_core/src/api_server/web3/backend_jsonrpsee/batch_limiter_middleware.rs
slowli: Sorry, I wasn't clear. I've meant a test in which a client would send a batch of requests of acceptable size that would be rate-limited (e.g., a session has 2 remaining request permits per minute and gets a batch of 3 requests).

core/lib/zksync_core/src/api_server/web3/tests/ws.rs
```diff
@@ -461,5 +471,38 @@ impl WsTest for LogSubscriptionsWithDelay {
...
+
+#[async_trait]
+impl WsTest for RateLimiting {
+    async fn test(
+        &self,
+        client: &WsClient,
+        _pool: &ConnectionPool,
+        _pub_sub_events: mpsc::UnboundedReceiver<PubSubEvent>,
+    ) -> anyhow::Result<()> {
+        let _ = client.chain_id().await.unwrap();
```
slowli, line 488: Nit: `let _ = ` looks redundant; the returned value is not `#[must_use]`.

montekki: Removed.

core/lib/zksync_core/src/api_server/web3/backend_jsonrpsee/batch_limiter_middleware.rs
montekki: renamed

core/lib/zksync_core/src/api_server/web3/tests/mod.rs
montekki: Brought it back for ws server via `max_connections`.

core/lib/zksync_core/src/api_server/web3/backend_jsonrpsee/batch_limiter_middleware.rs
montekki: I've added a new test for this

core/lib/zksync_core/src/api_server/web3/pubsub.rs
Deniallugo: I prefer the third solution. 
With an additional actor. 
It allows us not to care about slow clients at all.

The capacity shouldn't be an issue.

popzxc:
Meta-comment: I don't think it's a chore, as it contains a fair chunk of new functionality.

montekki:
@popzxc which is in no way exposed to the user, but I am happy to change to whatever fits

core/lib/zksync_core/src/api_server/web3/pubsub.rs
```diff
@@ -78,18 +93,13 @@ impl PubSubNotifier<typed::Sink<PubSubResult>> {
...
-                    PUB_SUB_METRICS.notify_subscribers_latency[&SubscriptionType::Blocks].start();
-                for sink in self.current_subscribers().await {
-                    for block in new_blocks.iter().cloned() {
-                        if sink.notify(Ok(PubSubResult::Header(block))).is_err() {
-                            // Subscriber disconnected.
-                            break;
-                        }
-                        PUB_SUB_METRICS.notify[&SubscriptionType::Blocks].inc();
+                for new_block in new_blocks {
+                    if !self.sender.receiver_count().is_zero() {
```
slowli, line 97: This seems prone to TOCTOU issues: broadcast receivers may appear or disappear during the call. IMO, it would be more straightforward just to remove this condition.

Sort of related: One thing that makes `broadcast` channel different from the previous implementation is that a receiver can now receive *a part* of changes produced during a single `PubSubNotifier` iteration. I don't think it could be avoided, but it seems benign (although logs may require some thinking – is it OK if a subscriber receives logs starting from a middle of a certain miniblock?).

core/lib/zksync_core/src/api_server/web3/pubsub.rs
```diff
@@ -213,23 +213,23 @@ impl PubSubNotifier<(typed::Sink<PubSubResult>, PubSubFilter)> {
...
+    blocks: broadcast::Sender<BlockHeader>,
+    transactions: broadcast::Sender<H256>,
+    logs: broadcast::Sender<Log>,
     events_sender: Option<mpsc::UnboundedSender<PubSubEvent>>,
 }
 
 impl EthSubscribe {
-    pub fn new(runtime_handle: tokio::runtime::Handle) -> Self {
+    pub fn new() -> Self {
+        let (blocks, _rx_blocks) = broadcast::channel(BROADCAST_CHANNEL_CAPACITY);
```
slowli, line 225: Nit: You can use `_` vars here. Using `_something` implies that it's important that the variable isn't *immediately* dropped (unlike `_`, `_something` vars live until the end of the scope), which doesn't seem to be the case here.

core/lib/zksync_core/src/api_server/web3/pubsub.rs
```diff
@@ -238,118 +238,174 @@ impl EthSubscribe {
...
+    }
+
+    async fn heads_subscriber(
+        pending_sink: PendingSubscriptionSink,
+        b: broadcast::Receiver<BlockHeader>,
+    ) {
+        let _guard = PUB_SUB_METRICS.active_subscribers[&SubscriptionType::Blocks].inc_guard(1);
+        let sink = pending_sink.accept().await.unwrap();
+        let closed = sink.closed().fuse();
+        let mut b = BroadcastStream::new(b);
```
slowli, line 257: You can probably use inherent `broadcast::Receiver` methods here (i.e., [`recv`](https://docs.rs/tokio/latest/tokio/sync/broadcast/struct.Receiver.html#method.recv)); it's not necessary to convert it to a `Stream`.

core/lib/zksync_core/src/api_server/web3/pubsub.rs
```diff
@@ -238,118 +238,174 @@ impl EthSubscribe {
...
+        loop {
+            tokio::select! {
+                new_log = b.next() => {
+                    if let Some(Ok(new_log)) = new_log {
+                        sink.send_timeout(
+                            SubscriptionMessage::from_json(&PubSubResult::Header(new_log))
+                            .expect("PubSubResult always serializable to json; qed"),
+                            SUBSCRIPTION_SINK_SEND_TIMEOUT,
+                        )
+                            .await.ok();
```
slowli, line 269: - Nit: incorrect formatting (Rust formatter is not very good at formatting inside macros like `tokio::select!` in this case).
- This implies that some events for a subscriber may be dropped, which I don't think is reasonable. Either we should stop the task on a timeout, or use `send()` (i.e., wait indefinitely).

core/lib/zksync_core/src/api_server/web3/pubsub.rs
```diff
@@ -238,118 +238,174 @@ impl EthSubscribe {
...
+            ErrorCode::InvalidParams.code(),
+            "Rejecting subscription - invalid parameters provided.",
+            None,
+        ))
+        .await;
+    }
+
+    async fn heads_subscriber(
+        pending_sink: PendingSubscriptionSink,
+        b: broadcast::Receiver<BlockHeader>,
```
slowli, line 252: I'd suggest to use `PubSubResult` as the channel element type so that this logic can be reused for blocks and transactions.

core/lib/zksync_core/src/api_server/web3/pubsub.rs
```diff
@@ -388,3 +445,17 @@ impl EthSubscribe {
...
+impl EthPubSubServer for EthSubscribe {
+    async fn subscribe(
+        &self,
+        pending: PendingSubscriptionSink,
+        sub_type: String,
+        filter: Option<PubSubFilter>,
+    ) -> SubscriptionResult {
+        self.sub(pending, sub_type, filter).await;
+
+        Ok(())
```
slowli, line 459: Dumb question: Does `jsonrpsee` make any assumptions as to when this method will be resolved (e.g., is `PendingSubscriptionSink` required to be accepted / rejected?). In any case, IMO, it'd make sense to resolve the sync *before* spawning a Tokio task for it.

core/lib/zksync_core/src/api_server/web3/pubsub.rs
```diff
@@ -238,118 +238,174 @@ impl EthSubscribe {
...
+        let _guard = PUB_SUB_METRICS.active_subscribers[&SubscriptionType::Logs].inc_guard(1);
+        let sink = pending_sink.accept().await.unwrap();
+        let closed = sink.closed().fuse();
+        let mut b = BroadcastStream::new(b);
+        tokio::pin!(closed);
+
+        loop {
+            tokio::select! {
+                new_log = b.next() => {
+                    let new_log = new_log.unwrap().unwrap();
```
slowli, line 326: I don't think `unwrap()`s are safe here; we may get `None` if the notifier task terminates, and an error if `broadcast::Receiver` is too far behind.

core/lib/zksync_core/src/api_server/web3/tests/mod.rs
slowli: That's what I meant by "hav[ing] doubts it worked as intended". The EN doc comment on `subscriptions_limit` reads: "Max possible limit of subscriptions to be in the API state at once", which setting the number of *connections* won't do (there may be multiple subscriptions per connection). Or am I missing something, and `max_connections` actually configures the number of subscriptions?

core/lib/zksync_core/src/api_server/web3/pubsub.rs
montekki: done

core/lib/zksync_core/src/api_server/web3/pubsub.rs
montekki: done

core/lib/zksync_core/src/api_server/web3/pubsub.rs
montekki: done

core/lib/zksync_core/src/api_server/web3/pubsub.rs
montekki: removed.

montekki: > s it OK if a subscriber receives logs starting from a middle of a certain miniblock?
No it's not ok

core/lib/zksync_core/src/api_server/web3/pubsub.rs
montekki: formatted

core/lib/zksync_core/src/api_server/web3/pubsub.rs
montekki: resolving moved before spawning

core/lib/zksync_core/src/api_server/web3/pubsub.rs
montekki: removed

core/lib/zksync_core/src/api_server/web3/pubsub.rs
montekki: @slowli wdyt about sending atomically a `Vec<PubSubEvent>` here? That would make it so that no matter what the client consistently sees _all_ things that have happend in a given block.

slowli: Yep, that seems the most reasonable option.

montekki: Done

core/lib/zksync_core/src/api_server/web3/tests/mod.rs
montekki: No I think that those two were used interchangably bc the assumption was that statistically they are more or less equal.

popzxc:
> which is in no way exposed to the user, but I am happy to change to whatever fits

@montekki people start to hack around the code too (and will do that more with zkStack), so we should start to treat the API our _code_ provides too; I'd propose it to be a `feat` so that this thing will appear in changelog.

core/lib/zksync_core/src/api_server/web3/pubsub.rs
```diff
@@ -238,118 +228,143 @@ impl EthSubscribe {
...
 
-    fn reject(subscriber: typed::Subscriber<PubSubResult>) {
-        subscriber
-            .reject(Error {
-                code: ErrorCode::InvalidParams,
-                message: "Rejecting subscription - invalid parameters provided.".into(),
-                data: None,
-            })
-            .unwrap();
+    async fn subscriber(
```
slowli, line 240: Bikeshedding: Non-getter methods are usually named using verb phrases. In this case, `run_subscriber` would be better.

core/lib/zksync_core/src/api_server/web3/pubsub.rs
```diff
@@ -238,118 +228,143 @@ impl EthSubscribe {
...
+        sink: SubscriptionSink,
+        subscription_type: SubscriptionType,
+        mut b: broadcast::Receiver<Vec<PubSubResult>>,
+        filter: Option<PubSubFilter>,
+    ) -> Result<(), anyhow::Error> {
```
slowli, line 245: Bikeshedding: AFAIU, the error here gets ignored, which IMO looks weird; i.e., errors are used here just for short-circuiting the task. I'd suggest to have the void return value and treat errors in `new_items` and `send_timeout` explicitly. As potential future work, we may meter the number of subscriptions that were closed because of lagging `broadcast::Receiver` and because of a timeout on `send_timeout`; more explicit error handling would help with this.

core/lib/zksync_core/src/api_server/web3/pubsub.rs
```diff
@@ -238,118 +228,143 @@ impl EthSubscribe {
...
+        let _guard = PUB_SUB_METRICS.active_subscribers[&subscription_type].inc_guard(1);
+        let closed = sink.closed().fuse();
+        tokio::pin!(closed);
+
+        loop {
+            tokio::select! {
+                new_items = b.recv() => {
+                    let new_items = new_items?;
+                    for item in new_items {
```
slowli, line 254: Bikeshedding: IMO, it'd make sense to extract items handling to a separate function:

```rust
async fn handle_new_items(
    sink: &SubscriptionSink,
    subscription_type: SubscriptionType,
    new_items: Vec<PubSubResult>,
    filter: Option<&PubSubFilter>,
) -> Result<(), SendTimeoutError> {
    // ...
}
```

This would help readability and minimize the amount of code inside `tokio::select!` with potential formatting problems.

core/lib/zksync_core/src/api_server/web3/tests/mod.rs
slowli: IMO, it'd make sense to rename the config param to "connections limit" so that it's less confusing. Can be done separately, of course.

core/lib/zksync_core/src/api_server/web3/tests/ws.rs
```diff
@@ -72,7 +74,10 @@ trait WsTest {
     ) -> anyhow::Result<()>;
 }
 
-async fn test_ws_server(test: impl WsTest) {
+async fn test_ws_server(
+    test: impl WsTest,
+    websocket_requests_per_minute_limit: Option<NonZeroU32>,
```
slowli, line 79: Bikeshedding: Instead of introducing this parameter, you could extend `WsTest` with the corresponding getter:

```rust
trait WsTest {
    fn websocket_requests_per_minute_limit(&self) -> Option<NonZeroU32> {
        None
    }
    // ...
}
```

In this way, less existing code would be changed.

slowli (approved):
I wonder how well the new approach behave under load, but the code looks nice.

core/lib/zksync_core/src/api_server/web3/tests/ws.rs
montekki: done

core/lib/zksync_core/src/api_server/web3/pubsub.rs
montekki: done

core/lib/zksync_core/src/api_server/web3/pubsub.rs
montekki: explicit error treating done

core/lib/zksync_core/src/api_server/web3/pubsub.rs
montekki: renamed

core/lib/zksync_core/src/api_server/web3/mod.rs
```diff
@@ -730,24 +509,38 @@ impl<G: 'static + Send + Sync + L1GasPriceProvider> ApiBuilder<G> {
...
+            .max_response_body_size(response_body_size_limit)
+            .set_batch_request_config(batch_request_config);
+
+        let (local_addr, server_handle) = if is_http {
+            // HTTP-specific settings
+            let server = server_builder
+                .http_only()
+                .build(addr)
+                .await
+                .with_context(|| format!("Failed building {transport_str} JSON-RPC server"))?;
```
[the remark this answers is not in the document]
montekki: done

core/lib/zksync_core/src/api_server/web3/mod.rs
```diff
@@ -730,24 +509,38 @@ impl<G: 'static + Send + Sync + L1GasPriceProvider> ApiBuilder<G> {
...
+            (server.local_addr(), server.start(rpc))
         } else {
-            ServerBuilder::default().ws_only()
+            // WS specific settings
+            let server = server_builder
+                .set_rpc_middleware(RpcServiceBuilder::new().layer_fn(move |a| {
+                    LimitMiddleware::new(a, websocket_requests_per_minute_limit)
+                }))
+                .set_id_provider(EthSubscriptionIdProvider)
+                .max_connections(subscriptions_limit.unwrap_or(5_000) as u32)
```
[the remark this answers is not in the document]
montekki: yeaaah so probably let's leave it for another PR, for some reason we used to set it without any options to `5_000` for HTTP and optionally apply if it exists the limit to websocket or also set it to `5_000`. I suggest reworking it in a separate PR that would involve two new values for config:
1. optional limit for http that would actually applicable to server
2. renamed parameter for websocket

Or make the same parameter with sane name that would be applied to both protocols.

core/lib/zksync_core/src/api_server/web3/tests/mod.rs
montekki: yes let's do it separately because that would lead to deployments change

popzxc (approved):
Nice, well done.

PR #694, opened 2023-12-15
feat: Remove data fetchers

popzxc:
## What ❔

Removes the data fetchers from the codebase.
The tables they've been writing to are left intact since we need to support the RPC methods that interact with them for a while.
This way these methods will be accessing the latest available data there.

Additionally, as discussed before, `get_well_known_token_addresses` no longer checks for `well_known = true`.

## Why ❔

This component is deprecated.

## Checklist


core/lib/dal/src/tokens_dal.rs
```diff
@@ -78,13 +78,12 @@ impl TokensDal<'_, '_> {
         }
     }
 
-    pub async fn get_well_known_token_addresses(&mut self) -> Vec<(Address, Address)> {
+    pub(crate) async fn get_well_known_token_addresses(&mut self) -> Vec<(Address, Address)> {
```
perekopskiy, line 81: There is also `get_well_known_tokens` in `tokens_web3_dal` where we should remove filter by `well_known`

popzxc: Done.

perekopskiy (approved):

Deniallugo (approved):

Deniallugo (approved):

hatemosphere (approved):
CI changes LGTM

PR #696, opened 2023-12-15
perf: reduce memory consumption of witness generation

joonazan:
## What ❔

Uses an [updated zkevm_test_harness](https://github.com/matter-labs/era-zkevm_test_harness/pull/57) that uses less memory. The updated crate takes callback functions instead of returning a vector of all results.

We may have to think about how to not keep the database open the whole time the callbacks are coming in.

## Why ❔

Proving batches currently uses memory proportional to the batch size, which is not ok for large batches.


RomanBrodetski (approved):
Merging this to test on stage, will revert if not working

prover/witness_generator/src/basic_circuits.rs
```diff
@@ -185,7 +180,10 @@ impl JobProcessor for BasicWitnessGenerator {
     const SERVICE_NAME: &'static str = "fri_basic_circuit_witness_generator";
 
     async fn get_next_job(&self) -> anyhow::Result<Option<(Self::JobId, Self::Job)>> {
-        let mut prover_connection = self.prover_connection_pool.access_storage().await.unwrap();
+        let block_number = L1BatchNumber(22);
+        let job = get_artifacts(block_number, &*self.object_store).await;
+        Ok(Some((block_number, job)))
+        /*let mut prover_connection = self.prover_connection_pool.access_storage().await.unwrap();
```
RomanBrodetski, line 186: wait, so will it only compute the proof for block `22`?

RomanBrodetski (changes requested):
To test on stage please revert the change that hardcoded the target l1 batch. Alternativelly we can introduce a config field `basic_witness_generation_target_l1_batch_override`

prover/witness_generator/src/basic_circuits.rs
RomanBrodetski: I updated the file, let's see if it builds

RomanBrodetski (approved):

RomanBrodetski (approved):

PR #701, opened 2023-12-18
feat: Transfer event removal

Artemka374:
## What ❔

Divide web3 server API into 2:
* Old one
* Test-drive API, which won't return ETH Transfer events

API that will be run per instance is defined via config:
- API_WEB3_JSON_RPC_ETH_TRANSFER_EVENTS(disabled/enabled) modes
- EN_ETH_TRANSFER_EVENTS(disabled/enabled) modes

Added new columns to `events` database to reassign indexes of logs correctly(added SQL migrations and Rust migrations, mostly inspired by https://github.com/matter-labs/zksync-era/pull/688):
- `event_index_in_block_without_eth_transfer`
- `event_index_in_tx_without_eth_transfer`

## Why ❔

To be EVM compatible.

## Checklist


core/lib/dal/src/events_web3_dal.rs
```diff
@@ -111,7 +128,11 @@ impl EventsWeb3Dal<'_, '_> {
         }
     }
 
-    fn build_get_logs_where_clause(&self, filter: &GetLogsFilter) -> (String, u8) {
+    fn build_get_logs_where_clause(
+        &self,
+        filter: &GetLogsFilter,
+        skip_transfer_event: bool,
```
montekki, line 134: I would suggest turning this boolean value into an `enum` something like

```rust
enum ApiMode {
  Legacy,
  Modern,
 }
 ```
 
For improved readability (which is which?) and future-proofing (agianst situations where we will be running more than two apis (we will).

File: core/lib/dal/src/events_web3_dal.rs [hunk 2]
```diff
@@ -127,65 +148,64 @@ impl EventsWeb3Dal<'_, '_> {
...
+                    WHERE miniblock_number > $1 {}
+                    ORDER BY miniblock_number ASC, event_index_in_block ASC
+                )
+                SELECT miniblocks.hash as "block_hash",
+                    address as "address", topic1 as "topic1", topic2 as "topic2", topic3 as "topic3", topic4 as "topic4", value as "value",
+                    miniblock_number as "miniblock_number", miniblocks.l1_batch_number as "l1_batch_number", tx_hash as "tx_hash",
+                    tx_index_in_block as "tx_index_in_block", event_index_in_block as "event_index_in_block", event_index_in_tx as "event_index_in_tx"
+                FROM events_select
+                INNER JOIN miniblocks ON events_select.miniblock_number = miniblocks.number
+                ORDER BY miniblock_number ASC, event_index_in_block ASC
```
montekki, line 191: Probably insubstantial (to the topic of the PR) formatting fixes should not be included (+this may make sql formatting checker unhappy?)

core/lib/dal/src/events_web3_dal.rs
```diff
@@ -127,65 +148,64 @@ impl EventsWeb3Dal<'_, '_> {
...
+            );
+
+            let mut query = sqlx::query_as(query.as_str());
+
+            query = query.bind(from_block.0 as i64);
+
+            if skip_transfer_event {
+                query = query.bind(L2_ETH_TOKEN_ADDRESS.as_bytes());
+                query = query.bind(TRANSFER_EVENT_TOPIC.as_bytes());
+            }
```
montekki, line 203: Judging by the code above, these are always constant so you do not have to parametrise the query and can embed these values there directly?

RomanBrodetski:
this is definitely not a `chore`. Remember that `chore` doesn't end up in changelog. This is a `feat`.

core/lib/dal/src/events_web3_dal.rs
```diff
@@ -97,21 +106,56 @@ impl EventsWeb3Dal<'_, '_> {
...
+                        || log.topic1 != TRANSFER_EVENT_TOPIC.as_bytes()
+                });
+
+                if !db_logs.is_empty() {
+                    let mut new_event_index_in_block = 0;
+                    let mut new_event_index_in_tx = 0;
+                    let mut current_block = db_logs[0].miniblock_number;
+                    let mut current_tx_hash = db_logs[0].tx_hash.clone();
+
+                    for i in 0..db_logs.len() {
```
slowli, line 132: Nit: It is more idiomatic to iterate over mutable reference to logs: `for db_log in &mut db_logs`.

core/lib/dal/src/events_web3_dal.rs
```diff
@@ -127,15 +171,25 @@ impl EventsWeb3Dal<'_, '_> {
             arg_index += 1;
         }
 
+        if api_mode == APIMode::Modern {
+            where_sql += &format!(
+                " AND NOT (address = ${} AND topic1 = ${})",
+                arg_index,
+                arg_index + 1
+            );
```
slowli, line 179: Dumb question: Is speedup from having this as a part of the query that large that it justifies breaking encapsulation? IMO, from the encapsulation perspective, it would make sense to leave DAL as-is and perform event filtering and index translation on the server level. Is this infeasible because of the performance degradation?

core/lib/dal/src/events_web3_dal.rs
```diff
@@ -97,21 +106,56 @@ impl EventsWeb3Dal<'_, '_> {
...
                 .with_arg("filter", &filter)
                 .with_arg("limit", &limit)
                 .fetch_all(self.storage.conn())
                 .await?;
+
+            if api_mode == APIMode::Modern {
+                db_logs.retain(|log| {
+                    log.address != L2_ETH_TOKEN_ADDRESS.as_bytes()
+                        || log.topic1 != TRANSFER_EVENT_TOPIC.as_bytes()
+                });
```
slowli, line 124: Dumb question: isn't such events filtered out on the DB query level?

core/lib/dal/src/events_web3_dal.rs
```diff
@@ -185,7 +239,38 @@ impl EventsWeb3Dal<'_, '_> {
             )
             .fetch_all(self.storage.conn())
             .await?;
+
+            if api_mode == APIMode::Modern {
```
slowli, line 243: Is it possible to extract shared code mapping logs between this method, `get_logs()` and `get_transaction_receipt`?

core/lib/zksync_core/src/api_server/web3/tests/mod.rs
```diff
@@ -283,7 +342,29 @@ async fn store_events(
             indexed_topics: vec![H256::repeat_byte(42), H256::repeat_byte(111)],
             value: (start_idx + 3).to_le_bytes().to_vec(),
         },
+        // ETH Transfer event with matching both address and topic
+        VmEvent {
```
slowli, line 346: Bikeshedding: IMO, it would be better to have a dedicated setup for ETH transfer events; i.e., a method like `store_eth_transfer_events()`. To avoid copy-paste, common code between `store_events` and `store_eth_transfer_events` can be extracted to a method that accepts `&[VmEvent]` as an argument. Or, you can modify `store_events` to accept such an argument and create methods to generate "default" and ETH transfer events.

core/lib/zksync_core/src/api_server/web3/tests/mod.rs
```diff
@@ -157,18 +210,23 @@ async fn spawn_server(
...
 }
 
 #[async_trait]
 trait HttpTest {
-    async fn test(&self, client: &HttpClient, pool: &ConnectionPool) -> anyhow::Result<()>;
+    async fn test(
+        &self,
+        client: &HttpClient,
+        pool: &ConnectionPool,
+        api_mode: APIMode,
```
slowli, line 225: This doesn't look right. A test should know on its own which kind of the server it's run with. So, you may add the corresponding getter:

```rust
trait HttpTest {
    fn api_mode(&self) -> ApiMode;
}
```

If you want to run some test for both API modes, you can do that by having it defined like this:

```rust
struct ParametricTest(ApiMode);

#[async_trait]
impl HttpTest for ParametricTest {
    fn api_mode(&self) -> ApiMode { self.0 }

    async fn test(/* ... */) {
        // Can access `self.0` if necessary.
    } 
}

#[test_casing(2, [ApiMode::Legacy, ApiMode::Modern])
#[tokio::test]
async fn parametric_test(mode: ApiMode) {
    test_http_server(ParametricTest(mode)).await;
}
```

core/lib/types/src/api/mod.rs
```diff
@@ -19,6 +19,14 @@ use crate::{
 
 pub mod en;
 
+/// Enum for choosing API mode
+#[derive(Copy, Clone, Debug, PartialEq, Display, Default)]
+pub enum APIMode {
```
slowli, line 24: Bikeshedding: [Rust guidelines recommend](https://rust-lang.github.io/rust-clippy/master/index.html#/upper_case_acronyms) not capitalizing acronyms; i.e., it should be `ApiMode`.

core/lib/types/src/api/mod.rs
```diff
@@ -19,6 +19,14 @@ use crate::{
...
+    #[default]
+    Modern,
+    Legacy,
```
slowli, line 27: Bikeshedding: IMO, these variant names (and possibly, the enum name) are too vague. Do we plan to introduce other changes in the API in the future? If not (i.e., variability is confined to ETH transfer event handling), then at least variant naming should reflect that, e.g., `EthTransfersIncluded` / `EthTransfersExcluded`.

core/bin/external_node/src/main.rs
```diff
@@ -279,11 +280,41 @@ async fn init_tasks(
             .with_tx_sender(tx_sender.clone(), vm_barrier.clone())
             .with_sync_state(sync_state.clone())
             .enable_api_namespaces(config.optional.api_namespaces())
-            .build(stop_receiver.clone())
+            .build(stop_receiver.clone(), APIMode::Modern)
             .await
             .context("Failed initializing HTTP JSON-RPC server")?;
 
+    let legacy_http_server_handles =
+        ApiBuilder::jsonrpsee_backend(config.clone().into(), connection_pool.clone())
```
slowli, line 288: - `ApiBuilder` may be made `Clone` to avoid this copy-paste.
- Dumb question: Do we *want* to run 2 instances of each API server? IMO, it makes little sense; instead, `ApiMode` should be a config param, and each server should run as a single instance taking this param into account.

slowli (changes requested):

core/lib/types/src/api/mod.rs
Artemka374: I am not sure about it tbh, but from what @montekki said, I think yes: 

> For improved readability (which is which?) and future-proofing (agianst situations where we will be running more than two apis (we will).

core/bin/external_node/src/main.rs
Artemka374: I guess it makes sense

core/lib/types/src/api/mod.rs
slowli: In this case, `Legacy` and `EthTransfersExcluded` would be better.

core/lib/dal/src/events_web3_dal.rs
Artemka374: Right here they are not. The thing is, that we need to somehow deal with indexes, so there will be no gaps if we get only filtered events. Generally we can put this filtering inside query, but it seems a bit more painful than doing it this way

.github/workflows/ci-core-reusable.yml
```diff
@@ -343,3 +452,123 @@ jobs:
...
+
+      - name: Show upgrade.log logs
+        if: always()
+        run: ci_run cat core/tests/upgrade-test/upgrade.log || true
+
+      - name: Show sccache logs
+        if: always()
+        run: |
+          ci_run sccache --show-stats
+          ci_run cat /tmp/sccache_log.txt
```
popzxc, line 574: Nit: Missing empty line.

core/lib/types/src/api/mod.rs
```diff
@@ -19,6 +19,39 @@ use crate::{
...
+                    "eth_transfer_included" => ApiMode::EthTransferIncluded,
+                    _ => return Err(E::custom(format!("Unsupported API mode: {}", value))),
+                };
+
+                Ok(result)
+            }
+        }
+        deserializer.deserialize_str(V)
+    }
+}
```
popzxc, line 53: I guess you can use [`#[serde(rename_all = "snake_case")]`](https://serde.rs/container-attrs.html#rename_all).

core/lib/config/Cargo.toml
```diff
@@ -11,6 +11,7 @@ categories = ["cryptography"]
 
 [dependencies]
 zksync_basic_types = { path = "../../lib/basic_types" }
+zksync_types = { path = "../../lib/types" }
```
popzxc, line 14: `zksync_types` is a very heavyweight (and dangerous, w.r.t. the vast amount of exposed APIs) crate, so I'd prefer to not depend on it. Instead, you can define your own mode here, and implement the conversion between types where required manually. It'll be slightly more boilerplate, but it'd be safer.

core/lib/types/src/api/mod.rs
```diff
@@ -19,6 +19,39 @@ use crate::{
 
 pub mod en;
 
+/// Enum for choosing API mode
+#[derive(Copy, Clone, Debug, PartialEq, Display, Default)]
+pub enum ApiMode {
+    #[default]
+    Modern,
+    EthTransferIncluded,
```
popzxc, line 27: I'm not sure if the names respect well the intent. Modern doesn't say much to someone who is familiar with the topic already. Probably `EthTransferFiltered` and `EthTransferIncluded` would be better.

popzxc: Actually, 1) I do not think that we're going to support more modes for our API, and 2) this is a temporary measure until we implement a proper protocol upgrade, so maybe it's more generic than needed.
Maybe it'll be even better to have an enum named `ApiEthTransferEvents` and values `Enabled` and `Disabled`.
This way this will be an exhaustive list of supported options, and you won't have to specify that "Currently we support these modes", as there may be only these 2.

core/tests/ts-integration/src/modifiers/balance-checker.ts
```diff
@@ -194,44 +197,75 @@ class ShouldChangeBalance extends MatcherModifier {
...
-        throw {
-            message: `No fee log was found in the following transaction receipt`,
-            receipt
-        };
-    }
 
-    const feeAmount = ethers.BigNumber.from(feeLog.data);
+    const apiMode = process.env.EN_API_MODE ?? process.env.API_WEB3_JSON_RPC_API_MODE;
+
+    if (apiMode == 'modern') {
```
popzxc, line 216: Wouldn't this approach work for "legacy" API as well? If so, why do we need both?

core/tests/ts-integration/src/modifiers/balance-checker.ts
```diff
@@ -194,44 +197,75 @@ class ShouldChangeBalance extends MatcherModifier {
...
-    });
-    if (!feeLog) {
-        throw {
-            message: `No fee log was found in the following transaction receipt`,
-            receipt
-        };
-    }
 
-    const feeAmount = ethers.BigNumber.from(feeLog.data);
+    const apiMode = process.env.EN_API_MODE ?? process.env.API_WEB3_JSON_RPC_API_MODE;
```
popzxc, line 214: This kind of implicit checks often becomes a huge PITA to fix, so I'd be extremely cautious about it and generally avoid at all costs.

popzxc (commented):
Review is pretty shallow, I expect someone else to make sure that the acceptance criteria is met.

core/tests/ts-integration/src/modifiers/balance-checker.ts
Artemka374: I am pretty sure that it will, but I think it is good to have separate tests for both API modes to understand and test difference between them

popzxc: But this isn't really a test, it's a modifier that is to be used by tests, so I'd expect it to do 1 thing in 1 way.
If you want to make sure that the second approach works, probably you can just extract it to the test which won't use this modifier and instead will make the calculations based on the 2nd approach.

.github/workflows/ci-core-reusable.yml
```diff
@@ -225,7 +226,115 @@ jobs:
           ci_run sccache --show-stats
           ci_run cat /tmp/sccache_log.txt
 
-  external-node:
+  integration-old-api:
```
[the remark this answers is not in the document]
Artemka374: Cool, didn't know about it, I will try

core/lib/dal/src/events_web3_dal.rs
```diff
@@ -13,16 +14,61 @@ pub struct EventsWeb3Dal<'a, 'c> {
...
+                if db_log.miniblock_number != current_block {
+                    current_block = db_log.miniblock_number;
+                    new_event_index_in_block = 0;
+                }
+                if db_log.tx_hash != current_tx_hash {
+                    current_tx_hash = db_log.tx_hash.clone();
+                    new_event_index_in_tx = 0;
+                }
+                db_log.event_index_in_block = new_event_index_in_block;
+                db_log.event_index_in_tx = new_event_index_in_tx;
```
perekopskiy, line 43: Maybe a stupid question, but how will it work when not all logs from block are queried? I.e. if block has 2 events: first with topic1=A, second with topic1=B, both events are not ETH transfer. If someone sends a request with filter.topic1=B, won't the response will have the only log with event_index_in_block=0, while it should be 1?

Artemka374: That's actually a good question, I didn't really think about it, and yes, it won't work properly. I will think how to fix that behaviour, but first question that comes to my mind - do you know how bad could it be if we try to filter topics/addresses on Rust side?

perekopskiy: hmm, I don't think it's possible with the current approach, we don't limit block range at all, so worst case we need to filter millions of events

perekopskiy: One way to make it work without too much changes:
query ETH transfer events regardless provided filter and then count how many ETH transfers were before each event within block/tx and subtract this number from indices

core/lib/zksync_core/src/state_keeper/io/event_indexes_migration.rs
```diff
@@ -0,0 +1,378 @@
+//! Temporary module for migrating fee addresses from L1 batches to miniblocks.
+
+use std::time::Duration;
+
+use anyhow::Context as _;
+use tokio::sync::watch;
+use zksync_dal::{ConnectionPool, StorageProcessor};
+use zksync_types::MiniblockNumber;
+
+/// Runs the migration for non-pending miniblocks. Should be run as a background task.
```
slowli, line 10: Nit: In the context of the fee address migration, "non-pending miniblocks" means miniblocks belonging to a sealed L1 batch. This distinction is made because the fee address is copied to miniblocks from the corresponding L1 batch record, which is not present in Postgres for pending miniblocks. AFAIU, this distinction is not applicable for this migration.

core/lib/dal/src/models/storage_event.rs
```diff
@@ -86,3 +86,65 @@ impl From<StorageL2ToL1Log> for L2ToL1Log {
...
+    pub tx_index_in_block: i32,
+    pub event_index_in_block: i32,
+    pub event_index_in_tx: i32,
+    pub event_index_in_block_without_eth_transfer: i32,
+    pub event_index_in_tx_without_eth_transfer: i32,
+}
+
+impl StorageWeb3LogExt {
+    pub fn into_storage_log(self, api_eth_transfer_events: ApiEthTransferEvents) -> StorageWeb3Log {
+        let event_index_in_block = match api_eth_transfer_events {
```
slowli, line 111: Dumb question: What happens if this conversion is called with `ApiEthTransferEvents::Disabled` for non-migrated logs? Note that the latest version of #688 [contains dedicated logic](https://github.com/matter-labs/zksync-era/pull/688/commits/393a96106bc9e9d75a4abb3f711883c522b31c79) in the DAL crate to return accurate data in this case. Not sure whether this is a concern in this case though; theoretically, we could only spin up servers with `ApiEthTransferEvents::Disabled` after the migration is complete.

core/lib/config/src/configs/api.rs
```diff
@@ -232,3 +239,12 @@ impl MerkleTreeApiConfig {
         3_072
     }
 }
+
+/// Enum for choosing API mode
```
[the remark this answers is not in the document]
Artemka374: Uh, forgot to change it here

core/lib/zksync_core/src/state_keeper/io/event_indexes_migration.rs
Artemka374: Yes, this one was mostly copy-pasted from fee address migration PR and then adjusted, but in the context of these changes, new events are saved already correctly. So that is just comment being not updated

core/lib/dal/src/models/storage_event.rs
Artemka374: I will think about how to resolve this issue, thanks for pointing out, but generally I think we can't assign the indexes without getting logs of full miniblock, so I think that we either try to call assigning everything at that point, or to do as you said

core/lib/dal/src/events_dal.rs
```diff
@@ -217,9 +235,170 @@ impl EventsDal<'_, '_> {
...
+impl EventsDal<'_, '_> {
+    /// Method assigns indexes, that should be present without ETH transfer events for all the events in the given range.
+    /// Note, that indexes are assigned starting from 1 to understand whether indexes were migrated.
+    pub async fn assign_indexes_without_eth_transfer(
+        &mut self,
+        numbers: ops::RangeInclusive<MiniblockNumber>,
+    ) -> sqlx::Result<u64> {
+        let count_eth_transfer_events = self
+            .assign_eth_transfer_event_indexes(numbers.clone())
+            .await?;
```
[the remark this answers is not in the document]
Artemka374: I wouldn't say so, at least it doesn't really matter in what order or when exactly they will be executed, that's just division by type of events

slowli:
A really late question, sorry for not raising it before. What if the event filtering is controlled via an additional field in `Filter` (used in `eth_newFilter` and `eth_getLogs`) / `PubSubFilter` (used in subscriptions) with a default value? In this case, a single API server could serve both new and legacy API (we may need to spin up 2 notifier tasks for logs instead of one; one with filtering and the other without it), and filtering would be controlled by the client. Would this contradict some other requirements / not work / require too much effort to integrate into Web3 clients? (I'd guess they should have interceptor functionality that would allow doing that, but I don't have much experience using Web3 clients.)

cc @RomanBrodetski @montekki @popzxc

core/lib/dal/src/events_dal.rs
```diff
@@ -217,9 +234,169 @@ impl EventsDal<'_, '_> {
...
+            TRANSFER_EVENT_TOPIC.as_bytes()
+        )
+        .execute(self.storage.conn())
+        .await?;
+
+        Ok(result.rows_affected())
+    }
+
+    /// Method checks, whether any event in the miniblock has `index_without_eth_transfer` equal to NULL, which means that indexes are not migrated.
+    pub async fn are_event_indexes_migrated(
```
slowli, line 340: Is it strictly necessary to check all miniblocks in the range? Note that in #688, it's sufficient to check only the starting miniblock of the range provided that ranges are guaranteed to be identical across node restarts. Would a similar assumption not apply here? AFAIU, checking the entire range would put significantly more load on the DB if the range is reasonably large.

core/lib/dal/src/events_dal.rs
```diff
@@ -217,9 +234,169 @@ impl EventsDal<'_, '_> {
...
+    async fn assign_eth_transfer_event_indexes(
+        &mut self,
+        numbers: ops::RangeInclusive<MiniblockNumber>,
+    ) -> sqlx::Result<u64> {
+        let result = sqlx::query!(
+            r#"
+            UPDATE events
+            SET
+                event_index_in_block_without_eth_transfer = 0,
+                event_index_in_tx_without_eth_transfer = 0
```
slowli, line 318: Just to clarify: Are potential duplicate values for `event_index_*` fields OK?

core/lib/zksync_core/src/state_keeper/io/event_indexes_migration.rs
```diff
@@ -0,0 +1,395 @@
...
+            .blocks_dal()
+            .insert_miniblock(&new_miniblock)
+            .await?;
+        let tx_location = IncludedTxLocation {
+            tx_hash: H256::repeat_byte(1),
+            tx_index_in_miniblock: 0,
+            tx_initiator_address: Address::repeat_byte(2),
+        };
+        let events = vec![
+            // Matches address, doesn't match topics
```
slowli, line 137: Nit: These comments don't make much sense here, and I'd argue that it would be sufficient to retain only the last 3 events (i.e., ones related to filtering logic).

popzxc:
> What if the event filtering is controlled via an additional field in Filter (used in eth_newFilter and eth_getLogs) / PubSubFilter (used in subscriptions) with a default value?

@slowli I believe most people use SDKs (ethers and such) for that, and in most cases modifying the logic of interacting with filters is hidden. Probably it won't work.

slowli:
@popzxc For ethers in particular, the interceptor abstraction I've talked about is encapsulated in the `Provider` interface. Specifically, `BaseProvider` (a base for other providers) [defines `_getFilter`](https://github.com/ethers-io/ethers.js/blob/0bfa7f497dc5793b66df7adfb42c6b846c51d794/packages/providers/src.ts/base-provider.ts#L1609), which AFAIU maps all event filters processed by the provider (this is in ethers v5 used in integration tests; in ethers v6, it's structured similarly). So it seems possible to do something like this:

```typescript
interface FilterWithEthTransfers extends Filter {
    filterEthTransfers?: boolean;
}

class CustomizableProvider extends zksync.Provider {
    private filterEthTransfers: boolean;

    override async _getFilter(filter: Filter | /* ... */): Promise<FilterWithEthTransfers | /* ... */> {
        const resolvedFilter = <FilterWithEthTransfers>await super._getFilter(filter);
        resolvedFilter.filterEthTransfers = this.filterEthTransfers; // passed to JSON-RPC call
        return resolvedFilter;
    }
}
```

IMO, controlling filtering on the client side makes the approach more maintainable and easier to reason about, both for us and clients. OTOH, I may be unaware of complexities associated with modifying and maintaining client code.

popzxc:
@slowli I think that if anything more complex than just "swap URL" would be required, people would just not bother. Though I may be wrong, cc @RomanBrodetski

core/tests/ts-integration/src/modifiers/balance-checker.ts
```diff
@@ -194,43 +196,26 @@ class ShouldChangeBalance extends MatcherModifier {
...
-    const fromAccountAddress = ethers.utils.hexZeroPad(ethers.utils.arrayify(from), 32);
-    // Fee log is one that sends money to the system contract account.
-    const feeLog = receipt.logs.find((log) => {
-        return log.topics.length == 3 && log.topics[1] == fromAccountAddress && log.topics[2] == systemAccountAddress;
-    });
-    if (!feeLog) {
+export async function extractFee(receipt: zksync.types.TransactionReceipt, wallet: zksync.Wallet): Promise<Fee> {
+    const trace = await wallet!.provider.send('debug_traceTransaction', [receipt.transactionHash]);
+
+    if (!trace) {
```
StanislavBreadless, line 205: why do we need this check? Having the `extractFee` as async function is not bad, but most of the diff in tests comes from it, so just curious

core/lib/dal/src/models/storage_event.rs
```diff
@@ -19,10 +20,49 @@ pub struct StorageWeb3Log {
...
+
+impl From<ExtendedStorageWeb3Log> for Log {
+    fn from(log_with_transfer_events_mode: ExtendedStorageWeb3Log) -> Log {
+        let log = log_with_transfer_events_mode.0;
+        let api_eth_transfer_events = log_with_transfer_events_mode.1;
+
+        let event_index_in_block = match api_eth_transfer_events {
+            ApiEthTransferEvents::Enabled => log.event_index_in_block,
+            ApiEthTransferEvents::Disabled => {
+                log.event_index_in_block_without_eth_transfer.unwrap_or(0)
```
StanislavBreadless, line 56: should it panic? I mean if it returns 0 in API it might lead to worse consequences

StanislavBreadless (approved):
Generally LGTM, I left a few nits

core/tests/ts-integration/src/modifiers/balance-checker.ts
Artemka374: With this method we are checking whether all the balances were updated correctly(along with fees) and since in this new mode we can't just ask for receipt for native transfers - this is the only way to get fee

core/lib/dal/migrations/20240111084243_events-without-eth-transfer-indexes.up.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE events
+    ADD COLUMN IF NOT EXISTS event_index_in_block_without_eth_transfer INT DEFAULT NULL;
+ALTER TABLE events
+    ADD COLUMN IF NOT EXISTS event_index_in_tx_without_eth_transfer INT DEFAULT NULL;
```
RomanBrodetski, line 4: According to quick research

> In Postgres 10 or older (such as OP's 9.6), any DEFAULT clause even DEFAULT NULL will require a full rewrite of the table, which might take essentially forever for large tables[1].

so this is not safe. We can try it on stage2 (although please copy a subset of table first, don't amend `eevents` table there) , but I think it won't work. 

We may find a better migration approach (just remove `DEFAULT`?) but even then we might negatively impact insert time to this table which will result in lower TPS. Is there an alternative solution?

popzxc: We don't use postgres 10 or older, don't we? I thought we're at 14 or at least 12?

popzxc: image

popzxc: E.g. from [the same response](https://stackoverflow.com/questions/47709593/postgresql-add-column-default-null-locks-and-performance) (I guess) below:

> This was changed with Postgres 11, Vao Tsun's answer is only correct for these, [per the Postgres 11 documentation](https://www.postgresql.org/docs/11/sql-altertable.html#SQL-ALTERTABLE-NOTES):
    When a column is added with ADD COLUMN and a non-volatile DEFAULT is specified, the default is evaluated at the time of the statement and the result stored in the table's metadata. That value will be used for the column for all existing rows. If no DEFAULT is specified, NULL is used. In neither case is a rewrite of the table required.
So as long as DEFAULT is a static value or a non-volatile expression the table will not need to be rewritten.

RomanBrodetski: oh sorry I misread it, you are right. Still wanted to ask if anyone sees an alternative solution so that we don't need to change schema, store more data per event and migrate that data as well

popzxc:
@Artemka374 is this PR still relevant or can it be closed?

Artemka374:
I don't know to be honest. The discussion about it was brought up again not that long ago, but I don't remember any decision was made. Maybe @RomanBrodetski or @StanislavBreadless can say more?

StanislavBreadless:
I believe it will be needed in the long term anyway, so better to finish it up and get into production

PR #706, opened 2023-12-18
feat(api): Do not return receipt if tx was not included to the batch

Deniallugo:
## What ❔

For compatibility with Ethereum, we have to return `Receipt` only if tx is included in the miniblock. Block hash is not null null. 

## Why ❔

For compatibility with Ethereum and for not breaking external clients, such as Metemask, they do not expect that block hash could be null

## Checklist

### DIFF
--- a/core/lib/dal/src/transactions_web3_dal.rs
+++ b/core/lib/dal/src/transactions_web3_dal.rs
@@ -50,7 +50,7 @@ impl TransactionsWeb3Dal<'_, '_> {
                     transactions.hash AS tx_hash,
                     transactions.index_in_block AS index_in_block,
                     transactions.l1_batch_tx_index AS l1_batch_tx_index,
-                    transactions.miniblock_number AS block_number,
+                    transactions.miniblock_number AS "block_number!",
                     transactions.error AS error,
                     transactions.effective_gas_price AS effective_gas_price,
                     transactions.initiator_address AS initiator_address,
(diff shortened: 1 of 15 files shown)


core/lib/types/src/api/mod.rs
```diff
@@ -209,7 +209,7 @@ pub struct TransactionReceipt {
     pub transaction_index: Index,
     /// Hash of the block this transaction was included within.
     #[serde(rename = "blockHash")]
-    pub block_hash: Option<H256>,
+    pub block_hash: H256,
```
perekopskiy, line 212: If you're changing the struct then it also makes sense to make `block_number`, `status` required

perekopskiy (dismissed):

core/lib/dal/src/transactions_web3_dal.rs
```diff
@@ -78,21 +78,17 @@ impl TransactionsWeb3Dal<'_, '_> {
             .fetch_optional(self.storage.conn())
             .await?
             .map(|db_row| {
-                let status = match (db_row.block_number, db_row.error) {
-                    (_, Some(_)) => Some(U64::from(0)),
-                    (Some(_), None) => Some(U64::from(1)),
-                    // tx not executed yet
-                    _ => None,
-                };
+                let status = db_row.error.map(|_| U64::zero()).unwrap_or_else(U64::one);
```
perekopskiy, line 81: unwrap_or_else(U64::one) looks weirds, it should be smth like `expect("Mined transaction must have status")`

core/lib/zksync_core/src/api_server/web3/namespaces/eth.rs
```diff
@@ -523,9 +523,7 @@ impl<G: L1GasPriceProvider> EthNamespace<G> {
                     .await
                     .map_err(|err| internal_error(METHOD_NAME, err))
                 {
-                    if main_node_receipt.status == Some(0.into())
-                        && main_node_receipt.block_number.is_none()
-                    {
+                    if main_node_receipt.status == 0.into() {
```
perekopskiy, line 526: Given that we don't return receipts for rejected transactions anymore, forwarding requests to main node doesn't make sense anymore.

core/lib/dal/src/transactions_web3_dal.rs
Deniallugo: no, we unwrap the error.

Deniallugo: it's just `if none then zero else one  `

RomanBrodetski (dismissed):

perekopskiy (approved):

RomanBrodetski (approved):

RomanBrodetski (approved):

PR #708, opened 2023-12-18
fix: silence cat errors on Show Logs step

tomg10:
## What ❔

Silencing errors like:
```
Run ci_run cat server.log
cat: server.log: No such file or directory
Error: Process completed with exit code 1.
```
That imo are just an additional needless noise that decrease the UX of development.

### DIFF
--- a/.github/workflows/ci-core-reusable.yml
+++ b/.github/workflows/ci-core-reusable.yml
@@ -111,7 +111,7 @@ jobs:
       - name: Show logs
         if: always()
         run: |
-          ci_run cat server.log
+          ci_run cat server.log || true
           ci_run sccache --show-stats
           ci_run cat /tmp/sccache_log.txt
   integration:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-core-reusable.yml
```diff
@@ -203,10 +203,10 @@ jobs:
...
         if: always()
         run: |
-          ci_run cat server.log
-          ci_run cat contract_verifier.log
-          ci_run cat core/tests/revert-test/revert.log
-          ci_run cat core/tests/upgrade-test/upgrade.log
+          ci_run cat server.log || true
+          ci_run cat contract_verifier.log || true
+          ci_run cat core/tests/revert-test/revert.log || true
+          ci_run cat core/tests/upgrade-test/upgrade.log || true
```
montekki, line 209: Can you please change this to `cat` a single file within the single `- name` directive while you are at it? It would improve things so much more if all logs were not squashed together.

tomg10: great idea! I added this change

RomanBrodetski (approved):

PR #711, opened 2023-12-19
feat: rewritten gossip sync to be async from block processing

pompon0:
* Simplified configs to not require genesis block.
* Removed consensus fields from jsonrpc sync api, to prevent having gaps in miniblock quorum certificates.
* Moved consensus fields to a separate table to make first/last entry lookup fast and reliable.
* Migrated consensus integration to new era-consensus API: https://github.com/matter-labs/era-consensus/pull/53
* Deduplicated code between EN consensus integration and main node consensus integration
* Implemented support for cloning a db in tests


core/lib/dal/src/connection/mod.rs
```diff
@@ -68,85 +60,116 @@ impl ConnectionPoolBuilder {
...
     pub(crate) inner: PgPool,
+    database_url: String,
     max_size: u32,
 }
 
 impl fmt::Debug for ConnectionPool {
     fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
         formatter
             .debug_struct("ConnectionPool")
+            .field("database_url", &self.database_url)
```
popzxc, line 81: It was previously discussed that `database_url` can contain sensitive information (e.g. database password), so it shouldn't be printed lightheartedly. Probably it makes sense to write a comment stating that this field is omitted intentionally.

core/lib/dal/src/consensus_dal.rs
```diff
@@ -1,13 +1,19 @@
...
 use crate::StorageProcessor;
 
+/// Storage access methods for `zksync_core::consensus` module.
 #[derive(Debug)]
 pub struct ConsensusDal<'a, 'c> {
     pub storage: &'a mut StorageProcessor<'c>,
 }
 
 impl ConsensusDal<'_, '_> {
+    /// Fetch the current BFT replica state.
```
popzxc, line 16: Bikeshedding: Though there is no formal guide right now, we most often write doc-comments in the 3rd form (e.g. "Fetches" instead of "Fetch"), and I see that you also use this form below. Feel free to ignore this comment, but if you'll be touching this file otherwise, you can make it easier for me to publish all the stuff on crates.io later :)

core/lib/dal/src/models/proto/mod.proto
```diff
@@ -2,9 +2,23 @@ syntax = "proto3";
 
 package zksync.dal;
 
-import "zksync/roles/validator.proto";
+message Transaction {
+  // Default derive(serde::Serialize) encoding of the zksync_types::Transaction.
+  // TODO(gprusak): it is neither efficient, unique, nor suitable for version control.
+  //     replace with a more robust encoding.
```
popzxc, line 8: In this PR? If not, it'd be great to have a task number here, e.g. `TODO(BFT-xxx): ...`.

popzxc (approved):
Looks good to me, but I feel that my domain knowledge isn't sufficient to grok changes fully. I'd highly recommend getting a 2nd review from someone with enough domain knowledge.

core/lib/dal/src/consensus_dal.rs
```diff
@@ -47,6 +54,158 @@ impl ConsensusDal<'_, '_> {
...
+
+    /// Inserts a certificate for the miniblock `cert.header().number`.
+    /// It verifies that
+    /// * the certified payload matches the miniblock in storage
+    /// * the `cert.header().parent` matches the parent miniblock.
+    /// * the parent block already has a certificate.
+    /// NOTE: This is an extra secure way of storing a certificate,
+    /// which will help us to detect bugs in the consensus implementation
+    /// while it is "fresh". If it turns out to take too long,
+    /// we can remove the verification checks later.
```
RomanBrodetski, line 164: @pompon0 ideally `dal` code should only serve as the DB interface and not do any business-logic. Do you think there is a better place for such validation? I'm not concerned about performance by the way - I like having those checks.

core/lib/dal/src/sync_dal.rs
```diff
@@ -64,40 +63,54 @@ impl SyncDal<'_, '_> {
...
         .await?;
+        Ok(transactions.into_iter().map(Transaction::from).collect())
+    }
 
-        let Some(storage_block_details) = storage_block_details else {
+    pub async fn sync_block(
+        &mut self,
+        block_number: MiniblockNumber,
+        current_operator_address: Address,
+        include_transactions: bool,
```
RomanBrodetski, line 102: ideally whether transactions are included should be reflected on the type level. Do you think there is a way to do this?

RomanBrodetski (approved):
Looks good! @slowli were you also planning to review this?

core/lib/dal/src/consensus_dal.rs
pompon0: these are storage consistent checks. I would like to avoid exposing public dal methods which allow to corrupt invariants that are checked here.

RomanBrodetski: @pompon0 I understand the reasoning but 

> avoid exposing public dal methods which allow to corrupt invariants

This is not how we approached the DAL abstraction in the first place - it's a dump wrapper around sql queries. We also try to enforce invariants on the schema design level (e.g. via normalization) - so that it's impossible to represent an invalid state. 

Having said that, I won't insist - and I know in other places we weren't that strict. Feel free to leave as is if you like - for me it was only important to let you know the reasoning

core/lib/dal/src/sync_dal.rs
pompon0: flag indicating whether or not transactions are requested is an argument of the rpc call: https://github.com/matter-labs/zksync-era/blob/39f2d5057a7bb5b489e858bafb72fb925ebc6c8f/core/lib/zksync_core/src/api_server/web3/namespaces/en.rs#L19 , so at best we could make this method always include transactions. Alternatively we can also
* make the the rpc call handler call 2 functions (one for header, one for transactions) instead, so that there is no boolean flag in dal.
* Create a newtype for this argument `struct IncludeTransactions(boolean)`, to make function calls more readable.

core/lib/zksync_core/src/consensus/storage.rs
```diff
@@ -1,419 +1,417 @@
...
-        })
+            operator_addr,
+        }
+    }
+
+    /// Converts `Store` into a `BlockStore`.
+    pub fn into_block_store(self) -> BlockStore {
+        BlockStore {
+            inner: self,
+            cursor: sync::Mutex::new(None),
```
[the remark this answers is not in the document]
pompon0: It is None for the main node, and Some for the external node.

core/lib/zksync_core/src/consensus/mod.rs
```diff
@@ -1,61 +1,212 @@
...
+        })
+        .await
+    }
+}
+
+/// External node consensus config.
+#[derive(Debug, Clone)]
+pub struct FetcherConfig {
+    executor: executor::Config,
+    operator_address: Address,
```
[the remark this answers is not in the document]
pompon0: It is intentionally verbose to show how many places depend on it. I'm looking forward to https://github.com/matter-labs/zksync-era/pull/688 getting merged.

core/lib/zksync_core/src/consensus/storage.rs
```diff
@@ -1,419 +1,417 @@
...
+    inner: FetcherCursor,
+    actions: ActionQueueSender,
+}
+
+/// Wrapper of `ConnectionPool` implementing `ReplicaStore` and `PayloadManager`.
+#[derive(Clone, Debug)]
+pub(super) struct Store {
     pool: ConnectionPool,
-    operator_address: Address,
+    operator_addr: Address,
```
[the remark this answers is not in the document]
pompon0: it is just long. Reverted

core/lib/dal/src/models/proto/mod.proto
```diff
@@ -2,9 +2,23 @@ syntax = "proto3";
...
+  optional uint32 protocol_version = 9; // required; u16
+  optional bytes hash = 1; // required; H256
+  optional uint32 l1_batch_number = 2; // required
+  optional uint64 timestamp = 3; // required; seconds since UNIX epoch
+  optional uint64 l1_gas_price = 4; // required; gwei
+  optional uint64 l2_fair_gas_price = 5; // required; gwei
+  optional uint32 virtual_blocks = 6; // required
+  optional bytes operator_address = 7; // required; H160
+  repeated Transaction transactions = 8;
+  optional bool last_in_batch = 10; // required
```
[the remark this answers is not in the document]
pompon0: Previous implementation was caching a block, which made implementing PersistentBlockStore correctly impossible. Additionally @popzxc said that we don't want to depend on the property that the last block is always empty.

core/lib/zksync_core/src/consensus/storage.rs
```diff
@@ -1,419 +1,417 @@
...
+                .wrap("access()")?;
+            let number = storage
+                .last_miniblock_number(ctx)
+                .await
+                .wrap("last_miniblock_number()")?;
+            if number >= block.header().number {
+                storage
+                    .insert_certificate(ctx, &block.justification, self.inner.operator_addr)
+                    .await
+                    .wrap("insert_certificate()")?;
```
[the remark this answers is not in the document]
pompon0: We don't reexecute the persisted miniblock without the certificate - we just backfill the certificate after restart and then continue executing the blocks.

core/lib/zksync_core/src/consensus/storage.rs
```diff
@@ -1,419 +1,417 @@
...
+                    operator_address: payload.operator_address,
+                    transactions: payload.transactions,
+                };
+                cursor
+                    .actions
+                    .push_actions(cursor.inner.advance(block))
+                    .await;
+            }
+        }
+        const POLL_INTERVAL: time::Duration = time::Duration::milliseconds(50);
```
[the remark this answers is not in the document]
pompon0: ack

core/lib/zksync_core/src/consensus/storage.rs
```diff
@@ -1,419 +1,417 @@
...
+        // This mutex prevents concurrent store_next_block calls.
+        let mut cursor = ctx.wait(self.cursor.lock()).await?;
+        if let Some(cursor) = &mut *cursor {
+            let number = MiniblockNumber(
+                u32::try_from(block.header().number.0)
+                    .context("Integer overflow converting block number")?,
+            );
+            let payload =
+                Payload::decode(&block.payload).context("Failed deserializing block payload")?;
+            if cursor.inner.next_miniblock <= number {
```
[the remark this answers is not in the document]
pompon0: added returning an error if next_miniblock < number.

core/lib/zksync_core/src/consensus/storage.rs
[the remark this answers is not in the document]
pompon0: added a test for that

core/lib/zksync_core/src/consensus/storage/testonly.rs
```diff
@@ -0,0 +1,47 @@
+//! Storage test helpers.
+use anyhow::Context as _;
+use zksync_concurrency::{ctx, error::Wrap as _, time};
+use zksync_consensus_roles::validator;
+use zksync_consensus_storage as storage;
+
+use super::{BlockStore, CtxStorage};
+
+impl BlockStore {
+    /// Waits for the `want` miniblock to have a certificate.
```
slowli, line 10: Nit: "`want` miniblock" -> "miniblock with the given `number`"?

core/lib/zksync_core/src/consensus/testonly.rs
```diff
@@ -235,72 +278,47 @@ impl StateKeeperHandle {
...
     const POLL_INTERVAL: time::Duration = time::Duration::milliseconds(100);
-    let mut n = L1BatchNumber(1);
+    let mut n = {
+        let mut storage = pool.access_storage().await.context("access_storage()")?;
+        storage
+            .blocks_dal()
+            .get_last_l1_batch_number_with_metadata()
+            .await
+            .context("get_last_l1_batch_number_with_metadata()")?
+            .context("no L1 batchers in Postgres")?
```
slowli, line 320: Typo: "batchers" -> "batches".

core/lib/zksync_core/src/consensus/testonly.rs
```diff
@@ -235,72 +278,47 @@ impl StateKeeperHandle {
...
-        ctx: &ctx::Ctx,
-        pool: &ConnectionPool,
-    ) -> anyhow::Result<()> {
+    /// Creates a new `BlockStore` for the underlying `ConnectionPool`.
+    pub fn store(&self) -> BlockStore {
+        Store::new(self.pool.clone(), self.operator_address).into_block_store()
+    }
+
+    // Wait for all pushed miniblocks to be produced.
+    pub async fn sync(&self, ctx: &ctx::Ctx) -> ctx::Result<()> {
```
slowli, line 293: Bikeshedding: I'd rename this method to something like `wait_for_miniblocks()` for clarity.

core/lib/zksync_core/src/lib.rs
```diff
@@ -1,4 +1,8 @@
-#![allow(clippy::upper_case_acronyms, clippy::derive_partial_eq_without_eq)]
+#![allow(
+    clippy::upper_case_acronyms,
+    clippy::derive_partial_eq_without_eq,
+    clippy::redundant_locals
```
slowli, line 4: Bikeshedding: IMO, more local placement of this switch (e.g., on the `consensus` module, or even on specific methods / functions) would be better.

core/lib/zksync_core/src/consensus/mod.rs
```diff
@@ -1,61 +1,212 @@
...
+            self.executor.node_key.public(),
+        );
+
+        scope::run!(ctx, |ctx, s| async {
+            let store = Store::new(pool, self.operator_address);
+            let mut block_store = store.clone().into_block_store();
+            block_store
+                .set_actions_queue(ctx, actions)
+                .await
+                .wrap("block_store.try_init_genesis()")?;
```
slowli, line 200: Nit: Wrong message; should be something like `set_actions_queue`.

core/lib/zksync_core/src/consensus/mod.rs
```diff
@@ -1,61 +1,212 @@
...
+            operator_address: cfg
+                .operator_address
+                .context("operator_address is required")?,
+        })
+    }
+}
+
+/// Main node consensus config.
+#[derive(Debug, Clone)]
 pub struct Config {
```
slowli, line 120: Bikshedding: Maybe, name this `MainNodeConfig` given that this module will presumably be used by both the main and external nodes?

core/lib/zksync_core/src/consensus/storage/mod.rs
```diff
@@ -0,0 +1,424 @@
...
+                    .context("Integer overflow converting block number")?,
+            );
+            let payload =
+                Payload::decode(&block.payload).context("Failed deserializing block payload")?;
+            let want = cursor.inner.next_miniblock;
+            if want < number {
+                return Err(anyhow::anyhow!("expected {want:?}, got {number:?}").into());
+            }
+            if want == number {
+                let block = FetchedBlock {
```
slowli, line 319: Nit: IMO, it'd still make sense to extract the `Payload` -> `FinalBlock` conversion as in the previous implementation so that it's more maintainable.

core/lib/zksync_core/src/consensus/storage/mod.rs
```diff
@@ -0,0 +1,424 @@
...
+                    l1_gas_price: payload.l1_gas_price,
+                    l2_fair_gas_price: payload.l2_fair_gas_price,
+                    virtual_blocks: payload.virtual_blocks,
+                    operator_address: payload.operator_address,
+                    transactions: payload.transactions,
+                };
+                cursor
+                    .actions
+                    .push_actions(cursor.inner.advance(block))
+                    .await;
```
slowli, line 335: Nit: This can be extracted to a `Cursor` method taking `Payload` as an input to improve code structuring.

core/lib/zksync_core/src/consensus/storage.rs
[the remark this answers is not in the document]
slowli: In this case, wouldn't `Option<sync::Mutex<_>>` make more sense?

core/lib/zksync_core/src/consensus/storage/mod.rs
```diff
@@ -0,0 +1,424 @@
...
+            .wrap("new_fetcher_cursor()")?;
+        *sync::lock(ctx, &self.cursor).await? = Some(Cursor { inner, actions });
+        Ok(())
+    }
+}
+
+#[async_trait::async_trait]
+impl PersistentBlockStore for BlockStore {
+    async fn state(&self, ctx: &ctx::Ctx) -> ctx::Result<BlockStoreState> {
+        // Ensure that genesis block has consensus field set in postgres.
```
slowli, line 251: Is this comment outdated? Also, how *is* the genesis block set for ENs? Or is it not ensured specifically? IIUC, in tests it's currently obtained from a Postgres-level snapshot. Is that the intended way to pass the genesis block data for ENs?

core/lib/zksync_core/src/consensus/mod.rs
```diff
@@ -1,61 +1,212 @@
...
 #[cfg(test)]
 mod tests;
 
-pub(crate) use self::{payload::Payload, storage::sync_block_to_consensus_block};
+use std::collections::{HashMap, HashSet};
+
+use serde::de::Error;
+use zksync_consensus_crypto::{Text, TextFmt};
+
+use self::storage::Store;
```
slowli, line 24: Please place all `use`s before module definitions so that they are appropriately grouped by the formatter.

core/lib/zksync_core/src/consensus/storage/mod.rs
```diff
@@ -0,0 +1,424 @@
...
+    /// matches the miniblock in the db.
+    async fn verify(
+        &self,
+        ctx: &ctx::Ctx,
+        block_number: validator::BlockNumber,
+        payload: &validator::Payload,
+    ) -> ctx::Result<()> {
+        let want = self.propose(ctx, block_number).await?;
+        let want = consensus::Payload::decode(&want).context("Payload::decode(want)")?;
+        let got = consensus::Payload::decode(payload).context("Payload::decode(got)")?;
```
slowli, line 418: Nit: `consensus::` qualification is redundant; `Payload` is imported into the scope directly.

core/lib/zksync_core/src/consensus/storage/mod.rs
```diff
@@ -0,0 +1,424 @@
...
+    /// Verify that `payload` is a correct proposal for the block `block_number`.
+    /// Currently (for the main node) it is implemented as checking whether the received payload
+    /// matches the miniblock in the db.
+    async fn verify(
+        &self,
+        ctx: &ctx::Ctx,
+        block_number: validator::BlockNumber,
+        payload: &validator::Payload,
+    ) -> ctx::Result<()> {
+        let want = self.propose(ctx, block_number).await?;
```
slowli, line 416: Nit: You can define a version of propose that returns domain-specific `Payload` and convert it to `validator::Payload` in `PayloadManager::propose()`. In this way, you wouldn't need to convert the `want` payload twice.

core/lib/zksync_core/src/consensus/mod.rs
pompon0: done

core/lib/zksync_core/src/lib.rs
pompon0: done

core/lib/zksync_core/src/consensus/testonly.rs
pompon0: done

core/lib/zksync_core/src/consensus/storage/testonly.rs
pompon0: done

core/lib/zksync_core/src/lib.rs
pompon0: done

core/lib/zksync_core/src/consensus/storage/mod.rs
pompon0: Comment removed. Yes, ENs are supposed to obtain the certificate via the snapshot for simplicity.

core/lib/zksync_core/src/consensus/storage/mod.rs
pompon0: done

core/lib/zksync_core/src/consensus/storage/mod.rs
pompon0: This only works for a single validator setup, I don't think there is a need to optimize that.

core/lib/dal/src/sync_dal.rs
```diff
@@ -64,40 +63,54 @@ impl SyncDal<'_, '_> {
...
         .with_arg("block_number", &block_number)
         .fetch_optional(self.storage.conn())
+        .await?
+        else {
+            return Ok(None);
+        };
+        Ok(Some(block.try_into()?))
+    }
+
+    pub(super) async fn sync_block_transactions(
```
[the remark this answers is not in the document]
pompon0: done

slowli (approved):
As discussed privately, there's a couple of outstanding questions (e.g., whether it make sense to have `Payload` defined / validated in the DAL crate), but they can be revisited later.

Also, an operational nit: I think this PR merits a "feat" prefix; it brings quite a few improvements, and "refactor" PRs don't show up in [the changelog](https://github.com/matter-labs/zksync-era/blob/main/core/CHANGELOG.md).

PR #725, opened 2023-12-21
chore(dal): bumps sqlx dependency to 0.7.3

montekki:
## What ❔

Bumps `sqlx` dependency to `0.7.3`.

## Checklist


core/lib/utils/Cargo.toml
```diff
@@ -26,6 +26,7 @@ hex = "0.4"
 reqwest = { version = "0.11", features = ["blocking"] }
 itertools = "0.10.5"
 metrics = "0.21"
+sqlx = "0.7.3"
```
popzxc, line 29: This crate may be used by something that doesn't need DAL, so I would really love to not depend on `sqlx` here, since we have tracked history of starting to used whatever we can use uncontrollably.
This can become a huge PITA in the future (look at `zksync_types` 🥲 ).

popzxc: Maybe we can move functions that use SQLx types to the DAL crate, since most likely they are used in the context where DAL is available?

montekki: Actually not needed, removed.

.github/workflows/ci-core-lint-reusable.yml
```diff
@@ -36,5 +36,5 @@ jobs:
           ci_run zk lint js --check
           ci_run zk lint ts --check
           ci_run zk lint md --check
-          ci_run zk db check-sqlx-data
 
+# ci_run zk db check-sqlx-data
```
popzxc, line 40: Debug artifact?

montekki: No, this check will not work because our CI images are still those generated with 0.5.* and they are not compatible. So we have to merge this -> images get built -> we bring this back

core/lib/dal/Cargo.toml
```diff
@@ -29,22 +29,21 @@ url = "2"
 prost = "0.12.1"
 rand = "0.8"
 tokio = { version = "1", features = ["full"] }
-sqlx = { version = "0.5.13", default-features = false, features = [
+sqlx = { version = "0.7.3", default-features = false, features = [
     "runtime-tokio-native-tls",
```
slowli, line 33: Nit: Judging by [the crate readme](https://crates.io/crates/sqlx), `runtime-tokio-native-tls` is soft-deprecated in favor of `runtime-tokio` + `tls-native-tls`.

popzxc (approved):
fluttershy_yay.gif

slowli (commented):
There's some new features that we could use (e.g., flattening [as mentioned here](https://github.com/matter-labs/zksync-era/blob/3db25cbea80180a1238531944d7e45a5408003dc/core/lib/dal/src/models/storage_block.rs#L121C13-L121C16), or less conversions from byte arrays to `Vec<u8>`), but they should be tackled in a follow-up PR, IMO.

core/lib/dal/Cargo.toml
montekki: Changed to those, any particular reason we don't use `rustls` here btw?

popzxc (dismissed):
still yay

core/lib/dal/Cargo.toml
popzxc: I vaguely remember some annoying issue that didn't let us do that back in the time, but don't even remember what it was.
Maybe it's even long resolved.

RomanBrodetski:
didn't look through every file, but really happy to see this PR. Let's monitor the DB performance when we deploy this as well

montekki:
The regression of our external node integration test seems to be introduced in launchbadge/sqlx#2039 but the size of the PR is so huge that it is almost impossible to pinpoint the exact reason. The main symptom of the regression is that it looks like something has happened with the way pool recycles connections and now our pool spends huge amounts of time waiting for acquiring connections in acuire_connection_retried

popzxc:
@montekki we've had problems with SQLx native pool in the past too; you can try swapping the native pool with deadpool.
Here's [the implementation for Lite (that seems to work be working just fine for years)](https://github.com/matter-labs/zksync-dev/blob/dev/core/lib/storage/src/connection/mod.rs)

core/lib/dal/src/blocks_dal.rs
```diff
@@ -2254,7 +2254,7 @@ impl BlocksDal<'_, '_> {
             WHERE
                 number = $1
             "#,
-            miniblock_number.0 as u32
+            miniblock_number.0 as i32
```
popzxc, line 2257: Side comment: probably we should create a task to switch all the counter types to u64, just so that we don't have our internal y2k38. Not urgent (lol), but still.

popzxc (approved):

.github/workflows/ci-core-lint-reusable.yml
artmakh: Add indentation please

artmakh (approved):

PR #728, opened 2023-12-21
chore(docs): Update list of repositories

mm-zk:
## What ❔

* Updated list of repositories in repositories.md

## Why ❔

* Since last update, we moved everything to opensource and introduced boojum (with its own set of repositories).


shahar4 (dismissed):
This is amazing! A couple of ultra nit suggestions

shahar4 (approved):

PR #732, opened 2023-12-21
chore: Retire the lightweight tree component

popzxc:
## What ❔

Removes the component-based option to enable the lightweight tree.
Also, it drops the support for `_new` postfix for the tree component.

## Why ❔

The tree can be configured via configuration options in a more elegant way.

## Checklist


infrastructure/zk/src/hyperchain_wizard.ts
```diff
@@ -489,7 +489,7 @@ async function startServer() {
     const results: any = await enquirer.prompt(questions);
 
     let components: string[] = [];
-    const defaultChoices = ['http_api', 'eth', 'state_keeper', 'housekeeper', 'tree_lightweight'];
+    const defaultChoices = ['http_api', 'eth', 'state_keeper', 'housekeeper', 'tree'];
```
popzxc, line 492: Given that even EN currently requires full tree to run, the ability to use it in hyperchain setup should be reconsidered (e.g. I'd assume that it may only work in a microservice-based setup where both full and lightweight tree are ran independently). However given that the performance of full tree is pretty good anyways, probably it's not a very big deal for the time being.
cc @githubdoramon

slowli (dismissed):

RomanBrodetski (approved):

PR #734, opened 2023-12-21
fix(prover): Reduce amount of prover connections per prover subcomponent

EmilLuta:
## What ❔

Configuration limit on number of DB connections per prover subcomponent.

## Why ❔

During surge of TPS, prover database was pummeled with connections. This change will makes sure the database will be protected from rogue connections. Assuming the patch is wrong, prover subcomponents will be throttled.

## Checklist

### DIFF
--- a/prover/prover_fri/src/main.rs
+++ b/prover/prover_fri/src/main.rs
@@ -123,7 +123,7 @@ async fn main() -> anyhow::Result<()> {
     // There are 2 threads using the connection pool:
     // 1. The prover thread, which is used to update the prover job status.
     // 2. The socket listener thread, which is used to update the prover instance status.
-    const MAX_POOL_SIZE_FOR_PROVER: i32 = 2;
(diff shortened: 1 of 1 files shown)


Deniallugo (approved):

perekopskiy (approved):

prover/witness_vector_generator/src/main.rs
```diff
@@ -58,13 +58,10 @@ async fn main() -> anyhow::Result<()> {
     let exporter_config = PrometheusExporterConfig::pull(config.prometheus_listener_port);
 
     let postgres_config = PostgresConfig::from_env().context("PostgresConfig::from_env()")?;
-    let pool = ConnectionPool::builder(
-        postgres_config.prover_url()?,
-        postgres_config.max_connections()?,
```
RomanBrodetski, line 63: what was the previous number configured here?

PR #741, opened 2023-12-22
fix: close Docker PostgreSQL to external connections

bogatyy:
PostgreSQL should only be accessible from the localhost interface, otherwise malicious actors can change or delete the database.

Bringing the `zk init` settings in line with `zk stack init` settings: https://github.com/matter-labs/zksync-era/blob/main/docker-compose-zkstack-common.yml#L28

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -107,7 +107,7 @@ services:
   postgres:
     image: "postgres:14"
     ports:
-      - 5432:5432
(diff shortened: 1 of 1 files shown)


shahar4 (approved):
Good catch! 

Seems like this exists also in `docker-compose-gpu-runner.yml` and in `docker-compose-cpu-runner.yml` and a couple of other files. Mind changing it there too?

PR #742, opened 2023-12-22
feat(en): Make consistency checker work with pruned data

slowli:
## What ❔

Modifies consistency checker so that it works with pruned node data during snapshot recovery. Adds tests for the checker.

## Why ❔

## Checklist


core/lib/zksync_core/src/consistency_checker/tests/commit_l1_batch_351000-351004_mainnet.calldata
slowli, line 1: If there are any suggestions to make these fixtures smaller, they're welcome. I've spent some time searching for smaller calldata sizes on the testnet / mainnet, but some fixtures are still quite large.

core/lib/zksync_core/src/consistency_checker/mod.rs
```diff
@@ -1,204 +1,300 @@
...
-            Some(1.into()),
-            "Main node gave us a failed commit tx"
-        );
+        // We can't get tx calldata from db because it can be fake.
+        let commit_tx_input_data = self
+            .l1_client
+            .transaction_input_data(commit_tx_hash)
+            .await?
+            .with_context(|| format!("Commit for tx {commit_tx_hash:?} not found on L1"))?;
+        // FIXME: shouldn't the receiving contract and selector be checked as well?
```
slowli, line 144: I'm not very familiar with the checker requirements, but the fact that it doesn't check either the recipient contract address, or the Solidity function selector makes it intuitively easy to fool.

core/lib/zksync_core/src/consistency_checker/mod.rs
```diff
@@ -1,204 +1,300 @@
...
-                .unwrap();
-            let batch_has_metadata = metadata
-                .map(|m| {
-                    m.metadata.bootloader_initial_content_commitment.is_some()
-                        && m.metadata.events_queue_commitment.is_some()
-                })
-                .unwrap_or(false);
+                .await?;
+            let batch_has_metadata = batch_with_metadata.map_or(false, |batch| {
+                // FIXME: how does this work for pre-Boojum batches?
```
slowli, line 265: Unless I'm missing something, pre-Boojum batches will always have `bootloader_initial_content_commitment` and `events_queue_commitment` set to `None`, so the checker will hang up on them.

core/lib/zksync_core/src/consistency_checker/mod.rs
```diff
@@ -1,204 +1,318 @@
...
+}
+
+impl From<zksync_dal::SqlxError> for CheckError {
+    fn from(err: zksync_dal::SqlxError) -> Self {
+        Self::Internal(err.into())
+    }
+}
+
+#[async_trait]
+trait L1Client: fmt::Debug + Send + Sync {
```
popzxc, line 36: Do we need a separate trait for that? Wouldn't `eth_client::EthInterface` suffice here?

core/lib/zksync_core/src/consistency_checker/mod.rs
popzxc: Good catch, agree. I'd prefer it to be implemented separately though.

core/lib/zksync_core/src/consistency_checker/mod.rs
```diff
@@ -1,204 +1,318 @@
...
+            l1_batch_updater: Box::new(()),
+            l1_data_mismatch_behavior: L1DataMismatchBehavior::Bail,
+            pool,
         }
     }
 
-    async fn check_commitments(&self, batch_number: L1BatchNumber) -> Result<bool, error::Error> {
-        let mut storage = self.db.access_storage().await.unwrap();
+    /// Configures this checker to log discovered L1 batch mismatches instead of bailing with an error.
+    pub fn log_l1_data_mismatches(&mut self) {
```
popzxc, line 107: The name provides an impression that it's additive (e.g. it just changes whether logs are emitted or not), while in fact it changes how the `run` function works.
Maybe it will be more clear the other way around, e.g. `bail_on_l1_data_mismatches`?

Additionally, it looks like the main "user" of the checker - EN - has to call this method in order to get the desired behavior, which also seems odd. I'd expect this component to match EN expectations by default.

slowli: I'm not sure what behavior is intended to be the default one long-term. [The change](https://github.com/matter-labs/zksync-era/pull/758) in behavior is very recent and AFAIU, it's temporary. @RomanBrodetski, could you please clarify whether this "log-only" operation mode will remain long-term?

core/lib/zksync_core/src/consistency_checker/mod.rs
slowli: `EthInterface` cannot be made into a trait object because of the parametric `call_contract_function` method. IMO, it'd make sense to refactor this method so that it's not parametric, and use an extension trait for the parametric version (optionally, add a builder pattern as well to manage argument count). Would you prefer me doing this in this PR, or separately? As a possible alternative (IMO, a suboptimal one), `ConsistencyChecker` itself may incorporate a `C: EthInterface` type param.

core/lib/zksync_core/src/consistency_checker/mod.rs
popzxc: Makes sense; but I'd probably still prefer a more expressive name. Maybe something `do_not_bail_on_l1_data_mismatches` or something of that sort.

core/lib/zksync_core/src/consistency_checker/mod.rs
popzxc: As discussed, would love to see object-safe implementation of `EthInterface`.

core/lib/zksync_core/src/consistency_checker/mod.rs
popzxc: How did it work before Boojum upgrade? Did we remove the corresponding logic? If so, probably it needs to be restored. Although realistically I don't think that it's going to fire, so the priority is probably low, but until we implement a sane policy w.r.t. backward compatibility, I'd always prefer to be consistently backward compatible.

core/lib/zksync_core/src/consistency_checker/tests/commit_l1_batch_351000-351004_mainnet.calldata
popzxc: Why not generate it on localhost tho?

RomanBrodetski (dismissed):
Looks great! Regarding the Bail/Log: we want to bring Bail back, but to do this we need to fix the method that generates pubdata on the server. @mm-zk already found the fundamental issue, but we didn;t change the method yet. Note that this method is not used anymore on the main node - we take the DB value instead

popzxc (dismissed):

core/lib/zksync_core/src/consistency_checker/tests/commit_l1_batch_351000-351004_mainnet.calldata
slowli: Slightly less reliable; taking this data from the mainnet / testnet guarantees that the fixtures are 100% valid. In any case, fixtures could be swapped later if necessary.

core/lib/zksync_core/src/consistency_checker/mod.rs
slowli: Added a TODO.

core/lib/zksync_core/src/consistency_checker/mod.rs
slowli: Before Boojum, such filtering [was not performed](https://github.com/matter-labs/zksync-era/commit/80273264a9512bc1e6f1d1f4372107f9167260b1#diff-a85ad3ae6a0d60fe1c81964a604bdcecd1eed05a76c87f08f45d992449503db7R163-R177); it was assumed that if an L1 batch has any metadata, it has all of it (i.e., metadata insertion is atomic). I've refactored this logic so that it's easier to reason about.

PR #745, opened 2023-12-22
fix: added waiting for prometheus to finish

tomg10:
## What ❔

added waiting for prometheus task to finish.

## Why ❔

We need to wait for prometheus to push one last 'batch of metrics'

### DIFF
--- a/core/bin/snapshots_creator/src/main.rs
+++ b/core/bin/snapshots_creator/src/main.rs
@@ -2,7 +2,10 @@
 
 use anyhow::Context as _;
 use prometheus_exporter::PrometheusExporterConfig;
-use tokio::sync::{watch, Semaphore};
+use tokio::{
+    sync::{watch, Semaphore},
+    task::JoinHandle,
+};
 use zksync_config::{configs::PrometheusConfig, PostgresConfig, SnapshotsCreatorConfig};
 use zksync_dal::ConnectionPool;
 use zksync_env_config::{object_store::SnapshotsObjectStoreConfig, FromEnv};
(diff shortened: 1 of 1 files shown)


core/bin/snapshots_creator/src/main.rs
```diff
@@ -27,7 +30,7 @@ mod tests;
 
 async fn maybe_enable_prometheus_metrics(
     stop_receiver: watch::Receiver<bool>,
-) -> anyhow::Result<()> {
+) -> anyhow::Result<Option<JoinHandle<anyhow::Result<()>>>> {
```
slowli, line 33: Nit: This function doesn't need to be `async`, and the outer `anyhow::Result<_>` is obsolete as well. (AFAIU, both of these could be caught with Clippy lints that are not enabled by default; you may try copying them from [another crate](https://github.com/matter-labs/zksync-era/blob/4fe284a273a0299528c00e01264a1c1c32f630c5/core/lib/merkle_tree/src/lib.rs#L41-L47) to see how many lints we're missing 🙃)

core/bin/snapshots_creator/src/main.rs
```diff
@@ -291,5 +295,10 @@ async fn main() -> anyhow::Result<()> {
     run(blob_store, replica_pool, master_pool, MIN_CHUNK_COUNT).await?;
     tracing::info!("Finished running snapshot creator!");
     stop_sender.send(true).ok();
+    if let Some(prometheus_exporter_task) = prometheus_exporter_task {
+        prometheus_exporter_task
```
slowli, line 299: Nit: It may make sense to have a reasonable timeout here (e.g., 10s) via `tokio::time::timeout` so that it's guaranteed that the process eventually terminates. Maybe, that's handled by K8s infra already; in any case, feel free to leave as-is if you don't think it's ebtter.

slowli (approved):

core/bin/snapshots_creator/src/main.rs
tomg10: AFAIK We have alerts if a job is running too long, so we'd know if this ever happens

core/bin/snapshots_creator/src/main.rs
tomg10: I'd leave it as it for now. I want to play with clippy settings after company-wide holidays and fix found issues in this crate

PR #747, opened 2023-12-22
feat(vm)!: Release v19 - remove allowlist

StanislavBreadless:
## What ❔

This upgrade contains mainly contract changes (more description of those [here](https://github.com/matter-labs/era-contracts/pull/133)), while the changes that impacted `zksync-era` repo mainly refer to the changes in how the testing / contract preprocessing is done: now, the system contracts are preprocessed into the `contracts-preprocessed` folder and then all the artifacts should be taken from there 

## Checklist


core/lib/zksync_core/src/api_server/tx_sender/mod.rs
```diff
@@ -88,7 +88,9 @@ impl MultiVMBaseSystemContracts {
             | ProtocolVersionId::Version15
             | ProtocolVersionId::Version16
             | ProtocolVersionId::Version17 => self.post_virtual_blocks_finish_upgrade_fix,
-            ProtocolVersionId::Version18 | ProtocolVersionId::Version19 => self.post_boojum,
+            ProtocolVersionId::Version18
+            | ProtocolVersionId::Version19
+            | ProtocolVersionId::Version20 => self.post_boojum,
```
perekopskiy, line 93: Bootloader was changed, so you need to branch here

StanislavBreadless: @perekopskiy its interface didn't, so I believe we can keep the old one

perekopskiy: But gas costs have been slightly changed.. I'm not sure if it affects API responses though. @Deniallugo what do you think?

StanislavBreadless: @perekopskiy it does affect them (for estimateGas), I just thought that we generally try to avoid adding new bootloaders if we can avoid doing it, but I don't mind. I'll add one

Deniallugo (approved):

artmakh (approved):

PR #757, opened 2023-12-25
fix(pubdata): use pubdata from bootloader in l1 operations

RomanBrodetski:


core/lib/types/src/commitment.rs
```diff
@@ -209,7 +209,7 @@ impl L1BatchWithMetadata {
                         .to_vec(),
                 ),
                 Token::Bytes(self.metadata.l2_l1_messages_compressed.clone()),
-                Token::Bytes(self.header.pubdata_input.clone().unwrap()),
+                Token::Bytes(self.header.pubdata_input.clone().unwrap()[32..]),
```
perekopskiy, line 212:
```suggestion
                Token::Bytes(self.header.pubdata_input[32..].clone().unwrap()),
```

PR #762, opened 2023-12-25
fix(core): do not unwrap unexisting calldata in commitment and regenerate it

montekki:
## What ❔

Solve panics on nonexistent calldata.

## Checklist

### DIFF
--- a/core/lib/types/src/commitment.rs
+++ b/core/lib/types/src/commitment.rs
@@ -209,13 +209,50 @@ impl L1BatchWithMetadata {
                         .to_vec(),
                 ),
                 Token::Bytes(self.metadata.l2_l1_messages_compressed.clone()),
-                Token::Bytes(self.header.pubdata_input.clone().unwrap()),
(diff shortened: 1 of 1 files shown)


RomanBrodetski (approved):

Deniallugo (approved):

core/lib/types/src/commitment.rs
```diff
@@ -209,14 +209,51 @@ impl L1BatchWithMetadata {
                         .to_vec(),
                 ),
                 Token::Bytes(self.metadata.l2_l1_messages_compressed.clone()),
-                Token::Bytes(self.header.pubdata_input.clone().unwrap()),
+                Token::Bytes(
+                    self.header
+                        .pubdata_input
+                        .clone()
+                        .unwrap_or(self.construct_pubdata()),
```
kpp, line 216: Should be unwrap_or_else, because `construct_pubdata()` is called every time no matter if you need it or not.

PR #791, opened 2024-01-01
feat(vm)!: fee model updates + 1.4.1

StanislavBreadless:
## What ❔

Adds a new VM version `VM_1_4_1` that will work with `zk_evm@1.4.1` and the [following contracts](https://github.com/matter-labs/era-contracts/pull/159). 

Generally, it contains the following features (on the server side):

- The new fee model
- 1.4.1 integration
- The Increase of the maximal number of transactions per batch to 10 thousand.

❗ Requires DB migration and new config params before being deployed❗

## Checklist


core/lib/zksync_core/src/api_server/tx_sender/mod.rs
```diff
@@ -875,13 +899,28 @@ impl<G: L1GasPriceProvider> TxSender<G> {
...
+            .blocks_dal()
+            .get_miniblock_protocol_version_id(block_args.resolved_block_number())
+            .await
+            .unwrap()
+            .unwrap_or(ProtocolVersionId::last_pre_boojum());
+        drop(connection);
+
+        let (base_fee, _) =
+            derive_base_fee_and_gas_per_pubdata(fee_model_params, protocol_version.into());
+
```
Deniallugo, line 923: I'd prefer to move it outside and use it as a pure function, how it was done before

Deniallugo (commented):
Not fully finished. 
But it's already some places to work

core/lib/types/src/commitment.rs
```diff
@@ -382,7 +382,7 @@ impl L1BatchAuxiliaryOutput {
             repeated_writes_compressed,
             system_logs_compressed,
             state_diffs_packed,
-        ) = if is_pre_boojum {
+        ) = if protocol_version.is_pre_boojum() {
```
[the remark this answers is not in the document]
koloz193: if its only used for the `l1_commit_data` then it can be removed but we just need to make sure it's not still being used to calculate gas limits, fees, or within metadata calculator under the hood

core/lib/multivm/src/versions/vm_latest/oracles/storage.rs
```diff
@@ -411,7 +414,7 @@ impl<S: WriteStorage, H: HistoryMode> VmStorageOracle for StorageOracle<S, H> {
                     }
                 };
 
-                let LogQuery { written_value, .. } = query.log_query;
+                let LogQuery { written_value, .. } = query.log_query.glue_into();
```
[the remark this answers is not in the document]
StanislavBreadless: Because the zksync_types uses the 1_3_2 ones. As per our discussion, I'll try to convert it into our custom `struct`, so that every version glues into it

core/lib/types/src/commitment.rs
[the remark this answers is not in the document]
Deniallugo: i'd prefer to remove it during this upgrade and keep only pre boojum stage here and for everything after boojum use the input from vm

core/lib/multivm/src/versions/vm_1_3_2/transaction_data.rs
```diff
@@ -58,12 +60,18 @@ impl From<Transaction> for TransactionData {
                     U256::zero()
                 };
 
+                let gas_per_pubdata_limit = if common_data.transaction_type.is_ethereum_type() {
```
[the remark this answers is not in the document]
StanislavBreadless: This part is actually tricky. With 1.4.1 we changed the maximal gas per pubdata. And in the bootloader there is a requirement that for legacy transactions (where the user does not sign gas per pubdata) the operator puts gas per pubdata that is not higher than the maximal one. 

So we get the following problem:
- If we put the new maximal limit, the old VM will not work
- If we put the old maximal limit, under high L1 gas price, the new VM may stop working.

That's why I moved resolving the "default" gas per pubdata for the ethereum (0,1,2) tx types to the VM's themselves

core/lib/multivm/src/versions/vm_m6/transaction_data.rs
```diff
@@ -60,12 +62,18 @@ impl From<Transaction> for TransactionData {
                     U256::zero()
                 };
 
+                let gas_per_pubdata_limit = if common_data.transaction_type.is_ethereum_type() {
+                    MAX_GAS_PER_PUBDATA_BYTE_PRE_1_4_1.into()
+                } else {
+                    common_data.fee.gas_per_pubdata_limit
+                };
```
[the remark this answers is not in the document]
StanislavBreadless: same as here: https://github.com/matter-labs/zksync-era/pull/791#discussion_r1439615795

core/lib/types/src/fee_model.rs
```diff
@@ -0,0 +1,179 @@
+use core::panic;
+
+use serde::{Deserialize, Serialize};
+use zksync_system_constants::L1_GAS_PER_PUBDATA_BYTE;
+
+use crate::ProtocolVersionId;
+
+/// Fee input to be provided into the VM. It contains two options:
+/// - L1Pegged: L1 gas price is provided to the VM, and the pubdata price is derived from it. Using this option is required for the
```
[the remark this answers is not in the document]
StanislavBreadless: would "pre 1.4.1" be a better wording?

core/lib/types/src/l2/mod.rs
```diff
@@ -35,6 +35,18 @@ pub enum TransactionType {
...
+    /// Returns whether a transaction type is an Ethereum transaction type.
+    pub fn is_ethereum_type(&self) -> bool {
+        match self {
+            TransactionType::LegacyTransaction => true,
+            TransactionType::EIP2930Transaction => true,
+            TransactionType::EIP1559Transaction => true,
+            _ => false,
+        }
+    }
+}
```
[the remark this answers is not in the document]
StanislavBreadless: This method is needed for this use-case : https://github.com/matter-labs/zksync-era/pull/791#discussion_r1439615795

core/lib/web3_decl/src/namespaces/zks.rs
```diff
@@ -97,6 +98,9 @@ pub trait ZksNamespace {
     #[method(name = "getL1GasPrice")]
     async fn get_l1_gas_price(&self) -> RpcResult<U64>;
 
+    #[method(name = "getMainNodeFeeParams")]
+    async fn get_main_node_fee_params(&self) -> RpcResult<MainNodeFeeParams>;
```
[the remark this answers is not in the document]
StanislavBreadless: not sure. `get_l1_gas_price` that only the `EN` uses is here

core/lib/zksync_core/src/api_server/execution_sandbox/apply.rs
```diff
@@ -173,10 +175,17 @@ pub(super) fn apply_vm_in_sandbox<T>(
         tracing::debug!("Prepared the storage view (took {storage_view_setup_time:?})",);
     }
 
+    if adjust_pubdata_price {
+        adjust_pubdata_price_for_tx(
+            &mut shared_args.batch_fee_model_input,
+            tx.gas_per_pubdata_byte_limit(),
+            protocol_version.into(),
+        );
+    }
```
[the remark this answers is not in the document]
StanislavBreadless: because before we do not have `protocol_version` yet

core/lib/constants/src/ethereum.rs
```diff
@@ -8,7 +8,8 @@ pub static ETHEREUM_ADDRESS: Address = Address::zero();
 /// This the number of pubdata such that it should be always possible to publish
 /// from a single transaction. Note, that these pubdata bytes include only bytes that are
 /// to be published inside the body of transaction (i.e. excluding of factory deps).
-pub const GUARANTEED_PUBDATA_PER_L1_BATCH: u64 = 4000;
+pub const GUARANTEED_PUBDATA_PER_L1_BATCH_PRE_1_4_1: u64 = 4000;
+pub const GUARANTEED_PUBDATA_PER_L1_BATCH_1_4_1: u64 = 2500;
```
RomanBrodetski, line 12: Looking at all these changes (including the one from https://github.com/matter-labs/zksync-era/pull/809), I'm wondering if it would make more sense to have all these inside the zk_evm crate. In that case we'd just need to reference the new crate version.

@Deniallugo I'm sure you've already thought about it - what's the current stance? It will probably only be feasible when we integarte JayVM

core/lib/dal/src/blocks_dal.rs
```diff
@@ -595,6 +596,16 @@ impl BlocksDal<'_, '_> {
     ) -> anyhow::Result<()> {
         let base_fee_per_gas = BigDecimal::from_u64(miniblock_header.base_fee_per_gas)
             .context("base_fee_per_gas should fit in u64")?;
+        let gas_per_pubdata_limit = if miniblock_header
+            .protocol_version
+            .unwrap_or(ProtocolVersionId::last_pre_boojum())
+            .is_1_4_1()
```
RomanBrodetski, line 602: I might be missing something, but won't it be a problem when we integrate the next VM version? You check if the current vm is `1.4.1` and **if not** you apply `MAX_GAS_PER_PUBDATA_BYTE_PRE_1_4_1`. When we create 1.5, won't we by mistake use the old value? Compiler won't help here.

core/lib/dal/src/models/storage_block.rs
```diff
@@ -522,20 +525,40 @@ pub struct StorageMiniblockHeader {
 
 impl From<StorageMiniblockHeader> for MiniblockHeader {
     fn from(row: StorageMiniblockHeader) -> Self {
+        let protocol_version = row.protocol_version.map(|v| (v as u16).try_into().unwrap());
+
+        let fee_input = protocol_version
+            .filter(|version: &ProtocolVersionId| version.is_1_4_1())
```
RomanBrodetski, line 531: Same here - will this work when we further increase the version?

core/lib/types/src/fee_model.rs
```diff
@@ -0,0 +1,179 @@
...
+    pub l1_gas_price: u64,
+}
+
+/// Pubdata price may be independent from L1 gas price. The L1 gas price is not needed.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct PubdataIndependentBatchFeeModelInput {
+    /// Fair L2 gas price to provide
+    pub fair_l2_gas_price: u64,
+    /// Fair pubdata price to provide.
+    /// In this version, it MUST be equal to 17 * l1_gas_price
```
RomanBrodetski, line 129: in what version?

core/lib/dal/src/blocks_dal.rs
RomanBrodetski: oh now I see that the implementation of `is_1_4_1` checks if its `1.4.1` **or later**. So it will work correctly, but it is still confusing. Perhaps `is_post_1-4-1`?

core/lib/zksync_core/src/api_server/execution_sandbox/execute.rs
```diff
@@ -108,40 +109,14 @@ pub(crate) async fn execute_tx_eth_call(
...
-
 /// This method assumes that (block with number `resolved_block_number` is present in DB)
 /// or (`block_id` is `pending` and block with number `resolved_block_number - 1` is present in DB)
 #[allow(clippy::too_many_arguments)]
 #[tracing::instrument(skip_all)]
-async fn execute_tx_in_sandbox(
+pub(crate) async fn execute_tx_in_sandbox(
     vm_permit: VmPermit,
     shared_args: TxSharedArgs,
+    adjust_pubdata_price: bool,
```
RomanBrodetski, line 119: please add a comment on this field. When do we need to adjust the pubdata price? When not?

RomanBrodetski: thinking more about this - I think it's a little bit of a code smell that we need to do something like this. So we pass an instance of `TxSharedArgs` where fee-related fields are _sometimes_ incomplete. Fixing this might require larger refactoring. 

Another code smell here (not introdeuced by this PR) is that we need to pass this `apply` lambda function. Did anyone already thing how this can be improved? cc @popzxc 

This is of course not for this PR

popzxc: Re `TxSharedArgs` / `apply`: probably we need a enum here.
Re `apply` lambda: I think it's caused by the fact that VM initialization is complex and very customizable. Probably it's possible to make it nicer, but we've already had 3 or 4 rounds of refactoring this module, and `apply` closure outlived all of them, so I'd say that likely it won't be easy to get rid of it.

core/lib/zksync_core/src/fee_model.rs
```diff
@@ -0,0 +1,126 @@
...
+        PubdataIndependentBatchFeeModelInput,
+    },
+    U256,
+};
+
+use crate::l1_gas_price::L1GasPriceProvider;
+
+/// Trait responsiblef for providign fee info for a batch
+pub trait BatchFeeModelInputProvider: fmt::Debug + 'static + Send + Sync {
+    fn get_batch_fee_input(&self, scale_l1_prices: bool) -> BatchFeeInput {
```
RomanBrodetski, line 15: `scale_l1_prices` is not really self-explonatory...

core/lib/constants/src/ethereum.rs
StanislavBreadless: These changes are just not inherent to `zk_evm`. Things like `MAX_BOOTLOADER_MEMORY`/`MAX_TXS_PER_BATCH`/`GUARANTEED_PUBDATA`/etc are constants that are related to a specific bootloader version. I used the "1_4_1" here as it is how the upgrade will be called (I can call it some other name)

core/lib/zksync_core/src/fee_model.rs
StanislavBreadless: I'll add a comment

core/lib/constants/src/ethereum.rs
RomanBrodetski: I agree. I'm just missing some abstraction that would encapsulate everything related to a partoicular protocol version.
No need for any changes in this PR

core/lib/types/src/fee_model.rs
StanislavBreadless: The comment is wrong, I'll delete it

core/lib/types/src/fee_model.rs
```diff
@@ -0,0 +1,179 @@
...
+        }
+    }
+
+    pub fn for_protocol_version(
+        protocol_version: ProtocolVersionId,
+        fair_l2_gas_price: u64,
+        fair_pubdata_price: Option<u64>,
+        l1_gas_price: u64,
+    ) -> Self {
+        if protocol_version.is_1_4_1() {
```
mm-zk, line 98: this is a little bit dangerous - if we roll out 1.5.0 - then this will return 'false' - and we'll switch back to L1Pegged.

core/lib/types/src/fee_model.rs
```diff
@@ -0,0 +1,179 @@
...
+                l1_gas_price,
+            })
+        }
+    }
+}
+
+/// Pubdata is only published via calldata and so its price is pegged to the L1 gas price.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct L1PeggedBatchFeeModelInput {
+    /// Fair L2 gas price to provide
```
mm-zk, line 117: we should explain more what 'fair' here means. I guess it should 'only' cover the CPU / prover costs, right?

core/lib/types/src/fee_model.rs
```diff
@@ -0,0 +1,179 @@
...
+            max_pubdata_per_batch: 120_000_000,
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+pub struct MainNodeFeeParams {
+    pub config: MainNodeFeeModelConfig,
+    pub l1_gas_price: u64,
+    pub l1_pubdata_price: u64,
```
mm-zk, line 178: comment

core/lib/types/src/fee_model.rs
```diff
@@ -0,0 +1,179 @@
...
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+pub struct MainNodeFeeModelConfig {
+    /// The factor by which the L1 gas price is scaled. This is used to account for the fact that the L1 gas price may fluctuate.
+    pub l1_gas_price_scale_factor: f64,
+    /// The factor by which the L1 pubdata price is scaled. This is used to account for the fact that the L1 pubdata price may fluctuate.
+    pub l1_pubdata_price_scale_factor: f64,
+    /// The minimal acceptable L2 gas price, i.e. the price that should include the cost of computation/proving as well
+    /// as potentially premium for congestion.
+    pub minimal_l2_gas_price: u64,
+    /// Tthe constant that represents the possibility that a batch can be sealed because of overuse of compute.
```
mm-zk, line 144: typo in 'Tthe'

core/lib/types/src/fee_model.rs
```diff
@@ -0,0 +1,179 @@
...
+    pub compute_overhead_percent: f64,
+    /// The constant that represents the possibility that a batch can be sealed because of overuse of pubdata.
```
mm-zk, line 146: so this should be between 0 & 1 ?

core/lib/types/src/fee_model.rs
```diff
@@ -0,0 +1,179 @@
...
+    pub pubdata_overhead_percent: f64,
+    /// The constant amount of L1 gas that is used as the overhead for the batch. It includes the price for batch verification, etc.
+    pub batch_overhead_l1_gas: u64,
+    /// The maximum amount of gas that can be used by the batch.
+    pub max_gas_per_batch: u64,
+    /// The maximum amount of pubdata that can be used by the batch.
```
mm-zk, line 152: mention, that it should not exceed the calldata limits for now (128k)

core/lib/types/src/fee_model.rs
```diff
@@ -0,0 +1,179 @@
...
+    /// The minimal acceptable L2 gas price, i.e. the price that should include the cost of computation/proving as well
+    /// as potentially premium for congestion.
+    pub minimal_l2_gas_price: u64,
+    /// Tthe constant that represents the possibility that a batch can be sealed because of overuse of compute.
+    pub compute_overhead_percent: f64,
+    /// The constant that represents the possibility that a batch can be sealed because of overuse of pubdata.
+    pub pubdata_overhead_percent: f64,
+    /// The constant amount of L1 gas that is used as the overhead for the batch. It includes the price for batch verification, etc.
+    pub batch_overhead_l1_gas: u64,
+    /// The maximum amount of gas that can be used by the batch.
```
mm-zk, line 150: mention what are the natural limitations - why people should restrict it and not set to 100B

core/lib/types/src/fee_model.rs
```diff
@@ -0,0 +1,179 @@
...
+
+use crate::ProtocolVersionId;
+
+/// Fee input to be provided into the VM. It contains two options:
+/// - L1Pegged: L1 gas price is provided to the VM, and the pubdata price is derived from it. Using this option is required for the
+/// older versions of Era.
+/// - PubdataIndependent: L1 gas price and pubdata price are not necessarily dependend on one another. This options is more suitable for the
+/// newer versions of Era.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum BatchFeeInput {
```
mm-zk, line 14: I'd suggest that we start with the PR that only introduces the BatchFeeInput (with L1Pegged). (as a part of breaking this large PR into pieces)

mm-zk (changes requested):

core/lib/types/src/fee_model.rs
StanislavBreadless: No, actually "fair" would need to cover the L1 batch overhead costs in the new fee model

core/lib/types/src/fee_model.rs
StanislavBreadless: yes

core/lib/types/src/fee_model.rs
StanislavBreadless: BTW, are you sure? I think that once we introduce the blob support we'll forget to update the comment. So far the variable name seems self explanatory

core/lib/types/src/fee_model.rs
StanislavBreadless: Yes: https://github.com/matter-labs/zksync-era/pull/817/files

core/lib/constants/src/ethereum.rs
Deniallugo: yeah, we could probably need an abstraction for bootloader and configs

core/lib/types/src/fee_model.rs
StanislavBreadless: I changed to `is_post_1_4_1`

core/lib/zksync_core/src/metadata_calculator/updater.rs
```diff
@@ -204,14 +206,14 @@ impl TreeUpdater {
             .await
             .unwrap();
 
-        let is_pre_boojum = header
+        // TODO(X): ensure that the protocol version is always available.
```
mm-zk, line 209: is this TODO still needed?

core/lib/zksync_core/src/metadata_calculator/mod.rs
```diff
@@ -191,10 +191,12 @@ impl MetadataCalculator {
         events_queue_commitment: Option<H256>,
         bootloader_initial_content_commitment: Option<H256>,
     ) -> L1BatchMetadata {
-        let is_pre_boojum = header
+        // The commitment generation pre-boojum is the same for all the version, so in case the version is not present, we just supply the
+        // last pre-boojum version.
+        // TODO(X): make sure that protocol version is not an Option
```
mm-zk, line 196: is this needed?

core/lib/multivm/src/versions/vm_latest/constants.rs
```diff
@@ -1,23 +1,38 @@
...
+// to be published inside the body of transaction (i.e. excluding of factory deps).
+pub(crate) const GUARANTEED_PUBDATA_PER_L1_BATCH: u64 = 2500;
+
+// The users should always be able to provide `MAX_GAS_PER_PUBDATA_BYTE` gas per pubdata in their
+// transactions so that they are able to send at least `GUARANTEED_PUBDATA_PER_L1_BATCH` bytes per
+// transaction.
+pub(crate) const MAX_GAS_PER_PUBDATA_BYTE: u64 =
+    MAX_L2_TX_GAS_LIMIT / GUARANTEED_PUBDATA_PER_L1_BATCH;
+
 // The maximal number of transactions in a single batch
```
mm-zk, line 26: add a comment that it was increased in this VM version.

core/lib/multivm/src/versions/vm_latest/constants.rs
```diff
@@ -1,23 +1,38 @@
...
-    USED_BOOTLOADER_MEMORY_WORDS,
-};
+use zksync_system_constants::{MAX_L2_TX_GAS_LIMIT, MAX_NEW_FACTORY_DEPS};
 
 use crate::vm_latest::old_vm::utils::heap_page_from_base;
 
+/// The size of the bootloader memory in bytes which is used by the protocol.
+/// While the maximal possible size is a lot higher, we restrict ourselves to a certain limit to reduce
+/// the requirements on RAM.
+pub(crate) const USED_BOOTLOADER_MEMORY_BYTES: usize = 24_000_000;
```
mm-zk, line 12: add a comment that it used to be smaller, but now it was increased.

core/lib/multivm/src/versions/vm_latest/constants.rs
```diff
@@ -101,17 +113,17 @@ pub const VM_HOOK_POSITION: u32 = RESULT_SUCCESS_FIRST_SLOT - 1;
 pub const VM_HOOK_PARAMS_COUNT: u32 = 2;
 pub const VM_HOOK_PARAMS_START_POSITION: u32 = VM_HOOK_POSITION - VM_HOOK_PARAMS_COUNT;
 
-pub(crate) const MAX_MEM_SIZE_BYTES: u32 = 16777216; // 2^24
+pub(crate) const MAX_MEM_SIZE_BYTES: u32 = 24000000;
```
mm-zk, line 116: mention why was it changed. And what's the relation to USED_BOOTLOADER_MEMORY_BYTES

core/lib/multivm/src/versions/vm_latest/constants.rs
```diff
@@ -124,7 +136,16 @@ pub(crate) const TX_OPERATOR_L2_BLOCK_INFO_OFFSET: usize =
 
 pub(crate) const TX_OPERATOR_SLOTS_PER_L2_BLOCK_INFO: usize = 4;
 pub(crate) const TX_OPERATOR_L2_BLOCK_INFO_SLOTS: usize =
-    (MAX_TXS_IN_BLOCK + 1) * TX_OPERATOR_SLOTS_PER_L2_BLOCK_INFO;
+    (MAX_TXS_IN_BATCH + 1) * TX_OPERATOR_SLOTS_PER_L2_BLOCK_INFO;
 
 pub(crate) const COMPRESSED_BYTECODES_OFFSET: usize =
     TX_OPERATOR_L2_BLOCK_INFO_OFFSET + TX_OPERATOR_L2_BLOCK_INFO_SLOTS;
+
+/// The maximal gas limit that gets passed into an L1->L2 transaction
```
mm-zk, line 144: maybe mention what are the constraints here (why is it 72M and not higher or lower)

core/lib/multivm/src/versions/vm_latest/constants.rs
```diff
@@ -124,7 +136,16 @@ pub(crate) const TX_OPERATOR_L2_BLOCK_INFO_OFFSET: usize =
...
+pub(crate) const PRIORITY_TX_MAX_GAS_LIMIT: usize = 72_000_000;
+
+/// The amount of gas to be charged for occupying a single slot of a transaction.
```
mm-zk, line 147: same here - how did we come up with this number.

core/lib/multivm/src/versions/vm_latest/constants.rs
```diff
@@ -124,7 +136,16 @@ pub(crate) const TX_OPERATOR_L2_BLOCK_INFO_OFFSET: usize =
...
+pub(crate) const TX_SLOT_OVERHEAD_GAS: u32 = 10_000;
+
+/// The amount of gas to be charged for occupying a single byte of the bootloader's memory.
+pub(crate) const TX_MEMORY_OVERHEAD_GAS: u32 = 10;
```
mm-zk, line 151: and here.

core/lib/multivm/src/versions/vm_latest/utils/overhead.rs
```diff
@@ -1,356 +1,9 @@
...
-        derive_overhead(
-            MAX_L2_TX_GAS_LIMIT as u32,
-            gas_per_pubdata_byte_limit,
-            encoded_len.as_usize(),
-            coefficients,
-        )
-    } else {
-        overhead
-    }
+pub(crate) fn derive_overhead(encoded_len: usize) -> u32 {
```
mm-zk, line 3: Add a comment.

And mention that in the past we used different rules, but with the new fee model, we only consider encoded len - as other part (like slots, memory, gas and pubdata are computed somewhere else)

core/lib/multivm/src/versions/vm_latest/utils/overhead.rs
```diff
@@ -1,356 +1,9 @@
...
-        test_params(
-            MAX_TX_ERGS_LIMIT + max_tx_overhead,
-            5000,
-            10000,
-            OverheadCoefficients::new_l2(),
-        );
-
-        test_params(115432560, 800, 2900, OverheadCoefficients::new_l1());
-    }
+pub(crate) fn get_amortized_overhead(encoded_len: usize) -> u32 {
```
mm-zk, line 7: mention why this method even exists ;-)

core/lib/multivm/src/versions/vm_latest/tracers/refunds.rs
```diff
@@ -146,6 +144,15 @@ impl<S> RefundsTracer<S> {
             U256::zero()
         });
 
+        tracing::trace!(
+            "Fee benchmark for transaction with hash {}",
+            hex::encode(tx_hash.as_bytes())
+        );
+        tracing::trace!("Gas Limit: {}", tx_gas_limit);
+        tracing::trace!("Gas spent on computation: {}", gas_spent_on_computation);
```
mm-zk, line 152: I wonder we we could persist some of this inside the execution_info json in DB. (this would make debugging a lot easier)

core/lib/multivm/src/versions/vm_latest/tracers/refunds.rs
```diff
@@ -334,7 +343,7 @@ pub(crate) fn pubdata_published<S: WriteStorage, H: HistoryMode>(
         })
         .filter(|log| log.sender != SYSTEM_CONTEXT_ADDRESS)
         .count() as u32)
-        * zk_evm_1_4_0::zkevm_opcode_defs::system_params::L1_MESSAGE_PUBDATA_BYTES;
+        * zk_evm_1_4_1::zkevm_opcode_defs::system_params::L1_MESSAGE_PUBDATA_BYTES;
```
mm-zk, line 346: can we move this zk_evm_1_4_1 import to the top (so that it is there with all the others?)

core/lib/types/src/commitment.rs
```diff
@@ -742,7 +753,8 @@ mod tests {
         expected_outputs: ExpectedOutput,
     }
 
-    #[test]
+    // TODO(X): restore this test
```
mm-zk, line 756: can we restore it now? (if not - can you explain more what is broken)

core/lib/zksync_core/src/metadata_calculator/updater.rs
StanislavBreadless: cc @perekopskiy ?

It is not needed for this PR, but I thought that maybe we do have plans to do it in the future

core/lib/zksync_core/src/metadata_calculator/mod.rs
StanislavBreadless: Same as here: https://github.com/matter-labs/zksync-era/pull/791#discussion_r1447282310

core/lib/types/src/commitment.rs
StanislavBreadless: the commitment calculation is broken, I would the new correct commitment to restore the test

StanislavBreadless: probably it is worth it to just wait until the prover integration is done, then we know for sure that this commitment calculation is right

core/lib/multivm/src/versions/vm_latest/utils/overhead.rs
StanislavBreadless: I'll remove it

core/lib/multivm/src/versions/vm_latest/tracers/refunds.rs
StanislavBreadless: cc @perekopskiy WDYT? Obviously out of scope of the PR

core/lib/multivm/src/versions/vm_m5/transaction_data.rs
```diff
@@ -59,12 +60,22 @@ impl From<Transaction> for TransactionData {
...
 
+                // Ethereum transactions do not sign gas per pubdata limit, and so for them we need to use
+                // some default value. We use the maximum possible value that is allowed by the bootloader
+                // (i.e. we can not use u64::MAX, because the bootloader requires gas per pubdata for such
+                // transactions to be higher than MAX_GAS_PER_PUBDATA_BYTE).
+                let gas_per_pubdata_limit = if common_data.transaction_type.is_ethereum_type() {
+                    MAX_GAS_PER_PUBDATA_BYTE.into()
+                } else {
+                    common_data.fee.gas_per_pubdata_limit
+                };
```
Deniallugo, line 71: why do you need to change it for old vms?

core/lib/multivm/src/versions/vm_latest/types/internals/transaction_data.rs
```diff
@@ -212,16 +211,16 @@ impl TransactionData {
...
-        )
+        derive_overhead(encoded_len)
     }
 
-    pub(crate) fn trusted_ergs_limit(&self, _block_gas_price_per_pubdata: u64) -> U256 {
+    pub(crate) fn trusted_ergs_limit(&self) -> U256 {
+        if self.tx_type == L1_TX_TYPE {
+            // In case we get a users' transactions with unexpected gas limit, we do not let it have more than
+            // a certain limit
+            return U256::from(PRIORITY_TX_MAX_GAS_LIMIT);
```
Deniallugo, line 221: should it be `min(self.gas_limit, PRIORITY_TX_MAX_GAS_LIMIT)`?

core/tests/ts-integration/tests/api/web3.test.ts
```diff
@@ -218,7 +218,7 @@ describe('web3 API compatibility tests', () => {
         await legacyTx.wait();
 
         const legacyApiReceipt = await alice.provider.getTransaction(legacyTx.hash);
-        expect(legacyApiReceipt.gasPrice).bnToBeEq(legacyTx.gasPrice!);
+        expect(legacyApiReceipt.gasPrice).bnToBeLte(legacyTx.gasPrice!);
```
Deniallugo, line 221: why?

core/lib/commitment_utils/src/lib.rs
```diff
@@ -1,23 +1,33 @@
 //! Utils for commitment calculation.
+use multivm::utils::get_used_bootloader_memory_bytes;
 use zkevm_test_harness::witness::utils::{
     events_queue_commitment_fixed, initial_heap_content_commitment_fixed,
```
perekopskiy, line 4: double-checking: there is no difference in commitment generation between 1.4.0 and 1.4.1, right?

core/lib/dal/sqlx-data.json
```diff
@@ -0,0 +1,3 @@
+{
+  "db": "PostgreSQL"
+}
```
perekopskiy, line 3: I guess this file should be removed

core/lib/multivm/src/versions/vm_latest/tracers/refunds.rs
perekopskiy: Yes, I think it's a good idea. Shouldn't be that hard to save it

core/lib/zksync_core/src/metadata_calculator/updater.rs
perekopskiy: We definitely have these plans, I'm just not sure when we will capacity to implement them

etc/env/base/chain.toml
```diff
@@ -45,8 +45,37 @@ reject_tx_at_eth_params_percentage=0.95
 # it takes more percentage of the max block gas capacity than this value.
 reject_tx_at_gas_percentage=0.95
 
-# The price the operator spends on 1 gas of computation in wei.
-fair_l2_gas_price=250000000
+# The minimal acceptable L2 gas price, i.e. the price that should include the cost of computation/proving as well
+# as potentially premium for congestion.
+minimal_l2_gas_price=100000000
```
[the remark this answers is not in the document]
StanislavBreadless: Discussed in private with @mm-zk, the comment is good enough

core/bin/system-constants-generator/src/main.rs
```diff
@@ -56,23 +60,22 @@ pub fn generate_l1_contracts_system_config(gas_constants: &IntrinsicSystemGasCon
         priority_tx_max_pubdata: (L1_TX_DECREASE * (MAX_PUBDATA_PER_BLOCK as f64)) as u32,
         fair_l2_gas_price: FAIR_L2_GAS_PRICE_ON_L1_CONTRACT,
         l1_gas_per_pubdata_byte: L1_GAS_PER_PUBDATA_BYTE,
-        block_overhead_l2_gas: BLOCK_OVERHEAD_GAS,
+        block_overhead_l2_gas: 0,
```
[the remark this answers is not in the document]
StanislavBreadless: Just of simplification, since L2 gas overhead <<< L1 gas overhead, I decided to remove it altogether to keep the calculations simple. I'll probably just remove the constant to not confuse anyone

core/lib/multivm/src/versions/vm_m5/transaction_data.rs
StanislavBreadless: because in the old VMs `MAX_GAS_PER_PUBDATA_BYTE` was lower than for the new one

StanislavBreadless: So basically this `if` is what used to be done implicitly on the API side. But now, since the constant has changed, it has to be done individually for each VM version

core/lib/multivm/src/versions/vm_latest/types/internals/transaction_data.rs
StanislavBreadless: Not necessarily (the bootloader would never give more gas than the transaction offers), but yes, better to add the `min` for more clarity

core/lib/commitment_utils/src/lib.rs
StanislavBreadless: to the best of my knowledge no, but I'll probably just use the newer methods just in case

StanislavBreadless: Making it the other way would create a big diff for this PR, so I'll double check with the crypto team + create a follow up PR

mm-zk (dismissed):
LGTM - but please take a look at the VM performance numbers - as it seems that VM is 20% slower after these changes - and I wonder why.

Deniallugo (approved):

Deniallugo (approved):

perekopskiy (dismissed):

perekopskiy (approved):

PR #800, opened 2024-01-02
chore(eth sender): Modify Default Gas Estimation Values

shahar4:
## What ❔

This PR changes the default values of the parameters responsible for gas estimations.

## Why ❔

During times of congestion, hyperchains operators have been running into `replacement transaction underpriced` errors. The new params will be less frugal in estimating gas which seems like the better choice for testnets.

## Checklist


Deniallugo:
It will drastically increase the price of l2

PR #803, opened 2024-01-03
refactor(state_keeper): Abstract ConditionalSealer

popzxc:
## What ❔

Defines a new trait: `ConditionalSealer` and provides two implementations: `SequencerSealer` and `NoopSealer`.

## Why ❔

- Prerequisite for ZK Stack configuration system.
- Leaves a single constructor for the state keeper.
- Removes `StateKeeperConfig` use from `TxSender`.
- Potentially makes it possible to create `ConditionalSealer` that contains some kind of sealing logic but does not rely on `StateKeeperConfig`.

## Checklist

### DIFF
--- a/core/lib/zksync_core/src/lib.rs
+++ b/core/lib/zksync_core/src/lib.rs
@@ -70,7 +70,9 @@ use crate::{
         MetadataCalculator, MetadataCalculatorConfig, MetadataCalculatorModeConfig,
     },
     metrics::{InitStage, APP_METRICS},
-    state_keeper::{create_state_keeper, MempoolFetcher, MempoolGuard, MiniblockSealer},
+    state_keeper::{
+        create_state_keeper, MempoolFetcher, MempoolGuard, MiniblockSealer, SequencerSealer,
+    },
 };
 
 pub mod api_server;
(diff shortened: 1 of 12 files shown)


core/lib/zksync_core/src/lib.rs
```diff
@@ -5,6 +5,7 @@ use std::{net::Ipv4Addr, str::FromStr, sync::Arc, time::Instant};
 use anyhow::Context as _;
 use futures::channel::oneshot;
 use prometheus_exporter::PrometheusExporterConfig;
+use state_keeper::SequencerSealer;
```
slowli, line 8: Nit: This is not an external import; it should be prefixed with `crate::` to make this clear.

slowli (approved):

core/lib/zksync_core/src/lib.rs
popzxc: Fixed. Kinda sad that `rustfmt` can't do that automatically, since looks like `rust-analyzer` autoimport functionality is somewhat random.

slowli (approved):

core/lib/zksync_core/src/state_keeper/seal_criteria/conditional_sealer.rs
```diff
@@ -114,3 +142,31 @@ impl ConditionalSealer {
         ]
     }
 }
+
+/// Implementation of [`ConditionalSealer`] that never seals the batch.
+/// Can be used in contexts where, for example, state keeper configuration is not available,
+/// or the decision to seal batch is taken by some other component.
+#[derive(Debug)]
```
RomanBrodetski, line 149: what are the examples of such contexts? EN is mentioned elsewhere - makes sense that the EN doesn't seal l1 batches (for now). But I think the reason for this is that we are afraid of divergencies - in theory the EN should apply the same sealing criteria and panic if the l1 batch is sealed "incorrectly".

RomanBrodetski (approved):

core/lib/zksync_core/src/state_keeper/seal_criteria/conditional_sealer.rs
popzxc: > what are the examples of such contexts? 

API server on EN that accepts any transaction and defers checks to the main node.

> But I think the reason for this is that we are afraid of divergencies

We're not afraid of divergencies. The sealer on the main node uses `StateKeeperConfig` and EN doesn't currently have a mechanism to synchronize the settings with the main node if they ever change (which absolutely can happen).

popzxc: Uh, and also unit tests, obviously.

PR #809, opened 2024-01-03
feat(vm): Make utils version-dependent

StanislavBreadless:
## What ❔

In the new 1.4.1 version, the fee formulas as well as max txs per batch will change. We'll prepare the main branch for it first to reduce the diff

## Checklist


core/lib/zksync_core/src/api_server/execution_sandbox/execute.rs
```diff
@@ -108,40 +109,41 @@ pub(crate) async fn execute_tx_eth_call(
...
+//         vm_permit,
+//         shared_args,
+//         execution_args,
+//         connection_pool,
+//         tx,
+//         block_args,
+//         vec![],
+//     )
+//     .await
+// }
```
Deniallugo, line 137: Why?

core/lib/zksync_core/src/api_server/execution_sandbox/validate.rs
```diff
@@ -13,44 +13,44 @@ use zksync_dal::{ConnectionPool, StorageProcessor};
...
+    //     mut self,
+    //     vm_permit: VmPermit,
+    //     connection_pool: ConnectionPool,
+    //     tx: L2Tx,
+    //     computational_gas_limit: u32,
+    // ) -> Result<(), ValidationError> {
+    //     let mut connection = connection_pool.access_storage_tagged("api").await.unwrap();
+    //     let block_args = BlockArgs::pending(&mut connection).await;
+    //     drop(connection);
+    //     self.adjust_l1_gas_price(tx.common_data.fee.gas_per_pubdata_limit);
```
Deniallugo, line 32: Why again? :)

core/lib/zksync_core/src/state_keeper/mempool_actor.rs
```diff
@@ -77,7 +80,18 @@ impl<G: L1GasPriceProvider> MempoolFetcher<G> {
...
+                .protocol_versions_dal()
+                .base_system_contracts_by_timestamp(seconds_since_epoch())
+                .await
+                .1;
+
+            let l2_tx_filter = l2_tx_filter(
+                self.l1_gas_price_provider.as_ref(),
+                fair_l2_gas_price,
+                protocol_version.into(),
+            );
```
Deniallugo, line 94: The more I read it, seems we need a kinda Oracle with the current protocol version of the system. 
@popzxc wdyt?

core/lib/zksync_core/src/api_server/tx_sender/mod.rs
```diff
@@ -618,6 +624,23 @@ impl TxSender {
...
+            .await
+            .unwrap();
+        let block_args = BlockArgs::pending(&mut connection).await;
+        // If protocol version is not present, we'll use the pre-boojum one
+        let protocol_version = connection
+            .blocks_dal()
+            .get_miniblock_protocol_version_id(block_args.resolved_block_number())
+            .await
+            .unwrap()
+            .unwrap_or(ProtocolVersionId::last_pre_boojum());
```
Deniallugo, line 641: i think it's safe to just unwrap

core/lib/zksync_core/src/state_keeper/mempool_actor.rs
popzxc: What do you mean? Some kind of in-memory store to avoid accessing DAL?
Probably it's a nice-to-have, but generally, it seems to be a premature optimization.

core/lib/zksync_core/src/api_server/execution_sandbox/execute.rs
```diff
@@ -108,40 +109,41 @@ pub(crate) async fn execute_tx_eth_call(
...
 
 /// This method assumes that (block with number `resolved_block_number` is present in DB)
 /// or (`block_id` is `pending` and block with number `resolved_block_number - 1` is present in DB)
 #[allow(clippy::too_many_arguments)]
 #[tracing::instrument(skip_all)]
-async fn execute_tx_in_sandbox(
+pub(crate) async fn execute_tx_in_sandbox(
     vm_permit: VmPermit,
     shared_args: TxSharedArgs,
+    adjust_pubdata_price: bool,
```
RomanBrodetski, line 146: this field deserves a comment

core/lib/multivm/src/utils.rs
```diff
@@ -0,0 +1,164 @@
...
+        VmVersion::VmVirtualBlocksRefundsEnhancement => {
+            crate::vm_refunds_enhancement::constants::BOOTLOADER_TX_ENCODING_SPACE
+        }
+        VmVersion::VmBoojumIntegration => {
+            crate::vm_boojum_integration::constants::BOOTLOADER_TX_ENCODING_SPACE
+        }
+    }
+}
+
+pub fn get_max_transactions_in_batch(version: VmVersion) -> usize {
```
RomanBrodetski, line 149: maybe specify that this has to do with the bootloader? Something like `get_bootloader_l1_batch_tx_limit`

core/lib/zksync_core/src/state_keeper/mempool_actor.rs
RomanBrodetski: Do we have a struct that represents every option/config parameter that can differ between VM implementations - obviously, apart from the VM itself which lives in zk_evm repo? Do I understand it right that currently to update VM version we need to do:
* everything otherwise needed to update protocol version
* reference corresponding tags from crypto deps in our cargo.toml
* implement multivm-related methods (or copy/use the implementaiton of the previous version) - such as the ones that this PR introduces

Is there a way to get an overview of all the particularities related to a particular VM version?

Deniallugo: in the database we have this struct 
image

Normally it represents everything.

Deniallugo: @popzxc It's a bit scary part right here. 
because we get the protocol version, not from the latest miniblock, but from the timestamp. 

The timestamp is not when we are doing an upgrade, but when we can start an upgrade and the upgrade will start with the next batch. 

So, it have to track somehow for sure what is the current protocol version of the system

popzxc: Hm, agree. It does sound kinda dangerous.

core/lib/zksync_core/src/api_server/execution_sandbox/execute.rs
StanislavBreadless: to pass `block_args` more efficiently, but yeah, I'll def remove this commented out code

core/lib/zksync_core/src/api_server/execution_sandbox/validate.rs
StanislavBreadless: some here

core/lib/zksync_core/src/api_server/tx_sender/mod.rs
StanislavBreadless: Then maybe we should remove `Option` there at all? cc @perekopskiy

core/lib/zksync_core/src/state_keeper/mempool_actor.rs
StanislavBreadless: @Deniallugo this is mempool fetcher that usually retrieves transactions that must not need to comply with the parameters of the current batch (for instance, it re-queries L1 gas price all the time). I can change the logic to query the protocol version from the latest miniblock. The worst thing that can happen is that the transactions won't be picked up for some very short period of time after which the server will be updated and everything will work as needed

core/lib/zksync_core/src/api_server/tx_sender/mod.rs
perekopskiy: It's not safe, ENs still don't have this data

perekopskiy: It's not safe, ENs don't have this data. There is a plan to perform a migration but it's not a top priority

core/lib/zksync_core/src/state_keeper/mempool_actor.rs
perekopskiy: @StanislavBreadless I think it's better to query protocol version of the latest miniblock, unless fee formulas will be quite different and there is a risk of getting stuck

StanislavBreadless: @perekopskiy they will be different, but we'll try to make the transition as smooth as possible

perekopskiy (approved):

Deniallugo (approved):

PR #813, opened 2024-01-04
chore: normalize file path

MattewGraham:
## What ❔

normalize file path

## Why ❔

Here's the difference between using path.resolve with and without __dirname:

When using path.resolve without __dirname, the function will resolve the path based on the current working directory (CWD) of the Node.js process. In other words, if you provide a relative path without __dirname, the function will assume that the path is relative to the CWD.

On the other hand, when using path.resolve with __dirname, the function will resolve the path based on the directory where the current module file is located. In other words, if you provide a relative path with __dirname, the function will assume that the path is relative to the directory where the module file is located.

The difference lies in how the function interprets the relative path. Without __dirname, the function assumes that the path is relative to the CWD, which may not be the desired behavior in some cases. For example, if you have a folder structure like this:

my_project/
│ index.js
│ node_modules/
│ lib/
└── main.js
If you call path.resolve(__dirname, '../lib/my_file.js'), without using __dirname, the function will resolve the path to /./lib/my_file.js, which may not be what you intended. However, if you use path.resolve(__dirname, 'my_file.js'), the function will resolve the path to my_file.js in the lib folder, which is the desired behavior.

In summary, using path.resolve with __dirname can help you write more predictable and robust code by ensuring that the relative path is resolved based on the correct directory.

## Checklist


core/tests/ts-integration/src/helpers.ts
```diff
@@ -23,7 +24,8 @@ export function getTestContract(name: string): ZkSyncArtifact {
  */
 export function getContractSource(relativePath: string): string {
     const contractPath = `${process.env.ZKSYNC_HOME}/core/tests/ts-integration/contracts/${relativePath}`;
-    const source = fs.readFileSync(contractPath, 'utf8');
+    const filePath = path.resolve(__dirname, contractPath);
```
perekopskiy, line 27: `ZKSYNC_HOME` is an absolute path, so `contractPath` is also absolute and there is no difference between resolving based on CWD on __dirname.

MattewGraham: > `ZKSYNC_HOME` is an absolute path, so `contractPath` is also absolute and there is no difference between resolving based on CWD on __dirname.

Ok, Thanks

PR #817, opened 2024-01-04
feat(vm): Add batch input abstraction

StanislavBreadless:
## What ❔

The new fee model will allow us to have separate L1 gas price and pubdata price (and it is needed for EIP4844). That's why the `BatchFeeInput` struct is needed as it provides the needed fields for both the VM version.

The version that is provided in this PR will require not additional changes on the server side (i.e. whatever config params that we have right now is enough, the full integration of the config params for the new fee model will be added to the main PR: https://github.com/matter-labs/zksync-era/pull/791/.

While the defined enum allows for both l1-gas-price pegged and independent pubdata pricing, in this PR only L1Pegged is ever instantiated 

❗ A new JSON-RPC method has been added: `zks_getMainNodeFeeParams`. It will be used by the EN and it will serve the same purpose as the `zks_getL1GasPrice`, i.e. retrieving the fee info from the main node, but now it will be extended to provide the fair l2 gas price as well as other config params that for now will be defined on the main node. The `zks_getL1GasPrice` will continue working as usual to not interrupt the existing external nodes, but may be removed in the long run.

## Why ❔

Splitting the integration of the 1.4.1 & the new fee model (https://github.com/matter-labs/zksync-era/pull/791/) into separate smaller PRs 

## Checklist


core/lib/zksync_core/src/fee_model.rs
```diff
@@ -0,0 +1,143 @@
...
+
+    BatchFeeInput::L1Pegged(L1PeggedBatchFeeModelInput {
+        l1_gas_price,
+        fair_l2_gas_price: params.config.minimal_l2_gas_price,
+    })
+}
+
+/// Calculates the batch fee input based on the main node parameters.
+/// This function uses the `V2` fee model, i.e. where the pubdata price does not include the proving costs.
+pub(crate) fn compute_batch_fee_model_input_v2(
```
mm-zk, line 93: can you add some unittests for this method in this file?

core/lib/zksync_core/src/fee_model.rs
```diff
@@ -0,0 +1,143 @@
...
+    params: MainNodeFeeParamsV2,
+    l1_gas_price_scale_factor: f64,
+    l1_pubdata_price_scale_factor: f64,
+) -> BatchFeeInput {
+    let MainNodeFeeParamsV2 {
```
mm-zk, line 98: optional: should we check that scale factor is valid (for example > 0,  and < 100 etc?)

core/lib/types/src/fee_model.rs
```diff
@@ -0,0 +1,185 @@
...
+
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+pub struct MainNodeFeeModelConfigV2 {
+    /// The minimal acceptable L2 gas price, i.e. the price that should include the cost of computation/proving as well
+    /// as potentially premium for congestion.
+    pub minimal_l2_gas_price: u64,
+    /// The constant that represents the possibility that a batch can be sealed because of overuse of computation resources.
+    /// It has range from 0 to 1. If it is 0, the compute will not depend on the cost for closing the batch.
+    /// If it is 1, the gas limit per batch will have to cover the entire cost of closing the batch.
+    pub compute_overhead_percent: f64,
```
mm-zk, line 133: btw 'percent' in the name suggests that it is in the range from 0 to 100...

maybe call it 'batch_sealed_due_to_compute_odds' ? 
(and then the other one would be 'batch_sealed_due_to_pubdata_odds' )

core/lib/zksync_core/src/fee_model.rs
```diff
@@ -0,0 +1,143 @@
...
+    let l1_pubdata_price = (l1_pubdata_price as f64 * l1_pubdata_price_scale_factor) as u64;
+
+    // While the final results of the calculations are not expected to have any overflows, the intermediate computations
+    // might, so we use U256 for them.
+    let l1_batch_overhead_wei = U256::from(l1_gas_price) * U256::from(batch_overhead_l1_gas);
+
+    let fair_l2_gas_price = {
+        let l1_batch_overhead_per_gas = l1_batch_overhead_wei / U256::from(max_gas_per_batch);
+        let gas_overhead_wei =
+            (l1_batch_overhead_per_gas.as_u64() as f64 * compute_overhead_percent) as u64;
```
mm-zk, line 123: add more comments about the logic here - that we are using 'compute_overhead_percent' and 'pubdata_overhead_percent' to 'fairly' distribute the cost of closing the batch between transactions.

core/lib/zksync_core/src/fee_model.rs
```diff
@@ -0,0 +1,143 @@
...
+
+        minimal_l2_gas_price + gas_overhead_wei
+    };
+
+    let fair_pubdata_price = {
+        let l1_batch_overhead_per_pubdata =
+            l1_batch_overhead_wei / U256::from(max_pubdata_per_batch);
+        let pubdata_overhead_wei =
+            (l1_batch_overhead_per_pubdata.as_u64() as f64 * pubdata_overhead_percent) as u64;
+        let pubdata_overhead_wei = pubdata_overhead_wei / max_pubdata_per_batch;
```
mm-zk, line 133: why do you divide it again? (you already did it in line 130)

core/lib/types/src/fee_model.rs
```diff
@@ -0,0 +1,185 @@
...
+    /// The constant that represents the possibility that a batch can be sealed because of overuse of pubdata.
+    /// It has range from 0 to 1. If it is 0, the pubdata will not depend on the cost for closing the batch.
+    /// If it is 1, the pubdata limit per batch will have to cover the entire cost of closing the batch.
```
mm-zk, line 136: shouldn't we also check somewhere that compute + pubdata percentages are summing up to 1 ? 
(in theory, they don't have to be 'exact' one, but the sum should be close, right?)

mm-zk (changes requested):

core/lib/types/src/fee_model.rs
```diff
@@ -0,0 +1,185 @@
...
+
+/// The enum which represents the version of the fee model. It is used to determine which fee model should be used for the batch.
+/// - V1, the first model that was used in zkSync Era. In this fee model, the pubdata price must be pegged to the L1 gas price.
+/// Also, the fair L2 gas price is expected to only include the proving/computation price for the operator and not the costs that come from
+/// processing the batch on L1.
+/// - V2, the second model that was used in zkSync Era. There the pubdata price might be independent from the L1 gas price. Also,
+/// The fair L2 gas price is expected to both the proving/computation price for the operator and the costs that come from
+/// processing the batch on L1.
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+pub enum MainNodeFeeModelConfig {
```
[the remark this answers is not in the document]
StanislavBreadless: We can, but (at least right now) it is set on the main node and EN would only request it from the main node

core/lib/web3_decl/src/namespaces/zks.rs
```diff
@@ -101,6 +102,9 @@ pub trait ZksNamespace {
     #[method(name = "getL1GasPrice")]
     async fn get_l1_gas_price(&self) -> RpcResult<U64>;
 
+    #[method(name = "getMainNodeFeeParams")]
+    async fn get_main_node_fee_params(&self) -> RpcResult<MainNodeFeeParams>;
```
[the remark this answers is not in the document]
StanislavBreadless: no, this is just for realtime access from ENs, no need for historical access. At least no need to have it right now

core/lib/zksync_core/src/fee_model.rs
StanislavBreadless: it is mainly an optional param that is used in the API... I would personally not add unneeded constraints to it

core/lib/types/src/fee_model.rs
StanislavBreadless: maybe better to call it `part`, since it basically defines which part of the batch overhead will the compute take?

StanislavBreadless: IMHO `odds` sounds a bit weird in this context (as it is not really a probability, but rather a percentage of the batch overhead to be paid by compute)

core/lib/types/src/fee_model.rs
StanislavBreadless: not necessarily, since these are not really probabilities, but rather just a % of the batch costs to be included for this param

core/lib/multivm/src/versions/vm_latest/utils/fee.rs
```diff
@@ -32,18 +39,46 @@ pub(crate) fn derive_base_fee_and_gas_per_pubdata(
...
+                / U256::from(17);
+
+            fee_input.l1_gas_price = new_l1_gas_price.as_u64();
+        }
+        BatchFeeInput::PubdataIndependent(fee_input) => {
+            // gasPerPubdata = ceil(fair_pubdata_price / fair_l2_gas_price)
+            // gasPerPubdata <= fair_pubdata_price / fair_l2_gas_price + 1
+            // fair_l2_gas_price(gasPerPubdata - 1) <= fair_pubdata_price
+            let new_fair_pubdata_price = U256::from(fee_input.fair_pubdata_price)
+                * (tx_gas_per_pubdata_limit - U256::from(1u32));
```
perekopskiy, line 77: If I understand correctly, then `fee_input.fair_l2_gas_price` should be used instead of `fee_input.fair_pubdata_price`

core/lib/zksync_core/src/fee_model.rs
```diff
@@ -0,0 +1,143 @@
...
+}
+
+/// The struct that represents the batch fee input provider to be used in the main node of the server, i.e.
+/// it explicitly gets the L1 gas price from the provider and uses it to calculate the batch fee input instead of getting
+/// it from other node.
+#[derive(Debug)]
+pub(crate) struct MainNodeFeeInputProvider<G: ?Sized> {
+    provider: Arc<G>,
+    config: MainNodeFeeModelConfig,
+}
```
perekopskiy, line 53: It probably makes sense to use `dyn L1GasPriceProvider` instead of generics like we do in other places, cc @popzxc

StanislavBreadless: makes sense

popzxc: Yes please.

core/lib/web3_decl/src/namespaces/zks.rs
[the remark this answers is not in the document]
Deniallugo: And my question again, is why it's in zks namespace? if it's suitable only for ens?

core/lib/types/src/fee_model.rs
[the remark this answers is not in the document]
Deniallugo: I'd prefer just FeeModelConfig. 
We shouldn't have divergency in names

core/lib/zksync_core/src/api_server/web3/namespaces/zks.rs
```diff
@@ -582,13 +583,31 @@ impl ZksNamespace {
...
+            .batch_fee_input_provider
+            .get_batch_fee_input()
+            .l1_gas_price();
 
         method_latency.observe();
         gas_price.into()
     }
 
+    #[tracing::instrument(skip(self))]
+    pub fn get_main_node_fee_params_impl(&self) -> MainNodeFeeParams {
```
Deniallugo, line 595: why do we have main_node here? 
seems external node also can expose it

Deniallugo (commented):
please address @mm-zk comments

core/lib/types/src/fee_model.rs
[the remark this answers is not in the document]
StanislavBreadless: Okay, I'll do it this way

core/lib/web3_decl/src/namespaces/zks.rs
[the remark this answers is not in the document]
StanislavBreadless: @Deniallugo For the same reason why `getL1GasPrice` is there (so I put it there for consistency/ease of integration). If you think it is better to move it to EN, okay, I'll do

core/lib/zksync_core/src/api_server/web3/namespaces/zks.rs
StanislavBreadless: they can expose it, but it is the main node that truly sets it... Okay, I'll rename it together with the config params

core/lib/web3_decl/src/namespaces/zks.rs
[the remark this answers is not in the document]
Deniallugo: get l1 gas price is much more generic. 
but probably is `main_node` in name is the most confusing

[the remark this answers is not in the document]
StanislavBreadless: I'll remove the `main node` thing. the namespace is of course optional, just keeping it in `zks` seems a bit easier to add

core/lib/multivm/src/versions/vm_latest/utils/fee.rs
StanislavBreadless: good catch

core/lib/types/src/fee_model.rs
mm-zk: we could also say 'probability' instead of odds..

so it would be:
'batch_sealed_due_to_compute_probability'

core/lib/zksync_core/src/fee_model.rs
mm-zk: I'd still check that it is > 0..

if someone sets it to -1 --> the effects will be very strange (and potentially dangerous)

core/lib/zksync_core/src/fee_model.rs
```diff
@@ -0,0 +1,386 @@
...
+
+    /// Returns the fee model parameters.
+    fn get_fee_model_params(&self) -> FeeParams;
+}
+
+/// The struct that represents the batch fee input provider to be used in the main node of the server, i.e.
+/// it explicitly gets the L1 gas price from the provider and uses it to calculate the batch fee input instead of getting
+/// it from other node.
+#[derive(Debug)]
+pub(crate) struct MainNodeFeeInputProvider {
```
mm-zk, line 53: I'd also suggest renaming this from 'MainNodeFeeInputProvider' to 'FeeInputProvider'.

(as we're also creating this object in external node ;-)

mm-zk (dismissed):
LGTM - after these small rename comments above.

perekopskiy (dismissed):

Deniallugo (approved):

core/lib/zksync_core/src/fee_model.rs
StanislavBreadless: @mm-zk I would personally keep it this way. While it is instantiated in the external node, it is only instantiated for the tx sender (i.e. API estimations). The fee input provider used by the state keeper of the EN is different

perekopskiy (approved):

PR #819, opened 2024-01-04
fix(prover): Remove prover-utils from core

EmilLuta:
Last batch of removing old prover from core.
Further improvements will follow up in the core <-> prover integration. These will be tackled as part of prover <-> core separation.

## Checklist


popzxc (approved):
yay.jpg

PR #826, opened 2024-01-06
feat(core): removes multiple tokio runtimes and worker number setting.

montekki:
## Checklist


montekki:
This generally looks to work in CI, but i think that we should add some config variables deprication warning

core/bin/external_node/src/config/mod.rs
```diff
@@ -355,8 +355,6 @@ pub struct RequiredENConfig {
     pub ws_port: u16,
     /// Port on which the healthcheck REST server is listening.
     pub healthcheck_port: u16,
-    /// Number of threads per API server
-    pub threads_per_server: usize,
```
popzxc, line 359: Yup, since we're changing the EN, I think we need to both
1) edit the [prepared configs](https://github.com/matter-labs/zksync-era/tree/main/docs/guides/external-node/prepared_configs), and
2) announce to the EN users that the variable is no longer in use and may be removed.

popzxc (commented):
Nice.

core/bin/external_node/src/config/mod.rs
montekki: Removed

popzxc (approved):

popzxc (approved):

PR #832, opened 2024-01-08
feat(state-keeper): Reject transactions that fail to publish bytecodes

perekopskiy:
## What ❔

State keeper and API reject transactions that fail to publish bytecodes

## Checklist

### DIFF
--- a/core/lib/zksync_core/src/state_keeper/batch_executor/mod.rs
+++ b/core/lib/zksync_core/src/state_keeper/batch_executor/mod.rs
@@ -452,9 +452,9 @@ impl BatchExecutor {
             vec![]
         };
 
-        if let Ok(result) =
-            vm.inspect_transaction_with_bytecode_compression(tracer.into(), tx.clone(), true)
-        {
+        let (published_bytecodes, mut result) =
+            vm.inspect_transaction_with_bytecode_compression(tracer.into(), tx.clone(), true);
+        if published_bytecodes.is_ok() {
             let compressed_bytecodes = vm.get_last_tx_compressed_bytecodes();
 
             let trace = Arc::try_unwrap(call_tracer_result)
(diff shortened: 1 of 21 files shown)


core/lib/zksync_core/src/state_keeper/batch_executor/mod.rs
```diff
@@ -445,58 +445,30 @@ impl BatchExecutor {
...
-        // is limited and the transactions do not pay for taking it.
-        // In order to not let the accounts spam the space of compressed bytecodes with bytecodes
-        // that will not be published (e.g. due to out of gas), we use the following scheme:
-        // We try to execute the transaction with compressed bytecodes.
-        // If it fails and the compressed bytecodes have not been published,
-        // it means that there is no sense in polluting the space of compressed bytecodes,
-        // and so we re-execute the transaction, but without compression.
-
-        // Saving the snapshot before executing
-        vm.make_snapshot();
```
Deniallugo, line 458: It will lead to discrepancy in EN execution

StanislavBreadless (approved):

core/lib/zksync_core/src/state_keeper/batch_executor/mod.rs
StanislavBreadless: @Deniallugo it won't, because if tx has failed to publish compressed bytecode, it won't even make it to the external ndoe

Deniallugo: If we are talking about executing only future txs, then yes. 
but for the past txs it's breaking

perekopskiy: Changed code, so EN behaviour is the same as it was before

StanislavBreadless (approved):

Deniallugo (approved):

PR #838, opened 2024-01-08
feat(api): Make Web3 API server work with pruned data

slowli:
## What ❔

Modifies the Web3 API server so that it works with pruned node data during snapshot recovery.

## Why ❔

## Checklist


core/lib/types/src/api/mod.rs
```diff
@@ -556,19 +556,19 @@ pub struct GetLogsFilter {
 
 /// Result of debugging block
 /// For some reasons geth returns result as {result: DebugCall}
-#[derive(Debug, Serialize, Deserialize, Clone)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
popzxc, line 559: -0 on such changes: they increase the diff while arguably the benefit is marginal/subjective.
When you have to read several 1k+ sized PRs, it adds up and makes reviewing harder than necessary.
Probably it'd be better to ship such changes as separate smallish PRs.

core/lib/zksync_core/src/api_server/execution_sandbox/apply.rs
```diff
@@ -320,55 +322,47 @@ impl BlockArgs {
...
+
+        if self.is_pending_miniblock() {
+            let sealed_l1_batch_number =
+                connection.blocks_dal().get_sealed_l1_batch_number().await?;
+            let sealed_miniblock_header = connection
+                .blocks_dal()
+                .get_last_sealed_miniblock_header()
+                .await?
+                .context("no miniblocks in storage")?;
+            // TODO: no miniblocks can be supported (requires snapshot miniblock timestamp)
```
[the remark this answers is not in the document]
popzxc: Do you mean that you can make it work if there are no miniblocks in the storage right after the snapshot sync? If so, _I think_ it's not worth it. With miniblocks created every second, the chances of anyone actually noticing this error are very slim, and it's not going to have any consequences even if it'll fire.

core/lib/zksync_core/src/api_server/execution_sandbox/mod.rs
```diff
@@ -227,6 +231,13 @@ impl BlockArgs {
     }
 
     /// Loads block information from DB.
+    ///
+    /// # Important!
+    ///
+    /// It is assumed that `block_id` is present in Postgres or can be present in the future
+    /// (i.e., it does not refer to a pruned block). If called for a pruned block, the returned value
+    /// (specifically, `l1_batch_timestamp_s`) will be nonsensical. Hence, the caller must take care
+    /// of pruned blocks separately.
```
popzxc, line 240: Maybe it makes sense to pass `BlockStartInfo` here and do the check internally?
I can imagine that _in theory_ we can forget to do the check (or accidentally remove it) during some quickfix to the API.
It'll be a bit more verbose, but I'd prefer us to return an error in such case rather than serve incorrect information.

core/lib/zksync_core/src/utils/mod.rs
```diff
@@ -87,6 +87,33 @@ pub(crate) async fn projected_first_l1_batch(
     Ok(snapshot_recovery.map_or(L1BatchNumber(0), |recovery| recovery.l1_batch_number + 1))
 }
 
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct BlockStartInfo {
```
popzxc, line 91: This structure seems to be only used in the `api_server` module. Do you anticipate it being useful elsewhere? If not, maybe it'll be better to move it there?
`utils` module feels like a somewhat awkward place to put it.

core/lib/zksync_core/src/api_server/web3/state.rs
```diff
@@ -182,6 +184,58 @@ impl RpcState {
...
+        }
+    }
+
+    pub(crate) fn check_pruned_l1_batch(&self, number: L1BatchNumber) -> Result<(), Web3Error> {
+        let first_local_l1_batch = self.start_info.first_l1_batch;
+        if number < first_local_l1_batch {
+            return Err(Web3Error::PrunedL1Batch(first_local_l1_batch));
+        }
+        Ok(())
+    }
```
popzxc, line 211: Idk if it'll be better, but as a suggestion:
If we will move `BlockStartInfo` to the `api_server`, we can in theory do roughly the following (pseudocode):

```rust
enum PruneQuery {
    Batch(L1BatchNumber),
    ApiBlock(api::BlockId),
}

impl BlockStartInfo {
  fn ensure_not_pruned(&self, query: impl Into<PruneQuery>) -> Result<(), Web3Error> { .. }
}
```

then you'll be able to just call `self.state.block_start_info.ensure_not_pruned(whatever)?`.

core/lib/zksync_core/src/utils/mod.rs
```diff
@@ -87,6 +87,33 @@ pub(crate) async fn projected_first_l1_batch(
...
+    /// Projected number of the first locally available miniblock. This miniblock is **not**
+    /// guaranteed to be present in the storage!
```
popzxc, line 93: Also, stupid question: looks like we have some places where the fact that there may be some state _but_ no miniblocks/batches in the storage after the snapshot sync is an edge case. Would it make sense to include the last L1 batch into the snapshot? This way we'll always be sure that there is at least one miniblock/batch in the storage and we'll be able to use it as an invariant.
The cost of +1 L1 batch sounds negligible in theory, but it can simplify stuff quite a bit.

slowli: We've discussed this with @tomg10 and @RomanBrodetski before and ultimately decided against it. The reasoning for it is that a miniblock / L1 batch header in Postgres *sort of* assumes that there's information related to it (transactions, events, etc.) is also present in Postgres. This would not be the case for the snapshot miniblock / L1 batch, so if we store their headers in Postgres, it would make more difficult to reason about storage invariants.

popzxc: But why won't we ship events/transactions/etc as well? For a single batch, it shouldn't add too much data.
Basically what I had in mind is that we'll _really_ fully sync a single block in addition to the rest of snapshot contents.

slowli: I would be *somewhat* against for a couple of reasons:

- It seems we cannot ensure that the data (events, L2-to-L1 logs etc.) produced by transaction execution in the snapshot miniblock or L1 batch is genuine. To ensure it's genuine, we'd need to actually execute transactions, which would require a VM state snapshot for the previous L1 batch 🙃 As long as we keep snapshot to *just* the VM state (i.e., 3 pieces of info: storage logs, initial writes and factory deps), it's moderately easy to authenticate it using root hash published on L1. AFAIU, the same mechanism could be used to authenticate factory dependencies (since their hashes are a part of the VM state). Authenticity of initial writes seem to be most complicated, but still doable. Or is it possible to verify authenticity of all data associated with a miniblock / L1 batch?
- More banal argument: If we keep snapshot data to minimum, it's easier to maintain. In your proposal, we'd need to ship *all* data associated with a miniblock / L1 batch, which can evolve with time, so old snapshots might cease to work, non-updated nodes might not understand new snapshots etc.

popzxc: Fair enough.

popzxc (approved):

PR #843, opened 2024-01-09
fix(state-keeper): Updates manager keeps track of fictive block metrics

perekopskiy:
## What ❔

Updates manager keeps track of fictive block metrics

## Why ❔

Some metrics (l1_gas, estimated_circuits) are needed in context of the whole batch and were calculated incorrectly previously

## Checklist

### DIFF
--- a/core/lib/zksync_core/src/state_keeper/batch_executor/mod.rs
+++ b/core/lib/zksync_core/src/state_keeper/batch_executor/mod.rs
@@ -257,7 +257,13 @@ impl BatchExecutorHandle {
         latency.observe();
     }
 
-    pub(super) async fn finish_batch(self) -> (FinishedL1Batch, Option<WitnessBlockState>) {
+    pub(super) async fn finish_batch(
+        self,
+    ) -> (
+        FinishedL1Batch,
+        Option<WitnessBlockState>,
+        ExecutionMetricsForCriteria,
+    ) {
         let (response_sender, response_receiver) = oneshot::channel();
         self.commands
             .send(Command::FinishBatch(response_sender))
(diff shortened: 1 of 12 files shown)


core/lib/zksync_core/src/state_keeper/batch_executor/mod.rs
```diff
@@ -330,7 +342,13 @@ impl BatchExecutor {
                     } else {
                         None
                     };
-                    resp.send((vm_block_result, witness_block_state)).unwrap();
+                    let block_tip_metrics = Self::get_execution_metrics(
+                        None,
+                        &vm_block_result.block_tip_execution_result,
+                    );
```
slowli, line 348: Dumb question: Can't the equivalent of this method be called on `vm_block_result` on the caller side (i.e., in `MiniblockUpdates::extend_from_fictive_transaction()`, so that it's not necessary to pass an additional value everywhere? E.g., rework this method as a crate-public function in the `gas_tracker` module, or as a `ExecutionMetricsForCriteria` constuctor. Or do we plan to change `block_tip_metrics` definition in the future, so that it won't be derivable from `vm_block_result` alone?

StanislavBreadless (approved):

core/lib/zksync_core/src/state_keeper/batch_executor/mod.rs
perekopskiy: Definitely makes sense to not pass it everywhere and can be derived at the end. I was trying to make it the same way as we do for non-fictive transactions. But as far as I see the reason why it's so for non-fictive ones is that deriving metrics for them requires `Transaction` object and we don't want to clone it before sending to batch executor

core/lib/zksync_core/src/state_keeper/updates/miniblock_updates.rs
```diff
@@ -59,13 +59,21 @@ impl MiniblockUpdates {
         }
     }
 
-    pub(crate) fn extend_from_fictive_transaction(&mut self, result: VmExecutionResultAndLogs) {
+    pub(crate) fn extend_from_fictive_transaction(
+        &mut self,
+        result: VmExecutionResultAndLogs,
+        l1_gas_count: BlockGasCount,
+        execution_metrics: ExecutionMetrics,
```
slowli, line 66: Bikshedding: Couldn't you instantiate `ExecutionMetricsForCriteria::new(None, &result)` directly inside this method to compact the things even further? Specifically, it would signify that `MiniblockUpdates` metrics are updated based on `result` only.

slowli (approved):

core/lib/zksync_core/src/state_keeper/updates/miniblock_updates.rs
perekopskiy: I'd like to keep this way to be consistent between `extend_from_fictive_transaction` and `extend_from_executed_transaction`. Maybe this whole part worth a refactoring, but it shouldn't be a part of this PR

PR #853, opened 2024-01-10
ci: migrate to builded era-contracts

otani88:
## What ❔
Migrate to pre-builded contracts

## Why ❔
Optimize CI execution

## Checklist


.github/workflows/build-docker-from-tag.yml
```diff
@@ -49,21 +49,17 @@ jobs:
...
-    needs: [setup, build-contracts]
+    needs: [setup]
     uses: ./.github/workflows/build-core-template.yml
     if: contains(github.ref_name, 'core')
     secrets:
       DOCKERHUB_USER: ${{ secrets.DOCKERHUB_USER }}
       DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
     with:
       image_tag_suffix: ${{ needs.setup.outputs.image_tag_suffix }}
+      era_contract_version: "main-6250292"
```
Deniallugo, line 62: Oh, it's a bad idea. I think we still want to have contracts as a submodule. Because we can't blindly use them and protocol will be ultra disappointed.

I'm ok if you keep this version and can calculate it from the submodule. 

Otherwise, you will see a lot of angry developers...because it works locally but doesn't in ci

PR #857, opened 2024-01-11
chore(docs): Adding a README with a list of VM versions

mm-zk:
## What ❔

* Adding a list with the current VM versions

## Why ❔

* To make it easy to see what is the chronological order (and what was changed).


popzxc (approved):
❤️

PR #873, opened 2024-01-15
 I fixed a simple word mistake

bolatfurkan:
I would be glad if I could make a small contribution to this project.


dutterbutter:
@bolatfurkan thanks for contribution but this has already been resolved.

PR #880, opened 2024-01-16
feat: ZK Stack framework MVP

popzxc:
⚠️ Please read the description before reviewing. ⚠️ 

This is the ZK Stack framework MVP that represents the core functionality of the new framework.
A lot of things are not polished yet: it'll need better logging/reporting, interfaces can be better, naming is off at some places (and I haven't decided on the usage of ZK Stack / zkSync there), we need more tests and examples, no README.md, etc.

Additionally, only a few resources and tasks are implemented right now. The reasons for that:
- Adapting the existing code to get a single interface takes time, as most of the tasks/resources were created in an ad-hoc manner.
- I wanted to focus on the framework functionality here, the implementations may come later as separate PRs.
- I didn't want the diff to be big.

In the current form, the framework contains a showcase of how interacting with it looks like, and generally, it seems to work fine.
For more details on motivation and goals, please read the spec.

With that in mind, *any* feedback is appreciated, but I don't guarantee that I'll proceed it in this PR. Here I'll focus on making sure that the framework works and there is no major disagreement w.r.t. its design, and then there will be N follow-up PRs bringing the missing functionality and polishing the code.


slowli (commented):
Not proposing to recreate [Spring](https://spring.io/) 🙃, but the resource approach looks remarkably similar to DI containers. So I think it could be a reasonable idea to have a node be defined as a DI container that supports a minimum viable set of functionality – multiple ~beans~ resources of the same type (maybe, with string qualifiers?); that seems to allow to model health checks (and maybe tasks themselves) as parts of the DI container. We can probably get away with not having much else that makes DI containers ~a living nightmare~ hard to reason about for large apps (ordering, conditionals, lazy resources etc.); we can probably rely on the user-provided ordering of updating the DI container to not bother with topologically sorting dependencies, detecting circular deps etc.).

core/lib/node/src/task/prometheus_exporter.rs
```diff
@@ -0,0 +1,48 @@
...
+
+#[async_trait::async_trait]
+impl ZkSyncTask for PrometheusExporterTask {
+    fn healthcheck(&mut self) -> Option<Box<dyn CheckHealth>> {
+        self.prometheus_health_check
+            .take()
+            .map(|c| Box::new(c) as Box<dyn CheckHealth>)
+    }
+
+    async fn run(self: Box<Self>, stop_receiver: StopReceiver) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
popzxc: Yep, actually [it was a resource](https://github.com/matter-labs/zksync-era/blob/a283a46c642ad4893ca93bdfea5578bd4d1a9291/core/lib/node/src/resource/stop_receiver.rs), but I decided against it, since I thought that it'd be really easy to forget to request it and create a task that hangs. It also was clunky, since this resource was the only one that [is implicitly added](https://github.com/matter-labs/zksync-era/blob/a283a46c642ad4893ca93bdfea5578bd4d1a9291/core/lib/node/src/node/mod.rs#L135) and I wanted to avoid it.

core/lib/node/src/node/mod.rs
```diff
@@ -0,0 +1,269 @@
...
+        Ok(self_)
+    }
+
+    /// Adds a task to the node.
+    ///
+    /// The task is not created at this point, instead, the constructor is stored in the node
+    /// and will be invoked during [`ZkSyncNode::run`] method. Any error returned by the constructor
+    /// will prevent the node from starting and will be propagated by the [`ZkSyncNode::run`] method.
+    pub fn add_task<
+        F: FnOnce(&NodeContext<'_>) -> Result<Box<dyn ZkSyncTask>, TaskInitError> + 'static,
```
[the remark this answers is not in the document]
popzxc: Yep, I still have the same question and haven't decided on it yet.
More feedback here is welcome.

core/lib/node/src/node/mod.rs
```diff
@@ -0,0 +1,269 @@
...
+            .map(|task| {
+                let task = task.task.take().expect(
+                    "Tasks are created by the node and must be Some prior to calling this method",
+                );
+                rt_handle.spawn(task).fuse()
+            })
+            .collect();
+
+        // Run the tasks until one of them exits.
+        // TODO (QIT-24): wrap every task into a timeout to prevent hanging.
```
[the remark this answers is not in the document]
popzxc: Yes, I want to work on this soon-ish, just thought that given just a few components, it's not a blocker to get this PR merged.

core/lib/node/src/node/mod.rs
```diff
@@ -0,0 +1,269 @@
...
+    /// - Its own healthcheck is ignored (the task is expected to handle its own checks itself).
+    /// - There may be only one healthcheck task per node.
+    ///
+    /// If the healthcheck task is already set, the provided one will override it.
+    pub fn with_healthcheck<
+        F: FnOnce(
+                &NodeContext<'_>,
+                Vec<Box<dyn CheckHealth>>,
+            ) -> Result<Box<dyn ZkSyncTask>, TaskInitError>
+            + 'static,
```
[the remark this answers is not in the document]
popzxc: Noted on the health checks -- and I think I kind of like this idea, this also allows one to implement other similar concepts, like circuit breakers (probably we want components to define their circuit breakers rather than have to initialize them as a separate entity). What I don't know right now is a good way to do it without having to rely on the invocation order (e.g. even if the healtcheck task is added first, it'll still observe all the healthchecks).

On the sync/async -- the tokio runtime handle is provided through the node context, but I'm still not sure if a freely available async context is a good thing.

core/lib/node/examples/main_node.rs
```diff
@@ -0,0 +1,86 @@
...
+    healthcheck::IntoHealthCheckTask,
+    node::ZkSyncNode,
+    resource::{pools::MasterPoolResource, Resource, ResourceProvider},
+    task::{
+        healtcheck_server::HealthCheckTask, metadata_calculator::MetadataCalculatorTask,
+        IntoZkSyncTask,
+    },
+};
+
+/// Resource provider for the main node.
```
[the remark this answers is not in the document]
popzxc: I think the answer to this question is also related to your main review message.

I think that we aren't building a _general-purpose_ framework like Spring. The use cases here will be somewhat limited, main being:
- Use different config formats (env/toml/yaml).
- Run different sets of tasks (main motivation is to run different components on different machines, like we do now; and probably either run or nor run setup-specific tasks (BWIP for main node, reorg detector for EN)).
- Customize tasks -- and that's where resources come in handy. When we are composing the node, we know exactly what behavior we want.

So it seems that it's OK for us to know all the possible resource definitions beforehand. Obviously, we can have some node-specific resources that are provided implicitly, but generally I don't find it too awful for the end crate to explicitly know what resources it provides. Especially that pretty often these resources require some kind of config that you have to provide anyways.

We lose some flexibility this way, but also what we obtain is that we don't have to deal BOTH with topological sorts and having to rely on the call ordering.

> An important requirement in this case would be to be able abstract the resource type from the implementation (i.e., have it defined something like Arc<dyn MyTrait>).

It is possible right now (see e.g. object store resource implementation), but it is resource provider's responsibility to choose which implementation it will give.

core/lib/node/src/resource/mod.rs
```diff
@@ -0,0 +1,35 @@
...
+/// A marker trait for anything that can be stored (and retrieved) as a resource.
+/// Requires `Clone` since the same resource may be requested by several tasks.
+pub trait Resource: 'static + Clone + std::any::Any {
+    /// Unique identifier of the resource.
+    /// Used to fetch the resource from the provider.
+    ///
+    /// It is recommended to name resources in form of `<scope>/<name>`, where `<scope>` is the name of the task
+    /// that will use this resource, or 'common' in case it is used by several tasks, and `<name>` is the name
+    /// of the resource itself.
+    const RESOURCE_NAME: &'static str;
```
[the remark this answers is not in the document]
popzxc: I thought that it will make sense to provide some kind of naming scheme that will also consider multiple implementations of the resource. I'll share a separate spec on that soon-ish, but the rough idea is:
Resource name is `common/object_store`, but what resource provider will work with is `common/object_store/gcp` or `common/object_store/in_memory` (or alternatively `state_keeper/io` -> `state_keeper/io/mempool_io` / `state_keeper/io/external_io`).

This will come in handy when the node configuration is defined by some kind of spec file (which is one of the main use cases for the framework).

core/lib/node/src/task/prometheus_exporter.rs
```diff
@@ -0,0 +1,48 @@
...
+    }
+}
+
+#[async_trait::async_trait]
+impl ZkSyncTask for PrometheusExporterTask {
+    fn healthcheck(&mut self) -> Option<Box<dyn CheckHealth>> {
+        self.prometheus_health_check
+            .take()
+            .map(|c| Box::new(c) as Box<dyn CheckHealth>)
+    }
```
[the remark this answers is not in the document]
popzxc: > Do we have a guarantee that healthcheck will be called no more than once for a task? 

[Yep](https://github.com/matter-labs/zksync-era/blob/776ec26e342993d419ce02ca80b743c82c01f98e/core/lib/node/src/task/mod.rs#L36).

> IMO, defining health check as a part of task initialization (e.g., by providing it into the DI container; see the other comment) could be more straightforward.

Similarly to the comment above, I agree that it'd be nicer, but I would love to find a way where we don't implement anything that is too complex/too generic (since the ROI is not clear), or has user-facing flaws like having to depend on the invocation order (IMHO it's worse). One could argue that [polling a non-fused future ](https://doc.rust-lang.org/std/future/trait.Future.html#panics) also has awkward guarantees, so generally having library-level contracts is OK-ish.

core/lib/node/src/node/mod.rs
```diff
@@ -0,0 +1,269 @@
...
+///
+/// Initialization flow:
+/// - Node instance is created with access to the resource provider.
+/// - Task constructors are added to the node. At this step, tasks are not created yet.
+/// - Optionally, a healthcheck task constructor is also added.
+/// - Once the `run` method is invoked, node
+///   - attempts to create every task. If there are no tasks, or at least task
+///     constructor fails, the node will return an error.
+///   - initializes the healthcheck task if it's provided.
+///   - waits for any of the tasks to finish.
```
pompon0, line 36: what about one-off tasks? We don't need support for those?

popzxc: IIRC whenever we use one-off tasks, we still want the node to exit upon completion, so it's kind of covered (e.g. task runs, returns `Ok(())`, and the node exits).
I don't think I'm aware of a use-case where we have a task that can complete but shouldn't cause the node to exit, do you have an example?

core/lib/node/src/node/mod.rs
```diff
@@ -0,0 +1,269 @@
...
+    /// Cache of resources that have been requested at least by one task.
+    resources: RefCell<HashMap<String, Box<dyn Any>>>,
+    /// List of task constructors.
+    task_constructors: Vec<(String, TaskConstructor)>,
+    /// Optional constructor for the healthcheck task.
+    healthcheck_task_constructor: Option<HealthCheckTaskConstructor>,
+
+    /// Sender used to stop the tasks.
+    stop_sender: watch::Sender<bool>,
+    /// Stop receiver to be shared with tasks.
```
pompon0, line 53: you can always subscribe to watch::Sender, there is no need to keep the receiver around.

core/lib/node/src/node/mod.rs
```diff
@@ -0,0 +1,269 @@
...
+        if tasks.is_empty() {
+            anyhow::bail!("No tasks to run");
+        }
+
+        // Initialize the healthcheck task, if any.
+        if let Some(healthcheck_constructor) = self.healthcheck_task_constructor.take() {
+            let healthcheck_task = healthcheck_constructor(&NodeContext::new(&self), healthchecks)
+                .map_err(|err| {
+                    tracing::error!("Healthcheck task can't be initialized: {err}");
+                    anyhow::format_err!("Healtcheck task can't be initialized: {err}")
```
pompon0, line 189: you can use TaskInitError implements Error, so you can just use `context("...")` instead (outside of map_err)

pompon0, line 189: logging AND returning an error is a bad practice (the error most likely will be logged outside, which will cause log duplication).

popzxc: Yup, I'll improve the error approach in the follow-up PRs.
Right now I did it this way, since we log all the errors for failed constructors, but here it's slightly different.

core/lib/node/src/node/context.rs
```diff
@@ -0,0 +1,69 @@
+use std::any::Any;
+
+use crate::{node::ZkSyncNode, resource::Resource};
+
+/// An interface to the node's resources provided to the tasks during initialization.
+/// Provides the ability to fetch required resources, and also gives access to the Tokio runtime used by the node.
+#[derive(Debug)]
+pub struct NodeContext<'a> {
```
pompon0, line 8: If we made the resource provider into a trait (`trait ResourceProvider<T>`) with a method like `provide(&self) -> T` and parametrized NodeContext and ZkSyncNode by ResourceProvider, then the tasks could accept NodeContext<P> where P : ResourceProvider<whatever the task needs>, and this way resource availability will be resolved at compilation time (although resource caching will still be implemented using Any wrappers.

core/lib/node/src/task/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+    fn create(
+        node: &NodeContext<'_>,
+        config: Self::Config,
+    ) -> Result<Box<dyn ZkSyncTask>, TaskInitError>;
+}
+
+/// A task implementation.
+#[async_trait::async_trait]
+pub trait ZkSyncTask: 'static + Send + Sync {
+    /// Gets the healthcheck for the task, if it exists.
```
pompon0, line 35: that looks like sth that can be moved to the IntyZkSyncTask::create() result.

pompon0: in fact, given that all these methods are supposed to be called at most once, this could be a struct rather than a trait. Like `struct Task { healthcheck: Option<...>, task: BoxFuture<...>, after_node_shutdown: BoxFuture<...> }`, assuming that we could move StopReceiver to NodeContext.

core/lib/node/src/resource/mod.rs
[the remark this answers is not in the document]
pompon0: you can always keep a single resource type which is a wrapper for a dynamic implementation: `Box<dyn MyResourceTrait>` can also be a resource.

core/lib/node/src/resource/mod.rs
```diff
@@ -0,0 +1,35 @@
...
+}
+
+/// An entity that knows how to initialize resources.
+///
+/// It exists to simplify the initialization process, as both tasks and *resources* can depend on other resources,
+/// and by having an entity that can initialize the resource on demand we can avoid the need to provide resources
```
pompon0, line 21: but resources cannot depend on tasks, so unless we want/need laziness, we could as well construct all the resources before initializing ZkSyncNode, therefore moving the dependency resolution to compilation time (i.e. profit).

core/lib/node/src/resource/mod.rs
```diff
@@ -0,0 +1,35 @@
...
+/// all the resource consumers will interact with the same resource instance, which may be important for getting
+/// the consistent state (e.g. to make sure that L1 gas price is the same for all the tasks).
+pub trait ResourceProvider: 'static + Send + Sync + fmt::Debug {
+    /// Returns a resource with the given name.
+    ///
+    /// In case it isn't possible to obtain the resource (for example, if some error occurred during initialization),
+    /// the provider is free to either return `None` (if it assumes that the node can continue without this resource),
+    /// or to panic.
+    // Note: we have to use `Box<dyn Any>` here, since we can't use `Box<dyn Resource>` due to it not being object-safe.
+    fn get_resource(&self, name: &str) -> Option<Box<dyn Any>>;
```
pompon0, line 34: we don't want this to be async?

core/lib/node/src/task/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+    /// a shutdown. Every task is expected to either await or periodically check the state of channel and stop
+    /// its execution once the channel is changed.
+    ///
+    /// Each task is expected to perform the required cleanup after receiving the stop signal (e.g. make sure that
+    /// spawned sub-tasks are awaited).
+    async fn run(self: Box<Self>, stop_receiver: StopReceiver) -> anyhow::Result<()>;
+
+    /// Asynchronous hook that will be called after *each task* has finished their cleanup.
+    /// It is guaranteed that no other task is running at this point, e.g. `ZkSyncNode` will invoke
+    /// this hook sequentially for each task.
```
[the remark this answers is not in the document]
popzxc: This hook is for stuff like automatic rollbacks if reorg detector needs them. We can't do that at the end of `run`, since some tasks may still be running when we'll start to rollback the database, and we can't really do that after `ZkSyncNode::run()`, since 1) `ZkSyncNode` owns the runtime that was used to initialize async stuff, and passing things through runtimes tends to cause problems, and 2) in the current form the task defines whether it needs to do something (e.g. it's a part of task logic), and if it's done outside of `ZkSyncNode`, it becomes somewhat arbitraty (e.g. `if task_was_resent { do_task_related_logic() } `).

core/lib/node/src/resource/mod.rs
popzxc: The problem here is that, for example, we can run the server with or without `eth_sender`, and depending on that, the operator's private key (and other configs in general) may not be available. So we kind of need laziness.

core/lib/node/src/task/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+pub mod prometheus_exporter;
+
+/// Factory that can create a task.
+// Note: This have to be a separate trait, since `ZkSyncTask` has to be object-safe.
+pub trait IntoZkSyncTask: 'static + Send + Sync {
+    /// Config type for the task.
+    /// It is highly recommended for tasks to have dedicated configs, not shared with other tasks and resources.
+    type Config: 'static + Send + Sync;
+
+    /// Creates a new task.
```
pompon0, line 23: document that this may block.

core/lib/node/src/node/context.rs
popzxc: I think I've tried something like this, but met a lot of errors related to object safety (which you get when you replace generics with `dyn Trait`). I'll check once again though.

core/lib/node/src/task/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+    }
+}
+
+/// An error that can occur during the task initialization.
+#[derive(thiserror::Error, Debug)]
+#[non_exhaustive]
+pub enum TaskInitError {
+    #[error("Resource {0} is not provided")]
+    ResourceLacking(&'static str),
+    #[error("Internal error: {0}")]
```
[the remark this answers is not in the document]
pompon0: or just `error(transparent)`

core/lib/node/src/resource/mod.rs
popzxc: That's a good question. Do we have reasons for this to be async, in your opinion (genuine question)?
Unless you have objections, I think we'll know the answer once I'll migrate most of resources to this model -- if it'll turn out that it'll provide better DevEx to have this method async, I'll change it.

core/lib/node/src/resource/mod.rs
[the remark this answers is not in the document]
popzxc: Exactly, that's how [object store resource ](https://github.com/matter-labs/zksync-era/blob/776ec26e342993d419ce02ca80b743c82c01f98e/core/lib/node/src/resource/object_store.rs) is implemented

core/lib/node/src/node/mod.rs
pompon0: we have this for some reason: https://github.com/matter-labs/zksync-era/blob/3a07890dacebf6179636c44d7cce1afd21ab49eb/core/bin/zksync_server/src/main.rs#L162

popzxc: Yep, but this was a kludge to implement one-off tasks.
_By default_ no actor is expected to exit, and if something exited, then depending on this boolean we will either exit with zero or non-zero code. I can't say I like this approach, so now tasks are allowed (but not required) to exit, and if task returns `Ok(())` this would be considered equivalent to `tasks_allowed_to_finish = true`.

core/lib/node/src/resource/mod.rs
pompon0: ok, I assumed that the feasible execution modes will be hardcoded (like external node/main node). Now I understand that you want the set of tasks to be defined in an external config, instead of writing a short rust program which composes the required components. Is that correct?

core/lib/node/src/resource/mod.rs
pompon0: It depends on whether we need the resources to be lazily computed on demand. If not, perhaps a resource provider should just be a wrapper of "Vec<Box<dyn Any>>" instead of a trait.

core/lib/node/src/resource/mod.rs
popzxc: Yup!

>instead of writing a short rust program which composes the required components. Is that correct?

Yes, though it's just a hypothesis we want to test, not a commitment for X years of LTS. We will ship it (as it's relatively cheap to try) and see how it's being used -- if people would use framework more than external configs, we can always change things to favor that.

core/lib/node/src/node/mod.rs
popzxc: I was able to get rid of these lines completely.

core/lib/node/src/node/context.rs
popzxc: Resource provider traits indeed cannot be generic due to object safety. I've played a bit with attempts to make it more typesafe but failed.
I think I'll keep it as is for now, but we certainly can improve it along the way.

core/lib/node/src/resource/mod.rs
popzxc: The resources are already lazily computed (only when this method is called), though in a blocking manner.
I guess if later we will make `IntoZkSyncTask::create` async (which is not unlikely), it'll certainly make sense to make this method async as well. For now, I don't think it'll severely limit the possibilities for resources, but again -- we'll see along the way.

core/lib/node/src/task/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+
+use crate::node::{NodeContext, StopReceiver};
+
+/// Factory that can create a task.
+// Note: This have to be a separate trait, since `ZkSyncTask` has to be object-safe.
+pub trait IntoZkSyncTask: 'static + Send + Sync {
+    /// Unique name of the task.
+    const NAME: &'static str;
+
+    /// Config type for the task.
```
pompon0, line 17: that would be trivial to enforce if instead of Config and IntoZkSyncTask, you would have ConfigBuilder trait with a method `fn create(self, node NodeContext) -> Result<Box<ZksyncTask> TaskInitError>`.

core/lib/node/src/node/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+///     constructor fails, the node will return an error.
+///   - initializes the healthcheck task if it's provided.
+///   - waits for any of the tasks to finish.
+///   - sends stop signal to all the tasks.
+///   - waits for the remaining tasks to finish.
+///   - calls `after_node_shutdown` hook for every task that has provided it.
+///   - returns the result of the task that has finished.
+pub struct ZkSyncNode {
+    /// Primary source of resources for tasks.
+    resource_provider: Box<dyn ResourceProvider>,
```
pompon0, line 36: IMO given that ResourceProvider is not the only source of resources now, why do we need it in the first place? Wouldn't it be more consistent to just have an add_resource method on ZkSyncNode to prepopulate resources if needed and otherwise just wait for tasks to provide them?

popzxc: I guess the main thing here is that ResourceProvider is lazy, it'll only provide resources if required.
Determining whether we need an `ObjectStore`, for example, would be hard -- we'll need to go through the list of tasks, somehow check whether any of them needs an `ObjectStore`, and then add it.

core/lib/node/src/task/mod.rs
popzxc: So, I've been thinking about it since yesterday, and am still torn. While probably it'll indeed look somewhat nicer (and also this trait would become object-safe, which would make it easier to make it async later), this would assume that each task _requires_ a config, but we might have tasks where `Config = ()` (e.g. if it some kind of an updater or logger or something that just interacts with resources provided by other tasks), and here it would result in unnecessary types.

In your opinion, is it a significant drawback?

popzxc: Also, if we'll go this route, we don't even have to mention configs anywhere, this may just be `TaskBuilder` trait -- so it's one less framework concept to think about.

popzxc: Although probably configless tasks wouldn't be a majority, so I guess it's OK. I'll implement it and will also check whether it will unblock async constructors.

pompon0: Types are cheap, and it is ok to have many types. From my experience, minimize-number-of-types mentality is what prevents people from writing type-safe code. Hence, I don't consider it a drawback really.

core/lib/node/src/node/mod.rs
pompon0: if tasks are supposed to be able to add resources as well, then one will have to do this reasoning anyway (since tasks are not lazy), no?

core/lib/node/src/resource/lazy_resource.rs
```diff
@@ -0,0 +1,84 @@
...
+
+/// A lazy resource represent a resource that isn't available at the time when the tasks start.
+/// Normally it's used to represent the resources that should be provided by one task to another one.
+/// Lazy resources are aware of the node lifecycle, so attempt to resolve the resource won't hang
+/// if the resource is never provided: the resolve future will fail once the stop signal is sent by the node.
+#[derive(Debug)]
+pub struct LazyResource<T: Resource> {
+    resource: Arc<Mutex<Option<T>>>,
+    resolve_receiver: watch::Receiver<bool>,
+    resolve_sender: Arc<Mutex<watch::Sender<bool>>>,
```
pompon0, line 17: why Mutex on watch? Watch access is atomic anyway (I think it has a mutex internally).

core/lib/node/src/node/mod.rs
popzxc: That's true, but then we'll have to ship config for dependency resources together with all the tasks.
OTOH this can be solved with "ConfigProvider", hm.

popzxc: Although not really -- we will have to make every task aware of every possible `ObjectStore` implementation, which is one more thing we want to avoid.

I think I'll stick with `ResourceProvider` for now, but maybe we'll brainstorm how to improve the system once the codebase is adapted to support the new framework.

core/lib/node/src/resource/resource_collection.rs
```diff
@@ -0,0 +1,106 @@
...
+            .await
+            .map_err(|_| ResourceCollectionError::AlreadyResolved(std::any::type_name::<T>()))?;
+
+        let mut handle = self.resources.write().unwrap();
+        if self.resolved.load(std::sync::atomic::Ordering::Relaxed) {
+            return Err(ResourceCollectionError::AlreadyResolved(
+                std::any::type_name::<T>(),
+            ));
+        }
+        let resources = std::mem::take(&mut *handle);
```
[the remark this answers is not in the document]
popzxc: This required a slight refactoring of the healthcheck server, which I initially wanted to avoid, but generally, I agree that it'd be more consistent/predictable. Changed.

core/lib/node/src/task/mod.rs
popzxc: Done, plus made everything async.

core/lib/node/src/task/metadata_calculator.rs
```diff
@@ -0,0 +1,69 @@
...
+
+        if object_store.is_none() {
+            tracing::info!(
+                "Object store is not provided, metadata calculator will run without it."
+            );
+        }
+
+        let metadata_calculator = node
+            .runtime_handle()
+            .block_on(MetadataCalculator::new(config, object_store.map(|os| os.0)));
```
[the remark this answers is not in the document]
popzxc: Everything is async now.

core/lib/node/src/node/mod.rs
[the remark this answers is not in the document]
popzxc: Everything is async now.

core/lib/node/src/resource/mod.rs
popzxc: Async now.

pompon0 (approved):
all my comments are optional BTW

core/lib/node/src/node/mod.rs
popzxc: Well, task-wise, yes, one is expected to provide a correct set of tasks.
I see your point, but right now it feels like creating a correct set of tasks is simpler than creating a correct set of resources (based on the current state of [`initialize_components`](https://github.com/matter-labs/zksync-era/blob/main/core/lib/zksync_core/src/lib.rs#L272)).

If it'll turn out to be untrue, it should be pretty straightforward to remove `ResourceProvider` completely later.

core/lib/node/src/resource/resource_collection.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// Flag indicating that the collection has been resolved.
+    wired: watch::Receiver<bool>,
+}
+
+impl<T: Resource> Resource for ResourceCollection<T> {
+    fn resource_id() -> ResourceId {
+        ResourceId::new("collection") + T::resource_id()
+    }
+
+    fn on_resoure_wired(&mut self) {
```
slowli, line 38: Typo: "resoure" -> resource.

core/lib/node/src/node/mod.rs
slowli: Regarding resource lazy instantiation, I believe it could be dealt with by lazy resources. IMO, it'd be relatively maintainable to define a "task" / set of "tasks" (`ResourceConfiguration` as per other line comment) that would fulfill the `ResourceProvider` role (i.e., provide resources based on a configuration). And we could still retain "statically known" (node config-based) set of resources. This would also allow making `StoredResource` crate-private (which, IMO, it should be).

core/lib/node/examples/main_node.rs
```diff
@@ -0,0 +1,85 @@
...
+    }
+}
+
+#[async_trait::async_trait]
+impl ResourceProvider for MainNodeResourceProvider {
+    async fn get_resource(&self, name: &ResourceId) -> Option<Box<dyn StoredResource>> {
+        match name {
+            name if name == &MasterPoolResource::resource_id() => {
+                let resource =
+                    Self::master_pool_resource().expect("Failed to create pools resource");
```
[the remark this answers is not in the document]
popzxc: Probably we should, but I'll address it later.

core/lib/dal/src/connection/mod.rs
```diff
@@ -65,6 +66,15 @@ impl ConnectionPoolBuilder {
             max_size: self.max_size,
         })
     }
+
+    /// Builds a connection pool that has a single connection.
+    pub async fn build_singleton(&self) -> anyhow::Result<ConnectionPool> {
```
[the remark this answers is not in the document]
popzxc: Probably it'll be easier to remove it once more components are migrated to the new framework.
I've added it to my to-check-later list though.

core/lib/node/src/task/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+use futures::future::BoxFuture;
+
+use crate::{
+    node::{NodeContext, StopReceiver},
+    resource::ResourceId,
+};
+
+/// Factory that can create a task.
+#[async_trait::async_trait]
+pub trait IntoZkSyncTask: 'static + Send + Sync {
```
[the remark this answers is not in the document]
popzxc: Probably will be implemented later.

core/lib/node/src/node/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+/// - Once the `run` method is invoked, node
+///   - attempts to create every task. If there are no tasks, or at least task
+///     constructor fails, the node will return an error.
+///   - initializes the healthcheck task if it's provided.
+///   - waits for any of the tasks to finish.
+///   - sends stop signal to all the tasks.
+///   - waits for the remaining tasks to finish.
+///   - calls `after_node_shutdown` hook for every task that has provided it.
+///   - returns the result of the task that has finished.
+pub struct ZkSyncNode {
```
RomanBrodetski, line 31: Do I understand it right that for multi-machine/multi-process deployment we'd be running multiple instances of `ZkSyncNode` that together consitute one logical zkSync Node? I don't see a way around it, but naming makes it a little confusing. Did you plan at all to represent such multi-machine setups within the framework design?

core/lib/node/src/node/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+
+/// "Manager" class of the node. Collects all the resources and tasks,
+/// then runs tasks until completion.
+///
+/// Initialization flow:
+/// - Node instance is created with access to the resource provider.
+/// - Task constructors are added to the node. At this step, tasks are not created yet.
+/// - Optionally, a healthcheck task constructor is also added.
+/// - Once the `run` method is invoked, node
+///   - attempts to create every task. If there are no tasks, or at least task
```
RomanBrodetski, line 23:
```suggestion
///   - attempts to create every task. If there are no tasks, or at least one task
```

RomanBrodetski (commented):
Had a look - very interesting!

Is there somewhere a list of potential resources? The main node example is helpful, would it only have `master_pool_resource`? I want to understand the motivation behind have a resource abstraction better. We have a nice invariant that components don't communicate with each other directly, I was hoping this would allow us to treat them independently on the code level as well. This would also make the fact that they can be run in different processes more transparent. 

Probably some examples that highlight usefullness of this abstractions would help

core/lib/node/src/node/mod.rs
popzxc: Yes and yes. Naming is subject to changes, of course, but a modular (or microservice, if you must) approach is certainly considered here.

core/lib/node/src/node/mod.rs
```diff
@@ -0,0 +1,207 @@
...
+    }
+}
+
+impl ZkSyncNode {
+    pub fn new<R: ResourceProvider>(resource_provider: R) -> anyhow::Result<Self> {
+        if tokio::runtime::Handle::try_current().is_ok() {
+            anyhow::bail!(
+                "Detected a Tokio Runtime. ZkSyncNode manages its own runtime and does not support nested runtimes"
+            );
+        }
```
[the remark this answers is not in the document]
popzxc: This became one of my personal places where I don't really like Rust -- the implicit state is everywhere, so we have to opt-in for such checks.

core/lib/node/src/resource/lazy_resource.rs
```diff
@@ -0,0 +1,92 @@
...
+
+use thiserror::Error;
+use tokio::sync::watch;
+
+use super::{Resource, ResourceId};
+use crate::node::StopReceiver;
+
+/// A lazy resource represents a resource that isn't available at the time when the tasks start.
+///
+/// Normally it's used to represent the resources that should be provided by one task to another one.
```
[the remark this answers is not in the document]
popzxc: Yep, later there will be doctests showcasing features, right now I didn't want to spend too much on it, given that stuff is likely to change along the way.

popzxc:
@RomanBrodetski 

> Is there somewhere a list of potential resources?

I think a few obvious examples are `ObjectStore` and `StateKeeperIO` (where we can choose different implementation -- that would be up to `ResourceProvider`). A less obvious example is a tree reader: probably it would be a lazy resource that is provided by the `metadata_calculator` and tree API will wait until it's available (actual logic will be more complex, but I think it gives an idea).

> The main node example is helpful, would it only have master_pool_resource?

No, it will evolve as more components are migrated to support the framework. I wanted to reduce diff size, so for now the example is incomplete.

> We have a nice invariant that components don't communicate with each other directly, I was hoping this would allow us to treat them independently on the code level as well.

This is mostly to alter the behavior of components via configuration on the top level, e.g. switch IO implementation in the state keeper. Also resource abstraction will allow to create 3rd party crates that aren't part of our codebase -- e.g. someone will be able to write AWS S3 object store and inject it as an `ObjectStoreResource`.

> Probably some examples that highlight usefullness of this abstractions would help

Will be done in subsequent PRs.

PR #881, opened 2024-01-16
feat: consensus component config for main node and external node

pompon0:
## What ❔

consensus component config for main node and external node

## Why ❔

Consensus component has been implemented, now we need a way to enable it.


core/lib/zksync_core/src/consensus/storage/mod.rs
```diff
@@ -255,14 +255,22 @@ impl BlockStore {
...
             .context("miniblock disappeared")?;
-        let (genesis, _) = zksync_consensus_bft::testonly::make_genesis(
-            &[validator_key.clone()],
-            payload.encode(),
-            number,
-        );
-        txn.insert_certificate(ctx, &genesis.justification, self.inner.operator_address)
-            .await
-            .wrap("insert_certificate()")?;
+        let mut genesis = validator::testonly::GenesisSetup {
```
popzxc, line 258: Side comment: using `testonly` outside of tests looks a bit odd.

core/lib/zksync_core/src/lib.rs
```diff
@@ -505,6 +506,35 @@ pub async fn initialize_components(
         tracing::info!("initialized State Keeper in {elapsed:?}");
     }
 
+    if let Some(cfg) = configs.consensus_config.clone() {
```
popzxc, line 509: Would it make sense to create a `consensus` component and exit with an error if such a component is requested, but config is not provided?

pompon0: Consensus is in practice a part of state keeper, but I can add it as a separate component if you want. Should I do so?

popzxc: Up to you. Personally, I think I'd prefer it to be explicitly stated whether we want to run consensus or not.
E.g. right now if you add the required variables to the gitops-kubernetes, it'll immediately restart with consensus enabled, and in order to disable it, if we ever want to, you'll have to remove all these variables.
With a component-based approach enabling and disabling will be handled by adding a CLI argument independently of whether the config values are present.

core/lib/zksync_core/src/consensus/storage/mod.rs
pompon0: done

popzxc (approved):

otani88 (dismissed):

popzxc (approved):

perekopskiy (approved):

hatemosphere (approved):

yorik (approved):

PR #882, opened 2024-01-16
feat: Adding EN snapshots applier

tomg10:
Parts of this component that I decided to do in separate PRs:
-resilience


core/lib/snapshots_applier/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+        let mut storage = connection_pool
+            .access_storage_tagged("snapshots_applier")
+            .await?;
+        let mut storage = storage.start_transaction().await?;
+
+        let mut applied_snapshot_status = storage
+            .snapshot_recovery_dal()
+            .get_applied_snapshot_status()
+            .await
+            .map_err(|_| Retryable("Unable to get applied snapshots status".to_string()))?;
```
slowli, line 91: You have `From<SqlxError> for SnapshotApplierError` conversion defined above; why is it not used here and in similar situations? If you want to add additional context for errors, one way to do this would be to define an extension trait similar to `anyhow::Context` that would convert `Result<T, E>` where `E: Into< SnapshotApplierError>` + context into `Result<T, SnapshotApplierError>`. Something like this:

```rust
trait SnapshotApplierContext<T> {
    fn wrap(self, context: &'static str) -> Result<T, SnapshotApplierError>;
    // Can define a lazily evaluated context (e.g., `impl FnOnce() -> String`) as another method
}

impl<T, E: Into<SnapshotApplierError>> SnapshotApplierContext<T> for Result<T, E> {
    fn wrap(self, context: &'static str) -> Result<T, SnapshotApplierError> {
        match self {
            Ok(value) => Ok(value),
            Err(err) => Err(err.into().context(context)),
        }
    }
}

impl SnapshotApplierError {
    fn context(self, context: &'static str) -> Self {
        // To be able to add context, all error variants should be `anyhow::Error`,
        // or it can be refactored to a struct with `anyhow::Error` and severity fields
    }
}

// usage:
let mut applied_snapshot_status = storage
    .snapshot_recovery_dal()
    .get_applied_snapshot_status()
    .await
    .wrap("Unable to get applied snapshots status")?;
```

core/lib/snapshots_applier/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+        main_node_client: Box<dyn SnapshotsApplierMainNodeClient>,
+    ) -> Result<SnapshotRecoveryStatus, SnapshotApplierError> {
+        let snapshot_response = main_node_client
+            .fetch_newest_snapshot()
+            .await
+            .map_err(|_| Retryable("Unable to fetch newest snapshot from main node".to_string()))?;
+        if snapshot_response.is_none() {
+            return Err(Canceled("Main node does not have any ready snapshots, skipping initialization from snapshot!".to_string()));
+        }
+        let snapshot = snapshot_response.unwrap();
```
slowli, line 150: Nit: It's idiomatic to use `ok_or` / `ok_or_else` to map `Option::None` to a `Result::Err`.

core/lib/snapshots_applier/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+            miniblock_root_hash,
+            storage_logs_chunks_ids_already_processed: vec![],
+            storage_logs_chunks_ids_to_process: chunk_ids,
+        })
+    }
+
+    async fn recover_factory_deps(
+        &mut self,
+        storage: &mut StorageProcessor<'_>,
+    ) -> anyhow::Result<(), SnapshotApplierError> {
```
slowli, line 182: Nit: Why not simply a `Result` if you don't use `anyhow::Error` as the error type?

core/lib/snapshots_applier/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+                Some(SnapshotsApplier::create_fresh_recovery_status(main_node_client).await?)
+        }
+
+        let mut recovery = Self {
+            connection_pool,
+            blob_store,
+            applied_snapshot_status: applied_snapshot_status.unwrap(),
+        };
+
+        recovery.recover_factory_deps(&mut storage).await?;
```
slowli, line 125: What if factory deps are already recovered?

core/lib/snapshots_applier/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+            "Found snapshot with data up to l1_batch {}, storage_logs are divided into {} chunk(s)",
+            l1_batch_number,
+            snapshot.storage_logs_chunks.len()
+        );
+
+        let miniblock_root_hash = main_node_client
+            .miniblock_hash(snapshot.miniblock_number)
+            .await
+            .map_err(|err| Fatal(err.into()))?
+            .ok_or_else(|| Fatal(Error::msg("Miniblock is missing")))?;
```
slowli, line 162: This feels like a hack. What if we need more data in `SnapshotRecoveryStatus`? (JFYI, preliminarily we'd need `l1_batch_timestamp`, `miniblock_timestamp` and `protocol_version` to support snapshot recovery for the state keeper.) Maybe, it'd make sense to add missing data to `snapshot_response`?

core/lib/snapshots_applier/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+        storage
+            .storage_dal()
+            .insert_factory_deps(
+                self.applied_snapshot_status.miniblock_number,
+                &all_deps_hashmap,
+            )
+            .await?;
+        Ok(())
+    }
+    async fn insert_initial_writes_chunk(
```
slowli, line 202: Nit: Missing empty lines between methods (ditto in a couple of other places).

core/lib/snapshots_applier/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+
+    async fn recover_storage_logs_single_chunk(
+        &mut self,
+        chunk_id: u64,
+    ) -> Result<(), SnapshotApplierError> {
+        let mut storage = self
+            .connection_pool
+            .access_storage_tagged("snapshots_applier")
+            .await?;
+        let mut storage = storage.start_transaction().await?;
```
slowli, line 237: Dumb question: What's the reason for connecting to the DB and opening a transaction before getting data from the blob store?

core/lib/snapshots_applier/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+
+        let storage_logs = &storage_snapshot_chunk.storage_logs;
+        self.insert_storage_logs_chunk(storage_logs, &mut storage)
+            .await?;
+
+        self.insert_initial_writes_chunk(storage_logs, &mut storage)
+            .await?;
+
+        self.applied_snapshot_status
+            .storage_logs_chunks_ids_already_processed
```
slowli, line 257: Don't you need to change `storage_logs_chunks_ids_to_process` as well? See a line comment about the recovery progress representation.

core/lib/snapshots_applier/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+}
+
+#[async_trait]
+pub trait SnapshotsApplierMainNodeClient: fmt::Debug + Send + Sync {
+    async fn miniblock_hash(&self, number: MiniblockNumber) -> Result<Option<H256>, RpcError>;
+
+    async fn fetch_newest_snapshot(&self) -> Result<Option<SnapshotHeader>, RpcError>;
+}
+impl<'a> SnapshotsApplier<'a> {
+    pub async fn load_snapshot(
```
slowli, line 77: Would this be the entry point used by EN code? I'd expect `SnapshotApplierError` to be an internal detail (cf. updated EN components with similar transient / fatal error distinction, such as the consistency checker), so that retries are handled internally in the snapshot applier, and it only returns fatal errors / cancellations outside. It could be the case that the retry strategy should be configurable in this case, since otherwise we risk getting indefinitely stuck if we improperly classify a fatal error as transient.

core/lib/snapshots_applier/src/lib.rs
tomg10: I wasn't aware that additional data would be needed here and it's easier to test it now, I also assumed it would be easy to change this to use full miniblock later, but if you say it's certain it will be needed, I will change it now

core/lib/snapshots_applier/src/lib.rs
tomg10: I think I will just write a python script for my personal use that adds such lines as this is at least 3rd time I've missed this :D

core/lib/dal/src/snapshot_recovery_dal.rs
```diff
@@ -12,6 +12,18 @@ impl SnapshotRecoveryDal<'_, '_> {
         &mut self,
         status: &SnapshotRecoveryStatus,
     ) -> sqlx::Result<()> {
```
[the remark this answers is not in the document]
tomg10: I forgot that discussion, I agree that it makes sense to add those methods

core/lib/dal/migrations/20240112194527_snapshots_applier_processed_chunks_format_change.up.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE snapshot_recovery DROP COLUMN last_finished_chunk_id;
+ALTER TABLE snapshot_recovery DROP COLUMN total_chunk_count;
+
+ALTER TABLE snapshot_recovery ADD COLUMN storage_logs_chunks_ids_to_process INT[] NOT NULL;
+ALTER TABLE snapshot_recovery ADD COLUMN storage_logs_chunks_ids_already_processed INT[] NOT NULL;
```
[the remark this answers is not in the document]
tomg10: yes, I've written applier in a way that assumes that chunk_ids are unique and nothing more, just as a defensive mechanism. I agree having a boolean array has nice properties, but I think I can also minimize the risks of invalid state in dal (I will introduce a separate method that 'marks one chunk as complete') and it will have some guard mechanisms. For me those two arrays just look a bit more readable, though it's highly personal

core/lib/snapshots_applier/src/lib.rs
slowli: I'm 99% sure that *at least* these fields are needed; I've surveyed the state keeper logic to arrive at this list. Maybe it's not a complete list.

core/lib/dal/migrations/20240112194527_snapshots_applier_processed_chunks_format_change.up.sql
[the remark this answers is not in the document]
tomg10: I've actually changed my mind and changed those two arrays into a bool array

core/lib/snapshots_applier/src/lib.rs
tomg10: I've used a convention, where opening transaction is the first thing done in a method without deeper thought, I see that moving it after acesing blob store has some benefits, but I'm not yet convinced it's a meaningful optimization in a non-high-load-server situations or I'm missing something

core/lib/snapshots_applier/src/lib.rs
tomg10: I plan to refactor the code to behave similarly as you described in resiliency PR

core/lib/snapshots_applier/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+}
+
+impl From<SqlxError> for SnapshotApplierError {
+    fn from(err: SqlxError) -> Self {
+        match err {
+            SqlxError::Database(_) => Fatal(err.into()),
+            SqlxError::RowNotFound => Fatal(err.into()),
+            SqlxError::ColumnNotFound(_) => Fatal(err.into()),
+            SqlxError::Configuration(_) => Fatal(err.into()),
+            SqlxError::TypeNotFound { type_name: _ } => Fatal(err.into()),
```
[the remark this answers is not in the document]
tomg10: nice! thanks

core/lib/snapshots_applier/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+    jsonrpsee::core::ClientError as RpcError,
+    namespaces::{EthNamespaceClient, ZksNamespaceClient},
+};
+
+use crate::SnapshotApplierError::*;
+
+mod tests;
+
+#[derive(thiserror::Error, Debug)]
+pub enum SnapshotApplierError {
```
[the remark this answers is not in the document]
tomg10: Yeah, I didn't want to spend too much effort deciding which potentially rare errors are retryable as the worst case scenario is that it would do a meaningless retry a few times

core/lib/snapshots_applier/src/lib.rs
tomg10: Initial checks and inserting those factory deps are inside a single transaction, so this situation cannot occur

core/lib/snapshots_applier/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+    zkevm_test_harness::k256::pkcs8::der::EncodeValue,
+    MiniblockNumber, H256,
+};
+use zksync_utils::bytecode::hash_bytecode;
+use zksync_web3_decl::{
+    jsonrpsee::core::ClientError as RpcError,
+    namespaces::{EthNamespaceClient, ZksNamespaceClient},
+};
+
+use crate::SnapshotApplierError::*;
```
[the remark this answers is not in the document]
tomg10: I don't have a strong opinion on that, corrected

core/lib/dal/src/storage_logs_dal.rs
```diff
@@ -67,6 +67,47 @@ impl StorageLogsDal<'_, '_> {
...
+                r"\\x{hashed_key:x}|\\x{address:x}|\\x{key:x}|\\x{value:x}|",
+                hashed_key = log.key.hashed_key(),
+                address = log.key.address(),
+                key = log.key.key(),
+                value = log.value
+            );
+            writeln_str!(
+                &mut buffer,
+                r"{}|\\x{:x}|{miniblock_number}|{now}|{now}",
+                log.enumeration_index,
```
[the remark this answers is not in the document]
tomg10: as discussed on slack, I'm leaving this as it is

core/lib/snapshots_applier/src/lib.rs
tomg10: update: I've moved it when adding metrics

core/lib/snapshots_applier/src/lib.rs
tomg10: I've actually just forgot to refactor those places, thanks for providing the snippet, I plan to use it when adding resiliency

core/lib/snapshots_applier/src/lib.rs
```diff
@@ -0,0 +1,356 @@
...
+pub trait SnapshotsApplierMainNodeClient: fmt::Debug + Send + Sync {
+    async fn fetch_l2_block(&self, number: MiniblockNumber) -> Result<Option<SyncBlock>, RpcError>;
+
+    async fn fetch_newest_snapshot(&self) -> Result<Option<SnapshotHeader>, RpcError>;
+}
+impl<'a> SnapshotsApplier<'a> {
+    pub async fn load_snapshot(
+        connection_pool: &'a ConnectionPool,
+        main_node_client: Box<dyn SnapshotsApplierMainNodeClient>,
+        blob_store: Box<dyn ObjectStore>,
```
slowli, line 125: Nit: Inconsistency among arguments: `connection_pool` is supplied by ref, while others are supplied by value. I'd suggest to supply them all by ref.

core/lib/snapshots_applier/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+license = "MIT OR Apache-2.0"
+keywords = ["blockchain", "zksync"]
+categories = ["cryptography"]
+
+
+[dependencies]
+zksync_dal = { path = "../../lib/dal" }
+zksync_types = { path = "../../lib/types" }
+zksync_object_store = { path = "../../lib/object_store" }
+zksync_core = { path = "../../lib/zksync_core" }
```
slowli, line 17: This dependency is not used.

core/lib/snapshots_applier/src/lib.rs
slowli: I've meant that the same deps may be inserted into Postgres multiple times, but this is fine given that the relevant DB statement has `ON CONFLICT (bytecode_hash) DO NOTHING`.

core/lib/snapshots_applier/src/lib.rs
```diff
@@ -0,0 +1,356 @@
...
+                .iter()
+                .filter(|x| !(**x))
+                .count(),
+        );
+
+        recovery.recover_factory_deps(&mut storage).await?;
+
+        storage
+            .snapshot_recovery_dal()
+            .insert_initial_recovery_status(&recovery.applied_snapshot_status)
```
slowli, line 189: This will error if the status is already present in Postgres (i.e., if it was loaded from it above).

core/lib/snapshots_applier/src/tests.rs
```diff
@@ -0,0 +1,229 @@
...
+    async fn fetch_newest_snapshot(&self) -> Result<Option<SnapshotHeader>, RpcError> {
+        if let Some(response) = self.fetch_newest_snapshot_response.clone() {
+            Ok(Some(response))
+        } else {
+            Ok(None)
+        }
+    }
+}
+#[cfg(test)]
+mod snapshots_applier_tests {
```
slowli, line 34: Bikeshedding: this module looks redundant. I'd rather move utils above in `tests::utils` module.

core/lib/dal/src/storage_logs_dal.rs
```diff
@@ -502,6 +545,65 @@ impl StorageLogsDal<'_, '_> {
             .collect()
     }
 
+    // Retrieves all storage log entries for testing purposes.
+    pub async fn get_all_storage_logs_for_tests(&mut self) -> Vec<StorageLogDbRow> {
```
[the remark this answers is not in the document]
tomg10: I agree that "dump_(...)" is a nice convention, but I'm not yet a very big fan of the separate impl block. I suspect that as soon as such block gets bigger, it would be very easy to miss that you are adding something to test-only impl instead of one you intend to if you just add the method at the end of the impl

core/lib/types/src/snapshots.rs
```diff
@@ -193,6 +193,25 @@ pub struct SnapshotRecoveryStatus {
...
     pub miniblock_number: MiniblockNumber,
     pub miniblock_root_hash: H256,
-    pub last_finished_chunk_id: Option<u64>,
-    pub total_chunk_count: u64,
+    pub storage_logs_chunks_processed: Vec<bool>,
+}
+
+// Used only in tests
+#[derive(Debug, PartialEq)]
+pub struct InitialWriteDbRow {
```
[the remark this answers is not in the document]
tomg10: I've seen those methods, but I think that using them would introduce too much logic in the tests and make the tests less readable enough to justify introducing dedicated methods

core/lib/snapshots_applier/src/lib.rs
```diff
@@ -0,0 +1,356 @@
...
+}
+
+impl From<SqlxError> for SnapshotsApplierError {
+    fn from(err: SqlxError) -> Self {
+        match err {
+            SqlxError::Database(_)
+            | SqlxError::RowNotFound
+            | SqlxError::ColumnNotFound(_)
+            | SqlxError::Configuration(_)
+            | SqlxError::TypeNotFound { type_name: _ } => SnapshotsApplierError::Fatal(err.into()),
```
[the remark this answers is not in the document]
tomg10: nice!

core/lib/snapshots_applier/src/lib.rs
tomg10: nice catch!

core/lib/snapshots_applier/src/lib.rs
```diff
@@ -0,0 +1,356 @@
...
+    }
+}
+
+impl From<RpcError> for SnapshotsApplierError {
+    fn from(err: RpcError) -> Self {
+        match err {
+            Error::Transport(_) | Error::RequestTimeout | Error::RestartNeeded(_) => {
+                SnapshotsApplierError::Retryable(
+                    "An error occurred while accessing main node".to_string(),
+                )
```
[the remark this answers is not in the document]
tomg10: I've refactored so that retryable error contains error instead of string. About first point, there are only 3 places in which I construct Canceled, so I don't think it's worth it.

core/lib/snapshots_applier/src/tests.rs
tomg10: which methods exactly do you suggest moving to test utils?

core/lib/types/src/snapshots.rs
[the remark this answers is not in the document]
slowli: IMO, testing that "real" DAL methods work when invoked with snapshot-related inputs is better than creating dedicated methods for testing purposes, but that's a personal opinion. Curious to see what's @RomanBrodetski's opinion on the matter. If we go with the current approach, please place `*Row` structs in the DAL crate; e.g., you can replace `DBStorageLog` because it doesn't seem to be used anywhere.

core/lib/snapshots_applier/src/lib.rs
```diff
@@ -0,0 +1,358 @@
...
+
+    async fn create_fresh_recovery_status(
+        main_node_client: &'a dyn SnapshotsApplierMainNodeClient,
+    ) -> anyhow::Result<SnapshotRecoveryStatus> {
+        let snapshot_response = main_node_client.fetch_newest_snapshot().await?;
+
+        let snapshot = snapshot_response.ok_or(SnapshotsApplierError::Canceled(
+            "Main node does not have any ready snapshots, skipping initialization from snapshot!"
+                .to_string(),
+        )).context("No snapshots were found in main node")?;
```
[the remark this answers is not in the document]
tomg10: About the first point, I think It's an extreme edge-case that and I prefer not to add this to the config, imo it will only confuse the users.
About the second point, I didn't yet catch what you mean, you suggest having an error attached to canceled instead of a string? If so, why is it better?

core/lib/snapshots_applier/src/lib.rs
```diff
@@ -0,0 +1,358 @@
...
+                Self::Retryable(error.into())
+            }
+            _ => Self::Fatal(error.into()),
+        }
+    }
+}
+
+#[async_trait]
+pub trait SnapshotsApplierMainNodeClient: fmt::Debug + Send + Sync {
+    async fn fetch_l2_block(&self, number: MiniblockNumber) -> Result<Option<SyncBlock>, RpcError>;
```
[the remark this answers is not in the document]
tomg10: What's wrong with what we are currently doing, i.e. what's your biggest motivation for suggesting this change?

core/lib/snapshots_applier/src/lib.rs
```diff
@@ -0,0 +1,358 @@
...
+        match error {
+            Error::Transport(_) | Error::RequestTimeout | Error::RestartNeeded(_) => {
+                Self::Retryable(error.into())
+            }
+            _ => Self::Fatal(error.into()),
+        }
+    }
+}
+
+#[async_trait]
```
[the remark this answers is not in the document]
tomg10: yes, ideally as a part of PR that ties everything together and adds typescript integration test

core/lib/snapshots_applier/src/lib.rs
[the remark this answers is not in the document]
slowli: It unnecessarily ties trait implementations to a particular implementation you have in mind (i.e., using `en_syncL2Block` RPC endpoint). Not saying that it's a dealbreaker, but I'd argue that such tight coupling is not necessary in this case, given that there are viable alternatives that I've listed above, and only a relatively small part of the returned `SyncBlock` is actually used in the recovery logic. I won't insist on this change; it's mostly a stylistic issue rather than a practical one.

core/lib/snapshots_applier/src/lib.rs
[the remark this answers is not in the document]
slowli: > imo it will only confuse the users.

Fair enough. Just to provide more context: I think snapshot recovery will be an experimental, opt-in feature initially; at least, I'd consider it safe to position it as such. So if we don't provide configurability in this scenario, a safe option would be to halt, so that a user can either fix their setup (network connectivity, GCS bucket options etc.) and retry, or switch off snapshot recovery entirely and sync using other means.

> you suggest removing that String from SnappshotsApplierError::Canceled and attaching message via context?

No, I don't propose that. My line or reasoning is as follows. `snapshot_response.ok_or(_)` returns `Result<_, SnapshotsApplierError>`. Then, you can `anyhow::Context::context` on this `Result`, wrapping an error in  `anyhow::Error`. The question is: why not just create an `anyhow::Error` directly using `snapshot_response.context(_)`? My guess that I've mentioned above is using `root_cause()` of the error when it is handled, which doesn't look like a particularly good idea.

core/lib/snapshots_applier/src/tests.rs
```diff
@@ -0,0 +1,234 @@
...
+    impl SnapshotsApplierMainNodeClient for MockMainNodeClient {
+        async fn fetch_l2_block(
+            &self,
+            number: MiniblockNumber,
+        ) -> Result<Option<SyncBlock>, RpcError> {
+            if let Some(response) = self.fetch_l2_block_responses.get(&number) {
+                Ok(Some((*response).clone()))
+            } else {
+                Ok(None)
+            }
```
slowli, line 33: This can be shortened to `Ok(self.fetch_l2_block_responses.get(&number).cloned())`. Ditto for the below method – can be shortened to `Ok(self.fetch_newest_snapshot_response.clone())`.

core/lib/snapshots_applier/src/lib.rs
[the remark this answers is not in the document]
tomg10: Corrected. About the config, I think it's best to return to this conversation on resiliency PR, but also I misunderstood earlier what you want to encode in node config, everything what you have written now makes perfect sense

core/lib/snapshots_applier/src/lib.rs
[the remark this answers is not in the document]
tomg10: I get your argument, but I'm not convinced the difference in code quality would be substantial enough to justify making this change

slowli (approved):
There are some minor nits (if you don't mind, I can fix them in a follow-up PR), but nothing preventing merging.

slowli (approved):

PR #900, opened 2024-01-19
feat(en): Make state keeper work with pruned data

slowli:
## What ❔

Modifies state keeper so that it works with pruned node data during snapshot recovery.

## Why ❔

## Checklist


core/lib/zksync_core/src/state_keeper/io/tests/mod.rs
```diff
@@ -431,6 +434,113 @@ async fn miniblock_and_l1_batch_processing_with_sync_sealer() {
     test_miniblock_and_l1_batch_processing(pool, 0).await;
 }
 
+// FIXME (PLA-589): Doesn't work because of missing system contracts, which cannot be added w/o a miniblock
```
slowli, line 437: Until foreign key constraints referencing miniblocks / L1 batches are removed, it seems impossible to meaningfully test state keeper I/O implementations. @tomg10 maybe it makes sense to extract the removal of these constraints into a separate PR?

core/lib/dal/src/transactions_dal.rs
```diff
@@ -1247,31 +1249,57 @@ impl TransactionsDal<'_, '_> {
...
+            .collect();
+
+        let mut data = Vec::with_capacity(transactions_by_miniblock.len());
+        let it = transactions_by_miniblock.into_iter().zip(miniblock_data);
+        for ((number, txs), miniblock_row) in it {
+            let prev_miniblock_number = number - 1;
+            let prev_block_hash = match prev_miniblock_hashes.get(&prev_miniblock_number) {
+                Some(hash) => *hash,
+                None => {
+                    // Can occur after snapshot recovery; the first previous miniblock may not be present
```
RomanBrodetski, line 1269: good catch, but I'm curious what we can do (or already doing) so that we don't forget such cases anywhere. Imagine we are adding a new component and ely on the fact that the previous miniblock hash is always present. Is there a way to model data differently so that we don't have to think about it? Probably not. An example would be to store prev miniblock hash in the next miniblock row, but that results in denormalization and is not practical

core/lib/zksync_core/src/state_keeper/io/common/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+    let max_wait_millis = max_wait.as_millis() as u64;
+    let delay_interval_millis = delay_interval.as_millis() as u64;
+    assert!(delay_interval_millis > 0, "delay interval must be positive");
+
+    ((max_wait_millis + delay_interval_millis - 1) / delay_interval_millis).max(1) as usize
+}
+
+/// Cursor of the miniblock / L1 batch progress used by [`StateKeeperIO`](super::StateKeeperIO) implementations.
+#[derive(Debug)]
+pub(crate) struct IoCursor {
```
RomanBrodetski, line 24: do I understand it right that now you use this struct both in EN and on the Main Node? Very nice that we are able to reuse!

RomanBrodetski (dismissed):
looks really good. For the amount of changes in invariants, I expected worse changes are needed on the code level.

Like the  `L1BatchParamsProvider` abstraction. I don't see a good way around it.

Are the new invariants documented anywhere? Like, what can programmer rely on when writing code that runs agains our DB? Something like:
* there are no gaps in miniblock/l1 batch indices
* if miniblock #N is missing, there is a snapshot with miniblock >N


Separately, would Main Node work with pruned data? I don't see why not.

core/lib/zksync_core/src/state_keeper/io/common/mod.rs
slowli: Yes, it's used in both I/O implementations.

slowli:
> Separately, would Main Node work with pruned data? I don't see why not.

Right now it won't because some components only run on the main node would need to be adapted to support snapshot recovery. Such adaptation seems straightforward though.

core/lib/zksync_core/src/state_keeper/io/mempool.rs
```diff
@@ -128,15 +161,9 @@ impl StateKeeperIO for MempoolIO {
         max_wait: Duration,
     ) -> Option<(SystemEnv, L1BatchEnv)> {
         let deadline = Instant::now() + max_wait;
-
+        // FIXME: why do we wait for hash immediately and not below? (changed in #809)
         let prev_l1_batch_hash = self.load_previous_l1_batch_hash().await;
```
slowli, line 165: Highlighting this. Not sure whether this influences performance in practice, but theoretically it could be negative. Maybe it makes sense to rename `load_previous_l1_batch_hash` to `wait_for_previous_l1_batch_hash` to be more explicit with what the method does? If this line stays, the "We only need to get the root hash when we're certain that we have a new transaction" comment below should be removed as obsolete. cc @StanislavBreadless

RomanBrodetski: The comment seems to not match the logic indeed - probably a bug/leftover. 

> but theoretically it could be negative
@slowli , under which circumstances? From the first glance, we need to wait both for a transaction and for the hash. It doesn't matter in what order to check unless there are heavy operations in betsween

RomanBrodetski (dismissed):

core/lib/zksync_core/src/state_keeper/io/mempool.rs
slowli: Yes, you are right; the order doesn't seem to matter for performance. There are some differences in method semantics though. In the old implementation, once a fitting transaction appears in the mempool, we'll wait for the previous L1 batch hash as long as it takes, and the method is guaranteed to return `Some(_)`. In the new implementation, if we wait for the previous L1 batch hash for more than `max_wait`, the method won't even check the mempool and will immediately return `None`. Now, since the method is only called from [a loop in `ZkSyncStateKeeper`](https://github.com/matter-labs/zksync-era/blob/a8699a20a5e6ff1c6ea99412019f9bfecc3b92f5/core/lib/zksync_core/src/state_keeper/keeper.rs#L258), it doesn't seem to make much difference; the method will presumably get the prev L1 batch hash immediately on the next iteration and will still return `Some(_)` if there are fitting transactions in the mempool. One side-effect of this is that the `wait_for_prev_hash_time` no longer works as intended; it reports bogus near-zero values on these repeated loop iterations.

RomanBrodetski: I see. So what would you propose? I'm fine to keep as is (but please remove comment) or change - whatever feels right for you

RomanBrodetski (approved):

PR #901, opened 2024-01-19
feat(api): Consider State keeper fee model input in the API

StanislavBreadless:
## What ❔

Before, under big spikes in L1 gas price, it was possible that the state keeper retained the inflated gas price, while the API got the smaller one. In this PR we ensure that the API takes into account the fee params of the latest sealed miniblock

## Checklist

### DIFF
--- a/core/lib/zksync_core/src/fee_model.rs
+++ b/core/lib/zksync_core/src/fee_model.rs
@@ -12,10 +12,11 @@ use zksync_utils::ceil_div_u256;
 use crate::l1_gas_price::L1GasPriceProvider;
 
 /// Trait responsible for providing fee info for a batch
+#[async_trait::async_trait]
 pub trait BatchFeeModelInputProvider: fmt::Debug + 'static + Send + Sync {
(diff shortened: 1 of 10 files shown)


core/lib/zksync_core/src/fee_model.rs
```diff
@@ -77,6 +79,59 @@ impl MainNodeFeeInputProvider {
...
+            .inner
+            .get_batch_fee_input_scaled(l1_gas_price_scale_factor, l1_pubdata_price_scale_factor)
+            .await;
+        let last_miniblock_params = self
+            .connection_pool
+            .access_storage_tagged("api_fee_input_provider")
+            .await
+            .unwrap()
+            .blocks_dal()
+            .get_last_sealed_miniblock_header()
```
perekopskiy, line 120: Let's add a specialised method to DAL that loads only `batch_fee_input`

StanislavBreadless: Discussed in private not needed right now

PR #902, opened 2024-01-19
refactor(api): Propagate DB errors in `TxSender`

slowli:
## What ❔

Propagates DB errors in `TxSender` instead of panicking on them.

## Why ❔

Rust panics are slow and are more difficult to meter compared to errors.

## Checklist


Deniallugo:
I'm a big fan of this change. but @RomanBrodetski was against it before

popzxc (approved):
❤️

PR #910, opened 2024-01-22
feat(docs): Article about 4844 (blobs)

mm-zk:
## What ❔

* Article explaining how we adopted our system to 4844

## Why ❔

* To increase the knowledge within the team and beyond


docs/guides/advanced/data/4844_data.md
```diff
@@ -0,0 +1,185 @@
...
+- Users must pay for a whole blob, which is 128kb, regardless of how much data they actually use. This differs from
+  calldata, where payment was based on the amount of data sent.
+- The contents of the blobs are not directly accessible. Instead, they contain unique KZG commitments and precompiles.
+
+**In the first version, we will be using up to 2 blobs per batch - so 256kb of pubdata**
+
+In this article, we will mainly discuss how we adapted our system, including our proving system, to work with these KZG
+commitments.
+
+**Why did Ethereum introduce these commitments instead of continuing with the previous hash-based system?**
```
StanislavBreadless, line 32: Bikeshedding: Probably "calldata-based" is more precise name

docs/guides/advanced/data/4844_data.md
```diff
@@ -0,0 +1,185 @@
...
+We will skip the technical details of 4844 blobs and focus on how zkSync specifically implements them.
+
+## Before 4844: The Era of Calldata
+
+Originally, before the introduction of 4844 blobs, the system worked differently. In this period, known as the pre-4844
+era, zkSync would collect all its pubdata, which could be up to 120kb. This data was then sent as 'calldata' of the
+commitBatches transaction.
+
+In these transactions, the contract would calculate a keccak hash of the data and use this hash in the proving process.
+More precisely, this hash was stored as part of the block commitment and later used in the `proveBatches` function.
```
StanislavBreadless, line 15: The pubdata was not used in the proving process.

in the proving process we only used the hash of all the state diffs (i.e. uncompressed ones). This hash is calculated on L2 and passed by system logs. It will continue being the case for the blob based system too.

Only together with blob introduction the commitments to the chunks of pubdata will be needed to prove the blob correctness

docs/guides/advanced/data/4844_data.md
```diff
@@ -0,0 +1,185 @@
...
+| Versioned Hash:            | Generate a versioned hash (`vh = hash(c)`) which is more flexible and smaller in size than the commitment. (32 bytes vs 48)                                                                                                                  |
+| Choose a Random Point:     | Select a 'random' point `x = hash(hash(data) \|\| vh)` using the [hash of the data and the versioned hash][random_point_selection]. This is ['Fiat-Shamir' heuristic][fiat_shamir]                                                           |
+| Compute Polynomial Value:  | Calculate the value `p(x) = y`.                                                                                                                                                                                                              |
+| Prepare for Circuit Input: | Ensure that `hash(vh \|\| x \|\| y)` is provided as a [public input for the circuit][scheduler_expect_4844], based on the [Schwarz-Zippel lemma][schwarz_zippel] - if 2 polynomials are equal in a random point, there are equal everywhere. |
```
StanislavBreadless, line 61: Worth mentioning that `hash(data)` is also provided

docs/guides/advanced/data/4844_data.md
```diff
@@ -0,0 +1,185 @@
...
+### What, but why
+
+#### Why does prover need to compute the commitment
+
+Prover needs to come up with a 'random' point 'x' (otherwise Schwarz-Zippel lemma would not work), so it has to generate
+based on something. The simpler solution would be to generate it based on the hash of the pubdata, but unfortunately the
+L1 contract doesn't have access to the hash - so it would not be able to verify that the 'x' was really generated based
+on that hash.
+
+#### Why do I need to pass blob proof to L1 contract
```
StanislavBreadless, line 91: not sure if I understand. there are two proofs that need to be provided with the transaction:

- Blob proof (tbh, I didn't dive into why do we need it yet)
- The point evaluation proof (that `p(x) = y`). 

If we take the analogy of "merkle root", then I don't think i would need to only provide point evaluation proof (i.e. merkle path) and not anything else

docs/guides/advanced/data/4844_data.md
```diff
@@ -0,0 +1,185 @@
...
+In these transactions, the contract would calculate a keccak hash of the data and use this hash in the proving process.
+More precisely, this hash was stored as part of the block commitment and later used in the `proveBatches` function.
+
+## The Change with 4844 Blob Transactions
+
+In January 2024, the introduction of 4844 blobs brought several significant changes:
+
+- Blob transactions have their own fee market. This means the cost of using blobs can be different from the cost of
+  using gas.
+- Users must pay for a whole blob, which is 128kb, regardless of how much data they actually use. This differs from
```
koloz193, line 23: maybe worth mentioning there can be more than 1 blob used and users will pay for all of it

docs/guides/advanced/data/4844_data.md
```diff
@@ -0,0 +1,185 @@
...
+| step                         | explanation                                            |
+| ---------------------------- | ------------------------------------------------------ |
+| Collect data                 | Get `c`, `x`, `y`, `blob_proof` from witness generator |
+| Create New Transaction Type: | Form a new blob transaction with a special `0x3` type. |
+
+### Inside the L1 contract
+
+| step                             | explanation                                                                                                                                                                                                              |
+| -------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| Receive Inputs:                  | Get `c`, `x`, `y`, and `blob_proof` as inputs from the [state keeper][executor_inputs].                                                                                                                                  |
```
koloz193, line 74: same comment on blob_proof as above

docs/guides/advanced/data/4844_data.md
```diff
@@ -0,0 +1,185 @@
...
+| Verification Hash:               | Obtain the verification hash through the BLOB_HASH precompile. [`_getBlobVersionedHash`][get_blob_vhash]                                                                                                                 |
+| Use Precompile for Verification: | Apply a [precompile][precompile_4844] to check if `p(x) == y` using the version hash, `c` and `blob_proof`. This is where the KZG scheme plays a crucial role, allowing the proof without needing the entire polynomial. |
```
koloz193, line 76: see https://github.com/matter-labs/zksync-era/pull/910/files#r1461989911

docs/guides/advanced/data/4844_data.md
koloz193: L1 contract only gets the opening proof, blob proof is used by nodes to ensure the blobs == kzg commitments == versioned hashes

docs/guides/advanced/data/4844_data.md
```diff
@@ -0,0 +1,185 @@
+# Understanding zkSync's Use of 4844 Blob Transactions
+
+This article explains how zkSync is utilizing a new type of transaction called 4844 blob transactions to transfer its
+public data (pubdata) to Ethereum Layer 1 (L1).
```
barakshani, line 4:
```suggestion
This article explains how zkSync is utilizing the new type of transaction called 4844
blob transactions to commit to the public data (pubdata) to Ethereum Layer 1 (L1).

File: docs/guides/advanced/data/4844_data.md [hunk 9]
```diff
@@ -0,0 +1,185 @@
...
+
+We will skip the technical details of 4844 blobs and focus on how zkSync specifically implements them.
```
barakshani, line 6: worth giving a reference

docs/guides/advanced/data/4844_data.md
```diff
@@ -0,0 +1,185 @@
...
+
+## Before 4844: The Era of Calldata
+
+Originally, before the introduction of 4844 blobs, the system worked differently. In this period, known as the pre-4844
+era, zkSync would collect all its pubdata, which could be up to 120kb. This data was then sent as 'calldata' of the
+commitBatches transaction.
```
barakshani, line 12: If you refer to this era in the past, then I guess "would collect" --> "collected" (also below).

I also feel like these 3 sentences could be written in 1: Before the introduction... zkSync collected... and sent...

docs/guides/advanced/data/4844_data.md
```diff
@@ -0,0 +1,185 @@
...
+
+In these transactions, the contract would calculate a keccak hash of the data and use this hash in the proving process.
+More precisely, this hash was stored as part of the block commitment and later used in the `proveBatches` function.
+
+## The Change with 4844 Blob Transactions
+
+In January 2024, the introduction of 4844 blobs brought several significant changes:
```
barakshani, line 19:
```suggestion
The introduction of 4844 blobs brought several significant changes:
```
The introduction of 4844 was before Jan '24

docs/guides/advanced/data/4844_data.md
barakshani, line 23: The term "users" is a bit confusing here, it's the L1 tx that must pay for it (not, e.g., L2 users).
It's probably worth mentioning that the data is represented as a 128kb commitment, and this is why even little data will pay for the whole thing. So maybe have this point after the next.

docs/guides/advanced/data/4844_data.md
```diff
@@ -0,0 +1,185 @@
...
+
+- Blob transactions have their own fee market. This means the cost of using blobs can be different from the cost of
```
barakshani, line 21: I don't think the following is addressed anywhere: what if the price-per-byte exceeds the calldata price?

docs/guides/advanced/data/4844_data.md
barakshani, line 32: the previous hash-based system is the zkSync way, not Ethereum. Maybe replace with "raw data".

docs/guides/advanced/data/4844_data.md
```diff
@@ -0,0 +1,185 @@
...
+The short answer is, that while the current version of Ethereum doesn't gain much from this change, the next version,
+known as 'full Danksharding,' will greatly benefit from these commitments, as they will allow the addition of many more
+blobs.
+
+## Technical details
+
+Let's go back to our problem statement:
+
+**We have 128kb of data in L2, and we want to prove on L1 that it was correctly 'attached' to L1 (that is - that it was
+made available).**
```
barakshani, line 43:
```suggestion
**We have 128kb of data in L2, and we want to prove on L1 that the KZG commitment was constructed correctly from this data.**

File: docs/guides/advanced/data/4844_data.md [hunk 14]
```diff
@@ -0,0 +1,185 @@
...
+
+This verification is done in 14 steps, divided into three parts: inside the prover, inside the state keeper, and finally
+within the L1 contract.
```
barakshani, line 46:
```suggestion
This proof and its verification are done in 14 steps, divided into three parts: inside the prover, inside the state 
keeper, and finally within the L1 contract.

File: docs/guides/advanced/data/4844_data.md [hunk 15]
```diff
@@ -0,0 +1,185 @@
...
+
+### Inside the prover
+
+Most of this is done in either witness generator or a new circuit type 4844:
+
+| step                       | explanation                                                                                                                                                                                                                                  |
+| -------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Split data                 | Divide the data into 4096 units of 256-bit points.                                                                                                                                                                                           |
```
barakshani, line 54:
```suggestion
| Split Data                 | Divide the data into 4096 units of 256-bit points.

File: docs/guides/advanced/data/4844_data.md [hunk 16]
```diff
@@ -0,0 +1,185 @@
...
+| Create Polynomial          | Form a polynomial `p` that passes through these points. For more clarity on this, refer to [Blobspace 101][blobspace_101]                                                                                                                    |
+| Trusted setup              | Obtain the [trusted setup][trusted_setup] from Ethereum's KZG ceremony (also known as the secret point 's').                                                                                                                                 |
```
barakshani, line 56:
```suggestion
| Trusted Setup              | Obtain the [trusted setup][trusted_setup] values from Ethereum's KZG ceremony (also known as the secret point 's').     
```
I'm not so sure about "the secret point 's'" -- the secret point is secret, this is not what you get from the ceremony.

docs/guides/advanced/data/4844_data.md
```diff
@@ -0,0 +1,185 @@
...
+| Compute Commitment         | [Calculate the commitment][compute_commitment] `(c = p(s))`.                                                                                                                                                                                 |
+| Versioned Hash:            | Generate a versioned hash (`vh = hash(c)`) which is more flexible and smaller in size than the commitment. (32 bytes vs 48)                                                                                                                  |
```
barakshani, line 58: why : all of a sudden?

docs/guides/advanced/data/4844_data.md
```diff
@@ -0,0 +1,185 @@
...
+| Choose a Random Point:     | Select a 'random' point `x = hash(hash(data) \|\| vh)` using the [hash of the data and the versioned hash][random_point_selection]. This is ['Fiat-Shamir' heuristic][fiat_shamir]                                                           |
```
barakshani, line 59: below we write that x is not generated using hash(data)...

heuristic --> transform

docs/guides/advanced/data/4844_data.md
barakshani, line 61: The lemma does not state that they are equal, but bounds the probability they are not. But why to even mention this lemma here?

docs/guides/advanced/data/4844_data.md
```diff
@@ -0,0 +1,185 @@
...
+| Compute Polynomial Value:  | Calculate the value `p(x) = y`.                                                                                                                                                                                                              |
+| Prepare for Circuit Input: | Ensure that `hash(vh \|\| x \|\| y)` is provided as a [public input for the circuit][scheduler_expect_4844], based on the [Schwarz-Zippel lemma][schwarz_zippel] - if 2 polynomials are equal in a random point, there are equal everywhere. |
+
+### Inside the state keeper
+
+| step                         | explanation                                            |
+| ---------------------------- | ------------------------------------------------------ |
+| Collect data                 | Get `c`, `x`, `y`, `blob_proof` from witness generator |
```
barakshani, line 67:
```suggestion
| Collect Data                 | Get `c`, `x`, `y`, `blob_proof` from witness generator |

File: docs/guides/advanced/data/4844_data.md (hunk 6 above)
Comment (barakshani) [inline docs/guides/advanced/data/4844_data.md:74]: What do you mean "from the state keeper"? Interaction with contracts are via transactions

File: docs/guides/advanced/data/4844_data.md [hunk 20]
```diff
@@ -0,0 +1,185 @@
...
+| Pass Public Input to Verifier    | Send `hash(vh \|\| x \|\| y)` as public input to the verifier using [`_batchAuxiliaryOutput`][executor_aux].                                                                                                             |
+
+This process simplifies how zkSync validates the attachment of 128kb of L2 data to L1, ensuring its correctness and
+integrity.
+
+### What, but why
+
+#### Why does prover need to compute the commitment
+
+Prover needs to come up with a 'random' point 'x' (otherwise Schwarz-Zippel lemma would not work), so it has to generate
```
barakshani, line 86:
```suggestion
The prover needs to come up with a 'random' point 'x' (otherwise we cannot apply the Schwarz-Zippel lemma ), so it has to be generated

File: docs/guides/advanced/data/4844_data.md [hunk 21]
```diff
@@ -0,0 +1,185 @@
...
+based on something. The simpler solution would be to generate it based on the hash of the pubdata, but unfortunately the
```
barakshani, line 87: as above, the point x is generates using the hash of the data...

docs/guides/advanced/data/4844_data.md
```diff
@@ -0,0 +1,185 @@
...
+L1 contract doesn't have access to the hash - so it would not be able to verify that the 'x' was really generated based
+on that hash.
+
+#### Why do I need to pass blob proof to L1 contract
+
+Execution layer only know that a consensus layer has a blob with a given hash, and nothing more. So in order to prove
+anything about the contents of the blob, you (as a caller), have to provide all the information. It is similar to
+someone storing a full merkle tree, and you knowing only the roto hash. If you want to prove the inclusion of the leaf,
```
barakshani, line 95:
```suggestion
The execution layer only knows that a consensus layer has a blob with a given hash, and nothing more. So in order to 
prove anything about the contents of the blob, you (as a caller) have to provide all the information. It is similar to
someone storing a full merkle tree, and you knowing only the root hash. If you want to prove the inclusion

File: docs/guides/advanced/data/4844_data.md [hunk 23]
```diff
@@ -0,0 +1,185 @@
...
+#### Why do I need to pass this information into the aux inputs of verifier
+
+While we say, that verifier accepts public inputs (plural) - in practice it accepts a single hash, that is an aggregated
+keccak of all the inputs.
+
+And passing all the information there, is the only way to make sure that operator was not malicious (otherwise, operator
+could pass some other data in a blob, with correct `x` and `y` etc, that would be completely different from the one that
+prover was actually proving).
+
+#### Why do you say that point 's' is secret if I can simply download it
```
barakshani, line 107: See above, I recommend removing this.

docs/guides/advanced/data/4844_data.md
mm-zk: clarified

docs/guides/advanced/data/4844_data.md
barakshani: I don't see a reason to explain it at all. I'm not even sure that "the secret point 's'" is _known_. You can probably replace with CRS/SRS if you want.
Note also that it's not Atheneum's ceremony, but Aztec's (https://aztec.network/blog/aztec-crs-the-biggest-mpc-setup-in-history-has-successfully-finished/).

docs/guides/advanced/data/4844_data.md
barakshani: I meant the colon symbol :
There are other places below as well

docs/guides/advanced/data/4844_data.md
StanislavBreadless: only the hash of the state diffs was part of batch commitment (i.e. the data that is used by prover)

the hash of the pubdata was published via system L2->L1 logs and accessed during `commitBatches` function call & it was there to make sure that the correct pubdata has been sent

deniallugo-ml:

PR #911, opened 2024-01-22
fix(error & syntax): Optimize error handling

wangenkai1024:
## What ❔
1. Before the if statement, the else keyword is used, which is incorrect. The unwrap_or_else method should be used before the if statement to handle possible errors.

2. Unwrap() is used to handle possible errors, which may cause the program to crash when an error is encountered. A better approach is to use the ? operator to propagate errors so that the function can return an error when it encounters it, rather than crashing.

I added two new error types StorageAccessFailed and FailedTransactionCheck to the CircuitBreakerError enumeration

## Checklist

### DIFF
--- a/core/bin/contract-verifier/src/main.rs
+++ b/core/bin/contract-verifier/src/main.rs
@@ -28,9 +28,9 @@ async fn update_compiler_versions(connection_pool: &ConnectionPool) {
         .unwrap()
         .filter_map(|file| {
             let file = file.unwrap();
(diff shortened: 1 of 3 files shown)


montekki:
Thank you for your contribution, unfortunately I am going to have to close it because it kind of breaks the way we are doing error handling currently. TLDR we want our services to go down fast on irrecoverable errors such as database going down and `.unwraps()` currently do exactly this for us. In the future we plan to improve this by cleaner shutdowns.

PR #913, opened 2024-01-22
feat: made external node tests use a snapshot of the main node db

pompon0:
## What ❔

Made external node tests use a snapshot of the main node db.
It is implemented by postgres dump & restore in "zk db setup" call.

## Why ❔

Current implementation of the gossip network-based block syncing for external node needs to know the certificate on the consensus genesis block (which might not be the first miniblock in the db) to be able to verify all subsequent blocks. 
This certificate cannot be generated by the external node, because it contains signature of the validator (aka main node). For simplicity, we assume that this certificate on genesis block is already present in the external node's db, which has been restored from a snapshot of the main node db.

Integration tests for external node start with an empty db instead, which disables fetching blocks over gossip network altogether. This PR is supposed to fix that.


etc/env/ext-node.toml
```diff
@@ -2,8 +2,9 @@
 # All the variables must be provided explicitly.
 # This is on purpose: if EN will accidentally depend on the main node env, it may cause problems.
 
-test_database_url = "postgres://postgres@localhost:5433/zksync_local_test_ext_node"
 database_url = "postgres://postgres@localhost/zksync_local_ext_node"
+template_database_url="postgres://postgres@localhost/zksync_local"
```
popzxc, line 6: Probably it makes sense to add a comment on this variable, e.g. "Optional variable. If set, `zk` tool will use this URL to dump the database contents. Not used by EN itself".

infrastructure/zk/src/database.ts
```diff
@@ -53,10 +53,20 @@ export async function setup() {
         // Remote database, we can't show the contents.
         console.log(`WARNING! Using prod db!`);
     }
-    await utils.spawn('cargo sqlx database create');
-    await utils.spawn('cargo sqlx migrate run');
-    if (process.env.DATABASE_URL!.startsWith(localDbUrl)) {
-        await utils.spawn('cargo sqlx prepare --check -- --tests || cargo sqlx prepare -- --tests');
+    if (process.env.TEMPLATE_DATABASE_URL !== undefined) {
+        // Dump and restore from template database (simulate backup)
```
popzxc, line 57: Maybe it makes sense to add something like
```js
console.log(`Template DB URL provided. Creating a DB via dump from ${process.env.TEMPLATE_DATABASE_URL}`);
```

popzxc (approved):

etc/env/ext-node.toml
pompon0: done

infrastructure/zk/src/database.ts
pompon0: done

popzxc (approved):

PR #916, opened 2024-01-22
feat: added unauthenticated version of gcs object store

tomg10:
Adding unauthenticated version of GCS object store.
It's needed for EN to be able to download snapshots from our GCS


tomg10:
One thing I'm concerned a bit is that I'm not aware of any good ways to unit-test this.

core/lib/object_store/src/gcs_unauthenticated.rs
```diff
@@ -0,0 +1,99 @@
...
+                Err(error) => {
+                    last_error = Some(error);
+                }
+            }
+
+            tracing::warn!(
+                "Failed to download {url} (attempt {}/{retry_count}). Backing off for 5 second",
+                retry_number + 1
+            );
+            tokio::time::sleep(Duration::from_secs(5)).await;
```
slowli, line 72: Nit: You can extract the timeout into a constant and use it in the log message above.

core/lib/object_store/src/gcs_unauthenticated.rs
```diff
@@ -0,0 +1,99 @@
...
+        )))
+    }
+
+    async fn put_raw(
+        &self,
+        _bucket: Bucket,
+        _key: &str,
+        _value: Vec<u8>,
+    ) -> Result<(), ObjectStoreError> {
+        unimplemented!("This store is read-only!")
```
slowli, line 85: Bikeshedding: I'd return an error from this method and `remove_raw` since we cannot guarantee at compile-time that these methods won't be called, and dealing with panics is more costly / difficult than dealing with errors. Ideally, it could make sense to move `get_raw` to a new supertrait for `ObjectStore` (`ReadonlyObjectStore`?) and use this trait in snapshot recovery logic, so that it *can* be guaranteed in compile-time. But the downside of this is that we won't be able to configure such a store using `ObjectStoreFactory` (or at least it'd require more effort), so I'm not sure it's worth the trouble.

tomg10: I see now, that it may lead to runtime error, I don't think it's worth it to enforce this on type level as probably for a long time this could be the only read-only implementation. I have one idea though, I can put ReadOnly into the name of the type: PublicReadOnlyGoogleCloudStorage, it should make it way more difficult to miss

slowli: IMO, it's not really necessary given how public buckets work in GCS, S3 etc.; they imply public readonly access, not full read-write permissions. But I'm OK with this long name as well 🙃

core/lib/object_store/src/gcs_public.rs
```diff
@@ -0,0 +1,163 @@
...
+    backoff_seconds: u64,
+}
+
+impl PublicReadOnlyGoogleCloudStorage {
+    pub fn new(
+        gcs_base_url: String,
+        bucket_prefix: String,
+        max_retries: u16,
+        timeout: u64,
+        backoff_seconds: u64,
```
slowli, line 25: Nit: It generally makes sense to supply `Duration`s in such cases, so that they are more configurable, have more conventional `Debug` representation etc.

core/lib/object_store/src/gcs_public.rs
```diff
@@ -0,0 +1,163 @@
...
+        }
+    }
+}
+
+#[async_trait]
+impl ObjectStore for PublicReadOnlyGoogleCloudStorage {
+    async fn get_raw(&self, bucket: Bucket, key: &str) -> Result<Vec<u8>, ObjectStoreError> {
+        let mut last_error: Option<anyhow::Error> = None;
+        let retry_count = self.max_retries;
+        for retry_number in 0..retry_count {
```
slowli, line 46: Does `retry_count = 0` means an instant panic? I'd suggest to check `max_retries` is not equal to zero in the constructor then. Alternatively, `max_retries = 0` may be defined to mean "a single attempt"; note that an adjacent `GoogleCloudStorage` implementation uses this definition, so it would be more consistent to use it here as well.

slowli (approved):

PR #298, opened 2023-10-24
feat!: shared bridge

kelemeno:
# What ❔
* Adding support for shared bridge (bridgehub, state transition manager etc).

## Why ❔
* Adds a way to combine the liquidity from multiple chains.

PR #927, opened 2024-01-23
chore: preparation for shared bridge migration (server)

ly0va:
## What ❔

This PR aims to reduce the diff between main and #298 to simplify the future review.

The changes do not affect current server behavior.
Only changes related to the zkSync server went into this PR.

What didn't go into this PR:
- various variable renamings
- integration tests
- changes that depend on the future version of era-contracts

Where applicable, stubs and dummy (for now) config variables were introduced, which can be replaced/populated later.

## Why ❔

Breaking up a PR into smaller chunks should be easier to review.

## Checklist


mm-zk:
@ly0va - let's change the name of the PR, to make it more understandble in the long term (for example, preparation for shared bridge migration etc).

core/lib/config/src/configs/contracts.rs
```diff
@@ -14,6 +14,13 @@ pub enum ProverAtGenesis {
 pub struct ContractsConfig {
     pub governance_addr: Address,
     pub mailbox_facet_addr: Address,
+
+    pub bridgehub_proxy_addr: Address,
```
mm-zk, line 18: please add comments (maybe also with names of the interfaces that should be deployed there)

core/lib/config/src/configs/contracts.rs
```diff
@@ -49,6 +56,10 @@ impl ContractsConfig {
     /// Same goes for hashes.
     pub fn for_tests() -> Self {
         Self {
+            bridgehub_proxy_addr: Address::repeat_byte(0x01),
+            bridgehub_impl_addr: Address::repeat_byte(0x01),
+            state_transition_proxy_addr: Address::repeat_byte(0x01),
+            state_transition_impl_addr: Address::repeat_byte(0x01),
```
mm-zk, line 62: let's use a different address for each one (like we do in the cases below)

core/lib/dal/src/protocol_versions_dal.rs
```diff
@@ -98,6 +98,45 @@ impl ProtocolVersionsDal<'_, '_> {
         db_transaction.commit().await.unwrap();
     }
 
+    pub async fn save_genesis_upgrade(&mut self, id: ProtocolVersionId, tx_hash: Option<H256>) {
```
mm-zk, line 101: should this be public?

core/lib/dal/src/protocol_versions_dal.rs
```diff
@@ -98,6 +98,45 @@ impl ProtocolVersionsDal<'_, '_> {
...
+            "#,
+            tx_hash.map(|tx_hash| tx_hash.0.to_vec()),
+            id as i32,
+        )
+        .execute(self.storage.conn())
+        .await
+        .unwrap();
+    }
+
+    pub async fn save_genesis_upgrade_with_tx(
```
mm-zk, line 118: let's add a comment (and mention that it also injects the transaction in the database)

core/lib/env_config/src/chain.rs
```diff
@@ -125,6 +125,8 @@ mod tests {
             CHAIN_STATE_KEEPER_SAVE_CALL_TRACES="false"
             CHAIN_STATE_KEEPER_UPLOAD_WITNESS_INPUTS_TO_GCS="false"
             CHAIN_STATE_KEEPER_ENUM_INDEX_MIGRATION_CHUNK_SIZE="2000"
+            CHAIN_STATE_KEEPER_VIRTUAL_BLOCKS_PER_MINIBLOCK="1"
```
mm-zk, line 128: surprising - what's the reason for this to be in this PR?

core/lib/env_config/src/contracts.rs
```diff
@@ -19,6 +19,10 @@ mod tests {
 
     fn expected_config() -> ContractsConfig {
         ContractsConfig {
+            bridgehub_proxy_addr: addr("d8dA6BF26964aF9D7eEd9e03E53415D37aA96045"),
```
mm-zk, line 22: use different values

core/lib/zksync_core/src/api_server/web3/namespaces/zks.rs
```diff
@@ -123,6 +123,11 @@ impl ZksNamespace {
             .map_err(|err| err.into_web3_error(method_name))
     }
 
+    #[tracing::instrument(skip(self))]
+    pub fn get_bridgehub_contract_impl(&self) -> Address {
```
mm-zk, line 127: I wonder if this should be `Optional<Address>`

core/lib/zksync_core/src/eth_watch/client.rs
```diff
@@ -57,6 +59,7 @@ pub struct EthHttpQueryClient {
     client: Box<dyn EthInterface>,
     topics: Vec<H256>,
     zksync_contract_addr: Address,
+    state_transition_manager_contract_addr: Address,
```
mm-zk, line 62: comment. What interface will this address support?

core/lib/zksync_core/src/api_server/execution_sandbox/mod.rs
```diff
@@ -284,7 +284,7 @@ pub(crate) struct BlockArgs {
 }
 
 impl BlockArgs {
-    pub(crate) async fn pending(connection: &mut StorageProcessor<'_>) -> anyhow::Result<Self> {
+    pub async fn pending(connection: &mut StorageProcessor<'_>) -> anyhow::Result<Self> {
```
mm-zk, line 287: if you make it public - add the comment

core/lib/zksync_core/src/eth_watch/event_processors/set_chain_id.rs
```diff
@@ -0,0 +1,98 @@
...
+pub struct SetChainIDEventProcessor {
+    diamond_proxy_address: Address,
+    set_chain_id_signature: H256,
+}
+
+impl SetChainIDEventProcessor {
+    pub fn new(diamond_proxy_address: Address) -> Self {
+        Self {
+            diamond_proxy_address,
+            set_chain_id_signature: H256::default(),
```
mm-zk, line 21: we should know the signature already, right?

mm-zk (changes requested):

core/lib/zksync_core/src/state_keeper/keeper.rs
```diff
@@ -136,25 +139,50 @@ impl ZkSyncStateKeeper {
...
-        } else if !pending_miniblocks.is_empty() && version_changed {
-            // Sanity check: if `txs_to_reexecute` is not empty and upgrade tx is present for this block
-            // then it must be the first one in `txs_to_reexecute`.
-            if self.io.load_upgrade_tx(protocol_version).await.is_some() {
-                let first_tx_to_reexecute = &pending_miniblocks[0].txs[0];
-                assert_eq!(
-                    first_tx_to_reexecute.tx_format(),
-                    TransactionType::ProtocolUpgradeTransaction
-                )
+        let mut protocol_upgrade_tx = self.io.load_upgrade_tx(protocol_version).await;
```
mm-zk, line 147: we're adding too much logic to this 'run_inner' function.

I'd like us to:
* move this logic to a separate method
* add unittests for this method.

core/lib/zksync_core/src/eth_watch/event_processors/set_chain_id.rs
```diff
@@ -0,0 +1,98 @@
...
+            let new_version = previous_version.apply_upgrade(upgrade, scheduler_vk_hash);
+
+            if let Some(tx) = new_version.tx.clone() {
+                storage
+                    .transactions_dal()
+                    .insert_system_transaction(tx.clone())
+                    .await;
+                storage
+                    .protocol_versions_dal()
+                    .save_genesis_upgrade_with_tx(version_id, tx)
```
perekopskiy, line 87: `insert_system_transaction` is called twice: a few lines above and inside `save_genesis_upgrade_with_tx`, I think it's not intended

core/lib/zksync_core/src/state_keeper/keeper.rs
```diff
@@ -136,25 +139,50 @@ impl ZkSyncStateKeeper {
 
         let previous_batch_protocol_version =
             self.io.load_previous_batch_version_id().await.unwrap();
+
+        let first_batch_in_shared_bridge = l1_batch_env.number == L1BatchNumber(1)
+            && protocol_version >= ProtocolVersionId::Version21;
```
perekopskiy, line 144: Let's add a method for `ProtocolVersionId` called `is_post_shared_bridge` or something and let it return false for now as we don't know if we would need to do another upgrade before shared bridge.

core/lib/zksync_core/src/eth_watch/event_processors/set_chain_id.rs
```diff
@@ -0,0 +1,98 @@
...
+impl SetChainIDEventProcessor {
+    pub fn new(diamond_proxy_address: Address) -> Self {
+        Self {
+            diamond_proxy_address,
+            set_chain_id_signature: H256::default(),
+            // TODO: uncomment when we merge the Shared Bridge contracts.
+            // state_transition_manager_contract()
+            //     .event("SetChainIdUpgrade")
+            //     .expect("SetChainIdUpgrade event is missing in ABI")
+            //     .signature(),
```
perekopskiy, line 26: I think you can hardcode event's ABI in core/lib/contracts and use it here.

core/lib/zksync_core/src/eth_watch/event_processors/set_chain_id.rs
```diff
@@ -0,0 +1,98 @@
...
+        });
+
+        // SetChainId does not go through the governance contract, so we need to parse it separately.
+        for event in filtered_events {
+            let timestamp = client
+                .get_block(event.block_hash.expect("event without block_hash"))
+                .await?
+                .expect("event's block not found")
+                .timestamp;
+            let upgrade = ProtocolUpgrade::decode_set_chain_id_event(event, timestamp.as_u64())
```
perekopskiy, line 54: It seems you don't need whole `ProtocolUpgrade`. I think `decode_set_chain_id_event` should return only `ProtocolVersionId` and upgrade tx. Then won't need nor querying timestamp, nor `get_block` method in client interface

core/lib/zksync_core/src/eth_watch/event_processors/set_chain_id.rs
```diff
@@ -0,0 +1,98 @@
...
+
+use crate::eth_watch::{
+    client::{Error, EthClient},
+    event_processors::EventProcessor,
+    metrics::{PollStage, METRICS},
+};
+
+/// Responsible for saving new protocol upgrade proposals to the database.
+#[derive(Debug)]
+pub struct SetChainIDEventProcessor {
```
perekopskiy, line 12: Please add unit tests for this processor

core/lib/env_config/src/chain.rs
ly0va: Yeah I was also surprised, but this test was failing for me locally. And if you look closely, these variables were indeed missing from the complete config. :shrug:

core/lib/zksync_core/src/eth_watch/event_processors/set_chain_id.rs
ly0va: Yes but I didn't want to hardcode it, since it already changed twice only in the last 2 weeks

core/lib/zksync_core/src/state_keeper/keeper.rs
```diff
@@ -136,25 +139,50 @@ impl ZkSyncStateKeeper {
 
         let previous_batch_protocol_version =
             self.io.load_previous_batch_version_id().await.unwrap();
+
+        let first_batch_in_shared_bridge = l1_batch_env.number == L1BatchNumber(1)
```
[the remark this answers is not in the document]
ly0va: there is already a comment about it a few lines below

core/lib/zksync_core/src/eth_watch/event_processors/set_chain_id.rs
```diff
@@ -0,0 +1,98 @@
...
+
+#[async_trait::async_trait]
+impl EventProcessor for SetChainIDEventProcessor {
+    async fn process_events(
+        &mut self,
+        storage: &mut StorageProcessor<'_>,
+        client: &dyn EthClient,
+        events: Vec<Log>,
+    ) -> Result<(), Error> {
+        let mut upgrades = Vec::new();
```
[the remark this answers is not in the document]
ly0va: we could if we didn't have to await a future inside of that closure

core/lib/zksync_core/src/eth_watch/event_processors/set_chain_id.rs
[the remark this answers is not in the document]
ly0va: yes that's fine, they are idempotent

core/lib/zksync_core/src/eth_watch/event_processors/set_chain_id.rs
```diff
@@ -0,0 +1,98 @@
...
+use zksync_types::{web3::types::Log, Address, ProtocolUpgrade, H256};
+
+use crate::eth_watch::{
+    client::{Error, EthClient},
+    event_processors::EventProcessor,
+    metrics::{PollStage, METRICS},
+};
+
+/// Responsible for saving new protocol upgrade proposals to the database.
+#[derive(Debug)]
```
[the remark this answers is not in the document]
ly0va: no, not really. I'll add a comment

core/lib/zksync_core/src/state_keeper/keeper.rs
```diff
@@ -223,9 +214,10 @@ impl ZkSyncStateKeeper {
                 .await
                 .ok_or(Error::Canceled)?;
 
-            let version_changed = system_env.version != sealed_batch_protocol_version;
+            let version_changed_or_first_batch =
+                system_env.version != sealed_batch_protocol_version || first_batch_in_shared_bridge;
```
mm-zk, line 218: and were do we mark first_batch_in_shared_bridge to false?

core/lib/zksync_core/src/state_keeper/keeper.rs
```diff
@@ -234,6 +226,59 @@ impl ZkSyncStateKeeper {
         Err(Error::Canceled)
     }
 
+    async fn load_protocol_upgrade_tx(
```
mm-zk, line 229: I wonder how this differs from self.io.load_upgrade_tx..

core/lib/zksync_core/src/state_keeper/keeper.rs
```diff
@@ -234,6 +226,59 @@ impl ZkSyncStateKeeper {
...
+        &mut self,
+        pending_miniblocks: &[MiniblockExecutionData],
+        protocol_version: ProtocolVersionId,
+        first_batch_in_shared_bridge: bool,
```
mm-zk, line 233: I don't like that this upgrade logic is spread in 3 places (one here, one in line 140 - before the loop, and one in 220 - at the end of the loop).

could we instead move the logic to this function, and remove stuff from line 220 and 140?

core/lib/zksync_core/src/state_keeper/keeper.rs
```diff
@@ -234,6 +226,59 @@ impl ZkSyncStateKeeper {
...
+            if self.is_canceled() {
+                return Err(Error::Canceled);
+            }
+            tokio::time::sleep(Duration::from_secs(1)).await;
+            protocol_upgrade_tx = self.io.load_upgrade_tx(protocol_version).await;
+        }
+
+        let protocol_upgrade_tx =
+            if pending_miniblocks.is_empty() && (version_changed || first_batch_in_shared_bridge) {
+                tracing::info!("We have upgrade tx to be executed in empty miniblock");
```
mm-zk, line 258: this log might not be correct (if version was changed, but for some reason the protocoL_upgrade_tx is still 'none')

core/lib/zksync_core/src/state_keeper/keeper.rs
```diff
@@ -234,6 +226,59 @@ impl ZkSyncStateKeeper {
...
+                && (version_changed || first_batch_in_shared_bridge)
+            {
+                // Sanity check: if `txs_to_reexecute` is not empty and upgrade tx is present for this block
+                // then it must be the first one in `txs_to_reexecute`.
+                if protocol_upgrade_tx.is_some() {
+                    let first_tx_to_reexecute = &pending_miniblocks[0].txs[0];
+                    assert_eq!(
+                        first_tx_to_reexecute.tx_format(),
+                        TransactionType::ProtocolUpgradeTransaction
+                    )
```
mm-zk, line 270: if we create assert like this - let's add more information here to make debugging easier. (for example, the transaction hash of the first transaction etc)

core/lib/types/src/protocol_version.rs
```diff
@@ -93,6 +93,11 @@ impl ProtocolVersionId {
         self <= &Self::Version17
     }
 
+    pub fn is_pre_shared_bridge(&self) -> bool {
+        // TODO: review this when we actually deploy shared bridge
+        self <= &Self::Version21
```
mm-zk, line 98: alternatively - for now just return true.

core/lib/zksync_core/src/eth_sender/eth_tx_aggregator.rs
```diff
@@ -403,50 +387,67 @@ impl EthTxAggregator {
...
                     self.functions
-                        .post_boojum_commit
+                        .pre_shared_bridge_commit
+                        .encode_input(&op.into_tokens())
+                        .expect("Failed to encode commit transaction data")
+                } else {
+                    let func = self
+                        .functions
+                        .post_shared_bridge_commit
                         .as_ref()
```
mm-zk, line 411: I'm surprised that this is differnt than the pre_shared_bridge flow (where we call 'encode_input' directly. Any reason for that?

core/lib/zksync_core/src/eth_watch/tests.rs
```diff
@@ -648,31 +758,35 @@ fn upgrade_into_governor_log(upgrade: ProtocolUpgrade, eth_block: u64) -> Log {
     }
 }
 
+fn tx_into_token(tx: ProtocolUpgradeTx) -> Token {
+    Token::Tuple(vec![
```
mm-zk, line 762: if we have such encodings - let's also include the link (or the name) of the solidity object (so that reader can easily see/verify )

mm-zk (changes requested):

core/lib/zksync_core/src/eth_sender/eth_tx_aggregator.rs
ly0va: The post shared bridge functions require an extra argument - the chain id.

core/lib/zksync_core/src/state_keeper/keeper.rs
ly0va: In this case it cannot be `None`. If `first_batch_in_shared_bridge` is `true`, we waited for it to be `Some` in the while-loop before. If `version_changed` is true, then `eth_watch` saved the upgrade transaction together with the new protocol version. But we can put an additional assert there just to be sure.

core/lib/zksync_core/src/state_keeper/keeper.rs
ly0va: on line 220 we're calling `self.io.load_upgrade_tx` in a loop - since the state keeper has to process upgrade transactions routinely, not only at startup. Therefore since it does not have to perform additional checks or waiting loops, calling `self.load_protocol_upgrade_tx` is not really appropriate.

I don't see how it's possible to incorporate this logic into the existing function.

As for line 140, would you propose passing the L1BatchNumber into the function instead of `first_batch_in_shared_bridge`?

mm-zk: yes, we could put l1 batch number  (as argument here).

core/lib/zksync_core/src/state_keeper/keeper.rs
```diff
@@ -234,6 +224,67 @@ impl ZkSyncStateKeeper {
         Err(Error::Canceled)
     }
 
+    pub(super) async fn load_protocol_upgrade_tx(
```
mm-zk, line 227: add a comment and metion that this method is called only once on init.

mm-zk (approved):

kelemeno (commented):
Stamping, Marcin approved previously

kelemeno (approved):

PR #928, opened 2024-01-24
DRAFT: multiVM to depend on the new geometry crate

mm-zk:
## What ❔

Changing multiVM to depend on the newly crated geometry_config crate.

## Why ❔

This crate is a lot smaller than the evm_test_harness one, and multiVM requires just the geometry information.


joonazan:
I see that the separate geometry config crate didn't happen. Why is that? Should we close this PR too?

PR #860, opened 2024-01-11
chore(vm): use internal types for log query

StanislavBreadless:
## What ❔

Reducing the number of `glue_into` in the multivm crate. It should improve performance and general refactoring. 

**Notes**

- The number of commits is large, since this PR used to be based on the 1.4.1 integration branch & it was squashed when merged to main.
- There are some optimizations that can be done (for instance, ensure that zkevm_test_harness @1.4.0 and @1.4.1 accept iterator instead of slice: https://github.com/matter-labs/era-zkevm_test_harness/pull/68

## Checklist

- [ ] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [ ] Tests for the changes have been added / updated.
- [ ] Documentation comments have been added / updated.
- [ ] Code has been formatted via `zk fmt` and `zk lint`.
- [ ] Spellcheck has been run via `zk spellcheck`.

PR #935, opened 2024-01-24
refactor: Remove zk_evm deps from zksync_types

popzxc:
## What ❔

- Moves zkevm glue from zksync_types to multivm

## Why ❔

- Glue is used by the multivm
- Multivm already has these deps
- zksync_types doesn't need these deps
- A lot of crates depend on zksync_types but also don't need these deps
- We need to think about our dependency graphs more.

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8906,8 +8906,6 @@ dependencies = [
  "thiserror",
  "tokio",
  "zk_evm 1.3.3 (git+https://github.com/matter-labs/era-zk_evm.git?tag=v1.3.3-rc2)",
- "zk_evm 1.4.0",
- "zk_evm 1.4.1",
  "zkevm_test_harness 1.3.3",
  "zksync_basic_types",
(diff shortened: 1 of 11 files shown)


Cargo.lock
```diff
@@ -8906,8 +8906,6 @@ dependencies = [
  "thiserror",
  "tokio",
  "zk_evm 1.3.3 (git+https://github.com/matter-labs/era-zk_evm.git?tag=v1.3.3-rc2)",
```
Deniallugo, line 8908: Is it necessary to keep 1.3.3? 
@StanislavBreadless you are also working on something like it

popzxc: I've raised this topic before, it's out of scope for this PR.

mm-zk (approved):

popzxc:
Closing in favor of https://github.com/matter-labs/zksync-era/pull/860

PR #939, opened 2024-01-24
chore(vm_utils): Make create vm more generic

Deniallugo:
## What ❔

Add more flexibility for creating vm in vm utils

## Why ❔

For future features, it's required to execute vm not only for the latest miniblock 

## Checklist


core/lib/state/src/postgres/mod.rs
```diff
@@ -393,6 +393,11 @@ impl<'a> PostgresStorage<'a> {
     fn values_cache(&self) -> Option<&ValuesCache> {
         Some(&self.caches.as_ref()?.values.as_ref()?.cache)
     }
+
+    /// Get the connection to the database.
+    pub fn connection(&mut self) -> &mut StorageProcessor<'a> {
+        &mut self.connection
+    }
```
popzxc, line 400: PostgresStorage assumes a unique ownership of the connection. I don't think we want to expose the underlying connection (generally, exposing something internal while there is a way to obtain the same resource otherwise is usually not a good idea).

core/lib/vm_utils/src/lib.rs
```diff
@@ -94,3 +43,36 @@ pub fn execute_tx<S: WriteStorage>(
...
+                "l1_batch_number {l1_batch_number:?} must have a previous miniblock to start from"
+            )
+        })?;
+
+    let mut pg_storage =
+        PostgresStorage::new(rt_handle.clone(), connection, miniblock_number, true);
+
+    let vm_env = VmEnvBuilder::new(l1_batch_number, u32::MAX, l2_chain_id)
+        .with_miniblock_number(miniblock_number)
+        .build(pg_storage.connection())
```
popzxc, line 69: You have the connection already, so you can just swap the lines so that you create `vm_env` first, and then create `pg_storage`. This way you don't need to expose `connection` method on the `PostgresStorage`.

core/lib/vm_utils/src/vm_env.rs
```diff
@@ -0,0 +1,183 @@
...
+    vm_latest::constants::BLOCK_GAS_LIMIT,
+};
+use zksync_contracts::BaseSystemContracts;
+use zksync_dal::StorageProcessor;
+use zksync_types::{
+    fee_model::BatchFeeInput, Address, L1BatchNumber, L2ChainId, MiniblockNumber,
+    ProtocolVersionId, H256, ZKPORTER_IS_AVAILABLE,
+};
+
+/// The environment required to initialize the VM.
```
popzxc, line 12: Please `#[derive(Debug)]` here and in other places.

slowli:
@Deniallugo JFYI, the changes in the `vm_utils` crate [significantly intersect](https://github.com/matter-labs/zksync-era/blob/168efbaa6c909992d74fa44be0597aadf9d73cd2/core/lib/vm_utils/src/storage.rs) with [my PR](https://github.com/matter-labs/zksync-era/pull/900) making the state keeper (and in particular VM batch executor) aware of snapshot recovery.

Deniallugo:
@slowli how can we merge it together? should we merge my version as is? for now? are you satisfied with the outcome?

slowli:
@Deniallugo I don't want to block you, so merging your PR as is could be the easiest option. I can't say I understand some of the changes made (since I don't understand the intended scope of the `vm_utils` crate), and it looks like I'll have to roll some of those back with #900 (e.g., moving `wait_for_l1_batch_params` from/to the crate).

core/lib/zksync_core/src/state_keeper/io/common.rs
```diff
@@ -75,13 +23,18 @@ pub(crate) async fn load_pending_batch(
     validation_computational_gas_limit: u32,
     chain_id: L2ChainId,
 ) -> Option<PendingBatchData> {
-    let (system_env, l1_batch_env) = load_l1_batch_params(
-        storage,
+    let previous_l1_batch_hash = wait_for_prev_l1_batch_params(storage, current_l1_batch_number)
```
slowli, line 26: IMO, this construction is unnecessarily brittle; nowhere does it guarantee that the hash supplied to `with_prev_batch_hash` won't be mismatched (e.g., by loading a hash for a different L1 batch).

core/lib/vm_utils/src/vm_env.rs
```diff
@@ -0,0 +1,185 @@
...
+        Self {
+            l1_batch_number,
+            validation_computational_gas_limit,
+            chain_id,
+            miniblock_number: None,
+            prev_batch_hash: None,
+            fee_account: None,
+        }
+    }
+    pub fn with_miniblock_number(mut self, miniblock_number: MiniblockNumber) -> Self {
```
slowli, line 45: Same problem as with `with_prev_batch_hash` here; unnecessarily brittle. In this particular case, you could access `miniblock_number ` from `L1BatchEnv` in `create_vm_for_l1_batch` (IIUC, that's the only use of the method).

core/lib/vm_utils/src/vm_env.rs
```diff
@@ -0,0 +1,185 @@
...
+        self.fee_account = Some(fee_account);
+        self
+    }
+
+    pub fn with_prev_batch_hash(mut self, prev_batch_hash: H256) -> Self {
+        self.prev_batch_hash = Some(prev_batch_hash);
+        self
+    }
+
+    pub async fn build(self, connection: &mut StorageProcessor<'_>) -> Option<VmEnv> {
```
slowli, line 60: Nit: It could make sense to propagate errors / add contexts for errors here instead of `unwrap()`ping them.

core/lib/zksync_core/src/state_keeper/io/common.rs
Deniallugo: Yeah, but sometimes it's easier to get it somewhere outside. 
And it's users responsibility to get it correct

core/lib/vm_utils/src/vm_env.rs
Deniallugo: It will be more use, where i need to have access to the different miniblock.

core/lib/vm_utils/src/vm_env.rs
Deniallugo: Yeah, probably and we could also return some kind of error if we found some discrepancy

core/lib/vm_utils/src/vm_env.rs
Deniallugo: but at the same time we can use this method for creating wrong data intentionaly.

popzxc (commented):
We've discussed this PR a bit in DMs with @Deniallugo , and in short: I agree with @slowli that the proposed interface for VM is too fragile.

Currently, VM exposes a lot of APIs that can lead to incorrect configurations and this new interface _appears_ to be safe, but in fact still enables these incorrect configurations, and even makes them more likely due to automatic substitutions it does.

I'm not sure if we should merge it in the current form. If we want a generic high-level interface for the VM, I'd much prefer it to decline invalid configurations.

core/lib/vm_utils/src/vm_env.rs
```diff
@@ -0,0 +1,243 @@
...
+            .0;
+
+        let (
+            first_miniblock_number_included_in_l1_batch,
+            last_miniblock_number_included_in_l1_batch,
+        ) = connection
+            .blocks_dal()
+            .get_miniblock_range_of_l1_batch(self.l1_batch_number)
+            .await?
+            .context(format!("L1 batch {} not found", self.l1_batch_number))?;
```
popzxc, line 81: Nit: [`with_context`](https://docs.rs/anyhow/latest/anyhow/trait.Context.html#tymethod.with_context) would be more appropriate here.

core/lib/vm_utils/src/vm_env.rs
```diff
@@ -0,0 +1,243 @@
...
+
+        self.miniblock_number = if let Some(pending_miniblock_number) = self.miniblock_number {
+            if first_miniblock_number_included_in_l1_batch >= pending_miniblock_number
+                || pending_miniblock_number >= last_miniblock_number_included_in_l1_batch
+            {
+                anyhow::bail!(
+                    "Miniblock {} is not included in L1 batch {}",
+                    pending_miniblock_number,
+                    self.l1_batch_number - 1
```
popzxc, line 90: I think there shouldn't be `- 1` here.

core/lib/vm_utils/src/vm_env.rs
```diff
@@ -0,0 +1,243 @@
...
+        self.build(connection).await
+    }
+
+    /// Builds the `VmEnv` for the pending L1 batch. Please note, that this method is waiting for
+    /// the previous L1 batch to be sealed.
+    pub async fn build_for_pending_pending_batch(
+        mut self,
+        connection: &mut StorageProcessor<'_>,
+    ) -> anyhow::Result<VmEnv> {
+        let batch_hash = wait_for_prev_l1_batch_params(connection, self.l1_batch_number - 1)
```
popzxc, line 107: What I don't like right now is that this method may hang if this info is not returned. What we can do is make user provide a _getter_ for the previous block hash, something like:

```rust
pub async fn build_for_pending_batch<F, Fut>(mut self, get_prev_batch_hash: F, connection: &mut StorageProcessor<'_>)
where
  F: FnOnce(&mut StorageProcessor<'_>, L1BatchNumber> -> Fut,
  Fut: Future<Output = anyhow::Result<H256>>
{
  let prev_batch_hash = get_prev_batch_hash(connection, self.l1_batch_number - 1).await?;
  ...
}
```

This way we still guarantee that the data is aligned (e.g. we use the same `l1_batch_number` everywhere, and the only thing user should worry about is to provide a correct getter function. 

WDYT?

core/lib/vm_utils/src/vm_env.rs
```diff
@@ -0,0 +1,243 @@
...
+        } else {
+            Some(first_miniblock_number_included_in_l1_batch)
+        };
+        self.prev_batch_hash = Some(batch_hash);
+        self.build(connection).await
+    }
+
+    /// Builds the `VmEnv` for the pending L1 batch. Please note, that this method is waiting for
+    /// the previous L1 batch to be sealed.
+    pub async fn build_for_pending_pending_batch(
```
popzxc, line 103: Nit: `s/pending_pending/pending`

core/lib/vm_utils/src/vm_env.rs
```diff
@@ -0,0 +1,243 @@
...
+            }
+            Some(pending_miniblock_number)
+        } else {
+            Some(first_miniblock_number_included_in_l1_batch)
+        };
+        self.prev_batch_hash = Some(batch_hash);
+        self.build(connection).await
+    }
+
+    /// Builds the `VmEnv` for the pending L1 batch. Please note, that this method is waiting for
```
popzxc, line 101: I'm still not sure if I understand the semantics here.
This method says that it's for pending batch, but it allows you to override the miniblock number. What would the behavior be if we do so? Will we apply all the previous miniblocks stored inside, or will we just change the number (thus creating a gap here)?

popzxc (commented):
This looks better than yesterday, but I still have some problems.
The main one is that we allow choosing some miniblock from the corresponding l1 batch range, but it doesn't look like that we also provide any guarantees that the state will actually be reexecuted.
Do we need to provide ability to override miniblock at all then? What if miniblock is overridden, but the state is not replayed till this miniblock?

core/lib/vm_utils/src/vm_env.rs
Deniallugo: that's why we have  a protection inside

PR #944, opened 2024-01-25
refactor: Remove zk_evm deps from zksync_types

popzxc:
## What ❔

- Remove `zk_evm_xxx` deps from `zksync_types`

## Why ❔

- These deps only needed for a few use cases, so these use cases are implemented elsewhere. Part of initiative of moving heavyweight stuff from `zksync_types`.

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8261,6 +8261,8 @@ name = "zksync_commitment_utils"
 version = "0.1.0"
 dependencies = [
  "multivm",
+ "zk_evm 1.3.3 (git+https://github.com/matter-labs/era-zk_evm.git?tag=v1.3.3-rc2)",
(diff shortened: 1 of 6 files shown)


StanislavBreadless (commented):
Do you think that importing `multivm` in commitment utils would be better? For me, it seems a bit aesthetically nicer if all major conversions happen in one `multivm` crate, but it is obviously opinionated

popzxc:
@StanislavBreadless I'll most likely remove `commitment_utils` crate soon (and there is a good chance that it'll become a part of `multivm`), so probably it isn't important.

Deniallugo (commented):
❤️

Deniallugo (approved):

Issue #955, opened 2024-01-27
bug: meet bug when build docker images


ChengYueJia:
#### 📝 Description
Meet error when try to build docker image.

#### 🔄 Reproduction Steps
Two ways to reproduction it: 
> PWD: project_root_dir
1. `docker build -f ./docker/server-v2/Dockerfile .` 
2. `zk stack docker-setup` 
      will meet error when execute first image, the command as below: 
     * `DOCKER_BUILDKIT=1 docker buildx build -t us-docker.pkg.dev/matterlabs-infra/matterlabs-docker/server-v2:docker-linux-amd64 -t matterlabs/server-v2:doker-linux-amd64  --platform=linux/amd64  -f ./docker/server-v2/Dockerfile .`

#### 🤔 Expected Behavior
Successfully build images using both ways mentioned above.

#### 😯 Current Behavior
will meet error 
#### 🖥️ Environment
22.04.1-Ubuntu  x86_64

#### 📋 Additional Context

Add any other context about the problem here. If applicable, add screenshots to help explain.

#### 📎 Log Output
```bash
$ docker build -f ./docker/server-v2/Dockerfile .
[+] Building 1.9s (7/23)                                                                                                                                docker:default
 => [internal] load .dockerignore                                                                                                                                 0.0s
 => => transferring context: 1.00kB                                                                                                                               0.0s
 => [internal] load build definition from Dockerfile                                                                                                              0.0s
 => => transferring dockerfile: 1.90kB                                                                                                                            0.0s
 => [internal] load metadata for docker.io/library/debian:bookworm-slim                                                                                           1.1s
 => CACHED [stage-1  1/14] FROM docker.io/library/debian:bookworm-slim@sha256:f4a83aa865a2b4a064ff142aa91c713180df9fcb86ce676b5de2981029379c37                    0.0s
 => CANCELED [internal] load build context                                                                                                                        0.6s
 => => transferring context: 90.29MB                                                                                                                              0.6s
 => CACHED [builder 2/6] WORKDIR /usr/src/zksync                                                                                                                  0.0s
 => ERROR [stage-1  2/14] RUN apt-get update && apt-get install -y curl libpq5 liburing-dev ca-certificates &&     rm -rf /var/lib/apt/lists/*                    0.6s
------                                                                                                                                                                 
[...]

PR #956, opened 2024-01-27
fix(docker): bugfix error when build docker image

ChengYueJia:
## What ❔

1. #955 
    Solution: https://stackoverflow.com/questions/53950082/docker-images-lacking-tmp
2. Add docs for setup for  `zkStack`

## Why ❔

Meet error when try to use `zk stack docker-setup`

## Checklist


hatemosphere:
@ChengYueJia this issue does not exist on any vanilla Ubuntu installation. thanks a lot for attempting to fix, but we can't support all the variants of Ubuntu configurations, including ones without /tmp dir existing by default

PR #966, opened 2024-01-30
fix: reorg check was performed on a too high block

pompon0:
## What ❔

reorg check at min(local head, remote head)

## Why ❔

So that divergence is determined as soon as it occurs, rather than waiting for the main node to produce enough miniblocks.


pompon0:
duplicate of https://github.com/matter-labs/zksync-era/pull/964

PR #967, opened 2024-01-30
refactor(en): Brush up snapshot applier code

slowli:
## What ❔

Brushes up the snapshot applier code according to nits from #882 code reviews.

## Why ❔

Improves snapshot applier maintainability.

## Checklist


core/lib/snapshots_applier/src/lib.rs
```diff
@@ -303,25 +278,27 @@ impl<'a, 'b> SnapshotsApplier<'a, 'b> {
         Ok(())
     }
 
+    #[tracing::instrument(level = "debug", err, skip(self))]
```
tomg10, line 281: why this line is needed?

core/lib/snapshots_applier/src/lib.rs
```diff
@@ -330,28 +307,22 @@ impl<'a, 'b> SnapshotsApplier<'a, 'b> {
             .connection_pool
             .access_storage_tagged("snapshots_applier")
             .await?;
-        let mut storage = storage.start_transaction().await?;
-
-        let storage_logs = &storage_snapshot_chunk.storage_logs;
+        let mut storage_transaction = storage.start_transaction().await?;
```
tomg10, line 310: I really like the convention of adding _transaction suffix to such variables!

core/lib/snapshots_applier/src/lib.rs
```diff
@@ -109,16 +74,27 @@ impl From<RpcError> for SnapshotsApplierError {
...
     async fn fetch_newest_snapshot(&self) -> Result<Option<SnapshotHeader>, RpcError>;
 }
-impl<'a, 'b> SnapshotsApplier<'a, 'b> {
-    pub async fn prepare_applied_snapshot_status(
+
+/// Applying application-level storage snapshots to the Postgres storage.
+#[derive(Debug)]
+pub struct SnapshotsApplier<'a> {
+    connection_pool: &'a ConnectionPool,
+    blob_store: &'a dyn ObjectStore,
```
tomg10, line 89: Lifetimes newbie question: I don't think I see, why it's better to have a single lifetime here

core/lib/snapshots_applier/src/lib.rs
```diff
@@ -109,16 +74,27 @@ impl From<RpcError> for SnapshotsApplierError {
     }
 }
 
+/// Main node API used by the [`SnapshotsApplier`].
```
tomg10, line 77: Nit: It's probably not worth to remove this comment, but does it add value? Maybe I've spent too much time in this part of the system, but for me the same set of info is encoded in the name of the type

core/lib/snapshots_applier/Cargo.toml
```diff
@@ -1,15 +1,14 @@
 [package]
 name = "zksync_snapshots_applier"
 version = "0.1.0"
-edition = "2018"
+edition = "2021"
```
tomg10, line 4: Out of curiosity, which features of 2021 edition did you need?

tomg10 (approved):
Looks very nice, thanks for doing this work!

core/lib/snapshots_applier/src/lib.rs
slowli: It's not strictly needed, but it's useful: It adds context (namely, a [span](https://docs.rs/tracing/latest/tracing/#spans) with `chunk_id` argument) to all logs emitted within the method (and all code that it calls). This allows to not repeat all context explicitly in all logs, so it's useful for debugging etc.

core/lib/snapshots_applier/src/lib.rs
slowli: It's a good question. Less lifetime params makes code easier to read / maintain, and in this case one lifetime is sufficient: because we use lifetimes for reference types in the same struct, their lifetimes are guaranteed to intersect (otherwise, we wouldn't be able to construct a stuct). This means that both refs can use this common lifetime `'a`; their original lifetimes will be automatically coerced to `'a` (in type theory terms, reference types are [covariant](https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science)) w.r.t. the lifetime).

As a counter-example, DAL types cannot use a single lifetime because they wrap `&'a mut StorageProcessor<'c>` which is invariant over `'c` (i.e., `'c` cannot be replaced with `'a`) because `&mut T` references are invariant over `T`.

tomg10 (approved):

PR #976, opened 2024-01-31
feat(core): adds eip4844 transactions support to eth-signer and eth-sender

montekki:
## What ❔

Adds the support of [EIP4844 transactions](https://eips.ethereum.org/EIPS/eip-4844) to the `eth-signer` and `eth-sender` components.

EIP4844 introduces blob transactions that expand the already existing EIP1559 transactions by adding the following information. The main purpose is sending the blobs of data to the network that will only temporarily remain available instead of forever staying in chain history. A single transaction may include multiple blobs.

* A new blob gas price field is added to the transaction introducing which introduces a new blob gas market to the protocol
* The versioned hashes of the blobs (these first two fields go into tx and remain on-chain forever)
* A "sidecar" addition info that includes the blobs themselves, the commitments and proofs of these blobs. This information is only temporarily kept around by CL.

The above mentioned EIP describes signing and network transmission format of these transactions which this PR implements.

## Checklist


core/lib/eth_signer/src/raw_ethereum_tx.rs
```diff
@@ -47,6 +48,10 @@ pub struct TransactionParameters {
     pub max_fee_per_gas: U256,
     /// miner bribe
     pub max_priority_fee_per_gas: U256,
+    /// Max fee per blob gas
+    pub max_fee_per_blob_gas: Option<U256>,
+    /// blob versioned hashes
+    pub blob_versioned_hashes: Option<Vec<H256>>,
```
montekki, line 54: @koloz193 q: do we compute these values once at tx creation? If so they should live in DB i suppose.

montekki:
@koloz193 the [testcases in the EIP4844 page](https://eips.ethereum.org/EIPS/eip-4844#test-cases) are not populated, we need to come up with a strategy to unit test this + integration test

core/lib/eth_signer/src/raw_ethereum_tx.rs
koloz193: yes, the versioned hashes correspond to the blobs being used and the blob gas should be whatever was used for the txn

koloz193:
> @koloz193 the [testcases in the EIP4844 page](https://eips.ethereum.org/EIPS/eip-4844#test-cases) are not populated, we need to come up with a strategy to unit test this + integration test

agreed, for integration tests we can assume that theyll all test using blobs if thats what the server is configured to use. maybe in the long run we may want a new CI step that runs all the tests where pubdata is used as calldata and where pubdata is used as blobs

core/lib/eth_signer/src/raw_ethereum_tx.rs
```diff
@@ -147,6 +155,42 @@ impl Transaction {
...
+
+        stream.append(&self.value);
+        stream.append(&self.data);
+
+        self.rlp_append_access_list(&mut stream);
+
+        stream.append(&self.max_fee_per_blob_gas.unwrap());
+        // TODO: is this the correct way to encode a vector of versioned hashes
+        // to RLP here?
+        stream.append_list(self.blob_versioned_hashes.as_ref().unwrap());
```
koloz193, line 185: i think this is right, i tested locally and under the hood it encodes the length and the header byte

core/lib/dal/src/models/storage_eth_tx.rs
```diff
@@ -22,6 +22,7 @@ pub struct StorageEthTx {
     pub updated_at: NaiveDateTime,
     // TODO (SMA-1614): remove the field
     pub sent_at_block: Option<i32>,
+    pub blob_sidecar: Option<Vec<u8>>,
```
mm-zk, line 25: Add a comment and explain how this data is generated (and that even if we have multiple blobs, they are still within a single sidecar)

core/lib/eth_client/src/clients/http/signing.rs
```diff
@@ -217,6 +217,8 @@ impl<S: EthereumSigner> BoundEthInterface for SigningClient<S> {
         data: Vec<u8>,
         contract_addr: H160,
         options: Options,
+        max_fee_per_blob_gas: Option<U256>,
```
mm-zk, line 220: shouldn't we check that if blob_versioned_hashes is None, than this is also None?

core/lib/eth_client/src/types.rs
```diff
@@ -176,4 +249,79 @@ mod tests {
...
+"8b81d1d09565b1a5e44eea2c1a00ee8cc1072aea32f7da9ab463e951e1be5c7875f9904be432787a666425c8f0a52178"
+        ).unwrap();
+
+        let signer = PrivateKeySigner::new(private_key);
+
+        let mut blob = vec![0u8; 131072];
+        for item in blob.iter_mut().take(12) {
+            *item = b'A';
+        }
+        blob[12] = 0x0a;
```
mm-zk, line 272: can you comment what's happening here?

core/lib/eth_client/src/types.rs
```diff
@@ -129,6 +134,64 @@ pub struct SignedCallResult {
...
+fn encode_blob_tx_with_sidecar(
+    raw_tx: &RawTransactionBytes,
+    sidecar: &EthTxBlobSidecar,
+) -> RawTransactionBytes {
+    let EthTxBlobSidecar::EthTxBlobSidecarV1(sidecar) = sidecar;
+    let blobs_count = sidecar.blobs.len();
+
+    let raw_tx = &raw_tx.0;
+    let mut stream_outer = RlpStream::new();
+    stream_outer.begin_list(4);
```
mm-zk, line 171: comment where 4 comes from (and preferably replace with a constant)

core/lib/eth_client/src/types.rs
```diff
@@ -129,6 +134,64 @@ pub struct SignedCallResult {
...
+    stream_outer.append_raw(&raw_tx[1..], 1);
```
mm-zk, line 172: comment what's happening here

core/lib/eth_client/src/types.rs
```diff
@@ -129,6 +134,64 @@ pub struct SignedCallResult {
...
+        }
+    }
+}
+
+fn encode_blob_tx_with_sidecar(
+    raw_tx: &RawTransactionBytes,
+    sidecar: &EthTxBlobSidecar,
+) -> RawTransactionBytes {
+    let EthTxBlobSidecar::EthTxBlobSidecarV1(sidecar) = sidecar;
+    let blobs_count = sidecar.blobs.len();
```
mm-zk, line 167: can we check that number of blobs < 3?

core/lib/eth_client/src/types.rs
```diff
@@ -129,6 +134,64 @@ pub struct SignedCallResult {
...
+
+    pub fn raw_tx(&self, blob_tx_sidecar: Option<&EthTxBlobSidecar>) -> RawTransactionBytes {
+        match blob_tx_sidecar {
+            None => self.raw_tx.clone(),
+            Some(sidecar) => encode_blob_tx_with_sidecar(&self.raw_tx, sidecar),
+        }
+    }
+}
+
+fn encode_blob_tx_with_sidecar(
```
mm-zk, line 162: can you add a simple 'smoke' unittest that checks the output of more than 1 blob.

it can be as simple as keccak(encode_blob_tx(sidecar=vec['data1', 'data2'])) == 0x1234

so that if we ever change this logic by mistake, the test would fail.

core/lib/eth_signer/src/pk_signer.rs
```diff
@@ -107,6 +109,8 @@ mod test {
             chain_id: 270,
             transaction_type: Some(U64::from(1u32)),
             access_list: None,
+            blob_versioned_hashes: None,
```
mm-zk, line 112: can you also add a test with these values set?

core/lib/eth_signer/src/raw_ethereum_tx.rs
```diff
@@ -47,6 +48,10 @@ pub struct TransactionParameters {
     pub max_fee_per_gas: U256,
     /// miner bribe
     pub max_priority_fee_per_gas: U256,
+    /// Max fee per blob gas
```
mm-zk, line 51: These comments don't help much ...

Maybe instead say that for 4844 blobs, are using separate fee market.
And mention what happens if blobs are present and this is not set and vice versa -- if I have blobs, does this have to be set too?

core/lib/eth_signer/src/raw_ethereum_tx.rs
```diff
@@ -47,6 +48,10 @@ pub struct TransactionParameters {
...
+    pub max_fee_per_blob_gas: Option<U256>,
+    /// blob versioned hashes
```
mm-zk, line 53: this comment doesn't say much..

instead - maybe mention about the transaction type. What happens when we use the 'legacy' transactoin type and set this?
Maybe we should change the transction_type to be enum (with blob & blob fee being available only if this is the 4844 transaction?)

core/lib/eth_signer/src/raw_ethereum_tx.rs
```diff
@@ -147,6 +155,42 @@ impl Transaction {
...
+        // cannot have the form of a create transaction.
+        stream.append(self.to.as_ref().unwrap());
+
+        stream.append(&self.value);
+        stream.append(&self.data);
+
+        self.rlp_append_access_list(&mut stream);
+
+        stream.append(&self.max_fee_per_blob_gas.unwrap());
+        // TODO: is this the correct way to encode a vector of versioned hashes
```
mm-zk, line 183: should this TODO stay here? ;-)

core/lib/eth_signer/src/raw_ethereum_tx.rs
```diff
@@ -147,6 +155,42 @@ impl Transaction {
...
+        // represent a 20-byte address. This means that blob transactions
+        // cannot have the form of a create transaction.
+        stream.append(self.to.as_ref().unwrap());
+
+        stream.append(&self.value);
+        stream.append(&self.data);
+
+        self.rlp_append_access_list(&mut stream);
+
+        stream.append(&self.max_fee_per_blob_gas.unwrap());
```
mm-zk, line 182: maybe these should be inside the transaction type enum? (this way we'll know that they are not set for other types of transactions)

core/lib/eth_signer/src/raw_ethereum_tx.rs
```diff
@@ -147,6 +155,42 @@ impl Transaction {
         stream
     }
 
+    /// Encodes a [`eip4844`] transaction: <https://eips.ethereum.org/EIPS/eip-4844>
+    fn encode_eip4844_payload(&self, chain_id: u64, signature: Option<&Signature>) -> RlpStream {
```
mm-zk, line 159: do we have any unittests for this? if not, I'd suggest adding at least one in this file

core/lib/eth_signer/src/raw_ethereum_tx.rs
```diff
@@ -147,6 +155,42 @@ impl Transaction {
...
+        let mut stream = RlpStream::new();
+
+        let list_size = if signature.is_some() { 14 } else { 11 };
```
mm-zk, line 162: where does 14 and 11 come from?

core/lib/types/src/eth_sender.rs
```diff
@@ -9,6 +28,7 @@ pub struct EthTx {
     pub tx_type: AggregatedActionType,
     pub created_at_timestamp: u64,
     pub predicted_gas_cost: u64,
+    pub blob_sidecar: Option<EthTxBlobSidecar>,
```
mm-zk, line 31: can you also add the info about this to the fmt below (so that it mentions how many blobs we have)

core/lib/zksync_core/src/eth_sender/eth_tx_aggregator.rs
```diff
@@ -404,12 +404,12 @@ impl EthTxAggregator {
         &self,
         op: &AggregatedOperation,
         contracts_are_pre_boojum: bool,
-    ) -> Vec<u8> {
+    ) -> (Vec<u8>, Option<Vec<u8>>) {
```
mm-zk, line 407: please add the comment to this method and mention what it returns

core/lib/zksync_core/src/eth_sender/eth_tx_aggregator.rs
```diff
@@ -448,7 +448,9 @@ impl EthTxAggregator {
                 f.encode_input(&op.into_tokens())
                     .expect("Failed to encode execute transaction data")
             }
-        }
+        };
+
+        (calldata, None)
```
mm-zk, line 453: so we always return None as secont param? that's surprising, please add commetn explaining

core/lib/zksync_core/src/eth_sender/eth_tx_manager.rs
```diff
@@ -410,7 +416,10 @@ impl EthTxManager {
                     opt.max_fee_per_gas = Some(U256::from(base_fee_per_gas + priority_fee_per_gas));
                     opt.max_priority_fee_per_gas = Some(U256::from(priority_fee_per_gas));
                     opt.nonce = Some(tx.nonce.0.into());
+                    opt.transaction_type = Some(EIP_1559_TX_TYPE.into());
```
mm-zk, line 419: I'm confused  - shouldn't the type depend on whether tx had blobs?

sdk/zksync-rs/src/ethereum/mod.rs
```diff
@@ -418,17 +424,20 @@ impl<S: EthereumSigner> EthereumProvider<S> {
                 tx_data,
                 Options::with(|f| {
                     f.gas = Some(U256::from(300000));
+                    f.transaction_type = Some(EIP_1559_TX_TYPE.into());
                     f.value = Some(value);
                     f.gas_price = Some(gas_price)
                 }),
+                None,
```
mm-zk, line 431: maybe consider having a separate method (sign_prepared_tx_with_blobs?)

this way you'll not have to add these None,None in all these places?

core/lib/eth_client/src/types.rs
```diff
@@ -129,6 +134,64 @@ pub struct SignedCallResult {
...
+        Self {
+            raw_tx,
+            max_priority_fee_per_gas,
+            max_fee_per_gas,
+            nonce,
+            hash,
+        }
+    }
+
+    pub fn raw_tx(&self, blob_tx_sidecar: Option<&EthTxBlobSidecar>) -> RawTransactionBytes {
```
mm-zk, line 154: can you explain what this method does? Also I'm a little bit surprised, that we are adding this 'blob tx sidecar' from the side (as it seems to really modify the interface).
Maybe have 2 methods instead? (raw_tx) & raw_tx_with_blob ? 

(the second one could also check - if we can - that the transaction is really a 'blob type' transaction? )

mm-zk (changes requested):

core/lib/eth_client/src/types.rs
montekki: why? this transaction can have more blobs: https://sepolia.blobscan.com/tx/0x9a22ccb0029bc8b0ddd073be1a1d923b7ae2b2ea52100bae0db4424f9107e9c0

core/lib/eth_client/src/types.rs
montekki: Before this patch the `SignedCallResult` contained a public field `raw_tx` which is essentially a transaction encoded into a byte form. We only supported 1559 transactions and so it made sense to be as it is. Important thing: for 1559 the signed transaction blob prefixed with a byte `0x02` to indicate the tx type is the network form of the transaction, end of story. With the instroduction of 4844 transactions that changes; the signed transaction blob (transaction + signature) + a byte prefix for type is no longer the complete network form for them, there is a new thing ("sidecar") that contains blobs, proofs and commitments + a new way it is all encoded together with the transaction (params + signature).

We want an interface that would leave no space for errors like: you wanted a 4844 tx, but actually went ahead and sent `raw_tx` field which in itself is not enough for it. All possible solutions to this are not ideal: you may misuse the current interface; you may misuse the interface proposed by you; currently we have no typesafe way of telling one tx from another.

sdk/zksync-rs/src/ethereum/mod.rs
montekki: it would be a bit less typesafe but a bit more pretty i suppose. i don't have a strong opinion on it

core/lib/eth_signer/src/raw_ethereum_tx.rs
montekki: We don't have a transaction type enum.

core/lib/eth_signer/src/raw_ethereum_tx.rs
montekki: this PR contains unittests.

core/lib/eth_client/src/clients/mock.rs
```diff
@@ -389,6 +389,8 @@ impl BoundEthInterface for MockEthereum {
         data: Vec<u8>,
         _contract_addr: H160,
         options: Options,
+        _max_fee_per_blob_gas: Option<U256>,
+        _blob_versioned_hashes: Option<Vec<H256>>,
```
Deniallugo, line 393: should `_max_fee_per_blob_gas` be a part of an options and `_blob_versioned_hashes` be part of the data?

core/lib/eth_client/src/types.rs
Deniallugo: Should blob sidecar be a part of the `new` and raw_tx will return the correct value depends on it?

core/lib/eth_client/src/types.rs
```diff
@@ -176,4 +283,198 @@ mod tests {
...
+        pretty_assertions::assert_eq!(raw_tx_str, expected_str);
+    }
+
+    #[tokio::test]
+    // Tests the encoding of the `EIP_4844_TX_TYPE` transaction to
+    // network format defined by the `EIP`. That is, a signed transaction
+    // itself and the sidecar containing the blobs.
+    async fn test_generating_signed_raw_transaction_with_4844_sidecar_two_blobs() {
+        let private_key =
+            H256::from_str("27593fea79697e947890ecbecce7901b0008345e5d7259710d0dd5e500d040be")
```
Deniallugo, line 376: i hope it's newly generated key.

core/lib/eth_signer/src/raw_ethereum_tx.rs
Deniallugo: should we at least verify that it's the correct tx_type?

core/lib/zksync_core/src/eth_sender/eth_tx_aggregator.rs
Deniallugo: Should it be `TxData` enum? With calldata option and eip4844 data?

core/lib/eth_client/src/types.rs
montekki: https://github.com/matter-labs/zksync-era/blob/636fcfd308cdde92c6407b886d7b329938623e9d/core/lib/env_config/src/eth_sender.rs#L79

core/lib/eth_client/src/clients/mock.rs
montekki: [`Options`](https://docs.rs/web3/latest/web3/contract/struct.Options.html) is the external struct from `web3`, we cannot add new fields there. But both fields are related and `_blob_versioned_hashes` is also the part of tx.

core/lib/eth_client/src/types.rs
montekki: the only place where we call this `new` method is in `sign_prepared_tx_for_addr` and so we would need to pass the optional sidecar also there although it would be never used there. Which is probably even less ergonomic

core/lib/dal/migrations/20240201195827_add_4844_specific_tx_fields.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE eth_txs ADD COLUMN blob_sidecar BYTEA;
+ALTER TABLE eth_txs_history ADD COLUMN blob_sidecar BYTEA;
```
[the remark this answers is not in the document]
montekki: good call, done https://github.com/matter-labs/zksync-era/pull/976/commits/3449587b2cb66de6d786ee5809db2cccbcec0e74

core/lib/eth_client/src/types.rs
montekki: comment is added

core/lib/eth_client/src/types.rs
montekki: comment added

core/lib/eth_signer/src/raw_ethereum_tx.rs
montekki: Expanded a comment a bit

core/lib/zksync_core/src/eth_sender/eth_tx_aggregator.rs
montekki: Created a `struct` instead of enum, that should be enough

core/lib/eth_client/src/clients/mock.rs
popzxc: Maybe we could define our own `Options` that would privately embed `web3::Options`? Currently the interface looks somewhat clunky, e.g. we provide two ways of grouping, which doesn't make much sense.

sdk/zksync-rs/src/ethereum/mod.rs
popzxc: I think this problem would also be solved if we would introduce a custom options type.

popzxc (commented):
Didn't thoroughly check everything, a few surface comments

core/lib/dal/src/models/storage_eth_tx.rs
```diff
@@ -22,6 +22,10 @@ pub struct StorageEthTx {
     pub updated_at: NaiveDateTime,
     // TODO (SMA-1614): remove the field
     pub sent_at_block: Option<i32>,
+    // A `EIP_4844_TX_TYPE` transaction blob sidecar.
+    //
+    // Format a `bincode`-encoded `EthTxBlobSidecar` enum.
+    pub blob_sidecar: Option<Vec<u8>>,
```
popzxc, line 28: Maybe a stupid question, but still. If I'm not mistaken, blobs are taken as whole, even if you use a fraction of it. So we may assume that at times, the blob won't be fully utilized (e.g. each time when pubdata wasn't the criterion to seal the batch), and there will be an "empty" reminder.

Do we want to use any kind of compression when we store/load data? Given that the state size is a problem already, probably we would like to optimize on this front whenever possible.

montekki: yes well my current idea here is that as you see `Vec<u8>` is used for all fields while from the protocol level we need three arrays: `[u8; 32]` `[u8; 48]` and `[u8; 32*4096]`. Two reasons for this:
1. `serde` does not implement traits for such const-size arrays
2. we actually have to store actual data with variable lengths for blob instead of a const array.

So as such there will always be a layer of conversion between a blob expanded to const size and `Vec<u8>` that is stored in DB and back. We just haven't figured out the ergonomics with @koloz193 but that is the current idea and things will remain as they are now. If we should use some compression for it is another topic; however the versioned blob struct should make it possible to do so in the future if necessary.

koloz193: we can definitely compress the data at the db level. there are a few things that happen before we store the blob:
1. we take the raw pubdata and split it into chunks of size `4096*31` lets call these `zksync_blobs`
2. for each `zksync_blob` we then convert it into an `ethereum_blob`
3. from the `ethereum_blob` we compute the other kzg values needed

while the `zksync_blob` might have trailing zeroes, im not sure if/how many trailing zeroes are in the `ethereum_blob`. for this we'd want to check if we want to store the `zksync_blob` or the `ethereum_blob`. this would introduce a crypto dependency to be able to convert `zksync_blob` -> `ethereum_blob`

mm-zk (approved):

montekki:
Closing in favor of #1254

PR #978, opened 2024-01-31
refactor: Remove dead code from zksync_contracts

popzxc:
## What ❔

- Removes dead code
- Reduces the visibility for some (unnecessary) public functions

## Why ❔

- Dead code bad
- Excessive public APIs bad

## Checklist


core/lib/contracts/src/lib.rs
```diff
@@ -381,12 +312,6 @@ impl BaseSystemContracts {
         BaseSystemContracts::load_with_bootloader(bootloader_bytecode)
     }
 
-    /// BaseSystemContracts with playground bootloader - used for handling eth_calls.
-    pub fn estimate_gas() -> Self {
-        let bootloader_bytecode = read_bootloader_code("fee_estimate");
-        BaseSystemContracts::load_with_bootloader(bootloader_bytecode)
-    }
```
Deniallugo, line 388: Looks like we have to keep this function and use it in api

popzxc: Do you mean that it is a bug that this function is not used? It's never invoked anywhere.

Deniallugo: it's supposed to be use for getting bootloader for the latest vm

StanislavBreadless (approved):

PR #992, opened 2024-02-01
feat(merkle-tree): Do not wait for tree initialization when starting node

slowli:
## What ❔

Makes tree initialization async in metadata calculator (i.e., moves it from the constructor to the `run()` method).

## Why ❔

It currently takes ~1m to initialize Merkle tree RocksDB for ENs.

## Checklist

### DIFF
--- a/core/lib/zksync_core/src/metadata_calculator/mod.rs
+++ b/core/lib/zksync_core/src/metadata_calculator/mod.rs
@@ -4,7 +4,7 @@
 use std::{
     future::{self, Future},
     sync::Arc,
-    time::Duration,
+    time::{Duration, Instant},
 };
 
 use anyhow::Context as _;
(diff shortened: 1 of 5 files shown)


slowli:
One thing that I'm not sure about is when the tree health should be marked as "ready":

- Immediately when the metadata calculator starts running
- When the tree starts recovering from a snapshot or is ready to process new L1 batches (this is how it's defined currently)
- Whet the tree is ready to process new L1 batches

popzxc:
> One thing that I'm not sure about is when the tree health should be marked as "ready"

Given that it's used for checking the health of the pod AFAIK, I think the right approach is as soon as the metadata calculator is running. Otherwise we may get alerts if the pod takes too long to start functioning.

core/lib/zksync_core/src/metadata_calculator/mod.rs
```diff
@@ -141,19 +132,47 @@ impl MetadataCalculator {
...
+        )
+        .await
+        .with_context(|| {
+            format!(
+                "failed opening Merkle tree RocksDB with configuration {:?}",
+                self.config
+            )
+        })?;
+        tracing::info!(
+            "Opened Merkle tree RocksDB with configuration {:?}",
```
popzxc, line 151: Would it make sense to also report how much time did it take?

popzxc (approved):

PR #995, opened 2024-02-01
feat(en): Take into account nonce from tx proxy

Deniallugo:
## What ❔

Take into account nonce from TxProxy, now if tx was submitted but miniblock was not synced yet. en will return correct nonce  

## Why ❔

[EN doesn't consider ](https://github.com/matter-labs/zksync-era/blob/main/core/lib/zksync_core/src/api_server/web3/namespaces/eth.rs#L303)transactions currently residing in the TxProxy when calculating pending nonce, which results in returned nonces being incorrect t times
## Checklist


core/bin/external_node/src/main.rs
```diff
@@ -154,6 +158,7 @@ async fn init_tasks(
             tokio::time::sleep(Duration::from_secs(10)).await;
         }
     }));
+    let tx_cache = Arc::new(RwLock::new(TxCache::default()));
```
slowli, line 161: Exposing `TxCache` internally looks like not the best idea; IMO, it'd make more sense to encapsulate it as with other API server tasks.

slowli (commented):
I have published [a branch](https://github.com/matter-labs/zksync-era/compare/deniallugo-pla-733-en-consider-txs-stored-in-proxy-when-calculating-pending...deniallugo-pla-733--aov) to illustrate changes in `TxCache` configuration / initialization. It **does not** address other concerns, such as correctness of nonce computation. The logic of updating `nonces_by_account` map *only* in the updater corresponds to transactions never evicted from the EN mempool until they are included in a miniblock; not entirely sure it's correct, but it's equivalent to your implementation with `TxProxy::forget_tx()` called only from `get_transaction_impl()` (in my version, the removed call in `TxSender:: submit_tx()` can be restored).

Regarding updater parameters: IMO, it's reasonable to hard-code the polling interval, at least initially. Regarding the use of a specific `ConnectionPool`, I'm not sure; IMO, your approach is OK (but if it stays, I'd advise to rename `last_miniblock_pool` etc. to `updaters_pool` or something like that).

core/lib/zksync_core/src/api_server/web3/namespaces/eth.rs
```diff
@@ -490,6 +491,24 @@ impl EthNamespace {
                 .map_err(|err| internal_error(method_name, err))?;
         }
 
+        // Number of txs accepted by the node.
+        // We should add the nonces from txs in our cache for making the nonce
+        // aligned with the main node. If the nonce is not sequential,
+        // then we should not increment nonce.
```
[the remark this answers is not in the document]
Deniallugo: But TxCache lives only in memory and we do restart for all reverts, so it will be empty after the revert. 

For now we have either mempool or tx cache (mainnode vs en) so they don't work together

core/lib/dal/src/storage_web3_dal.rs
```diff
@@ -32,6 +32,46 @@ impl StorageWeb3Dal<'_, '_> {
         Ok(decompose_full_nonce(full_nonce).0)
     }
 
+    /// Returns the current *stored* nonces (i.e., w/o accounting for pending transactions) for the specified accounts.
+    pub async fn get_nonces_for_addresses(
+        &mut self,
+        addresses: &[Address],
+    ) -> sqlx::Result<HashMap<Address, Nonce>> {
+        let res = sqlx::query!(
```
slowli, line 40: This is not a query I had in mind; it should be based on `storage_logs`, essentially encapsulating [`StorageWeb3Dal::get_values()`](https://github.com/matter-labs/zksync-era/blob/758f4877433ebdd86cc90dd592db9a7d87598586/core/lib/dal/src/storage_web3_dal.rs#L56) (you could even refactor [`MempoolFetcher`](https://github.com/matter-labs/zksync-era/blob/758f4877433ebdd86cc90dd592db9a7d87598586/core/lib/zksync_core/src/state_keeper/mempool_actor.rs#L140) to use this method as well, although it's by no means necessary – I can do this in a follow-up PR). Sorry for not being clear. One case in which your query certainly won't work is snapshot recovery; a recovered DB won't contain historic transactions (but it will contain appropriate storage values).

Relatively minor compared to the problem above, your query contradicts the doc comment; it *does* take pending non-errored transactions into account.

core/lib/dal/src/transactions_dal.rs
```diff
@@ -1344,6 +1344,36 @@ impl TransactionsDal<'_, '_> {
         .unwrap()
         .map(|tx| tx.into())
     }
+
+    pub async fn check_tx_hashes(
```
slowli, line 1348: This query looks unused.

core/lib/zksync_core/src/api_server/web3/namespaces/eth.rs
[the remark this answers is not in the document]
slowli: > so it will be empty after the revert

I get this, but the EN will still report transaction count *including* the reverted transactions, as if they are present in the mempool. Maybe this is a feature, not a bug, I just want to highlight that it's inconsistent with the "ordinary" EN restart (after which the EN forgets all transactions in the mempool).

> For now we have either mempool or tx cache (mainnode vs en) so they don't work together

This is non-local reasoning; I feel that it should be explained as a comment. Also, it could be easily avoided if the code is refactored as follows:

```rust
if matches!(block_id, BlockId::Number(BlockNumber::Pending)) {
    let pending_nonces: Vec<Nonce> = if let Some(proxy) = &self.state.tx_sender.0.proxy {
        // EN: get pending nonces from the transaction cache
        proxy.get_nonces_by_account(address).await
    } else {
        // Main node: load pending nonces from the database
        connection
            .transactions_web3_dal()
            .pending_nonces_by_initiator_account(start_nonce) // minor refactoring of `next_nonce_by_initiator_account`
            .await
            .map_err(|err| internal_error(method_name, err))?
    };
    // find out the next free nonce based on `pending_nonces`.
}
```

Maybe, it makes sense instead to refactor `get_nonces_by_account` so that it has the same semantics as `pending_nonces_by_initiator_account` (i.e., accepts a starting nonce and returns the next free nonce). It could be beneficial because it makes it easier to detect a currently present bug: `get_nonces_by_account` may return outdated nonces (i.e., ones lesser than `actual_tx_number`) because they are updated asynchronously. In this case, the current logic won't work correctly.

Also, you haven't addressed the concern with non-pending block IDs. Unless I'm missing something obvious, pending nonces should only be taken into account for `BlockNumber::Pending`, but currently they are taken into account for all requests on the EN. This can lead to bogus values returned (e.g., for `BlockNumber::Latest`).

core/lib/dal/src/storage_web3_dal.rs
Deniallugo: In case of nonces from transactions, there are many more cases where we take into account nonces from txs table. 
Probably it deserves to be addressed all together

core/lib/zksync_core/src/api_server/web3/namespaces/eth.rs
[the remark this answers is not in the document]
Deniallugo: Finally, i get it. 
Yes you are right. 
We have mempool in main node and we have tx proxy in EN, we have to use tx proxy as mempool. 

It doesn't make sense to get pending nonce from the db for EN, because it's always empty

core/lib/zksync_core/src/api_server/tx_sender/proxy.rs
```diff
@@ -1,22 +1,110 @@
...
+            drop(storage); // Don't hold both `storage` and lock on `inner` at the same time.
+
+            let mut inner = self.inner.write().await;
+            inner.nonces_by_account.retain(|address, account_nonces| {
+                let stored_nonce = nonces_for_accounts
+                    .get(address)
+                    .copied()
+                    .unwrap_or(Nonce(0));
+                // Retain only nonces exceeding the stored one.
+                *account_nonces = account_nonces.split_off(&(stored_nonce + 1));
```
slowli, line 92: I think there's an off-by-one error (introduced by me, sorry). `stored_nonce` is actually the next expected nonce for the account (e.g., if the corresponding storage slot is empty – i.e., we have `Nonce(0)` – it means that the *next* nonce should be 0). So here, it should be `split_off(stored_nonce)`. Ideally, this should be covered by a unit test 🙃

core/lib/zksync_core/src/api_server/tx_sender/proxy.rs
```diff
@@ -25,20 +113,40 @@ impl TxProxy {
...
+    pub async fn get_nonces_by_account(&self, account_address: Address) -> Vec<Nonce> {
+        self.tx_cache.get_nonces_for_account(account_address).await
+    }
+
+    pub async fn next_nonce_by_initiator_account(
+        &self,
+        account_address: Address,
+        current_nonce: u32,
+    ) -> Nonce {
+        let mut exprected_nonce = current_nonce;
```
perekopskiy, line 141: nit: typo `exprected_nonce` -> `expected_nonce`

core/lib/zksync_core/src/lib.rs
```diff
@@ -1100,15 +1100,15 @@ async fn run_http_api<G: L1GasPriceProvider>(
     }
     namespaces.push(Namespace::Snapshots);
 
-    let last_miniblock_pool = ConnectionPool::singleton(postgres_config.replica_url()?)
+    let updaters_pool = ConnectionPool::singleton(postgres_config.replica_url()?)
```
perekopskiy, line 1103: I think you should create pool with 2 connections, so that sealed_miniblock and nonce_sweeper tasks don't block each other

perekopskiy: Well, if nonce_sweeper query takes much time for some reason it will drastically affect sealed_miniblock task which is bad

core/lib/zksync_core/src/api_server/tx_sender/proxy.rs
```diff
@@ -25,20 +113,41 @@ impl TxProxy {
...
     }
 
-    pub async fn save_tx(&self, tx_hash: H256, tx: L2Tx) {
-        self.tx_cache.write().await.insert(tx_hash, tx);
+    pub async fn next_nonce_by_initiator_account(
+        &self,
+        account_address: Address,
+        current_nonce: u32,
+    ) -> Nonce {
+        let mut expected_nonce = current_nonce.saturating_sub(1);
```
slowli, line 141: Dumb question: Shouldn't this use the same logic as in `next_nonce_by_initiator_account()`, i.e.:

```rust
let mut expected_nonce = current_nonce;
// ...
if pending_nonce.0 == expected_nonce {
    expected_nonce += 1;
}
Nonce(expected_nonce)
```

I *think* these two approaches are always equivalent (including edge cases when `current_nonce == 0` and/or `nonces` start with 0), so I'd prefer using the same approach as in another place (which is tested / has been used in production).

core/lib/zksync_core/src/api_server/tx_sender/proxy.rs
```diff
@@ -25,20 +113,41 @@ impl TxProxy {
...
+        let nonces = self.get_nonces_by_account(account_address).await;
```
slowli, line 142: As discussed earlier, we may have `nonces` less than `current_nonce` because they are not cleaned up yet; these should be ignored. IIUC, in the current approach not ignoring them can lead to bogus value returned; e.g., `current_nonce == 2` and `nonces == [1, 2, 3]` should output 4, but will output 2. To ignore nonces efficiently, you can use [`BTreeSet::range()`](https://doc.rust-lang.org/std/collections/struct.BTreeSet.html#method.range).

slowli (commented):
Still got a couple of questions regarding nonce computations, but otherwise looks nice 👍

core/lib/zksync_core/src/api_server/tx_sender/proxy.rs
```diff
@@ -1,22 +1,110 @@
...
 
+#[derive(Debug, Clone, Default)]
+pub(crate) struct TxCache {
+    inner: Arc<RwLock<TxCacheInner>>,
+}
+
+#[derive(Debug, Default)]
+struct TxCacheInner {
+    tx_cache: HashMap<H256, L2Tx>,
+    nonces_by_account: HashMap<Address, BTreeSet<Nonce>>,
```
perekopskiy, line 29: Probably we should limit number of addresses stored in the cache

core/lib/zksync_core/src/lib.rs
perekopskiy: I would insist on either limiting max number of accounts stored in nonce cache or having separate DB connection for each update task. First would be better IMO

perekopskiy (approved):

core/lib/zksync_core/src/api_server/tx_sender/proxy.rs
RomanBrodetski: @Deniallugo @slowli is this resolved?

perekopskiy: As far as I see it's

core/lib/zksync_core/src/api_server/tx_sender/proxy.rs
Deniallugo: We kinda can limit it, but I'm not sure it deserves it.

PR #997, opened 2024-02-02
feat: Migrations linter

tomg10:
I've added a small linter, that makes it significantly harder to add a migration that is not backwards-compatible.
It's a sort of a temporary solution until we come up with a better one.


CODEOWNERS
```diff
@@ -2,3 +2,4 @@
 **/CHANGELOG.md @RomanBrodetski @perekopskiy @Deniallugo @popzxc
 CODEOWNERS @RomanBrodetski @perekopskiy @Deniallugo @popzxc
 .github/workflows/** @matter-labs/devops
+not-backwards-compatible-migrations-whitelist.txt @RomanBrodetski @perekopskiy @Deniallugo @popzxc
```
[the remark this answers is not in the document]
tomg10: corrected

infrastructure/zk/src/lint.ts
```diff
@@ -38,6 +39,10 @@ async function proverClippy() {
     await utils.spawn('cargo clippy --tests --locked -- -D warnings -A incomplete_features');
 }
 
+async function migrations() {
```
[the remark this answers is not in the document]
tomg10: I've removed that method all together and switched to directly using validateMigrations

tomg10:
It is relatively strict, but such migrations are rare, in last quarter, there have been on average only 2 per month

infrastructure/zk/src/validate_migrations.ts
```diff
@@ -0,0 +1,39 @@
+import * as utils from './utils';
+import fs from 'fs';
+
+async function validateMigration(filePath: string) {
+    const content = await fs.promises.readFile(filePath, { encoding: 'utf-8' });
+    const isBackwardsCompatible = !content.includes('RENAME') && !content.includes('DROP');
```
pompon0, line 6: are these checks case sensitive? sql is case insensitive.

tomg10: corrected

infrastructure/zk/src/validate_migrations.ts
```diff
@@ -0,0 +1,40 @@
+import * as utils from './utils';
+import fs from 'fs';
+
+async function validateMigration(filePath: string) {
+    let content = await fs.promises.readFile(filePath, { encoding: 'utf-8' });
+    content = content.toUpperCase();
+    const isBackwardsCompatible = !content.includes('RENAME') && !content.includes('DROP');
```
slowli, line 7: As discussed, I don't really like this check. E.g., AFAIU, it will mark [this migration](https://github.com/matter-labs/zksync-era/blob/b947d02c522a30026c28fc915d7fda1b4a50ffac/core/lib/dal/migrations/20240130151508_add_extra_fields_to_snapshot_recovery.up.sql) as false positive:

```sql
-- `miniblock_root_hash` should be renamed to `miniblock_hash`, but we cannot do it straightforwardly
```

(notice the "rename" use). As a comprehensive solution, you'd need to parse SQL (there should be packages for this, right?) and blacklist specific statement types.

infrastructure/zk/src/validate_migrations.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export async function validateMigrations() {
+    const allowlistedMigrations = await getallowlistedMigrations();
+    let { stdout: filesRaw } = await utils.exec('find core/lib/dal -type f -name "*up.sql"');
+    const files = filesRaw
+        .trim()
+        .split('\n')
+        .sort()
+        .filter((filepath) => !allowlistedMigrations.includes(filepath));
+    const validateResults = await Promise.all(files.map((file) => validateMigration(file)));
```
slowli, line 33: Nit: AFAIU, you can shorten `map` invocation `files.map(validateMigration)`, like in Rust.

slowli (changes requested):

infrastructure/zk/src/validate_migrations.ts
slowli: I don't think it's OK to have a false positive for the explicit opposite of the tested predicate. @RomanBrodetski WDYT?

RomanBrodetski: Taking a step back, I'd like to solve it in a more general way, for example running int testds against the old server version with new DB schema.

I think those false positives are indeed a bummer (I can even imagine a case when someone adds migration to the exceptions list because of said comment, and then adds a rename/delete statement without realising it's now breaking - artificial example of course, but still). But also parsing SQL just to see if there is a drop or rename is an overkill.

@tomg10 would having a CI job that tests new schema against old code be sufficient in place of this PR?

slowli (approved):
As discussed in personal communication, having an imperfect implementation is better than having none at all (and in the worst case, it could be switched off).

perekopskiy:

PR #1002, opened 2024-02-02
feat: Better errors for JsonRPC calls

tomg10:
## What ❔

Provides additional context for JSON-RPC client calls used by the EN with the help of an extension trait.

## Why ❔

Some of client errors are currently quite obscure and thus difficult to debug.

## Checklist


core/lib/web3_decl/src/error.rs
```diff
@@ -42,3 +43,65 @@ pub enum Web3Error {
...
+    pub inner_error: RpcError,
+    pub method: String,
+    pub args: HashMap<String, String>,
+}
+impl error::Error for RpcErrorWithDetails {}
+
+impl fmt::Display for RpcErrorWithDetails {
+    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let args_formatted: Vec<String> =
+            self.args.iter().map(|(k, v)| format!("{k}: {v}")).collect();
```
slowli, line 60: Why not format `args` with the `Debug` formatter (i.e., `{args:?}`)? It outputs args in a quite similar format.

core/lib/web3_decl/src/error.rs
```diff
@@ -42,3 +43,65 @@ pub enum Web3Error {
...
+
+use std::{collections::HashMap, error, fmt, fmt::Debug};
+
+#[derive(Debug)]
+pub struct RpcErrorWithDetails {
+    pub inner_error: RpcError,
+    pub method: String,
+    pub args: HashMap<String, String>,
+}
+impl error::Error for RpcErrorWithDetails {}
```
slowli, line 55: You should override [`source()`](https://doc.rust-lang.org/std/error/trait.Error.html#method.source) method to point at the original `RpcError`.

core/lib/web3_decl/src/error.rs
```diff
@@ -42,3 +43,65 @@ pub enum Web3Error {
     #[error("Tree API is not available")]
     TreeApiUnavailable,
 }
+
+use std::{collections::HashMap, error, fmt, fmt::Debug};
+
+#[derive(Debug)]
+pub struct RpcErrorWithDetails {
+    pub inner_error: RpcError,
+    pub method: String,
```
slowli, line 52: Are there use cases when `method` is not known in compile time (i.e., cannot be a `&'static str`)? Same question regarding `args` keys.

core/lib/web3_decl/src/error.rs
```diff
@@ -42,3 +43,65 @@ pub enum Web3Error {
...
+pub trait WithArgRpcError<T> {
+    fn with_arg(self, arg: &str, value: &dyn fmt::Debug) -> Result<T, RpcErrorWithDetails>;
+}
+
+impl<T> WithArgRpcError<T> for Result<T, RpcErrorWithDetails> {
+    fn with_arg(self, arg: &str, value: &dyn Debug) -> Result<T, RpcErrorWithDetails> {
+        match self {
+            Ok(t) => Ok(t),
+            Err(error) => {
+                let mut new_args = error.args.clone();
```
slowli, line 97: You probably don't need cloning here; the error is consumed by the method.

core/lib/web3_decl/src/error.rs
```diff
@@ -42,3 +43,65 @@ pub enum Web3Error {
...
+            Err(error) => Err(RpcErrorWithDetails {
+                inner_error: error,
+                method: method.to_string(),
+                args: HashMap::default(),
+            }),
+        }
+    }
+}
+
+pub trait WithArgRpcError<T> {
```
slowli, line 88: Bikeshedding: 2 distinct traits is why I haven't proposed this option when discussing how best to implement this functionality; IMO, it looks mildly awkward. If you really think that it's the best option, I won't oppose (personally, I'd like the `Future`-wrapping approach better; it has a single trait and also allows tracking long-executing requests).

core/lib/web3_decl/src/error.rs
```diff
@@ -42,3 +43,65 @@ pub enum Web3Error {
     #[error("Tree API is not available")]
     TreeApiUnavailable,
 }
+
+use std::{collections::HashMap, error, fmt, fmt::Debug};
+
+#[derive(Debug)]
+pub struct RpcErrorWithDetails {
+    pub inner_error: RpcError,
```
slowli, line 51: It looks weird for errors to have public fields. I'd suggest to make fields private and define a getter for `inner_error` (e.g., `fn inner(&self) -> &RpcError`) instead.

core/lib/web3_decl/src/error.rs
tomg10: I wanted to make it a bit nicer, but it does make sense to just use Debug formatter

core/lib/web3_decl/src/error.rs
tomg10: I decided to just leave it as it is

slowli (approved):
Good enough 👍 There are some nits and potential improvements that I'll address in a follow-up PR if you don't mind.

core/lib/web3_decl/src/error.rs
```diff
@@ -42,3 +55,152 @@ pub enum Web3Error {
     #[error("Tree API is not available")]
     TreeApiUnavailable,
 }
+
+/// RPC error with additional details: the method name and arguments of the called method.
+///
+/// The wrapped error can be accessed using [`AsRef`].
+#[derive(Debug)]
+pub struct EnrichedRpcError {
```
popzxc, line 63: This is generally for _clients_ to add additional information about requests made, right?
I was initially a bit confused assuming that it can be used on the server side too, and it doesn't look like this.

If so, I wonder whether we make it a bit less confusing, e.g. by changing the naming (`Enrich`_`Client`_`RpcError`) or moving it elsewhere (create some client-specific module? `client_error_ext`?).

core/lib/web3_decl/src/error.rs
```diff
@@ -1,5 +1,18 @@
...
+    error,
+    error::Error,
+    fmt,
+    future::Future,
+    mem,
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+use jsonrpsee::core::ClientError as RpcError;
```
popzxc, line 14: Nit: This alias contributed to the confusion, tbh. jsonrpsee has a lot of error types, and I didn't really understand which error type it was.

popzxc (approved):
I like this addition.

core/lib/web3_decl/src/error.rs
slowli: Yeah, I think using the original `ClientError` name instead of `RpcError` would make things clearer.

popzxc (approved):
👍

core/lib/web3_decl/src/error.rs
```diff
@@ -42,3 +55,152 @@ pub enum Web3Error {
...
+        impl fmt::Debug for DebugArgs<'_> {
+            fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+                formatter.write_str("(")?;
+                for (i, (name, value)) in self.0.iter().enumerate() {
+                    write!(formatter, "{name}={value}")?;
+                    if i + 1 < self.0.len() {
+                        formatter.write_str(", ")?;
+                    }
+                }
+                formatter.write_str(")")
```
RomanBrodetski, line 120: just curious - could we do something like `.join(", ")` here? Did you want to avoid allocating a vector for this?

RomanBrodetski (commented):
nice

PR #1004, opened 2024-02-04
fix(db): miniblocks index

RomanBrodetski:
We added this index to the DB manually, need to have it in the migrations


RomanBrodetski:
to this PR? Because why not - I don't need this history, it's only a distraction

RomanBrodetski:
Thanks for the swift review btw

Deniallugo (approved):

PR #1012, opened 2024-02-05
feat(shared bridge): preparation for shared bridge migration (server)

ly0va:
## What ❔

This PR aims to reduce the diff between main and #298 to simplify the future review.

The changes do not affect current server behavior.
Only changes related to the zkSync server went into this PR.

What didn't go into this PR:

- various variable renamings
- integration tests
- changes that depend on the future version of era-contracts

Where applicable, stubs and dummy (for now) config variables were introduced, which can be replaced/populated later.

## Why ❔

Breaking up a PR into smaller chunks should be easier to review.

## Checklist


core/lib/protobuf_config/src/contracts.rs
```diff
@@ -122,6 +122,36 @@ impl ProtoRepr for proto::Contracts {
...
+                .as_ref()
+                .map(|x| parse_h160(x))
+                .transpose()
+                .context("bridgehub_proxy_addr")?,
+            bridgehub_impl_addr: self
+                .bridgehub_impl_addr
+                .as_ref()
+                .map(|x| parse_h160(x))
+                .transpose()
+                .context("bridgeub_impl_addr")?,
```
perekopskiy, line 136: nit
```suggestion
                .context("bridgehub_impl_addr")?,
```

core/lib/zksync_core/src/eth_sender/eth_tx_aggregator.rs
```diff
@@ -403,50 +387,64 @@ impl EthTxAggregator {
...
+                if contracts_are_pre_shared_bridge {
+                    self.functions
+                        .pre_shared_bridge_commit
+                        .encode_input(&op.into_tokens())
+                        .expect("Failed to encode commit transaction data")
                 } else {
+                    args.extend(op.into_tokens());
                     self.functions
-                        .post_boojum_commit
+                        .post_shared_bridge_commit
```
perekopskiy, line 410: Please add unit tests for post_shared_bridge eth sender

core/lib/zksync_core/src/eth_watch/event_processors/set_chain_id.rs
```diff
@@ -0,0 +1,78 @@
...
+
+use crate::eth_watch::{
+    client::{Error, EthClient},
+    event_processors::EventProcessor,
+    metrics::{PollStage, METRICS},
+};
+
+/// Responsible for saving `setChainId` upgrade transactions to the database.
+#[derive(Debug)]
+pub struct SetChainIDEventProcessor {
```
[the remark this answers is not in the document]
ly0va: I like this idea, this would simplify a lot of things. But don't we need to run genesis before deploying the contracts? I think it logs some config variables used in contracts later, like the genesis batch commitment etc

core/lib/zksync_core/src/eth_sender/tests.rs
```diff
@@ -111,8 +111,9 @@ impl EthSenderTester {
             Address::random(),
             contracts_config.l1_multicall3_addr,
             Address::random(),
-            0,
-        );
+            270.into(),
```
slowli, line 114: Nit: IMO, `L2ChainId::default()` would be more idiomatic here.

core/lib/zksync_core/src/eth_watch/event_processors/set_chain_id.rs
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        let ids_str: Vec<_> = upgrades
+            .iter()
+            .map(|(id, _tx)| format!("{}", *id as u16))
+            .collect();
+        tracing::debug!(
+            "Received setChainId upgrade with version_id: {}",
+            ids_str.join(", ")
+        );
```
slowli, line 62: Nit: As I understand, this code will be refactored anyway, but this looks overly complicated. You can collect IDs into `ids: Vec<u16>` and then output its `Debug` representation via `{ids:?}`.

core/lib/zksync_core/src/state_keeper/io/mempool.rs
```diff
@@ -455,7 +455,7 @@ impl MempoolIO {
     }
 
     async fn load_previous_l1_batch_hash(&self) -> U256 {
-        tracing::info!(
+        tracing::trace!(
```
slowli, line 458: Dumb question: What's the reason for these changes? Do these logs become spammy with this PR? Or are they already spammy?

core/lib/zksync_core/src/state_keeper/keeper.rs
```diff
@@ -234,6 +218,72 @@ impl ZkSyncStateKeeper {
...
+        // If it's not present, we have to wait for it to be picked up by the event watcher.
+        // The setChainId upgrade does not bump the protocol version, rather attaches an upgrade
+        // transaction to the genesis protocol version.
+        while first_batch_in_shared_bridge && protocol_upgrade_tx.is_none() {
+            tracing::info!("Waiting for setChainId upgrade transaction to be picked up");
+            if self.is_canceled() {
+                return Err(Error::Canceled);
+            }
+            tokio::time::sleep(Duration::from_secs(1)).await;
+            protocol_upgrade_tx = self.io.load_upgrade_tx(protocol_version).await;
```
slowli, line 250: To repeat the question we've discussed previously: for ENs, `load_upgrade_tx()` [always returns](https://github.com/matter-labs/zksync-era/blob/758f4877433ebdd86cc90dd592db9a7d87598586/core/lib/zksync_core/src/sync_layer/external_io.rs#L544-L545) `None`.  So it seems that the EN will enter an infinite loop if `first_batch_in_shared_bridge` is true. Is `load_protocol_upgrade_tx()` not supposed to be called for ENs? Is this implemented right now?

ly0va: This loop will be removed when [Misha's suggestion](https://github.com/matter-labs/zksync-era/pull/1012#discussion_r1481209538) is implemented, so ENs should still work.

core/lib/zksync_core/src/state_keeper/io/mempool.rs
ly0va: They are already spammy :)

core/lib/zksync_core/src/eth_watch/event_processors/set_chain_id.rs
ly0va: Sure. It's done the same way in the upgrade even processor, I'll change it there as well

core/lib/zksync_core/src/eth_sender/eth_tx_aggregator.rs
```diff
@@ -403,50 +387,64 @@ impl EthTxAggregator {
...
         op: &AggregatedOperation,
-        contracts_are_pre_boojum: bool,
+        contracts_are_pre_shared_bridge: bool,
     ) -> Vec<u8> {
-        let operation_is_pre_boojum = op.protocol_version().is_pre_boojum();
+        let operation_is_pre_shared_bridge = op.protocol_version().is_pre_shared_bridge();
+        assert_eq!(
+            contracts_are_pre_shared_bridge,
+            operation_is_pre_shared_bridge
+        );
```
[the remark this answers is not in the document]
ly0va: Yes.

core/lib/zksync_core/src/state_keeper/keeper.rs
ly0va: Hmm, it will break an assertion a few lines below though. I'll fix it

core/bin/zksync_server/src/main.rs
```diff
@@ -37,6 +37,10 @@ struct Cli {
     /// Generate genesis block for the first contract deployment using temporary DB.
     #[arg(long)]
     genesis: bool,
+    /// Wait for the `setChainId` event during genesis.
+    /// If `--genesis` is not set, this flag is ignored.
+    #[arg(long)]
+    set_chain_id: bool,
```
RomanBrodetski, line 43: why is it configueable/optional? Is it just for the time being?

core/lib/dal/src/protocol_versions_dal.rs
```diff
@@ -98,6 +98,47 @@ impl ProtocolVersionsDal<'_, '_> {
         db_transaction.commit().await.unwrap();
     }
 
+    async fn save_genesis_upgrade(&mut self, id: ProtocolVersionId, tx_hash: Option<H256>) {
```
RomanBrodetski, line 101: nit: 
```suggestion
    async fn save_genesis_upgrade_tx_hash(&mut self, id: ProtocolVersionId, tx_hash: Option<H256>) {
```

core/lib/zksync_core/src/eth_sender/eth_tx_aggregator.rs
ly0va: I can add some, but currently there's no point: the place where behavior differs is mocked away

core/lib/zksync_core/src/eth_watch/mod.rs
```diff
@@ -92,6 +93,31 @@ impl EthWatch {
         }
     }
 
+    /// `EthWatch` is only meant to be created this way during genesis, since
+    /// we only need to save the `setChainId` event once -- after the chain creation.
+    pub async fn new_set_chain_id_watch(
```
RomanBrodetski, line 98: is this meant to be pub?

core/lib/zksync_core/src/lib.rs
```diff
@@ -102,7 +102,9 @@ pub async fn genesis_init(
     eth_sender: &ETHSenderConfig,
     network_config: &NetworkConfig,
     contracts_config: &ContractsConfig,
+    ethwatch_config: &ETHWatchConfig,
     eth_client_url: &str,
+    wait_for_set_chain_id: bool,
```
RomanBrodetski, line 107: what does this flag depend on? Will it be remived in the future, or will chains be able to spin up both ways?

core/lib/dal/src/protocol_versions_dal.rs
```diff
@@ -98,6 +98,47 @@ impl ProtocolVersionsDal<'_, '_> {
...
+        tx: ProtocolUpgradeTx,
+    ) {
+        let tx_hash = Some(tx.common_data.hash());
+
+        let mut db_transaction = self.storage.start_transaction().await.unwrap();
+
+        db_transaction
+            .transactions_dal()
+            .insert_system_transaction(tx)
+            .await;
```
RomanBrodetski, line 132: why do we combine this on the dal level? Dal shouldn't have any business logic. The fact that you need to update two tables is better represented on the app layer (`eth-watch` in this case). So please create the transaction there and call two dal methods from there

core/lib/zksync_core/src/state_keeper/keeper.rs
```diff
@@ -147,27 +150,9 @@ impl ZkSyncStateKeeper {
...
-                    first_tx_to_reexecute.tx_format(),
-                    TransactionType::ProtocolUpgradeTransaction
-                )
-            }
-
-            None
-        } else {
-            None
-        };
+        let mut protocol_upgrade_tx = self
```
RomanBrodetski, line 153: much better in a separate method, agreed. I would also make it clear that it's an `Option` - e.g. by naming it as `protocol_upgrade_tx_option` or putting the type explicitely

core/lib/zksync_core/src/state_keeper/keeper.rs
```diff
@@ -241,6 +225,57 @@ impl ZkSyncStateKeeper {
...
+                    first_tx_to_reexecute.tx_format(),
+                    TransactionType::ProtocolUpgradeTransaction,
+                    "Expected an upgrade transaction to be the first one in pending_miniblocks, but found {:?}",
+                    first_tx_to_reexecute.hash()
+                );
+            }
+            tracing::info!("We have no upgrade tx to execute");
+            None
+        } else {
+            tracing::info!("We are not changing protocol version");
```
RomanBrodetski, line 272: nit:

```suggestion
            tracing::info!("There is no protocol version change");
```

Also, can you please add more comments to every if clause? what's the difference between no changing protocol version and no ugprade tx to execute?

RomanBrodetski (commented):
Overall looks good, mostly left some minor nits

RomanBrodetski (commented):
Please wait for @perekopskiy 's oapproval and feel free to merge - unless @slowli has some request changes as well

core/bin/zksync_server/src/main.rs
ly0va: Firstly, there's no need for it while we don't have a shared bridge. Secondly, with the shared bridge setup, we already run `zk server --genesis` twice (first time only to generate commitments, before contract deployment, IIRC @kelemeno?). During the second time, we will run it with `--set-chain-id`.

core/lib/zksync_core/src/lib.rs
ly0va: As per comment above, `zk server --genesis` will have to be run twice - with and without this flag.

core/lib/dal/src/protocol_versions_dal.rs
ly0va: it's the same as [`save_protocol_version_with_tx`](https://github.com/matter-labs/zksync-era/blob/689d38c3aea4d63da259362f9f96c51c3a359191/core/lib/dal/src/protocol_versions_dal.rs#L75) though?

core/lib/zksync_core/src/eth_sender/eth_tx_aggregator.rs
perekopskiy: I would expect from test to check that eth sender provides args compatible with ABI. I don't insist on adding test now if it would more convenient later, then feel free to add it later

core/lib/zksync_core/src/lib.rs
```diff
@@ -174,6 +176,20 @@ pub async fn genesis_init(
...
+        eth_watch::wait_for_set_chain_id(
+            ethwatch_config.clone(),
+            pool.clone(),
+            Box::new(eth_client),
+            contracts_config.diamond_proxy_addr,
+            contracts_config
+                .state_transition_proxy_addr
+                .expect("state_transition_proxy_addr is not set, but needed for genesis"),
+        )
+        .await?;
```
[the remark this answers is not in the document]
perekopskiy: Hmm, maybe it wasn't fully clear in my initial suggestion but I had in mind to drop all the changes to eth watch from the PR and replace them with smth like this pseudo-code extracted into a separate function in `genesis.rs` file
```
let eth_client = QueryClient::new(eth_client_url);
let filter = Filter { address, topic, from_block: 0}; 
let logs = eth_client.get_logs(filter);
assert_eq!(logs.len(), 1); // There should be exactly one event we look for.
let upgrade_tx = parse(logs[0]);
dal. save_genesis_upgrade_tx(upgrade_tx);
```

core/lib/types/src/protocol_version.rs
```diff
@@ -462,6 +367,164 @@ impl TryFrom<Log> for ProtocolUpgrade {
     }
 }
 
+pub fn decode_set_chain_id_event(
+    event: Log,
+) -> Result<(ProtocolVersionId, ProtocolUpgradeTx), crate::ethabi::Error> {
+    let transaction_param_type = ParamType::Tuple(vec![
```
slowli, line 373: Nit: This definition is duplicated from `impl TryFrom<Log> for ProtocolUpgrade` above; maybe, it makes sense to extract it to a function?

core/lib/zksync_core/src/eth_watch/event_processors/set_chain_id.rs
```diff
@@ -0,0 +1,73 @@
...
+        // SetChainId does not go through the governance contract, so we need to parse it separately.
+        let upgrades = events
+            .into_iter()
+            .filter(|log| {
+                log.topics[0] == self.set_chain_id_signature
+                    && log.topics[1] == self.diamond_proxy_address.into()
+            })
+            .map(|event| {
+                decode_set_chain_id_event(event)
+                    .map_err(|err| Error::LogParse(format!("{:?}", err)))
```
slowli, line 47: Off-topic: I feel that `LogParse` wraps a wrong type (should be some error, maybe `L1TxParseError`), but that has little to do with the PR 🙃

core/lib/zksync_core/src/eth_watch/mod.rs
```diff
@@ -131,13 +157,15 @@ impl EthWatch {
...
+    async fn run(
         &mut self,
         pool: ConnectionPool,
         stop_receiver: watch::Receiver<bool>,
+        limit: Option<usize>,
     ) -> anyhow::Result<()> {
         let mut timer = tokio::time::interval(self.poll_interval);
-        loop {
+        let mut processed_events = 0;
+        while limit.map(|l| processed_events < l).unwrap_or(true) {
```
slowli, line 168: Nit: You can simplify this call chain to `.map_or(true, |l| processed_events < l)`.

core/lib/zksync_core/src/eth_watch/mod.rs
```diff
@@ -131,13 +157,15 @@ impl EthWatch {
         }
     }
 
-    pub async fn run(
+    async fn run(
         &mut self,
         pool: ConnectionPool,
         stop_receiver: watch::Receiver<bool>,
+        limit: Option<usize>,
```
slowli, line 164: Bikeshedding: I feel it would be better to stop the component via `stop_receiver` by either watching its side-effects (that is, the storage), or by having the component emit events that can be subscribed to. One way to achieve the latter is to add `watch::Sender<bool>` to `SetChainIDEventProcessor` (initially `false`; set to `true` once at least one event is received). This would mean you don't need to change `EventProcessor` signature etc.

core/lib/zksync_core/src/eth_watch/event_processors/set_chain_id.rs
```diff
@@ -0,0 +1,73 @@
...
+#[derive(Debug)]
+pub struct SetChainIDEventProcessor {
+    /// Address of the `DiamondProxy` contract of the chain that the `SetChainId` event is for.
+    diamond_proxy_address: Address,
+    /// Signature of the `SetChainIdUpgrade` event.
+    /// The event is emitted by the `StateTransitionManager` contract.
+    set_chain_id_signature: H256,
+}
+
+impl SetChainIDEventProcessor {
```
slowli, line 21: Bikeshedding: Conventionally, acronyms in Rust are spelled with just the first letter capitalized (i.e., `SetChainIdEventProcessor`).

core/lib/zksync_core/src/eth_watch/mod.rs
```diff
@@ -178,27 +213,59 @@ impl EthWatch {
...
+    );
+
+    let mut eth_watch = EthWatch::new_set_chain_id_watch(
+        diamond_proxy_addr,
+        Box::new(eth_client),
+        config.poll_interval(),
+    )
+    .await;
+
+    let (_stop_sender, stop_receiver) = watch::channel(false);
```
slowli, line 251: It feels strange that this method is non-responsive (i.e., doesn't accept `stop_receiver` from outside). Is this intentional? Is it assumed that it won't really need to wait for long realistically?

core/lib/zksync_core/src/lib.rs
```diff
@@ -597,8 +615,13 @@ pub async fn initialize_components(
             contracts_config.validator_timelock_addr,
             contracts_config.l1_multicall3_addr,
             main_zksync_contract_address,
-            nonce.as_u64(),
-        );
+            configs
+                .network_config
+                .as_ref()
+                .context("netowrk_config")?
```
slowli, line 621: Typo: "netowrk_config" -> "network_config".

core/lib/zksync_core/src/lib.rs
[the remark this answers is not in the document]
ly0va: It does indeed look simpler, but is it flexible enough? What if we send a tx with the event, but block is not validated fast enough and we miss it? I think the polling logic is handy here, and it's already implemented in `EthWatch`

core/lib/zksync_core/src/state_keeper/keeper.rs
```diff
@@ -241,6 +225,62 @@ impl ZkSyncStateKeeper {
...
+            && (version_changed || first_batch_in_shared_bridge)
+        {
+            // We already processed the upgrade tx but did not seal the batch it was in.
+            // Sanity check: if `txs_to_reexecute` is not empty and upgrade tx is present for this block
+            // then it must be the first one in `txs_to_reexecute`.
+            if protocol_upgrade_tx.is_some() {
+                let first_tx_to_reexecute = &pending_miniblocks[0].txs[0];
+                assert_eq!(
+                    first_tx_to_reexecute.tx_format(),
+                    TransactionType::ProtocolUpgradeTransaction,
```
slowli, line 266: Does it make sense to compare `first_tx_to_reexecute` with `protocol_upgrade_tx` more thoroughly, e.g., by hash?

core/lib/zksync_core/src/state_keeper/keeper.rs
```diff
@@ -241,6 +225,62 @@ impl ZkSyncStateKeeper {
...
+
+        let version_changed = protocol_version != previous_batch_protocol_version;
+        let protocol_upgrade_tx = self.io.load_upgrade_tx(protocol_version).await;
+
+        let protocol_upgrade_tx = if pending_miniblocks.is_empty()
+            && (version_changed || first_batch_in_shared_bridge)
+        {
+            // We have a new upgrade transaction - either a regular protocol upgrade or a `setChainId` upgrade.
+            tracing::info!("There is an new upgrade tx to be executed in this batch");
+            protocol_upgrade_tx
```
slowli, line 255: Unless I'm missing something, it isn't guaranteed that `protocol_upgrade_tx` is actually `Some(_)` at this point. Should it be checked?

core/lib/zksync_core/src/lib.rs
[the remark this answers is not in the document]
perekopskiy: Well, if event is not ready/present, then `zk server --genesis` should panic with message "Couldn't find event ..." and you rerun it when event is ready

core/lib/zksync_core/src/state_keeper/keeper.rs
ly0va: It is `None` if we're running external node :)

core/lib/zksync_core/src/eth_watch/mod.rs
ly0va: Yes, this method is supposed to exit after a single `setChainId` event is found.

core/lib/zksync_core/src/eth_watch/mod.rs
ly0va: won't I have to call an `.unwrap()` after that anyway?

core/lib/dal/src/protocol_versions_dal.rs
```diff
@@ -98,6 +98,47 @@ impl ProtocolVersionsDal<'_, '_> {
...
+                upgrade_tx_hash = $1
+            WHERE
+                id = $2
+            "#,
+            tx_hash.map(|tx_hash| tx_hash.0.to_vec()),
+            id as i32,
+        )
+        .execute(self.storage.conn())
+        .await
+        .unwrap();
```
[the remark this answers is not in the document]
ly0va: I think this should be done as a separate PR, since there are a lot of places in DAL where we don't propagate DB errors yet

core/lib/zksync_core/src/lib.rs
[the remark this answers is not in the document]
RomanBrodetski: makes sense - agreed with @perekopskiy on this one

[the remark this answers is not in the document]
ly0va: implemented

perekopskiy (dismissed):

core/lib/dal/src/protocol_versions_dal.rs
RomanBrodetski: I think both have to be reworked, but there is no need to do it now. I'm resolving the convo.

RomanBrodetski (dismissed):
please wait for @slowli's approval

core/lib/zksync_core/src/genesis.rs
```diff
@@ -419,6 +422,41 @@ pub(crate) async fn save_genesis_l1_batch_metadata(
...
+        )
+        .from_block(from.into())
+        .to_block(BlockNumber::Latest)
+        .build();
+    let mut logs = eth_client.logs(filter, "fetch_chain_id_tx").await?;
+    assert_eq!(
+        logs.len(),
+        1,
+        "Expected a single set_chain_id event, got {:?}",
+        logs.len()
```
slowli, line 450: Nit: `logs.len()` would be output as the first arg in a failed `assert_eq` assertion anyway. IMO, it'd be more useful to output entire `logs` here, so that debugging is easier.

core/lib/zksync_core/src/genesis.rs
```diff
@@ -419,6 +422,41 @@ pub(crate) async fn save_genesis_l1_batch_metadata(
...
+            Some(vec![SET_CHAIN_ID_EVENT.signature()]),
+            Some(vec![diamond_proxy_address.into()]),
+            None,
+            None,
+        )
+        .from_block(from.into())
+        .to_block(BlockNumber::Latest)
+        .build();
+    let mut logs = eth_client.logs(filter, "fetch_chain_id_tx").await?;
+    assert_eq!(
```
slowli, line 446: Nit: Since this function propagates errors, it could make sense to use `anyhow::ensure!(logs.len() == 1, ..)` instead.

core/lib/zksync_core/src/lib.rs
```diff
@@ -174,6 +175,20 @@ pub async fn genesis_init(
...
     )
     .await?;
+
+    if wait_for_set_chain_id {
+        genesis::save_set_chain_id_tx(
+            eth_client_url,
+            contracts_config.diamond_proxy_addr,
+            contracts_config
+                .state_transition_proxy_addr
+                .expect("state_transition_proxy_addr is not set, but needed for genesis"),
```
slowli, line 185: Nit: Since this function propagates errors, `.context(_)` would be more idiomatic here.

core/lib/zksync_core/src/state_keeper/keeper.rs
```diff
@@ -241,6 +225,63 @@ impl ZkSyncStateKeeper {
...
+
+        let version_changed = protocol_version != previous_batch_protocol_version;
+        let protocol_upgrade_tx = self.io.load_upgrade_tx(protocol_version).await;
+
+        let protocol_upgrade_tx = if pending_miniblocks.is_empty()
+            && (version_changed || first_batch_in_shared_bridge)
+        {
+            // We have a new upgrade transaction - either a regular protocol upgrade or a `setChainId` upgrade.
+            // If we are running an EN, `protocol_upgrade_tx` will be `None` here since it is fetched from the main node.
+            tracing::info!("There is an new upgrade tx to be executed in this batch");
```
slowli, line 255: Nit: Here and in other log messages in this method, it could make sense to add context (e.g., "this batch" -> "batch #{l1_batch_number}"; info about `protocol_upgrade_tx`) to simplify debugging.

core/lib/zksync_core/src/genesis.rs
```diff
@@ -419,6 +422,41 @@ pub(crate) async fn save_genesis_l1_batch_metadata(
...
+        .topics(
+            Some(vec![SET_CHAIN_ID_EVENT.signature()]),
+            Some(vec![diamond_proxy_address.into()]),
+            None,
+            None,
+        )
+        .from_block(from.into())
+        .to_block(BlockNumber::Latest)
+        .build();
+    let mut logs = eth_client.logs(filter, "fetch_chain_id_tx").await?;
```
slowli, line 445: Dumb question: If `save_set_chain_id_tx` is called prematurely, so that there's no event emitted yet, terminating the node is the best course of action, right?

slowli (approved):
There are also a couple of unresolved nits / questions from the previous review, but nothing major.

core/lib/zksync_core/src/genesis.rs
ly0va: Yes, and since this is only called during genesis, that's not a big deal at all

core/lib/zksync_core/src/state_keeper/keeper.rs
ly0va: since it has little to do with the current PR, let's move do it in a separate one

perekopskiy (approved):

PR #904, opened 2024-01-19
chore(docs): Update documentation about our new fee model

mm-zk:
## What ❔

* Documentation about our new fee model

## Why ❔

* With 1.4.1 we changed the fee model, and added more flexibility.

PR #1015, opened 2024-02-05
feat: Validium mode abstraction

ilitteri:
<h2>Initiating the system</h2>
<p>The system is initiated at compile time by running <code>zk init --validium-mode</code>. This flag sets different configurations across the system to convert the system into a Validium operator as described below.</p>
<h2>Not sending <code>pubdata</code> to L1</h2>
<p>L1 batches are committed by the <code>commitBatches</code> method of the <code>Executor.sol</code> contract. It receives a <code>StoredBatchInfo</code> (last committed batch info) and an array of <code>CommitBatchInfo</code> (data needed to commit the batch) as parameters. The pubdata is part of the <code>CommitBatchInfo</code> therefore not sending it to L1 means not including it in said struct.</p>

```solidity
/// @inheritdoc IExecutor
function commitBatches(
    StoredBatchInfo memory _lastCommittedBatchData,
    CommitBatchInfo[] calldata _newBatchesData
) external nonReentrant onlyValidator {
    // Check that we commit batches after last committed batch
    require(s.storedBatchHashes[s.totalBatchesCommitted] == _hashStoredBatchInfo(_lastCommittedBatchData), &quot;i&quot;); // incorrect previous batch data
    require(_newBatchesData.length &gt; 0, &quot;No batches to commit&quot;);

    bytes32 systemContractsUpgradeTxHash = s.l2SystemContractsUpgradeTxHash;
    // Upgrades are rarely done so we optimize a case with no active system contracts upgrade.
    if (systemContractsUpgradeTxHash == bytes32(0) || s.l2SystemContractsUpgradeBatchNumber != 0) {
        _commitBatchesWithoutSystemContractsUpgrade(_lastCommittedBatchData, _newBatchesData);
    } else {
        _commitBatchesWithSystemContractsUpgrade(
            _lastCommittedBatchData,
            _newBatchesData,
            systemContractsUpgradeTxHash
        );
    }

    s.totalBatchesCommitted = s.totalBatchesCommitted + _newBatchesData.length;
}
```

<p>The transaction that executes a call to <code>commitBatches</code> with its calldata is constructed and sent by the server. Particularly, the <code>eth_tx_aggregator</code> constructs it and saves it in the Postgres database and then it is picked up by the <code>eth_tx_manager</code> and sent to the L1 (as explained above).</p>
<p>In our proposal, the calldata saved in the database by the <code>eth_tx_aggregator</code> does not contain the pubdata for every <code>CommitBatchInfo</code> in Validium mode, it does in Rollup mode.</p>
<p><code>CommitBatches</code> is in charge of encoding the calldata for <code>commitBatches</code>, it implements the trait <code>Tokenize</code> to perform such a task. Said implementation iterates over every batch to commit, and wraps them into <code>CommitBatchInfo</code> (Rust counterpart of Solidity’s) which also implements <code>Tokenize</code> which finally encodes it. Here it lays the logic we need to abstract.</p>
<p>To abstract this, we defined the <code>L1BatchCommitGenerator</code> trait one method <code>l1_commit_data(&amp;self, l1_batch_with_metadata: &amp;L1BatchWithMetadata)</code>.</p>

```rust
pub trait L1BatchCommitDataGenerator
where
    Self: std::fmt::Debug + Send + Sync,
{
    fn l1_commit_data(&amp;self, l1_batch_with_metadata: &amp;L1BatchWithMetadata) -&gt; Token;
}
```

<p>For the Rollup mode, we have a <code>RollupModeL1BatchCommitGenerator</code> struct implementing <code>L1BatchCommitGenerator</code> in such a way that <code>l1_commit_data</code> includes the pubdata, and a <code>ValidiumModeL1BatchCommitGenerator</code> struct for Validium mode, not including the pubdata in <code>l1_commit_data</code>.</p>

```rust
#[derive(Debug, Clone)]
pub struct RollupModeL1BatchCommitDataGenerator {}

#[derive(Debug, Clone)]
pub struct ValidiumModeL1BatchCommitDataGenerator {}

impl L1BatchCommitDataGenerator for RollupModeL1BatchCommitDataGenerator {
    fn l1_commit_data(&amp;self, l1_batch_with_metadata: &amp;L1BatchWithMetadata) -&gt; Token {
        Token::Tuple(rollup_mode_l1_commit_data(l1_batch_with_metadata))
    }
}

impl L1BatchCommitDataGenerator for ValidiumModeL1BatchCommitDataGenerator {
    fn l1_commit_data(&amp;self, l1_batch_with_metadata: &amp;L1BatchWithMetadata) -&gt; Token {
        Token::Tuple(validium_mode_l1_commit_data(l1_batch_with_metadata))
    }
}
```

<p><code>validium_mode_l1_commit_data</code> is implemented in such a way that it does not include the pubdata in the <code>CommitBatchInfo</code> encoding</p>

```rust
fn validium_mode_l1_commit_data(l1_batch_with_metadata: &amp;L1BatchWithMetadata) -&gt; Vec&lt;Token&gt; {
    let header = &amp;l1_batch_with_metadata.header;
    let metadata = &amp;l1_batch_with_metadata.metadata;
    let commit_data = vec![
        // `batchNumber`
        Token::Uint(U256::from(header.number.0)),
        // `timestamp`
        Token::Uint(U256::from(header.timestamp)),
        // `indexRepeatedStorageChanges`
        Token::Uint(U256::from(metadata.rollup_last_leaf_index)),
        // `newStateRoot`
        Token::FixedBytes(metadata.merkle_root_hash.as_bytes().to_vec()),
        // `numberOfLayer1Txs`
        Token::Uint(U256::from(header.l1_tx_count)),
        // `priorityOperationsHash`
        Token::FixedBytes(header.priority_ops_onchain_data_hash().as_bytes().to_vec()),
        // `bootloaderHeapInitialContentsHash`
        Token::FixedBytes(
            metadata
                .bootloader_initial_content_commitment
                .unwrap()
                .as_bytes()
                .to_vec(),
        ),
        // `eventsQueueStateHash`
        Token::FixedBytes(
            metadata
                .events_queue_commitment
                .unwrap()
                .as_bytes()
                .to_vec(),
        ),
        // `systemLogs`
        Token::Bytes(metadata.l2_l1_messages_compressed.clone()),
    ];
    commit_data
}
```

<p>In <code>rollup_mode_l1_commit_data</code> the same as in <code>validium_mode_l1_commit_data</code> is being done but the pubdata is added at the end of the commit data.</p>

```rust
fn rollup_mode_l1_commit_data(l1_batch_with_metadata: &amp;L1BatchWithMetadata) -&gt; Vec&lt;Token&gt; {
    let mut commit_data = validium_mode_l1_commit_data(l1_batch_with_metadata);
    commit_data.push(Token::Bytes(utils::construct_pubdata(
        l1_batch_with_metadata,
    )));
    commit_data
}
```

<p>A field <code>l1_batch_commit_data_generator: Arc&lt;dyn L1BatchCommitGenerator&gt;</code> was added to <code>CommitBatchInfo</code> to perform the L1 commit data encoding.</p>

```rust
/// Encoding for `CommitBatchInfo` from `IExecutor.sol`
#[derive(Debug)]
pub struct CommitBatchInfo&lt;'a&gt; {
    pub l1_batch_with_metadata: &amp;'a L1BatchWithMetadata,
    pub l1_batch_commit_data_generator: Arc&lt;dyn L1BatchCommitDataGenerator&gt;,
}

impl&lt;'a&gt; CommitBatchInfo&lt;'a&gt; {
    pub fn new(
        l1_batch_with_metadata: &amp;'a L1BatchWithMetadata,
        l1_batch_commit_data_generator: Arc&lt;dyn L1BatchCommitDataGenerator&gt;,
    ) -&gt; Self {
        Self {
            l1_batch_with_metadata,
            l1_batch_commit_data_generator,
        }
    }
}

impl&lt;'a&gt; Tokenizable for CommitBatchInfo&lt;'a&gt; {
    fn from_token(_token: Token) -&gt; Result&lt;Self, Web3ContractError&gt;
    where
        Self: Sized,
    {
        // Currently there is no need to decode this struct.
        // We still want to implement `Tokenizable` trait for it, so that *once* it's needed
        // the implementation is provided here and not in some other inconsistent way.
        Err(Web3ContractError::Api(Web3ApiError::Decoder(
            &quot;Not implemented&quot;.to_string(),
        )))
    }

    fn into_token(self) -&gt; Token {
        if self
            .l1_batch_with_metadata
            .header
            .protocol_version
            .unwrap()
            .is_pre_boojum()
        {
            pre_boojum_into_token(self.l1_batch_with_metadata)
        } else {
            self.l1_batch_commit_data_generator
                .l1_commit_data(self.l1_batch_with_metadata)
        }
    }
}
```

<p>The same field was also added to the <code>Aggregator</code> and <code>CommitBatches</code> structs. The first one instantiates <code>CommitBatches</code>s and the latter instantiates <code>CommitBatchInfo</code>s.</p>

```rust
// Input required to encode `commitBatches` call.
#[derive(Debug, Clone)]
pub struct CommitBatches {
    pub last_committed_l1_batch: L1BatchWithMetadata,
    pub l1_batches: Vec&lt;L1BatchWithMetadata&gt;,
    pub l1_batch_commit_data_generator: Arc&lt;dyn L1BatchCommitDataGenerator&gt;,
}

#[derive(Debug)]
pub struct Aggregator {
    commit_criteria: Vec&lt;Box&lt;dyn L1BatchPublishCriterion&gt;&gt;,
    proof_criteria: Vec&lt;Box&lt;dyn L1BatchPublishCriterion&gt;&gt;,
    execute_criteria: Vec&lt;Box&lt;dyn L1BatchPublishCriterion&gt;&gt;,
    config: SenderConfig,
    blob_store: Arc&lt;dyn ObjectStore&gt;,
    l1_batch_commit_data_generator: Arc&lt;dyn L1BatchCommitDataGenerator&gt;,
}
```

<p>Which of the mentioned trait implementors will be instantiated as the <code>l1_batch_commit_generator</code> value of the <code>Aggregator</code>'s field will depend on the <code>l1_batch_commit_generator_mode</code> value in the <code>chain.toml</code> file, <code>state_keeper</code> section.</p>

```rust
let l1_batch_commit_generator: Arc&lt;dyn L1BatchCommitGenerator&gt; =
    match state_keeper_config.l1_batch_commit_generator_mode {
        Arc::new(ValidiumModeL1BatchCommitGenerator {})
    } else {
        Arc::new(RollupModeL1BatchCommitGenerator {})
    };
```

<h3>Contracts changes</h3>
<p>The following lines in <code>Executor.sol</code> are being skipped by solpp when the Validium mode flag is set to <code>true</code>:</p>
<ul>
<li><code>bytes32 providedL2ToL1PubdataHash = keccak256(_newBatch.totalL2ToL1Pubdata);</code></li>
<li><code>require(logSender == L2_TO_L1_MESSENGER_SYSTEM_CONTRACT_ADDR, &quot;ln&quot;);</code></li>
<li><code>require(providedL2ToL1PubdataHash == logValue, &quot;wp&quot;);</code></li>
</ul>
<p>We also have to change how the <code>DiamondInit</code> is initialized due to the new fee model changes:</p>
<ul>
<li>The <code>pubdataPricingMode</code> is fixed to <code>PubdataPricingMode.Rollup</code>, this value will depend on the <code>VALIDIUM_MODE</code> env now.</li>
</ul>
<h2>Not charging for <code>pubdata</code></h2>
<p>When running the stack as a Validium, gas-used calculations have to account for the fact that <code>pubdata</code> is no longer published, and thus certain operations like storage become cheaper (they don’t pay for <code>pubdata</code> inclusion on L1). Under the new fee model that was recently merged, every change needed here is managed through the following configuration values. With this in mind, all that’s needed is, when running <code>zk init</code>, if the user is initializing the stack in Validium mode (passing the <code>--validium-mode</code> flag) we should load the following config specified for Validium in the file <code>etc/env/base/chain.toml</code>:</p>
<p><a href="https://github.com/matter-labs/zksync-era/pull/904/files#diff-a9f4b95e25a034628b5e4474de26a3e4bd325d926a8e443d13b8e0859f07731c">chore(docs): Update documentation about our new fee model by mm-zk · Pull Request #904 · matter-labs/zksync-era</a></p>

flag | rollup with calldata | rollup with 4844 (blobs) | value for validium | value for DA
-- | -- | -- | -- | --
l1_pubdata_price | 510'000'000'000 | 3'000'000'000 | 0 | 5'000
max_pubdata_per_batch | 120'000 | 250'000 | 1'000'000'000'000 | 1'000'000
pubdata_overhead_part | 0.7 | 0.4 | 0 | 0.1
compute_overhead_part | 0.5 | 0.7 | 1 | 1
batch_overhead_l1_gas | 1'000'000 | 1'000'000 | 1'000'000 | 1'400'000

<p>The only flag not available in the <code>chain.toml</code> is the <code>l1_pubdata_price</code> as it is computed and set by the function  <code>estimate_effective_pubdata_price()</code> defined in the <code>L1GasPriceProvider</code> trait, implemented by <code>GasAdjuster</code>. To make this value equal to zero, we need to set the const <code>L1_GAS_PER_PUBDATA_BYTE</code> to zero for validium, and 17 for rollup. For this, we’ve added a new field in the <code>GasAdjusterConfig</code> struct which is loaded from the <code>eth_sender.toml</code> file (hence the same field was added there to be read).</p>


RomanBrodetski:
Hmm for some reason in the description the code has some encoding issues:
```
    commit_criteria: Vec&lt;Box&lt;dyn L1BatchPublishCriterion&gt;&gt;,
    proof_criteria: Vec&lt;Box&lt;dyn L1BatchPublishCriterion&gt;&gt;,
```
Can you please paste again?

core/bin/external_node/src/main.rs
```diff
@@ -267,7 +268,11 @@ async fn init_tasks(
         .context("failed to build a tree_pool")?;
     let tree_handle = task::spawn(metadata_calculator.run(tree_pool, tree_stop_receiver));
 
-    let consistency_checker_handle = tokio::spawn(consistency_checker.run(stop_receiver.clone()));
+    let l1_batch_commit_data_generator = Arc::new(RollupModeL1BatchCommitDataGenerator {});
```
popzxc, line 271: I think you should make sure that EN can work as well. EN verifies the pubdata sent to L1, so if main node is in validium mode, EN (in its current form) will probably not work.

core/lib/dal/sqlx-data.json
```diff
@@ -0,0 +1,3 @@
+{
```
popzxc, line 1: Is it required?

core/lib/l1_contract_interface/Cargo.toml
```diff
@@ -13,6 +13,7 @@ readme = "README.md"
 [dependencies]
 zksync_types = { path = "../types" }
 zksync_prover_interface = { path = "../prover_interface" }
+zksync_config = { path = "../config" }
```
popzxc, line 16: Is it used anywhere?

core/lib/types/src/utils.rs
```diff
@@ -77,6 +77,38 @@ pub fn deployed_address_create(sender: Address, deploy_nonce: U256) -> Address {
     Address::from_slice(&keccak256(&bytes)[12..])
 }
 
+/// Packs all pubdata needed for batch commitment in boojum into one bytes array. The packing contains the
+/// following: logs, messages, bytecodes, and compressed state diffs.
+/// This data is currently part of calldata but will be submitted as part of the blob section post EIP-4844.
+pub fn construct_pubdata(l1_batch_with_metadata: &L1BatchWithMetadata) -> Vec<u8> {
```
popzxc, line 83: This was a method on the structure, but now it's a free function that still accepts the same argument (e.g. it was `&self` before, now it's `l1_batch_with_metadata`). I think it slightly worsens the logical structure of the crate (e.g. it was in an expected namespace -- as a method of the corresponding function -- but now it's still in the global namespace, but in an unexpected place).

Is there any strong reasoning for it? If not, I'd suggest to roll it back.

core/lib/types/src/l1_batch_commit_data_generator.rs
```diff
@@ -0,0 +1,74 @@
...
+
+pub trait L1BatchCommitDataGenerator
+where
+    Self: std::fmt::Debug + Send + Sync,
+{
+    fn l1_commit_data(&self, l1_batch_with_metadata: &L1BatchWithMetadata) -> Token;
+}
+
+#[derive(Debug, Clone)]
+pub struct RollupModeL1BatchCommitDataGenerator {}
```
popzxc, line 13: Nit: you can just remove braces for empty structs both here and when you construct it.

core/lib/config/src/configs/chain.rs
```diff
@@ -45,6 +45,18 @@ impl Default for FeeModelVersion {
...
+pub enum L1BatchCommitDataGeneratorMode {
+    Rollup,
+    Validium,
+}
+
+impl Default for L1BatchCommitDataGeneratorMode {
+    fn default() -> Self {
+        Self::Rollup
+    }
+}
```
popzxc, line 58: I've taken a look at the PR to the contracts and it seems that info [on whether we're in validium mode is available on the contract](https://github.com/lambdaclass/era-contracts/pull/14/files#diff-6466768dc18a351a71e348fcf8c13a415a9aeb9d676b6d671ebd61660ab90f53R27).

Can we increase the safety by querying the contract and panicking if the chosen config option doesn't match what's enabled on the contract?

This probably would require you to create a corresponding getter in contracts, but I guess it isn't too hard.

infrastructure/zk/src/init.ts
```diff
@@ -32,6 +32,7 @@ export async function init(initArgs: InitArgs = DEFAULT_ARGS) {
     let envFileContent = fs.readFileSync(process.env.ENV_FILE!).toString();
     envFileContent += `VALIDIUM_MODE=${validiumMode}\n`;
     fs.writeFileSync(process.env.ENV_FILE!, envFileContent);
+    process.env.VALIDIUM_MODE = validiumMode.toString();
```
popzxc, line 35: Is it (and code above) still required given that we now have a proper configuration option?

era-contracts-lambda
popzxc, line 1: PR porting artifact?

ilitteri:
> Hmm for some reason in the description the code has some encoding issues:
> 
> ```
>     commit_criteria: Vec&lt;Box&lt;dyn L1BatchPublishCriterion&gt;&gt;,
>     proof_criteria: Vec&lt;Box&lt;dyn L1BatchPublishCriterion&gt;&gt;,
> ```
> 
> Can you please paste again?

Yes! This was because it was copied from Notion 😅.

core/lib/dal/sqlx-data.json
ilitteri: Nope it isn't, thought it was deleted though, removing it asap.

core/lib/l1_contract_interface/Cargo.toml
ilitteri: Nope, it was used in a previous version where `zksync_config::L1BatchCommitDataGeneratorMode` was being used. Removing it asap.

core/lib/types/src/utils.rs
ilitteri: In terms of design, there are two things: one is that it's a little bit smelly that this was a method for `L1BatchWithMetadata` as it doesn't make a lot of sense for such an entity to be responsible for pubdata construction. Secondly, the responsibility of whether the pubdata is constructed or not was delegated to the new abstraction and we kinda can say again that it doesn't make sense anymore where it is. What do you think?

If this thought is not enough for making such a change, we roll back and let the method where it was and then the abstraction, as it knows the `L1BatchWithMetadata`, could call it without any further issues.

core/lib/config/src/configs/chain.rs
ilitteri: It will be less complex to read the `VALIDIUM_MODE` env as it is the same that the contract reads (no more contract logic is needed nor making a query to it). Any thoughts on this? This will also reduce the code complexity as the config and enum won't be required any longer.

infrastructure/zk/src/init.ts
ilitteri: This case is very particular. That line is there only because the L1 contracts (`Executor.sol`) need that variable set before compilation. I really thought about improving here but nothing came to my mind. Let me know if, now that you have a complete context of the design decision, you come up with a new strategy for this.

era-contracts-lambda
ilitteri: Also a thing to remove, I'll handle it.

core/lib/config/src/configs/chain.rs
popzxc: The contracts are deployed and the server is run usually by different people at different times (and moreover, configs for server can change over time), so it's not guaranteed that this env variable would have the same value all the time, so this would serve as a defensive programming measure to guarantee that we won't ever run server in incompatible mode by accident.

RomanBrodetski: hmm @ilitteri so what you mean is to read env directly from one of the `L1BatchCommitDataGeneratorMode` structs and make sure that e.g. if Rollup one is instantiated, the env var is set to `Rollup`? But we define which `L1BatchCommitDataGeneratorMode` to create based on this very config variable! I mean, we could mess up env parsing into variable, but it's not worth it to check again in run-time :)

What @popzxc suggests is to panic if we mistakenly run server configured in rollup mode against contracts that are validium.

ilitteri: > hmm @ilitteri so what you mean is to read env directly from one of the L1BatchCommitDataGeneratorMode structs and make sure that e.g. if Rollup one is instantiated, the env var is set to Rollup? But we define which L1BatchCommitDataGeneratorMode to create based on this very config variable! I mean, we could mess up env parsing into a variable, but it's not worth it to check again in run-time :)

I'm saying that rolling back to [this](https://github.com/lambdaclass/zksync-era/pull/75#discussion_r1474245028) will be easier in terms of complexity (if this solution fits), and given that there's only one `VALIDIUM_MODE` env variable, we are sure that its value is the same for the contracts and the server. This way we also do not need to panic because this env variable is set in the `init.ts`. 

Let me know if I'm missing something and if there's still a chance of having mixed configurations.

popzxc: I think our point is that `VALIDIUM_MODE` may have different values at different times (e.g. you could deploy contracts with `VALIDIUM_MODE=true` and then run the server with `VALIDIUM_MODE=false` by accident, which would be a mistake during the configuration.

So the proposal here is not about simplification, but rather about making the code more resistant to incorrect configuration. You will still have an option for the mode (and it doesn't matter much how is it called) -- it will express the desire of the node operator to run the node with a certain config, but at the same time the node itself will check whether it was properly configured and will not let the node to run in an incorrect mode.

infrastructure/zk/src/init.ts
popzxc: Wouldn't it be better to set this variable in the script that deploys the corresponding contract, and pass it to the script from init via a CLI argument?
Behavior that depends on some environment variable is efficiently a global state, so by passing it via CLI you'll make the process more explicit, and the env variable would only exist in a much more local context.

core/lib/types/src/utils.rs
popzxc: > In terms of design, there are two things: one is that it's a little bit smelly that this was a method for L1BatchWithMetadata as it doesn't make a lot of sense for such an entity to be responsible for pubdata construction. 

While I certainly agree with this statement, I'm still not convinced that a free function in a generally random module would express the domain adherence better. If we are to consider separation of interfaces and/or domain knowledge, I think it has to do with better encapsulation in the first place, and `utils` module probably doesn't provide any improvements on that front.

> Secondly, the responsibility of whether the pubdata is constructed or not was delegated to the new abstraction and we kinda can say again that it doesn't make sense anymore where it is. What do you think?

Given the above, I think that _majorly_ the design hasn't changed, the overall logic is still the same: it's either to call this method or not to call this method.

---

So, to summarize, I feel like this function was moved from one inappropriate place to another inappropriate place. I'd suggest to roll it back for now. But to your point, it certainly makes sense to revisit this place better to find a better logical structure.

core/lib/config/src/configs/chain.rs
StanislavBreadless: Just in case, @ilitteri if you do decide to create such a getter, please make sure that it is marked as unstable in the comments. In the future we might wanna allow different chains to have different DA layers (i.e. not a binary "Rollup"/"Validium") and so a different method so a different double checking method will be required.

Though I am all for adding a temporary getter for double checking to avoid mistakes

core/lib/types/src/l1_batch_commit_data_generator.rs
```diff
@@ -0,0 +1,74 @@
...
+    }
+}
+
+impl L1BatchCommitDataGenerator for ValidiumModeL1BatchCommitDataGenerator {
+    fn l1_commit_data(&self, l1_batch_with_metadata: &L1BatchWithMetadata) -> Token {
+        Token::Tuple(validium_mode_l1_commit_data(l1_batch_with_metadata))
+    }
+}
+
+fn validium_mode_l1_commit_data(l1_batch_with_metadata: &L1BatchWithMetadata) -> Vec<Token> {
```
StanislavBreadless, line 30: You still need the `bytes pubdata` at the end of the encoding. It can have zero length, but you still need it. Otherwise, the ABI encoding wont match

core/lib/config/src/configs/chain.rs
```diff
@@ -45,6 +45,18 @@ impl Default for FeeModelVersion {
     }
 }
 
+#[derive(Debug, Clone, Copy, Deserialize, PartialEq, Eq)]
+pub enum L1BatchCommitDataGeneratorMode {
+    Rollup,
+    Validium,
+}
+
+impl Default for L1BatchCommitDataGeneratorMode {
```
slowli, line 54: Nit: You can derive `Default` and add `#[default]` attribute for the `Rollup` variant in order to be more concise and idiomatic.

core/lib/l1_contract_interface/src/i_executor/methods/commit_batches.rs
```diff
@@ -10,6 +15,7 @@ use crate::{
 pub struct CommitBatches {
     pub last_committed_l1_batch: L1BatchWithMetadata,
     pub l1_batches: Vec<L1BatchWithMetadata>,
+    pub l1_batch_commit_data_generator: Arc<dyn L1BatchCommitDataGenerator>,
```
slowli, line 18: Bikeshedding: IMO, such an interface looks bad. Previously, `CommitBatches` was plain old data, but now it includes behavior. Dumb question: Why is `L1BatchCommitDataGenerator` needed in the first place? Don't its implementations correspond 1-to-1 to `L1BatchCommitDataGeneratorMode`? If so, wouldn't it make sense to pass `L1BatchCommitDataGeneratorMode` instead and attach logic currently present in the `L1BatchCommitDataGenerator` trait to it?

core/lib/zksync_core/src/consistency_checker/mod.rs
```diff
@@ -248,7 +252,11 @@ impl ConsistencyChecker {
             .await?)
     }
 
-    pub async fn run(mut self, mut stop_receiver: watch::Receiver<bool>) -> anyhow::Result<()> {
+    pub async fn run(
+        mut self,
+        mut stop_receiver: watch::Receiver<bool>,
+        l1_batch_commit_data_generator: Arc<dyn L1BatchCommitDataGenerator>,
```
slowli, line 258: Please add parameters to the `ConsistencyChecker` structure itself, rather than to the `run()` method.

core/lib/zksync_core/src/eth_sender/publish_criterion.rs
```diff
@@ -212,16 +216,20 @@ impl L1BatchPublishCriterion for DataSizeCriterion {
         _storage: &mut StorageProcessor<'_>,
         consecutive_l1_batches: &[L1BatchWithMetadata],
         _last_sealed_l1_batch: L1BatchNumber,
+        l1_batch_commit_data_generator: Arc<dyn L1BatchCommitDataGenerator>,
```
slowli, line 219: Please move `l1_batch_commit_data_generator` to `DataSizeCriterion` so that it doesn't pollute `L1BatchPublishCriterion` interface.

core/lib/zksync_core/src/eth_sender/tests.rs
```diff
@@ -477,7 +510,14 @@ async fn three_scenarios() -> anyhow::Result<()> {
 #[tokio::test]
 async fn failed_eth_tx() {
     let connection_pool = ConnectionPool::test_pool().await;
-    let mut tester = EthSenderTester::new(connection_pool.clone(), vec![100; 100], false).await;
+    let l1_batch_commit_data_generator = Arc::new(RollupModeL1BatchCommitDataGenerator {});
```
slowli, line 513: Dumb question: here, and in quite a few other tests you pass "hard-coded" `RollupModeL1BatchCommitDataGenerator` as a parameter throughout the test. Do you plan to adapt these tests to run with the other generator as well? If not, it could be inlined down the call chain (e.g., into the `Aggregator::new()` constructor in `EthSenderTester::new()` in this case), so that the changeset is smaller and the PR is easier to review. In this particular case, a comment right above `Aggregator::new()` says "Aggregator - unused", so I'm not sure having a generator as an argument would even make sense.

core/lib/l1_contract_interface/src/i_executor/methods/commit_batches.rs
popzxc: This would likely change. The current proposal is as follows (the comment went missing for some reason on this PR, so I shared it to Lambda directly):
image

core/lib/types/src/l1_batch_commit_data_generator.rs
ilitteri: You're right! We filed an issue last week and will tackle it this week!

core/lib/l1_contract_interface/src/i_executor/methods/commit_batches.rs
mationorato: Hello @popzxc , thank you for taking the time to review the PR and provide your suggestions. They all make sense to us too, and we've already filed an issue to track them, along with other suggestions from you and others. While most of them have been addressed, some are significant changes that we believe go beyond the scope of what we've discussed for this initial release.

We intend to revisit these improvements after this PR is merged, as the Validium feature will continue its development after the first delivery. We will not be closing the feature development.

core/lib/types/src/utils.rs
ilitteri: Tracked in https://github.com/lambdaclass/zksync-era/issues/141

core/lib/config/src/configs/chain.rs
ilitteri: Coming back to this conversation, we've recently merged this PR https://github.com/lambdaclass/zksync-era/pull/138 which makes the `zk init` command update the config files automatically correspondingly to whether the `--validium-mode` flag is set or not, leaving the possibility of having mix configurations to the only case of deliberately changing the `dev.env` file before running `zk server`. The `VALIDIUM_MODE` env variable is only used for solpp in the L1 contracts compilation. There is no way to incorrectly set up the whole stack using the `zk` tool.

infrastructure/zk/src/init.ts
ilitteri: Indeed it is, we'll update it asap.

core/lib/types/src/l1_batch_commit_data_generator.rs
ilitteri: Solved in this PR https://github.com/matter-labs/zksync-era/pull/1015/commits/d31d1a66e01a143136068b7f189f86963110d1a9

ilitteri: Solved here https://github.com/lambdaclass/zksync-era/pull/136

core/lib/zksync_core/src/consistency_checker/mod.rs
ilitteri: Tracked in https://github.com/lambdaclass/zksync-era/issues/132

core/lib/zksync_core/src/eth_sender/publish_criterion.rs
ilitteri: Tracked in https://github.com/lambdaclass/zksync-era/issues/133

core/lib/zksync_core/src/eth_sender/tests.rs
ilitteri: They have been merged already, resolving this conversation

core/lib/config/src/configs/chain.rs
ilitteri: Addressed in [34f9f37](https://github.com/matter-labs/zksync-era/pull/1015/commits/34f9f37e3bb212277fad6af3fee04f6e6c6f2db2)

infrastructure/zk/src/init.ts
```diff
@@ -138,6 +140,92 @@ export async function submoduleUpdate() {
     await utils.exec('git submodule update');
 }
 
+function updateConfigFile(path: string, modeConstantValues: Record<string, number | string | null>) {
```
arose0, line 143: This approach looks a little suspicious, like we're cleaning something up in a way that feels fragile and/or not maintainable (will we remember why we're doing all of this in a few months?). 

I haven't looked super carefully here but would like to understand what we're doing here better.

core/lib/config/src/configs/chain.rs
popzxc: @ilitteri the `zk` tool is only used to initialize the network for the first run. However, once the system in production, configs live elsewhere (not in `dev.env`) and are often edited. For example, we have dedicated infrastructure for managing configuration of our environments. So the problem we're discussing may occur *long after* the chain is initialized when it's in production. The check, if added, would happen each time the node is restarted (and you have to restart the node when you change configuration), so it'll act as a circuit breaker not allowing it to start if somehow somebody changed the configuration to an incorrect setting.

mationorato:
I will close this as the merge was done here #1461

PR #1016, opened 2024-02-05
feat: Add `zks_getBatchPubdata` endpoint

ilitteri:
## Description

This PR adds an endpoint to the RPC that'll let a user like Validium's DA Manager get some batch's pubdata to store in some DA Layer.

## How to test

1. Run the hole stack (`zk && zk clean --all && zk init`)
2. Run the following `curl`
```
curl http://localhost:3050  -X POST \
  --data '{"method":"zks_getBatchPubdata","params":[1],"id":1,"jsonrpc":"2.0"}'
```


core/lib/dal/src/blocks_dal.rs
```diff
@@ -2148,6 +2148,21 @@ impl BlocksDal<'_, '_> {
         .await?
         .map(|row| row.virtual_blocks as u32))
     }
+
+    pub async fn get_batch_pubdata(
```
slowli, line 2152: Bikeshedding: I feel that DAL should not be responsible for domain logic (mapping L1 batch metadata to pubdata in this case). It would be better to use `get_l1_batch_metadata()` in the endpoint handler directly.

core/lib/zksync_core/src/api_server/web3/namespaces/zks.rs
```diff
@@ -639,4 +639,25 @@ impl ZksNamespace {
...
+        const METHOD_NAME: &str = "get_batch_pubdata";
+
+        let method_latency = API_METRICS.start_call(METHOD_NAME);
+        self.state.start_info.ensure_not_pruned(l1_batch_number)?;
+        let mut storage = self.access_storage(METHOD_NAME).await?;
+        let pubdata = storage
+            .blocks_dal()
+            .get_batch_pubdata(l1_batch_number)
+            .await
+            .map_err(|_err| Web3Error::PubdataNotFound)?
```
slowli, line 657: Why not return `None` in this case? Most other methods both in this namespace and in `eth` one return `None` in case of missing data.

core/lib/dal/src/blocks_dal.rs
ilitteri: LGTM, I'll file an issue in our fork to tackle this. Thanks!

core/lib/zksync_core/src/api_server/web3/namespaces/zks.rs
ilitteri: Can do that to keep this method consistent.

core/lib/dal/src/blocks_dal.rs
ilitteri: We created this issue to tackle this https://github.com/lambdaclass/zksync-era/issues/146

ilitteri:
@slowli, I solved both suggestions, let me know if there's anything else to do for this PR.

core/lib/web3_decl/src/error.rs
```diff
@@ -41,4 +41,6 @@ pub enum Web3Error {
     InvalidFilterBlockHash,
     #[error("Tree API is not available")]
     TreeApiUnavailable,
+    #[error("Pubdata not found for given batch id")]
+    PubdataNotFound,
```
slowli, line 45: This variant doesn't seem to be used any more; if so, please remove.

core/lib/web3_decl/src/namespaces/zks.rs
```diff
@@ -118,4 +118,8 @@ pub trait ZksNamespace {
         keys: Vec<H256>,
         l1_batch_number: L1BatchNumber,
     ) -> RpcResult<Proof>;
+
+    #[method(name = "getBatchPubdata")]
+    async fn get_batch_pubdata(&self, l1_batch_number: L1BatchNumber)
+        -> RpcResult<Option<Vec<u8>>>;
```
slowli, line 124: Please use `zksync_type::Bytes` instead of `Vec<u8>` so that the returned value is hex-encoded rather than being a JSON array of `u8`s.

core/lib/zksync_core/src/api_server/web3/namespaces/zks.rs
```diff
@@ -639,4 +640,25 @@ impl ZksNamespace {
...
+        const METHOD_NAME: &str = "get_batch_pubdata";
+
+        let method_latency = API_METRICS.start_call(METHOD_NAME);
+        self.state.start_info.ensure_not_pruned(l1_batch_number)?;
+        let mut storage = self.access_storage(METHOD_NAME).await?;
+        let l1_batch_with_metadata = storage
+            .blocks_dal()
+            .get_l1_batch_metadata(l1_batch_number)
+            .await
+            .unwrap_or(None);
```
slowli, line 658: This doesn't look right; DB-level errors should not be swallowed like this, but rather should be converted to internal server errors. See other methods in this module for the reference.

slowli (approved):
Not sure whether CI failure is expected, but AFAIU it's caused by the base branch; the changes in this PR look fine.

popzxc:
@ilitteri let's keep this on hold until we merge validium mode abstraction. I want us to have a round of discussion on 3rd party DA first -- maybe we'll change the design.

popzxc:
Closing as stale.

PR #1018, opened 2024-02-05
feat: Validium DA Manager example

ilitteri:
> DISCLAIMER: This PR makes use of #1016 changes, if you want to run the example please follow the instructions of [our internal PR](https://github.com/lambdaclass/zksync-era/pull/80).

## DA Manager (separate binary)

A `DA Manager` is introduced, a component that periodically queries for new batches and uploads their `pubdata` to some data availability solution, as a separate binary. The way this works is the following:

- To construct the `pubdata`, the operator will query Postgres for the four components of pubdata, then convert it into its byte representation using the logic in the
`pub fn construct_pubdata(&self) -> Vec<u8>` method of the `L1BatchWithMetadata` struct.
- A user wanting to incorporate a new DA solution simply writes their own DA manager in the programming language they prefer, as an application that uses the endpoint introduced above to periodically query for new `pubdata` and post it to the DA server. For this scenario, we should provide some examples showing what you should implement for your own DA solution.
- This `DA Manager` runs independently, from the operator, i.e. as an entirely separate `OS` process or even on a different machine (in particular, it’s independent from both the `eth_tx_aggregator` and the `eth_tx_manager`)


deniallugo-ml:

PR #1020, opened 2024-02-06
fix(prover): QoL socket utilization

EmilLuta:
## What ❔

This PR adds necessary timeout for WVG <-> prover socket communication. On top of it, also improves the debugging experience by adding logs on what connections are being established to what IPs.

## Why ❔

Without the timeout, we could have WVGs stuck for as long as the connection takes. In practice, we've observed over 30 minutes.

This will limit such instances to ~1 minute (10 retries * 6 seconds timeout). Whilst this is a duct-tape, it will do until we get around implementing the long-term solution.

## Checklist

### DIFF
--- a/prover/prover_fri_utils/src/socket_utils.rs
+++ b/prover/prover_fri_utils/src/socket_utils.rs
@@ -18,6 +18,7 @@ pub fn send_assembly(
     let started_at = Instant::now();
     let mut error_messages = vec![];
 
+    // 10 attempts, each waiting for 6 seconds => 1 minute wait time at most
     for _ in 0..10 {
         match TcpStream::connect_timeout(socket_address, Duration::from_secs(6)) {
             Ok(mut stream) => {


prover/prover_fri_utils/src/socket_utils.rs
```diff
@@ -19,7 +19,7 @@ pub fn send_assembly(
     let mut error_messages = vec![];
 
     for _ in 0..10 {
-        match TcpStream::connect(socket_address) {
+        match TcpStream::connect_timeout(socket_address, Duration::from_secs(6)) {
```
popzxc, line 22: Nit: In the PR description you mention that we limit the time to 1 minute here, and I think it'd make a great comment. I first looked at diff and wondered "Why 6 seconds though?", then looked at the PR description, and then it all made sense.

popzxc (dismissed):

popzxc (approved):

PR #1022, opened 2024-02-06
fix(metrics): Use latest block for non pos ethereum as safe

Deniallugo:
## What ❔

We have a hack and if we set number of confirmations, it's possible to use our network in PoW Ethereum, but it wasn't the case for our metrics. 

## Why ❔

We need to run zk stack locally in PoW Ethereum 

## Checklist


zk-Lumi (commented):
Fixes issue for me with the `safe` block height error in local node  👍

montekki (approved):

PR #1024, opened 2024-02-07
feat(commitment): Commitment component

perekopskiy:
## What ❔

- Adding new component that calculate commitment for L1 batches.
- Metadata calculator now only calculates things that require tree, i.e. root hash and merkle paths in case full tree mode is on.
- Soft removal of some `l1_batches` columns: `parent_hash`, `l2_l1_compressed_messages`, `compressed_write_logs`,  `compressed_contracts`.
- Fixed field `factory_deps` of `L1BatchWithMetadata`, it's renamed to `raw_published_factory_deps`, now it represents all published bytecodes for pre-boojum batches and published uncompressed bytecodes for post-boojum batches (compressed bytecodes are published as a part of long L2 to L1 messages).

## Why ❔

- To follow the single responsibility principle
- To improve the performance of full tree which directly affects EN performance

## Checklist


core/lib/types/src/commitment.rs
```diff
@@ -577,181 +573,134 @@ impl L1BatchCommitment {
...
+}
+
+#[cfg(test)]
+mod tests {
+    use serde::{Deserialize, Serialize};
+
+    use super::{
+        CommitmentInput, L1BatchAuxiliaryOutput, L1BatchCommitment, L1BatchCommitmentHash,
+        L1BatchMetaParameters, L1BatchPassThroughData,
+    };
```
slowli, line 665: Nit: It's acceptable to `use super::*` in tests.

core/lib/types/src/commitment.rs
```diff
@@ -218,185 +250,194 @@ impl SerializeCommitment for StateDiffRecord {
...
-    system_logs_linear_hash: H256,
-    state_diffs_hash: H256,
-    state_diffs_compressed: Vec<u8>,
-    bootloader_heap_hash: H256,
-    events_state_queue_hash: H256,
     protocol_version: ProtocolVersionId,
 }
 
+/// Block Output produced by Virtual Machine
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
```
slowli, line 260: Are `Serialize` / `Deserialize` derives here and in other types in this module only required for tests? If so, I'd suggest making this clear:

```rust
#[derive(Debug, Clone, Eq, PartialEq)]
#[cfg_attr(test, derive(Serialize, Deserialize))]
enum L1BatchAuxiliaryOutput {
     // ...
}
```

...so that the serialization format doesn't get relied upon somewhere eventually.

core/lib/types/src/commitment.rs
```diff
@@ -218,185 +250,194 @@ impl SerializeCommitment for StateDiffRecord {
...
+        common: L1BatchAuxiliaryCommonOutput,
+        system_logs_linear_hash: H256,
+        state_diffs_compressed: Vec<u8>,
+        state_diffs_hash: H256,
+        aux_commitments: AuxCommitments,
+    },
+}
+
 impl L1BatchAuxiliaryOutput {
     #[allow(clippy::too_many_arguments)]
```
slowli, line 280: Nit: This attribute is obsolete.

core/lib/types/src/commitment.rs
```diff
@@ -577,181 +573,134 @@ impl L1BatchCommitment {
...
+        hashes: L1BatchCommitmentHash,
     }
 
-    #[test]
-    fn commitment_test() {
+    fn run_test(test_name: &str) {
         let zksync_home = std::env::var("ZKSYNC_HOME").unwrap_or_else(|_| ".".into());
         let path = std::path::Path::new(&zksync_home)
-            .join("etc/commitment_tests/zksync_testharness_test.json");
+            .join(format!("etc/commitment_tests/{test_name}.json"));
```
slowli, line 679: Sort of off-topic: Specifying paths outside of the package makes it really difficult to publish it because package tests are no longer even theoretically self-contained. (I believe we have other places with similar issues, but the less such places, the better.) In this case, it would be easy to move (or symlink, if they are used anywhere else) the necessary files inside the package directory.

core/lib/types/src/commitment.rs
```diff
@@ -577,181 +573,134 @@ impl L1BatchCommitment {
...
-        pass_through_bytes: Vec<u8>,
-        pass_through_hash: H256,
-        #[serde_as(as = "serde_with::hex::Hex")]
-        meta_params_bytes: Vec<u8>,
-        meta_params_hash: H256,
-        #[serde_as(as = "serde_with::hex::Hex")]
-        auxiliary_bytes: Vec<u8>,
-        auxiliary_hash: H256,
-        commitment_hash: H256,
+    pub fn artifacts(&self) -> L1BatchCommitmentArtifacts {
```
slowli, line 577: Dumb question: Can't `L1BatchCommitment` be used in `save_l1_batch_commitment_artifacts()` directly (IIUC, that's the only non-test use of `L1BatchCommitmentArtifacts`)? Would it break an abstraction boundary?

core/lib/zksync_core/src/commitment_generator/metrics.rs
```diff
@@ -0,0 +1,29 @@
...
+    Calculate,
+    SaveResults,
+}
+
+/// Metrics for the commitment generator.
+#[derive(Debug, Metrics)]
+#[metrics(prefix = "server_commitment_generator")]
+pub(super) struct CommitmentGeneratorMetrics {
+    /// Latency of generating commitment per stage.
+    #[metrics(buckets = Buckets::LATENCIES)]
```
slowli, line 18: Nit: It's better to specify unit of measurement for new metrics: `unit = Unit::Seconds` (here, `Unit` should be imported into the scope from `vise`). We don't add UoM for the existing metrics because it influences metric name (as per the Open Metrics spec).

core/lib/zksync_core/src/commitment_generator/mod.rs
```diff
@@ -0,0 +1,235 @@
...
+        let events_queue = conn
+            .blocks_dal()
+            .get_events_queue(l1_batch_number)
+            .await?
+            .context("Events queue is required for post-boojum batch")?;
+        let events_queue_commitment = events_queue_commitment(&events_queue, protocol_version)
+            .context("Events queue commitment is required for post-boojum batch")?;
+        METRICS
+            .events_queue_commitment_latency
+            .observe(started_at.elapsed());
```
slowli, line 50: Nit: You can use

```rust
let latency = METRICS.events_queue_commitment_latency.start();
// ...
latency.observe();
```

core/lib/zksync_core/src/commitment_generator/mod.rs
```diff
@@ -0,0 +1,235 @@
...
+            connection_pool,
+            stop_receiver,
+        }
+    }
+
+    async fn calculate_aux_commitments(
+        conn: &mut StorageProcessor<'_>,
+        l1_batch_number: L1BatchNumber,
+        protocol_version: ProtocolVersionId,
+    ) -> anyhow::Result<Option<AuxCommitments>> {
```
slowli, line 39: Does this method ever return `Ok(None)`? If not, the return type can be simplified.

core/lib/zksync_core/src/commitment_generator/mod.rs
```diff
@@ -0,0 +1,235 @@
...
+        };
+        let touched_slots = connection
+            .storage_logs_dal()
+            .get_touched_slots_for_l1_batch(l1_batch_number)
+            .await?;
+        let touched_hashed_keys: Vec<_> =
+            touched_slots.keys().map(|key| key.hashed_key()).collect();
+        let previous_values = connection
+            .storage_logs_dal()
+            .get_previous_storage_values(&touched_hashed_keys, l1_batch_number)
```
slowli, line 106: Dumb question: In metadata calculator, `get_previous_storage_values()` is intentionally avoided (it's sort of an expensive query) in favor of using protective reads to remove "not really touched" slots. Would the same approach work here? Or is it too complex relative to potential performance benefits?

slowli (commented):
Will try to perform more in-depth review tomorrow.

core/lib/dal/src/blocks_dal.rs
```diff
@@ -724,161 +746,120 @@ impl BlocksDal<'_, '_> {
...
-                rollup_last_leaf_index = $9,
-                pass_through_data_hash = $10,
-                meta_parameters_hash = $11,
-                compressed_state_diffs = $12,
+                commitment = $1,
+                aux_data_hash = $2,
+                pass_through_data_hash = $3,
+                meta_parameters_hash = $4,
+                l2_l1_merkle_root = $5,
+                zkporter_is_available = $6,
```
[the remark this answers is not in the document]
perekopskiy: Yes, it can be saved when L1 batch sealed but for some reason it was saved with metadata previously. This field is never used when it's read from DB, so I thought about removing it but decided to do it separately to keep diff smaller. It doesn't matter when it's saved and I plan to remove it completely soon, so if you don't mind I will not change this behaviour in this PR

core/lib/dal/src/blocks_dal.rs
```diff
@@ -887,32 +868,45 @@ impl BlocksDal<'_, '_> {
...
-                metadata.root_hash,
-                metadata.l2_l1_merkle_root
+                commitment_artifacts.commitment_hash.commitment
             );
         }
+
+        sqlx::query!(
+            r#"
+            INSERT INTO
+                commitments (l1_batch_number, events_queue_commitment, bootloader_initial_content_commitment)
```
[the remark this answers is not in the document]
perekopskiy: Yes, that's the plan

core/lib/zksync_core/src/commitment_generator/mod.rs
perekopskiy: Good catch, yes, it doesn't return None, I'll remove `Option`

core/lib/zksync_core/src/commitment_generator/mod.rs
```diff
@@ -0,0 +1,230 @@
...
+                }
+            }
+            state_diffs.sort_unstable_by_key(|rec| (rec.address, rec.key));
+
+            CommitmentInput::PostBoojum {
+                common,
+                system_logs: header.system_logs,
+                state_diffs,
+                aux_commitments,
+            }
```
RomanBrodetski, line 174: would it make sense to extract this (and pre-bojum) to a method? Or maybe even a small method of `StateDiffRecord`?

Not sure it's worth it, just wanted you to consider this

RomanBrodetski (approved):
Looks good!

core/lib/dal/migrations/20240126114931_add-l1_batches-commitment-index.up.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX IF NOT EXISTS l1_batches_commitment ON l1_batches (commitment);
```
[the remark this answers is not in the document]
perekopskiy: Query `get_next_l1_batch_ready_for_commitment_generation` needs it. I just reused the same approach we use in metadata calculator and eth sender, we already have indices for `hash`, `eth_commit/prover/execute_tx_id`

core/lib/zksync_core/src/commitment_generator/mod.rs
perekopskiy: Makes sense

perekopskiy: Actually, just tried it and the code doesn't look much better, so I would probably leave it the current way

StanislavBreadless (approved):

core/lib/types/src/commitment/mod.rs
```diff
@@ -0,0 +1,666 @@
...
+                unsorted_factory_deps,
+                raw_published_bytecode_hashes,
+            ),
+            header,
+            metadata,
+        }
+    }
+
+    /// Iterates over factory deps in the order in which they appeared in this L1 batch.
+    pub fn factory_deps_in_appearance_order(
```
slowli, line 127: Nit: This function can be private (unless it's going to be use somewhere in the future).

core/lib/zksync_core/src/commitment_generator/mod.rs
```diff
@@ -0,0 +1,230 @@
...
+
+        let latency = METRICS.bootloader_content_commitment_latency.start();
+        let initial_bootloader_contents = conn
+            .blocks_dal()
+            .get_initial_bootloader_heap(l1_batch_number)
+            .await?
+            .context("Bootloader initial heap is missing")?;
+        let bootloader_initial_content_commitment =
+            bootloader_initial_content_commitment(&initial_bootloader_contents, protocol_version)
+                .context("Bootloader content commitment is required for post-boojum batch")?;
```
slowli, line 58: Given that calculating a commitment takes ~2s, it could make sense to refactor this method into 2 parts: loading data from Postgres (after which the DB connection is unused and may be dropped) and computations. Theoretically, computations could be put into `spawn_blocking()` task, but I don't see much benefit from it, TBH.

core/lib/zksync_core/src/commitment_generator/mod.rs
```diff
@@ -0,0 +1,230 @@
...
+                let hashed_key = key.hashed_key();
+                if previous_values[&hashed_key].unwrap_or_default() != value {
+                    let (initial_write_l1_batch_number, index) =
+                        l1_batches_for_initial_writes[&hashed_key];
+                    if initial_write_l1_batch_number == l1_batch_number {
+                        initial_writes.push(InitialStorageWrite {
+                            index,
+                            key: key.hashed_key_u256(),
+                            value,
+                        })
```
slowli, line 122: Nit: Missing semicolon here and after `repeated_writes.push`.

core/lib/zksync_core/src/commitment_generator/mod.rs
```diff
@@ -0,0 +1,230 @@
...
+                    } else {
+                        repeated_writes.push(RepeatedStorageWrite { index, value })
```
slowli, line 124: Maybe it makes sense to do a sanity check (or a condition if we're not sure it'll always hold) that `initial_write_l1_batch_number < l1_batch_number`? Ditto for post-Boojum batches.

core/lib/zksync_core/src/commitment_generator/mod.rs
```diff
@@ -0,0 +1,230 @@
...
+            .blocks_dal()
+            .save_l1_batch_commitment_artifacts(l1_batch_number, &artifacts)
+            .await?;
+        latency.observe();
+
+        Ok(())
+    }
+
+    pub async fn run(self) -> anyhow::Result<()> {
+        let mut connection = self
```
slowli, line 207: AFAIU, having a single long-living connection for the entire duration of the app lifetime is an anti-pattern; it's better to have shorter-lived connections, so that pooled connections can be utilized more efficiently.

core/lib/zksync_core/src/eth_sender/eth_tx_aggregator.rs
```diff
@@ -383,12 +385,25 @@ impl EthTxAggregator {
...
+                METRICS.pubdata_size[&PubdataKind::StateDiffs]
+                    .observe(batch.metadata.state_diffs_compressed.len());
+                METRICS.pubdata_size[&PubdataKind::UserL2ToL1Logs]
+                    .observe(batch.header.l2_to_l1_logs.len() * UserL2ToL1Log::SERIALIZED_SIZE);
+                METRICS.pubdata_size[&PubdataKind::LongL2ToL1Messages].observe(
+                    batch
+                        .header
+                        .l2_to_l1_messages
+                        .iter()
+                        .map(|message| message.len())
```
slowli, line 397: Nit: You can use `.map(Vec::len)` to be shorter / more idiomatic. Ditto below.

core/lib/zksync_core/src/genesis.rs
```diff
@@ -101,24 +104,36 @@ pub async fn ensure_genesis_state(
...
-        H256::zero(),
-        H256::zero(),
-        *protocol_version,
-    );
+        rollup_root_hash: genesis_root_hash,
+        bootloader_code_hash: base_system_contracts_hashes.bootloader,
+        default_aa_code_hash: base_system_contracts_hashes.default_aa,
+        protocol_version: *protocol_version,
+    };
+    let commitment_input = if protocol_version.is_pre_boojum() {
```
slowli, line 115: Nit: Maybe it makes sense to have constructor `CommitmentInput::for_genesis_batch(CommitmentCommonInput)` that would encapsulate this logic?

core/lib/zksync_core/src/metadata_calculator/mod.rs
```diff
@@ -184,87 +176,4 @@ impl MetadataCalculator {
...
-        let unsorted_factory_deps = storage
-            .blocks_dal()
-            .get_l1_batch_factory_deps(header.number)
-            .await
-            .unwrap();
-        let commit_gas_cost =
-            commit_gas_count_for_l1_batch(header, &unsorted_factory_deps, metadata);
-        storage
-            .blocks_dal()
-            .update_predicted_l1_batch_commit_gas(header.number, commit_gas_cost)
```
slowli, line 206: Dumb question: Is this update no longer required? If so, does it make sense to remove `update_predicted_l1_batch_commit_gas()` from DAL?

core/lib/zksync_core/src/metadata_calculator/updater.rs
```diff
@@ -125,40 +124,14 @@ impl TreeUpdater {
...
-
             let save_postgres_latency = METRICS.start_stage(TreeUpdateStage::SavePostgres);
-            let is_pre_boojum = header
-                .protocol_version
-                .map(|v| v.is_pre_boojum())
-                .unwrap_or(true);
+            let tree_data = L1BatchTreeData {
+                hash: metadata.root_hash,
+                rollup_last_leaf_index: metadata.rollup_last_leaf_index,
+            };
```
slowli, line 131: Since `TreeMetadata.initial_writes` are only used for consistency checking and `repeated_writes` are not used at all (other than in tests), maybe it makes sense removing them? Could be done separately, of course.

core/lib/zksync_core/src/commitment_generator/mod.rs
perekopskiy: I see value in it, I will also make these two (events and bootloader) commitments to be calculated concurrently

core/lib/zksync_core/src/metadata_calculator/updater.rs
perekopskiy: Agree, I created a task for that

core/lib/zksync_core/src/commitment_generator/mod.rs
```diff
@@ -0,0 +1,264 @@
...
+        let events_queue = connection
+            .blocks_dal()
+            .get_events_queue(l1_batch_number)
+            .await?
+            .context("Events queue is required for post-boojum batch")?;
+        let initial_bootloader_contents = connection
+            .blocks_dal()
+            .get_initial_bootloader_heap(l1_batch_number)
+            .await?
+            .context("Bootloader initial heap is missing")?;
```
slowli, line 53: You can drop `connection` after here so that it isn't held during computationally expensive tasks.

core/lib/zksync_core/src/commitment_generator/mod.rs
```diff
@@ -0,0 +1,264 @@
...
+            .storage_logs_dal()
+            .get_previous_storage_values(&touched_hashed_keys, l1_batch_number)
+            .await?;
+        let l1_batches_for_initial_writes = connection
+            .storage_logs_dal()
+            .get_l1_batches_and_indices_for_initial_writes(&touched_hashed_keys)
+            .await?;
+        drop(connection);
+
+        let input = if protocol_version.is_pre_boojum() {
```
slowli, line 138: Nit: IMO, this `prepare_input` method is uncomfortably large. It would be easier to read if computing `CommitmentInput` for pre-Boojum / post-Boojum batches was extracted to separate methods.

slowli (approved):
Looks really good! Optimally, there should be a couple of unit tests for the main component, but integration tests give *some* level of assurance that it works.

RomanBrodetski (dismissed):

perekopskiy:
> Optimally, there should be a couple of unit tests for the main component, but integration tests give some level of assurance that it works.

Create a task, will address soon

core/lib/zksync_core/src/commitment_generator/mod.rs
perekopskiy: I tried separating it, but to avoid unnecessary code duplication for loading data from DB, methods for pre-boojum and post-boojum input prep should take lots of args and it doesn't look nice as well, so I decided to keep it this way

StanislavBreadless (approved):

artmakh (approved):

core/lib/types/src/commitment/mod.rs
Deniallugo, line 1: Should we move it out? To some new library?

core/lib/zksync_core/src/commitment_generator/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+                Ok(bootloader_initial_content_commitment)
+            });
+
+        let events_queue_commitment = events_commitment_task
+            .await
+            .context("`events_commitment_task` failed")??;
+        let bootloader_initial_content_commitment = bootloader_memory_commitment_task
+            .await
+            .context("`bootloader_memory_commitment_task` failed")??;
+
```
Deniallugo, line 86: Does it make sense to spawn them in parallel?

core/lib/types/src/commitment/mod.rs
perekopskiy: Ideally yes, but I didn't take into in a scope of this PR

core/lib/zksync_core/src/commitment_generator/mod.rs
perekopskiy: Yes, events commitment take around 500ms on mainnet at the moment and bootloader memory around 2s, so we 500ms by running in parallel

PR #1026, opened 2024-02-07
feat(api): Start API server after first L1 batch

slowli:
## What ❔

- Delays the API server start until there's at least one L1 batch present in the node DB.
- Refactors the VM sandbox so that it's more modular / maintainable.
- Tests VM-related API server methods.

## Why ❔

Fixes API server operation when working on the DB after snapshot recovery.

## Checklist

### DIFF
--- a/core/lib/zksync_core/src/api_server/execution_sandbox/apply.rs
+++ b/core/lib/zksync_core/src/api_server/execution_sandbox/apply.rs
@@ -59,12 +59,10 @@ impl<'a> Sandbox<'a> {
         block_args: BlockArgs,
     ) -> anyhow::Result<Sandbox<'a>> {
         let resolve_started_at = Instant::now();
-        dbg!(&block_args);
         let resolved_block_info = block_args
             .resolve_block_info(&mut connection)
             .await
             .with_context(|| format!("cannot resolve block numbers for {block_args:?}"))?;
-        dbg!(&resolved_block_info);
         let resolve_time = resolve_started_at.elapsed();
         // We don't want to emit too many logs.
         if resolve_time > Duration::from_millis(10) {
@@ -99,7 +97,7 @@ impl<'a> Sandbox<'a> {
             shared_args,
             execution_args,
             &resolved_block_info,
-            dbg!(next_l2_block_info),
+            next_l2_block_info,
         );
 
         Ok(Self {
@@ -124,8 +122,6 @@ impl<'a> Sandbox<'a> {
         )
         .await
         .context("failed reading L2 block info")?;
-
-        dbg!(&current_l2_block_info);
 
         let next_l2_block_info = if is_pending_block {
             L2BlockEnv {
@@ -278,8 +274,8 @@ impl<'a> Sandbox<'a> {
 
         let storage_view = self.storage_view.to_rc_ptr();
         let vm = Box::new(VmInstance::new_with_specific_version(
-            dbg!(self.l1_batch_env),
-            dbg!(self.system_env),
+            self.l1_batch_env,
+            self.system_env,
             storage_view.clone(),
             protocol_version.into_api_vm_version(),
         ));
--- a/core/lib/zksync_core/src/api_server/execution_sandbox/apply.rs
+++ b/core/lib/zksync_core/src/api_server/execution_sandbox/apply.rs
@@ -38,8 +38,7 @@ use super::{
     BlockArgs, TxExecutionArgs, TxSharedArgs, VmPermit,
 };
 
-type SandboxStorageView<'a> = StorageView<PostgresStorage<'a>>;
-type BoxedVm<'a> = Box<VmInstance<SandboxStorageView<'a>, HistoryDisabled>>;
+type BoxedVm<'a> = Box<VmInstance<StorageView<PostgresStorage<'a>>, HistoryDisabled>>;
 
 #[derive(Debug)]
 struct Sandbox<'a> {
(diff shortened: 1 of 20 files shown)


slowli:
Sorry for combining multiple features into a single PR; it occurred naturally when covering VM-related API methods in tests and debugging related errors. I could try separating this PR into smaller ones (one per each of the 3 items mentioned in "What"?), but I'm not sure it would be worth it.

core/lib/zksync_core/src/utils/mod.rs
```diff
@@ -126,6 +126,31 @@ pub(crate) async fn projected_first_l1_batch(
     Ok(snapshot_recovery.map_or(L1BatchNumber(0), |recovery| recovery.l1_batch_number + 1))
 }
 
+/// Obtains a protocol version projected to be applied for the next miniblock. This is either the version used by the last
+/// sealed miniblock, or (if there are no miniblocks), one referenced in the snapshot recovery record.
+pub(crate) async fn pending_protocol_version(
```
Deniallugo, line 131: There are two potential problems: 
What if there are no miniblocks yet or on goerli for very ancient blocks there is no protocol version set.

slowli: - The case with no miniblocks after snapshot recovery *is* being handled, so this leaves the case without blocks and without snapshot recovery. I don't think we can handle such a case; realistically, it would be most probably caused by `pending_protocol_version()` called "too early" in the node lifecycle. @RomanBrodetski WDYT?
- We load the *latest* miniblock header, so no protocol version for some *ancient* miniblocks is not a concern. The latest miniblock on all nodes should have its protocol version set, right? Even if they don't, it's not a big deal: we could use `ProtocolVersionId::last_potentially_undefined()` as a fallback. This would make `pending_protocol_version()` completely equivalent to the current implementation.

Deniallugo: > we could use ProtocolVersionId::last_potentially_undefined() as a fallback

It w'd be great

core/lib/zksync_core/src/api_server/execution_sandbox/apply.rs
```diff
@@ -36,7 +37,250 @@ use super::{
...
+                prev_block_hash: current_l2_block_info.l2_block_hash,
+                // For simplicity we assume each miniblock create one virtual block.
+                // This may be wrong only during transition period.
+                max_virtual_blocks_to_create: 1,
+            }
+        } else if current_l2_block_info.l2_block_number == 0 {
+            // Special case:
+            // - For environments, where genesis block was created before virtual block upgrade it doesn't matter what we put here.
+            // - Otherwise, we need to put actual values here. We cannot create next L2 block with block_number=0 and `max_virtual_blocks_to_create=0`
+            //   because of SystemContext requirements. But, due to intrinsics of SystemContext, block.number still will be resolved to 0.
```
RomanBrodetski, line 138: given that virtual blocks migration is over, can this be simplified? @slowli good that you extracted this method, but the logic is still a little hard to follow

core/lib/zksync_core/src/api_server/web3/mod.rs
```diff
@@ -468,6 +459,16 @@ impl FullApiParams {
         local_addr_sender: oneshot::Sender<SocketAddr>,
         health_updater: HealthUpdater,
     ) -> anyhow::Result<()> {
+        // We don't want to start the server until at least one L1 batch is stored in Postgres.
+        let earliest_l1_batch_number =
+            wait_for_l1_batch(&self.pool, self.polling_interval, &mut stop_receiver)
+                .await
+                .context("error while waiting for L1 batch in Postgres")?;
```
RomanBrodetski, line 466: can we print a log line that we are waiting for it and when it's there? Just to make it easier to debug if we see api server not starting/not accepting requests. If we do this we can also output the block number

core/lib/zksync_core/src/utils/mod.rs
RomanBrodetski: > so this leaves the case without blocks and without snapshot recovery

So it can only happen if we run state keeper before genesis (or snapshotshot recovery) was complete. I think we need to be more explicit in expectations that component can have, but currently I think the state keeper behaviour is undefined. Now, I can imagine a problem if this method is called from e.g. snapshot recovery logic or genesis logic (as @slowli mentioned, too early in node lifecycle). This shouldn't happen, but it still concerns me that we have methods with undefined behaviour for some DB states.

I guess the easiest would be to call this pre-genesis state (point `1` [from this slack message](https://matter-labs-workspace.slack.com/archives/C05ANUCGCKV/p1707226824450369?thread_ts=1707209111.843459&cid=C05ANUCGCKV)) a special case that the components don't have to handle. This way they can always rely on the fact that there are `miniblocks`/`l1_batches` OR `snapshot_reovery` entires.

RomanBrodetski (approved):
Looks good, left some very minor comments

core/lib/zksync_core/src/utils/mod.rs
slowli: > undefined behaviour

Nitpick: this isn't UB in Rust terms, and I think returning an error with intelligible cause is the best possible course of action in such a case, which this logic does.

core/lib/zksync_core/src/api_server/execution_sandbox/apply.rs
slowli: As you've said, all work on `Sandbox` is essentially just moving code around. I'd prefer to do any semantics changes separately, so it'd be easier to review and revert if necessary.

RomanBrodetski (approved):

PR #1027, opened 2024-02-07
feat(db): Instrument DB connection lifecycle

slowli:
## What ❔

Instruments DB connection lifecycle (acquiring connection call sites, long-living connections). Traces connections for test pools and allows to configure test pool size.

## Why ❔

- Can provide more insights as to which components utilize connections suboptimally.
- Simplifies testing / DevEx in general.

## Checklist


core/lib/dal/src/connection/mod.rs
```diff
@@ -24,12 +27,20 @@ impl fmt::Debug for ConnectionPoolBuilder {
...
+            .field("acquire_timeout", &self.acquire_timeout)
             .field("statement_timeout", &self.statement_timeout)
             .finish()
     }
 }
 
 impl ConnectionPoolBuilder {
+    /// Sets the acquire timeout for a single connection attempt. There are multiple attempts (currently 3)
+    /// before `access_storage*` methods return an error.
+    pub fn set_acquire_timeout(&mut self, timeout: Duration) -> &mut Self {
```
slowli, line 39: We could theoretically allow configuring this timeout via node config. For now, it's only called by test pools.

core/lib/dal/src/connection/mod.rs
```diff
@@ -170,24 +162,68 @@ impl TestTemplate {
...
+    /// Test pools trace their active connections. If acquiring a connection fails (e.g., with a timeout),
+    /// the returned error will contain information on all active connections.
     pub async fn test_pool() -> ConnectionPool {
-        TestTemplate::empty().unwrap().create_db().await.unwrap()
+        const DEFAULT_CONNECTIONS: u32 = 50; // Expected to be enough for any unit test.
+        Self::constrained_test_pool(DEFAULT_CONNECTIONS).await
+    }
+
+    /// Same as [`Self::test_pool()`], but with a configurable number of connections. This is useful to test
+    /// behavior of components that rely on singleton / constrained pools in production.
```
slowli, line 204: It may be my problem, but I've had issues with the state keeper logic a couple of times where it passed unit tests, but hanged up on integration tests / loadtest because of a connection acquisition deadlock (the state keeper is provided with a singleton pool). Now it's possible to test this using unit tests and pinpoint where the extra connection has been created.

core/lib/dal/src/connection/processor.rs
```diff
@@ -0,0 +1,222 @@
...
+            .debug_struct("PooledStorageProcessor")
+            .field("tags", &self.tags)
+            .field("created_at", &self.created_at)
+            .finish_non_exhaustive()
+    }
+}
+
+impl Drop for PooledStorageProcessor<'_> {
+    fn drop(&mut self) {
+        const LONG_CONNECTION_THRESHOLD: Duration = Duration::from_secs(5);
```
slowli, line 109: Not sure what the reasonable value for this param should be.

core/lib/dal/src/connection/processor.rs
```diff
@@ -0,0 +1,222 @@
...
+
+use sqlx::{pool::PoolConnection, types::chrono, Connection, PgConnection, Postgres, Transaction};
+
+use crate::metrics::CONNECTION_METRICS;
+
+/// Tags that can be associated with a connection.
+#[derive(Debug, Clone, Copy)]
+pub(super) struct StorageProcessorTags {
+    pub requester: &'static str,
+    pub location: &'static Location<'static>,
```
slowli, line 20: We could add user-defined tags (e.g., as `Vec<(&'static str, &dyn fmt::Debug)>`) as well similar to instrumenting queries, but I'm not sure whether it would bring much value.

core/lib/dal/src/connection/mod.rs
```diff
@@ -218,34 +254,50 @@ impl ConnectionPool {
...
 
     /// A version of `access_storage` that would also expose the duration of the connection
-    /// acquisition tagged to the `requester` name.
+    /// acquisition tagged to the `requester` name. It also tracks the caller location for the purposes
+    /// of logging (e.g., long-living connections) and debugging (when used with a test connection pool).
     ///
     /// WARN: This method should not be used if it will result in too many time series (e.g.
     /// from witness generators or provers), otherwise Prometheus won't be able to handle it.
-    pub async fn access_storage_tagged(
+    #[track_caller] // In order to use it, we have to de-sugar `async fn`
```
slowli, line 262: Theoretically, tracking caller should only introduce some executable size overhead due to additional code locations included. This seems acceptable given that we don't have that many call locations (my IDE tells there are 96).

core/lib/dal/src/connection/processor.rs
```diff
@@ -0,0 +1,222 @@
...
+        };
+        write!(formatter, "[{timestamp} - {tags_display}]")
+    }
+}
+
+/// Traced active connections for a connection pool.
+#[derive(Default)]
+pub(super) struct TracedConnections {
+    connections: Mutex<HashMap<usize, TracedConnectionInfo>>,
+    next_id: AtomicUsize,
```
slowli, line 56: Maybe, there's a better way to track connections? Since `HashMap` entries are dropped when connections are dropped, the `HashMap` shouldn't grow unboundedly.

slowli (commented):
An example of a connection timeout error with enabled connection tracing:

```text
called `Result::unwrap()` on an `Err` value: acquire_connection_retried()

Caused by:
    Run out of retries getting a DB connection (requested by `state_keeper` at core/lib/zksync_core/src/state_keeper/io/fee_address_migration.rs:98), last error: pool timed out while waiting for an open connection
    Alive connections: {
        [2024-02-07 10:12:02.526407167 UTC - requested by `state_keeper` at core/lib/zksync_core/src/state_keeper/io/fee_address_migration.rs:74],
    }
```

core/lib/dal/src/connection/mod.rs
RomanBrodetski: I understand it's set to 30 seconds by defau;lt? I think this is a very large number. I feel like it should be something like 1 second, so def a very good idea to have this in config.
Although I'm not sure how most of the components react when they receive error when getting connection. Probably panic? But either way, I feel like we want to give up way earlier. If for API server it takes that much time to get a DB connection, we can assume that the system health is poor enough that we can just respond user with `500`. A more fundamental solution would be to introduce DB health to the circuit breaker.

core/lib/dal/src/connection/processor.rs
RomanBrodetski: I cannot wait to see this data!!!

As for the threshold, I think something between 2 and 5 is good.

RomanBrodetski (approved):
I love this, cannot wait to see the data

core/lib/dal/src/connection/mod.rs
slowli: I'll introduce the corresponding config parameter in a follow-up PR then.

PR #1029, opened 2024-02-07
fix(prover): Decouple core/ prover database management

EmilLuta:
This is PR 1 out of many PRs to come. In first step, prover database
needs to be managed by code. The changes adds a new DAL/ migration for
prover that initializes the database with all the needed tables (as
observed in production).

Additional changes are made to `zk` to not break local development. Note
that from this point forward, development will be done on 2 databases
(as opposed to 1 database model). Alongside these changes, stage env
will be moved to dual database model.

This has been long overdue (8+ months) after core and prover has been
split from infra side during an outage. This is the continuation (and
with following PRs, finalization) of said effort.

Note that whilst there's no automated proper testing of the typescript code modified, I've reran and tested working condition locally manually multiple times. I expect no breakage, but can't guarantee it.

### DIFF
--- a/infrastructure/zk/src/database.ts
+++ b/infrastructure/zk/src/database.ts
@@ -14,8 +14,8 @@ enum DalPath {
 }
 
 export interface DbOpts {
-    server: boolean,
-    prover: boolean,
+    server: boolean;
+    prover: boolean;
 }
 
 function getDals(opts: DbOpts): Map<DalPath, string> {


infrastructure/zk/src/database.ts
```diff
@@ -1,101 +1,199 @@
...
+}
+
+export async function generateMigration(opts: any, name: String) {
     console.log('Generating migration... ');
-    process.chdir('core/lib/dal');
-    await utils.exec(`cargo sqlx migrate add -r ${name}`);
+    let dals = getDals(opts);
+    for (const [dalPath, dalDb] of dals.entries()) {
+        await generateMigrationForDal(dalPath, dalDb, name);
+    }
```
Deniallugo, line 100: this is kinda unexpected, i can't imagine, why should i create migration for two dals

infrastructure/zk/src/database.ts
```diff
@@ -1,101 +1,201 @@
...
-        console.log(`Template DB URL provided. Creating a DB via dump from ${process.env.TEMPLATE_DATABASE_URL}`);
-        await utils.spawn('cargo sqlx database drop -y');
-        await utils.spawn('cargo sqlx database create');
+    await utils.spawn(`cargo sqlx database create --database-url ${dalDb}`);
+    await utils.spawn(`cargo sqlx migrate run --database-url ${dalDb}`);
+    if (dalDb.startsWith(localDbUrl)) {
         await utils.spawn(
-            `pg_dump ${process.env.TEMPLATE_DATABASE_URL} -F c | pg_restore -d ${process.env.DATABASE_URL}`
+            `cargo sqlx prepare --check --database-url ${dalDb} -- --tests || cargo sqlx prepare --database-url ${dalDb} -- --tests`
         );
```
Deniallugo, line 119: this functionality is lost. and it's important for EN

infrastructure/zk/src/database.ts
```diff
@@ -1,101 +1,201 @@
...
 
-export async function wait(tries: number = 4) {
+export async function setup(opts: any) {
+    let dals = getDals(opts);
+    for (const [dalPath, dalDb] of dals.entries()) {
+        await setupForDal(dalPath, dalDb);
+    }
+}
+
+async function waitForDal(dalDb: string, tries: number) {
```
Deniallugo, line 132: seems it's better to call it just `DB`.
There is nothing related to dal

infrastructure/zk/src/database.ts
```diff
@@ -1,101 +1,199 @@
...
-export async function generateMigration(name: String) {
+export async function migrate(opts: any) {
+    await utils.confirmAction();
+    let dals = getDals(opts);
+    for (const [dalPath, dalDb] of dals.entries()) {
+        await migrateForDal(dalPath, dalDb);
+    }
+}
+
+async function generateMigrationForDal(dalPath: string, dalDb: string, name: String) {
```
RomanBrodetski, line 90: we could have made used of typescript's strong typing here. Using enum for `dalPath`, not using `any` for opts etc. 
I'm not insisting though - if it's working like it does so we can keep as is to save time

infrastructure/zk/src/database.ts
```diff
@@ -1,101 +1,199 @@
...
+    .description('reinitialize the database for test')
+    .option('-p, --prover')
+    .option('-s, --server')
+    .action(resetTest);
+command
+    .command('check-sqlx-data')
+    .description('check sqlx-data.json is up to date')
+    .option('-p, --prover')
+    .option('-s, --server')
+    .action(checkSqlxData);
```
RomanBrodetski, line 201: I gave it a little thought and I agree with this interface - having separate, but consistent parameters to specify if you want server or prover

prover/prover_dal/migrations/20240131134938_initial-prover-migration.up.sql
```diff
@@ -0,0 +1,287 @@
...
+    protocol_version           INTEGER REFERENCES prover_fri_protocol_versions (id),
+    picked_by                  TEXT
+);
+
+CREATE INDEX IF NOT EXISTS idx_witness_inputs_fri_status_processing_attempts
+    ON witness_inputs_fri (processing_started_at, attempts)
+    WHERE (status = ANY (ARRAY['in_progress'::text, 'failed'::text]));
+CREATE INDEX IF NOT EXISTS idx_witness_inputs_fri_queued_order
+    ON witness_inputs_fri (l1_batch_number)
+    WHERE (status = 'queued'::text);
```
RomanBrodetski, line 287: very well, separte thanks for documenting this. By the way it's possible to document them in SQL (so that description is not only available on this file, but in the DB itself - Postgres clients will render this description.
Optional.

RomanBrodetski (dismissed):
Looks really good!

infrastructure/zk/src/database.ts
```diff
@@ -1,101 +1,201 @@
...
 }
 
-export async function resetTest() {
-    process.env.DATABASE_URL = process.env.TEST_DATABASE_URL;
-    await utils.confirmAction();
+function getDals(opts: any): Map<string, string> {
+    let dals = new Map<string, string>();
+    if (!opts.prover && !opts.server) {
+        dals.set('core/lib/dal', process.env.DATABASE_URL!);
+        dals.set('prover/prover_dal', process.env.DATABASE_PROVER_URL!);
```
AnastasiiaVashchuk, line 15: nit: might it be better to put `core/lib/dal` and `prover/prover_dal` into the conts?

infrastructure/zk/src/database.ts
```diff
@@ -1,101 +1,201 @@
...
     const localDbUrl = 'postgres://postgres@localhost';
-    if (process.env.DATABASE_URL!.startsWith(localDbUrl)) {
-        console.log(`Using localhost database:`);
-        console.log(`DATABASE_URL = ${process.env.DATABASE_URL}`);
+    if (dalDb.startsWith(localDbUrl)) {
+        console.log(`Using localhost database -- ${dalDb}`);
     } else {
         // Remote database, we can't show the contents.
-        console.log(`WARNING! Using prod db!`);
+        console.log(`WARNING! Using prod main db!`);
```
AnastasiiaVashchuk, line 112: why main?

infrastructure/zk/src/init.ts
```diff
@@ -67,8 +67,8 @@ export async function reinit() {
     await announced('Setting up containers', up());
     await announced('Compiling JS packages', run.yarn());
     await announced('Compile l2 contracts', compiler.compileAll());
-    await announced('Drop postgres db', db.drop());
-    await announced('Setup postgres db', db.setup());
+    await announced('Drop postgres db', db.drop({}));
+    await announced('Setup postgres db', db.setup({}));
```
popzxc, line 71: We have plenty of implicit behavior in `zk` and it has proven to cause problems. I think it's better to require everything to be passed explicitly, even at the cost of being more verbose.

infrastructure/zk/src/hyperchain_wizard.ts
```diff
@@ -268,7 +268,7 @@ async function setHyperchainMetadata() {
         feeReceiverAddress = richWallets[3].address;
 
         await up('docker-compose-zkstack-common.yml');
-        await announced('Ensuring databases are up', db.wait());
+        await announced('Ensuring databases are up', db.wait({ server: true, prover: false }));
```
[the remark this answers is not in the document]
githubdoramon: 100% it doesn't, especially on this initialization :)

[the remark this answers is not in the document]
Deniallugo: Hyperchain has prover, so it's important

AnastasiiaVashchuk (approved):

infrastructure/zk/src/hyperchain_wizard.ts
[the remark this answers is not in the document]
githubdoramon: @Deniallugo but this part of the script is literally just initializing the db to do the initial setup (deploying L1 contracts, setting genesis root hash, etc). Provers come way after in the process. Why would it make sense to wait for the DB at this point (or does the prover DB also gets some init values from the chain)?

infrastructure/zk/src/database.ts
```diff
@@ -1,101 +1,233 @@
...
+            '[aborted] please specify a single database to generate migration for (i.e. to generate a migration for server `zk db new-migration --server name_of_migration`'
+        );
+    }
+    let name = '';
+    if (opts.prover) {
+        name = opts.prover;
+    }
+    if (opts.server) {
+        name = opts.server;
+    }
```
popzxc, line 112: This is a bit clunky. Why not just introduce a public enum `export enum DBType { Server, Prover }` and accept it as an argument?

infrastructure/zk/src/database.ts
```diff
@@ -1,101 +1,233 @@
...
+        dals.set(DalPath.ProverDal, process.env.TEST_DATABASE_PROVER_URL!);
+    }
+    if (opts.prover) {
+        dals.set(DalPath.ProverDal, process.env.TEST_DATABASE_PROVER_URL!);
+    }
+    if (opts.server) {
+        dals.set(DalPath.CoreDal, process.env.TEST_DATABASE_URL!);
+    }
+    return dals;
+}
```
popzxc, line 44: Generally speaking, I feel like this whole file became a bit boilerplate-ish (if that's even a word).
We deal with a set of `{server, prover} x {test, non_test}` here for which we maintain some currently implicit state (exposed by these methods and by `xxx` and `xxxTest` function variants) and I feel like we can express it better by having something like (pseudocode):

```typescript
export enum DBType { Server, Prover };

export class DBManager {
    public constructor(type: DBType, test: boolean) {  /* Load required envs, setup paths, etc */ }

    public async drop() { .. }
    public async migrate() { .. }
    // .. whatever
}
```

infrastructure/zk/src/database.ts
```diff
@@ -1,101 +1,233 @@
...
+    await setupForDal(dalPath, dbUrl);
     console.log('disallowing connections to the template');
     await utils.spawn(
-        `psql "${process.env.DATABASE_URL}" -c "update pg_database set datallowconn = false where datname = current_database()"`
+        `psql "${dbUrl}" -c "update pg_database set datallowconn = false where datname = current_database()"`
     );
 }
 
-export async function drop() {
+export async function resetTest(opts: any) {
```
popzxc, line 59: Btw, why `any`? We can declare an interface for that to be typesafe.

Though if we will consider an approach with a public enum for the dal type, I think we don't need it at all. We can either take an array of `DBType` or even let the caller deal with it (if they have to, they can just call the method twice, once for `Server` and once for `Prover`.

infrastructure/zk/src/database.ts
popzxc: I think TypeScript should be able to coerce interfaces, e.g. if you declare your own interface that matches the interface of what Commander would give you, it should work.

I've checked out your branch, and the following seems to compile just fine:

```
function getDals(opts: { server: boolean, prover: boolean }): Map<DalPath, string>
```

as well as 

```
export interface Opts {
    server: boolean,
    prover: boolean,
}

function getDals(opts: Opts): Map<DalPath, string> {
```

infrastructure/zk/src/database.ts
popzxc: You can always wrap stuff into a closure for commander's sake, no?
E.g. `.action((opts) => (new DBManager(opts, false)).drop())` or something like that.

infrastructure/zk/src/hyperchain_wizard.ts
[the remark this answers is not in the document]
Deniallugo: The question is when we are supposed to initialize it? why not during init?

RomanBrodetski (approved):

popzxc (approved):

infrastructure/zk/src/hyperchain_wizard.ts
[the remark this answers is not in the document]
Deniallugo: I'd say that it's a negligible amount of time to migrate prover db and it's better to avoid possible mistakes

PR #1031, opened 2024-02-07
feat: Add l1 token address endpoint

jrchatruc:
## What ❔

This PR adds an RPC method that answers the chain's base token, or `0x0..01` if the base token is `eth`.
## Why ❔
To make clients like block explorers or cli tools able to know which token the chain they point to is operating on, now that it could be one different from `eth`.

## Checklist


core/lib/zksync_core/src/api_server/web3/namespaces/zks.rs
```diff
@@ -646,4 +646,14 @@ impl ZksNamespace {
...
     }
+
+    #[tracing::instrument(skip_all)]
+    pub async fn get_base_token_l1_address_impl(&self) -> Result<Address, Web3Error> {
+        const METHOD_NAME: &str = "get_base_token_l1_address_impl";
+        let l1_token_env_var = env::var("CONTRACTS_BASE_TOKEN_ADDR")
+            .map_err(|_err| internal_error(METHOD_NAME, "Missing env variable for base token"))?;
+        Ok(H160::from_str(&l1_token_env_var).map_err(|_err| {
+            internal_error(METHOD_NAME, "Set value for base token is not an address")
+        })?)
```
popzxc, line 657: Please do not access environment variables directly from within any module.
Environment variables are not the only way to define the configuration for the application, use a dedicated configuration approach for that.
You may have a look at how other methods that return contract addresses are implemented in this file.

fkrause98: Done! Please let me know if there's something else to improve.

core/bin/external_node/src/config/mod.rs
```diff
@@ -66,7 +67,10 @@ impl RemoteENConfig {
                 .context("Failed to fetch L1 chain ID")?
                 .as_u64(),
         );
-
+        let base_token_addr = client
+            .get_base_token_l1_address()
+            .await
+            .context("Failed to fetch base token address")?;
```
popzxc, line 73: This probably wouldn't work if the new EN is released before the server version is updated (e.g. EN will attempt to call this method, but it doesn't exist yet on the older version of server) -- and this is likely to happen, because EN is almost always released first, since we have to test the new version on testnet for some time.
Probably what we want to do is to handle the error: if the request fails with "method does not exist" error, we fallback to the existing behavior (base token is ETH). We still can fail on other errors.

popzxc (commented):
It does look better now!

core/bin/external_node/src/config/mod.rs
fkrause98: Hi @popzxc!.

I've handled the case for the missing rpc method [here](https://github.com/matter-labs/zksync-era/pull/1031/files#diff-e52f3f81b2bcbc96468f7c93c1731a3a0c7aaab6a93c95b5e5e42173b1467740R74).
I'm not quite sure if this is the proper way to match with jsonrpsee's Error type, let me know if there's another and proper way to do it.

Also, one thing that to note is that I've hardcoded ETH's base token value [here](https://github.com/matter-labs/zksync-era/pull/1031/files#diff-e52f3f81b2bcbc96468f7c93c1731a3a0c7aaab6a93c95b5e5e42173b1467740R76), I'm guessing this should be properly set as a `const` value somewhere, but I haven't found where this should be done.

popzxc: Is `0x0000000000000000000000000000000000000001` an expected address?
We have L1 ETH address defined [here](https://github.com/matter-labs/zksync-era/blob/main/core/lib/constants/src/ethereum.rs#L6) and L2 ETH [here](https://github.com/matter-labs/zksync-era/blob/main/core/lib/constants/src/contracts.rs#L56) as constants, you should be able to reuse them.

core/lib/env_config/src/contracts.rs
```diff
@@ -70,6 +73,8 @@ mod tests {
             snark_wrapper_vk_hash: hash(
                 "0x4be443afd605a782b6e56d199df2460a025c81b3dea144e135bece83612563f2",
             ),
+            base_token_addr: Address::from_str("0x0000000000000000000000000000000000000001")
+                .unwrap(),
```
popzxc, line 77: You can use `addr` convenience function here similarly to above, this way you won't need to import `Address` or `FromStr`.

core/bin/external_node/src/config/mod.rs
```diff
@@ -66,7 +71,12 @@ impl RemoteENConfig {
                 .context("Failed to fetch L1 chain ID")?
                 .as_u64(),
         );
-
+        let base_token_addr = match client.get_base_token_l1_address().await {
+            Err(ClientError::Call(err)) if ErrorCode::MethodNotFound.code() == err.code() => {
+                Address::from_str("0x0000000000000000000000000000000000000001")?
```
popzxc, line 76: You can add a comment that it's a fallback for the case when EN version tries to interact with an older main node that doesn't have this method yet.

core/lib/config/src/configs/contracts.rs
```diff
@@ -49,6 +51,7 @@ pub struct ContractsConfig {
     pub fri_recursion_leaf_level_vk_hash: H256,
     pub prover_at_genesis: ProverAtGenesis,
     pub snark_wrapper_vk_hash: H256,
+    pub base_token_addr: Address,
```
popzxc, line 54: This is a mandatory config parameter, which means that it _has_ to be set before the server launches.
If we'll roll this change out without updating configs beforehand, the servers will enter the crashloop.

Would it make sense to use `#[serde(default = "...")]` here to implement a graceful fallback to ETH token if the variable is not set?

popzxc (commented):
Mostly LGTM. I guess to get CI passing, you need to merge recent main.

core/bin/external_node/src/config/mod.rs
slowli, line 76: Nit: You can use `"...".parse()` instead of importing `FromStr`; ditto in other places. Also, I'd suggest to create a constant for the default (fallback?) base token L1 address (probably somewhere in the `zksync_types` crate) so that there's no risk that it's mismatched in various places it's used.

core/lib/zksync_core/src/api_server/web3/backend_jsonrpsee/namespaces/zks.rs
```diff
@@ -20,6 +20,7 @@ use zksync_web3_decl::{
 use crate::api_server::web3::{backend_jsonrpsee::into_jsrpc_error, ZksNamespace};
 
 #[async_trait]
+
```
slowli, line 23: Looks like a bogus change to me; why is it necessary?

core/lib/zksync_core/src/api_server/web3/namespaces/zks.rs
```diff
@@ -646,4 +646,9 @@ impl ZksNamespace {
             storage_proof,
         })
     }
+
+    #[tracing::instrument(skip_all)]
+    pub async fn get_base_token_l1_address_impl(&self) -> Result<Address, Web3Error> {
```
slowli, line 651: Nit: It doesn't make sense to make this function `async`; AFAIU, it's not like it's expected to become async in the future.

core/bin/external_node/src/config/mod.rs
fkrause98: This is now set from the env, but the previous hardcoded value of`0x0000000000000000000000000000000000000001` was because the zk init set the env var CONTRACTS_BASE_TOKEN_ADDR=0x0000000000000000000000000000000000000001
when not specifiying a non-eth base token

popzxc: @fkrause98 Not sure if I understand. We want to fallback to ETH if there is no dedicated config set for that (this may happen when we'll be rolling out this version of server on existing zkSync Era deployments). Why would we use `0x0000000000000000000000000000000000000001 ` in that case? We want to use ETH address, no?

fkrause98: The new changes for the shared bridge determine an eth-based chain based on the base token's address value.
For example, the sdk now has the following code:
- [isEthBasedChain function](https://github.com/zksync-sdk/zksync-ethers/blob/ethers-v5-feat/bridgehub/src/adapters.ts#L97).
- [ETH_ADDRESS_IN_CONTRACTS constant](https://github.com/zksync-sdk/zksync-ethers/blob/ethers-v5-feat/bridgehub/src/utils.ts#L34).

So, when a chain is eth-based, the base token address is 0x..01, else is a proper address to an erc20 contract on L1.

popzxc: But we're discussing the fallback for the case when shared bridge is not yet deployed, aren't we?
E.g. imagine that this feature is merged. This will happen, most likely, before we actually deploy a shared bridge and enable it for existing zkSync Era. The method will be there, but `0x00..01` does not represent an actual ETH token available back then.

I may be missing context, but right now it looks like we have 3 different cases:
1. Shared bridge is added, base token is ETH.
2. Shared bridge is added, base token is NOT ETH.
3. No shared bridge added (e.g. the development is still ongoing).

What I'm trying to explain is that we need to have sensible fallbacks for (3) if we want to merge it to main anytime soon.

jrchatruc: I think it makes sense to just make it return the zero address, it makes all 3 cases work and nothing really breaks from it so I would go that route. It is important to make sure users are aware of it though, as it means some APIs will return the zero address when the base token is `eth` and others (like the `baseToken` function in the new bridgehub) will return `0x00..01` instead.

Tagging @kelemeno for visibility as this pertains to contract/SDK changes.

core/lib/env_config/src/contracts.rs
fkrause98: Done

core/lib/zksync_core/src/api_server/web3/namespaces/zks.rs
fkrause98: Fixed!

core/lib/zksync_core/src/api_server/web3/backend_jsonrpsee/namespaces/zks.rs
fkrause98: Removed.

core/lib/config/src/configs/contracts.rs
```diff
@@ -83,6 +85,8 @@ impl ContractsConfig {
             governance_addr: Address::repeat_byte(0x13),
             prover_at_genesis: ProverAtGenesis::Fri,
             snark_wrapper_vk_hash: H256::repeat_byte(0x09),
+            base_token_addr: Address::from_str("0x0000000000000000000000000000000000000001")
+                .unwrap(),
```
popzxc, line 89: Nit: Given that we use fake addresses here anyways, I think it'd be fine to use `Address::repeat_byte(0x14)`, this way you'll avoid having to import `FromStr` and doing an `unwrap`.

core/lib/constants/src/contracts.rs
```diff
@@ -120,3 +120,8 @@ pub const MINT_AND_BURN_ADDRESS: H160 = H160::zero();
 
 // The `storage_log.value` database value for a contract that was deployed in a failed transaction.
 pub const FAILED_CONTRACT_DEPLOYMENT_BYTECODE_HASH: H256 = H256::zero();
+
+pub const ETHEREUM_SHARED_BRIDGE_ADDRESS: Address = H160([
```
popzxc, line 124: A doc-comment here would be appreciated.

core/lib/contracts/src/lib.rs
```diff
@@ -39,10 +39,10 @@ const MULTICALL3_CONTRACT_FILE: &str =
...
 const VERIFIER_CONTRACT_FILE: &str =
     "contracts/l1-contracts/artifacts/cache/solpp-generated-contracts/state-transition/Verifier.sol/Verifier.json";
-const IERC20_CONTRACT_FILE: &str =
-    "contracts/l1-contracts/artifacts/cache/solpp-generated-contracts/common/interfaces/IERC20.sol/IERC20.json";
-const FAIL_ON_RECEIVE_CONTRACT_FILE: &str =
-    "contracts/l1-contracts/artifacts/cache/solpp-generated-contracts/zksync/dev-contracts/FailOnReceive.sol/FailOnReceive.json";
+// const IERC20_CONTRACT_FILE: &str =
+//     "contracts/l1-contracts/artifacts/cache/solpp-generated-contracts/common/interfaces/IERC20.sol/IERC20.json";
+// const FAIL_ON_RECEIVE_CONTRACT_FILE: &str =
+//     "contracts/l1-contracts/artifacts/cache/solpp-generated-contracts/zksync/dev-contracts/FailOnReceive.sol/FailOnReceive.json";
```
popzxc, line 45: Commented code.

core/lib/env_config/src/contracts.rs
```diff
@@ -103,6 +104,7 @@ CONTRACTS_FRI_RECURSION_NODE_LEVEL_VK_HASH="0x5a3ef282b21e12fe1f4438e5bb158fc506
 CONTRACTS_FRI_RECURSION_LEAF_LEVEL_VK_HASH="0x72167c43a46cf38875b267d67716edc4563861364a3c03ab7aee73498421e828"
 CONTRACTS_PROVER_AT_GENESIS="fri"
 CONTRACTS_SNARK_WRAPPER_VK_HASH="0x4be443afd605a782b6e56d199df2460a025c81b3dea144e135bece83612563f2"
+CONTRACTS_BASE_TOKEN_ADDR=0x0000000000000000000000000000000000000001
```
popzxc, line 107: You set `base_token_addr: addr("0x0000000000000000000000000000000000000000"),` above, so probably this test would fail.

core/bin/external_node/src/config/mod.rs
popzxc: BTW, I've consulted @vladbochok on this topic and it appears that this is still WIP, but we indeed leaning towards having it as `0x00..01` everywhere, including public interfaces.

This unfortunately renders my proposal incorrect -- sorry for that.
Now I believe that the right thing would be to make the base token optional, e.g. the corresponding config variable would have an `Option<Address>` instead of `Address` and if it's `None` then in the API we'll return `Err(Web3Error::NotImplemented)`, otherwise we return whatever is stored in the config.

On EN we'll have two fallbacks then: first on `MethodNotFound` and second on `NotImplemented`, both resulting in setting base token to `None`. Otherwise it'll be `Some` whatever returned by API.

This way we'll have no additional hackery to support base token until it's explicitly enabled in existing envs.

Does it make sense?

fkrause98: Yup, it does make sense, I'll make those changes and let you know.

core/lib/contracts/src/lib.rs
fkrause98: Uncommented.

core/lib/env_config/src/contracts.rs
fkrause98: Fixed

core/bin/external_node/src/config/mod.rs
fkrause98: @popzxc I've implemented and tested the 3 cases you stated above and works as expected.

core/lib/config/src/configs/contracts.rs
fkrause98: This is now an Option<Address> so this is not needed.

core/bin/external_node/src/config/mod.rs
fkrause98: Done, this is now a constant over on  lib/constants/src/contracts.rs file

core/lib/config/src/configs/contracts.rs
fkrause98: Done

core/lib/constants/src/contracts.rs
fkrause98: Done, wdy?

core/bin/external_node/src/config/mod.rs
```diff
@@ -55,6 +60,14 @@ impl RemoteENConfig {
             .get_main_contract()
             .rpc_context("get_main_contract")
             .await?;
+        let base_token_addr = match client.get_base_token_l1_address().await {
+            Err(ClientError::Call(err)) if ErrorCode::MethodNotFound.code() == err.code() => {
+                // This is the fallback case for when the EN tries to interact
+                // with a node that does not implement the zks_baseTokenL1Address endpoint.
+                ETHEREUM_ADDRESS
+            }
```
popzxc, line 68: We need to have this fallback on `Web3Error::NotImplemented` as well.

popzxc (commented):
Server-side implementation LGTM, one small fix for EN and that should be it.

core/bin/external_node/src/config/mod.rs
fkrause98: I don't think I quite follow you, you mean this match should also handle the `Web3Error::NotImplemented` error, right? If so, I've done it [here](https://github.com/matter-labs/zksync-era/pull/1031/commits/f544652953edfd66390240d5b294a9a4ba47b106).

fkrause98: @popzxc 
Tagging for visibility.

mationorato: @popzxc could you see this?

core/bin/external_node/src/config/mod.rs
mationorato: @popzxc can we resolve this?

core/bin/external_node/src/config/mod.rs
popzxc: Sorry for the delay. Yeah, that will do. There is a bug in the API server implementation right now (we use an incorrect code for the `NotImplemented` error), but we'll fix that separately, and until then the approach you used should work. Thanks!

popzxc (approved):
PR LGTM, but the CI seems to be failing.

core/bin/external_node/src/config/mod.rs
mationorato: Nice! thanks!

lferrigno:
> PR LGTM, but the CI seems to be failing.

CI is failing in the destination branch. @kelemeno is working on that separately

PR #1032, opened 2024-02-08
feat(en): Integrate snapshots recovery into EN

tomg10:
## What ❔

- Integrates snapshot recovery logic into EN.
- Tests recovering an EN from a snapshot in integration tests.

## Why ❔

This is the final step to implement snapshot recovery (still needs to be thoroughly tested, of course).

## Checklist


core/lib/config/src/configs/object_store.rs
```diff
@@ -1,19 +1,34 @@
...
+
+impl ObjectStoreConfig {
+    const fn default_max_retries() -> u16 {
+        5
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Deserialize)]
+#[serde(tag = "mode")]
+pub enum ObjectStoreMode {
```
slowli, line 20: Sort of unrelated change, but I've found `ObjectStoreConfig` as-is overly convoluted and not maintainable. The new format makes invalid states non-representable. I *think* the change is backward-compatible (other than the associated Protobuf declarations, but AFAIU, they aren't used yet).

core/tests/ts-integration/package.json
```diff
@@ -10,16 +10,15 @@
...
+        "build-yul": "hardhat run scripts/compile-yul.ts"
     },
     "devDependencies": {
         "@matterlabs/hardhat-zksync-deploy": "^0.6.1",
         "@matterlabs/hardhat-zksync-solc": "0.4.2",
         "@matterlabs/hardhat-zksync-vyper": "^1.0.0",
         "@nomiclabs/hardhat-vyper": "^3.0.5",
         "@types/jest": "^29.0.3",
-        "@types/node": "^14.14.5",
+        "@types/node": "^18.19.15",
```
slowli, line 21: AFAIU, we use Node 18 in `zk-environment`, so this update should be safe.

infrastructure/zk/src/config.ts
```diff
@@ -75,24 +49,47 @@ export function collectVariables(config: any, prefix: string = ''): Map<string,
...
+        loadConfigRecursive(config, configPath, []);
     }
+    return config;
+}
 
-    const overridesPath = `${process.env.ZKSYNC_HOME}/etc/env/${env}.toml`;
-    if (fs.existsSync(overridesPath)) {
-        const overrides = loadConfigFile(overridesPath);
-        deepExtend(config, overrides);
+function loadConfigRecursive(config: object, configPath: string, calledFrom: string[]) {
```
slowli, line 66: Another sort of unrelated change: I've found the current config compilation logic to be overly brittle, so I've devised something similar to Spring 🙃 I think it should be easy to decipher from the changed config TOMLs, but essentially `_metadata.base` is an array of paths to config files (relative to `etc/env` dir) that this config inherits. In this way, copy-paste between Docker and "ordinary" variants of configs is eliminated, and the `dev` config doesn't require special handling.

popzxc: I'd say -0 on this change. It wasn't pretty before, it is more complex now. `zk` seems to be a hot topic and point of confusion lately, so probably I'd expect a bit more discussion here.

However, given that 
a) it's already done, and
b) `zk` would likely be replaced with something else mid-term.

Still, probably it makes sense to contact @ly0va to make sure that it doesn't conflict with the work he and Kalmon are doing for hyperchains.

core/bin/snapshots_creator/src/creator.rs
```diff
@@ -266,6 +266,10 @@ impl SnapshotCreator {
         config: SnapshotsCreatorConfig,
         min_chunk_count: u64,
     ) -> anyhow::Result<()> {
+        tracing::info!(
+            "Starting snapshot creator with object store {:?} and config {config:?}",
+            self.blob_store
+        );
```
RomanBrodetski, line 272: unrelated question: what happens if we run creator on a pruned node?

core/bin/external_node/src/init.rs
```diff
@@ -0,0 +1,103 @@
...
+                 `--enable-snapshots-recovery` command-line arg to the node binary, or reset the node storage \
+                 to sync from genesis"
+            );
+
+            tracing::warn!("Proceeding with snapshot recovery. This is an experimental feature; use at your own risk");
+            let recovery_config = read_snapshots_recovery_config()?;
+            let blob_store = ObjectStoreFactory::new(recovery_config.snapshots_object_store)
+                .create_store()
+                .await;
+            // FIXME: change error handling once #1036 is merged; this is not always correct.
```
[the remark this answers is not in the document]
slowli: Yeah, it's sort of a preliminary review at this point, although changes related to #1036 shouldn't be that large.

core/bin/external_node/src/init.rs
```diff
@@ -0,0 +1,103 @@
...
+
+    let decision = match (genesis_l1_batch, snapshot_recovery) {
+        (Some(batch), Some(snapshot_recovery)) => {
+            anyhow::bail!(
+                "Node has both genesis L1 batch: {batch:?} and snapshot recovery information: {snapshot_recovery:?}. \
+                 This is not supported and can be caused by broken snapshot recovery."
+            );
+        }
+        (Some(batch), None) => {
+            tracing::info!("Node has a genesis L1 batch: {batch:?} and no snapshot recovery info");
```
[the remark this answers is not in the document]
slowli: I don't think so. IMO, it'd be more intuitive if `--enable-snapshot-recovery` enables, but doesn't force snapshot recovery. Roughly speaking, it should be possible to use `--enable-snapshot-recovery` on an existing EN (it wouldn't have any effect there). This is completely subjective though, so I'm wondering what other people think about it. cc @RomanBrodetski

core/tests/snapshot-recovery-test/tests/snapshot-recovery.test.ts
```diff
@@ -0,0 +1,255 @@
...
+        await promisify(exec)('killall -q -KILL zksync_external_node');
+    } catch (err) {
+        const typedErr = err as ChildProcessError;
+        if (typedErr.code === 1) {
+            // No matching processes were found; this is fine.
+        } else {
+            throw err;
+        }
+    }
+}
```
RomanBrodetski, line 255: honestly, I thought it will be even more bothersome to do this test, so, well done!

RomanBrodetski (approved):

core/tests/snapshot-recovery-test/tests/snapshot-recovery.test.ts
```diff
@@ -0,0 +1,255 @@
...
+    readonly l1BatchNumberOfInitialWrite: number;
+    readonly enumerationIndex: number;
+}
+
+/**
+ * Assumptions:
+ *
+ * - Main node is run for the duration of the test.
+ * - `ZKSYNC_ENV` variable is not set (checked at the start of the test). For this reason,
+ *   the test doesn't have a `zk` wrapper; it should be launched using `yarn`.
```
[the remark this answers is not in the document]
slowli: If you mean gossip-based syncing, it doesn't support snapshot recovery yet; consensus team are aware of this. I haven't addressed this yet; I'll add an early bail-out if both `--enable-consensus` and `--enable-snapshot-recovery` options are specified, and run the snapshot recovery integration test only in the CI job with `consensus=false`. Right now, although the test runs with `consensus=true`, it actually doesn't enable consensus on the EN, which is quite misleading.

core/bin/external_node/src/init.rs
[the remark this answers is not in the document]
RomanBrodetski: I don't have an opinon on this, feel free to leave as is

core/tests/snapshot-recovery-test/tests/snapshot-recovery.test.ts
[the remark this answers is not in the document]
RomanBrodetski: oh ok. Yes, misleading, thanks for the explanation

core/bin/snapshots_creator/src/creator.rs
slowli: I haven't tested it, but theoretically it should work fine. There's nothing in the snapshot creator logic that requires complete node history.

RomanBrodetski (approved):

popzxc (approved):

PR #1033, opened 2024-02-08
chore(db): Instrument db connection lifecycle (follow-ups)

slowli:
## What ❔

- Makes instrumentation-related DB timings configurable.
- Outputs DB connection tags for slow / errored queries.

## Why ❔

Improves DevEx.

## Checklist

### DIFF
--- a/core/lib/dal/src/connection/mod.rs
+++ b/core/lib/dal/src/connection/mod.rs
@@ -1,4 +1,13 @@
-use std::{env, fmt, future::Future, panic::Location, sync::Arc, time::Duration};
+use std::{
+    env, fmt,
+    future::Future,
+    panic::Location,
+    sync::{
+        atomic::{AtomicU64, Ordering},
+        Arc,
+    },
+    time::Duration,
+};
 
 use anyhow::Context as _;
 use rand::Rng;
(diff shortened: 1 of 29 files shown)


popzxc (approved):

slowli:
Would it make sense to allow configuring the new DB params on the external node? E.g., if it works with a slower Postgres, the operator might want to increase the slow query / long-living connection thresholds, so that the node doesn't spam useless logs.

popzxc (approved):

core/lib/dal/src/connection/mod.rs
```diff
@@ -298,11 +362,10 @@ impl ConnectionPool {
         &self,
         tags: Option<&StorageProcessorTags>,
     ) -> anyhow::Result<PoolConnection<Postgres>> {
-        const DB_CONNECTION_RETRIES: u32 = 3;
-        const BACKOFF_INTERVAL: Duration = Duration::from_secs(1);
+        const DB_CONNECTION_RETRIES: usize = 3;
```
RomanBrodetski, line 365: no need to change anything in this PR, but this got me thinking. I think we need different approach when API and core components fail to connect to the DB. 

On API side, we might want to fail early and respond with 5xx on the first time when we fail to get the connection. Of course we should monitor how often and under what circumstances it happens, but my intuition it should be a very rare occurance.

On backend side, e.g. in state keeper, we obviously want to try until we get the connection - otherwise state keeper cannot continue. But thinking about it further, depending on how often this happens, we can discuss letting state keeper panic, fail and restart. This way we can ditch the connection retry concept altogether.

Not so sure it's a good idea but wanted to bring up

PR #1034, opened 2024-02-08
perf: upgrade harness version to improve witness generation memory spike

joonazan:
This should further lower witness generation memory use, especially during the brief spike when computing RAM circuits.

Also upgrades boojum and shivini because a commit upgrading harness to newer boojum is under my commit in harness.

### DIFF
--- a/prover/Cargo.lock
+++ b/prover/Cargo.lock
@@ -805,7 +805,7 @@ dependencies = [
 [[package]]
 name = "circuit_definitions"
 version = "0.1.0"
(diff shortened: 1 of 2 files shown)


perekopskiy (dismissed):

joonazan:
Just below my commit there is a commit that upgrades boojum. But harness version 1.4.0 and 1.3.3 haven't been upgraded, so we need to get rid of those dependencies before continuing this.

perekopskiy (approved):

PR #1042, opened 2024-02-11
test(node): add basic unit tests for LazyResource

AnastasiiaVashchuk:
## Checklist

### DIFF
--- a/core/node/node_framework/src/resource/lazy_resource.rs
+++ b/core/node/node_framework/src/resource/lazy_resource.rs
@@ -83,10 +83,104 @@ impl<T: Resource + Clone> LazyResource<T> {
     }
 }
 
-#[derive(Debug, Error)]
+#[derive(Debug, Error, PartialEq)]
 pub enum LazyResourceError {
     #[error("Node is shutting down")]
     NodeShutdown,
     #[error("Resource is already provided")]
     ResourceAlreadyProvided,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[derive(Debug, Clone, PartialEq)]
+    struct TestResource(Arc<u8>);
+
+    impl Resource for TestResource {
+        fn resource_id() -> ResourceId {
+            ResourceId::new("test_resource")
+        }
+    }
+
+    fn init_lazy_resource() -> LazyResource<TestResource> {
+        let (stop_sender, _stop_receiver) = watch::channel(false);
+        LazyResource::<TestResource>::new(StopReceiver(stop_sender.subscribe()))
+    }
+
+    fn init_test_resource() -> TestResource {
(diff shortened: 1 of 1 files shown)


AnastasiiaVashchuk:
@popzxc  I was hesitating about adding comments for the tests I wrote. Please review and let me know whether it is worth adding!

core/node/node_framework/src/resource/lazy_resource.rs
```diff
@@ -83,10 +83,104 @@ impl<T: Resource + Clone> LazyResource<T> {
     }
 }
 
-#[derive(Debug, Error)]
+#[derive(Debug, Error, PartialEq)]
```
AnastasiiaVashchuk, line 86: will not be superfluous imo.

core/node/node_framework/src/resource/lazy_resource.rs
```diff
@@ -83,10 +83,104 @@ impl<T: Resource + Clone> LazyResource<T> {
...
+        let (lr_for_task2, _) = (lazy_resource.clone(), test_resource.clone());
+
+        let task2 = tokio::spawn(async move {
+            assert_eq!(
+                lr_for_task2.resolve().await,
+                Err(LazyResourceError::NodeShutdown)
+            );
+        });
+
+        futures::future::join_all([task1, task2]).await;
```
popzxc, line 184: I think here we can use just one tokio task:
1. You spawn a task where we call `resolve`, and then
2. You send the `true` through the `stop_sender`, and then
3. We `.await` the join handle where we called `resolve()`.

Additionally, we may change the semantics a bit: instead of asserting that we've returned an error _inside_ of the tokio task, we just return what we got there, and in the main test body we make sure that return value is `Ok(Err(LazyResourceError::NodeShutdown))` (`Ok` means that the task was completed successfully and didn't panic; and `Err(..)` is what we return from the task.

popzxc (commented):
A good starting point!

core/node/node_framework/src/resource/lazy_resource.rs
```diff
@@ -83,10 +83,104 @@ impl<T: Resource + Clone> LazyResource<T> {
...
+
+    #[tokio::test]
+    async fn test_already_provided_resource_case() {
+        let (lazy_resource, test_resource) = init_resources();
+
+        let (mut lr_for_task1, tr_for_task1) = (lazy_resource.clone(), test_resource.clone());
+
+        let task1 = tokio::spawn(async move {
+            lr_for_task1.provide(tr_for_task1).await.unwrap();
+        });
```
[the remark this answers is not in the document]
AnastasiiaVashchuk: make sense!

popzxc (approved):

PR #1043, opened 2024-02-12
feat(node_framework): Support StateKeeper in the framework

popzxc:
## What ❔

- ⚠️ Docs are missing or outdated in many places -- I found myself spending too much time keeping them up to date while the framework in active development, so I gave up. Will update all the docs at once when the framework is more or less finalized.
- ❗ Removes the `ResourceProvider` abstraction. State keeper task finally showed that it brings more problems than solutions. Main implication: now ordering of layers matters ("base" layers should go first), but hopefully it shouldn't cause too much problems.
- Makes it possible to add resources to the context without 
- Creates `PoolsLayer` that was previously handled by `ResourceProvider`.
- Creates `SequencerFeeInputLayer` and `ObjectStoreLayer`.
- Adds notion of unique resource (to have `Box<smth>` as a non-shareable resource, like IO).
- Adds support for state keeper (in form of `StateKeeperLayer`, `MempoolIOLayer`, and `MainNodeBatchExecutorBuilderLayer`).
- Rewamps the main node example to look better.

Current "external" point of view for the framework:

```rust
        .with_log_format(log_format)
        .build();

    MainNodeBuilder::new()
        .add_pools_layer()?
        .add_fee_input_layer()?
        .add_object_store_layer()?
        .add_state_keeper_layer()?
        .add_healthcheck_layer()?
        .build()
        .run()?;

    Ok(())
}
```


core/node/node_framework/src/service/context.rs
```diff
@@ -98,4 +90,15 @@ impl<'a> ServiceContext<'a> {
     pub async fn get_resource_or_default<T: Resource + Clone + Default>(&mut self) -> T {
         self.get_resource_or_insert_with(T::default).await
     }
+
+    /// Adds a resource to the service.
+    /// If the resource with the same name is already provided, the method will return an error.
+    pub fn add_resource<T: Resource>(&mut self, resource: T) -> Result<(), WiringError> {
```
slowli, line 96: Ah, inserting resources makes the `ServiceContext` container look more like a traditional DI container (IMO, that's a good thing). 

- Bikeshedding: I feel that per Rust naming conventions, `insert_resource` would be more idiomatic (cf. map / set methods in `std`).
- Here and in other `ServiceContext` methods, it would make sense to log operations to make potential DI problems easier to investigate. Could be done later, of course.

core/lib/zksync_core/src/state_keeper/keeper.rs
```diff
@@ -60,15 +61,15 @@ pub struct ZkSyncStateKeeper {
     stop_receiver: watch::Receiver<bool>,
     io: Box<dyn StateKeeperIO>,
     batch_executor_base: Box<dyn L1BatchExecutorBuilder>,
```
slowli, line 63: Bikeshedding / off-topic (?): I feel that it makes sense to shorten this name to `L1BatchExecutor` (and, naturally, shorten names for the related types introduced in this PR correspondingly, so that we don't need to deal with enterprise™ names like `MainNodeBatchExecutorBuilderLayer`).

core/lib/zksync_core/src/lib.rs
```diff
@@ -758,15 +758,15 @@ async fn add_state_keeper_to_task_futures(
...
-    let next_priority_id = state_keeper_pool
-        .access_storage()
-        .await
-        .unwrap()
-        .transactions_dal()
-        .next_priority_id()
-        .await;
-    let mempool = MempoolGuard::new(next_priority_id, mempool_config.capacity);
-    mempool.register_metrics();
+    let mempool = {
```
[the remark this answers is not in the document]
popzxc: I'd say that it's the fairly idiomatic thing in Rust. Alternatively, we can call `drop(connection)`, but personally I like scoped blocks more (since the "lifetime" of the connection is indented).

core/node/node_framework/examples/main_node.rs
```diff
@@ -2,52 +2,109 @@
...
-            _ => None,
-        }
+    fn add_object_store_layer(mut self) -> anyhow::Result<Self> {
+        let object_store_config = ObjectStoreConfig::from_env()?;
+        self.node
+            .add_layer(ObjectStoreLayer::new(object_store_config));
+        Ok(self)
+    }
+
+    fn add_metadata_calculator_layer(mut self) -> anyhow::Result<Self> {
```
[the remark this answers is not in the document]
popzxc: We're dynamic at the framework level (you can do what you want) but static at _this particular example_ level.
You're right that this example lacks flexibility -- but at the same time it doesn't need it, we're gradually re-implementing the main node (which is generally static). Moreover, this example for now is mostly a playground for reviewers to understand the framework better and for me to test that layers/tasks/resources work as they should, for the lack of other tests (for now).

It's when we will need to use the same code to spawn either EN or main node, we will need flexibility.

core/node/node_framework/src/implementations/layers/fee_input.rs
```diff
@@ -0,0 +1,82 @@
...
+            .context("GasAdjuster::new()")?;
+        let gas_adjuster = Arc::new(adjuster);
+
+        let batch_fee_input_provider = Arc::new(MainNodeFeeInputProvider::new(
+            gas_adjuster.clone(),
+            FeeModelConfig::from_state_keeper_config(&self.state_keeper_config),
+        ));
+        context.add_resource(FeeInputResource(batch_fee_input_provider))?;
+
+        context.add_task(Box::new(GasAdjusterTask::new(gas_adjuster.clone())));
```
slowli, line 57: Nit: `clone()` is redundant here. Also, is `GasAdjusterTask::new()` intentionally public? It could be replaced by more direct `GasAdjusterTask { gas_adjuster }`.

core/node/node_framework/src/implementations/layers/pools_layer.rs
```diff
@@ -0,0 +1,99 @@
...
+#[derive(Debug)]
+pub struct PoolsLayer {
+    config: PostgresConfig,
+    with_master: bool,
+    with_replica: bool,
+    with_prover: bool,
+}
+
+#[async_trait::async_trait]
+impl WiringLayer for PoolsLayer {
```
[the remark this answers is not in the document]
popzxc: I'm not sure either. I think we'll see what's better along the way -- similarly to how I first decided to keep `ResourceProvider` but now removed it when it became clear that it doesn't really work.

core/node/node_framework/src/implementations/layers/fee_input.rs
popzxc: > Also, is GasAdjusterTask::new() intentionally public? 

Nope, I should just more carefully check what copilot produces for me 😅

core/lib/zksync_core/src/lib.rs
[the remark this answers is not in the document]
popzxc: I look at it from the different angle: you don't even need to understand that something intermediate is instantiated and dropped there. The code tells you that _nothing but mempool_ outlives this block, and if you don't know specifics, it's generally all the information you should have, as if this was a separate function.

core/lib/zksync_core/src/state_keeper/keeper.rs
popzxc: That's a good idea, but I think I'll ship it as a separate PR to not pack too many things together.

core/node/node_framework/src/service/context.rs
popzxc: > Could be done later, of course.

Yup, I'm currently thinking about the right logging structure that would keep track of the initialization process (e.g. "Wiring Layer X requested resource Y"), but I want to tackle it later.

core/node/node_framework/src/resource/unique.rs
```diff
@@ -0,0 +1,35 @@
+use std::sync::{Arc, Mutex};
+
+/// Wrapper for resources that only support one consumer.
+///
+/// Normally, all the resources should support sharing between several tasks,
+/// but there are some cases where a resource should only be consumed by a single task.
+#[derive(Debug)]
+pub struct Unique<T: 'static + Send> {
+    inner: Arc<Mutex<Option<T>>>,
```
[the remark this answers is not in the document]
popzxc: 1: Maybe, what I'm afraid of is that it's not an approach widely used in our codebase, so it may not be intuitive.
2. Again maybe, what I'm afraid of here is that `Unique` is opt-in: you have to explicitly say that you're working with an unique resource on both ends. `take_resource` may be used by accident and cause problems.

I'll think about both points more, and maybe will implement in upcoming PRs.

core/node/node_framework/examples/main_node.rs
```diff
@@ -2,52 +2,109 @@
...
     service::ZkStackService,
 };
 
-/// Resource provider for the main node.
-/// It defines which resources the tasks will receive. This particular provider is stateless, e.g. it always uses
-/// the main node env config, and always knows which resources to provide.
-/// The resource provider can be dynamic, however. For example, we can define a resource provider which may use
-/// different config load scheme (e.g. env variables / protobuf / yaml / toml), and which resources to provide
-/// (e.g. decide whether we need MempoolIO or ExternalIO depending on some config).
-#[derive(Debug)]
```
AnastasiiaVashchuk, line 26: not debug anymore?

popzxc: Good catch, will fix in the next PR!

core/node/node_framework/src/task.rs
```diff
@@ -7,7 +7,7 @@ use crate::service::StopReceiver;
 
 /// A task implementation.
 #[async_trait::async_trait]
-pub trait Task: 'static + Send + Sync {
```
AnastasiiaVashchuk, line 10: why not `Sync` anymore?

popzxc: It doesn't have to be `Sync` in fact (no task is accessed from two thread simultaneously), so it was an incorrect bound.

PR #1047, opened 2024-02-13
feat(dal): `zksync_types::Transaction` to use protobuf for wire encoding (BFT-407)

moshababo:
## What ❔

Adding `zksync_types::Transaction` support for cannonical-protobuf encoding. 
Existing `serde` encoding remains intact—protobuf is used only for the construction of the consensus block payload. 

Closing [BFT-407](https://linear.app/matterlabs/issue/BFT-407/optimize-rlp-encoding-of-transaction-struct).

---

In addition, `zksync_consensus_*` dependencies were updated to include https://github.com/matter-labs/era-consensus/pull/66, and the relocated types were adjusted.

---

## Why ❔

* Better composability with the consensus block payload 
* Support for potential version control / backward compatibility 
* Having a deterministic + unique scheme

## Checklist


core/lib/dal/src/models/storage_sync.rs
```diff
@@ -220,3 +225,272 @@ impl Payload {
         validator::Payload(zksync_protobuf::encode(self))
     }
 }
+
+impl ProtoRepr for proto::Transaction {
+    type Type = Transaction;
+
```
pompon0, line 231: please add a reencoding test (see protobuf_config tests)

core/lib/dal/src/models/storage_sync.rs
```diff
@@ -154,12 +166,8 @@ impl ProtoFmt for Payload {
...
     fn read(message: &Self::Proto) -> anyhow::Result<Self> {
         let mut transactions = Vec::with_capacity(message.transactions.len());
-        for (i, tx) in message.transactions.iter().enumerate() {
-            transactions.push(
-                required(&tx.json)
-                    .and_then(|json_str| Ok(serde_json::from_str(json_str)?))
-                    .with_context(|| format!("transaction[{i}]"))?,
-            );
+        for tx in message.transactions.iter() {
+            transactions.push(tx.read()?)
```
pompon0, line 170: add error context, as it was previously

core/lib/protobuf_config/src/lib.rs
```diff
@@ -26,7 +26,7 @@ mod snapshots_creator;
 mod witness_generator;
 
 pub mod proto;
-mod repr;
+pub mod repr;
```
pompon0, line 29: sorry, but making dal depend on protobuf_config is not acceptable.

pompon0 (commented):
In general lg, we should optimize the proto encoding next (remove duplicate data, perhaps simplify layout).

core/lib/dal/src/models/storage_sync.rs
moshababo: Fixed, changes here: https://github.com/matter-labs/zksync-era/pull/1047/commits/d2bcc7edcfa6777e3353b14928a40e60a31a0f2c

core/lib/protobuf_config/src/lib.rs
moshababo: Extracted to `zksync_protobuf` (https://github.com/matter-labs/era-consensus/pull/66).

core/lib/dal/src/models/storage_sync.rs
moshababo: Added.

slowli (dismissed):
It seems that the transaction schema is still 2x (or 3x?) redundant. Are we OK with this?

core/lib/dal/src/models/proto/mod.proto
```diff
@@ -23,3 +16,75 @@ message Payload {
...
+  optional Execute execute = 4; // required
+  optional uint64 received_timestamp_ms = 5; // required
+  optional bytes raw_bytes = 6; // optional
+}
+
+message L1TxCommonData {
+  optional bytes sender_address = 1; // required; H160
+  optional uint64 serial_id = 2; // required
+  optional uint64 deadline_block = 3; // required
+  optional bytes layer_2_tip_fee = 4; // required; U256
```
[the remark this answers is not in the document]
pompon0: it is definitely not more efficient

"bytes" field requires 1 tag byte + 1 length byte (32 is encode'able as 1 byte varint) + data

your proposal requires 1 tag bytes + 1 length byte for the U256 message + 4* (1 tag byte + 10 varint bytes), which can be reduced to 2 + 4*(1 tag byte + 8 bytes) by using fixed64 instead

core/lib/dal/src/models/proto/mod.proto
```diff
@@ -23,3 +16,75 @@ message Payload {
...
+
+message Transaction {
+  oneof common_data {
+    L1TxCommonData L1 = 1;
+    L2TxCommonData L2 = 2;
+    ProtocolUpgradeTxCommonData ProtocolUpgrade = 3;
+  }
+  optional Execute execute = 4; // required
+  optional uint64 received_timestamp_ms = 5; // required
+  optional bytes raw_bytes = 6; // optional
```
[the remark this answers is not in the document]
pompon0: yes, we should remove it. @moshababo wanted to do that in a separate pr I think.

core/lib/dal/src/models/consensus/tests.rs
```diff
@@ -0,0 +1,105 @@
...
+    assert_transaction_eq(&tx1, &tx2);
+}
+
+/// Checks equality. Top-level `PartialEq` of `Transaction` checks only
+/// the hash hence it's not usable here.
+fn assert_transaction_eq(tx1: &Transaction, tx2: &Transaction) {
+    assert_eq!(tx1.received_timestamp_ms, tx2.received_timestamp_ms);
+    assert_eq!(tx1.raw_bytes, tx2.raw_bytes);
+    assert_eq!(tx1.execute, tx2.execute);
+    match (&tx1.common_data, &tx2.common_data) {
```
pompon0, line 26: this doesn't look good. Instead define a custom type X, which has all the data from Transaction that you want to compare for exact equality, then implement conversion Transaction -> X, and derive PartialEq for X.

core/lib/dal/src/models/proto/mod.proto
```diff
@@ -23,3 +16,75 @@ message Payload {
...
 }
+
+message Transaction {
+  oneof common_data {
+    L1TxCommonData L1 = 1;
+    L2TxCommonData L2 = 2;
+    ProtocolUpgradeTxCommonData ProtocolUpgrade = 3;
+  }
+  optional Execute execute = 4; // required
+  optional uint64 received_timestamp_ms = 5; // required
```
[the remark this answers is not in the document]
slowli: Actually, there's an open question whether it makes sense to transmit / use this field on EN because doing so [seemingly makes](https://linear.app/matterlabs/issue/PLA-782/fix-pending-transactions-subcription-on-en) pending transaction subscriptions buggy. But yeah, right now this field is in use, so it's more of a JFYI comment.

core/lib/dal/src/models/proto/mod.proto
```diff
@@ -23,3 +16,75 @@ message Payload {
...
+  optional bytes contract_address = 1; // required; H160
+  optional bytes calldata = 2; // required
+  optional bytes value = 3; // required; U256
+  repeated bytes factory_deps = 4; // optional
+}
+
+message InputData {
+  optional bytes data = 1; // required;
+  optional bytes hash = 2; // required; H256
+}
```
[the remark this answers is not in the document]
moshababo: > ...but it's somewhat justified given that we need to be able to query transaction fields

That justifies the storing of the normalized data, but what about the storing the raw data? I guess it's a tradeoff to re-encoding. But is that currently used?

Even if not, I'm not sure it makes sense to remove it from the protobuf format, without also removing it entirely from `zksync_types::Transaction` and the DB table, because it will create discrepancies during the lifecycle of the object. Same with the top-level `raw_bytes` which is optional because it's only relevant for L2 transactions, while I'm not sure if it's indeed required in that case. `input` _is_ required due to the hash field (and maybe also due to some other usage), but it can be fixed. 

Another option is to create a derivative struct for `zksync_types::Transaction` which won't have these fields, and will be associated with the protobuf format. But it seems like another workaround. 

Any thoughts? @slowli @pompon0

core/lib/dal/src/models/consensus/tests.rs
moshababo: Fixed, changes here: https://github.com/matter-labs/zksync-era/pull/1047/commits/d0dc588b8818a7eb9c3143a52d596813ec704a99

core/lib/dal/src/models/proto/mod.proto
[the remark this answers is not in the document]
pompon0: yup

core/lib/dal/src/models/proto/mod.proto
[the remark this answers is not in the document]
slowli: > but what about the storing the raw data

I may be mistaken, but I think that fields -> raw data conversion is not always possible, or more precisely, it isn't guaranteed to arrive at identical raw data (i.e., raw data -> fields conversion may be lossy). This would justify Postgres, and to lesser degree, in-memory transaction representation. I may be completely wrong about this though; cc @RomanBrodetski @popzxc

core/lib/dal/Cargo.toml
```diff
@@ -18,9 +18,10 @@ zksync_system_constants = { path = "../constants" }
 zksync_contracts = { path = "../contracts" }
 zksync_types = { path = "../types" }
 zksync_health_check = { path = "../health_check" }
-zksync_consensus_roles = { version = "0.1.0", git = "https://github.com/matter-labs/era-consensus.git", rev = "5b3d383d7a65b0fbe2a771fecf4313f5083be9ae" }
-zksync_consensus_storage = { version = "0.1.0", git = "https://github.com/matter-labs/era-consensus.git", rev = "5b3d383d7a65b0fbe2a771fecf4313f5083be9ae" }
-zksync_protobuf = { version = "0.1.0", git = "https://github.com/matter-labs/era-consensus.git", rev = "5b3d383d7a65b0fbe2a771fecf4313f5083be9ae" }
+zksync_consensus_roles = { version = "0.1.0", git = "https://github.com/matter-labs/era-consensus.git", rev = "97d139969476a004c50f8b4a31ece748e5bee14e" }
+zksync_consensus_storage = { version = "0.1.0", git = "https://github.com/matter-labs/era-consensus.git", rev = "97d139969476a004c50f8b4a31ece748e5bee14e" }
+zksync_protobuf = { version = "0.1.0", git = "https://github.com/matter-labs/era-consensus.git", rev = "97d139969476a004c50f8b4a31ece748e5bee14e" }
+zksync_protobuf_config = { path= "../protobuf_config" }
```
pompon0, line 24: remove this unused dependency

prover/Cargo.lock
```diff
@@ -7137,6 +7138,20 @@ dependencies = [
  "syn 2.0.48",
 ]
 
+[[package]]
+name = "zksync_protobuf_config"
+version = "0.1.0"
+dependencies = [
```
pompon0, line 7144: this shouldn't be here

pompon0 (dismissed):

core/lib/dal/src/models/proto/mod.proto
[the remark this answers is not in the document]
popzxc: Raw data <-> [`TransactionRequest`](https://github.com/matter-labs/zksync-era/blob/98f87cddaf4a41591d5f9861d52d3a21e698e9c4/core/lib/types/src/transaction_request.rs#L202) are expected to *not* be lossy, e.g. the roundtrip should be possible.
`TransactionRequest` -> `L2Tx` is lossy, e.g. it misses chain ID (there actually exists a method to convert `L2Tx` to RLP: [`L2Tx::get_rlp_bytes`](https://github.com/matter-labs/zksync-era/blob/98f87cddaf4a41591d5f9861d52d3a21e698e9c4/core/lib/types/src/l2/mod.rs#L237) which only takes said chain ID as an argument, so this may be the only difference, but I didn't check that.

> I don't know how scary is zksync_types::Transaction refactoring

Very scary, and comes with a high number of wtf/minute. Though if some brave soul would commit to that, I think it may make `zksync_types` much better.

prover/Cargo.lock
moshababo: It is part of the needed lock file update. The build with `--locked` fails without it.

core/lib/dal/Cargo.toml
moshababo: Fixed.

core/lib/dal/src/models/consensus/tests.rs
```diff
@@ -0,0 +1,137 @@
...
+            eth_block: data.eth_block,
+            canonical_tx_hash: data.canonical_tx_hash,
+            to_mint: data.to_mint,
+            refund_recipient: data.refund_recipient,
+        }
+    }
+}
+
+/// Derivative of `ProtocolUpgradeTxCommonData` to facilitate equality comparisons.
+#[derive(PartialEq, Debug)]
```
slowli, line 108: Dumb question: Why is this type required? `ProtocolUpgradeTxCommonData` already implements `PartialEq`, and it seems to have the same fields. And for `L1TxCommonData` I'd argue it'd be acceptable to derive `PartialEq` as well, since it's used in L2 / upgrade counterparts and makes semantic sense.

core/lib/dal/src/models/proto/mod.proto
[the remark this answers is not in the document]
slowli: @moshababo Future work: Since space efficiency isn't the main concern, `bytes` works fine, but I'd still investigate this alternative encoding (e.g., by sampling transactions from the mainnet and looking how much it would save). Can be done separately (of course, before the transaction wire format is finalized).

core/lib/dal/src/models/proto/mod.proto
```diff
@@ -23,3 +16,75 @@ message Payload {
   repeated Transaction transactions = 8;
   optional bool last_in_batch = 10; // required
 }
+
+message Transaction {
+  oneof common_data {
+    L1TxCommonData L1 = 1;
+    L2TxCommonData L2 = 2;
+    ProtocolUpgradeTxCommonData ProtocolUpgrade = 3;
```
slowli, line 24: Bikeshedding: I think it's idiomatic to name `oneof` variants as fields (since on the wire level they are), i.e. in `snake_case`.

core/lib/dal/src/models/proto/mod.proto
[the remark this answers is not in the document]
slowli: > As far as I'm concerned we can have yet another type for representing a normalized transaction, if that's useful.

We can leave the existing codebase as is for the moment, but if we care about normalization, we can implement it for the `Transaction` message, right? @moshababo I'm fine with you doing it in a follow-up PR (although IIUC, it'll mean that some work in this PR – such as structured L2 tx representation in Proobuf – is obsolete), just want to make sure that it'll be done eventually.

prover/Cargo.lock
pompon0: it shouldn't be needed once you removed the dependency from dal/Cargo.roml

core/lib/dal/src/models/consensus/tests.rs
moshababo: > Dumb question: Why is this type required? ProtocolUpgradeTxCommonData already implements PartialEq, and it seems to have the same fields.

You're right, I somehow missed that it's deriving `PartialEq`, unlike `L1TxCommonData`. 

> And for L1TxCommonData I'd argue it'd be acceptable to derive PartialEq as well, since it's used in L2 / upgrade counterparts and makes semantic sense. 

I tried to avoid changing behavior, but I guess it's fine. (same for the `ExecuteTransactionCommon` enum)

---

So as you suggested, I removed the test types for the common data, keeping only the top-level `ComparableTransaction`, needed to override `Transaction`'s `PartialEq` implementation.

core/lib/dal/src/models/proto/mod.proto
moshababo: Fixed.

prover/Cargo.lock
moshababo: I guess I had some local discrepancy. It was fixed when I ran `cargo tree`.

pompon0 (dismissed):

brunoffranca (approved):

pompon0 (approved):

core/lib/dal/src/models/consensus/mod.rs
```diff
@@ -0,0 +1,372 @@
...
+                            .map(h256_to_u256)
+                            .context("common_data.gas_limit")?,
+                        gas_per_pubdata_limit: required(&common_data.gas_per_pubdata_limit)
+                            .and_then(|x| parse_h256(x))
+                            .map(h256_to_u256)
+                            .context("common_data.gas_per_pubdata_limit")?,
+                        op_processing_type: required(&common_data.op_processing_type)
+                            .and_then(|x| {
+                                OpProcessingType::try_from(u8::try_from(*x)?)
+                                    .map_err(|_| anyhow!("u8::try_from"))
```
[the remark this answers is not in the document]
moshababo: The inner `u8::try_from` is still needed because the value is `u32`.
Also, if fails, it returns the error. The `anyhow!` is just for `OpProcessingType::try_from` associated error, who's type is `()`.

core/lib/dal/src/models/consensus/mod.rs
```diff
@@ -0,0 +1,372 @@
...
+    }
+
+    fn build(this: &Self::Type) -> Self {
+        let common_data = match &this.common_data {
+            ExecuteTransactionCommon::L1(data) => {
+                proto::transaction::CommonData::L1(proto::L1TxCommonData {
+                    sender_address: Some(data.sender.as_bytes().into()),
+                    serial_id: Some(data.serial_id.0),
+                    deadline_block: Some(data.deadline_block),
+                    layer_2_tip_fee: Some(u256_to_h256(data.layer_2_tip_fee).as_bytes().into()),
```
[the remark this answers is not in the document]
moshababo: Agree, but the implicit endianness conversions are already widespread in the codebase, in addition to clearly choosing big endian in the implementation, so I'd rather keep it as is.

brunoffranca (approved):

PR #1048, opened 2024-02-13
refactor: Get rid of L1GasPriceProvider trait

popzxc:
## What ❔

- Removes `L1GasPriceProvider` trait.
- Changes the rest of the code to use `BatchFeeModelInputProvider` trait.

## Why ❔

`BatchFeeModelInputProvider` is a successor to `L1GasPriceProvider` and it defines a real interface that modules should use.
`L1GasPriceProvider` was only used to instantiate `BatchFeeModelInputProvider`, so it was generally useless as an abstraction, while still increasing the cognitive complexity.

## Checklist

### DIFF
--- a/core/lib/zksync_core/src/l1_gas_price/mod.rs
+++ b/core/lib/zksync_core/src/l1_gas_price/mod.rs
@@ -23,11 +23,11 @@ pub trait L1GasPriceProvider: fmt::Debug + 'static + Send + Sync {
     fn estimate_effective_pubdata_price(&self) -> u64;
 }
 
-/// Extended version of `L1GasPriceProvider` that can provide parameters
-/// to set the fee for an L1 transaction, taking the desired mining time into account.
+/// Abstraction that provides parameters to set the fee for an L1 transaction, taking the desired
+/// mining time into account.
 ///
 /// This trait, as a bound, should only be used in components that actually sign and send transactions.
-pub trait L1TxParamsProvider: L1GasPriceProvider {
+pub trait L1TxParamsProvider: fmt::Debug + 'static + Send + Sync {
     /// Returns the recommended `max_fee_per_gas` value (EIP1559).
     fn get_base_fee(&self, time_in_mempool: u32) -> u64;
 
--- a/core/lib/zksync_core/src/l1_gas_price/mod.rs
+++ b/core/lib/zksync_core/src/l1_gas_price/mod.rs
@@ -10,19 +10,6 @@ mod gas_adjuster;
 mod main_node_fetcher;
 pub mod singleton;
 
-/// Abstraction that provides information about the L1 gas price currently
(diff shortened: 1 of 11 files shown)


core/lib/zksync_core/src/lib.rs
```diff
@@ -402,6 +402,10 @@ pub async fn initialize_components(
                 .get_or_init()
                 .await
                 .context("gas_adjuster.get_or_init()")?;
+            let batch_fee_input_provider = Arc::new(MainNodeFeeInputProvider::new(
+                bounded_gas_adjuster,
+                FeeModelConfig::from_state_keeper_config(&state_keeper_config),
+            ));
```
Deniallugo, line 408: do you really need to construct FeeInput provider 3 times?

core/lib/zksync_core/src/l1_gas_price/mod.rs
```diff
@@ -10,24 +10,11 @@ mod gas_adjuster;
...
 pub mod singleton;
 
-/// Abstraction that provides information about the L1 gas price currently
-/// observed by the application.
-pub trait L1GasPriceProvider: fmt::Debug + 'static + Send + Sync {
-    /// Returns a best guess of a realistic value for the L1 gas price.
-    /// Return value is in wei.
-    fn estimate_effective_gas_price(&self) -> u64;
-
-    /// Returns a best guess of a realistic value for the L1 pubdata price.
```
Deniallugo, line 20: Was it used for previous fee model only?

core/lib/zksync_core/src/lib.rs
popzxc: Just like gas adjuster, which is also created via `get_or_init` 3 times (so that it is only instantiated on demand).
There exists an open question of whether we want to merge `GasAdjuster` (or `GasAdjusterSingleton`) and `MainNodeFeeInputProvider`, but I think it's out of scope for this PR (as it's about getting rid of abstraction).

core/lib/zksync_core/src/l1_gas_price/mod.rs
popzxc: No, `estimate_effective_gas_price` seems to be invoked for both old and new model, and `estimate_effective_pubdata_price` seems to be invoked only for the new model.
They still exist, but now they're just methods on `GasAdjuster`, not a part of some trait.

core/lib/zksync_core/src/lib.rs
Deniallugo: yeah, but get or init returns the only instance. 
but here you have 3 instances. 
could it lead to unnecessary requests to the server?

popzxc: No

Deniallugo (approved):

PR #1050, opened 2024-02-13
feat: improving verification key generation

mm-zk:
## What ❔

* Removed vk_generator.rs that was no longer used
* Added more logging to VK generation
* Added an early check to detect if CRS file is missing

## Why ❔

* vk_generator logic code is already present in witness_generator crate
* more logging helps to see what is happening
* early check saves people 30 minutes, if they didn't configure their setup correctly.


prover/vk_setup_data_generator_server_fri/src/vk_generator.rs
```diff
@@ -27,11 +56,125 @@ fn main() -> anyhow::Result<()> {
...
 }
 
+fn get_all_basic_circuits(
+    geometry: &GeometryConfig,
+) -> Vec<
+    ZkSyncBaseLayerCircuit<GoldilocksField, VmWitnessOracle<GoldilocksField>, Poseidon2Goldilocks>,
+> {
+    vec![
+        ZkSyncBaseLayerCircuit::MainVM(ZkSyncUniformCircuitInstance {
+            witness: AtomicCell::new(Some(VmCircuitWitness::<
```
shamatar, line 66: None is enough, correct and serves as self-protection for VK generation

prover/vk_setup_data_generator_server_fri/src/vk_generator.rs
```diff
@@ -27,11 +56,125 @@ fn main() -> anyhow::Result<()> {
...
+                LinearHasherCircuitInstanceWitness::<GoldilocksField>::default(),
+            )),
+            config: Arc::new(geometry.limit_for_l1_messages_pudata_hasher as usize),
+            round_function: Arc::new(Poseidon2Goldilocks),
+            expected_public_input: None,
+        }),
+    ]
+}
+
 pub fn generate_basic_circuit_vks() -> anyhow::Result<()> {
```
shamatar, line 173: Is it useful with such signature? Maybe change is to accept `&mut dyn DataSource` ?

prover/vk_setup_data_generator_server_fri/src/vk_generator.rs
mm-zk: thanks - applied these changes in test_harness directly.

prover/vk_setup_data_generator_server_fri/src/vk_generator.rs
mm-zk: moved changes to test_harness directly.

RomanBrodetski:
@mm-zk why `chore`? You don't want this to appear in the Changelog? THat's what Chore does

mm-zk:
@RomanBrodetski - good point. Updated to feat.

PR #1053, opened 2024-02-13
fix: update harness to include fix to new boojum OOM

joonazan:
Upgrading boojum causes OOM. @aikixd made a fix to zkevm_test_harness that changes the mode of boojum to one that uses less memory. This PR updates harness to that version.


jespercv:
nice catch.. was about to do this

PR #1054, opened 2024-02-13
test(node): add basic unit tests for ResourceCollection

AnastasiiaVashchuk:
## Checklist

### DIFF
--- a/core/node/node_framework/src/resource/resource_collection.rs
+++ b/core/node/node_framework/src/resource/resource_collection.rs
@@ -64,7 +64,7 @@ impl<T> fmt::Debug for ResourceCollection<T> {
     }
 }
 
-#[derive(Debug, Error)]
+#[derive(Debug, Error, PartialEq)]
 pub enum ResourceCollectionError {
     #[error("Adding resources to the collection is not allowed after the wiring is complete")]
     AlreadyWired,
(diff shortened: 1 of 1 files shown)


core/node/node_framework/src/resource/resource_collection.rs
```diff
@@ -104,3 +104,64 @@ impl<T: Resource + Clone> ResourceCollection<T> {
...
+        let mut collection = ResourceCollection::<TestResource>::new();
+
+        collection.on_resource_wired();
+
+        let resolved = collection.resolve().await;
+        assert_eq!(resolved, vec![]);
+
+        // check the case that it is not allowed to resolve the collection when it is not wired
+        let collection = ResourceCollection::<TestResource>::new();
+        let result = timeout(Duration::from_secs(1), collection.resolve()).await;
```
popzxc, line 164: I think 50ms would be sufficient for the purpose of this test. 1 second is a lot of time, and we don't want to make the tests longer than they should be.

popzxc: Also I think it makes sense to extend this test by calling `on_resource_wired()` and making sure that it can be resolved after that.

core/node/node_framework/src/resource/resource_collection.rs
```diff
@@ -104,3 +104,64 @@ impl<T: Resource + Clone> ResourceCollection<T> {
...
+    struct TestResource(Arc<u8>);
+
+    impl Resource for TestResource {
+        fn resource_id() -> ResourceId {
+            ResourceId::new("test_resource")
+        }
+    }
+
+    #[tokio::test]
+    async fn test_push() {
```
popzxc, line 126: Probably it makes sense to ensure that we can push several different items to the collection.

core/node/node_framework/src/resource/resource_collection.rs
```diff
@@ -104,3 +104,67 @@ impl<T: Resource + Clone> ResourceCollection<T> {
...
+        assert_eq!(
+            rc_clone.push(resource),
+            Err(ResourceCollectionError::AlreadyWired)
+        );
+    }
+
+    #[tokio::test]
+    async fn test_resolve() {
+        let mut collection = ResourceCollection::<TestResource>::new();
+        let timeout = Duration::from_millis(500);
```
popzxc, line 156: 500 milliseconds is still a lot, 50 would be sufficient :)

popzxc (approved):

PR #1058, opened 2024-02-14
chore: Use gauge for call tracer metrics

Artemka374:
## What ❔

Use Gauge instead of histogram for CallTracer metrics

## Why ❔

For better obsevability

## Checklist


RomanBrodetski:
but.. why?

Artemka374:
It was request from @Deniallugo, they want to see more concrete maximum values of the call stack depth. I couldn't find a better way to do so, except adjusting buckets/not using histogram.

PR #1065, opened 2024-02-14
refactor(state_keeper): Rename batch executor family of types

popzxc:
## What ❔

- Does a set of renames:
  - `L1BatchExecutorBuilder` -> `BatchExecutor`
  - `MainBatchExecutorBuilder` -> `MainBatchExecutor`
  - `BatchExecutor` -> `CommandReceiver` (as it's just a "processing" part of `MainBatchExecutor`)
  - Similar changes in the framework.
- Moves the implementation of `MainBatchExecutor` to a separate file.
- Removes some temporary misplaced constructor for `BatchExecutorHandle`.

## Why ❔

- Less Enterprise™
- Better logical structure

## Checklist

### DIFF
--- a/core/bin/external_node/src/main.rs
+++ b/core/bin/external_node/src/main.rs
@@ -25,8 +25,8 @@ use zksync_core::{
     reorg_detector::ReorgDetector,
     setup_sigint_handler,
     state_keeper::{
-        seal_criteria::NoopSealer, L1BatchExecutorBuilder, MainBatchExecutorBuilder,
-        MiniblockSealer, MiniblockSealerHandle, ZkSyncStateKeeper,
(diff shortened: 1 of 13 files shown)


popzxc:
>Seems vague, I would've picked something more obvious (VmBatchExecturor?)

It is kinda vague, because it really implements the "main" flow:
1. It loads state from Postgres
2. It puts it in RocksDB
3. It really executes the batches on VM.

So it's like "real" flow. But we can other flows in theory (and there has been demand for it in the past), like, for example, do not use RocksDB and execute over Postgres directly. Or customize it to not interact with VM. Or whatever.

I don't like the naming either, but wasn't able to come up with something that is short and precise, so thought that it's fine for now.

PR #1075, opened 2024-02-15
chore(node_framework): Emit logs on context/resource access

popzxc:
## What ❔

- Adds more logs when the node is being configured. We can certainly improve it further (e.g. print task name in resource-related logs or use prettier type output for Unique), but I think it's a good starting point.

## Why ❔

- Logs are helpful for debugging, especially in dynamic systems like the new framework.

New logs look roughly as follows:
image

### DIFF
--- a/core/node/node_framework/src/implementations/resources/pools.rs
+++ b/core/node/node_framework/src/implementations/resources/pools.rs
@@ -1,10 +1,18 @@
+use std::sync::{
+    atomic::{AtomicU32, Ordering},
+    Arc,
+};
+
 use zksync_dal::{connection::ConnectionPoolBuilder, ConnectionPool};
 
 use crate::resource::Resource;
 
 /// Represents a connection pool to the master database.
 #[derive(Debug, Clone)]
-pub struct MasterPoolResource(ConnectionPoolBuilder);
+pub struct MasterPoolResource {
+    connections_count: Arc<AtomicU32>,
+    builder: ConnectionPoolBuilder,
+}
 
 impl Resource for MasterPoolResource {
     fn resource_id() -> crate::resource::ResourceId {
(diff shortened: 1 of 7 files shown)


core/node/node_framework/src/implementations/resources/pools.rs
```diff
@@ -1,10 +1,18 @@
...
+
 use zksync_dal::{connection::ConnectionPoolBuilder, ConnectionPool};
 
 use crate::resource::Resource;
 
 /// Represents a connection pool to the master database.
 #[derive(Debug, Clone)]
-pub struct MasterPoolResource(ConnectionPoolBuilder);
+pub struct MasterPoolResource {
+    connections_count: Arc<AtomicU32>,
```
Deniallugo, line 13: should we set a limit?  Otherwise i don't see much value here

popzxc: I don't think we should. It doesn't change behavior, but it'll make the debugging easier in the future.
Today we don't know how much connections do we use, and we can certainly create more connections than DB supports (which already happened in our CI and caused problems).

core/node/node_framework/src/implementations/resources/pools.rs
```diff
@@ -14,21 +22,48 @@ impl Resource for MasterPoolResource {
...
     pub async fn get(&self) -> anyhow::Result<ConnectionPool> {
-        self.0.build().await
+        let result = self.builder.build().await;
+
+        if result.is_ok() {
+            self.connections_count
+                .fetch_add(self.builder.max_size(), Ordering::Relaxed);
+            let total_connections = self.connections_count.load(Ordering::Relaxed);
+            tracing::info!(
+                "Created a new master pool. Master pool total connections count: {total_connections}"
```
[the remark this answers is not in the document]
popzxc: Total possible connections to the database by a single ZkStackService.

popzxc:
> I wonder what will be the impact on logs

@EmilLuta Next to none. 1. These logs are printed once at node launch, and most of impact comes from stuff like logging in API request handlers. 2. For one-shot machines you can always configure logging as `zksync_node_framework=warn` so that info level logs do not appear.

PR #1077, opened 2024-02-15
feat(vlog): Remove env getters from vlog

popzxc:
## What ❔

- Removes `*_from_env` functions from `vlog` crate.
- Introduces `ObservabilityConfig` for core and EN with env variable based constructors that match the old behavior.
- Changes the binaries to use new approach.

## Why ❔

- `vlog` is a generic library, so it should have no assumptions on the configuration system. It should be for users of `vlog` to decide where they get the configuration from.


core/bin/external_node/src/config/observability.rs
```diff
@@ -0,0 +1,39 @@
...
+    let sentry_url = if let Ok(sentry_url) = std::env::var("MISC_SENTRY_URL") {
+        if sentry_url == "unset" {
+            None
+        } else {
+            Some(sentry_url)
+        }
+    } else {
+        None
+    };
+    let sentry_environment = std::env::var("EN_SENTRY_ENVIRONMENT").ok().or_else(|| {
```
popzxc, line 15: Note: this is different from `ObservabilityConfig::from_env()`.

core/lib/config/src/configs/observability.rs
```diff
@@ -0,0 +1,12 @@
...
+/// logging and sentry integration.
+#[derive(Debug, Clone)]
+pub struct ObservabilityConfig {
+    /// URL of the Sentry instance to send events to.
+    pub sentry_url: Option<String>,
+    /// Name of the environment to use in Sentry.
+    pub sentry_environment: Option<String>,
+    /// Format of the logs as expected by the `vlog` crate.
+    /// Currently must be either `plain` or `json`.
+    pub log_format: String,
```
popzxc, line 11: I could've made it an enum, but it felt a bit off; having `impl FromStr for vlog::LogFormat` felt enough for that purpose.

core/bin/external_node/src/config/observability.rs
```diff
@@ -0,0 +1,39 @@
...
+        }
+    });
+    let log_format = if let Ok(log_format) = std::env::var("MISC_LOG_FORMAT") {
+        if log_format != "plain" && log_format != "json" {
+            anyhow::bail!("MISC_LOG_FORMAT has an unexpected value {}", log_format);
+        }
+        log_format
+    } else {
+        "plain".to_string()
+    };
```
Deniallugo, line 32: why you don't parse the string to vlog::LogFormat here?

core/lib/config/src/configs/observability.rs
Deniallugo: You do it every time anyway, so why not just do it here?

popzxc: What do you mean? Make `zksync_config` depend on `vlog`? This will make any crate that depends on `zksync_config` (most of which are libraries) transitively depend on sentry and tracing-subscriber. Make `vlog` depend on `zksync_config`? 
This also doesn't feel right, since we're trying to move away from specific crates being aware of the whole stack.

What we can do is to introduce a `enum LogFormat { Plain, Json }` here and do the conversion of `zksync_config::LogFormat` to `vlog::LogFormat`, but we'll still have to do it each time.

Another option would be to introduce one more "common" crate for `LogFormat` enum, but that'd _certainly_ be an overkill.

core/bin/external_node/src/config/observability.rs
popzxc: To keep logic similar to `zksync_env_config` crate, where we don't have direct access to `vlog`.

core/lib/config/src/configs/observability.rs
popzxc: Invalid in what sense? You can provide a wrong sentry URL as well. Your code will just fail not at the phase of config deserialization, but when you will try to parse the log format - which is generally the very next thing you would do after deserializing the config. And similarly, that'd also be the step where the code will fail if you'll provide the invalid Sentry URL, so it's pretty similar in that sense.

Deniallugo (dismissed):

perekopskiy (commented):
I think you should add `ObservabilityConfig` to protobug config crate as well

perekopskiy (approved):

Deniallugo (approved):

PR #1078, opened 2024-02-15
feat(api): add a config flag for disabling filter api

itegulov:
## What ❔

Added a config flag to disable all filter-related methods. If turned on, they report that the method is not implemented.

## Why ❔

If a node is behind a load balancer then there is no way for client to reliably hit the same node where he/she created the filter. Hence supporting this is a very awkward UX. Making this a flag since external nodes might want to turn this on if they only have a single node mapped to a static address.

## Checklist


itegulov:
FYI since `eth_getLogs` is stateless and can works fine even behind load balancer, I decided to make it unaffected by `filters_disabled`

core/bin/external_node/src/config/mod.rs
```diff
@@ -136,6 +136,14 @@ pub struct OptionalENConfig {
...
     /// Enabled JSON RPC API namespaces.
     api_namespaces: Option<Vec<Namespace>>,
+    /// Whether to support methods installing filters and querying filter changes.
+    ///
+    /// Rationale for setting this to `false`:
+    /// Filters are local to the specific node they were created at. Meaning if
+    /// there are multiple nodes behind a load balancer the client cannot reliably
+    /// query the previously created filter as the request might get routed to a
+    /// different node.
+    pub filters_enabled: Option<bool>,
```
slowli, line 146: - Why is this value optional? It's possible to set a `bool` value to the default using `#[serde(default = ..)]`, like for other fields in this struct. IMO, it'd be simpler to reason about and more idiomatic. Ditto for a similar field in other configs.
- Bikeshedding: IMO, it would be easier to reason about an inverse variable (`filters_disabled`). In this case, the absence of the variable is equivalent to a `false` value, which looks more intuitive. Additionally, you'd be able to use `#[serde(default)]` instead of `#[serde(default = ..)]` 🙃

core/bin/external_node/src/config/mod.rs
```diff
@@ -136,6 +136,14 @@ pub struct OptionalENConfig {
     latest_values_cache_size_mb: usize,
     /// Enabled JSON RPC API namespaces.
     api_namespaces: Option<Vec<Namespace>>,
+    /// Whether to support methods installing filters and querying filter changes.
+    ///
+    /// Rationale for setting this to `false`:
```
slowli, line 141: Bikeshedding: I readily admit I'm not an expert on English, but the first time I read this comment, it looked like it explained why `false` is the default value for this field (which isn't the case; the default value is `true`). IMO, it'd be more straightforward to formulate this something like "When to set this value to `false`".

core/lib/zksync_core/src/api_server/web3/namespaces/eth.rs
```diff
@@ -229,6 +229,9 @@ impl EthNamespace {
         const METHOD_NAME: &str = "get_filter_logs";
 
         let method_latency = API_METRICS.start_call(METHOD_NAME);
+        if !self.state.api_config.filters_enabled {
```
slowli, line 232: This works, but IMO it'd be more idiomatic if `RpcState.installed_filters` would be made `Option`al and set to `None` when filters are disabled. If this was the case, you couldn't accidentally miss filter uses even if you wanted.

core/lib/zksync_core/src/api_server/web3/tests/filters.rs
```diff
@@ -263,3 +265,42 @@ impl HttpTest for LogFilterChangesWithBlockBoundariesTest {
...
+
+#[derive(Debug)]
+struct DisableFiltersTest;
+
+#[async_trait]
+impl HttpTest for DisableFiltersTest {
+    async fn test(&self, client: &HttpClient, _pool: &ConnectionPool) -> anyhow::Result<()> {
+        use jsonrpsee::{core::client::Error, types::error::ErrorCode};
+
+        fn assert_not_implemented<T: Debug>(result: Result<T, Error>) {
```
slowli, line 277: Nit: I'd suggest to move this function to the module level; it'd be easier to read, and it could be later reused in other tests.

core/lib/zksync_core/src/api_server/web3/tests/mod.rs
```diff
@@ -141,10 +144,13 @@ async fn spawn_server(
     websocket_requests_per_minute_limit: Option<NonZeroU32>,
     tx_executor: MockTransactionExecutor,
     stop_receiver: watch::Receiver<bool>,
+    api_config: Option<InternalApiConfig>,
```
slowli, line 147: Bikeshedding: This looks weird. When we've talked about this, I've meant passing `api_config: InternalApiConfig` to this function and `spawn_http_server` *instead of* `network_config`; sorry for not being clear. `InternalApiConfig` can be constructed on the caller side; it'd lead to some code duplication, but it'd be fairly minor.

core/bin/external_node/src/config/mod.rs
itegulov: Fixed in https://github.com/matter-labs/zksync-era/pull/1078/commits/5df4df4ec48805e91c24e23945087d364c0b8c47 and https://github.com/matter-labs/zksync-era/pull/1078/commits/53c747832abc619e453f94a59c703132f512b973

> IMO, it would be easier to reason about an inverse variable (filters_disabled)

I usually try to avoid "negative" config values as you can sometimes find yourself using it with negation and it is hard to reason about double negation. Not a strong opinion on this though

core/lib/zksync_core/src/api_server/web3/namespaces/eth.rs
itegulov: Very good point, fixed in https://github.com/matter-labs/zksync-era/pull/1078/commits/71bb815d592c8c119cd0ad568bba4a60548ca885

core/lib/zksync_core/src/api_server/web3/tests/filters.rs
itegulov: Hmm, my reasoning behind making this scoped was that it uses collision-heavy types (`Error` and `ErrorCode`) but fair enough

On this point, do we have a styling guide perhaps? Just so I can get aligned on these things with everyone else

core/lib/zksync_core/src/api_server/web3/namespaces/eth.rs
```diff
@@ -613,6 +612,9 @@ impl EthNamespace {
         const METHOD_NAME: &str = "new_filter";
 
         let method_latency = API_METRICS.start_call(METHOD_NAME);
+        let Some(installed_filters) = &self.state.installed_filters else {
+            return Err(Web3Error::NotImplemented);
+        };
```
slowli, line 617: Nit: You can use `ok_or` to convert an `Option` to a `Result` more idiomatically.

slowli (approved):
Nice.

popzxc (approved):

popzxc (approved):

slowli (approved):

PR #1080, opened 2024-02-15
refactor(state-keeper): Propagate I/O errors in state keeper

slowli:
## What ❔

- Propagates I/O errors in the state keeper using `anyhow::Result`, adding context to calls where appropriate.
- Propagates DB errors in `FactoryDepsDal`.

## Why ❔

Propagating errors makes it easier to attach additional context to them; thus, it improves DevEx and code maintainability.

## Checklist


core/lib/zksync_core/src/api_server/web3/namespaces/zks.rs
```diff
@@ -548,12 +548,16 @@ impl ZksNamespace {
         &self,
         hash: H256,
     ) -> Result<Option<Vec<u8>>, Web3Error> {
+        // FIXME: is `Vec<u8>` vs `Bytes` intentional here?
```
slowli, line 551: It looks weird; most of other RPC methods use `Bytes` under similar circumstances. This RPC method is documented though (although AFAICT by the API dashboard, it's not used by anyone), so it may be too late to change the response format.

RomanBrodetski: I don't think it's intentional

popzxc (approved):

core/lib/zksync_core/src/api_server/web3/namespaces/zks.rs
slowli: So can it be changed? If so, does this require a specific procedure, or can I do this in this PR?

popzxc: This seems to be invoked by `MainNodeClient::fetch_system_contract_by_hash` if the system contract is absent on the EN, but this indeed shouldn't happen often. I guess we can take the risk and just do it.

popzxc: (though as usual I'll prefer a separate PR for that)

slowli: Certainly makes sense to implement separately; I've created a task for it in Linear.

popzxc (approved):

PR #1091, opened 2024-02-15
Update index.ts

mankzan:
This is the description text that is provided as It describes what the f command does.


Pofiz:
#airdropfarming

PR #1096, opened 2024-02-16
fix(state-keeper): Add GasForBatchTip criterion

perekopskiy:
## What ❔

- add GasForBatchTip criterion instead of checking gas in batch executor
- removes the assumption that a single tx can't consume whole batch gas (with or without batch tip)

## Why ❔

That check in batch executor
```
if !vm.has_enough_gas_for_batch_tip() {
    return TxExecutionResult::BootloaderOutOfGasForBlockTip;
}
```
conceptually was a criterion and should be extracted to a separate seal criteria. This way EN doesn't perform this check which is a correct behavior and code is cleaner.

## Checklist


core/lib/multivm/src/glue/types/vm/vm_partial_execution_result.rs
```diff
@@ -36,9 +36,11 @@ impl GlueFrom<crate::vm_m6::vm_instance::VmPartialExecutionResult>
             statistics: crate::interface::VmExecutionStatistics {
                 contracts_used: value.contracts_used,
                 cycles_used: value.cycles_used,
-                gas_used: value.computational_gas_used,
```
Deniallugo, line 39: It was before, why it's zero now?

core/lib/zksync_core/src/api_server/tx_sender/mod.rs
```diff
@@ -953,11 +953,11 @@ impl TxSender {
...
         let protocol_version = ProtocolVersionId::latest();
         let seal_data = SealData::for_transaction(transaction, tx_metrics, protocol_version);
-        if let Some(reason) = self
-            .0
-            .sealer
-            .find_unexecutable_reason(&seal_data, protocol_version)
-        {
+        if let Some(reason) = self.0.sealer.find_unexecutable_reason(
+            &seal_data,
+            tx_metrics.gas_remaining,
```
StanislavBreadless, line 958: maybe include it into the `seal_data`?

core/lib/zksync_core/src/state_keeper/seal_criteria/criteria/gas_for_batch_tip.rs
```diff
@@ -0,0 +1,95 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use assert_matches::assert_matches;
+
+    use super::*;
+
+    #[test]
+    fn test_slots_seal_criterion() {
```
StanislavBreadless, line 49: the name of the test is incorrect

StanislavBreadless (commented):
overall nice, I left a few comments

core/lib/multivm/src/glue/types/vm/vm_partial_execution_result.rs
perekopskiy: I think it was a bug before, I decided to explicitly pass 0 instead of "something"

core/lib/zksync_core/src/api_server/tx_sender/mod.rs
perekopskiy: Then it would duplicated for `block_seal_data` and `tx_seal_data`. I'll check, maybe it indeed makes sens to include it and have less args in methods definitions

StanislavBreadless (approved):

Deniallugo (approved):

PR #1099, opened 2024-02-16
fix: Clean prover tables from core database

EmilLuta:
This is PR 2 in the Prover/ Core DAL split (#1029).

This PR cleans unused/ unnnecessary tables that currently exist in core database. All these tables have been officially moved to prover database. All environemnts (staging, dev env) have been migrated and are using two databases (core db for core components, prover db for prover components).

This has long been a tech debt and source of confusion for anyone trying to understand the system. The cleanup gets rid of "`witness_inputs_fri` from which database, again?" or "I don't see it in the tables; Did you connect to the right database?" kind of questions.


RomanBrodetski (approved):

RomanBrodetski:
@EmilLuta we also have some tables related to proofs that are not used, **and are there from ZkSync Lite**. I think it's `aggregated_proof` and `proof`. Let's also delete them here maybe?

PR #1102, opened 2024-02-16
test(node_framework): Add set of tests for ZkStackService component

AnastasiiaVashchuk:
## What ❔

Created 7 unit tests to check the behaviour of the `ZkStackService` component in the node framework.
The next step is to start to work on `proof_data_handler ` layer.
Also I have fixed small typo: `ZksyncNode` ->  `ZkStackService`

## Checklist


core/node/node_framework/src/service/tests.rs
```diff
@@ -0,0 +1,235 @@
...
+        "remaining_task"
+    }
+    async fn run(self: Box<Self>, stop_receiver: StopReceiver) -> anyhow::Result<()> {
+        loop {
+            if *stop_receiver.0.borrow() {
+                let mut guard = self.0.lock().await;
+                *guard = false;
+                break;
+            }
+        }
```
popzxc, line 191: I guess you can get rid of the loop by using `stop_receiver.0.changed().await` before setting `guard` to `true`.

core/node/node_framework/src/service/tests.rs
```diff
@@ -0,0 +1,235 @@
+use std::sync::Arc;
+
+use anyhow::anyhow;
+use futures::lock::Mutex;
```
popzxc, line 4: You can use mutexes from `std` to avoid dealing with multiple tokio Runtimes, there is no need for async mutexes here.

core/node/node_framework/src/service/tests.rs
```diff
@@ -0,0 +1,235 @@
...
+    }
+}
+
+// Check ZkStack Service's `run()` method expected behavior.
+// There are 2 tests, one for successful task run and another for remaining task run.
+// Second test case relies on the first one.
+#[test]
+fn test_task_run() {
+    // case 1: run the task and check it updates resource successfully before stop signal received.
+    let resource = Arc::new(Mutex::new(false));
```
popzxc, line 202: Better to choose more descriptive names. E.g. here it would be `task_run_invoked` and below it would be `task_stop_signal_received`.

core/node/node_framework/src/service/tests.rs
```diff
@@ -0,0 +1,235 @@
...
+    let mut zk_stack_service = ZkStackService::new().unwrap();
+    zk_stack_service
+        .add_layer(SuccessfulTaskLayer(resource.clone()))
+        .add_layer(RemainingTaskLayer(resource.clone()));
+
+    assert!(zk_stack_service.run().is_ok());
+    tokio::runtime::Runtime::new()
+        .unwrap()
+        .block_on(async { assert!(!*resource.lock().await) });
```
popzxc, line 224: I think the issue would resolve itself with use of std mutexes and more descriptive variable names, but as a general comment: mixing (except for trivial cases) some random logic and test assertions looks confusing. Usually it's better to keep them separate, e.g. 

```rust
let thing_i_want_to_test = ...; 
assert(thing_i_want_to_test , "What went wrong");
```

core/node/node_framework/src/service/tests.rs
```diff
@@ -0,0 +1,235 @@
...
+}
+
+// ZkStack Service's `add_layer()` method has to add multiple layers into `self.layers`.
+#[test]
+fn test_add_layer() {
+    let mut zk_stack_service = ZkStackService::new().unwrap();
+    zk_stack_service
+        .add_layer(DefaultLayer)
+        .add_layer(DefaultLayer);
+    assert_eq!(2, zk_stack_service.layers.len());
```
popzxc, line 45: It's almost always a good idea to provide a message for assertions to make it easier to debug if some of tests fail. E.g. here the default message would be something like "Assertion failed, expected 2 but got 1". If you add a message like "Incorrect number of layers in the service", it'll be much more helpful.

core/node/node_framework/src/service/tests.rs
```diff
@@ -0,0 +1,235 @@
...
+
+// Check ZkStack Service's `stop_receiver()` method returns new `Receiver` for it's own sender,
+// and not arbitrary one.
+#[test]
+fn test_stop_receiver() {
+    let zk_stack_service = ZkStackService::new().unwrap();
+    let _receiver1 = zk_stack_service.stop_sender.subscribe();
+    let _receiver2 = zk_stack_service.stop_receiver();
+    assert_eq!(2, zk_stack_service.stop_sender.receiver_count());
+}
```
popzxc, line 235: I'd argue that it's an implementation detail. You have a test that checks if the stop receiver works in general -- it's sufficient. The implementation may change, and gluing tests to some specific implementation makes development unnecessarily hard.

core/node/node_framework/src/service/tests.rs
```diff
@@ -0,0 +1,235 @@
...
+#[test]
+fn test_new_with_nested_runtime() {
+    let runtime = Runtime::new().unwrap();
+
+    runtime.block_on(async {
+        assert_eq!(
+            ZkStackService::new().unwrap_err().to_string(),
+            "Detected a Tokio Runtime. ZkStackService manages its own runtime and does not support nested runtimes".to_string()
+        );
+    })
```
popzxc, line 21: Btw here is an example where you can detach logic from assertion.

You can have `let run_result = runtime.block_on(...);` and then check if `run_result` is an expected error.

AnastasiiaVashchuk: why this is better?

core/node/node_framework/src/service/tests.rs
AnastasiiaVashchuk: Do you suggest getting rid of this?

core/node/node_framework/src/service/tests.rs
popzxc: https://github.com/matter-labs/zksync-era/pull/1102#discussion_r1498776007

core/node/node_framework/src/service/tests.rs
popzxc: Yup

core/node/node_framework/src/service/tests.rs
```diff
@@ -0,0 +1,232 @@
...
+    // case 1: run the task and check it updates resource successfully before stop signal received.
+    let some_bool_resource = Arc::new(Mutex::new(false));
+    let mut zk_stack_service = ZkStackService::new().unwrap();
+    zk_stack_service.add_layer(SuccessfulTaskLayer(some_bool_resource.clone()));
+
+    assert!(
+        zk_stack_service.run().is_ok(),
+        "ZkStackService run failed, but it shouldn't"
+    );
+    let actual_resource_value = *some_bool_resource.lock().unwrap();
```
popzxc, line 211: I think we can find a name for this variable that would represent the logical meaning.
In this case, for example, how about `task_was_run` (because the flag is set when `run()` method is invoked)?

core/node/node_framework/src/service/tests.rs
```diff
@@ -0,0 +1,232 @@
...
+    let mut zk_stack_service = ZkStackService::new().unwrap();
+    zk_stack_service
+        .add_layer(SuccessfulTaskLayer(some_bool_resource.clone()))
+        .add_layer(RemainingTaskLayer(some_bool_resource.clone()));
+
+    assert!(
+        zk_stack_service.run().is_ok(),
+        "ZkStackService run failed, but it shouldn't"
+    );
+    let actual_resource_value = *some_bool_resource.lock().unwrap();
```
popzxc, line 228: ...and here it would be `stop_signal_sent` (because it's set when we receive the stop signal).

core/node/node_framework/src/service/tests.rs
```diff
@@ -0,0 +1,232 @@
...
+    }
+}
+
+// Check ZkStack Service's `run()` method expected behavior.
+// There are 2 tests, one for successful task run and another for remaining task run.
+// Second test case relies on the first one.
+#[test]
+fn test_task_run() {
+    // case 1: run the task and check it updates resource successfully before stop signal received.
+    let some_bool_resource = Arc::new(Mutex::new(false));
```
popzxc, line 203: ...which would show us that we can find better name here too! Would you like `run_flag_resource`?

core/node/node_framework/src/service/tests.rs
```diff
@@ -0,0 +1,232 @@
...
+        "ZkStackService run failed, but it shouldn't"
+    );
+    let actual_resource_value = *some_bool_resource.lock().unwrap();
+    // task changed resource value to true
+    assert!(actual_resource_value, "Incorrect resource value");
+
+    // case 2: the idea is to check we are still running remaining tasks after stop signal received.
+    // The flow is that successful_task updates resource to true (as checked in the case 1 above)
+    // and remaining_task updates it back to false.
+    let some_bool_resource = Arc::new(Mutex::new(false));
```
popzxc, line 218: I'll leave this one as an exercise to a reader 🙃

core/node/node_framework/src/service/tests.rs
```diff
@@ -0,0 +1,232 @@
...
+        zk_stack_service.run().is_ok(),
+        "ZkStackService run failed, but it shouldn't"
+    );
+    let actual_resource_value = *some_bool_resource.lock().unwrap();
+    // task changed resource value to true
+    assert!(actual_resource_value, "Incorrect resource value");
+
+    // case 2: the idea is to check we are still running remaining tasks after stop signal received.
+    // The flow is that successful_task updates resource to true (as checked in the case 1 above)
+    // and remaining_task updates it back to false.
```
popzxc, line 217: The spellcheck CI job seems to be complaining about stuff like "remaining_task" and such. To make it work, use markdown ticks, e.g. it wouldn't complain about "\`remaining_task\`"

core/node/node_framework/src/service/tests.rs
```diff
@@ -0,0 +1,232 @@
...
+    async fn run(self: Box<Self>, _stop_receiver: StopReceiver) -> anyhow::Result<()> {
+        let mut guard = self.0.lock().unwrap();
+        *guard = true;
+        Ok(())
+    }
+}
+
+// Layer for task that updates resource and returns only after stop signal received
+#[derive(Debug)]
+struct RemainingTaskLayer(Arc<Mutex<bool>>);
```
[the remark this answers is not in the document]
AnastasiiaVashchuk: Not sure I understand

core/node/node_framework/src/service/tests.rs
AnastasiiaVashchuk: imho `run_flag_resource` here is also good.

core/node/node_framework/src/service/tests.rs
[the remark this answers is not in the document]
popzxc: You have `SuccessfulTaskLayer` and `RemainingTaskLayer`, but this seems to be excessive. You can add two tasks from a single layer.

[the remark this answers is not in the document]
AnastasiiaVashchuk: If I add 2 tasks, `RemainingTask` and `SuccessfulTask`, into one layer, I cannot check that the flag has changed. 
I mean...it will not be possible to check a) the flag changed successfully  to `true` after `SuccessfulTask` finish.
Because  `RemainingTask` will be run after stop signal and again flag value will be false.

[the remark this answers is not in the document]
popzxc: You can have two fields in your layer (e.g. `Arc<Mutex<bool>>` for the first task and `Arc<Mutex<bool>>` for the second task).

core/node/node_framework/src/service/tests.rs
popzxc: But it's a different resource, no? It's for a flag that checks if the stop signal has been sent.

core/node/node_framework/src/service/tests.rs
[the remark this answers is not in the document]
AnastasiiaVashchuk: you are right..

popzxc (approved):

popzxc:
@AnastasiiaVashchuk  Ah, spellcheck is still failing. Please fix that. As I mentioned before, using markdown quoting (e.g. \`like_this\` for function names and such would make the spellchecker happy. E.g. you have stuff like "ZkStack Service" which linter isn't happy about, but it'd be fine with "\`ZkStackService\`" (note \` symbols).

popzxc (approved):

PR #1103, opened 2024-02-16
feat(api): Create RPC method to return all tokens

slowli:
## What ❔

- Creates `en_syncTokens` RPC method that returns all tokens optionally filtered by the miniblock at which the token was deployed.
- Uses this method during snapshot recovery to recover tokens.

## Why ❔

It's necessary to recover tokens in order for the VM sandbox to work properly.

## Checklist


core/lib/dal/src/storage_logs_dal.rs
```diff
@@ -300,6 +300,51 @@ impl StorageLogsDal<'_, '_> {
...
+            .map(|address| {
+                let hashed_key = get_code_key(&address).hashed_key().0;
+                (hashed_key, (hashed_key, address))
+            })
+            .unzip();
+        let max_miniblock_number = at_miniblock.map_or(u32::MAX, |number| number.0);
+        let rows = sqlx::query!(
+            r#"
+            SELECT DISTINCT
+                ON (hashed_key) hashed_key,
```
slowli, line 320: I think `SELECT DISTINCT ON` can be used in other queries as well; I plan to work on using this approach separately. cc @Deniallugo; we've recently discussed a query that could be seemingly made more efficient using this approach.

core/lib/dal/src/tokens_web3_dal.rs
```diff
@@ -31,17 +56,45 @@ impl TokensWeb3Dal<'_, '_> {
...
-                    symbol: record.symbol,
-                    decimals: record.decimals as u8,
-                },
-            })
-            .collect())
+        Ok(records.into_iter().map(Into::into).collect())
+    }
+
+    /// Returns information about all tokens.
+    pub async fn get_all_tokens(
```
slowli, line 63: - Another way to implement this method would be to add `tokens.miniblock_number` column. AFAIU, it'd require a programmatic migration to fill data because there's no way to compute a hashed key from address + key inside an SQL query.
- Or we can not filter tokens on the server side at all, and use client-side filtering.

core/lib/snapshots_applier/src/lib.rs
```diff
@@ -518,4 +530,70 @@ impl<'a> SnapshotsApplier<'a> {
...
+        let bogus_tokens = tokens.iter().filter(|token| {
+            // We need special handling for L2 ether; its `l2_address` doesn't have a deployed contract
+            !token.l2_address.is_zero() && !filtered_addresses.contains_key(&token.l2_address)
+        });
+        let bogus_tokens: Vec<_> = bogus_tokens.collect();
+        if !bogus_tokens.is_empty() {
+            let err = anyhow::anyhow!(
+                "Main node returned bogus tokens that are not deployed on L2: {bogus_tokens:?}"
+            );
+            return Err(SnapshotsApplierError::Retryable(err));
```
slowli, line 585: Not sure whether an error here should be retryable. My reasoning for making it such was that the main node may have an issue that could be resolved later. We still trust the main node not to omit any tokens from the response.

core/tests/snapshot-recovery-test/tests/snapshot-recovery.test.ts
```diff
@@ -34,6 +35,12 @@ interface StorageLog {
     readonly enumerationIndex: number;
 }
 
+interface TokenInfo {
+    // FIXME: why isn't case changed?
```
slowli, line 39: Unlike other DTOs return by Web3 RPCs, `TokenInfo` doesn't use `#[serde(rename_all = "camelCase")]`. Is it intentional? Since `TokenInfo` is also returned by `get_confirmed_tokens`, changing casing wouldn't be quite easy (unless `get_confirmed_tokens` is removed 🙃).

core/lib/web3_decl/src/namespaces/en.rs
```diff
@@ -20,4 +20,11 @@ pub trait EnNamespace {
         block_number: MiniblockNumber,
         include_transactions: bool,
     ) -> RpcResult<Option<SyncBlock>>;
+
+    /// Lists all tokens created at or before the specified `block_number`.
+    ///
+    /// This method is used by EN after snapshot recovery in order to recover token records.
+    #[method(name = "syncTokens")]
+    async fn sync_tokens(&self, block_number: Option<MiniblockNumber>)
+        -> RpcResult<Vec<TokenInfo>>;
```
slowli, line 29: As an alternative, this method could be placed in the `zks` namespace as `getAllTokens`. It depends on whether we want to make this method (eventually) public / usable for use cases beyond snapshot recovery.

core/lib/dal/src/storage_logs_dal.rs
```diff
@@ -300,6 +300,51 @@ impl StorageLogsDal<'_, '_> {
...
+                miniblock_number
+            FROM
+                storage_logs
+            WHERE
+                hashed_key = ANY ($1)
+                AND miniblock_number <= $2
+                AND value != $3
```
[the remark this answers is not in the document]
slowli: Good catch! I've slightly amended the query to return latest `value` so that zero values can be filtered out on the DAL level (unless I'm missing something, it seems impossible to do on the DB query level, like via `HAVING` clause).

perekopskiy (approved):

core/lib/dal/src/storage_logs_dal.rs
RomanBrodetski: yes and thanks for creating the [ticket](https://linear.app/matterlabs/issue/PLA-803/use-select-distinct-on-queries)

core/lib/dal/src/tokens_dal.rs
```diff
@@ -79,34 +76,25 @@ impl TokensDal<'_, '_> {
...
+        let all_token_addresses = self.get_all_l2_token_addresses().await?;
+        let token_deployment_data = self
+            .storage
+            .storage_logs_dal()
+            .filter_deployed_contracts(all_token_addresses.into_iter(), None)
+            .await?;
+        let token_addresses_to_be_removed: Vec<_> = token_deployment_data
+            .into_iter()
+            .filter_map(|(address, deployed_at)| (deployed_at > block_number).then_some(address.0))
+            .collect();
```
RomanBrodetski, line 90: Wow I think it fixes https://linear.app/matterlabs/issue/PLA-726/block-reversion-tool-tokens-are-not-rolled-back-properly ?? That's great!

RomanBrodetski (approved):
Nice!

PR #1109, opened 2024-02-17
docs: fix typo in introduction

hawks920:
The typo in the introduction is in the sentence:

"Value needs to be secured, and only blockchains are able provide the level of security that the internet needs."

The correction would be to add "to" between "are" and "able":

"Value needs to be secured, and only blockchains are able to provide the level of security that the internet needs."

## Checklist


jespercv:
good.

PR #1138, opened 2024-02-19
feat(commitment-generator): `events_queue` shadow mode

perekopskiy:
## What ❔

Commitment generator calculates `events_queue` based on VM events and checks that there is the same queue in `events_queue` DB table.

## Why ❔

Preparation for getting rid of events' data duplication (both in VM return type and DB schema).

## Checklist


core/lib/multivm/src/versions/vm_boojum_integration/vm.rs
```diff
@@ -86,7 +86,7 @@ impl<S: WriteStorage, H: HistoryMode> VmInterface<S, H> for Vm<S, H> {
     /// This method should be used only after the batch execution.
     /// Otherwise it can panic.
     fn get_current_execution_state(&self) -> CurrentExecutionState {
-        let (deduplicated_events_logs, raw_events, l1_messages) = self.state.event_sink.flatten();
```
Deniallugo, line 89: I'm not sure that you have to change old vms behaviour

core/lib/types/src/event.rs
```diff
@@ -339,6 +344,73 @@ pub struct VmEventGroupKey {
...
+                .indexed_topics
+                .iter()
+                .map(|h| h256_to_u256(*h))
+                .chain(event.value.chunks(32).map(U256::from_big_endian));
+
+            // And now we process these values in chunks by two.
+            let value_chunks = values.chunks(2);
+            let other_logs = value_chunks.into_iter().map(|mut chunk| {
+                // The first value goes to `log_query.key`.
+                let key = chunk.next().unwrap();
```
Deniallugo, line 387: why is it safe to unwrap?

core/lib/types/src/event.rs
```diff
@@ -339,6 +344,73 @@ pub struct VmEventGroupKey {
     pub topic: (u32, H256),
 }
 
+/// Each `VmEvent` can be translated to several log queries.
+/// This methods converts each event from input to log queries and returns all produced log queries.
+pub fn convert_vm_events_to_log_queries(events: &[VmEvent]) -> Vec<LogQuery> {
```
Deniallugo, line 349: could you add some tests for it? looks a bit non trivial

Deniallugo (commented):
Looks good! 
but a bit of a dangerous, could you probably run it in a shadow more for some time and remove the table after?

core/lib/multivm/src/versions/vm_boojum_integration/vm.rs
perekopskiy: Why not? The behavior is the same as for `vm_latest`, if `deduplicated_events_logs` is not needed for commitment generator then we can just remove it

core/lib/types/src/event.rs
perekopskiy: That's how chunking works, there are some chunks with len=2 and optionally the last one with len=1. For both cases chunk.count() >= 1 is held

core/lib/multivm/src/versions/vm_boojum_integration/vm.rs
Deniallugo: just for better compatibility. 
i'd prefer to freeze the code there as much as possible

perekopskiy: This means we should keep `deduplicated_events_logs` in response type forever (or at least with the current multivm approach). In general I don't mind but would prefer to remove code that is not needed

Deniallugo: i think, you can remove it from interface, but just don't remove the implementation

core/lib/zksync_core/src/commitment_generator/mod.rs
```diff
@@ -46,11 +47,28 @@ impl CommitmentGenerator {
...
+        let events_queue_calculated = {
+            let events = connection
+                .events_dal()
+                .get_vm_events_for_l1_batch(l1_batch_number)
+                .await?
+                .context("Missing events for batch")?;
+            convert_vm_events_to_log_queries(&events)
+        };
+
+        if events_queue_from_db != events_queue_calculated {
```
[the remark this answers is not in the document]
perekopskiy: VM also sets `timestamp: 0` for logs in `deduplicated_events_logs`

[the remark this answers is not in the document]
perekopskiy: https://github.com/matter-labs/zksync-era/blob/perekopskiy-pla-649-deduplicating-events-data-in-the-currentexecutionstate/core/lib/multivm/src/versions/vm_latest/old_vm/event_sink.rs#L113-L127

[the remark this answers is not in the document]
StanislavBreadless: @perekopskiy could you please add a comment on why it is safe then? (in code)

core/lib/types/src/event.rs
```diff
@@ -339,6 +344,73 @@ pub struct VmEventGroupKey {
...
+            };
+
+            // The next logs hold an information about remaining topics and `event.value`.
+            // Each log can hold at most two values each of 32 bytes.
+            // The following piece of code prepares these 32-byte values.
+            let values = event
+                .indexed_topics
+                .iter()
+                .map(|h| h256_to_u256(*h))
+                .chain(event.value.chunks(32).map(U256::from_big_endian));
```
StanislavBreadless, line 381: I am not 100% sure `U256::from_big_endian` will work fine with `value` that is not divisible by 32

perekopskiy: Is it possible that value is not divisible by 32? I couldn't find an example that gives `value.len() % 32 != 0`

StanislavBreadless: Maybe noone does that because a typical Solidity contract uses ABI encoded values which pad bytes with 0s to ensure that they are divisible by 32. However, there is nothing that in principle fobids value from being 1 byte long

core/lib/types/src/event.rs
```diff
@@ -339,23 +344,84 @@ pub struct VmEventGroupKey {
...
+                address: EVENT_WRITER_ADDRESS,
+                key,
+                read_value: U256::zero(),
+                written_value: address_to_u256(&event.address),
+                rw_flag: false,
+                rollback: false,
+                is_service: true,
+            };
+
+            // The next logs hold an information about remaining topics and `event.value`.
```
StanislavBreadless, line 376: nit typo

```suggestion
            // The next logs hold information about remaining topics and `event.value`.
```

core/lib/types/src/event.rs
```diff
@@ -502,4 +568,105 @@ mod tests {
...
...ad_value": "0x0", "written_value": "0x8008", "tx_number_in_block": 0}, {"key": "0x27fe8c0b49f49507b9d4fe5968c9f49edfe5c9df277d433a07a0717ede97638d", "address": "0x000000000000000000000000000000000000800d", "rw_flag": false, "aux_byte": 0, "rollback": false, "shard_id": 0, "timestamp": 0, "is_service": false, "read_value": "0x0", "written_value": "0x0", "tx_number_in_block": 0}, {"key": "0x1", "address": "0x000000000000000000000000000000000000800d", "rw_flag": false, "aux_byte": 0, "rollback": false, "shard_id": 0, "timestamp": 0, "is_service": false, "read_value": "0x0", "written_value": "0x0", "tx_number_in_block": 0}, {"key": "0x8001", "address": "0x000000000000000000000000000000000000800d", "rw_flag": false, "aux_byte": 0, "rollback": false, "shard_id": 0, "timestamp": 0, "is_service": false, "read_value": "0x0", "written_value": "0x835d4d504d1bf838056aeca450a2768b400a55ff247b31b29e98cc392bae7119", "tx_number_in_block": 0}, {"key": "0x1", "address": "0x000000000000000000000000000000000000800d", "rw_flag": false, "aux_byte": 0, "rollback": false, "shard_id": 0, "timestamp": 0, "is_service": false, "read_value": "0x0", "written_value": "0x0", "tx_number_in_block": 0}]"#).unwrap();
```
StanislavBreadless, line 581: would it be possible to format that list? it is a bit hard to read this way

StanislavBreadless (approved):

StanislavBreadless (approved):

PR #1144, opened 2024-02-20
feat: Metrics for block tip

Artemka374:
## What ❔

Add metrics for block tip

## Checklist


core/lib/zksync_core/src/state_keeper/metrics.rs
```diff
@@ -395,3 +397,41 @@ pub(super) struct ExecutorMetrics {
...
+            .observe(execution_result.statistics.gas_used as usize);
+        self.pubdata_published
+            .observe(execution_result.statistics.pubdata_published as usize);
+        self.circuit_statistic
+            .observe(execution_result.statistics.circuit_statistic.total());
+        self.execution_metrics_size
+            .observe(execution_result.get_execution_metrics(None).size());
+
+        let logs_iter = execution_result.logs.storage_logs.iter();
+        let writes_metrics = StorageWritesDeduplicator::apply_on_empty_state(logs_iter);
```
Deniallugo, line 428: it's relatively hard operation, let's not deduplicate

Artemka374: Please check again, I also didn’t compressed values size

core/lib/zksync_core/src/state_keeper/metrics.rs
```diff
@@ -395,3 +399,63 @@ pub(super) struct ExecutorMetrics {
...
+            .iter()
+            .filter(|log| log.log_type == StorageLogQueryType::RepeatedWrite)
+            .count();
+        let total_updated_values_size = execution_result.logs.storage_logs.len() * (32 + 1); // 32 bytes per value + 1 metadata byte
+
+        let writes_metrics = DeduplicatedWritesMetrics {
+            initial_storage_writes: initial_writes,
+            repeated_storage_writes: repeated_writes,
+            total_updated_values_size,
+        };
```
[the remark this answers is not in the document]
perekopskiy: I think in the current form it doesn't make much sense, without taking into deduplication and compression values we get will be very inaccurate. What I would do instead is to measure the difference between [these](https://github.com/matter-labs/zksync-era/blob/main/core/lib/zksync_core/src/state_keeper/keeper.rs#L694)
metrics and write metrics before batch tip was applied, i.e. code will look like
```
let block_writes_metrics = updates_manager
                    .storage_writes_deduplicator
                    .apply_and_rollback(logs_to_apply_iter.clone());
let diff = block_writes_metrics - updates_manager.storage_writes_deduplicator.metrics();
```
There is a caveat, I suspect diff would be negative, so just using "-" might not work, instead you can cast every metric field to signed type and subtract then. Diff may be negative because in batch tip some storage slots are reset to zero, so writes to them will be deduplicated

core/lib/zksync_core/src/state_keeper/metrics.rs
```diff
@@ -395,3 +399,63 @@ pub(super) struct ExecutorMetrics {
 
 #[vise::register]
 pub(super) static EXECUTOR_METRICS: vise::Global<ExecutorMetrics> = vise::Global::new();
+
+#[derive(Debug, Metrics)]
+#[metrics(prefix = "block_tip")]
+pub(crate) struct BlockTipMetrics {
+    #[metrics(buckets = Buckets::exponential(60000.0..=80000000.0, 2.0))]
+    gas_used: Histogram<usize>,
```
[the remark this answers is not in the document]
Artemka374: That’s true, but we were discussing similar thing about VM Call Tracer metrics - and it’s really hard to get definitive values with Prometheus metrics. Gauges will throw us some random values that were observed in 30 seconds, so it’s not the best way to find the maximum. With histograms we will be able to detect to which range the maximum belongs, and if the range will be too big - we always can simply adjust the buckets

perekopskiy (commented):
General comment: AFAIU we will run this for some time, this will allow us to estimate how much resources batch tip consumes. But how will we reestimate them in future if bootloader/system contract change? Perhaps it's better to have it as a VM unit test, like [here](https://github.com/matter-labs/zksync-era/blob/main/core/lib/multivm/src/versions/vm_latest/tests/block_tip.rs#L203)?

Artemka374:
That’s true, for now I want to measure just raw bounds, hardcode them to https://github.com/matter-labs/zksync-era/pull/1076 as constants and most likely will add test in the PR as well

core/lib/zksync_core/src/state_keeper/metrics.rs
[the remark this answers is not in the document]
Artemka374: How big the difference between values could be? Can it be that big, that if we use raw approach it could break something? Or we can just say that it has some overhead(in case metrics won’t report the biggest values)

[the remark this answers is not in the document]
perekopskiy: Your current approach measures an upper bound for state_diffs size published. But this upper bound is very inaccurate. For example, let's simplify batch tip execution and assume the only action it does is the following: there is a storage slot which is always written when VM executes any transaction, and at the end of the batch at "batch tip" this slot value is set back to 0. Your approach will return that during batch tip there was 1 initial storage write, 0 repeated storage writes, total_updated_values_size = 33. But in fact state diffs size didn't increase by `(length of 1 initial write + total_updated_values_size)`, instead it decreased by that value.
Note that an example with storage slot set to 0 is not far from reality, in the actual batch tip execution there are several such slots that are rolled back.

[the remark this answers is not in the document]
Artemka374: Okay, I got your point, I don’t think that negative diff really matters to us since we only need to know upper bound, so we probably can just set 0 values where it really doesn’t matter for us

core/lib/zksync_core/src/state_keeper/metrics.rs
```diff
@@ -395,3 +397,53 @@ pub(super) struct ExecutorMetrics {
...
+            .observe(execution_result.get_execution_metrics(None).size());
+    }
+
+    pub fn observe_writes_metrics(
+        &self,
+        initial_writes_metrics: &DeduplicatedWritesMetrics,
+        applied_writes_metrics: &DeduplicatedWritesMetrics,
+    ) {
+        let size_diff = applied_writes_metrics.size(ProtocolVersionId::latest()) as i128
+            - initial_writes_metrics.size(ProtocolVersionId::latest()) as i128;
```
Deniallugo, line 436: The latest is wrong. you have to use proper version

Deniallugo: And could you describe what is going on here..
i'm really confused

Artemka374: Probably I should’ve documented it better. These are the difference between writes metrics before applying and deduplicating them to calculate overhead of batch tip writes metrics(see the comment of @perekopskiy above). Since in seal criteria we are using only size of them - I have shortened it to simple substraction of sizes, and to check a suggestion about negative diffs we have 2 different histograms - for negative diffs and for positive

perekopskiy: still using latest version is wrong, instead pass one from `updates_manager` (this is the version used for the batch)

perekopskiy (commented):
Looks good! Please fix protocol version used and I will approve

core/lib/zksync_core/src/state_keeper/metrics.rs
Artemka374: Updated

perekopskiy (approved):

PR #1145, opened 2024-02-20
feat(node_framework): Support Eth Watch in the framework

AnastasiiaVashchuk:
## What ❔
- Small change in the  `EthWatch` component. Now  `Eth Watch` 's `pool` field is used instead of  `ConnectionPool`  passed in the Eth Watch's `run()` method.
- Adds `Eth Watch` implementation, i. e. `EthWatchLayer` and `EthWatchTask`, in the same way it was done for `StateKeeper ` [here](https://github.com/matter-labs/zksync-era/pull/1043/files#diff-8ff8babf7b83c79dbf96f4998cf71d888beaeb265c7ce33192b0bb0c808f662b).

Current "external" point of view for the framework(the only thing changed is that we add ` .add_eth_watch_layer()?` after `.add_state_keeper_layer()?`):

```rust
        .with_log_format(log_format)
        .build();

    MainNodeBuilder::new()
        .add_pools_layer()?
        .add_fee_input_layer()?
        .add_object_store_layer()?
        .add_state_keeper_layer()?
        .add_eth_watch_layer()?
        .add_healthcheck_layer()?
        .build()
        .run()?;

    Ok(())
}
```


core/lib/zksync_core/src/eth_watch/mod.rs
```diff
@@ -131,12 +134,9 @@ impl EthWatch {
...
     }
 
-    pub async fn run(
-        &mut self,
-        pool: ConnectionPool,
-        stop_receiver: watch::Receiver<bool>,
-    ) -> anyhow::Result<()> {
+    pub async fn run(&mut self, stop_receiver: watch::Receiver<bool>) -> anyhow::Result<()> {
         let mut timer = tokio::time::interval(self.poll_interval);
+        let pool = self.pool.clone();
```
AnastasiiaVashchuk, line 139: This is done to avoid swearing from borrow checker

core/lib/zksync_core/src/eth_watch/mod.rs
```diff
@@ -55,6 +56,7 @@ impl EthWatch {
         pool: &ConnectionPool,
         poll_interval: Duration,
     ) -> Self {
+        let connection_pool = pool.clone();
```
popzxc, line 59: Given that the very first thing you do is clone the pool, probably it makes sense to take an owned object instead of a reference in this method. This way you won't need to clone it inside.

core/lib/zksync_core/src/eth_watch/mod.rs
```diff
@@ -212,7 +212,7 @@ pub async fn start_eth_watch(
     )
     .await;
 
-    Ok(tokio::spawn(async move {
-        eth_watch.run(pool, stop_receiver).await
-    }))
+    Ok(tokio::spawn(
+        async move { eth_watch.run(stop_receiver).await },
```
popzxc, line 216: Btw you probably don't need `async move { .. }` block here.

popzxc (commented):
Looks pretty good!

core/lib/zksync_core/src/eth_watch/mod.rs
AnastasiiaVashchuk: why

core/lib/zksync_core/src/eth_watch/mod.rs
popzxc: Add `drop(storage);` after `Self::initialize_state(..)`.

core/lib/zksync_core/src/eth_watch/mod.rs
AnastasiiaVashchuk: but `move` is still needed

popzxc: Just change `run` method to take `mut self` instead of `&mut self`.

popzxc (approved):

PR #1166, opened 2024-02-21
feat(healthcheck): Various healthcheck improvements

slowli:
## What ❔

- Adds `HeathStatus::ShuttingDown` set immediately after a component receives a termination signal. Makes the `/health` endpoint conforming to K8s readiness probe expectations.
- Makes slow / hard time limits for health checks configurable and decreases their values by default.
- Adds metric for slow, timed out and dropped health checks.

## Why ❔

Improves healthcheck observability.

## Checklist

### DIFF
--- a/core/lib/health_check/README.md
+++ b/core/lib/health_check/README.md
@@ -1,0 +1,121 @@
+# Health Monitoring
+
+Healthcheck infrastructure for node components allowing components to signal their current health state.
+Health states for all components run by the node are aggregated and are exposed as an HTTP `GET /health` endpoint
+bound to a dedicated healthcheck port, both for the main node and external node. This endpoint can be used
+as a readiness probe for Kubernetes, or used in other automations.
+
+## Main concepts
+
+**Component** is a logically isolated part of a node that affects the ability of the node to handle requests (aka node health).
+Components are supposed to run indefinitely until the node receives a stop signal.
+
+- Internal components correspond to one or more Tokio tasks. Examples of internal components are: JSON-RPC API server, Merkle tree,
+  consistency checker, reorg detector.
+- External components correspond to another process that the node communicates with. Examples of external components are:
+  Postgres connection pool, main node JSON-RPC (for the external node).
+
+Each component can report its health, which consists of 2 parts:
+
+- **Status**, e.g., "not ready", "ready", "shut down", "panicked"; see the crate code for a full list.
+- **Details**, a JSON value with the component-specific schema. E.g., Merkle tree reports its L1 batch "cursor" as a part of this information.
+
+Health from all components is aggregated into **application health**, which has its own status computed as the worst of
+component statuses. Application health is returned by the `/health` endpoint.
+
+## `/health` endpoint format
+
+`/health` will return current application health encoded as a JSON object. The HTTP status of the response is 20x
+if the application is healthy, and 50x if it is not.
+
+> **Warning.** The schema of data returned by the `/health` endpoint is not stable at this point and can change
+> without notice. Use at your own risk.
+
+<details>
+<summary>Example of endpoint output for an external node:</summary>
+
+```json
+{
+  "status": "ready",
+  "components": {
+    "sync_state": {
+      "status": "ready",
+      "details": {
+        "is_synced": true,
+        "local_block": 91,
+        "main_node_block": 91
+      }
+    },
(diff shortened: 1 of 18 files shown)


slowli:
Just in case: I've checked that if `/heath` is requested with a small client timeout (e.g., using `curl -m ..`) so that it doesn't complete in time, then `axum` drops the handling future together with pending futures it depends on (in particular, `CheckHealth::check_health()` implementations). So a drop guard added in this PR will actually be triggered in this case.

montekki (approved):

core/lib/health_check/README.md
```diff
@@ -0,0 +1,122 @@
...
+  are: Postgres connection pool, main node JSON-RPC (for the external node).
+
+Each component can report its health, which consists of 2 parts:
+
+- **Status**, e.g., "not ready", "ready", "shut down", "panicked"; see the crate code for a full list.
+- **Details**, a JSON value with the component-specific schema. E.g., Merkle tree reports its L1 batch "cursor" as a
+  part of this information.
+
+Health from all components is aggregated into **application health**, which has its own status computed as the worst of
+component statuses. Application health is returned by the `/health` endpoint.
```
RomanBrodetski, line 25: I think it's accurate now, but want to ping @popzxc to make sure we update it after the node framework is used by

core/lib/health_check/README.md
```diff
@@ -0,0 +1,122 @@
...
+      "status": "ready",
+      "details": {
+        "l1_batch_number": 25
+      }
+    },
+    "http_api": {
+      "status": "ready"
+    }
+  }
+}
```
RomanBrodetski, line 119: very nice!

core/lib/health_check/src/lib.rs
```diff
@@ -23,6 +30,9 @@ pub enum HealthStatus {
     Ready,
     /// Component is affected by some non-fatal issue. The component is still considered healthy.
     Affected,
```
RomanBrodetski, line 32: Affected? Very interesting. What's an example for this? Would we want to return this status when state keeper is working with Postgres storage? cc @itegulov

core/lib/health_check/src/metrics.rs
```diff
@@ -0,0 +1,43 @@
...
+struct AbnormalCheckLabels {
+    component: &'static str,
+    result: CheckResult,
+}
+
+#[derive(Debug, Metrics)]
+#[metrics(prefix = "healthcheck")]
+pub(crate) struct HealthMetrics {
+    /// Latency for abnormal checks. Includes slow, dropped and timed out checks (distinguished by the "result" label);
+    /// skips normal checks.
```
RomanBrodetski, line 25: Hmm why do we want to skip normal checks? Because there will a lot of very fast data points? Outliers will still be visible on the heatmap..

RomanBrodetski (approved):

core/lib/health_check/README.md
popzxc: I think it'd be accurate even after we switch to the node framework. The server will still collect a set of health checks, but it'd be up to the concrete layers whether they add their health checks to the set or not.

RomanBrodetski: that's true, but I'm more referring to the terminology. If you are fine with it, just resolve this

core/lib/health_check/src/metrics.rs
slowli: Yes, the issue is with the number of data points. IIUC, this is why we have time series for the corresponding built-in K8s metrics disabled (e.g., no `prober_probe_duration_seconds_bucket` time series, while `prober_probe_duration_seconds_sum` and `*_count` time series are present).

core/lib/health_check/src/lib.rs
slowli: Two uses of this status rn are:

- If the EN is out of sync with the main node
- If commitment mismatch was detected by the consistency checker

Just in case: this status doesn't lead to an erroneous `/health` response status.

core/lib/health_check/src/metrics.rs
RomanBrodetski: this is not directly related to this PR, so can be tackled separately. The PR I'd like to merge ASAP

popzxc (approved):

PR #1190, opened 2024-02-21
feat(dal): add query for pubdata to blobs

koloz193:
## What ❔

Add a DAL function to pull the pubdata for a given batch and return it in the form of blobs.

## Why ❔

We need this information as part of witness generation

## Checklist


core/lib/dal/src/blocks_dal.rs
```diff
@@ -2319,6 +2320,30 @@ impl BlocksDal<'_, '_> {
             .context("delete_initial_writes_inner()")?;
         Ok(())
     }
+
+    /// Retrieves the blobs for a given batch
```
mm-zk, line 2324: please clearly mention what happens when a given batch didn't have blobs (for example for older protocol versions)

PR #1203, opened 2024-02-22
feat(api): Monitor server RPC errors

slowli:
## What ❔

- Introduces RPC-level metadata middleware.
- Uses this middleware to monitor server RPC errors and report stats as metrics and logs.

## Why ❔

Increased API server observability.

## Checklist

- [x] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [x] Tests for the changes have been added / updated.
- [x] Documentation comments have been added / updated.
- [x] Code has been formatted via `zk fmt` and `zk lint`.
- [x] Spellcheck has been run via `zk spellcheck`.
- [x] Linkcheck has been run via `zk linkcheck`.

PR #1213, opened 2024-02-22
block revert support for consensus component

pompon0:
Reverting blocks in the main node translates to a hard fork of the consensus blockchain.
This means that every time block reverter is executed, the state of the consensus has to be reset (aka forked).
To achieve that a `Genesis` of the consensus is introduced. Among other things, it specifies the number of the fork (incremented at every forking) and the miniblock number that the fork should start from (set to the last sealed miniblock at the time of forking). External node learns the current Genesis from the main node via jsonrpc (eventually Genesis should be moved to L1 contract, so that hard forks are managed by governance).

Consensus doesn't support yet synchronization of blocks before the Genesis first block, therefore for now external node is using jsonrpc for syncing the blocks before genesis first block.


core/lib/dal/src/blocks_dal.rs
```diff
@@ -2196,7 +2196,7 @@ impl BlocksDal<'_, '_> {
                 )
             WHERE
                 l1_batch_number IS NULL
-                AND fee_account_address = '\x0000000000000000000000000000000000000000'::bytea
+                AND fee_account_address = 'x0000000000000000000000000000000000000000'::bytea
```
popzxc, line 2199: IIRC I've seen this before and it caused some troubles. I think it's a brainfart of query linter and probably should be rolled back.

core/lib/dal/src/consensus_dal.rs
```diff
@@ -30,54 +30,119 @@ impl ConsensusDal<'_, '_> {
...
+    /// Noop if the new genesis is the same as the current one.
+    /// Resets the stored consensus state otherwise and purges all certificates.
+    pub async fn try_update_genesis(&mut self, genesis: &validator::Genesis) -> anyhow::Result<()> {
+        let mut txn = self.storage.start_transaction().await?;
+        if let Some(got) = txn.consensus_dal().genesis().await? {
+            // Exit if the genesis didn't change.
+            if &got == genesis {
+                return Ok(());
+            }
+            if got.fork.number >= genesis.fork.number {
```
popzxc, line 50: Would it make sense to enforce that new fork number is old fork number + 1?

core/lib/dal/src/consensus_dal.rs
```diff
@@ -30,54 +30,119 @@ impl ConsensusDal<'_, '_> {
...
         Ok(())
     }
 
-    /// Fetches the first consensus certificate.
-    /// Note that we didn't backfill the certificates for the past miniblocks
-    /// when enabling consensus certificate generation, so it might NOT be the certificate
-    /// for the genesis miniblock.
-    pub async fn first_certificate(&mut self) -> anyhow::Result<Option<validator::CommitQC>> {
-        let Some(row) = sqlx::query!(
+    pub async fn fork(&mut self) -> anyhow::Result<()> {
```
popzxc, line 91: A doc comment on this method would be nice, as I'd say it's not straightforward.

Separately I'll mention that it differs from how we normally use DAL: usually it's a pretty stupid wrapper around SQL that exists to prevent spreading `sqlx` all over the place, and the actual logic lives elsewhere. I don't really have an opinion on whether it's good or bad, but it also adds to the fact that the documentation is needed here.

core/lib/zksync_core/src/consensus/mod.rs
```diff
@@ -61,61 +58,166 @@ impl MainNodeConfig {
...
+                    sync::wait_for(ctx, sub, |state| state.main_node_block() >= *n).await?;
+                    send.send(
+                        ctx,
+                        s.spawn(async {
+                            let n = n;
+                            loop {
+                                match ctx
+                                    .wait(
+                                        self.client
+                                            .fetch_l2_block(*n, /*`with_transaction=`*/ true),
```
popzxc, line 127: This indentation is a bit cursed ngl.

popzxc (commented):
Overall looks pretty good.

core/lib/dal/src/consensus_dal.rs
pompon0: well, technically 2 forks may happen between en executions, so not really

core/lib/zksync_core/src/consensus/mod.rs
pompon0: done

core/lib/dal/src/consensus_dal.rs
pompon0: added documentation

brunoffranca (dismissed):

core/lib/dal/src/consensus_dal.rs
```diff
@@ -30,54 +30,125 @@ impl ConsensusDal<'_, '_> {
...
+    pub async fn try_update_genesis(&mut self, genesis: &validator::Genesis) -> anyhow::Result<()> {
+        let mut txn = self.storage.start_transaction().await?;
+        if let Some(got) = txn.consensus_dal().genesis().await? {
+            // Exit if the genesis didn't change.
+            if &got == genesis {
+                return Ok(());
+            }
+            if got.fork.number >= genesis.fork.number {
+                anyhow::bail!("transition to a past fork is not allowed");
+            }
```
slowli, line 52: Nit: You can use `anyhow::ensure` for these checks. Also, it could make sense to include details (`fork.number`?) into the error message for debug purposes.

core/lib/zksync_core/src/block_reverter/mod.rs
```diff
@@ -96,13 +100,15 @@ pub struct BlockReverter {
 
 impl BlockReverter {
     pub fn new(
+        main_node: bool,
```
slowli, line 103: Bikeshedding: It could make sense to use an enum instead, like `NodeRole::{Main, External}` for increased readability.

core/lib/zksync_core/src/reorg_detector/mod.rs
```diff
@@ -21,8 +21,8 @@ use crate::{
 mod tests;
 
 #[derive(Debug, thiserror::Error)]
-enum HashMatchError {
-    #[error("RPC error calling main node")]
+pub enum Error {
+    #[error("RPC error calling main node: {0}")]
```
slowli, line 25: Nit: AFAIU, the wrapped error info is displayed automatically for `{:?}` and `{:#}` presentations; that's why a similar example in [`thiserror` crate docs](https://docs.rs/thiserror/latest/thiserror/#example) doesn't include `: {0}` for the `Disconnect` variant.

core/lib/zksync_core/src/reorg_detector/mod.rs
```diff
@@ -38,21 +38,31 @@ enum HashMatchError {
...
     )]
     EarliestL1BatchTruncated(L1BatchNumber),
-    #[error("Internal error")]
-    Internal(#[from] anyhow::Error),
+    #[error("reorg detected, restart the node to revert to the last correct L1 batch #{0}.")]
+    ReorgDetected(L1BatchNumber),
+    #[error("storage error: {0:#}")]
+    Storage(anyhow::Error),
+    #[error("no remote L1 batch")]
+    NoRemoteL1Batch,
```
slowli, line 46: Bikeshedding: While I appreciate that it allows simplifying code quite a bit, IMO, having this error variant exposed publicly makes it more difficult to reason about public `ReorgDetector` methods, because in some cases, this variant isn't really an error. It could make sense to have an internal error enum representing a sum of `Error` and `NoRemoteL1Batch` variants.

core/lib/zksync_core/src/reorg_detector/mod.rs
```diff
@@ -38,21 +38,31 @@ enum HashMatchError {
         Using an earlier snapshot could help."
     )]
     EarliestL1BatchTruncated(L1BatchNumber),
-    #[error("Internal error")]
-    Internal(#[from] anyhow::Error),
+    #[error("reorg detected, restart the node to revert to the last correct L1 batch #{0}.")]
+    ReorgDetected(L1BatchNumber),
+    #[error("storage error: {0:#}")]
+    Storage(anyhow::Error),
```
slowli, line 44: Dumb question: Are `Internal` variant and the corresponding `From<anyhow::Error>` conversion too generic / implicit for your taste?

core/lib/zksync_core/src/sync_layer/fetcher.rs
```diff
@@ -168,21 +168,23 @@ impl IoCursor {
...
 enum FetcherError {
-    #[error("JSON-RPC error communicating with main node")]
+    #[error("JSON-RPC error communicating with main node: {0}")]
     Web3(#[from] EnrichedClientError),
-    #[error("internal fetcher error")]
+    #[error("internal fetcher error: {0:#}")]
     Internal(#[from] anyhow::Error),
 }
 
+pub(crate) fn is_transient(err: &EnrichedClientError) -> bool {
```
slowli, line 177: Nit: I think at this point, it could make sense to move this logic to an `EnrichedClientError` method; it seems to be more-or-less general-purpose.

core/lib/zksync_core/src/consensus/mod.rs
```diff
@@ -61,61 +60,155 @@ impl MainNodeConfig {
...
+    }
+
+    /// Fetches (with retries) the given block from the main node.
+    async fn fetch_block(&self, ctx: &ctx::Ctx, n: MiniblockNumber) -> ctx::Result<FetchedBlock> {
+        const RETRY_INTERVAL: time::Duration = time::Duration::seconds(5);
+        loop {
+            let res = ctx.wait(self.client.fetch_l2_block(n, true)).await?;
+            match res {
+                Ok(Some(block)) => return Ok(block.try_into()?),
+                Ok(None) => {}
```
slowli, line 110: Just to clarify: Is not having a block intentionally considered an error for the purpose of retries?

core/lib/zksync_core/src/consensus/mod.rs
```diff
@@ -61,61 +60,155 @@ impl MainNodeConfig {
...
+                Err(err) if is_transient(&err) => {}
+                Err(err) => {
+                    return Err(anyhow::format_err!("client.fetch_l2_block({}): {err}", n).into());
```
slowli, line 113: Nit: `format_err!` macro is just an alias for `anyhow!`, so it usually makes sense to use that. Also, shouldn't  `err` here be convertible to `anyhow::Error`? If so, it would be more idiomatic to use something like

```rust
anyhow::Error::new(err).context(format!("client.fetch_l2_block({n})))
```

core/lib/zksync_core/src/consensus/mod.rs
```diff
@@ -61,61 +60,155 @@ impl MainNodeConfig {
...
     }
-}
 
-/// External node consensus config.
-#[derive(Debug, Clone)]
-pub struct FetcherConfig {
-    pub executor: executor::Config,
-}
+    /// Fetches blocks from the main node in range `[cursor.next()..end)`.
+    async fn fetch_blocks(
```
slowli, line 121: It looks weird that EN now has two JSON-RPC-based fetchers. Could [the existing fetcher](https://github.com/matter-labs/zksync-era/blob/10e3a3ec50c84b3533af111d532e0b528c0b4966/core/lib/zksync_core/src/sync_layer/fetcher.rs#L193) be adapted to be used here, or, if it's bad, could it be entirely replaced with this one?

core/lib/zksync_core/src/consensus/mod.rs
```diff
@@ -61,61 +60,155 @@ impl MainNodeConfig {
...
+                .await?;
+
+            // Monitor the genesis of the main node.
+            // If it changes, it means that a hard fork occurred and we need to reset the consensus state.
+            s.spawn_bg::<()>(async {
+                let want = genesis;
+                loop {
+                    if let Ok(got) = self.fetch_genesis(ctx).await {
+                        if got != want {
+                            return Err(anyhow::format_err!("genesis changed").into());
```
slowli, line 184: Nit: You can include some details in the error message for better debugging.

core/lib/zksync_core/src/consensus/mod.rs
```diff
@@ -61,61 +60,155 @@ impl MainNodeConfig {
...
+            // Fetch blocks before the genesis.
+            self.fetch_blocks(ctx, &mut cursor, genesis.fork.first_block)
+                .await?;
+
+            // Monitor the genesis of the main node.
+            // If it changes, it means that a hard fork occurred and we need to reset the consensus state.
+            s.spawn_bg::<()>(async {
+                let want = genesis;
+                loop {
+                    if let Ok(got) = self.fetch_genesis(ctx).await {
```
slowli, line 182: Checking genesis could be the scope of responsibility of the reorg detector, or maybe another small component ("consensus reorg detector"? "fork detector"?) that would work similarly to the reorg detector. I don't have a particularly strong opinion regarding this, so feel free to ignore.

core/lib/dal/src/consensus_dal.rs
pompon0: done

core/lib/zksync_core/src/reorg_detector/mod.rs
pompon0: done

core/lib/zksync_core/src/reorg_detector/mod.rs
pompon0: It is an error in a sense that comparison couldn't be performed. The real issue is that all the methods are using the same error type.

core/lib/zksync_core/src/reorg_detector/mod.rs
pompon0: I wanted to clarify that all the "internal" errors of these functions are actually storage errors.

core/lib/zksync_core/src/block_reverter/mod.rs
pompon0: done

core/lib/zksync_core/src/sync_layer/fetcher.rs
pompon0: done

core/lib/zksync_core/src/consensus/mod.rs
pompon0: It is definitely an error. The question is whether it should be considered transient or fatal. Main node is expected to have all the blocks before the genesis, so we can consider it fatal. It may happen that None will be received if the main node forks while we are syncing, but that will be detected by the genesis change detector, so here it is ok for it to be considered transient.

core/lib/zksync_core/src/consensus/mod.rs
pompon0: I like `format_err` better, as it better expresses what it does compared to `anyhow`.

pompon0: it is not like anyone is expected to rely on the error structure, so IMO it is not important to be idiomatic about it.

core/lib/zksync_core/src/consensus/mod.rs
pompon0: done

core/lib/zksync_core/src/consensus/mod.rs
pompon0: This "fork detector" is implemented as a subtask of `Fetcher::run()`. I don't see a reason to flatten the task structure here and expose "fork detector" to the outside logic. Fork detector needs to start only once the consensus genesis for this execution is fixed, so it is not like it can be independent from `Fetcher::run()` execution.

core/lib/zksync_core/src/reorg_detector/mod.rs
pompon0: reverted `Storage` to `Internal`

core/lib/zksync_core/src/consensus/mod.rs
popzxc: Huge +1 on `format_err` vs `anyhow`.

core/lib/zksync_core/src/consensus/mod.rs
pompon0: deduplicated

core/tests/revert-test/tests/revert-and-restart-en.test.ts
```diff
@@ -0,0 +1,253 @@
+import * as utils from 'zk/build/utils';
```
popzxc, line 1: This file looks pretty much identical to the `revert-and-restart.test.ts` (at least I tried to find the difference by eye, but failed). Do we really need to copy-pase it? Would it be worse to use an already present `revert-and-restart.test.ts` (and for whatever difference there change the behavior dynamically)?

core/bin/external_node/src/main.rs
```diff
@@ -488,16 +482,36 @@ async fn main() -> anyhow::Result<()> {
...
+        NodeRole::External,
+        config.required.state_cache_path.clone(),
+        config.required.merkle_tree_path.clone(),
+        None,
+        connection_pool.clone(),
+        L1ExecutedBatchesRevert::Allowed,
+    );
+
+    let mut reorg_detector = ReorgDetector::new(main_node_client.clone(), connection_pool.clone());
+    match reorg_detector.check_consistency().await {
```
popzxc, line 502: Probably it makes sense to add a comment here explaining that we're checking for the reorg in the beginning because we expect that if reorg is detected during the node lifecycle, the node will exit the same way as it does with any other critical error, and would restart. Then, on the 2nd launch reorg would be detected here, and processed, and the node will be able to operate normally.

popzxc (approved):
Nice! I especially liked how the new fetcher in the consensus module unifies old/new approach.

core/tests/revert-test/tests/revert-and-restart-en.test.ts
pompon0: also, it fails now

core/lib/dal/src/consensus_dal.rs
```diff
@@ -30,54 +30,129 @@ impl ConsensusDal<'_, '_> {
...
+    /// [Main node only] creates a new consensus fork starting at
+    /// the last sealed miniblock. Resets the state of the consensus
+    /// by calling `try_update_genesis()`.
+    pub async fn fork(&mut self) -> anyhow::Result<()> {
+        let mut txn = self.storage.start_transaction().await?;
+        let Some(old) = txn.consensus_dal().genesis().await? else {
+            return Ok(());
+        };
+        let last = txn.blocks_dal().get_sealed_miniblock_number().await?;
+        let first_block = validator::BlockNumber(last.map(|n| n.0.into()).unwrap_or(0));
```
slowli, line 107: - Why is having no miniblocks in the storage OK here? Why is the default value zero? What about snapshot recovery? (At least in the decentralized network, and maybe even before that, we'll enable snapshot recovery for main node as well.)
- Repeating the responsibility question from the line comment above: Why should it be responsibility of DAL (vs the calling app) to decide whether having no miniblocks in the storage is OK? If this would be the app's responsibility, we could reason that e.g. consensus components never run with empty storage (not saying that it necessarily makes sense, just that app-level reasoning wouldn't have to be lowered to the DAL level).

core/lib/dal/src/consensus_dal.rs
```diff
@@ -30,54 +30,129 @@ impl ConsensusDal<'_, '_> {
...
+                first_parent: None,
+            },
+        };
+        txn.consensus_dal().try_update_genesis(&new).await?;
+        txn.commit().await?;
+        Ok(())
+    }
+
+    /// Fetches the current BFT replica state.
+    pub async fn replica_state(&mut self) -> anyhow::Result<ReplicaState> {
```
slowli, line 123: Dumb question: Non-optionality of the returned state is caused by the changes in the consensus logic (e.g., replica state never being read before it's written), right? From the DAL perspective, I don't think it's reasonable to assume that the `consensus_replica_state` table is not empty. Sure, we won't panic if it is, but IMO, it'd be reasonable to return `Result<Option<_>>` and deal with `Ok(None)` on the app level.

core/lib/zksync_core/src/api_server/web3/namespaces/en.rs
```diff
@@ -15,12 +15,33 @@ impl EnNamespace {
         Self { state }
     }
 
+    pub async fn consensus_genesis_impl(&self) -> Result<Option<en::ConsensusGenesis>, Web3Error> {
+        async {
```
slowli, line 19: That's neat, but will be obsoleted by #1203, which removes `internal_error` altogether 🙃 Also, not very maintainable, because if you'd need to add a non-internal error response somewhere in the `async` block, it'd be silently converted to an internal error.

core/lib/zksync_core/src/api_server/web3/namespaces/en.rs
```diff
@@ -15,12 +15,33 @@ impl EnNamespace {
...
+                .await?
+                .consensus_dal()
+                .genesis()
+                .await?
+            else {
+                return Ok(None);
+            };
+            anyhow::Ok(Some(en::ConsensusGenesis(
+                zksync_protobuf::serde::serialize(&genesis, serde_json::value::Serializer).unwrap(),
+            )))
```
slowli, line 33: Nit: Would it make sense to add a constructor to `ConsensusGenesis` that encapsulates this conversion? It looks weirdly low-level.

core/lib/zksync_core/src/reorg_detector/mod.rs
```diff
@@ -231,122 +238,147 @@ impl ReorgDetector {
...
+            .pool
+            .access_storage()
+            .await
+            .context("access_storage()")?;
+        let first_l1_batch = storage
+            .blocks_dal()
+            .get_earliest_l1_batch_number_with_metadata()
+            .await
+            .context("get_earliest_l1_batch_number_with_metadata")?
+            .context("all L1 blocks dissapeared")?;
```
slowli, line 302: Nit: "all L1 batches"; "L1 blocks" can be interpreted as blocks of the L1 blockchain.

core/lib/zksync_core/src/reorg_detector/mod.rs
```diff
@@ -231,122 +238,147 @@ impl ReorgDetector {
...
+        drop(storage);
+        match self.root_hashes_match(first_l1_batch).await {
+            Ok(true) => {}
+            Ok(false) => return Err(Error::EarliestL1BatchMismatch(first_l1_batch)),
+            Err(HashMatchError::RemoteHashMissing) => {
+                return Err(Error::EarliestL1BatchTruncated(first_l1_batch))
```
slowli, line 308: Nit: missing semicolon after the statement.

core/lib/zksync_core/src/reorg_detector/mod.rs
```diff
@@ -361,127 +393,31 @@ impl ReorgDetector {
...
         self.event_handler.initialize();
-        loop {
-            match self.run_inner(&mut stop_receiver).await {
-                Ok(l1_batch_number) => return Ok(l1_batch_number),
-                Err(HashMatchError::Rpc(err)) if is_transient_err(&err) => {
-                    tracing::warn!("Following transport error occurred: {err}");
+        while !*stop_receiver.borrow() {
+            match self.check_consistency().await {
+                Err(err) if !err.is_transient() => return Err(err),
+                Err(err) => {
```
slowli, line 411: Bikeshedding: IMO, it'd be more straightforward to place this clause first with `if err.is_transient()`; in this way, the condition wouldn't have to be inverted.

core/lib/zksync_core/src/consensus/fetcher.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    /// Fetches blocks from the main node in range `[cursor.next()..end)`.
+    pub(super) async fn fetch_blocks(
+        &self,
+        ctx: &ctx::Ctx,
+        cursor: &mut storage::Cursor,
+        end: Option<validator::BlockNumber>,
+    ) -> ctx::Result<()> {
+        const MAX_CONCURRENT_REQUESTS: usize = 100;
+        let done = |next| matches!(end, Some(end) if next>=end);
```
slowli, line 175: Nit: There should be spaces around `>=`. Also, completely subjectively, not using a closure would read equally well: `while end.map_or(false, |end| next < end) { ... }`. Weird that rustfmt doesn't see it (most probably, it's the usual "formatting code in macros" problem).

core/lib/zksync_core/src/consensus/fetcher.rs
```diff
@@ -0,0 +1,197 @@
...
+        let mut next = cursor.next();
+        scope::run!(ctx, |ctx, s| async {
+            let (send, mut recv) = ctx::channel::bounded(MAX_CONCURRENT_REQUESTS);
+            s.spawn(async {
+                let send = send;
+                while !done(next) {
+                    let n = MiniblockNumber(next.0.try_into().unwrap());
+                    self.sync_state.wait_for_main_node_block(ctx, n).await?;
+                    send.send(ctx, s.spawn(self.fetch_block(ctx, n))).await?;
```
slowli, line 184: JFYI, @tomg10 has run into problems recently with rate limiting caused by executing many requests concurrently. The issues were caused by the `MainNodeFetcher`, but it looks like they could still be present in this implementation as well – AFAIU, it starts fetching first `MAX_CONCURRENT_REQUESTS` from the main node almost at the same time, which could trigger rate limiting.

slowli (commented):
Mostly looks good; agree with @popzxc that the unified JSON-RPC block fetcher looks neat 👍

core/lib/zksync_core/src/api_server/web3/namespaces/en.rs
pompon0: ConsensusGenesis exists only so that zksync_types doesn't depend on zksync_consensus_roles.

core/lib/zksync_core/src/consensus/fetcher.rs
pompon0: what rate limit should I add then?

core/bin/external_node/src/main.rs
```diff
@@ -169,51 +169,46 @@ async fn init_tasks(
...
+    let reorg_detector = ReorgDetector::new(main_node_client.clone(), connection_pool.clone());
+    app_health.insert_component(reorg_detector.health_check().clone());
+    task_handles.push(tokio::spawn({
+        let stop = stop_receiver.clone();
+        async move {
+            reorg_detector
+                .run(stop)
+                .await
+                .context("reorg_detector.run()")
+        }
```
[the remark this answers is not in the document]
pompon0: I disagree that it is an overkill, just standard way of annotating errors. Can be removed once error is appropriately annotated elsewhere.

core/lib/zksync_core/src/api_server/web3/namespaces/en.rs
pompon0: outdated

core/bin/external_node/src/main.rs
pompon0: done

core/lib/dal/src/consensus_dal.rs
pompon0: * here you go, disabled forking with no blocks
* as long as multiple independent components access the same storage through dal, it is responsibility of dal to enforce invariants. Otherwise one component has no guarantee that the other component didn't mess up the state. It is a really bad idea to scatter the responsibility for maintaining state invariants over a large codebase. Precedence that it is already the case doesn't decrease the value of the effort of moving invariants to the place where they can be actually enforced.

core/lib/zksync_core/src/reorg_detector/mod.rs
pompon0: done

core/lib/zksync_core/src/reorg_detector/mod.rs
pompon0: done

core/lib/zksync_core/src/consensus/fetcher.rs
pompon0: done

core/lib/zksync_core/src/consensus/fetcher.rs
pompon0: added rate limiter

core/lib/zksync_core/src/sync_layer/tests.rs
```diff
@@ -589,180 +582,3 @@ async fn external_io_with_multiple_l1_batches() {
     assert_eq!(fictive_miniblock.timestamp, 2);
     assert_eq!(fictive_miniblock.l2_tx_count, 0);
 }
-
-#[tokio::test]
-async fn fetcher_basics() {
-    let pool = ConnectionPool::test_pool().await;
```
slowli, line 595: Dumb question: Is it possible to adapt the removed tests for the new fetcher? IMO, they provide useful coverage for the fetcher logic.

core/tests/revert-test/tests/revert-and-restart-en.test.ts
```diff
@@ -0,0 +1,365 @@
...
+    }
+
+    // Sends SIGINT to the node process and waits for it to exit.
+    public async terminate(): Promise<void> {
+        this.proc.kill('SIGINT');
+        expect(await this.wait_for_exit()).to.equal(0);
+    }
+
+    // Waits for the node process to exit.
+    public async wait_for_exit(): Promise<number> {
```
slowli, line 195: Nit: Here and in other function / var names: TS generally uses camelCase for these.

core/tests/revert-test/tests/revert-and-restart-en.test.ts
```diff
@@ -0,0 +1,365 @@
...
+            }
+            await utils.sleep(1);
+        }
+        const alice2 = await alice.getBalance();
+        console.log('Terminate the main node');
+        await main_node.terminate();
+
+        console.log('Ask block_reverter to suggest to which L1 batch we should revert');
+        let revert_env = { ...process.env };
+        revert_env.RUST_LOG = 'off';
```
slowli, line 263: This variable looks unused.

infrastructure/zk/src/database.ts
```diff
@@ -140,25 +140,6 @@ export async function setupForDal(dalPath: DalPath, dbUrl: string) {
 }
 
 export async function setup(opts: DbOpts) {
-    if (process.env.TEMPLATE_DATABASE_URL) {
```
slowli, line 143: Dumb question: Is the `TEMPLATE_DATABASE_URL` functionality not useful? Personally, it does look somewhat unintuitive, but it could be useful in some cases. As a possible alternative, it could be possible to remove `template_database_url` from the EN config, but still retain the functionality here. If you do remove `template_database_url` from the EN config, please remove [its mention in the snapshot recovery test](https://github.com/matter-labs/zksync-era/blob/0adab9ea84412c62d46893cc5b66368b2294d760/core/tests/snapshot-recovery-test/tests/snapshot-recovery.test.ts#L192).

core/tests/revert-test/tests/revert-and-restart-en.test.ts
```diff
@@ -0,0 +1,365 @@
...
+        while (this.proc.exitCode == null) {
+            await utils.sleep(1);
+        }
+        return this.proc.exitCode;
+    }
+}
+
+describe('Block reverting test', function () {
+    compileBinaries();
+    let main_logs: fs.WriteStream = fs.createWriteStream(main_logs_path, { flags: 'a' });
```
slowli, line 205: Nit: In TS, `let` should be idiomatically used if the variable is reassigned (i.e., has shallow mutability); in all other cases, one should use `const`. Not a big deal, obviously.

core/bin/external_node/src/main.rs
```diff
@@ -168,51 +168,54 @@ async fn init_tasks(
...
+    task_handles.push(tokio::spawn({
+        let ctx = ctx::root();
         let mut stop_receiver = stop_receiver.clone();
-        let sync_state = sync_state.clone();
-        let main_node_client = main_node_client.clone();
-
-        tokio::spawn(async move {
-            scope::run!(&ctx::root(), |ctx, s| async {
+        let p2p_config = config.consensus.clone();
+        let fetcher = consensus::Fetcher {
```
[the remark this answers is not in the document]
pompon0: rate limit should be eventually provided via config

core/lib/zksync_core/src/consensus/fetcher.rs
```diff
@@ -0,0 +1,198 @@
...
+            Ok(()) | Err(ctx::Error::Canceled(_)) => Ok(()),
+            Err(ctx::Error::Internal(err)) => Err(err),
+        }
+    }
+
+    /// Periodically fetches the head of the main node
+    /// and updates `SyncState` accordingly.
+    async fn fetch_state_loop(&self, ctx: &ctx::Ctx) -> ctx::Result<()> {
+        const DELAY_INTERVAL: time::Duration = time::Duration::milliseconds(500);
+        const RETRY_INTERVAL: time::Duration = time::Duration::seconds(5);
```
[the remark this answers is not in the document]
pompon0: These constants are arbitrary and specific to this loop.

core/lib/zksync_core/src/consensus/fetcher.rs
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    /// Fetches blocks from the main node in range `[cursor.next()..end)`.
+    pub(super) async fn fetch_blocks(
+        &self,
+        ctx: &ctx::Ctx,
+        cursor: &mut storage::Cursor,
+        end: Option<validator::BlockNumber>,
+    ) -> ctx::Result<()> {
+        const MAX_CONCURRENT_REQUESTS: usize = 30;
```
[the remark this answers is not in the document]
pompon0: This is a separate parameter of the system.

core/lib/zksync_core/src/sync_layer/tests.rs
pompon0: You mean to check for the specific sequence of actions? I would prefer to make the conversion FetchedBlock -> actions an implementation detail of ExternalIO instead.

core/tests/revert-test/tests/revert-and-restart-en.test.ts
```diff
@@ -0,0 +1,365 @@
...
+    // if enable_consensus is set, consensus component will be started in the main node.
+    // if enable_execute is NOT set, main node will NOT send L1 transactions to execute L1 batches.
+    public static async spawn(
+        logs: fs.WriteStream,
+        enable_consensus: boolean,
+        enable_execute: boolean
+    ): Promise<MainNode> {
+        let env = fetchEnv(main_env);
+        env.ETH_SENDER_SENDER_AGGREGATED_BLOCK_EXECUTE_DEADLINE = enable_execute ? '1' : '10000';
+        // Set full mode for the Merkle tree as it is required to get blocks committed.
```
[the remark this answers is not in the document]
pompon0: that's not mine, I just copy-pasted this piece of config with the comment

core/tests/revert-test/tests/revert-and-restart-en.test.ts
```diff
@@ -0,0 +1,365 @@
...
+    // Spawns an external node.
+    // If enable_consensus is set, the node will use consensus P2P network to fetch blocks.
+    public static async spawn(logs: fs.WriteStream, enable_consensus: boolean): Promise<ExtNode> {
+        let env = fetchEnv(ext_env);
+        console.log(`DATABASE_URL = ${env.DATABASE_URL}`);
+        let args = [];
+        if (enable_consensus) {
+            args.push('--enable-consensus');
+        }
+        let proc = spawn('./target/release/zksync_external_node', args, {
```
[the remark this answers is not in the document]
pompon0: I want to control the variables used to spawn the process, without clumsy middleware.

infrastructure/zk/src/database.ts
pompon0: It definitely doesn't make sense to maintain it if restoring db from snapshot requires tweaking its content  - it is too fragile. We can reintroduce it if a need arises.

core/tests/revert-test/tests/revert-and-restart-en.test.ts
pompon0: done

core/bin/external_node/src/main.rs
```diff
@@ -555,6 +527,53 @@ async fn main() -> anyhow::Result<()> {
...
+            .context(
+                "Cannot roll back pending L1 batch since there are no L1 batches in Postgres",
+            )?;
+        drop(connection);
+
+        tracing::info!("Rolling back to l1 batch number {sealed_l1_batch_number}");
+        reverter
+            .rollback_db(sealed_l1_batch_number, BlockReverterFlags::all())
+            .await;
+        tracing::info!("Rollback successfully completed");
```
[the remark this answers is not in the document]
pompon0: The reason is to not change the number of restarts wrt the previous behavior. Ideally no restarts should be needed anyway, but I'm not changing that for now.

.github/workflows/ci-core-reusable.yml
```diff
@@ -306,19 +306,10 @@ jobs:
...
-          ci_run sleep 20
-          ci_run grep -q 'Rollback successfully completed' ext-node.log
-          # Restart the EN
-          ci_run zk server --components=$SERVER_COMPONENTS &>>server.log &
-          ci_run sleep 30
-          ZKSYNC_ENV=ext-node-docker ci_run zk external-node $EXT_NODE_FLAGS &>>ext-node.log &
+          ENABLE_CONSENSUS=${{ matrix.consensus }} PASSED_ENV_VARS=ENABLE_CONSENSUS ci_run zk test i revert-en
+          # test terminates the nodes, so we restart them.
+          ZKSYNC_ENV=docker ci_run zk server --components=$SERVER_COMPONENTS &>server.log &
+          ZKSYNC_ENV=ext-node-docker ci_run zk external-node $EXT_NODE_FLAGS &>ext-node.log &
```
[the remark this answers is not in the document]
pompon0: done

core/tests/revert-test/tests/revert-and-restart-en.test.ts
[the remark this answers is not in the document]
pompon0: we should have a canonical way of fetching env though

core/tests/revert-test/tests/revert-and-restart-en.test.ts
```diff
@@ -0,0 +1,365 @@
...
+        console.log('Terminate the main node');
+        await main_node.terminate();
+
+        console.log('Ask block_reverter to suggest to which L1 batch we should revert');
+        let revert_env = { ...process.env };
+        revert_env.RUST_LOG = 'off';
+        const values_json = runBlockReverter(['print-suggested-values', '--json']);
+        console.log(`values = ${values_json}`);
+        let values = parseSuggestedValues(values_json);
+        assert(last_executed == values.lastL1BatchNumber);
```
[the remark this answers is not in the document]
pompon0: I don't like "expect" because I can't use it consistently (for comparing BigNumbers for example)

core/tests/revert-test/tests/revert-and-restart-en.test.ts
pompon0: consted everythings

core/tests/revert-test/tests/revert-and-restart-en.test.ts
pompon0: camelCased everything

core/tests/revert-test/tests/revert-and-restart-en.test.ts
```diff
@@ -0,0 +1,365 @@
...
+import { BigNumber, ethers } from 'ethers';
+import { expect, assert } from 'chai';
+import fs from 'fs';
+import * as child_process from 'child_process';
+import * as dotenv from 'dotenv';
+
+const main_env: string = 'docker';
+const ext_env: string = 'ext-node-docker';
+const main_logs_path: string = 'revert_main.log';
+const ext_logs_path: string = 'revert_ext.log';
```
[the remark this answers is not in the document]
pompon0: camelCased everything

core/tests/revert-test/tests/revert-and-restart-en.test.ts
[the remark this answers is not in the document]
popzxc: In the past we used to use `expect(a.eq(b)).to.be.true` for BigNumbers, but it's indeed kinda clunky.

.github/workflows/ci-core-reusable.yml
```diff
@@ -120,6 +120,7 @@ jobs:
   integration:
     name: Integration (consensus=${{ matrix.consensus }})
     strategy:
+      fail-fast: false
```
popzxc, line 123: Leftover (here and below)?
We can, in theory, discuss whether we want to disable fail-fast in CI, but I'd expect it to be done separately and with a prior discussion internally.

core/bin/external_node/src/main.rs
```diff
@@ -567,6 +540,53 @@ async fn main() -> anyhow::Result<()> {
...
+        NodeRole::External,
+        config.required.state_cache_path.clone(),
+        config.required.merkle_tree_path.clone(),
+        None,
+        connection_pool.clone(),
+        L1ExecutedBatchesRevert::Allowed,
+    );
+
+    let mut reorg_detector = ReorgDetector::new(main_node_client.clone(), connection_pool.clone());
+    // We're checking for the reorg in the beginning because we expect
```
popzxc, line 554: Micronit: the spacing feels a bit odd here with comments being shorter even than surrounding code lines, usually (though not consistently) we align comments to 120.

core/tests/ts-integration/src/env.ts
```diff
@@ -25,22 +25,19 @@ export async function waitForServer() {
...
 
     reporter.startAction('Connecting to server');
-    let ready = false;
     for (let i = 0; i < maxAttempts; ++i) {
         try {
             await l2Provider.getNetwork(); // Will throw if the server is not ready yet.
-            ready = true;
             reporter.finishAction();
+            // TODO (PLA-842): [workaround] sleep for 10s.
+            await zksync.utils.sleep(10 * 1000);
```
popzxc, line 33: Nit: Such sleeps are often forgotten, so maybe it makes sense to do

```typescript
reporter.warn("TODO (PLA-842): Connected to server, waiting 10 seconds for it to warm up");
```

this may incentivize people to prioritize the corresponding ticket 😅

popzxc (commented):
Mostly LGTM, just a few minor things.

core/tests/ts-integration/src/env.ts
pompon0: done

core/bin/external_node/src/main.rs
pompon0: done

.github/workflows/ci-core-reusable.yml
pompon0: done

core/bin/external_node/src/main.rs
```diff
@@ -555,6 +527,53 @@ async fn main() -> anyhow::Result<()> {
...
+        Err(reorg_detector::Error::ReorgDetected(last_correct_l1_batch)) => {
+            tracing::info!("Rolling back to l1 batch number {last_correct_l1_batch}");
+            reverter
+                .rollback_db(last_correct_l1_batch, BlockReverterFlags::all())
+                .await;
+            tracing::info!("Rollback successfully completed");
+        }
+        Err(err) => return Err(err).context("reorg_detector.check_consistency()"),
+    }
+    if opt.revert_pending_l1_batch {
```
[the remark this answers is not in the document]
pompon0: IMO  revert_pending_l1_batch should be treated as a separate command: if set, it should revert and then terminate unconditionally. Then there is no conflict.

popzxc (dismissed):

popzxc (dismissed):

PR #1225, opened 2024-02-23
feat(prover): Fixing snark verification keys

mm-zk:
## What ❔

* Compute the snark verification key in rust code
* Added historical values and tests to make sure that computations are correct
* Write the small json file with commitments when generating new verification key.
* Stopped loading CONTRACTS_SNARK_WRAPPER_VK_HASH from env variable, and computing it locally instead.

## Why ❔

* In the past, the snark verification key generation was a complex process, that involved running a separate binaries
* Now it is all the part of the key generator, with a simple utils library (and without including the old test harness)
* Historical files will allow us to quickly check the correctness of the older blocks


prover/vk_setup_data_generator_server_fri/src/utils.rs
```diff
@@ -35,3 +47,95 @@ pub fn get_leaf_vk_params(
     }
     Ok(leaf_vk_commits)
 }
+
+/// Calculates the hash of a snark verification key.
+// This function corresponds 1:1 with the following solidity code: https://github.com/matter-labs/era-contracts/blob/3e2bee96e412bac7c0a58c4b919837b59e9af36e/ethereum/contracts/zksync/Verifier.sol#L260
+pub fn calculate_snark_vk_hash(keystore: &Keystore) -> anyhow::Result<H256> {
```
koloz193, line 53: Are we not able to use the existing function in the boojum cli validator?

koloz193 (approved):

PR #1233, opened 2024-02-23
feat(node_framework): Support Proof Data Handler in the framework

AnastasiiaVashchuk:
## What ❔
- Adds `Proof data handler` implementation, i. e. `ProofDataHandlerLayer` and `ProofDataHandlerTask`, in the same way it was done for `StateKeeper ` [here](https://github.com/matter-labs/zksync-era/pull/1043/files#diff-8ff8babf7b83c79dbf96f4998cf71d888beaeb265c7ce33192b0bb0c808f662b).

Current "external" point of view for the framework(the only thing changed is that we add 
`.add_proof_data_handler_layer()?` after `.add_eth_watch_layer()?`):

```rust
        .with_log_format(log_format)
        .build();

    MainNodeBuilder::new()
        .add_pools_layer()?
        .add_fee_input_layer()?
        .add_object_store_layer()?
        .add_state_keeper_layer()?
        .add_eth_watch_layer()?
        .add_proof_data_handler_layer()?
        .add_healthcheck_layer()?
        .build()
        .run()?;

    Ok(())
}
```


core/node/node_framework/src/implementations/layers/proof_data_handler.rs
```diff
@@ -0,0 +1,86 @@
...
+impl WiringLayer for ProofDataHandlerLayer {
+    fn layer_name(&self) -> &'static str {
+        "proof_data_handler_layer"
+    }
+
+    async fn wire(self: Box<Self>, context: ServiceContext<'_>) -> Result<(), WiringError> {
+        let pool_resource = context.get_resource::<MasterPoolResource>().await?;
+        let main_pool = pool_resource.get().await.unwrap();
+
+        let object_store = context.get_resource::<ObjectStoreResource>().await.unwrap();
```
popzxc, line 47: I don't think we should unwrap here.

core/node/node_framework/src/implementations/layers/proof_data_handler.rs
```diff
@@ -0,0 +1,86 @@
...
+    fn name(&self) -> &'static str {
+        "proof_data_handler"
+    }
+
+    async fn run(self: Box<Self>, stop_receiver: StopReceiver) -> anyhow::Result<()> {
+        proof_data_handler::run_server(
+            self.proof_data_handler_config
+                .clone()
+                .context("proof_data_handler_config")?,
+            self.contracts_config.clone().context("contracts_config")?,
```
popzxc, line 79: These lines don't seem right.

popzxc (commented):
Looks like CI is failing due to some compile errors.

core/node/node_framework/src/implementations/layers/proof_data_handler.rs
AnastasiiaVashchuk: `await?` then?

core/node/node_framework/src/implementations/layers/proof_data_handler.rs
AnastasiiaVashchuk: indeed.

PR #1234, opened 2024-02-23
feat(commitment-generator): Commitment for 1.4.2

perekopskiy:
## What ❔

Adds support for 1.4.2 commitment (blob hashes are taken into account)

## Checklist


core/lib/l1_contract_interface/src/i_executor/structures/commit_batch_info.rs
```diff
@@ -7,6 +9,16 @@ use zksync_types::{
...
 
+/// Loads KZG settings from the file system.
+pub fn load_kzg_settings() -> KzgSettings {
+    static KZG_SETTINGS: Lazy<KzgSettings> = Lazy::new(|| {
+        let zksync_home = std::env::var("ZKSYNC_HOME").unwrap_or_else(|_| ".".into());
+        let path = std::path::Path::new(&zksync_home).join("trusted_setup.json");
+        KzgSettings::new(path.to_str().unwrap())
+    });
+    KZG_SETTINGS.clone()
+}
```
Deniallugo, line 20: @popzxc what should we do with? 
seems we have to provide the whole path

core/lib/types/src/l2_to_l1_log.rs
```diff
@@ -44,6 +36,22 @@ impl L2ToL1Log {
         }
     }
 
+    /// Returns the minimum number of items in the Merkle tree built from L2-to-L1 logs
+    /// for a certain protocol version.
+    pub fn min_tree_size(protocol_version: ProtocolVersionId) -> usize {
```
Deniallugo, line 41: I'm not sure that it's the correct place for this function. 
 
separate functions `min_l2_l1_messages_tree_size` seems more appropriate

Deniallugo: L2L1Log knows nothing about tree

core/lib/zksync_core/src/commitment_generator/mod.rs
```diff
@@ -210,11 +214,32 @@ impl CommitmentGenerator {
...
+                    .pubdata_input
+                    .expect("pubdata_input must be present for post 1.4.2 batches")
+                    .chunks(ZK_SYNC_BYTES_PER_BLOB)
+                    .map(|blob| {
+                        let kzg_info = KzgInfo::new(&kzg_settings, blob.to_vec());
+                        H256(kzg_info.to_blob_commitment())
+                    })
+                    .collect::<Vec<_>>();
+                [
+                    blob_commitments.get(0).copied().unwrap_or_default(),
```
mm-zk, line 230: Add a comment what happens when pubdata_input is < 128kb
(that the second commitment will be computed as 0s)

mm-zk (dismissed):

core/lib/types/src/l2_to_l1_log.rs
perekopskiy: Agree, moved to a separate function

core/lib/types/src/commitment/mod.rs
```diff
@@ -327,20 +330,32 @@ impl L1BatchAuxiliaryOutput {
                 system_logs,
                 state_diffs,
                 aux_commitments,
+                blob_commitments,
             } => {
+                // TODO (PLA-815): Remove `IGNORE_1_4_2_UPGRADE`.
+                let protocol_version = if std::env::var("IGNORE_1_4_2_UPGRADE").is_ok() {
```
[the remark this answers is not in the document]
Deniallugo: should we just switch era contracts?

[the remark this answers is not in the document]
perekopskiy: We can't update era-contracts currently, we either need this kludge so upgrade test passes or merge these changes in the final PR with contracts updated. I think first is better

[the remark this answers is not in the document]
Deniallugo: ok, I'll try to fix it with a new tool ;)

core/lib/types/src/l2_to_l1_log.rs
```diff
@@ -63,6 +55,22 @@ impl L2ToL1Log {
     }
 }
 
+/// Returns the minimum number of items in the Merkle tree built from L2-to-L1 logs
+/// for a certain protocol version.
+pub fn min_l2_to_l1_logs_tree_size(protocol_version: ProtocolVersionId) -> usize {
```
koloz193, line 60: this wont work if we're using 1.4.2 with 1 blob since we'll only be able to handle 2048 logs

perekopskiy: Why? This function sets a height for a tree. Pubdata seal criterion (which has to do with this function) would close batches by setting pubdata limit depending on config

koloz193: doesn't it affect the size of the mini merkle tree thats constructed? which would change the root [here](https://github.com/matter-labs/zksync-era/blob/80c50291c8f8439b1a9c11d680ebd00c85959e52/core/lib/types/src/commitment/mod.rs#L302)

perekopskiy: Yes, it affects. But I still don't get why does it matter

koloz193: just wanted to double check that we didn't end up with more logs than we could fit in a batch. it sound like this is a non issue though so ill resolve

core/lib/l1_contract_interface/src/i_executor/structures/commit_batch_info.rs
```diff
@@ -7,6 +9,16 @@ use zksync_types::{
 
 use crate::Tokenizable;
 
+/// Loads KZG settings from the file system.
+pub fn load_kzg_settings() -> KzgSettings {
+    static KZG_SETTINGS: Lazy<KzgSettings> = Lazy::new(|| {
+        let zksync_home = std::env::var("ZKSYNC_HOME").unwrap_or_else(|_| ".".into());
```
popzxc: I don't think it's a good idea to make `l1_contract_interface` depend on `ZKSYNC_HOME` (as it's an otherwise pretty much a pure crate that can be used anyhwere).

You can move this logic to the place that actually uses it (I guess commitment generator).

core/lib/l1_contract_interface/src/i_executor/structures/commit_batch_info.rs
```diff
@@ -7,6 +9,16 @@ use zksync_types::{
 
 use crate::Tokenizable;
 
+/// Loads KZG settings from the file system.
+pub fn load_kzg_settings() -> KzgSettings {
```
popzxc: Also, I don't think that you need to clone the whole object each time. You can return a `&'static KzgSettings` instead -- `once_cell` allows that.

core/lib/l1_contract_interface/src/i_executor/structures/commit_batch_info.rs
popzxc: Responded elsewhere. I think it's fine to depend on `ZKSYNC_HOME`, just in a place that is aware of `ZKSYNC_HOME` anyways.

core/lib/types/src/commitment/mod.rs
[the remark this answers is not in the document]
popzxc: Sorry for unresolving. Can we make this kludge a bit more verbose by 
1. Changing the env var to `IGNORE_1_4_2_UPGRADE_FOR_UPGRADE_TEST`?
2. Add an explanation comment: "// This is used only in upgrade test, because our system isn't flexible enough right now to support WIP changes to contracts" or something like this. Issues in linear are getting removed after a while, so we may forget it and then keep wondering whether it's needed for some long forgotten tooling.
3. Add a `tracing::error!("Ignoring the 1.4.2 upgrade");` just in case.

core/lib/zksync_core/src/api_server/web3/namespaces/zks.rs
```diff
@@ -334,17 +334,12 @@ impl ZksNamespace {
...
             .protocol_version
-            .map(|v| v.is_pre_boojum())
-            .unwrap_or(true)
-        {
-            Some(L2ToL1Log::PRE_BOOJUM_MIN_L2_L1_LOGS_TREE_SIZE)
-        } else {
-            Some(L2ToL1Log::MIN_L2_L1_LOGS_TREE_SIZE)
-        };
+            .unwrap_or_else(ProtocolVersionId::last_potentially_undefined);
+        let min_tree_size = min_l2_to_l1_logs_tree_size(protocol_version);
```
mm-zk, line 340: or alternatively - we could allow min_l2_to_l1_logs_tree_size to accept Option<ProtocolVersionId> instead.

core/lib/types/src/l2_to_l1_log.rs
```diff
@@ -63,6 +55,22 @@ impl L2ToL1Log {
     }
 }
 
+/// Returns the minimum number of items in the Merkle tree built from L2-to-L1 logs
```
mm-zk, line 58: minimum or maximum?

core/lib/zksync_core/src/commitment_generator/mod.rs
```diff
@@ -210,11 +214,35 @@ impl CommitmentGenerator {
             }
             state_diffs.sort_unstable_by_key(|rec| (rec.address, rec.key));
 
+            let kzg_settings = load_kzg_settings();
```
mm-zk: load_kzg_settings is heavy - maybe move inside the 'if' (so that we don't invoke it in non-1.4.2 flow?)

(unless this is done on purpose to test things - but then let's mention it in the comment)

core/lib/types/src/commitment/mod.rs
```diff
@@ -350,6 +365,23 @@ impl L1BatchAuxiliaryOutput {
...
+                    })
+                    .unwrap_or_default();
+                let blob2_linear_hash = system_logs
+                    .iter()
+                    .find_map(|log| {
+                        (log.0.sender == PUBDATA_CHUNK_PUBLISHER_ADDRESS
+                            && log.0.key == H256::from_low_u64_be(BLOB2_LINEAR_HASH_KEY as u64))
+                        .then_some(log.0.value)
+                    })
+                    .unwrap_or_default();
```
popzxc, line 383: 1. I understand that it's unlikely to become a bottleneck, but still iterating through the logs twice seems to be unnecessary here.
2. Would we want a sanity check that either none or both logs are present?

Probably I'd imagine a simple loop here, like

```rust
let (mut blob1_linear_hash, mut blob2_linear_hash) = (None, None);
for log in &system_logs {
    if (/* first_blob_criterion */) {
        blob1_linear_hash = Some(log.0.value);
    } else if (/* second_blob_criterion */) {
        blob2_linear_hash = Some(log.0.value);
    }
    if blob1_linear_hash.is_some() && blob2_linear_hash.is_some() {
        break;
    }
}

match (&blob1_linear_hash, &blob2_linear_hash) {
   (Some(_), None) | (None, Some(_)) => panic!("Only one blob hash was found in system logs"),
   _ => {}
}
```

core/lib/zksync_core/src/commitment_generator/mod.rs
```diff
@@ -210,11 +214,35 @@ impl CommitmentGenerator {
...
+
+            let blob_commitments = if protocol_version.is_post_1_4_2() {
+                let blob_commitments = header
+                    .pubdata_input
+                    .expect("pubdata_input must be present for post 1.4.2 batches")
```
mm-zk, line 221: mention l1_batch_number inside the expect. Will make debugging a lot easier.

core/lib/zksync_core/src/commitment_generator/mod.rs
```diff
@@ -210,11 +214,35 @@ impl CommitmentGenerator {
             }
             state_diffs.sort_unstable_by_key(|rec| (rec.address, rec.key));
 
+            let kzg_settings = load_kzg_settings();
+
+            let blob_commitments = if protocol_version.is_post_1_4_2() {
+                let blob_commitments = header
```
mm-zk, line 219: consider moving this logic into kzg.rs itself - and having a function 

```
pubdata_to_blob_commitments(Vec<u8>) -> [H256;2].

```

(this way we keep all this logic in one place)

core/lib/types/src/commitment/mod.rs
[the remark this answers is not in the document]
mm-zk: please mention it in the comment in the code above.

core/lib/types/src/l2_to_l1_log.rs
popzxc, line 60: Maybe it makes sense to make it a static method on `L2ToL1Log`?
E.g. `L2ToL1Log::min_tree_size(protocol_version)`.

core/lib/zksync_core/src/commitment_generator/mod.rs
popzxc: I think you don't need to load KZG settings for pre-1.4.2, so you probably should move this logic to the `if` body (this way for non-1.4.2 the corresponding `once_cell` will never be instantiated).

popzxc: It isn't heavy due to it being an once_cell, but moving it inside `if` certainly makes sense.

core/lib/types/src/l2_to_l1_log.rs
perekopskiy: @Deniallugo suggested vice versa https://github.com/matter-labs/zksync-era/pull/1234#discussion_r1500913056, i.e. to replace static method with a function. I don't have strong opinion, IMO both variants are ok

popzxc: lol, so be it

core/lib/l1_contract_interface/src/i_executor/structures/commit_batch_info.rs
perekopskiy: function was removed `load_kzg_settings`

core/lib/zksync_core/src/api_server/web3/namespaces/zks.rs
mm-zk: ack.

core/lib/types/src/commitment/mod.rs
```diff
@@ -350,6 +365,23 @@ impl L1BatchAuxiliaryOutput {
                 let state_diffs_hash = H256::from(keccak256(&(state_diffs_packed)));
                 let state_diffs_compressed = compress_state_diffs(state_diffs);
 
+                let blob1_linear_hash = system_logs
+                    .iter()
+                    .find_map(|log| {
+                        (log.0.sender == PUBDATA_CHUNK_PUBLISHER_ADDRESS
+                            && log.0.key == H256::from_low_u64_be(BLOB1_LINEAR_HASH_KEY as u64))
```
StanislavBreadless, line 372: Note, that for the 1.4.1 version the hash key `7` used the key for the upgrade tx hash, in 1.4.2 it was just shifted:
image

In other words, the code here will not work fine for the batch with the 1.4.1 upgrade (may be a problem for EN) or the batch where we had to do some emergency upgrade and unlike 1.4.2 we did not change the bootloader

core/lib/types/src/commitment/mod.rs
StanislavBreadless: I'd suggest to keep things as they are... Note that the same `find_map` construction is used multiple times after the changed lines (so to "fully" prevent multiple iterations through the array we'll need 4 mutables variables that we'll track). Also, the total length of the array is unlikely to ever exceed 30 (right now it has 10 elements max)

core/lib/types/src/commitment/mod.rs
perekopskiy: Nice catch! I will fix it by checking if version is post 1.4.2 before reading blob hashes from system logs

PR #1235, opened 2024-02-23
feat(Validium): add validium support for external node

lferrigno:
## What

This pull request addresses the need to dynamically set the L1CommitGeneratorMode of the external node based on the configuration value l1_batch_commit_data_generator_mode in the etc/env/ext-node.toml file. If the configuration value is set, the external node will be configured for Validium; otherwise, the default mode will be Rollup.
Changes Made
Added logic to check for the presence of l1_batch_commit_data_generator_mode in etc/env/ext-node.toml.
If the value is set, the L1CommitGeneratorMode of the external node is set to "validium."
If the value is not set, the default mode "rollup" is used.
Users can manually update the configuration by adding l1_batch_commit_data_generator_mode="Validium" to the ext-node.toml file.

## Why ❔

Need to add Valiidium support to EN

## Checklist

### DIFF
--- a/core/bin/external_node/src/config/mod.rs
+++ b/core/bin/external_node/src/config/mod.rs
@@ -4,6 +4,7 @@ use anyhow::Context;
 use serde::Deserialize;
 use url::Url;
 use zksync_basic_types::{Address, L1ChainId, L2ChainId};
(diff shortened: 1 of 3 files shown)


mationorato:
Hey @popzxc , thanks for the review.

Do you think we should add a special test case for this?

popzxc:
@mationorato I think it's fine to merge as-is.

PR #1237, opened 2024-02-23
fix(Validium): Improved changes for the `zk` tool

ilitteri:
## What ❔

Create a new constants file exclusively for configuration files, replacing the old one.
Move the implementation of writing configuration files from `init.ts` to `config.ts`.

## Why ❔

Validium introduces some changes to `zk` tool (init.ts mainly, but not related to it) that do not follow common patterns.

The new behavior is undocumented and does not follow the hierarchical structure of the module.

## Checklist


infrastructure/zk/src/config.ts
```diff
@@ -112,6 +172,17 @@ export function compileConfig(environment?: string) {
     console.log(`Configs compiled for ${environment}`);
 }
 
+// If running ValidiumMode config files should be updated and recompiled to get the right constant values
+// UpdateConfig should be run on init
+export function updateConfig(validiumMode: boolean) {
+    updateChainConfig(validiumMode);
+    updateEthSenderConfig(validiumMode);
```
popzxc, line 179: I guess this no longer will be needed after https://github.com/matter-labs/zksync-era/pull/1238 is merged

infrastructure/zk/src/config.ts
```diff
@@ -112,6 +172,17 @@ export function compileConfig(environment?: string) {
     console.log(`Configs compiled for ${environment}`);
 }
 
+// If running ValidiumMode config files should be updated and recompiled to get the right constant values
+// UpdateConfig should be run on init
+export function updateConfig(validiumMode: boolean) {
```
popzxc, line 177: We have recently added support of defining a dynamic set of config files for compilation.
An example can be found here: [`etc/env/dev.toml`](https://github.com/matter-labs/zksync-era/blob/main/etc/env/dev.toml).
This mechanism also supports overrides natively, e.g. [here](https://github.com/matter-labs/zksync-era/blob/main/etc/env/ext-node-docker.toml).

Basically, in `[_metadata]` you specify your dependencies and otherwise you can write the overrides you need.
With that, you can define a `dev_validium.toml` which will have the overrides and then in `[_metadata]` if would have `base = ["dev.toml"]`.

Then, when you compile configs, you'll choose either `dev.toml` or `dev-validium.toml` depending on the mode.

With this mechanism, I think you can implement a more generic approach:
1. Collect variables that differ for rollup/validium.
2. Create `rollup_configs.toml` and `validium_configs.toml` where you define the values.
3.

popzxc: I think for now it's fine to keep `--validium-mode` CLI flag -- it actually makes the process explicit, so it's not really a bad thing.

We have redesign of `zk` in progress, so we'll probably think on how to express it better in the next iteration of this tool.

ilitteri: Ok, we'll create `dev_validium.toml` but we'll keep the flag 👍

infrastructure/zk/src/config_constants.ts
```diff
@@ -0,0 +1,36 @@
+// Include constants to be used on config files.
```
popzxc, line 1: Is this file used anywhere now?

ilitteri: No it is not, removing it asap

popzxc (approved):

PR #1238, opened 2024-02-23
feat(Validium): Abstract `GasAdjuster` struct into a trait

ilitteri:
## What ❔

Abstracts `GasAdjuster` struct into a trait which is implemented by new structs `RollupGasAdjuster` and `ValidiumGasAdjuster`

## Why ❔

It was agreed in this comment https://github.com/matter-labs/zksync-era/pull/1017#discussion_r1481485699,

## Checklist


core/lib/zksync_core/src/l1_gas_price/gas_adjuster/validium_gas_adjuster.rs
```diff
@@ -0,0 +1,61 @@
...
+        &self.statistics
+    }
+
+    fn into_l1_tx_params_provider(self: Arc<Self>) -> Arc<dyn L1TxParamsProvider> {
+        self
+    }
+
+    fn estimate_effective_pubdata_price(&self) -> u64 {
+        0
+    }
```
popzxc, line 36: This seems to be the only difference in behavior for now, with a lot of other interfaces being exposes just for the sake of avoiding code duplication.
However, we can also avoid code duplication by abstracting over the things that are different. In this case, the pubdata pricing. So, we can define a trait for that specifically:

```rust
pub trait PubdataPricing {
    /// Returns the amount of L1 gas to publish a single byte of pub data.
    fn pubdata_byte_gas(&self) -> u64;
}
```

Then we can have `ValidiumPubdataPricing` and `RollupPubdataPricing` implementing this trait.
`GasAdjuster` will no longer have to be a trait as a whole, but instead it'll hold `Box<dyn PubdataPricing>` inside, which will be called inside of `estimate_effective_pubdata_price`.

ilitteri: I agree with this, it is a much simpler approach. We came to this solution from what was discussed here https://github.com/matter-labs/zksync-era/pull/1017#discussion_r1481485699 with @RomanBrodetski. Let us know what you prefer and we'll go in that direction.

popzxc: Yeah, I've seen this discussion, and I think that @RomanBrodetski in fact had something like this in mind, he just described a very rough draft of the idea (e.g. even in the current implementation, most of functionality remains in GasAdjuster, so it doesn't need to be abstracted itself, only its parts).

Let's switch to having a trait for pubdata pricing only.

core/lib/env_config/src/eth_sender.rs
```diff
@@ -64,15 +64,12 @@ mod tests {
...
-                l1_gas_per_pubdata_byte: 17,
             },
         }
     }
 
     fn validium_expected_config() -> ETHSenderConfig {
-        let mut validium_config = rollup_expected_config();
-        validium_config.gas_adjuster.l1_gas_per_pubdata_byte = 0;
-        validium_config
+        rollup_expected_config()
```
popzxc, line 72: I guess the different logic for rollup/validium can be removed altogether now.

ilitteri: Yes! We left this here because its being reworked with the hole batch commit data interface

ilitteri:
Superseded by #1357

PR #1240, opened 2024-02-23
fix(Validium): Add commitment mode cross check

ilitteri:
## What ❔

Adds a cross-check of the local mode with L1 when setting up the server.

## Why ❔

It is now possible to run the node with an incorrect config that doesn’t match the L1 config.

We should force the node to ask the pre-deployed contracts on L1 in which mode it should run.

## Checklist


ilitteri:
For this PR to be ready for review we need to use the `call_contract_function` method which requires the contract ABI. For that, we'll create a function `executor_contract` in the `zksync_contracts` module.

core/lib/zksync_core/src/lib.rs
```diff
@@ -630,6 +630,25 @@ pub async fn initialize_components(
                     Arc::new(ValidiumModeL1BatchCommitDataGenerator {})
                 }
             };
+
+        let args = CallFunctionArgs::new("getPubdataPriceMode", ()).for_contract(
```
popzxc, line 634: Also, probably it would make sense to encapsulate this logic to a separate function. For now I think we can keep it in the same file.

core/lib/zksync_core/src/lib.rs
```diff
@@ -630,6 +630,25 @@ pub async fn initialize_components(
...
+        let current_commitment_mode_eth_response = eth_client.call_contract_function(args).await?;
+
+        let current_commitment_mode =
+            L1BatchCommitDataGeneratorMode::from_tokens(current_commitment_mode_eth_response)?;
+
+        // contracts mode == server mode
+        assert_eq!(
+            current_commitment_mode,
+            state_keeper_config.l1_batch_commit_data_generator_mode,
+            "The selected L1BatchCommitDataGeneratorMode ({:?}) does not match the existing commitment mode ({:?})",
```
popzxc, line 647: Maybe replace "existing commitmend mode" with "commitment mode used on L1 contract" to make it more clear?

popzxc (commented):
Nice!

core/lib/zksync_core/src/lib.rs
ilitteri: Done in [d58e1ac](https://github.com/matter-labs/zksync-era/pull/1240/commits/d58e1acd5b3d4d8393fa8395f038d8ec731f10ab)

core/lib/zksync_core/src/lib.rs
```diff
@@ -630,6 +630,25 @@ pub async fn initialize_components(
                     Arc::new(ValidiumModeL1BatchCommitDataGenerator {})
                 }
             };
+
+        let args = CallFunctionArgs::new("getPubdataPriceMode", ()).for_contract(
+            contracts_config.executor_facet_addr,
+            zksync_contracts::executor_contract(),
+        );
+        let current_commitment_mode_eth_response = eth_client.call_contract_function(args).await?;
```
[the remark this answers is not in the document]
ilitteri: > This version of the server can be merged before the new contracts are actually deployed to testnet/mainnet, so we must add a check against that, e.g. we only do this check if the call succeeds. A simple way to do it is to match on an error in call_contract_function. If it's a zksync_eth_client::Error::Contract, you just emit a warning that the contract does not support the required getter and proceed. If it's Ok, you do the check. And you bail on any other error.

Handled here [9586ecc](https://github.com/matter-labs/zksync-era/pull/1240/commits/9586ecc3e45dbe53232edbe55efcf3754afe3c67)

> you should write a unit test for it to make sure that locally the call can succeed.

Which would be a proper place for this test?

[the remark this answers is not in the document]
popzxc: You can create `zksync_core/src/tests.rs` and put it there.

core/lib/zksync_core/src/lib.rs
```diff
@@ -298,6 +298,50 @@ impl FromStr for Components {
...
+    eth_client: &impl EthInterface,
+) -> anyhow::Result<()> {
+    let selected_l1_batch_commit_data_generator_mode = state_keeper_config
+        .l1_batch_commit_data_generator_mode
+        .clone();
+    match get_pubdata_pricing_mode(&contracts_config, eth_client).await {
+        // Getters contract support getPubdataPricingMode method
+        Ok(l1_contract_pubdata_pricing_mode) => {
+            let l1_contract_batch_commitment_mode =
+                L1BatchCommitDataGeneratorMode::from_tokens(l1_contract_pubdata_pricing_mode)?;
```
popzxc, line 313: Small nit: it's usually a good idea to add [`context`](https://docs.rs/anyhow/latest/anyhow/trait.Context.html#tymethod.context) to the errors you propagate. E.g. in this case you can do `.context("Unable to parse L1BatchCommitDataGeneratorMode received from L1 contract")?`.

This helps greatly when something happens, because by default you just see some bottommost error and you may even have no idea where it comes from (if there was no `context` added up the callstack).

core/lib/config/src/configs/chain.rs
```diff
@@ -52,6 +60,31 @@ pub enum L1BatchCommitDataGeneratorMode {
...
+// The cases are extracted from the `PubdataPricingMode` enum in the L1 contracts,
+// And knowing that, in Ethereum, the response is the index of the enum case.
+// If the Uint token is 0 I want the Rollup case,
+// If the Uint token is 1 I want the Validium case,
+// Else, an error.
+impl Detokenize for L1BatchCommitDataGeneratorMode {
+    fn from_tokens(tokens: Vec<ethabi::Token>) -> Result<Self, Web3ContractError> {
+        let error = Err(Web3ContractError::Api(Web3ApiError::Decoder(
+            "L1BatchCommitDataGeneratorMode::from_tokens".to_string(),
+        )));
```
popzxc, line 72: Tip: It's usually helpful to include information that would help to debug the error if it ever happens, e.g. in this case the tokens you couldn't parse. And to avoid doing extra work in an `Ok(..)` cases, you can make it a function, e.g.

```rust
fn from_tokens(tokens: Vec<ethabi::Token>) -> Result<Self, Web3ContractError> {
  fn error(tokens: &[ethabi::Token]) -> Web3ContractError {
    Web3ContractError::Api(Web3ApiError::Decoder(
            format!("L1BatchCommitDataGeneratorMode::from_tokens: {:?}", tokens)
        ))
  }
  
 match tokens.as_slice() {
    // ...
    _ => Err(error(&tokens))
 }
}
```

popzxc (approved):
Generally LGTM, feel free to merge once you fix nits/add a test. If you want, also feel free to submit the test as a separate PR.

popzxc:
@ilitteri also btw now that it's extracted to a separate function, it probably should be easy to integrate this check to EN as well (as a separate PR).

ilitteri:
> @ilitteri also btw now that it's extracted to a separate function, it probably should be easy to integrate this check to EN as well (as a separate PR).

Thanks a lot for the nits, I'll make new PRs for adding this cross-check to the EN and for the unit test.

ilitteri:
> @ilitteri also btw now that it's extracted to a separate function, it probably should be easy to integrate this check to EN as well (as a separate PR).

@popzxc  I'm currently working on supporting this cross-check in the EN and found out that the main_node_client does not implement `EthInterface`, thus the method cannot be exactly replicated.

Is it ok to replace the type of the `main_node_client` for a `Arc<dyn EthInterface>`? Or is it better to implement `EthInterface` for `HttpClient`?

popzxc:
@ilitteri On EN `eth_client_url` is a part of required config, and you can see it being used e.g. to instantiate the consistency checker. Right now, `ConsistencyChecker` takes a URL and instantiates `QueryClient` inside `new` just to coerce it to `Arc<dyn EthInterface>`. You can change it so that `ConsistencyChecker::new` accepts `Arc<dyn EthInterface>` and `QueryClient` is instantiated outside -- this way you'll have access to it during ETH initialization logic. Does it make sense?

ilitteri:
> @ilitteri On EN `eth_client_url` is a part of required config, and you can see it being used e.g. to instantiate the consistency checker. Right now, `ConsistencyChecker` takes a URL and instantiates `QueryClient` inside `new` just to coerce it to `Arc<dyn EthInterface>`. You can change it so that `ConsistencyChecker::new` accepts `Arc<dyn EthInterface>` and `QueryClient` is instantiated outside -- this way you'll have access to it during ETH initialization logic. Does it make sense?

This totally makes sense to me, thanks

PR #1242, opened 2024-02-23
fix(Validium): Remove obscure use of `VALIDIUM_MODE` env

ilitteri:
## Disclaimer ⚠️

This PR has a counterpart in the `era-contracts` repo. The PR in question is https://github.com/matter-labs/era-contracts/pull/225. The `era-contracts` submodule is pointing to said PR's branch. Before merging this PR, after it is approved and the `era-contracts` counterpart is merged, we need to update this PR's submodule commit.

## What ❔

This PR removes the use of `VALIDIUM_MODE` env for the contracts side and replaces it with the passthrough of the flag to the `l1-contracts` deploy script.

## Why ❔

`zk` tool is setting `VALIDIUM_MODE` environment variable, which is only used by a script in a different repository (era-contracts). This is a very implicit and hard to track approach.

## Checklist


infrastructure/zk/src/init.ts
```diff
@@ -240,6 +237,10 @@ async function checkEnv() {
     }
 }
 
+function additionalArgumentsForRedeployingL1(validium: boolean) {
+    return validium ? ['--validium-mode'] : [];
+}
```
popzxc, line 242: It is a bit unexpected that `init.ts` has to know arguments for a particular contracts script.
I think it'll be better to hold the arguments close to the script that uses them.
For example, you can change the signature of `redeployL1` to accept not only args, but also `validium: bool`. If it'll be `true`, you'll push an extra argument to the list of provided args.

popzxc: Also, to make it more typesafe you may introduce an enum, e.g. `DeploymentMode { Rollup, Validium }` to use instead of a boolean flag.

ilitteri: Done in [e1e01dd](https://github.com/matter-labs/zksync-era/pull/1242/commits/e1e01dd55a9a86ebabc06e0b1737d0c2cccab83f)

popzxc (approved):
Nice!

PR #1253, opened 2024-02-26
feat: enhance unit test for batch tip

Artemka374:
## What ❔

Add a few more hardcoded constants for batch tip unit test.

## Why ❔

To measure correct values for further removing of the `batch_dry_run`

## Checklist


Artemka374:
`gas_used` metric is the same as `max_used_gas`, so didn’t add this constant. All the values are multiplied by 2 to have an overhead(in case not all edge cases are checked). Also wasn’t added constant for block writes - not sure that we can do this in this test.

core/lib/multivm/src/versions/vm_latest/tests/block_tip.rs
```diff
@@ -211,77 +230,105 @@ fn test_dry_run_upper_bound() {
...
+    assert!(
+        statistics
+            .iter()
+            .map(|s| s.pubdata_published)
+            .max()
+            .unwrap()
+            * 2
+            <= MAX_BATCH_TIP_PUBDATA_PUBLISHED as u32,
+        "MAX_BATCH_TIP_PUBDATA_PUBLISHED is too low"
+    );
```
Deniallugo, line 313: Now you lacking of info about the test by itself. 
You can't distinguish where was a failure

Artemka374: Not sure that I understand your point. Do you mean like test case?

perekopskiy: Yes, I think Danil is talking about providing test case name in panic message

core/lib/multivm/src/versions/vm_latest/tests/block_tip.rs
```diff
@@ -168,7 +179,15 @@ fn execute_test(test_data: L1MessengerTestData) -> u32 {
 
     let ergs_after = vm.vm.state.local_state.callstack.current.ergs_remaining;
 
-    ergs_before - ergs_after
+    assert_eq!(ergs_before - ergs_after, result.statistics.gas_used);
+
+    TestStatistics {
+        max_used_gas: ergs_before - ergs_after,
+        pubdata_published: result.statistics.pubdata_published,
```
perekopskiy, line 186: VM doesn't calculate `pubdata_published` for batch tip. It does it only for transactions when it calculates refunds for them. That's why you have `MAX_BATCH_TIP_PUBDATA_PUBLISHED: u64 = 0`

core/lib/multivm/src/versions/vm_latest/tests/block_tip.rs
Artemka374: Added logging for tests

core/lib/multivm/src/versions/vm_latest/constants.rs
```diff
@@ -9,6 +9,9 @@ use crate::vm_latest::old_vm::utils::heap_page_from_base;
 /// The amount of ergs to be reserved at the end of the batch to ensure that it has enough ergs to verify compression, etc.
 pub(crate) const BOOTLOADER_BATCH_TIP_OVERHEAD: u32 = 80_000_000;
 
+pub(crate) const MAX_BATCH_TIP_CIRCUIT_STATISTICS: u64 = 2000;
+pub(crate) const MAX_BATCH_TIP_METRICS_SIZE: u64 = 1500;
```
perekopskiy, line 13: Let's make naming consistent for all 3 constants

perekopskiy (commented):
Please, resolve conflicts

perekopskiy (approved):

PR #1256, opened 2024-02-27
feat(state-keeper): implement asynchronous RocksDB cache

itegulov:
## What ❔

This PR adds a stateful storage abstraction `CachedStorage` that encapsulates both RocksDB and Postgres behind short-lived `Box<dyn StorageRead>` connections. Internally `CachedStorage` keeps track if RocksDB is behind Postgres and tries to catch it up in the background if so.

Apologies if the code is a little hairy, I tried to preserve the existing invariants as much as possible to limit the impact of this on other parts of the codebase.

Tested this locally by running load tests for 3 hours (~3k miniblocks), deleting RocksDB state and then restarting. Seems to be working fine, but it catches up really fast and I am not sure how to generate more data faster. So I am hoping to test this on testnet/staging with one of the external nodes.

Also trying to come up with a way to unit test this so that is WIP for now. In the meanwhile any feedback is welcome!

Note: metric for RocksDB-Postgres lag already [exists](https://github.com/matter-labs/zksync-era/blob/main/core/lib/state/src/rocksdb/metrics.rs#L13-L14)

## Why ❔

The main design goal here was to ensure liveliness of the system by providing a way to get a `ReadStorage` implementation while avoiding blocking operations as much as possible. If RocksDB ever gets wiped for whatever reason, State Keeper will keep working on Postgres-backed `StorageRead` connections for a few hours while `CachedStorage` is trying to catch RocksDB up.

## Checklist


core/lib/zksync_core/src/state_keeper/cached_storage.rs
```diff
@@ -0,0 +1,253 @@
...
+                match rocksdb_sync_handle.await?? {
+                    Some(rocksdb_storage) => {
+                        // Caught-up successfully
+                        self.state = CachedStorageState::CaughtUp {
+                            rocksdb_l1_batch_number: rocksdb_storage
+                                .l1_batch_number()
+                                .await
+                                .unwrap_or_default(),
+                        };
+                        Ok(Some(Self::access_storage_pg(rt_handle, &self.pool).await?))
```
RomanBrodetski, line 194: why do we use postgres in this case? IIUC, we can use rocksDB

RomanBrodetski (commented):
I need to look deeper, but I really like it from the first glance - it's very well abstracted from state keeper. 

I wanted to make a step back and think about `CachedStorageState` state transitions.
* Maybe naive, but do we really need "not initialized" status? can we start catching up and determine the correct state in `new` already? 
* Note that during state keeper operation, the cache **cannot fall behind by more than one block**. When state keeper is running, there is no other process that would be able to add blocks to L1 batches table. This is not the case on startup - on startup we might be way behind (reason (1) - SSD failed and we replaced it, reason (2) - we are recovering from snapshot). Thus, I think we don't need logic that would switch **from** rocksDB **to** Postgres - only the other way around. With this I think we can simplify the logic. 


I might definitely be off here - also curious what @slowli thinks

core/lib/zksync_core/src/state_keeper/cached_storage.rs
```diff
@@ -0,0 +1,270 @@
...
+type BoxReadStorage<'a> = Box<dyn ReadStorage + Send + 'a>;
+
+/// Encapsulates a storage that can produce a short-lived [`ReadStorage`] implementation backed by
+/// either Postgres or RocksDB (if it's caught up). Maintains internal state of how behind RocksDB
+/// is compared to Postgres and actively tries to catch it up in the background.
+///
+/// This struct's main design purpose is to be able to produce [`ReadStorage`] implementation with
+/// as little blocking operations as possible to ensure liveliness.
+#[derive(Debug)]
+pub struct CachedStorage {
```
slowli, line 18: I certainly agree with @RomanBrodetski in that this looks overly complex. Conceptually, we want to represent two cases: either the storage has caught up (=> use RocksDB) or not (=> use Postgres). We can assume that the storage doesn't fall behind unless we let it to. So this should roughly correspond to

```rust
#[derive(Clone)]
enum CachedStorageFactory {
    Postgres(ConnectionPool),
    Rocksdb(RocksDB<StateKeeperColumnFamily>),
}

impl CachedStorageFactory {
    pub async fn access_storage(&self) -> Box<dyn ReadStorage + Send + '_> {
        // trivial. We may even not box the output type, but rather have an enum for it
        // (theoretically, can lead to better performance due to more
        // compiler optimizations being possible, so I'd go for it).
    }
}

pub struct CachedStorage {
    inner: Arc<Mutex<CachedStorageFactory>>,
    // ^ the background update task will replace `inner` once caught up.
}

impl CachedStorage {
    fn factory(&self) -> CachedStorageFactory {
         // it's important that we don't hold the lock for long;
         // that's why `CachedStorageFactory` implements `Clone`
         self.inner.lock().expect("poisoned").clone()
    }
}
```

- With this approach, an update task can be treated in the same way as other server / EN tasks; you wouldn't need to retain its handle locally.
- If you'd want to retain `ConnectionPool` in `CachedStorageFactory` even after it caught up (e.g., for it to be able to update to the latest L1 batch), this is possible as well by slightly changing its shape.

core/lib/zksync_core/src/state_keeper/batch_executor/main_executor.rs
```diff
@@ -94,13 +84,22 @@ impl BatchExecutor for MainBatchExecutor {
...
+        let stop_receiver = stop_receiver.clone();
         let handle = tokio::task::spawn_blocking(move || {
-            executor.run(
-                secondary_storage,
-                l1_batch_params,
-                system_env,
-                upload_witness_inputs_to_gcs,
-            )
+            let rt_handle = Handle::current();
+            let mut cached_storage = rt_handle.block_on(cached_storage.write());
```
slowli, line 91: - Nit: If you only use `RwLock::write()`, it makes sense to replace it with `Mutex`.
- In the general case, this would look like a *suspicious* use of `RwLock`: you lock it exclusively for an extended period of time. AFAIU, it works right now because L1 batches are processed strictly sequentially, but in general reducing the lock time to a minimum (like in the approach I've outlined in another comment) is always preferable.

core/lib/zksync_core/src/state_keeper/cached_storage.rs
```diff
@@ -0,0 +1,270 @@
...
+    /// Returns `Ok(None)` if the process is interrupted using `stop_receiver`.
+    ///
+    /// # Errors
+    ///
+    /// Propagates RocksDB and Postgres errors.
+    pub async fn access_storage(
+        &mut self,
+        rt_handle: Handle,
+        stop_receiver: watch::Receiver<bool>,
+    ) -> anyhow::Result<Option<BoxReadStorage>> {
```
slowli, line 174: Nit: You should add a lifetime here to make clear that `BoxReadStorage` borrows from `self`. This could make long-term `RwLock` locking more visible as well.

core/lib/zksync_core/src/state_keeper/batch_executor/main_executor.rs
itegulov: Right, I think I was confused by lifetimes here and presumed that the guard gets dropped after `access_storage`. So your other comment about explicit lifetime makes a lot of sense!

core/lib/zksync_core/src/state_keeper/cached_storage.rs
itegulov: Indeed, I tried to account for the possibility that RocksDB somehow falls behind Postgres in the middle of the execution but if we don't care about that it simplifies a lot of things.

I have tried out your approach and it works perfectly. Just wondering what do you think I should do about errors? The existing codebase seems to be pretty split between:
a. Just unwrapping
b. Using `anyhow` for everything
c. Proper structured errors.

I feel like structured errors could make some sense here even if most underlying code is just `anyhow`. Also, representing interruption via an error would look cleaner than `None` as it is now (inherited approach from `RocksdbStorage`).

core/lib/zksync_core/src/state_keeper/cached_storage.rs
```diff
@@ -0,0 +1,253 @@
...
+/// either Postgres or RocksDB (if it's caught up). Maintains internal state of how behind RocksDB
+/// is compared to Postgres and actively tries to catch it up in the background.
+///
+/// This struct's main design purpose is to be able to produce [`ReadStorage`] implementation with
+/// as little blocking operations as possible to ensure liveliness.
+#[derive(Debug)]
+pub struct CachedStorage {
+    pool: ConnectionPool,
+    state_keeper_db_path: String,
+    enum_index_migration_chunk_size: usize,
```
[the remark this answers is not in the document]
itegulov: I think I would rather leave this for a separate PR - this one is trying to do a lot of stuff as is

itegulov:
@RomanBrodetski @slowli I have taken your comments into consideration and refactored the code according to them. Specifically, I am now operating under presumption that RocksDB can never fall more than one block behind after catching up which indeed simplified a lot of things.

I have also made the solution a bit more generic by introducing a trait `ReadStorageFactory` that regulates which `ReadStorage` implementations can be returned which is good for testing - I made so that we test some features specifically on Postgres-based storage or specifically on RocksDB-based storage. Also allowed me to write a test that can mimic RocksDB falling behind by executing a bunch of txs on Postgres first. Hopefully it did not make the PR too overengineered :)

PTAL when you can!

core/lib/zksync_core/src/state_keeper/state_keeper_storage.rs
```diff
@@ -0,0 +1,243 @@
...
+    Postgres(ConnectionPool),
+    Rocksdb(RocksDB<StateKeeperColumnFamily>, ConnectionPool),
+}
+
+impl AsyncRocksdbCache {
+    /// Returns a [`ReadStorage`] implementation backed by Postgres
+    async fn access_storage_pg(
+        rt_handle: Handle,
+        pool: &ConnectionPool,
+    ) -> anyhow::Result<PgOrRocksdbStorage> {
```
slowli, line 105: Please use `<'_>` for `PgOrRocksdbStorage` to cleanly specify that it has a lifetime.

core/lib/zksync_core/src/state_keeper/state_keeper_storage.rs
```diff
@@ -0,0 +1,243 @@
...
+            .context("Failed initializing RocksDB storage")?;
+        let rocksdb = rocksdb_builder
+            .synchronize(conn, stop_receiver)
+            .await
+            .context("Failed to catch up state keeper RocksDB storage to Postgres")?;
+        if let Some(rocksdb) = &rocksdb {
+            let rocksdb_l1_batch_number = rocksdb.l1_batch_number().await.unwrap_or_default();
+            tracing::debug!(%rocksdb_l1_batch_number, "Using RocksDB-based storage");
+        } else {
+            tracing::warn!("Interrupted")
```
slowli, line 157: Nit: missing semicolon. Also, I wouldn't consider this log warranting WARN level; interruption is a part of the normal node function.

Also, completely subjectively, a let-else would read better here:

```rust
let Some(rocksdb) = rocksdb else {
    tracing::info!("Synchronizing RocksDB interrupted");
    return Ok(None);
};
// work with `rocksdb`
```

core/lib/zksync_core/src/state_keeper/state_keeper_storage.rs
```diff
@@ -0,0 +1,243 @@
...
+
+impl StateKeeperStorage<AsyncRocksdbCache> {
+    pub fn async_rocksdb_cache(
+        pool: ConnectionPool,
+        state_keeper_db_path: String,
+        enum_index_migration_chunk_size: usize,
+    ) -> Self {
+        let inner = Arc::new(Mutex::new(AsyncRocksdbCache::Postgres(pool.clone())));
+        let factory = inner.clone();
+        tokio::task::spawn(async move {
```
slowli, line 218: Please define a type for the task so that it can be used in the [node framework](https://github.com/matter-labs/zksync-era/tree/main/core/node/node_framework) that @popzxc is developing. Also, in any case, you want to make this task cancelable (i.e., having it take `stop_receiver` rather than creating a surrogate one), and to propagate errors for it (so that if an errors occurs in it, it can be monitored and leads to the node termination). Maybe, we'd even want to have a [healthcheck](https://github.com/matter-labs/zksync-era/tree/main/core/lib/health_check) for this task to monitor syncing progress (naturally, this can be done in a follow-up PR). You can look at the API server as an example of the component with multiple tasks.

core/lib/zksync_core/src/state_keeper/state_keeper_storage.rs
```diff
@@ -0,0 +1,243 @@
...
+use zksync_storage::RocksDB;
+
+/// Encapsulates a storage that can produce a [`ReadStorageFactory`] on demand.
+///
+/// This struct's main design purpose is to be able to produce a (cheap) owned [`ReadStorageFactory`]
+/// implementation that can then be used to obtain a short-lived [`ReadStorage`] handle bound to
+/// the respective factory.
+#[derive(Debug, Clone)]
+pub struct StateKeeperStorage<T: ReadStorageFactory> {
+    // FIXME: `Arc<Mutex<_>>` is a bit of a leaky abstraction here as it exists mostly because of
```
slowli, line 20: You could encapsulate `Arc<Mutex<_>>` wrapper inside `AsyncRocksdbCache` with something like

```rust
pub struct AsyncRocksdbCache(Arc<Mutex< AsyncRocksdbCacheInner>>);

enum AsyncRocksdbCacheInner {
    Postgres(ConnectionPool),
    Rocksdb(RocksDB<StateKeeperColumnFamily>, ConnectionPool),
}
```

One other thing you could try is to use [`OnceCell`](https://docs.rs/once_cell/latest/once_cell/sync/struct.OnceCell.html) to avoid a `Mutex` wrapper, so something like

```rust
struct AsyncRocksdbCache {
     pool: ConnectionPool,
     rocksdb: Arc<OnceCell<RocksDB<StateKeeperColumnFamily>>>,
}
```

core/lib/zksync_core/src/state_keeper/state_keeper_storage.rs
```diff
@@ -0,0 +1,243 @@
...
+#[derive(Debug)]
+pub enum PgOrRocksdbStorage<'a> {
+    Postgres(PostgresStorage<'a>),
+    RocksdbStorage(RocksdbStorage),
+}
+
+impl<'a> ReadStorage for PgOrRocksdbStorage<'a> {
+    fn read_value(&mut self, key: &zksync_types::StorageKey) -> zksync_types::StorageValue {
+        match self {
+            PgOrRocksdbStorage::Postgres(postgres) => postgres.read_value(key),
```
slowli, line 50: Nit: You can use `Self::*` instead of `PgOrRocksdbStorage::*` in match clauses here and in other `ReadStorage` methods.

core/lib/zksync_core/src/state_keeper/state_keeper_storage.rs
```diff
@@ -0,0 +1,243 @@
...
+
+/// A [`ReadStorage`] implementation that uses either [`PostgresStorage`] or [`RocksdbStorage`]
+/// underneath.
+#[derive(Debug)]
+pub enum PgOrRocksdbStorage<'a> {
+    Postgres(PostgresStorage<'a>),
+    RocksdbStorage(RocksdbStorage),
+}
+
+impl<'a> ReadStorage for PgOrRocksdbStorage<'a> {
```
slowli, line 47: Nit: You can elide the lifetime here: `impl ReadStorage for PgOrRocksdbStorage<'_>`.

core/lib/zksync_core/src/state_keeper/batch_executor/main_executor.rs
```diff
@@ -22,67 +24,48 @@ use crate::{
...
     },
 };
 
 /// The default implementation of [`BatchExecutor`].
 /// Creates a "real" batch executor which maintains the VM (as opposed to the test builder which doesn't use the VM).
 #[derive(Debug, Clone)]
-pub struct MainBatchExecutor {
-    state_keeper_db_path: String,
-    pool: ConnectionPool,
+pub struct MainBatchExecutor<T: ReadStorageFactory = AsyncRocksdbCache> {
```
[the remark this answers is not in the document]
popzxc: `Arc<dyn Trait>` is our best friend now.

core/lib/state/src/utils.rs
```diff
@@ -0,0 +1,21 @@
...
+use zksync_storage::RocksDB;
+
+use crate::StateKeeperColumnFamily;
+
+/// Opens a RocksDB database with the given path and family descriptors used by state keeper.
+///
+/// # Errors
+///
+/// Propagates RocksDB I/O errors.
+pub async fn open_state_keeper_rocksdb(
```
slowli, line 13: Bikeshedding: IMO, this change can be avoided; you can still retain `RocksdbStorage::builder()` method as a way to initialize the storage + convert the builder to `RocksDB<StateKeeperColumnFamily>` if required.

core/lib/zksync_core/src/state_keeper/batch_executor/tests/mod.rs
```diff
@@ -29,15 +38,86 @@ fn assert_reverted(execution_result: &TxExecutionResult) {
...
+    Postgres,
+}
+
+impl StorageType {
+    const ALL: [Self; 3] = [Self::AsyncRocksdbCache, Self::Rocksdb, Self::Postgres];
+
+    async fn create_batch_executor(&self, tester: &Tester) -> BatchExecutorHandle {
+        let (l1_batch_env, system_env) = tester.default_batch_params();
+        match self {
+            StorageType::AsyncRocksdbCache => tester.create_batch_executor().await,
```
slowli, line 54: Bikeshedding: the methods in `StorageType` look like they belong to `Tester`, since they mostly operate on it, if this makes sense.

core/lib/zksync_core/src/state_keeper/batch_executor/tests/mod.rs
```diff
@@ -69,11 +149,22 @@ impl SnapshotRecoveryMutation {
     }
 }
 
+const EXECUTE_L2_TX_AFTER_SNAPSHOT_RECOVERY_CASES: TestCases<(
+    Option<SnapshotRecoveryMutation>,
+    StorageType,
+)> = cases!(itertools::iproduct!(
+    SnapshotRecoveryMutation::ALL,
+    StorageType::ALL
+));
```
slowli, line 158: Nit: There's [`Product`](https://docs.rs/test-casing/latest/test_casing/struct.Product.html) in `test-casing` that provides the same functionality without a need to use another crate.

core/lib/zksync_core/src/state_keeper/state_keeper_storage.rs
```diff
@@ -0,0 +1,249 @@
...
+                    (
+                        snapshot_recovery.miniblock_number,
+                        snapshot_recovery.l1_batch_number,
+                    )
+                }
+            };
+
+        tracing::debug!(%l1_batch_number, %miniblock_number, "Using Postgres-based storage");
+        Ok(
+            PostgresStorage::new_async(rt_handle, connection, miniblock_number, true)
```
slowli, line 133: Nit: IMO, it wouldn't be controversial to retrieve a `Handle` here via [`Handle::current()`](https://docs.rs/tokio/latest/tokio/runtime/struct.Handle.html#method.current), or even in `PostgresStorage::new_async()` so that it doesn't need to be passed around.

core/lib/zksync_core/src/state_keeper/batch_executor/tests/read_storage_factory.rs
```diff
@@ -0,0 +1,104 @@
...
+        rt_handle: Handle,
+        _stop_receiver: &watch::Receiver<bool>,
+    ) -> anyhow::Result<Option<PgOrRocksdbStorage<'a>>> {
+        let mut connection = self
+            .pool
+            .access_storage()
+            .await
+            .context("Failed accessing Postgres storage")?;
+
+        let snapshot_recovery = connection
```
slowli, line 27: This looks incorrect; looks like you've fixed snapshot-related logic in the main code, but not here. Wouldn't it be possible to reuse the code from `AsyncRocksdbCache` here and in `impl ReadStorageFactory for RocksdbFactory` so that such mistakes are eliminated by design?

core/lib/zksync_core/src/state_keeper/state_keeper_storage.rs
```diff
@@ -0,0 +1,249 @@
...
+    StateKeeperColumnFamily,
+};
+use zksync_storage::RocksDB;
+use zksync_types::{L1BatchNumber, MiniblockNumber};
+
+/// Factory that can produce a [`ReadStorage`] implementation on demand.
+#[async_trait]
+pub trait ReadStorageFactory: Debug + Send + Sync + 'static {
+    async fn access_storage<'a>(
+        &'a self,
```
slowli, line 19: Nit: The lifetime can be elided here; Rust assumes that the output borrows from `self` by default.

core/lib/zksync_core/src/state_keeper/state_keeper_storage.rs
```diff
@@ -0,0 +1,249 @@
...
+    ) -> anyhow::Result<Option<(MiniblockNumber, L1BatchNumber)>> {
+        let mut dal = connection.blocks_dal();
+        let Some(l1_batch_number) = dal.get_sealed_l1_batch_number().await? else {
+            return Ok(None);
+        };
+        let Some((_, miniblock_number)) =
+            dal.get_miniblock_range_of_l1_batch(l1_batch_number).await?
+        else {
+            return Ok(None);
+        };
```
slowli, line 100: Nit: I'd consider returning `get_miniblock_range_of_l1_batch` returning `Ok(None)` an error here because the L1 batch is guaranteed to exist. So, it would make sense to convert it to an error using `.context(_)`.

core/lib/zksync_core/src/state_keeper/state_keeper_storage.rs
```diff
@@ -0,0 +1,249 @@
...
+                None => {
+                    tracing::info!("Could not find latest sealed miniblock, loading from snapshot");
+                    let snapshot_recovery = connection
+                        .snapshot_recovery_dal()
+                        .get_applied_snapshot_status()
+                        .await
+                        .context("failed getting snapshot recovery info")?
+                        .ok_or_else(|| {
+                            anyhow::anyhow!("Could not find snapshot, no state available")
+                        })?;
```
slowli, line 123: Nit: You can use `anyhow::Context::context()` on `Option`s as well, so this can be shortened to `.context("Could not find snapshot, no state available")`.

core/lib/zksync_core/src/state_keeper/state_keeper_storage.rs
```diff
@@ -0,0 +1,249 @@
...
+}
+
+impl AsyncRocksdbCache {
+    /// Load latest sealed miniblock from the latest sealed L1 batch (ignores sealed miniblocks
+    /// from unsealed batches).
+    async fn load_latest_sealed_miniblock(
+        connection: &mut StorageProcessor<'_>,
+    ) -> anyhow::Result<Option<(MiniblockNumber, L1BatchNumber)>> {
+        let mut dal = connection.blocks_dal();
+        let Some(l1_batch_number) = dal.get_sealed_l1_batch_number().await? else {
```
slowli, line 93: Nit: May make sense to add `context` to DAL calls to understand which call has errored.

core/lib/zksync_core/src/state_keeper/state_keeper_storage.rs
```diff
@@ -0,0 +1,249 @@
...
+        Ok(Some(rocksdb.into()))
+    }
+
+    async fn access_storage_inner<'a>(
+        &'a self,
+        rt_handle: Handle,
+        stop_receiver: &watch::Receiver<bool>,
+    ) -> anyhow::Result<Option<PgOrRocksdbStorage<'a>>> {
+        match self.rocksdb_cell.get() {
+            Some(rocksdb) => {
```
slowli, line 167: Bikshedding: `if let Some(_) = self.rocksdb_cell.get() { .. } else { .. }` is slightly shorter and sometimes considered more idiomatic.

core/lib/zksync_core/src/state_keeper/state_keeper_storage.rs
```diff
@@ -0,0 +1,249 @@
...
+        stop_receiver: watch::Receiver<bool>,
+    ) -> Self {
+        let rocksdb_cell = Arc::new(OnceCell::new());
+        let task = AsyncCatchupTask {
+            pool: pool.clone(),
+            state_keeper_db_path,
+            enum_index_migration_chunk_size,
+            rocksdb_cell: rocksdb_cell.clone(),
+        };
+        tokio::task::spawn(async move { task.run(stop_receiver).await.unwrap() });
```
slowli, line 198: I'd return `AsyncCatchupTask` from this method so that the caller can manage it. E.g., it'd make sense to add it to the list of managed tasks on the main node / EN (rn you'd need to call [`pending().await`](https://doc.rust-lang.org/std/future/fn.pending.html) so that when the task is logically complete, it doesn't bring the node down with it).

itegulov: I used `stop_receiver.0.changed().await?` as that seemed like a slightly better alternative to `pending().await`

core/lib/state/src/utils.rs
itegulov: Yeah, I managed to return RocksDB-related code to almost what it used to be before this PR.

Also, I got rid of all `From` implementations as they didn't fit the official `From` [guidelines](https://doc.rust-lang.org/std/convert/trait.From.html). Props to @popzxc for his Slack post about this the other day

core/lib/zksync_core/src/state_keeper/state_keeper_storage.rs
itegulov: Oh, nice, had no idea it worked like that

RomanBrodetski (approved):
Looks great! Please wait for @slowli 's final approval before merging

core/lib/zksync_core/src/state_keeper/state_keeper_storage.rs
```diff
@@ -0,0 +1,242 @@
...
+use zksync_state::{
+    PostgresStorage, ReadStorage, RocksdbStorage, RocksdbStorageBuilder, StateKeeperColumnFamily,
+};
+use zksync_storage::RocksDB;
+use zksync_types::{L1BatchNumber, MiniblockNumber};
+
+/// Factory that can produce a [`ReadStorage`] implementation on demand.
+#[async_trait]
+pub trait ReadStorageFactory: Debug + Send + Sync + 'static {
+    async fn access_storage(
```
slowli, line 17: I wonder whether it would make sense to have a more defensive interface that takes the executed L1 batch number as an argument. Could be dealt with later, I suppose.

core/node/node_framework/src/implementations/layers/state_keeper/main_batch_executor.rs
```diff
@@ -32,17 +35,35 @@ impl WiringLayer for MainBatchExecutorLayer {
...
-            self.state_keeper_config.enum_index_migration_chunk_size(),
             false,
         );
 
         context.insert_resource(BatchExecutorResource(Unique::new(Box::new(builder))))?;
+        context.add_task(Box::new(RocksdbCatchupTask(task)));
+        Ok(())
+    }
+}
+struct RocksdbCatchupTask(AsyncCatchupTask);
```
slowli, line 56: Nit: Missing `Debug` implementation and an blank line before the type declaration.

slowli (commented):
Looks good; I only have minor nits other than spawning the syncing task.

core/lib/dal/src/blocks_dal.rs
```diff
@@ -2328,6 +2339,7 @@ impl BlocksDal<'_, '_> {
         self.delete_initial_writes_inner(None)
             .await
             .context("delete_initial_writes_inner()")?;
+        self.delete_logs_inner().await.context("no")?;
```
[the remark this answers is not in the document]
itegulov: Good catch, no idea why I wrote "no" lol

core/lib/zksync_core/src/state_keeper/batch_executor/main_executor.rs
```diff
@@ -1,4 +1,4 @@
-use std::sync::Arc;
+use std::{fmt::Debug, sync::Arc};
```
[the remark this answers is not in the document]
itegulov: I imported it because I had to implement `Debug` for some custom type manually in the earlier stages of PR. But good catch that we don't need it anymore.

core/bin/external_node/src/main.rs
```diff
@@ -75,13 +75,18 @@ async fn build_state_keeper(
...
 
-    let batch_executor_base: Box<dyn BatchExecutor> = Box::new(MainBatchExecutor::new(
-        state_keeper_db_path,
+    let (storage_factory, task) = AsyncRocksdbCache::new(
         connection_pool.clone(),
+        state_keeper_db_path,
+        config.optional.enum_index_migration_chunk_size,
+    );
+    let stop_receiver_clone = stop_receiver.clone();
+    tokio::task::spawn(async move { task.run(stop_receiver_clone).await.unwrap() });
```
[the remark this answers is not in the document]
itegulov: Ahh, I didn't realize that's what you meant by "managed task". I assumed only node framework is capable of that. Should be better now!

core/lib/zksync_core/src/state_keeper/state_keeper_storage.rs
itegulov: Hmm, defensive in the sense that it would sanity check that the current state of PG matches the latest executed L1 batch or in a sense that it would force `AsyncRocksdbCache` to cut off state at that specific L1 batch somehow?

core/lib/zksync_core/src/block_reverter/mod.rs
```diff
@@ -221,7 +221,7 @@ impl BlockReverter {
         tracing::info!("opening DB with state keeper cache...");
         let sk_cache = RocksdbStorage::builder(self.state_keeper_cache_path.as_ref())
             .await
-            .expect("Failed initializing state keeper cache");
+            .expect("Failed initializing state keeper");
```
slowli, line 224: Nit: Not sure this change is warranted; the log above talks about the state keeper cache.

core/lib/zksync_core/src/state_keeper/batch_executor/tests/tester.rs
```diff
@@ -82,45 +88,131 @@ impl Tester {
...
+    /// Creates a batch executor instance with the specified storage type.
     /// This function intentionally uses sensible defaults to not introduce boilerplate.
-    pub(super) async fn create_batch_executor(&self) -> BatchExecutorHandle {
-        // Not really important for the batch executor - it operates over a single batch.
-        let (l1_batch_env, system_env) = self.batch_params(L1BatchNumber(1), 100);
-        self.create_batch_executor_inner(l1_batch_env, system_env)
-            .await
+    pub(super) async fn create_batch_executor(
+        &mut self,
+        storage_type: &StorageType,
```
slowli, line 103: Nit: It's unlikely that this type is changed in the future, so you can implement `Copy` for it and pass it by value.

core/lib/zksync_core/src/state_keeper/state_keeper_storage.rs
slowli: The first option, i.e. a sanity check.

slowli (approved):
Very nice.

slowli (approved):

PR #1258, opened 2024-02-27
feat(node_framework): Add Web3 API layers

popzxc:
With some hacks, adds support of our Web3 API to the framework.
For more context on hacks/tradeoffs, see comments in the PR.

With this PR, the node finally passes a significant number of tests. Once the `eth_sender` is migrated, I expect all the tests to pass.

### DIFF
--- a/.github/workflows/vm-perf-comparison.yml
+++ b/.github/workflows/vm-perf-comparison.yml
@@ -28,7 +28,7 @@ jobs:
         run: echo "sha=$(git merge-base HEAD FETCH_HEAD)" >> $GITHUB_OUTPUT
 
       - name: checkout divergence point
-        run: git checkout ${{ steps.merge_base.outputs.sha }}
+        run: git checkout ${{ steps.merge_base.outputs.sha }} --recurse-submodules
 
       - name: setup-env
         run: |


core/lib/zksync_core/src/api_server/web3/mod.rs
```diff
@@ -265,16 +280,11 @@ impl ApiBuilder {
...
-impl ApiBuilder {
-    pub async fn build(
-        self,
-        stop_receiver: watch::Receiver<bool>,
-    ) -> anyhow::Result<ApiServerHandles> {
-        self.into_full_params()?.spawn_server(stop_receiver).await
+impl ApiServer {
+    pub fn health_check(&self) -> ReactiveHealthCheck {
+        self.health_check.clone()
     }
```
slowli, line 286: Storing `health_check` is redundant; it can be instantiated from `health_updater` using `subscribe()` method.

core/lib/zksync_core/src/api_server/web3/mod.rs
```diff
@@ -579,6 +585,8 @@ impl FullApiParams {
         let closing_vm_barrier = vm_barrier.clone();
         let health_updater = Arc::new(health_updater);
```
slowli, line 586: This creates an `Arc<Arc<_>>`, which is probably not what you need. One option is to change `health_updater` field type to `Option<HealthUpdater>` and take it out before consuming `self`. Not ideal, but IMO better than wrapping the field in `Arc`.

core/node/node_framework/examples/main_node.rs
```diff
@@ -126,6 +137,95 @@ impl MainNodeBuilder {
...
+        self.node.add_layer(Web3ServerLayer::http(
+            rpc_config.http_port,
+            InternalApiConfig::new(&network_config, &rpc_config, &contracts_config),
+            optional_config,
+        ));
+
+        Ok(self)
+    }
+
+    fn add_ws_web3_api_layer(mut self) -> anyhow::Result<Self> {
```
slowli, line 196: It looks like with some effort, copy-paste between this method and `add_http_web3_api_layer` could be avoided, but I understand that it's not the focus rn.

core/node/node_framework/src/implementations/layers/web3_api/server.rs
```diff
@@ -0,0 +1,225 @@
...
+/// Set of optional variables that can be altered to modify the behavior of API builder.
+#[derive(Debug, Default)]
+pub struct Web3ServerOptionalConfig {
+    pub namespaces: Option<Vec<Namespace>>,
+    pub filters_limit: Option<usize>,
+    pub subscriptions_limit: Option<usize>,
+    pub batch_request_size_limit: Option<usize>,
+    pub response_body_size_limit: Option<usize>,
+    pub websocket_requests_per_minute_limit: Option<NonZeroU32>,
+    pub tree_api_url: Option<String>,
```
slowli, line 25: Future work (?): It could make sense to make tree API client a resource. Could be done separately.

core/node/node_framework/src/implementations/layers/web3_api/server.rs
```diff
@@ -0,0 +1,225 @@
...
+}
+
+pub struct Web3ServerLayer {
+    transport: Transport,
+    port: u16,
+    internal_api_config: InternalApiConfig,
+    optional_config: Web3ServerOptionalConfig,
+}
+
+impl std::fmt::Debug for Web3ServerLayer {
```
slowli, line 69: Dumb question: What's the problem with deriving `Debug` here?

core/node/node_framework/src/implementations/layers/web3_api/tx_sender.rs
```diff
@@ -0,0 +1,160 @@
...
+        "tx_sender_layer"
+    }
+
+    async fn wire(self: Box<Self>, mut context: ServiceContext<'_>) -> Result<(), WiringError> {
+        // Get required resources.
+        let tx_sink = context.get_resource::<TxSinkResource>().await?.0;
+        let pool_resource = context.get_resource::<ReplicaPoolResource>().await?;
+        let replica_pool = pool_resource.get().await?;
+        let sealer = match context.get_resource::<ConditionalSealerResource>().await {
+            Ok(sealer) => Some(sealer.0.clone()),
```
slowli, line 64: Nit: `clone()` is redundant; `sealer` is returned by value.

core/node/node_framework/src/implementations/layers/web3_api/tx_sender.rs
```diff
@@ -0,0 +1,160 @@
...
+    fn layer_name(&self) -> &'static str {
+        "tx_sender_layer"
+    }
+
+    async fn wire(self: Box<Self>, mut context: ServiceContext<'_>) -> Result<(), WiringError> {
+        // Get required resources.
+        let tx_sink = context.get_resource::<TxSinkResource>().await?.0;
+        let pool_resource = context.get_resource::<ReplicaPoolResource>().await?;
+        let replica_pool = pool_resource.get().await?;
+        let sealer = match context.get_resource::<ConditionalSealerResource>().await {
```
slowli, line 63: Future work (?): May make sense to add `get_optional_resource` method that would return `Result<Option<T>, _>`; seems like a somewhat common pattern.

core/node/node_framework/src/implementations/layers/web3_api/tx_sender.rs
```diff
@@ -0,0 +1,160 @@
...
+}
+
+#[async_trait::async_trait]
+impl Task for PostgresStorageCachesTask {
+    fn name(&self) -> &'static str {
+        "postgres_storage_caches"
+    }
+
+    async fn run(self: Box<Self>, _stop_receiver: StopReceiver) -> anyhow::Result<()> {
+        tokio::task::spawn_blocking(self.task).await?
```
slowli, line 134: Future work (?): May make sense to make this task cancel-aware and maybe have a nameable type.

core/node/node_framework/src/implementations/resources/sync_state.rs
```diff
@@ -0,0 +1,12 @@
+use zksync_core::sync_layer::SyncState;
+
+use crate::resource::{Resource, ResourceId};
+
+#[derive(Debug, Clone)]
+pub struct SyncStateResource(pub SyncState);
+
+impl Resource for SyncStateResource {
+    fn resource_id() -> ResourceId {
+        "sync_state".into()
```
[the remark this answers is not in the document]
slowli: I think this name makes sense; at least, the corresponding health check is named the same 🙃

core/node/node_framework/src/implementations/resources/pools.rs
```diff
@@ -95,10 +95,14 @@ impl ReplicaPoolResource {
     }
 
     pub async fn get_singleton(&self) -> anyhow::Result<ConnectionPool> {
+        self.get_custom(1).await
+    }
+
+    pub async fn get_custom(&self, size: u32) -> anyhow::Result<ConnectionPool> {
```
slowli, line 101: I don't quite understand how this method works. Wouldn't it always build a singleton pool, regardless of the `size` supplied? It seems that it only influences logs.

core/node/node_framework/src/implementations/resources/pools.rs
```diff
@@ -95,10 +95,14 @@ impl ReplicaPoolResource {
...
         let result = self.builder.build_singleton().await;
 
         if result.is_ok() {
-            self.connections_count.fetch_add(1, Ordering::Relaxed);
+            self.connections_count.fetch_add(size, Ordering::Relaxed);
             let total_connections = self.connections_count.load(Ordering::Relaxed);
```
slowli, line 106: Nit: Since `fetch_add` returns the previous value, it could make sense to compute total connections as `fetch_add(_) + size` for atomicity. Not a big deal, obviously.

core/node/node_framework/examples/main_node.rs
popzxc: Yup, this will be worked on separately.

core/node/node_framework/src/implementations/layers/web3_api/server.rs
popzxc: Yup, also thought about it.

core/node/node_framework/src/implementations/layers/web3_api/server.rs
popzxc: Leftover from WIP state. I was experimenting and had some tries to expose the builder that we already have for API, and at that time `Debug` could've not been derived. Then I gave up and went with an extra type for builder fields and forgot to change the implementation.

Thanks for noticing!

core/node/node_framework/src/implementations/layers/web3_api/tx_sender.rs
popzxc: Yeah, I'm also thinking about this. Not sure about it just yet, since such optional resources themselves feel a bit wrong -- maybe the correct approach would be to know for sure whether you need the resource or not. Otherwise, we can create a surface for implicit bugs where some component is missing by mistake, but the task still runs without it, even if that's not what we want.

core/node/node_framework/src/implementations/resources/pools.rs
popzxc: Oh, right. What was I even thinking? 🤦

core/lib/zksync_core/src/api_server/web3/mod.rs
popzxc: Oh, I just forgot to remove this line. Of course, since we already have an `Arc`, we don't need to wrap it twice.
I'm not sure about `Option` though -- it would add a bunch of `unwrap`s, e.g. in places where we subscribe to get a health check object.

popzxc:
Well, that's unexpected 😅 
image

core/node/node_framework/src/implementations/layers/web3_api/tx_sender.rs
popzxc: Makes sense, will tackle it in a separate PR.

slowli (approved):

core/lib/zksync_core/src/api_server/web3/mod.rs
```diff
@@ -244,15 +248,24 @@ impl ApiBuilder {
...
+}
+
+impl ApiBuilder {
+    pub fn build(self) -> anyhow::Result<ApiServer> {
+        let transport = self.transport.context("API transport not set")?;
+        let health_check_name = match &transport {
+            ApiTransport::Http(_) => "http_api",
+            ApiTransport::WebSocket(_) => "ws_api",
+        };
+        let (_health_check, health_updater) = ReactiveHealthCheck::new(health_check_name);
```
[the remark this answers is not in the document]
popzxc: Will do in one of the follow-ups.

hatemosphere (approved):

AnastasiiaVashchuk:
LGTM!

PR #1261, opened 2024-02-27
feat(config): add pubdata sending method to config

koloz193:
## What ❔

Add the pubdata sending method to eth sender config

## Why ❔

We want an easy way for indicating to the server what DA layer should be used for pubdata

## Checklist

### DIFF
--- a/core/lib/protobuf_config/src/eth_sender.rs
+++ b/core/lib/protobuf_config/src/eth_sender.rs
@@ -1,5 +1,5 @@
 use anyhow::Context as _;
-use zksync_config::configs;
+use zksync_config::configs::{self, eth_sender::PubdataSendingMode};
 use zksync_protobuf::required;
 
 use crate::{
@@ -109,6 +109,7 @@ impl ProtoRepr for proto::Sender {
                 .and_then(|x| Ok(proto::ProofLoadingMode::try_from(*x)?))
                 .context("proof_loading_mode")?
                 .parse(),
+            pubdata_sending_mode: PubdataSendingMode::Calldata,
         })
     }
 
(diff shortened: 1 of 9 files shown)


etc/env/base/chain.toml
```diff
@@ -65,8 +65,11 @@ batch_overhead_l1_gas=800000
 # The maximum amount of gas that can be used by the batch. This value is derived from the circuits limitation per batch.
 max_gas_per_batch=200000000
 
-# The maximum amount of pubdata that can be used by the batch. Note that if the calldata is used as pubdata, this variable should not exceed 128kb.
-max_pubdata_per_batch=100000
+# This will also determine how many blobs are used, when enabled.
+# If the value is less than 126kb -> Calldata or 1 Blob
+# If the value is > 126 kb and < 252kb then 2 blobs are used.
+# Note, the max at this moment is 252kb
+max_pubdata_per_batch=110000
```
perekopskiy, line 72: Let's keep it 100k for now, and change later. We use 100k for all our envs including mainnet for now. Changing this variable affects fees, so let's do it separately

core/lib/protobuf_config/src/eth_sender.rs
```diff
@@ -109,6 +109,7 @@ impl ProtoRepr for proto::Sender {
                 .and_then(|x| Ok(proto::ProofLoadingMode::try_from(*x)?))
                 .context("proof_loading_mode")?
                 .parse(),
+            pubdata_sending_mode: PubdataSendingMode::Calldata,
```
perekopskiy, line 112: Should be loaded from protobuf config

koloz193: agreed, if we're okay with the enum then ill raise the pr in consensus repo

montekki (approved):

perekopskiy (approved):

PR #1268, opened 2024-02-28
feat: zk init rework

benceharomi:


infrastructure/zk/src/config.ts
```diff
@@ -151,7 +151,7 @@ export function pushConfig(environment?: string, diff?: string) {
         false
     );
 
-    env.modify('DATABASE_URL', `postgres://postgres@localhost/${environment}`, l2InitFile, false);
+    env.modify('DATABASE_URL', `postgres://postgres:notsecurepassword@localhost/${environment}`, l2InitFile, false);
```
ly0va, line 154: We should probably add the same `env.modify` for `DATABASE_PROVER_URL` and `TEST_DATABASE_PROVER_URL`

ly0va (commented):
Why did most normal functions were rewritten as arrow functions?

infrastructure/zk/src/config.ts
benceharomi: Good idea. Added!

benceharomi:
> Why did most normal functions were rewritten as arrow functions?

Arrow and normal functions were mixedly used, so I rewritten all of them to be arrow functions in that file so they would be the same. It is just a personal preference, we can also go with normal functions, but then I would change all of them to be normal functions.

ly0va (approved):
:+1:

PR #1270, opened 2024-02-28
fix(Validium): Fix broken integration test

SantiagoPittella:
Related to: #107 

To be correctly compatible with legacy transactions, the effective gas calculation should change and follow the specification in:
https://eips.ethereum.org/EIPS/eip-1559

## What ❔

- Updates an integration test to expect a lower or equal gas price instead of an equal gas price

## Why ❔

- [A test](https://github.com/lambdaclass/zksync-era/blob/fix-effective-gas-price-calculation/core/tests/ts-integration/tests/api/web3.test.ts#L220) that checks the gas price of a transaction (sent vs received in the receipt) was failing in Validium mode.

## Checklist


core/lib/types/src/fee.rs
```diff
@@ -74,8 +74,14 @@ impl Fee {
...
         assert!(self.max_priority_fee_per_gas <= self.max_fee_per_gas);
 
-        // For now, we charge only for base fee.
-        block_base_fee_per_gas
+        // The following calculation is based on the EIP-1559 formula:
+        // https://eips.ethereum.org/EIPS/eip-1559
+        let priority_fee_per_gas = std::cmp::min(
+            self.max_priority_fee_per_gas,
+            self.max_fee_per_gas - block_base_fee_per_gas,
+        );
```
popzxc, line 82: Are you sure it's a correct fix? There are other places where we also calculate `effective_gas_price`, e.g. [here](https://github.com/lambdaclass/zksync-era/blob/fix-effective-gas-price-calculation/core/lib/multivm/src/versions/vm_boojum_integration/tracers/refunds.rs#L120). It feels a bit suspicious that you only change a single place for that.

cc @StanislavBreadless @perekopskiy -- do you guys have any feedback on this?

core/lib/types/src/fee.rs
```diff
@@ -74,8 +74,14 @@ impl Fee {
         assert!(block_base_fee_per_gas <= self.max_fee_per_gas);
         assert!(self.max_priority_fee_per_gas <= self.max_fee_per_gas);
 
-        // For now, we charge only for base fee.
-        block_base_fee_per_gas
```
perekopskiy, line 78: Current fee model used is not EIP-1559 compliant. Priority fee must be ignored.

ilitteri:
I dove deeper into this, I think this fix could not necessarily be correct even if it makes the test pass. 

For Validium we're setting `compute_overhead_part` to `1.0` as it is suggested here. This affects the `fair_l2_gas_price` computation as now [we're adding `gas_overhead_wei` to it](https://github.com/matter-labs/zksync-era/blob/main/core/lib/zksync_core/src/fee_model.rs#L190) (if I remove that variable from the sum, the test passes).

With the above context, now focusing on the test itself, after discarding that this is exclusively a Legacy Tx issue (comment the line that overrides the type and run the test to see this), I concluded that there's an incompatibility between the `gasPrice` from the transaction being sent and the transaction retrieved by the Web3 API. The endpoint gets the transaction from the database implying that the transaction being sent is not exactly the same as the one stored in the database.

Do you have any idea of what could be happening here? Should we ignore this test in Validium? @popzxc  @perekopskiy  @StanislavBreadless

StanislavBreadless:
@ilitteri The reason for that is since `eth_getGasPrice` will return not exact gas price now, but rather a one with a multiplier (to account for the fact that the L1 gas price may rise in the near term), the `gasPrice` which is sent by your tx in the test is necessarily larger than the effective one (i.e. the actual one that did not have the multiplier). 

My recommendation would be probably to remove the multiplier from the `eth_getGasPrice` estimation (at a risk having users sending txs that get temporarily stuck in the mempool in case of spikes) or, alternatively amend the test to allow gas price to be <= to the one provided in the tx. We do ignore priority fee and it is a common practice among L2s.

ilitteri:
> @ilitteri The reason for that is since `eth_getGasPrice` will return not exact gas price now, but rather a one with a multiplier (to account for the fact that the L1 gas price may rise in the near term), the `gasPrice` which is sent by your tx in the test is necessarily larger than the effective one (i.e. the actual one that did not have the multiplier).
> 
> My recommendation would be probably to remove the multiplier from the `eth_getGasPrice` estimation (at a risk having users sending txs that get temporarily stuck in the mempool in case of spikes) or, alternatively amend the test to allow gas price to be <= to the one provided in the tx. We do ignore priority fee and it is a common practice among L2s.

Thanks for the explanation! We've proceeded on updating the test so the expected `gasPrice` is lower or equal to the tx's

ilitteri:
@popzxc if this new change is ok, I think we're done here (the linter and unit-tests are fixed in https://github.com/matter-labs/zksync-era/pull/1312)

popzxc (approved):

PR #1277, opened 2024-02-28
feat(Validium): add a Validium config file for external node 

toni-calvin:
At this point the external node uses a `consistency_checker` that has a field `l1_batch_commit_data_generator`. Depending on which mode is run this field would be set to `Validium` or `Rollup`. 

We have added a `ext-node-validium.toml` that uses `ext-node.toml` config as a base and adds the field `l1_batch_commit_data_generator` in Validium. By default the field is in Rollup so nothing will break in the already defined config.


popzxc (approved):
👍

PR #1281, opened 2024-02-28
feat(Gas Oracle): Use normal anyhow::Error for the `NativeErc20Fetcher`

juan518munoz:
## What ❔

Remove `Error` struct that holds `anyhow::Error` inside.

## Why ❔

Keep error structure clean.

## Checklist

### DIFF
--- a/core/lib/zksync_core/src/l1_gas_price/singleton.rs
+++ b/core/lib/zksync_core/src/l1_gas_price/singleton.rs
@@ -16,18 +16,8 @@ use crate::{l1_gas_price::GasAdjuster, native_token_fetcher::ConversionRateFetch
 pub struct GasAdjusterSingleton {
     web3_url: String,
     gas_adjuster_config: GasAdjusterConfig,
-    singleton: OnceCell<Result<Arc<GasAdjuster<QueryClient>>, Error>>,
+    singleton: OnceCell<anyhow::Result<Arc<GasAdjuster<QueryClient>>>>,
     erc20_fetcher_dyn: Option<Arc<dyn ConversionRateFetcher>>,
-}
-
-#[derive(thiserror::Error, Debug, Clone)]
-#[error(transparent)]
-pub struct Error(Arc<anyhow::Error>);
-
-impl From<anyhow::Error> for Error {
-    fn from(err: anyhow::Error) -> Self {
-        Self(Arc::new(err))
-    }
 }
 
 impl GasAdjusterSingleton {
@@ -44,8 +34,8 @@ impl GasAdjusterSingleton {
         }
     }
 
-    pub async fn get_or_init(&mut self) -> Result<Arc<GasAdjuster<QueryClient>>, Error> {
-        let adjuster = self
+    pub async fn get_or_init(&mut self) -> anyhow::Result<Arc<GasAdjuster<QueryClient>>> {
+        match self
             .singleton
             .get_or_init(|| async {
                 let query_client =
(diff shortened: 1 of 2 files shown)


core/lib/zksync_core/src/l1_gas_price/singleton.rs
```diff
@@ -59,17 +49,23 @@ impl GasAdjusterSingleton {
                 .context("GasAdjuster::new()")?;
                 Ok(Arc::new(adjuster))
             })
-            .await;
-        adjuster.clone()
+            .await
+        {
+            Ok(adjuster) => Ok(adjuster.clone()),
+            Err(_e) => Err(anyhow::anyhow!("Failed to get or initialize GasAdjuster")),
+        }
```
popzxc, line 56: Can't you use `?` in here?

core/lib/zksync_core/src/native_token_fetcher/mod.rs
```diff
@@ -39,24 +27,32 @@ impl NativeTokenFetcherSingleton {
...
             })
-            .await;
-        adjuster.clone()
+            .await
+        {
+            Ok(fetcher) => Ok(fetcher.clone()),
+            Err(_e) => Err(anyhow::anyhow!(
+                "Failed to get or initialize NativeTokenFetcher"
+            )),
+        }
```
popzxc, line 44: Same.

core/lib/zksync_core/src/l1_gas_price/singleton.rs
```diff
@@ -59,17 +49,23 @@ impl GasAdjusterSingleton {
...
 
     pub fn run_if_initialized(
         self,
         stop_signal: watch::Receiver<bool>,
     ) -> Option<JoinHandle<anyhow::Result<()>>> {
-        let gas_adjuster = self.singleton.get()?.clone();
+        let gas_adjuster = match self.singleton.get()? {
+            Ok(gas_adjuster) => gas_adjuster.clone(),
+            Err(_e) => return None,
+        };
```
popzxc, line 66: Nit: I guess you can do it as `self.singleton.get()?.ok()?.clone()`.

core/lib/zksync_core/src/l1_gas_price/singleton.rs
juan518munoz: Replacing the `match` with the `?` operator results in the following compilation error:
```
the `?` operator can only be applied to values that implement `Try`
the trait `Try` is not implemented for `&std::result::Result<Arc<GasAdjuster<QueryClient>>, anyhow::Error>`
the trait `Try` is implemented for `std::result::Result<T, E>`
```

The output of `get_or_init()` is not a `Result`, but a reference (`&`) to it. The simplest solution would be to `.clone()` the reference to get it's inner value, but this is not possible, because `anyhow::Error` is not cloneable, so when calling `clone()` on it, instead of getting the contents of the reference, we get the same reference again.

Let me know if I explained something poorly.

core/lib/zksync_core/src/native_token_fetcher/mod.rs
juan518munoz: Same answer as above.

core/lib/zksync_core/src/l1_gas_price/singleton.rs
juan518munoz: Similar to `get_or_init()`, the `get()` method result can't be dereferenced because the `anyhow::Error` is not cloneable.

popzxc (approved):

PR #1282, opened 2024-02-28
feat(Gas Oracle): Improve Fetcher Error Handling

juan518munoz:
## What ❔

Replace `unwraps()` with better error handling.

## Why ❔

Upon failure, current request failures will bring down the whole system.

## Checklist

### DIFF
--- a/core/lib/zksync_core/src/native_token_fetcher/mod.rs
+++ b/core/lib/zksync_core/src/native_token_fetcher/mod.rs
@@ -1,4 +1,4 @@
-use std::{sync::Arc, time::Duration};
+use std::{cmp::min, sync::Arc, time::Duration};
 
 use async_trait::async_trait;
 use metrics::atomics::AtomicU64;
@@ -8,7 +8,36 @@ use tokio::{
 };
 use zksync_config::configs::native_token_fetcher::NativeTokenFetcherConfig;
 
-const MAX_CONSECUTIVE_NETWORK_ERRORS: u8 = 10;
+#[derive(Debug)]
+struct ErrorReporter {
+    current_try: u8,
+    alert_spawned: bool,
+}
+
+impl ErrorReporter {
+    const MAX_CONSECUTIVE_NETWORK_ERRORS: u8 = 10;
+
+    fn new() -> Self {
+        Self {
+            current_try: 0,
+            alert_spawned: false,
+        }
+    }
+
+    fn reset(&mut self) {
+        self.current_try = 0;
+        self.alert_spawned = false;
+    }
+
+    fn process(&mut self, err: anyhow::Error) {
+        self.current_try = min(self.current_try + 1, Self::MAX_CONSECUTIVE_NETWORK_ERRORS);
+        tracing::error!("Failed to fetch native token conversion rate from the server: {err}");
+        if self.current_try >= Self::MAX_CONSECUTIVE_NETWORK_ERRORS && !self.alert_spawned {
+            vlog::capture_message(&err.to_string(), vlog::AlertLevel::Warning);
+            self.alert_spawned = true;
+        }
+    }
+}
 
 /// Trait used to query the stack's native token conversion rate. Used to properly
 /// determine gas prices, as they partially depend on L1 gas prices, denominated in `eth`.
@@ -49,9 +78,9 @@ impl NativeTokenFetcherSingleton {
         match self
             .singleton
             .get_or_init(|| async {
-                let fetcher =
-                    NativeTokenFetcher::new(self.native_token_fetcher_config.clone()).await;
-                Ok(Arc::new(fetcher))
+                Ok(Arc::new(
+                    NativeTokenFetcher::new(self.native_token_fetcher_config.clone()).await?,
+                ))
             })
             .await
         {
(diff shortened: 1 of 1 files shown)


core/lib/zksync_core/src/native_token_fetcher/mod.rs
```diff
@@ -84,20 +89,35 @@ impl NativeTokenFetcher {
...
+
+                    tracing::error!(
+                        "Failed to fetch native token conversion rate from the server: {err}"
+                    );
+
+                    if network_consecutive_errors >= MAX_CONSECUTIVE_NETWORK_ERRORS {
+                        vlog::capture_message(&err.to_string(), vlog::AlertLevel::Warning);
+
+                        // reset the error counter to prevent sending multiple sentry errors consecutively
+                        network_consecutive_errors = 0;
```
popzxc, line 117: Nice! However, if the polling interval is set to 3 seconds, we would still get 2 alerts per minute, which is probably not a good idea.
Additionally, the retry logic seems to be mixed with the actual application logic.
Maybe we declare a helper structure that would encapsulate it. Something like this (pseudocode/not prettified):

```rust
struct ErrorReporter {
  current_try: u64,
  alert_spawned: bool,
}

impl ErrorReporter {
  const MAX_CONSECUTIVE_NETWORK_ERRORS: u8 = 10;
  fn reset(&mut self) {
    self.current_try = 0;
    self.alert_spawned = false;
  }
  fn process(&mut self, err: Error) {
    self.current_try += std::cmp::min(self.current_try + 1, Self::MAX_CONSECUTIVE_NETWORK_ERRORS);  
    tracing::error!("Failed to fetch native token conversion rate from the server: {err}");
    if self.current_try >= Self::MAX_CONSECUTIVE_NETWORK_ERRORS && !self.alert_spawned {
        vlog::capture_message(&err.to_string(), vlog::AlertLevel::Warning);
        self.alert_spawned = true;
    }
  }
}
```

popzxc: This way we will only spawn a single alert per server going down. We will still, however, spawn multiple alerts if the server goes up and down, but that's probably a good thing: otherwise someone may check that in this exact moment "huh, it's up right now, probably it's fine".

core/lib/zksync_core/src/native_token_fetcher/mod.rs
```diff
@@ -70,12 +72,15 @@ pub(crate) struct NativeTokenFetcher {
...
-            .unwrap();
+        let conversion_rate = match reqwest::get(format!("{}/conversion_rate", config.host)).await {
+            Ok(response) => response.json::<u64>().await.unwrap_or(1), // Prevent program from crashing altogether if the first request fails
+            Err(err) => {
+                tracing::error!(
+                    "Failed to fetch native token conversion rate from the server: {err}"
+                );
+                1
+            }
+        };
```
popzxc, line 83: I'm not 100% sure about that. When I wrote about [running with unsynchronized ratios](https://github.com/matter-labs/zksync-era/pull/996#discussion_r1503677218) I was thinking about using some outdated, but still real value (e.g. persisting the history of fetched ratios) -- which is why I wrote that we can tackle it separately, since it's a big topic.

What you proposed in the same discussion -- panic (or, even better, return `anyhow::Result<Self>`) if the reliable rate cannot be fetcher for the first time -- seems to be a better short-term approach.

core/lib/zksync_core/src/native_token_fetcher/mod.rs
juan518munoz: I agree, moving the error handling to a separate structure keeps code clean. I have two questions about your pseudocode:

1. In what case should we `reset()` the `ErrorReporter`? 
2. Should we keep this struct inside the same file? Or move it so it can be ussed by multiple components?

core/lib/zksync_core/src/native_token_fetcher/mod.rs
juan518munoz: Maybe we can just `.process()` this possible initial with the new `ErrorReporter` structure discussed above.

core/lib/zksync_core/src/native_token_fetcher/mod.rs
juan518munoz: I've implemented a simple `ErrorReporter`, be sure to check it out whenever you can.

core/lib/zksync_core/src/native_token_fetcher/mod.rs
```diff
@@ -82,41 +113,45 @@ pub(crate) struct NativeTokenFetcher {
...
-                .await?
-                .json::<u64>()
                 .await
-                .unwrap();
-
-            self.latest_to_eth_conversion_rate
-                .store(conversion_rate, std::sync::atomic::Ordering::Relaxed);
+            {
+                Ok(response) => {
+                    let conversion_rate = response.json::<u64>().await?;
```
popzxc, line 148: I guess it's fair to exit if the response is 200 but cannot be deserialized, but if done as implemented, the error will likely be pretty confusing and hard to analyze. Please add [`context`](https://docs.rs/anyhow/latest/anyhow/trait.Context.html#tymethod.context) here, e.g. `.await.context("Unable to parse the response of the native token conversion rate server")?`.

core/lib/zksync_core/src/native_token_fetcher/mod.rs
```diff
@@ -8,6 +8,37 @@ use tokio::{
 };
 use zksync_config::configs::native_token_fetcher::NativeTokenFetcherConfig;
 
+#[derive(Debug)]
```
popzxc, line 11: It's a very nitpicky thing, but maybe we could move this structure to the bottom of the file?
We normally (though surely not always) structure files in a way that represents the "important" stuff at the top and "less important" stuff at the bottom.

Though it's nothing significant, so feel free to leave as is if you don't feel like it.

core/lib/zksync_core/src/native_token_fetcher/mod.rs
juan518munoz: Addressed this and comment bellow in latest commit.

popzxc (approved):
Nice!

PR #1284, opened 2024-02-29
feat(state-keeper): miniblock max payload size (BFT-417)

moshababo:
## What ❔

Adding miniblock sealing according to max payload size.

## Why ❔

Compliance with max payload size imposed in consensus modules.

## Notes
* Building the payload incrementally isn’t feasible because the miniblock’s transactions are flushed to DB before being reconstructed to be encoded as the payload. So the payload size accumulator added to `MiniblockUpdates` is just a heuristic done by encoding every executed transaction only to find out the estimated encoding size.
* `miniblock_max_payload_size` was added to `StateKeeperConfig`, decoupled from `max_payload_size` in `consensus::config::Config`. It is possible to only use the latter, or to make the latter define/override the former. The state keeper max size should be less than 1/2 the consensus max size, because it never rejects the latest transaction, and doesn’t account for some metadata fields. Actual value in that respect was not set yet, because I wasn’t sure about values used for different non-test environments, or if we even want to keep this approach. 
* `MiniblockMaxPayloadSizeSealer` was added, implementing `should_seal_miniblock()` alone (without `should_seal_l1_batch_unconditionally`, as in trait `IoSealCriteria`). This isn’t an ideal solution, so other ideas are welcome. 

The applied config are: 
* Statekeeper threshold: 1mb
* Consensus threshold: 2.5mb

---

Closing [BFT-417](https://linear.app/matterlabs/issue/BFT-417/limit-l2-block-size)

## Checklist


core/lib/protobuf_config/src/proto/chain.proto
```diff
@@ -29,28 +29,29 @@ message StateKeeper {
...
-  optional uint64 max_gas_per_batch = 18; // required; wei?
-  optional uint64 max_pubdata_per_batch = 19; // required; bytes?
-  optional FeeModelVersion fee_model_version = 20; // required
-  optional uint32 validation_computational_gas_limit = 21; // required; wei?
-  optional bool save_call_traces = 22; // required
-  optional uint32 virtual_blocks_interval = 23; // required
-  optional uint32 virtual_blocks_per_miniblock = 24; // required
-  optional bool upload_witness_inputs_to_gcs = 25; // required
-  optional uint64 enum_index_migration_chunk_size = 26; // optional
+  optional uint64 miniblock_max_payload_size = 5; // required
```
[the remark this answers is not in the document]
moshababo: Fixed.

core/lib/zksync_core/src/state_keeper/seal_criteria/mod.rs
```diff
@@ -177,6 +177,29 @@ impl IoSealCriteria for TimeoutSealer {
...
+    }
+}
+
+impl IoSealCriteria for MiniblockMaxPayloadSizeSealer {
+    fn should_seal_l1_batch_unconditionally(&mut self, _manager: &UpdatesManager) -> bool {
+        false
+    }
+
+    fn should_seal_miniblock(&mut self, manager: &UpdatesManager) -> bool {
+        manager.miniblock.payload_encoding_size >= self.max_payload_size
```
slowli, line 199: Just to understand better: Are we OK with the situation where the resulting block size significantly exceeds the desired one? You've written about this in the PR description (IMO, it'd make sense to add the corresponding doc comment on `MiniblockMaxPayloadSizeSealer` so that it's not lost), but I'm not entirely sure what prevents a situation when the miniblock size exceeds the consensus threshold.

core/lib/zksync_core/src/state_keeper/seal_criteria/mod.rs
```diff
@@ -177,6 +177,29 @@ impl IoSealCriteria for TimeoutSealer {
...
+        Self {
+            max_payload_size: config.miniblock_max_payload_size,
+        }
+    }
+}
+
+impl IoSealCriteria for MiniblockMaxPayloadSizeSealer {
+    fn should_seal_l1_batch_unconditionally(&mut self, _manager: &UpdatesManager) -> bool {
+        false
+    }
```
slowli, line 196: Since this method doesn't really make sense for `MiniblockMaxPayloadSizeSealer`, I'd suggest to implement `should_seal_miniblock()` as an inherent method instead.

core/lib/zksync_core/src/state_keeper/updates/miniblock_updates.rs
```diff
@@ -130,6 +132,8 @@ impl MiniblockUpdates {
         self.l1_gas_count += tx_l1_gas_this_tx;
         self.block_execution_metrics += execution_metrics;
         self.txs_encoding_size += tx.bootloader_encoding_size();
+        self.payload_encoding_size +=
+            zksync_protobuf::repr::encode::<zksync_dal::models::proto::Transaction>(&tx).len();
```
slowli, line 136: Dumb question: All executed transactions are preserved in `executed_transactions` (`TransactionExecutionResult` has a `transaction: Transaction` field). Wouldn't it be possible to reconstruct the payload size based on this info instead of introducing an extra field? Or would that be too inefficient?

core/lib/zksync_core/src/state_keeper/updates/miniblock_updates.rs
```diff
@@ -26,6 +26,7 @@ pub struct MiniblockUpdates {
     pub l1_gas_count: BlockGasCount,
     pub block_execution_metrics: ExecutionMetrics,
     pub txs_encoding_size: usize,
+    pub payload_encoding_size: usize,
```
slowli, line 29: Future work (?): Could make sense to observe the payload size estimate as a histogram to have stats on it readily available. OTOH, at least for analytics it could be reconstructed later.

slowli (commented):
Future work: It would probably make sense to somehow de-prioritize / space out large transactions so that they don't significantly increase latency of processing other transactions. (Miniblocks must have distinct timestamps, so if a large transaction causes a miniblock to be sealed early, the latency of the transactions following it in the mempool would suffer.)

core/lib/zksync_core/src/state_keeper/seal_criteria/mod.rs
moshababo: Do you mean as a regular method implemented for `MempoolIO`?

core/lib/zksync_core/src/state_keeper/updates/miniblock_updates.rs
moshababo: Yes, it's possible, but inefficient because it will require to encode all transactions each time, instead of just the appended transaction. So it's just a typical computation-storage tradeoff.

core/lib/zksync_core/src/state_keeper/updates/miniblock_updates.rs
moshababo: What's the best way to check these kind of changes on local dev env?

core/lib/zksync_core/src/state_keeper/seal_criteria/mod.rs
moshababo: This approach assumes that a miniblock with a single transaction won't exceed the consensus threshold. I'm not sure if `L1`/`ProtocolUpgrade` tx types have a max size limit (beyond EVM constrains), but they cannot be rejected as single transactions anyway. `L2` can be rejected, but have a limit in API level. 

Then, given a single transaction in a miniblock, the problematic case is when it's close but below the state keeper threshold, and then a large transaction is added. For that, there should be a significant gap between the state keeper config, and the actual consensus threshold config, which should also account for additional fields in payload, and potential errors in estimating the encoded payload size. 

A different approach might be to support the exclusion of a secondary transaction in a miniblock, to be rolled over to the next one. This will allow to reduce the gap between the state keeper config and the consensus threshold, but it adds some complexity, still must accept a single transaction, and anyway won't eliminate the need to keep some gap, because it's currently not feasible to build the payload incrementally, and so the size value is just a heuristic.

core/lib/zksync_core/src/state_keeper/updates/miniblock_updates.rs
slowli: You can spin up Prometheus and Grafana locally, e.g. using `docker compose`, and check changes in Grafana. I think we have a config [on Notion](https://www.notion.so/matterlabs/Running-Grafana-Prometheus-locally-8770d8a67c4d43a8a65dd8fe2289ad09).

core/lib/zksync_core/src/state_keeper/seal_criteria/mod.rs
slowli: A regular method in `MiniblockMaxPayloadSizeSealer`.

moshababo: Fixed.

etc/env/base/chain.toml
```diff
@@ -20,6 +20,7 @@ max_allowed_l2_tx_gas_limit=4000000000
 block_commit_deadline_ms=2500
 miniblock_commit_deadline_ms=1000
 miniblock_seal_queue_capacity=10
+miniblock_max_payload_size=5000000
```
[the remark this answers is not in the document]
moshababo: Yes it was just a temp setup. I've now reduced it to 40% of the consensus threshold.

moshababo:
The newly applied config are:
* Statekeeper threshold: 1mb
* Consensus threshold: 2.5mb

core/lib/dal/src/lib.rs
```diff
@@ -47,7 +47,7 @@ pub mod fri_witness_generator_dal;
 pub mod healthcheck;
 mod instrument;
 mod metrics;
-mod models;
+pub mod models;
```
popzxc, line 50: This doesn't feel right. DAL models are majorly unsafe and leaking them publicly sounds like a bad idea.
If I understand correctly, proto models, on the other hand, are OK to be used publicly? 

If so, maybe it makes sense to move them out of `models` to their own top-level module (`proto_models`?) and make it public instead. If proto models are not meant to generally be public as well, we probably need to expose functionality related to encoding sizes somehow else.

cc @pompon0

popzxc (commented):
Generally LGTM except for public models module.

core/lib/dal/src/lib.rs
pompon0: In this particular case the Payload struct (not the autogenerated proto code) is part of the public api of consensus_dal. The data of the payload (~miniblock) is stored in the db in a format unrelated to protobuf. The proto serialization is used in consensus, so it is part of the public API. In consensus_dal, we use protobuf encoding for hash validation in insert_certificate.

popzxc: Then probably it makes sense to create a module for actually publicly exposed API

moshababo: @popzxc @pompon0 I've extracted a top-level `consensus` module within `dal`. If later there would be additional public proto/Rust definitions for other domains, it can be rearranged.

core/lib/protobuf_config/src/proto/chain.proto
```diff
@@ -33,6 +33,7 @@ message StateKeeper {
   optional bool save_call_traces = 22; // required
   optional uint64 enum_index_migration_chunk_size = 26; // optional
   optional uint64 max_circuits_per_batch = 27; // required
+  optional uint64 miniblock_max_payload_size = 28; // required
```
pompon0, line 36: I don't remember if we already talked about this: have you already added to configs in all envs? If not, it cannot be required yet.

pompon0 (approved):

core/lib/protobuf_config/src/proto/chain.proto
moshababo: Added here: https://github.com/matter-labs/gitops-kubernetes/pull/4871

pompon0 (approved):

PR #1285, opened 2024-02-29
feat(eth_sender): set blob gas fee when blobs DA is used

koloz193:
## Checklist


core/lib/zksync_core/src/eth_sender/eth_tx_manager.rs
```diff
@@ -233,13 +233,16 @@ impl EthTxManager {
             .used_priority_fee_per_gas
             .observe(priority_fee_per_gas);
 
+        let blob_gas_price = if tx.blob_sidecar.is_some() {
+            Some(U256::from_big_endian(
+                &blob_base_fee_per_gas.unwrap_or(1u64).to_be_bytes(),
```
perekopskiy, line 238: Do you think it's better to use 1 as default instead of panicking?

koloz193: It depends, we should always have the price but if we don’t it’s likely because the value has dropped all the way down or it was just launched and 1 is the absolute min. 

A potential bad scenario would be we just don’t have the price and so doubling until we got to what mainnet expects would take a while. 

@montekki what do you think?

perekopskiy: Hmm, I don't agree. We should always have `blob_gas_price` provided to this function (obviously if it's needed, i.e. blob is sent with a tx). `GasAdjuster` is ready from the start to provide this price. So basically missing price at line 238 is a bug and I think it's better to panic right away instead of silently sending tx with base fee 1. Each tx resend forces us to double priority fee, so starting with blob_base_fee = 1 is a bad idea

montekki: It is expected to be set and it should panic in case it is not. If it is not set then we have a fatal bug and should not proceed

perekopskiy (approved):

PR #1286, opened 2024-02-29
refactor(db): Use `sqlx` query templates

slowli:
## What ❔

Introduces `sqlx` query templates using the fact that sqlx queries support concatenating query literals using `+`.

## Why ❔

Static queries perform more type checks and the output rows are more typesafe / easier to use.

## Checklist


core/lib/web3_decl/src/namespaces/eth.rs
```diff
@@ -87,7 +87,10 @@ pub trait EthNamespace {
     ) -> RpcResult<Option<U256>>;
 
     #[method(name = "getBlockReceipts")]
-    async fn get_block_receipts(&self, block_id: BlockId) -> RpcResult<Vec<TransactionReceipt>>;
+    async fn get_block_receipts(
+        &self,
+        block_id: BlockId,
+    ) -> RpcResult<Option<Vec<TransactionReceipt>>>;
```
slowli, line 93: AFAIU, "canonical" Ethereum Web3 implementations return `null` for non-existing blocks, so I've changed this method to match this behavior.

core/lib/zksync_core/src/api_server/web3/namespaces/eth.rs
```diff
@@ -517,16 +550,40 @@ impl EthNamespace {
...
+                let Ok(idx) = u32::try_from(idx) else {
+                    return Ok(None); // index overflow means no transaction
+                };
+                let Some(block_number) = self
+                    .state
+                    .resolve_block_unchecked(&mut storage, block_id, METHOD_NAME)
+                    .await
+                    .map_err(|err| internal_error(METHOD_NAME, err))?
+                else {
+                    return Ok(None);
```
slowli, line 577: I wasn't particularly attentive with measuring method latency on early returns, since #1203 will make it obsolete anyway.

core/lib/zksync_core/src/api_server/web3/state.rs
```diff
@@ -240,6 +241,33 @@ impl RpcState {
...
     }
 
+    /// Resolves the specified block ID to a block number, which is **not** guaranteed to be present in the node storage.
+    /// Returns `None` if the block is known to not be present in the storage (e.g., it's a "finalized" block ID and no blocks
+    /// were finalized yet).
+    ///
+    /// This method is more efficient than [`Self::resolve_block()`] (it doesn't query the storage if block ID maps to a known
+    /// block number), but is more difficult to reason about. You should use it only if the block number consumer correctly handles
+    /// non-existing blocks.
+    pub(crate) async fn resolve_block_unchecked(
```
slowli, line 251: Future work: block resolution caching could be put somewhere here.

slowli (commented):
Questions up to discussion:

- Is this approach desirable in principle? IMO, type safety it provides is better than the lack of it, but the introduced macro can take some time to get into and maintain. (I should probably document how it works in more detail.)
- Does it make sense to use this approach in `BlocksWeb3Dal` / `TransactionsWeb3Dal`? This increases the number of DB queries for the affected methods. E.g., getting an API block with transactions can now require 3 requests instead of one (resolve block ID to block number; get block header; get transactions). OTOH, it creates more opportunities to cache query outputs (obviously, I didn't do this in this PR, since it's large as is).
- The approach has limitations. E.g., I haven't managed to come up with a feasible way to replace [dynamic queries](https://github.com/matter-labs/zksync-era/blob/533f0131342dc40eea1bb82334828d0007043d19/core/lib/dal/src/events_web3_dal.rs#L38) in `EventsWeb3Dal` because they have too many dynamic parts (4 topic fields + address field with 2 variants for each; i.e., 32 variants in total). AFAICT, that's the only remaining DAL struct with dynamic queries, other than prover-related stuff [like this](https://github.com/matter-labs/zksync-era/blob/533f0131342dc40eea1bb82334828d0007043d19/core/lib/dal/src/fri_proof_compressor_dal.rs#L243), [this system query](https://github.com/matter-labs/zksync-era/blob/533f0131342dc40eea1bb82334828d0007043d19/core/lib/dal/src/system_dal.rs#L15) and [storage log queries w/ custom tables](https://github.com/matter-labs/zksync-era/blob/533f0131342dc40eea1bb82334828d0007043d19/core/lib/dal/src/storage_logs_dal.rs#L778) (can they be removed? cc @perekopskiy).
- Bikeshedding: specific syntax of the macro. I've tried to keep it close to idiomatic Rust (e.g., in the sense of using `match`), but I'm not sure how best to express substituted query parts and args.

montekki (dismissed):
No objections from my side, this is great.

perekopskiy:
> and [storage log queries w/ custom tables](https://github.com/matter-labs/zksync-era/blob/533f0131342dc40eea1bb82334828d0007043d19/core/lib/dal/src/storage_logs_dal.rs#L778) (can they be removed? cc @perekopskiy).

Yes, they can be removed along with `storage_logs_dedup_migration`

perekopskiy (commented):
Looks really good!
> Does it make sense to use this approach in BlocksWeb3Dal / TransactionsWeb3Dal?

IIRC we thought that processing API request with single DB request should be more performance but I don't remember that we actually validate it, cc @RomanBrodetski. IMO, it now looks better than it was and even if it leads to a small performance degradation it's ok (but I doubt it will lead to any)

perekopskiy (approved):

PR #1289, opened 2024-02-29
fix(snapshots_creator): Fix snapshot generation query

slowli:
## What ❔

- Fixes the DB query returning query chunks, so that it doesn't incorrectly include storage logs that would be initially written to in future L1 batches.
- Adds basic sanity checks to Postgres recovery.

## Why ❔

These extra logs lead to incorrect tree recovery.

## Checklist


tomg10 (approved):

popzxc (approved):

RomanBrodetski (commented):
👍

PR #1290, opened 2024-02-29
feat(node_framework): Add a custom setup hook

popzxc:
⚠️ To be superseded by a different approach.

## What ❔

Adds the ability to implement a custom setup hook with access to context able to retrieve resources and spawn tasks early.

## Why ❔

Often there may be some code that we want to run before we actually launch the node. As a simple case, check if the genesis batch info is present in the DB. It would be nice to do it:
- From within the framework to avoid responsibility split.
- With access to the service lifecycle (e.g. if we're performing some long operation, we would want to first spawn the healthcheck server).

## Checklist

### DIFF
--- a/core/node/node_framework/examples/main_node.rs
+++ b/core/node/node_framework/examples/main_node.rs
@@ -159,7 +159,7 @@ fn main() -> anyhow::Result<()> {
         .add_proof_data_handler_layer()?
         .add_healthcheck_layer()?
         .build()
-        .with_setup(|pre_run| {
+        .run_with_setup(|pre_run| {
             Box::pin(async move {
                 // We want to make sure that the healthcheck server is the very first
                 // thing that starts.
@@ -171,7 +171,4 @@ fn main() -> anyhow::Result<()> {
                 Ok(())
             })
         })
-        .run()?;
-
-    Ok(())
 }
--- a/core/node/node_framework/examples/main_node.rs
+++ b/core/node/node_framework/examples/main_node.rs
@@ -158,6 +158,18 @@ fn main() -> anyhow::Result<()> {
         .add_proof_data_handler_layer()?
         .add_healthcheck_layer()?
         .build()
+        .with_setup(|pre_run| {
+            Box::pin(async move {
+                // We want to make sure that the healthcheck server is the very first
+                // thing that starts.
+                pre_run.launch_task("healthcheck_server")?;
+
+                // ...imagine that there is some setup work here -- ensuring genesis state, running some migration,
+                // loading some data, etc..
+
+                Ok(())
+            })
+        })
         .run()?;
 
     Ok(())
--- a/core/node/node_framework/examples/main_node.rs
+++ b/core/node/node_framework/examples/main_node.rs
@@ -17,7 +17,7 @@ use zksync_node_framework::{
     implementations::layers::{
         eth_watch::EthWatchLayer,
         fee_input::SequencerFeeInputLayer,
-        healtcheck_server::HealthCheckLayer,
+        healtcheck_server::{HealthCheckLayer, HealthCheckTask},
         metadata_calculator::MetadataCalculatorLayer,
         object_store::ObjectStoreLayer,
         pools_layer::PoolsLayerBuilder,
@@ -29,6 +29,7 @@ use zksync_node_framework::{
         },
     },
     service::ZkStackService,
+    task::Task,
 };
 
 struct MainNodeBuilder {
@@ -162,7 +163,7 @@ fn main() -> anyhow::Result<()> {
             Box::pin(async move {
                 // We want to make sure that the healthcheck server is the very first
                 // thing that starts.
-                pre_run.launch_task("healthcheck_server")?;
+                pre_run.launch_task(HealthCheckTask::name())?;
 
                 // ...imagine that there is some setup work here -- ensuring genesis state, running some migration,
                 // loading some data, etc..
--- a/core/node/node_framework/src/implementations/layers/fee_input.rs
+++ b/core/node/node_framework/src/implementations/layers/fee_input.rs
@@ -70,7 +70,7 @@ struct GasAdjusterTask {
 
 #[async_trait::async_trait]
 impl Task for GasAdjusterTask {
-    fn name(&self) -> &'static str {
+    fn name() -> &'static str {
         "gas_adjuster"
     }
 
--- a/core/node/node_framework/src/service/mod.rs
+++ b/core/node/node_framework/src/service/mod.rs
@@ -3,6 +3,7 @@ use std::{collections::HashMap, fmt};
 use futures::{future::BoxFuture, FutureExt};
 use tokio::{runtime::Runtime, sync::watch};
 
+use self::pre_run::PreRun;
 pub use self::{context::ServiceContext, stop_receiver::StopReceiver};
 use crate::{
     resource::{ResourceId, StoredResource},
@@ -11,9 +12,12 @@ use crate::{
 };
 
 mod context;
+mod pre_run;
 mod stop_receiver;
 #[cfg(test)]
 mod tests;
+
+pub type SetupHook = Box<dyn FnOnce(&mut PreRun) -> BoxFuture<anyhow::Result<()>> + Send>;
 
 /// "Manager" class for a set of tasks. Collects all the resources and tasks,
 /// then runs tasks until completion.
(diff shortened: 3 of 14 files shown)


core/node/node_framework/examples/main_node.rs
```diff
@@ -158,7 +159,16 @@ fn main() -> anyhow::Result<()> {
...
-
-    Ok(())
+        .run_with_setup(|pre_run| {
+            Box::pin(async move {
+                // We want to make sure that the healthcheck server is the very first
+                // thing that starts.
+                pre_run.launch_task(HealthCheckTask::name())?;
+
+                // ...imagine that there is some setup work here -- ensuring genesis state, running some migration,
+                // loading some data, etc..
```
slowli, line 169: I think this approach works (other than minor interface nits), but isn't quite declarative. I wonder whether we could do better. One idea that I've come up with is to describe task preconditions (e.g., "initialized node storage") that would essentially lead to tasks being wrapped like this:

```rust
pub trait Task {
    // existing contents

    fn preconditions(&self) -> Preconditions {
        Preconditions::initialized_storage()
    }
    // or you can define `add_task_with_preconditions` in `ServiceContext`
}

struct WrappedTask<T: Task> {
    preconditions: Preconditions,
    inner: task,
}

impl<T: Task> WrappedTask<T> {
    async fn run(self, stop_receiver: StopReceiver) -> anyhow::Result<()> {
        self.preconditions.wait(stop_receiver.clone()).await?; // cancellation handling omitted for brevity
        self.inner.run(stop_receiver).await
    }
}
```

While it's possible to inject preconditions into each task manually, this approach avoids boilerplate and allows to define default preconditions, so that they can be opted *out of* (which would be the case with the healthcheck server, storage recovery etc.), rather than opted *into*.

I'm not sure how preconditions should be organized internally. It's tempting to say that preconditions should be fulfilled by other tasks (to fit them into the resource / task model); not entirely sure this makes sense. E.g., for the "initialized storage" precondition, you might say that storage initialization is a task that checks / performs initialization. In this case, preconditions can be implemented using watch channels. If fulfilled preconditions are declared in the task interface, it would be possible to "statically" (i.e., given a node config) check that all requested preconditions are actually fulfilled by other tasks etc.; but you'd probably need to support finite tasks for this. A simpler option would be to make preconditions a templated resource.

Another view on preconditions in "node lifecycle"; i.e., assigning different tasks to different parts of the node lifecycle ("initialization", "main phase" etc.). Intuitively, this sounds more restrictive and less expressive than the preconditions model.

core/node/node_framework/src/service/mod.rs
```diff
@@ -82,6 +90,22 @@ impl ZkStackService {
...
+    ///
+    /// This hook will be invoked after all the wiring layers have been processed and all the tasks
+    /// have been added to the service, but before any of the tasks are started. During the setup, the
+    /// hook can launch any added tasks and access resources through the [`PreRun`] object.
+    ///
+    /// The node will block on the hook execution if its provided, so if it's a long-running operation,
+    /// it is advised to start the supplementary tasks, such as healthcheck server, in the hook itself.
+    pub fn run_with_setup<F>(mut self, setup: F) -> anyhow::Result<()>
+    where
+        F: FnOnce(&mut PreRun) -> BoxFuture<anyhow::Result<()>> + Send + 'static,
```
slowli, line 103: Since the method has type params anyway, is there anything preventing replacing `BoxFuture` with another type param? Would make calling this method more eloquent.

core/node/node_framework/src/service/mod.rs
```diff
@@ -132,17 +156,30 @@ impl ZkStackService {
...
-            })
-            .collect();
+        let mut join_handles: Vec<_> = pre_run.join_handles;
+        let mut tasks = pre_run.unstarted_tasks;
+
+        for task in &mut tasks {
+            let Some(task) = task.task.take() else {
+                // The task was started during the pre-run.
+                continue;
+            };
```
slowli, line 180: Bikeshedding: You can use `if let` here to have more "linear" control flow.

core/node/node_framework/examples/main_node.rs
popzxc: Funnily, that roughly resembles the design I initially came up with and then discarded.
The problem that I see with an opt-out approach is that for the setup you probably want to have a greater control on the situation.
Imagine that two independent groups of developers work on two big features and decide to implement this opt-out pattern as separate tasks (for whatever reason), and both of them involve some kind of breaking change (e.g. running some kind of storage migration implemented in Rust). This can be merged unnoticed -- because you can never see the "setup" code easily, or even suspect that it exists.

This is why at least for now I chose the opt-in approach where we assume that the "node composer" understands the state of the node and can compose a sequence of actions required to alter its state.

popzxc: But at the same time I certainly can relate to your message, I just kinda feel that if there is a tradeoff between explicitness and declarativity, now is the time to choose explicitness.

More opinions are appreciated, cc @RomanBrodetski

core/node/node_framework/src/service/mod.rs
popzxc: I like the explicit `continue` because it comes with a comment explaining how can it happen and why is it fine, so probably I'll keep it as is.

core/node/node_framework/src/service/mod.rs
popzxc: `BoxFuture` is used within the `SetupHook` type, and I'm not sure how to perform a conversion from `dyn FnOnce(..) -> Fut` (where `Fut: Future<..>`) to `dyn FnOnce(..) -> BoxFuture<..>` without wrapping it into an additional closure.
Could you please give me a hint here?

core/node/node_framework/src/implementations/layers/fee_input.rs
```diff
@@ -70,7 +70,7 @@ struct GasAdjusterTask {
 
 #[async_trait::async_trait]
 impl Task for GasAdjusterTask {
-    fn name(&self) -> &'static str {
+    fn name() -> &'static str {
```
popzxc, line 73: dd

core/node/node_framework/examples/main_node.rs
slowli: Well, my proposal could easily be reworked to have opt-in if we remove the default implementation of `Task::preconditions()`. I still think it makes sense because it allows making waiting for first L1 batch in the storage (e.g., a part of API server) to be more declarative. As another option that requires less new notions:

- Add a resource for preconditions. Could be a templated one, but could be not, e.g. we may have `InitializedStorage` resource.
- Provide `async fn wait_until_ready()` as a part of this resource; in case of `InitializedStorage`, it may return general storage info (e.g., optional snapshot recovery metadata).
- Require this resource in the relevant components. Call `wait_until_ready()` at the start of the corresponding task(s).
- Provide this resource using a separate task (`InitializeStorageTask`?) which would encapsulate genesis / snapshot recovery.

I think task definitions could be made more declarative by having a generic wrapper for tasks with preconditions, so that the main task code still doesn't change:

```rust
struct MyTask {
    // ...
}

#[async_trait]
impl Task for MyTask {
    // ...
}

// in wiring code
let task = MyTask { /* ... */ };
let initialized_storage = context.get_resource::<InitializedStorage>()?;
let task = WithPreconditions::new(task, vec![initialized_storage]);
// ^ can probably be made more eloquent via `TaskExt` extension trait:
let task = task.with_precondition(initialized_storage);
context.add_task(task);

struct WithPreconditions<T> {
    task: T,
    preconditions: Vec<Box<dyn Precondition>>,
}

#[async_trait]
trait Precondition: Resource {
    async fn wait_until_ready(&self); // probably doesn't make sense to define output
}
```

core/node/node_framework/src/service/mod.rs
slowli: You're right, it doesn't look possible; it'd require async closures or syntax like

```rust
pub fn run_with_setup<F, Fut>(mut self, setup: F) -> anyhow::Result<()>
where
    F: for<'a> FnOnce(&'a mut PreRun) -> Fut<'a> + Send + 'static,
    for<'a> Fut<'a>: Future<Output = anyhow::Result<()>> + Send + 'a,
```

core/node/node_framework/examples/main_node.rs
popzxc: But then again -- with an opt-in scheme implemented this way one may forget to add a ready storage precondition (e.g. because they're simply not familiar with the framework) when it would really be required for the component to work correctly.

I try to judge from the perspective of easiness to misuse or do something wrong because of a lack of familiarity with the framework. The way I implemented it is kinda "stupid", but it's *understandably* stupid -- you don't need to know the framework to see all the code that runs before the node has started.

It surely comes with a significant disadvantage: you have to write the code for preconditions yourself, and you must be aware of the set of tasks that has been added to the service. I can already see that it may be a problem for a generic runner that would compose the node based on the node spec.

But at the same time, I'm afraid that we may overcomplicate the framework too much for people to give up on using it as intended.

PR #1293, opened 2024-02-29
feat!: support loading verifier and protocol version from db

mm-zk:
## What ❔

* Support loading the verification key and protocol version from DB

## Why ❔

* This way proving system can pick the right provers when processing a given batch (based on the batch protocol version rather than environment variables)

### DIFF
--- a/core/lib/zksync_core/src/proof_data_handler/request_processor.rs
+++ b/core/lib/zksync_core/src/proof_data_handler/request_processor.rs
@@ -104,17 +104,41 @@ impl RequestProcessor {
             .await
             .map_err(RequestProcessorError::ObjectStore)?;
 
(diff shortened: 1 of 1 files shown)


core/lib/types/src/protocol_version.rs
```diff
@@ -140,22 +140,43 @@ impl TryFrom<U256> for ProtocolVersionId {
...
+    Version19,
+    Version20,
+    Version21,
+    Version22,
 }
 
 impl FriProtocolVersionId {
     pub fn latest() -> Self {
-        Self::Version1
+        Self::Version21
```
mm-zk, line 175: thanks for fixing..

Are these "latest" and "next" actually used anywhere ?

PR #1296, opened 2024-02-29
fix(validium): Change assert for small gasPerPubdataByte test

ColoCarletti:
## What ❔

Modify the asserts of [test](https://github.com/lambdaclass/zksync-era/blob/8525ce9f324fb8888be734a1f51abab56f1f7d98/core/tests/ts-integration/tests/system.test.ts#L75) as different behaviors are expected depending on whether it is in validium mode or rollup mode.

## Why ❔

The test in rollup checks that a transaction with a very low value for gasPerPubdataByte is accepted but not executed, to verify this it makes a transaction with these characteristics and then another with the same nonce and expects it to be overwritten. When operating in validium mode, the transaction with a low gasPerPubdataByte is executed, so the second transaction with the same nonce has to be rejected.

## Checklist

### DIFF
--- a/core/tests/ts-integration/tests/system.test.ts
+++ b/core/tests/ts-integration/tests/system.test.ts
@@ -24,12 +24,10 @@ const contracts = {
 describe('System behavior checks', () => {
     let testMaster: TestMaster;
     let alice: zksync.Wallet;
-    let bob: zksync.Wallet;
 
     beforeAll(() => {
         testMaster = TestMaster.getInstance(__filename);
         alice = testMaster.mainAccount();
-        bob = testMaster.newEmptyAccount();
     });
 
     test('Network should be supporting Cancun+Deneb', async () => {


contracts
popzxc, line 1: Is the change of contracts submodule expected?

core/tests/ts-integration/tests/system.test.ts
```diff
@@ -73,25 +75,30 @@ describe('System behavior checks', () => {
...
-            to: alice.address,
-            customData: {
-                gasPerPubdata: smallGasPerPubdata
-            }
-        });
-
-        // Now send the next tx with the same nonce: it should override the previous one and be executed.
-        await expect(
-            alice.sendTransaction({
+        // Check if the gasPerPubdata is to low, making the test inapplicable.
```
popzxc, line 81:
```suggestion
        // Check if the gasPerPubdata is too low, making the test inapplicable.
```

core/tests/ts-integration/tests/system.test.ts
```diff
@@ -24,10 +24,12 @@ const contracts = {
 describe('System behavior checks', () => {
     let testMaster: TestMaster;
     let alice: zksync.Wallet;
+    let bob: zksync.Wallet;
```
popzxc, line 27: Is it used anywhere?

core/tests/ts-integration/tests/system.test.ts
popzxc: Also, for extra expressivity you can do the following:

```typescript
// A safe low value to determine whether we can run this test.
// It's higher than `smallGasPerPubdata` to not make the test flaky.
const gasPerPubdataThreshold = 5;
```

popzxc (commented):
Mostly looks good

popzxc (approved):

PR #1301, opened 2024-02-29
feat(Gas Oracle): Add token address to API subdirectory

SantiagoPittella:
## What ❔

Includes the token address in the conversion rate API.

## Why ❔

This was discussed in: https://github.com/matter-labs/zksync-era/pull/996#discussion_r1500314337

## Checklist


core/lib/config/src/configs/native_token_fetcher.rs
```diff
@@ -4,4 +4,5 @@ use serde::Deserialize;
 pub struct NativeTokenFetcherConfig {
     pub poll_interval: u64,
     pub host: String,
+    pub token_address: String,
```
popzxc, line 7: Maybe make it `Address` instead of `String` to keep it a bit more typesafe?

PR #1312, opened 2024-03-01
feat(Validium): Adding CI jobs for running integration tests in Validium mode

jordibonet-lambdaclass:
## What ❔

Using the [matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-expanding-configurations) functionality of GitHub Actions (similar to the `consensus` implementation), add jobs to the CI to run **the integration tests** and the **External Node** in Validium Mode.

### Continuous Integration

Update [ci-core-reusable.yml](https://github.com/matter-labs/zksync-era/pull/1312/files#diff-293b35dbbaf364cec87320ca4091f93d974165e35cc640d7f44cfb9f51ebed0d) to run the integration tests and the External Node in Validium Mode using matrix functionality. 

A new [dev_validium_docker.toml](https://github.com/matter-labs/zksync-era/pull/1312/files#diff-b1ea9b0f5054b358511597ff2db6b6587cb653b4143a3ee7d71db3bf02aa38a3) and [ext-node-validium-docker.toml](https://github.com/matter-labs/zksync-era/pull/1312/files#diff-cf656ef48cad5b659ce909abde22cf2754b70cb4d4c0fc9490c8dfe7a7f1289b) is also needed to use the correct database configuration.

### Fixing linter errors
Update [system.test.ts](https://github.com/matter-labs/zksync-era/pull/1312/files#diff-df2630e2ff72274deeda2ad93f62736b2f15eea583c9f44f9e587206ce30e026) and [commit_batch_info.rs](https://github.com/matter-labs/zksync-era/pull/1312/files#diff-2f6b2e1dcb43abbceafd9074128cf256a047d2cde65f109938bce7129999ca69) to make the linter pass.

### Fixing unit test
Fix [tests.rs](https://github.com/matter-labs/zksync-era/pull/1312/files#diff-153a6b12ef1eba752b68e91b2a89fca2bd994f0c938303c250600f7253e68dfd) to pass unit tests. The failing test was `correct_order_for_confirmations`, you can run it using this command:
```sh
zk test rust --package  zksync_core eth_sender::tests::correct_order_for_confirmations
```
This was due to the fact that contracts_are_pre_shared_bridge was set to false, which meant that we were using post shared bridge contracts and this is not the case.

### Revert test and Upgrade test

Update [revert-and-restart.test.ts](https://github.com/matter-labs/zksync-era/pull/1312/files#diff-f126bf01f609604b0b6d94ac0cc449cbd79a6bee5393f9159808fc52fc590345) and [upgrade.test.ts](https://github.com/matter-labs/zksync-era/pull/1312/files#diff-aeb6f672ca2d9517a6bedb530c3818508d5c700b8bb51c9e2e36c6470b7c715e) so that the value that the sender has once it has been returned is greater than or equal to the value it had before. this is done in this way and we think it makes sense since in Validium it spends less gas.

### Snapshot recovery test
Update [snapshot-recovery.test.ts](https://github.com/matter-labs/zksync-era/pull/1312/files#diff-ca70d9c7a06d1c2bb63a24b0b54215f7dc9312fd1e528fa3097594b7147586bb) to use Validium external node configuration in order to pass the test. The test failed because of a [timeout waiting for the connection](https://github.com/matter-labs/zksync-era/actions/runs/8281933516/job/22668795414?pr=1312). You can check the correction of this test by running the server in validium mode (`zk && zk clean --all && zk env dev_validium && zk init --validium-mode && zk server`) and running the following command in another terminal:
```sh
echo IS_VALIDIUM=1 >> .env
zk config compile ext-node-validium
yarn snapshot-recovery-test snapshot-recovery-test
```
We distinguish between Rollup and Validium using this `IS_VALIDIUM` variable set [here](https://github.com/matter-labs/zksync-era/pull/1312/files#diff-293b35dbbaf364cec87320ca4091f93d974165e35cc640d7f44cfb9f51ebed0dR143), similar to `IN_DOCKER`.


ilitteri (changes requested):
`ci-integration-tests-validium.yml` looks fine, but it is not enough for the CI to run it. We still need to modify `ci.yml`. Take a look at this to use as reference https://github.com/lambdaclass/zksync-era/pull/159/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR88

jordibonet-lambdaclass:
While adapting the CI for Validium of the integration tests and the external node, we found a problem with the connection string for the databases: in Rollup it was `postgres://postgres@postgres/prover_local` and in Validium `postgres://postgres@localhost/prover_local`.
We fixed it with the configuration file `dev_validium_docker.toml`, adding the same configuration as `docker.toml`.

Currently we have two doubts:
- We can't find where the connection string for databases has been configured to be `postgres://postgres@localhost/prover_local`.
- The contracts submodule seems not to be pointing to the right commit despite it is correctly set in `.gitmodules`, which causes the `Verifier` deployment to fail, because we need the deployment to take a parameter that has been added now, which is `--validium-mode`. You can check it in the [last CI run](https://github.com/matter-labs/zksync-era/actions/runs/8160772169/job/22308165141?pr=1312). Is the submodule commit being updated somewhere else than in the _Checkout system-contracts submodule step_?

ilitteri:
>     * The contracts submodule seems not to be pointing to the right commit despite it is correctly set in `.gitmodules`, which causes the `Verifier` deployment to fail, because we need the deployment to take a parameter that has been added now, which is `--validium-mode`. You can check it in the [last CI run](https://github.com/matter-labs/zksync-era/actions/runs/8160772169/job/22308165141?pr=1312). Is the submodule commit being updated somewhere else than in the _Checkout system-contracts submodule step_?

Probably there's something wrong with the commit that is pushed, I'll double check this

popzxc:
@jordibonet-lambdaclass I think that by default the prover database URL is taken from `etc/env/private.toml` 

image

jordibonet-lambdaclass:
Thank you @popzxc. 
Apologies, I made a mistake, we were actually referring to the other one (`postgres://postgres@postgres/prover_local`). 
We can't find in which part of the CI it is assigned that one and not the default one in `private.toml`.

popzxc:
@jordibonet-lambdaclass I'm not sure if I fully understand the issue, but as a rough hypothesis: recently we've changed the approach to the databases, instead of using "trust" auth method everywhere, we introduced a hardcoded password ("no secure password"). The connection string you mentioned has no password in it, so maybe you're trying to use an old connection string somewhere. If you're interested to see where the Postgres in configured in general, then it's [this docker-compose file](https://github.com/matter-labs/zksync-era/blob/main/docker-compose.yml).

If that doesn't help, please elaborate what you're trying to do, what goes wrong, and what would you expect to work.

ilitteri:
The integration tests that are failing are fixed in https://github.com/matter-labs/zksync-era/pull/1270 and https://github.com/matter-labs/zksync-era/pull/1296. We can wait for #1270 to be merged to pull from the base branch, but this PR can be reviewed in the mean time

core/tests/ts-integration/tests/api/web3.test.ts
```diff
@@ -217,17 +217,17 @@ describe('web3 API compatibility tests', () => {
...
-        const legacyTx = await alice.sendTransaction({
-            type: LEGACY_TX_TYPE,
-            to: alice.address
-        });
-        await legacyTx.wait();
-
-        const legacyApiReceipt = await alice.provider.getTransaction(legacyTx.hash);
-        expect(legacyApiReceipt.gasPrice).bnToBeEq(legacyTx.gasPrice!);
-    });
+    // test('Should check transactions from API / Legacy tx', async () => {
```
popzxc, line 220: Don't forget to uncomment later.

popzxc (commented):
Generally looks good, let's wait for all the prerequisites to be merged.

popzxc (approved):
I'll approve this PR tentatively to not block you -- feel free to merge as soon as you feel it's ready.
If there will be any issue with it post factum, we'll catch them during the review of the main PR.

PR #1323, opened 2024-03-01
test: add info about how many instructions it takes to run benchmarks

joonazan:
This makes the automatic VM performance reporting also report changes in number of instructions executed. Those changes can be used to identify cases where performance hasn't actually changed but gas prices have changed.

### DIFF
--- a/core/tests/vm-benchmark/Cargo.toml
+++ b/core/tests/vm-benchmark/Cargo.toml
@@ -37,3 +37,7 @@ path = "src/compare_iai_results.rs"
 [[bin]]
 name = "find-slowest"
 path = "src/find_slowest.rs"
+
+[[bin]]
+name = "instruction-counts"
+path = "src/instruction_counts.rs"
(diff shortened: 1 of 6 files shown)


joonazan:
I think it's good now. It is supposed to say N/A, as there's no opcode count for older commits.

core/tests/vm-benchmark/Cargo.toml
```diff
@@ -37,3 +37,7 @@ path = "src/compare_iai_results.rs"
 [[bin]]
 name = "find-slowest"
 path = "src/find_slowest.rs"
+
+[[bin]]
+name = "instruction-counts"
+path = "src/instruction_counts.rs"
```
popzxc, line 43: Nit: Missing newline at the end of the file.

core/tests/vm-benchmark/src/instruction_counts.rs
```diff
@@ -0,0 +1,16 @@
+use vm_benchmark_harness::{cut_to_allowed_bytecode_size, get_deploy_tx, BenchmarkingVm};
```
popzxc, line 1: Nit: Doc-comments are always appreciated, e.g. a comment explaining the purpose of the binary as a topmost module doc-comment, and doc-comments for the main structures introduced (like `InstructionCounter`) help a lot.

popzxc (approved):

joonazan:
@popzxc Thanks for the suggestions. I added (short) doc comments. But looks like I need to get a review from someone else to be able to merge.

PR #1324, opened 2024-03-01
feat(Gas Oracle): Move gas prices to be `U256` to avoid overflows

juan518munoz:
## What ❔

Replace variables types from `u64` to `U256`.

## Why ❔

Prevent possible overflows. 

## Checklist


popzxc (commented):
Not sure if this PR is ready to be reviewed, but I think that if we're going to switch to `U256` because `u64` can now overflow with custom base token, it is certainly in scope to address all the places where an overflow can occur (otherwise there isn't much value in the change).

jrchatruc:
> Not sure if this PR is ready to be reviewed, but I think that if we're going to switch to `U256` because `u64` can now overflow with custom base token, it is certainly in scope to address all the places where an overflow can occur (otherwise there isn't much value in the change).

That makes sense, do you want to add someone else to review this PR or are you checking it internally?

core/lib/dal/src/blocks_dal.rs
```diff
@@ -558,8 +558,8 @@ impl BlocksDal<'_, '_> {
             initial_bootloader_contents,
             used_contract_hashes,
             base_fee_per_gas,
-            header.l1_gas_price as i64,
-            header.l2_fair_gas_price as i64,
+            (header.l1_gas_price).as_u64() as i64,
+            (header.l2_fair_gas_price).as_u64() as i64, // TODO: this might overflow
```
jrchatruc, line 562: Fixing this overflow requires a database migration moving the field from `bigint` into `numeric(80,0)`. What is the policy for postgres migrations? Do we change the existing field or should we create a new one to start using now?

core/lib/dal/src/blocks_dal.rs
```diff
@@ -474,8 +475,7 @@ impl BlocksDal<'_, '_> {
         // Serialization should always succeed.
         let used_contract_hashes = serde_json::to_value(&header.used_contract_hashes)
             .expect("failed to serialize used_contract_hashes to JSON value");
-        let base_fee_per_gas = BigDecimal::from_u64(header.base_fee_per_gas.as_u64()) // TODO: this might overflow
-            .context("block.base_fee_per_gas should fit in u64")?;
+        let base_fee_per_gas = BigDecimal::from_str(&header.base_fee_per_gas.to_string())?;
```
juan518munoz, line 478: fixes possible overflow by converting _directly_ from `U256` to `BigDecimal`.

popzxc:
@jrchatruc I've asked @perekopskiy to review and assist you with this PR.

core/lib/dal/src/models/storage_block.rs
```diff
@@ -79,16 +79,18 @@ impl From<StorageL1BatchHeader> for L1BatchHeader {
...
             used_contract_hashes: serde_json::from_value(l1_batch.used_contract_hashes)
                 .expect("invalid value for used_contract_hashes in the DB"),
-            base_fee_per_gas: l1_batch
-                .base_fee_per_gas
-                .to_u64()
-                .expect("base_fee_per_gas should fit in u64"),
+            base_fee_per_gas: U256::from(
+                l1_batch
+                    .base_fee_per_gas
+                    .to_u64()
```
perekopskiy, line 85: You should avoid casting to u64 here

core/lib/dal/src/models/storage_block.rs
```diff
@@ -559,7 +564,7 @@ impl From<StorageMiniblockHeader> for MiniblockHeader {
             hash: H256::from_slice(&row.hash),
             l1_tx_count: row.l1_tx_count as u16,
             l2_tx_count: row.l2_tx_count as u16,
-            base_fee_per_gas: row.base_fee_per_gas.to_u64().unwrap(),
+            base_fee_per_gas: U256::from(row.base_fee_per_gas.to_u64().unwrap()),
```
perekopskiy, line 567: Same

core/lib/dal/src/models/storage_sync.rs
```diff
@@ -38,7 +39,7 @@ pub(crate) struct SyncBlock {
     pub l1_batch_number: L1BatchNumber,
     pub last_in_batch: bool,
     pub timestamp: u64,
-    pub l1_gas_price: u64,
+    pub l1_gas_price: U256,
     pub l2_fair_gas_price: u64,
     pub fair_pubdata_price: Option<u64>,
```
perekopskiy, line 44: Why only type of `l1_gas_price` is changed?

core/lib/multivm/src/versions/vm_1_3_2/refunds.rs
```diff
@@ -36,11 +36,10 @@ impl<H: HistoryMode, S: WriteStorage> VmInstance<S, H> {
         // For now, bootloader charges only for base fee.
         let effective_gas_price = self.block_context.base_fee;
 
-        let bootloader_eth_price_per_pubdata_byte = U256::from(effective_gas_price)
```
perekopskiy, line 39: You should avoid changing multivm version except for latest. Instead you should glue results

core/lib/types/src/api/en.rs
```diff
@@ -25,7 +25,7 @@ pub struct SyncBlock {
     /// L2 block timestamp.
     pub timestamp: u64,
     /// L1 gas price used as VM parameter for the L1 batch corresponding to this L2 block.
-    pub l1_gas_price: u64,
+    pub l1_gas_price: U256,
     /// L2 gas price used as VM parameter for the L1 batch corresponding to this L2 block.
     pub l2_fair_gas_price: u64,
     /// The pubdata price used as VM parameter for the L1 batch corresponding to this L2 block.
```
perekopskiy, line 31: Same here, why did you change only `l1_gas_price`?

perekopskiy (commented):
Some general questions/comments
- Do I get it right that `base_fee_per_gas`/`l2_fair_gas_price` may overflow `u64` if base token price is low?
- Do I get it right that `l1_gas_price` may overflow `u64` if L1 is not Ethereum but some other chain with cheap base token?
- `u64` is sufficient for `gas_per_pubdata`, because all VM gas fits into `u32`, so even `u32` is enough. So please revert all changes you've done to `gas_per_pubdata`

jrchatruc:
> Some general questions/comments
> 
> * Do I get it right that `base_fee_per_gas`/`l2_fair_gas_price` may overflow `u64` if base token price is low?
> * Do I get it right that `l1_gas_price` may overflow `u64` if L1 is not Ethereum but some other chain with cheap base token?
> * `u64` is sufficient for `gas_per_pubdata`, because all VM gas fits into `u32`, so even `u32` is enough. So please revert all changes you've done to `gas_per_pubdata`

- Indeed both `base_fee_per_gas` and `l2_fair_gas_price` may overflow 64 bits if the base token price is low.
- Yes, I believe gas prices in Ethereum are also technically `U256`, so I think it's a good idea to change it so it matches.
- Makes sense, we'll revert them

core/lib/dal/src/models/storage_block.rs
jrchatruc: Addressed here [37fccd5](https://github.com/matter-labs/zksync-era/pull/1324/commits/37fccd5b0468b99de0d861dbde9c23f5a43bb293)

core/lib/dal/src/models/storage_block.rs
jrchatruc: Same as above, addressed here [37fccd5](https://github.com/matter-labs/zksync-era/pull/1324/commits/37fccd5b0468b99de0d861dbde9c23f5a43bb293)

core/lib/dal/src/models/storage_sync.rs
jrchatruc: Yeah you're right, we will also update `l2_fair_gas_price` and `fair_pubdata_price`. Also I realized the type for prices need to be updated for sync blocks along with the protobuf `.proto` files used for external node sync messages (writing it here to track it)

core/lib/types/src/api/en.rs
jrchatruc: Same comment as [this](https://github.com/matter-labs/zksync-era/pull/1324#discussion_r1513527901)

core/lib/dal/src/blocks_dal.rs
jrchatruc: I added a migration creating new columns for the values that need to be moved to U256 here [518ee35](https://github.com/matter-labs/zksync-era/pull/1324/commits/518ee352e221b836fb5311b7962c2d4db5a146a4). I don't believe it's possible to have an old server version work with a DB with applied migrations if we simply modify the column type.

Essentially the migration creates new columns with a `_u256` suffix, copies all the existing values on the non-`_u256` column over to it, and the rust `dal`/`sqlx` code is updated to make selects take the `_u256` version and map it to the existing struct fields. Inserts are updated to insert into the new `_u256` column, and an `SQL` trigger is created for inserts and updates to those tables to also insert/update the old column with the same value. The only issue here is what happens when the value inserted/updated does not fit in a `bigint`/`i64`. The trigger implementation I pushed will set the old column value to `-1` in that case to signify an error. 

Let me know what you think of this, in theory that scenario should not be a problem, as in the currently deployed environments the value should always fit in a `bigint`, but it might be better to error out in those environments to make absolutely nothing weird happens.

core/lib/dal/src/models/storage_sync.rs
jrchatruc: We have updated both `l2_fair_gas_price` and `fair_pubdata_price` and the `.proto` files, but I realized they are not actually used for syncing, as that's done through the regular RPC api the operator exposes.

Protobuf types were updated here [a478592](https://github.com/matter-labs/zksync-era/pull/1324/commits/a47859255385ed7cfedbd46bc96c9507cba3bfab) and a simple unit test for the change was  added here [2c75075](https://github.com/matter-labs/zksync-era/pull/1324/commits/2c750754c8ee8ab22e5d39c48289f43c6d6fd344), but to make sure: what is the whole protobuf/storagesync stuff used for? How do we know it's working outside of some basic unit tests?

core/lib/dal/migrations/20240307190434_change_to_u256.up.sql
```diff
@@ -0,0 +1,51 @@
...
+ALTER TABLE miniblocks ADD l1_gas_price_u256 numeric(80,0) NOT NULL DEFAULT 0;
+ALTER TABLE miniblocks ADD l2_fair_gas_price_u256 numeric(80,0) NOT NULL DEFAULT 0;
+ALTER TABLE miniblocks ADD fair_pubdata_price_u256 numeric(80,0) NOT NULL DEFAULT 0;
+
+UPDATE l1_batches SET l1_gas_price_u256=l1_gas_price;
+UPDATE l1_batches SET l2_fair_gas_price_u256=l2_fair_gas_price;
+
+UPDATE miniblocks SET l1_gas_price_u256=l1_gas_price;
+UPDATE miniblocks SET l2_fair_gas_price_u256=l2_fair_gas_price;
+UPDATE miniblocks SET fair_pubdata_price_u256=fair_pubdata_price;
```
perekopskiy, line 13: This should be a programmatic migration, see https://github.com/matter-labs/zksync-era/tree/main/core/lib/dal#programmatic-migrations

core/lib/dal/src/blocks_dal.rs
```diff
@@ -1574,8 +1581,7 @@ impl BlocksDal<'_, '_> {
         default_aa_hash: H256,
         protocol_version_id: ProtocolVersionId,
     ) -> anyhow::Result<Vec<L1BatchWithMetadata>> {
-        let raw_batches = sqlx::query_as!(
-            StorageL1Batch,
+        let raw_batches = sqlx::query_as::<_, StorageL1Batch>(
```
perekopskiy, line 1584: Hmm, why are you changing macro to function?

core/lib/dal/migrations/20240307190434_change_to_u256.up.sql
jrchatruc: Got it, I'll remove the updates on the migration and instead add a task that every few seconds updates some of the rows in batches of around 5 or 10k until it's done. Is it reasonable to write the task in Rust/Tokio inside the operator or do you have another method of doing it? We couldn't find an example in the codebase.

perekopskiy: Check `fee_address_migration` as an example

core/lib/multivm/src/versions/vm_1_3_2/refunds.rs
jrchatruc: The problem here is we have updated the `L1PeggedBatchFeeModelInput` struct to move gas prices to `U256`. This struct is defined on the `zksync_types` crate under fee model, outside the vm/multivm stuff. This change breaks all the `multivm` methods/functions that use it because they expect  their fields to still be `u64` (for example the `derive_base_fee_and_gas_per_pubdata` function).

I'm not sure what the best way to proceed here is; we can introduce a new struct (something like `L1PeggedBatchFeeModelInputU256`) that the latest vm uses, while the old versions keep using the non `U256` version and make the `derive_base_fee_and_gas_per_pubdata` function in utils dispatch accordingly. Is that reasonable?

perekopskiy: Yes! IMO it's better to update `L1PeggedBatchFeeModelInput` to use `U256` and introduce struct `L1PeggedBatchFeeModelInputLegacy` with u64 and describe why we need it with a doc-comment. And in old VM version modules you can do `use ...::L1PeggedBatchFeeModelInputLegacy as L1PeggedBatchFeeModelInput` so the diff will be really small

core/lib/dal/src/blocks_dal.rs
jrchatruc: The reason for this is the `query_as!` macro [needs the struct field names to be the exact same as the ones in postgres](https://docs.rs/sqlx/latest/sqlx/macro.query_as.html), but with the new columns mapped to the old struct fields, so the names do not match anymore. The `query_as` function allows for this

core/lib/multivm/src/versions/vm_1_3_2/refunds.rs
jrchatruc: After trying out the changes I realized that doing this ends up requiring going all the way to the top to the `Vm` struct and `VmInterface` trait, because it contains an they both contain an `L1BatchEnv`, which holds the gas prices that need to be changed from `u64` to `U256`. 

This is a lot of structs/traits (at least 7) that would now have their `Legacy` version, which is a lot of changes. We can go that route but I want to double check if it's reasonable or if there are other alternatives.

jrchatruc:
The change required for [this ticket](https://github.com/lambdaclass/zksync-era/issues/201) (related to [this discussion](https://github.com/matter-labs/zksync-era/pull/996#discussion_r1500273515)) was merged into this PR (here [a11e12c](https://github.com/matter-labs/zksync-era/pull/1324/commits/a11e12cb38fc9487c5c8edd6a2ea2b6838d4f9f3)) since it requires its changes to do it.

core/lib/multivm/src/versions/vm_1_3_2/refunds.rs
lferrigno: @perekopskiy @popzxc can you please check the previous message. Thanks!

PR #1340, opened 2024-03-04
refactor: Refactor KZG trusted setup

slowli:
## What ❔

- Refactors the KZG trusted setup so that it doesn't require reading files in runtime.
- Changes commitment data checks in the consistency checker so that the DA source is determined dynamically based on the reference commitment.

## Why ❔

Simplifies maintenance.

## Checklist


prover/witness_generator/src/main.rs
```diff
@@ -184,15 +182,23 @@ async fn main() -> anyhow::Result<()> {
...
-                let kzg_config = KzgConfig::from_env().context("KzgConfig::from_env()")?;
+                let trusted_setup_path =
+                    std::env::var("KZG_TRUSTED_SETUP_PATH").or_else(|err| match err {
+                        std::env::VarError::NotPresent => Ok("./trusted_setup.json".to_owned()),
+                        std::env::VarError::NotUnicode(_) => {
+                            let err = anyhow::Error::new(err)
+                                .context("`KZG_TRUSTED_SETUP_PATH` env variable is invalid");
+                            Err(err)
+                        }
+                    })?;
```
slowli, line 193: I didn't manage to switch to static KZG settings here because they are passed as a file path (rather than `KzgSettings`) inside `zkevm_test_harness` functions.

core/lib/l1_contract_interface/src/i_executor/commit/trusted_setup.rs
```diff
@@ -0,0 +1,4176 @@
...
+        roots_of_unity_brp: roots_of_unity.try_into().unwrap(),
+        setup_g2_1,
+        lagrange_setup_brp: lagrange_setup.try_into().unwrap(),
+    }
+});
+
+// Taken from the C KZG library: https://github.com/ethereum/c-kzg-4844/blob/main/src/trusted_setup.txt
+// with added `Hex.decode(_)` wrappers.
+#[allow(long_running_const_eval)]
+const L1_LAGRANGE: [[u8; 48]; ELEMENTS_PER_4844_BLOCK] = [
```
slowli, line 79: There are a couple of other ways to include these constants, e.g. parse them using a build script. I've chosen this approach because it's more transparent and requires less privileges during the build.

popzxc: Not a strong opinion, but given that it's already wrapped into `once_cell::Lazy` (especially with `L1_LAGRANGE.iter()`, maybe we'll have a `trusted_setup.txt` as a separate file and use `include_str!` here?

prover/witness_generator/src/main.rs
popzxc: Maybe to unify the approach and simplify the requirements on inputs, we'll do a [lil' hack here](https://docs.rs/tempfile/3.10.1/tempfile/fn.tempfile.html) (obviously wrapped into an `once_cell` for efficiency)? At the same time, we can ask the protocol team to change the public interface in the harness so that `KzgSettings` are passed directly and we can remove it later.

popzxc (approved):
Nice!

core/lib/l1_contract_interface/src/i_executor/commit/trusted_setup.rs
slowli: Done. This means ~2x more space occupied in the binary, but I guess it's not the largest concern, and we could switch to using a build script later if this becomes an issue.

popzxc (dismissed):

core/lib/protobuf_config/src/kzg.rs
```diff
@@ -1,22 +0,0 @@
-use anyhow::Context as _;
```
perekopskiy, line 1: Please remove core/lib/protobuf_config/src/proto/kzg.proto as well

perekopskiy (approved):
Nice!

koloz193 (approved):

montekki (approved):

PR #1342, opened 2024-03-04
fix(en): Fix pending transactions subscription

slowli:
## What ❔

- Fixes `pendingTransactions` subscription notifier logic on EN. This logic implicitly relies on the "received at" timestamp of transactions being causally ordered (i.e., transactions persisted into the storage later should have a later timestamp), but this wasn't the case for ENs – ENs would copy this timestamp from the main node. This is fixed by overriding the "received at" timestamp on ENs before saving a transaction to the storage.
- Correspondingly, the "received at" timestamp is removed from the consensus payload because consensus wouldn't agree on it.
- Disables reporting transaction inclusion latencies for ENs, which is meaningless and can lead to spammy logs.

## Why ❔

- Bugs bad.
- Spammy logs bad.

## Checklist


core/lib/zksync_core/src/state_keeper/io/seal_logic.rs
```diff
@@ -272,45 +271,61 @@ impl MiniblockSealCommand {
...
-            let progress = MINIBLOCK_METRICS.start(MiniblockSealStage::PreInsertTxs, is_fictive);
-            for tx in &self.miniblock.executed_transactions {
-                if let Ok(l1_tx) = L1Tx::try_from(tx.transaction.clone()) {
+    async fn insert_transactions(&self, transaction: &mut StorageProcessor<'_>) {
+        for tx_result in &self.miniblock.executed_transactions {
+            let mut tx = tx_result.transaction.clone();
+            // Override the "received at" timestamp for the transaction so that they are causally ordered (i.e., transactions
+            // with an earlier timestamp are persisted earlier). Without this property, code relying on causal ordering may work incorrectly;
+            // e.g., `pendingTransactions` subscriptions notifier can skip transactions.
+            tx.received_timestamp_ms = unix_timestamp_ms();
```
RomanBrodetski, line 280: Feels a little uncomfortable to set it here - the transaction was actually _received_ earlier.. But I understand the reasonning and cannot propose a better solution, so let's keep it

RomanBrodetski: We'll likely need to change this logic anyway when we have a different/more sophisticated mempool, so I'm fine both with the current solution and DAL. I personally prefer DAL, so if it doesn't add additional hustle let's change. As for `created_at` - yeah, let's not use it in app logic and keep just it only for journaling/debug

slowli: JFYI, I have moved the timestamp assignment earlier into the transaction lifecycle, when a `Transaction` is converted into a `SyncAction`.

core/lib/zksync_core/src/sync_layer/sync_action.rs
```diff
@@ -140,6 +143,8 @@ pub(crate) enum SyncAction {
         timestamp: u64,
         virtual_blocks: u32,
     },
+    /// **Important.** Please only use [`Self::transaction()`] to convert transactions to this variant;
+    /// it sets a correct "received at" timestamp for the transaction.
     Tx(Box<Transaction>),
```
popzxc, line 148: If we want I guess we can make it impossible by introducing, for example, `struct AdjustedTransaction(Box<Transaction>)` with all the `Deref`/`DerefMut`/`From` conversions and using it here instead of `Box<Transaction>`. This way it'll also be easily visible that we somehow change the transaction during the flow. Dunno if that's overkill.

popzxc (dismissed):

core/lib/zksync_core/src/sync_layer/sync_action.rs
```diff
@@ -140,7 +140,9 @@ pub(crate) enum SyncAction {
         timestamp: u64,
         virtual_blocks: u32,
     },
-    Tx(Box<Transaction>),
+    /// **Important.** Please only use [`Self::transaction()`] to convert transactions to this variant;
+    /// it sets a correct "received at" timestamp for the transaction.
```
popzxc, line 144: No longer relevant?

popzxc (approved):

PR #1343, opened 2024-03-04
feat(Validium): Commit data generation mode cross-check in EN

ilitteri:
## What ❔

This PR:
- Adds a cross-check of the commit data generation mode (Validium or Rollup) between the deployed L1 contracts and the config set in the server.
- Updates the ConsistencyChecker `new` method to receive an `impl EthInterface` instead of receiving an Ethereum client URL to instantiate one.
- Moves the function that contains the main logic to `utils` module given that it is used both in EN and in `zksync_core`.

## Why ❔

The mode was being cross-checked in the server, but it should also be cross-checked in the EN.

## Checklist

## Aditional Information

Unit tests for the function in question will be added in the `zksync_core` module in another PR

### DIFF
--- a/core/lib/zksync_core/src/consistency_checker/mod.rs
+++ b/core/lib/zksync_core/src/consistency_checker/mod.rs
@@ -5,7 +5,7 @@ use serde::Serialize;
 use tokio::sync::watch;
 use zksync_contracts::PRE_BOOJUM_COMMIT_FUNCTION;
 use zksync_dal::{ConnectionPool, StorageProcessor};
-use zksync_eth_client::{clients::QueryClient, Error as L1ClientError, EthInterface};
+use zksync_eth_client::{Error as L1ClientError, EthInterface};
 use zksync_health_check::{Health, HealthStatus, HealthUpdater, ReactiveHealthCheck};
 use zksync_l1_contract_interface::{i_executor::structures::CommitBatchInfo, Tokenizable};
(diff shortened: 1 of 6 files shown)


core/lib/zksync_core/src/consistency_checker/mod.rs
```diff
@@ -209,18 +209,17 @@ impl ConsistencyChecker {
     const DEFAULT_SLEEP_INTERVAL: Duration = Duration::from_secs(5);
 
     pub fn new(
-        web3_url: &str,
+        l1_client: impl EthInterface,
```
popzxc, line 212: Nit: `impl SomeTrait` is a hidden generic, which means that this function has to be monomorphized, which comes with a compilation time penalty. At the same time, this function already accepts `Arc<dyn Trait>`, which introduces an inconsistency. Given that there is no real need for this function to be generic, it's probably be better to accept `Box<dyn EthInterface>` directly. (and if we are to add a wider context, though it's even more of FYI, in the notion of the WIP [framework](https://github.com/matter-labs/zksync-era/tree/main/core/node/node_framework) `EthInterface` represents [a shared resource](https://github.com/matter-labs/zksync-era/blob/main/core/node/node_framework/src/implementations/resources/eth_interface.rs), so it can be made `Arc`).

Not a big deal, however, so feel free to leave as is.

popzxc (approved):

PR #1344, opened 2024-03-04
feat: enabled loading yaml config for the main node

pompon0:
Added config_path flag to the main node binary.
If set, binary will load the config from a yaml file instead of env vars.
I've moved each proto descriptor file to its own package, because prost seems to be unable to import package from multiple files.
I've moved secrets out of TempConfigStore, so that they are delivered separately.


core/bin/external_node/src/config/mod.rs
```diff
@@ -471,16 +471,20 @@ impl PostgresConfig {
...
-    })
+    Ok(Some(decode_json(&cfg).context("failed decoding JSON")?))
+}
+
+pub(crate) fn read_consensus_config() -> anyhow::Result<Option<consensus::Config>> {
+    let Ok(path) = std::env::var("EN_CONSENSUS_CONFIG_PATH") else {
+        return Ok(None);
+    };
+    let cfg = std::fs::read_to_string(&path).context(path)?;
+    Ok(Some(decode_json(&cfg).context("failed decoding JSON")?))
```
Deniallugo, line 487: why json? 
if we use yaml in other places?

Deniallugo (approved):
question are only to consensus, i'd love to resolve them now, but feel free to implement it later

core/bin/external_node/src/config/mod.rs
```diff
@@ -471,16 +471,20 @@ impl PostgresConfig {
...
-    Ok(consensus::FetcherConfig {
-        executor: cfg.executor_config(node_key),
-    })
+    Ok(Some(decode_json(&cfg).context("failed decoding JSON")?))
+}
+
+pub(crate) fn read_consensus_config() -> anyhow::Result<Option<consensus::Config>> {
+    let Ok(path) = std::env::var("EN_CONSENSUS_CONFIG_PATH") else {
+        return Ok(None);
+    };
```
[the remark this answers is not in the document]
pompon0: consensus is configured always via config, even if the rest of the config is in env vars. If everything is in a file, consensus is also embedded in the file, if everything is in env vars, this is the env var pointing to the consensus configuration.

core/bin/external_node/src/config/mod.rs
pompon0: nobody protested so far. I can change it to yaml.

core/bin/external_node/src/config/mod.rs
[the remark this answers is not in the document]
pompon0: if we remove env var configuration, these files will also be gone.

popzxc (dismissed):
Nice!

popzxc:
Although we have [this fellow](https://github.com/matter-labs/zksync-era/blob/main/core/lib/config/src/configs/eth_sender.rs#L128). This was added as a quick fix to make it possible to not have a private key in each env we use, not sure if this in or out of scope for this PR.

Deniallugo:
> Although we have [this fellow](https://github.com/matter-labs/zksync-era/blob/main/core/lib/config/src/configs/eth_sender.rs#L128). This was added as a quick fix to make it possible to not have a private key in each env we use, not sure if this in or out of scope for this PR.

I'd say it's out of scope and i'll move it to secret part

core/bin/external_node/src/config/mod.rs
pompon0: done

Deniallugo (approved):

PR #1345, opened 2024-03-04
fix(eth_sender): makes pubdata sending mode config backwards compatible

montekki:
## What ❔

This config variable has become non-`Option`-al breaking all existing configs. This change makes this new config variable backwards compatible with existing configs.

## Checklist

### DIFF
--- a/core/lib/protobuf_config/src/eth_sender.rs
+++ b/core/lib/protobuf_config/src/eth_sender.rs
@@ -1,5 +1,5 @@
 use anyhow::Context as _;
-use zksync_config::configs::{self};
+use zksync_config::configs::{self, eth_sender::PubdataSendingMode};
 use zksync_protobuf::{read_required_repr, required, ProtoRepr};
 
 use crate::proto;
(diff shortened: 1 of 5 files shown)


core/lib/protobuf_config/src/eth_sender.rs
```diff
@@ -124,10 +124,12 @@ impl ProtoRepr for proto::Sender {
...
-            pubdata_sending_mode: required(&self.pubdata_sending_mode)
-                .and_then(|x| Ok(proto::PubdataSendingMode::try_from(*x)?))
-                .context("pubdata_sending_mode")?
-                .parse(),
+            pubdata_sending_mode: Some(
+                required(&self.pubdata_sending_mode)
+                    .and_then(|x| Ok(proto::PubdataSendingMode::try_from(*x)?))
+                    .context("pubdata_sending_mode")?
+                    .parse(),
+            ),
```
montekki, line 132: Although protobuf config should also be optional

PR #1346, opened 2024-03-04
fix(api): SQL: use = instead of ANY where possible in events-related queries

RomanBrodetski:
We have the following request:

```
	FROM events 
	WHERE (miniblock_number >= $6) AND (miniblock_number <= $7) AND (address = ANY($1)) AND (topic1 = ANY($2)) AND (topic2 = ANY($3)) AND (topic3 = ANY($4)) 
	ORDER BY miniblock_number ASC, event_index_in_block ASC 
```

currently regardless of amount of elements in every filter, we apply ANY clause. This PR changes this so that if there is exactly one element, `ANY` is replaced with `=`.

**WHY**

we have a hypotesis that this will significantly improve DB performance of this query


core/lib/dal/src/events_web3_dal.rs
```diff
@@ -118,18 +128,69 @@ impl EventsWeb3Dal<'_, '_> {
...
+    fn bind_params_for_optional_filter_query<'q>(
+        query: Query<'q, Postgres, PgArguments>,
+        values: Vec<&'q [u8]>,
+    ) -> Query<'q, Postgres, PgArguments> {
+        match values.len() {
+            0 => query,
+            1 => query.bind(values[0]),
+            _ => query.bind(values),
+        }
+    }
```
RomanBrodetski, line 192: This is not good - but I couldn't find a way to write this in a generic manner (`.bind` is defined sepraately on `Query` and `QueryAs`. `QueryAs`  has `inner` of type `Query`, but it's `pub(crate)`

perekopskiy (approved):
Please make lints pass

perekopskiy (approved):

PR #1350, opened 2024-03-05
feat: Move prover DAL to prover workspace

Artemka374:
## What ❔

- Move prover methods to prover DAL crate
- Generalize StorageProcessor and move it to another crate
- Rename StorageProcessor used in ConnectionPool to ConnectionOperator

## Why ❔

For further creating typesafe connections and prevent accessing tables from other databases with one connection pool.

## Checklist


Artemka374:
It shouldn't, I didn't touch it, all the previous runs were successful though

core/lib/types/Cargo.toml
```diff
@@ -32,6 +32,8 @@ num_enum = "0.6"
 hex = "0.4"
 prost = "0.12.1"
 
+sqlx = "0.7.3"
```
popzxc, line 35: This is **very** dangerous. One of the main purposes of DAL crate was not to give direct access to SQLx to any dependent crates and make sure that all of our DB-related stuff is well encapsulated.

At the same time, `zksync_types` is a crate where people move stuff when they need to share something but don't know where to put it. Which means that virtually everything in zkSync stack depends on `zksync_types`. This crate is already pretty toxic in APIs it exposes, and it makes it even more toxic: people use what they can use and in a rush they may do hacks that are hard to fix later.

AFAICS you only need sqlx to expose `chrono` types, which are [just reexports](https://github.com/launchbadge/sqlx/blob/main/sqlx-core/src/types/mod.rs#L40). Instead you can depend on [the same version of chrono](https://github.com/launchbadge/sqlx/blob/v0.7.3/sqlx-core/Cargo.toml#L59) and simply use these types directly. Which [seemingly we already do](https://github.com/matter-labs/zksync-era/blob/cadffe6410385bd6fb62c5a7646ae2b8272d56e2/core/lib/types/Cargo.toml#L22) so probably you don't need this dependency at all.

core/lib/dal/Cargo.toml
```diff
@@ -21,6 +21,8 @@ zksync_health_check = { path = "../health_check" }
 zksync_consensus_roles = { version = "0.1.0", git = "https://github.com/matter-labs/era-consensus.git", rev = "97d139969476a004c50f8b4a31ece748e5bee14e" }
 zksync_consensus_storage = { version = "0.1.0", git = "https://github.com/matter-labs/era-consensus.git", rev = "97d139969476a004c50f8b4a31ece748e5bee14e" }
 zksync_protobuf = { version = "0.1.0", git = "https://github.com/matter-labs/era-consensus.git", rev = "97d139969476a004c50f8b4a31ece748e5bee14e" }
+zksync_db_connection = { path = "../db_connection" }
+zksync_prover_dal = { path = "../../../prover/prover_dal" }
```
popzxc, line 25: This is to minimize the diff and will be removed in follow-up PRs, right?

core/lib/zksync_core/src/consensus/storage/mod.rs
```diff
@@ -16,7 +16,7 @@ use crate::{
     sync_layer::{fetcher::FetchedBlock, sync_action::ActionQueueSender},
 };
 
-/// Context-aware `zksync_dal::StorageProcessor` wrapper.
+/// Context-aware `zksync_dal::StorageProcessorWrapper` wrapper.
```
popzxc, line 19: Glob replace?

prover/prover_dal/src/fri_prover_dal.rs
```diff
@@ -249,27 +26,23 @@ impl FriProverDal<'_, '_> {
...
                 circuit_blob_url,
-                is_node_final_proof,
+                false,
                 protocol_version_id,
             )
             .await;
         }
-        drop(latency);
+        // drop(latency);
     }
```
popzxc, line 46: WIP?

prover/prover_dal/src/fri_prover_dal.rs
```diff
@@ -488,10 +261,10 @@ impl FriProverDal<'_, '_> {
             blob_url,
             id as i64,
         )
-        .instrument("save_fri_proof")
-        .report_latency()
-        .with_arg("id", &id)
-        .fetch_optional(self.storage)
+        // .instrument("save_fri_proof")
+        // .report_latency()
+        // .with_arg("id", &id)
```
popzxc, line 266: Same

core/lib/db_connection/src/lib.rs
```diff
@@ -113,7 +115,8 @@ impl Drop for PooledStorageProcessor<'_> {
             let lifetime = self.created_at.elapsed();
             CONNECTION_METRICS.lifetime[&tags.requester].observe(lifetime);
 
-            if lifetime > ConnectionPool::global_config().long_connection_threshold() {
+            // TODO: value is hardcoded for now, but should be in sync with ConnectionPool::global_config()
+            if lifetime > Duration::from_millis(5_000) {
```
popzxc, line 119: Generally I'd expect TODOs that aren't considered to be in scope for this exact PR to be labeled with a linear ticket ID.

prover/witness_vector_generator/src/generator.rs
```diff
@@ -8,7 +8,7 @@ use anyhow::Context as _;
 use async_trait::async_trait;
 use tokio::{task::JoinHandle, time::sleep};
 use zksync_config::configs::FriWitnessVectorGeneratorConfig;
-use zksync_dal::{fri_prover_dal::types::GpuProverInstanceStatus, ConnectionPool};
+use zksync_dal::ConnectionPool;
```
popzxc, line 11: It feels clunky that `ConnectionPool` is not a part of `zksync_db_connection` crate and you cannot really use `zksync_prover_dal` without it.

I thought that the logic would be as follows:
- `zksync_db_connection` defines _all_ the machinery required to interact with the database, and only it.
- `zksync_dal` (optionally renamed to `zksync_sequencer_dal`) defines the sequencer DALs and only them.
- `zksync_prover_dal` defines the prover DALs and only them.

With segregations completed, it'd mean that crates in the core workspace would depend *only* on `zksync_dal` and crates in the prover workspace would depend *only* on `zksync_prover_dal`.

popzxc (commented):
Had a pretty shallow review for now, but overall the PR feels to be in a WIP state.

prover/witness_vector_generator/src/generator.rs
Artemka374: that's true, but it will be done in the next PR, where we will separate ConnectionPool's, in this PR we would need a separate crates for both storage processor and connection pool, which will be avoided in the next one - so consider this one more as a preparation PR to make it more easy to review(although this one is big too)

core/lib/db_connection/src/lib.rs
Artemka374: Also will be done in the next PR

core/lib/types/Cargo.toml
Artemka374: thanks, nice suggestion

Artemka374:
Generally some things in this PR will be reworked in the follow-up PR, so this one is kind of preparation to decrease diff and make it easier to review. And right, I just missed some comments, that I should've uncomment

core/lib/db_connection/src/metrics.rs
```diff
@@ -0,0 +1,14 @@
+use std::time::Duration;
+
+use vise::{Buckets, Histogram, LabeledFamily, Metrics, Unit};
+
+#[derive(Debug, Metrics)]
+#[metrics(prefix = "sql_connection")]
+pub(crate) struct ConnectionMetrics {
```
slowli, line 7: I'd expect that *all* metrics defined in the DAL crate would be moved here, together with the `instrument` module and macros (the macros would need to be exported in order to work). Are there arguments against sharing these among DAL crates?

core/lib/db_connection/src/lib.rs
```diff
@@ -90,7 +92,7 @@ impl TracedConnections {
     }
 }
 
-struct PooledStorageProcessor<'a> {
+pub struct PooledStorageProcessor<'a> {
```
slowli, line 95: It's worrisome that these internal structs need to be made public; looks like inappropriate abstraction boundaries. As a benchmark, this crate should export the least amount of types possible; ideally, only `ConnectionPool` / `StorageProcessor` equivalents, and helpers like macros, instrumentation etc.

core/lib/dal/src/blocks_web3_dal.rs
```diff
@@ -568,10 +569,11 @@ impl BlocksWeb3Dal<'_, '_> {
 
         // FIXME (PLA-728): remove after 2nd phase of `fee_account_address` migration
         #[allow(deprecated)]
-        self.storage
-            .blocks_dal()
-            .maybe_load_fee_address(&mut details.operator_address, details.number)
-            .await?;
+        BlocksDal {
+            storage: self.storage,
+        }
```
slowli, line 574: Dumb question: Could `Dal` structs wrap `&mut StorageProcessor` instead? (Obviously, it would make sense to implement `fn conn(&mut self) -> &mut PgConnection` to minimize the changeset, and `DerefMut<Target = BasicStorageProcessor` for the same purpose; although in the latter case, I'd argue that replacing `self.storage` with `&mut self.storage.0` doesn't make the PR *that* harder to review.) Having to construct DALs manually looks quite clunky.

core/lib/dal/src/models/storage_prover_job_info.rs
slowli, line 1: Is this module unused? (If so, please remove.) It's logically part of `prover_dal`, isn't it?

core/lib/dal/Cargo.toml
slowli, line 25: It sounds quite counter-intuitive to depend on a local crate outside the workspace. IMO, it'd make sense to temporarily place `prover_dal` in the main workspace and place it in `core/lib` directory, and move it to a proper directory / workspace in a follow-up PR.

slowli (commented):
I don't think that (doubtlessly noble) considerations about minimizing the changeset should lead to noticeably worsening code quality in `main` (leaky abstractions etc.). If we go with this logic, it could make sense to create a feature branch and target it.

core/lib/dal/src/blocks_web3_dal.rs
Artemka374: It can't wrap StorageProcessor - that's the most problematic part of this implementation. Since we don't want to actually separate types of `ConnectionPool -> StorageProcessor` in this PR(that is why this exact PR was made) - we aren't able to implement these `get_dal` methods for StorageProcessor because we will have some cyclic dependencies there - so we are using one type that we have inside of Dal structs - and the other one to actually be able to access Dal structs. I agree that it doesn't look good though - but I couldn't find a better way to implement this.

Artemka374:
@slowli Generally I agree with the point of making another branch as a target - but then we won't be able to test if everything works good with this DAL separation. So, my plan was to merge this one first, assure that everything works as intended - and then the one with ConnectionPool separation (I already started to work on it [here](https://github.com/matter-labs/zksync-era/pull/1334)).

Artemka374:
Thanks for reviewing though, I will try to get it in a better state

Artemka374:
Closing in favor of https://github.com/matter-labs/zksync-era/pull/1334

PR #1351, opened 2024-03-05
feat(db): Add Postgres table size metrics

slowli:
## What ❔

Exports Postgres table sizes as gauge metrics.

## Why ❔

Useful to monitor state growth, especially once we have state pruning, and/or have access to metrics from the partners' nodes.

## Checklist


core/lib/dal/src/connection/mod.rs
```diff
@@ -307,6 +307,12 @@ impl ConnectionPool {
         self.max_size
     }
 
+    /// Uses this pool to report Postgres-wide metrics (e.g., table sizes). Should be called sparingly to not spam
+    /// identical metrics from multiple places. The returned future runs indefinitely and should be spawned as a Tokio task.
+    pub async fn run_postgres_metrics_scraping(self, scrape_interval: Duration) {
```
slowli, line 312: I have first tried using a [`Collector`](https://matter-labs.github.io/vise/vise/struct.Collector.html) to scrape these metrics, but since table sizes are not changed *that* rapidly, I think that updating the corresponding metrics on a schedule would be preferable.

popzxc (approved):

perekopskiy (approved):

core/bin/external_node/src/main.rs
```diff
@@ -544,6 +544,14 @@ async fn main() -> anyhow::Result<()> {
...
         app_health.clone(),
     );
+    // Start scraping Postgres metrics before store initialization as well.
+    let metrics_pool = connection_pool.clone();
+    let mut task_handles = vec![tokio::spawn(async move {
+        metrics_pool
+            .run_postgres_metrics_scraping(Duration::from_secs(60))
+            .await;
+        Ok(())
+    })];
```
RomanBrodetski, line 554: it's now only run for EN, we need to make sure we include it when we run nodes via yaml files. Will it be a separate component? @popzxc curious how'd you build id into the Framework

RomanBrodetski (approved):

core/bin/external_node/src/main.rs
popzxc: Well, it certainly will be a task. Where would it be added? A lot of options, e.g.
1. A dedicated layer that fetches the pool as a resource and spawns a task. Maybe parametrized for a pool type.
2. Can be spawned by the pools layer directly if we want. Maybe in a form `PoolLayer::new().export_extra_metrics(true).build()`
3. Have an umbrella layer for the observability stuff (e.g. we have other tasks that mostly exist for metrics -- we can either spawn them all together or none).

Depends on how we see the use cases.

slowli: It's run on the main node as well, btw, as a part of the house keeper component.

PR #1354, opened 2024-03-05
feat(node_framework): Add timeouts for remaining tasks to avoid hang outs

AnastasiiaVashchuk:
## What ❔
We need to wrap every task future into tokio timeout when we are shutting down.

## Why ❔
Right now, the tasks are awaited, and if some task doesn't check the stop signal, it can hang the node.

## Checklist


core/node/node_framework/src/service/mod.rs
```diff
@@ -145,30 +148,43 @@ impl ZkStackService {
...
-                tracing::error!("Task {task_name} panicked");
+                tracing::error!("Task {resolved_task_name} panicked");
                 true
             }
         };
 
+        let remaining_tasks_with_timeout: Vec<_> = remaining
+            .into_iter()
+            .map(|task|
+                rt_handle.spawn(async move {
```
popzxc, line 173: `task` is already spawned via runtime, you don't really need to wrap it twice.

core/node/node_framework/src/service/mod.rs
```diff
@@ -145,30 +148,43 @@ impl ZkStackService {
...
+                    if tokio::time::timeout(TASK_SHUTDOWN_TIMEOUT, task).await.is_err()
+                    {
+                        tracing::warn!(
```
popzxc, line 176: This would put a log line per each task that timed out, which looks weird given that we don't include the task name.

core/node/node_framework/src/service/mod.rs
```diff
@@ -145,30 +148,43 @@ impl ZkStackService {
...
+                    }
+                })
+            ).collect();
+
         // Send stop signal to remaining tasks and wait for them to finish.
         // Given that we are shutting down, we do not really care about returned values.
         self.stop_sender.send(true).ok();
-        self.runtime.block_on(futures::future::join_all(remaining));
+        self.runtime
+            .block_on(futures::future::join_all(remaining_tasks_with_timeout));
```
popzxc, line 187: I think you can just wrap the tasks into `timeout` and join the wrapped futures here. Then you can iterate through the results to see if any of them is a timeout error. Then you can emit a log on timeouts.

core/node/node_framework/src/service/mod.rs
AnastasiiaVashchuk: as turned out, it is not trivial to add task name into logs here..

>This would put a log line per each task that timed out, which looks weird given that we don't include the task name.

at least we will know how much tasks reached timeout. What would u propose?

core/node/node_framework/src/service/mod.rs
AnastasiiaVashchuk: not sure I understand what you mean...

popzxc: I mean that you can
1. Wrap every remaining task into `tokio::timeout(..)`.
2. Run `join_all` here as before, but store the results, e.g. `let results = self.runtime.block_on(..)`.
3. Then `results` will have an array of `Result`s, and you can check if any of them is an `Err`, then at least one task timed out.
4. After that, you can emit a single log line stating how many tasks timed out, without mentioning their name, but also withour repeating logs.

core/node/node_framework/src/service/mod.rs
popzxc: See the comment below -- I don't propose including task names, I propose writing a log just once (and there I've described how you can do it).

AnastasiiaVashchuk: yeah...it will be at least cheaper

core/node/node_framework/src/service/mod.rs
AnastasiiaVashchuk: then it panics:
```
thread 'service::tests::test_task_run' panicked at core/node/node_framework/src/service/mod.rs:172:25:
there is no reactor running, must be called from the context of a Tokio 1.x runtime
```

prover/prover_fri/src/gpu_prover_job_processor.rs
```diff
@@ -3,8 +3,9 @@ pub mod gpu_prover {
     use std::{collections::HashMap, sync::Arc, time::Instant};
 
     use anyhow::Context as _;
-    use shivini::gpu_proof_config::GpuProofConfig;
-    use shivini::{gpu_prove_from_external_witness_data, ProverContext};
+    use shivini::{
+        gpu_proof_config::GpuProofConfig, gpu_prove_from_external_witness_data, ProverContext,
+    };
```
AnastasiiaVashchuk, line 8: zk fmt requires it

PR #1360, opened 2024-03-05
feat(Gas Oracle): Remove Fetcher singleton

jrchatruc:
## What ❔

This PR removes the `NativeTokenFetcherSingleton` as per the discussion [here](https://github.com/matter-labs/zksync-era/pull/996#discussion_r1500358343)


core/lib/zksync_core/src/native_token_fetcher/mod.rs
```diff
@@ -150,35 +82,71 @@ impl ConversionRateFetcher for NativeTokenFetcher {
...
+        Ok(())
+    }
 }
 
 #[derive(Debug)]
 struct ErrorReporter {
-    current_try: u8,
-    alert_spawned: bool,
+    current_try: AtomicU8,
+    alert_spawned: AtomicBool,
```
popzxc, line 115: Nit: I think having a single atomic variable to implement interior mutability is usually fine, but two is sus (especially given that if this would ever actually be used in a shared context, it may introduce bugs). Probably the right approach here would be to put it under `Mutex` (actually, `RefCell` should be fine since we won't hold the guard across an await point, but up to you).

jrchatruc: After looking into it a bit, I realized the way the logic worked, the `alert_spawned` boolean could be removed, as essentially all we are doing is sending an alert when the `MAX_CONSECUTIVE_NETWORK_ERRORS` threshold is reached. 

Removed it here [6e0c812](https://github.com/matter-labs/zksync-era/pull/1360/commits/6e0c81294b7f78d85d8768356ba0cd3815c82de2) so now there's just one atomic variable.

popzxc: I'm not sure if the new code is equivalent to the old behavior. Given that the variable type is u8, with polling period of 3 seconds it'll overflow in ~13 minutes and spawn an alert again. The new logic also uses non-atomic increments through plain `u8`, which means that in debug mode it'll panic (as Rust in debug mode checks for integer overflows), which is probably not desired.

I think it's better to still keep two variables and use a lock on the structure. Given that there will be only a single consumer for this lock, it's not a subject for contention and overall it's not a performance-critical part of the code.

etc/env/l1-inits/.init.env
```diff
@@ -1,29 +1,29 @@
 CONTRACTS_LATEST_PROTOCOL_VERSION=22
 CONTRACTS_CREATE2_FACTORY_ADDR=0x3c4CE2E9D8b03BbCd568E47465A483EE86893c49
-CONTRACTS_VERIFIER_ADDR=0xCf42AEFfB17dED8de2d4116D33716354F53e51e6
-CONTRACTS_L1_MULTICALL3_ADDR=0x69D85c64032337E58064A35020d6F45Dab350883
+CONTRACTS_VERIFIER_ADDR=0xF025a57cfda291db967Cbd88120fb8997232bC5E
```
popzxc, line 3: Not sure if this file was intended to be committed to the repo in the first place, looks like it should've been added to `.gitignore`.

core/lib/zksync_core/src/native_token_fetcher/mod.rs
jrchatruc: Got it, I reverted the changes and went for an `Arc<Mutex>` implementation here [fbf66b1](https://github.com/matter-labs/zksync-era/pull/1360/commits/fbf66b15e11767592b05a5f4fecd82aca26edd21)

etc/env/l1-inits/.init.env
jrchatruc: I reverted the changes to the file here [a19be60](https://github.com/matter-labs/zksync-era/pull/1360/commits/a19be607ea8b287339c040d78bc13d2ba8c8f18d), I believe it was commited in the shared bridge PR for some reason.

core/lib/zksync_core/src/native_token_fetcher/mod.rs
```diff
@@ -1,19 +1,18 @@
...
 
 use anyhow::Context;
 use async_trait::async_trait;
 use hex::ToHex;
 use metrics::atomics::AtomicU64;
-use tokio::{
-    sync::{watch, OnceCell},
-    task::JoinHandle,
-};
+use tokio::sync::Mutex;
```
popzxc, line 7: Side comment: mutexes from std are perfectly fine to use in an async context, and generally they are more performant than `tokio` mutex. Usually, you only need `tokio` mutex if you need to hold a guard across an `await` point (and even in such situations, it's often a code smell). See the [tokio docs for context](https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use).

This is a minor thing though and can be kept as is.

yarn.lock
```diff
@@ -759,10 +759,10 @@
     minimatch "^3.1.2"
     strip-json-comments "^3.1.1"
 
-"@eslint/js@8.56.0":
-  version "8.56.0"
-  resolved "https://registry.yarnpkg.com/@eslint/js/-/js-8.56.0.tgz#ef20350fec605a7f7035a01764731b2de0f3782b"
-  integrity sha512-gMsVel9D7f2HLkBma9VbtzZRehRogVRfbr++f06nL2vnCGCNlzOD+/MUov/F4p8myyAHspEhVobgjpX64q5m6A==
+"@eslint/js@8.57.0":
```
popzxc, line 762: Not sure why this file has changed. Nothing wrong with bumping dependencies, just looks a bit unrelated.
Can be kept as is.

popzxc (approved):
Nice!

PR #1367, opened 2024-03-06
feat(en): Check recipient contract and function selector in consistency checker

slowli:
## What ❔

- Checks the recipient contract (should be equal to the main L1 contract address) and the Solidity function selector (should correspond to `commitBlocks` / `commitBatches` selector) in the consistency checker.
- Reworks error handling in the consistency checker to be more maintainable.

## Why ❔

Not checking these values allows to more-or-less easily fool the checker into accepting a commit transaction which is not really a commit transaction.

## Checklist


core/lib/zksync_core/src/consistency_checker/mod.rs
```diff
@@ -358,9 +413,50 @@ impl ConsistencyChecker {
...
     }
 
+    async fn sanity_check_validator_timelock_addr(&self) -> Result<(), CheckError> {
+        let Some(address) = self.validator_timelock_addr else {
+            return Ok(());
+        };
+        tracing::debug!("Performing sanity checks for validator timelock contract {address:?}");
+
+        let call =
+            CallFunctionArgs::new("getName", ()).for_contract(address, self.contract.clone());
```
slowli, line 423: Not sure what exactly should we call in the contract; I went with the simplest option with human-readable output. 

For some reason, it looks like not all `view` functions from the contract declaration at `self.contract` (i.e., [`IZkSync`](https://github.com/matter-labs/era-contracts/blob/main/l1-contracts/contracts/zksync/interfaces/IZkSync.sol)) are actually present in the validator timelock contract; is this expected? E.g., I've initially wanted to fetch the protocol version from it (part of the `IGetters` interface), but it's not actually present.

core/lib/zksync_core/src/consistency_checker/mod.rs
```diff
@@ -28,13 +32,22 @@ mod tests;
...
-impl From<zksync_dal::SqlxError> for CheckError {
-    fn from(err: zksync_dal::SqlxError) -> Self {
-        Self::Internal(err.into())
+impl CheckError {
+    fn is_transient(&self) -> bool {
+        matches!(
+            self,
+            Self::Web3(L1ClientError::EthereumGateway(
+                web3::Error::Unreachable | web3::Error::Transport(_) | web3::Error::Io(_)
+            ))
```
slowli, line 49: Not 100% sure this matcher is exhaustive; would appreciate feedback on it.

infrastructure/zk/src/env.ts
```diff
@@ -64,11 +64,16 @@ export function set(env: string, print: boolean = false) {
...
+function loadInit(zksyncEnv: string) {
     if (fs.existsSync('etc/env/.init.env')) {
         const initEnv = dotenv.parse(fs.readFileSync('etc/env/.init.env'));
         for (const envVar in initEnv) {
             process.env[envVar] = initEnv[envVar];
+            if (zksyncEnv.startsWith('ext-node')) {
+                // We want to name env vars used by external nodes idiomatically, hence the added `EN_` prefix.
+                // The original names are still used in some integration tests, so we expose the original var as well.
+                process.env[`EN_${envVar}`] = initEnv[envVar];
+            }
```
slowli, line 76: I haven't come up with a good way to copy these variables from `.init.env` (rn, we require the only variable: `CONTRACTS_VALIDATOR_TIMELOCK_ADDR`). Theoretically, the EN could ready this env var directly using `std::env::var`, but IMO it looks too ad-hoc.

core/bin/external_node/src/config/mod.rs
```diff
@@ -237,6 +237,11 @@ pub struct OptionalENConfig {
     /// 0 means that sealing is synchronous; this is mostly useful for performance comparison, testing etc.
     #[serde(default = "OptionalENConfig::default_miniblock_seal_queue_capacity")]
     pub miniblock_seal_queue_capacity: usize,
+    /// Address of the L1 validator contract used by the consistency checker to match with the recipient of
+    /// L1 batch commit transactions. If not set, the recipient will not be verified.
+    // This is intentionally not a part of `RemoteENConfig` because fetching this info from the main node would defeat
+    // its purpose; the consistency checker assumes that the main node may provide false information.
+    pub contracts_validator_timelock_addr: Option<Address>,
```
perekopskiy, line 244: I think `contracts_validator_timelock_addr` is not the best address to check against. It can be changed, e.g. with the upcoming EIP4844 upgrade it will be changed for every env.
Instead you can check if commit transaction contains a `BlockCommit` event coming from the diamond proxy address (this address can't change!)

core/lib/zksync_core/src/consistency_checker/mod.rs
perekopskiy: Timelock validator only implements `IExecutor`. If you switch to using DiamondProxy then all getters from IGetters will be achievable

perekopskiy (approved):

infrastructure/zk/src/env.ts
RomanBrodetski: pinging @Deniallugo for visibility - how will this change in the new CLI?

RomanBrodetski (approved):
Very nice!

infrastructure/zk/src/env.ts
Deniallugo: We won't use env variables anymore.

We will just provide files with necessary date. 

but, what to do with tests is a good question

PR #1377, opened 2024-03-06
chore: Replace  test harness with just opcode_defs in constants generator

mm-zk:
## What ❔

* Replacing old evm_test_harness with opcode defs

## Why ❔

* To improve compilation speed, and not depend on unneeded methods.

### DIFF
--- a/core/bin/system-constants-generator/Cargo.toml
+++ b/core/bin/system-constants-generator/Cargo.toml
@@ -16,8 +16,8 @@ zksync_types = { path = "../../lib/types" }
 zksync_utils = { path = "../../lib/utils" }
 zksync_contracts = { path = "../../lib/contracts" }
 multivm = { path = "../../lib/multivm" }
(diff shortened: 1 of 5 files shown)


core/bin/system-constants-generator/Cargo.toml
```diff
@@ -16,8 +16,8 @@ zksync_types = { path = "../../lib/types" }
 zksync_utils = { path = "../../lib/utils" }
 zksync_contracts = { path = "../../lib/contracts" }
 multivm = { path = "../../lib/multivm" }
-zkevm_test_harness_1_3_3 = { git = "https://github.com/matter-labs/era-zkevm_test_harness.git", branch = "v1.3.3", package = "zkevm_test_harness" }
 
+zkevm_opcode_defs = { git = "https://github.com/matter-labs/era-zkevm_opcode_defs.git", branch = "v1.4.1"}
```
StanislavBreadless, line 20: you should be able to import it from `multivm::zk_evm_latest`

mm-zk: ack. good idea.

popzxc (approved):

PR #1386, opened 2024-03-08
feat(en): Enable Merkle tree client on EN

slowli:
## What ❔

Enables Merkle tree client on EN so that `zks_getProof` can work properly.

## Why ❔

This endpoint won't produce overhead and could be useful for some EN operators. Also, it allows checking Merkle tree consistency after snapshot recovery more easily.

## Checklist

### DIFF
--- a/core/lib/zksync_core/src/api_server/web3/namespaces/zks.rs
+++ b/core/lib/zksync_core/src/api_server/web3/namespaces/zks.rs
@@ -516,7 +516,6 @@ impl ZksNamespace {
             Err(TreeApiError::NotReady) => return Err(Web3Error::TreeApiUnavailable),
             Err(TreeApiError::NoVersion(err)) => {
                 return if err.missing_version > err.version_count {
-                    // FIXME: should we distinguish the case when the tree is catching up?
                     Ok(None)
                 } else {
                     Err(Web3Error::InternalError(anyhow::anyhow!(
--- a/core/lib/zksync_core/src/api_server/web3/namespaces/zks.rs
+++ b/core/lib/zksync_core/src/api_server/web3/namespaces/zks.rs
@@ -26,10 +26,7 @@ use zksync_web3_decl::{
     types::{Address, Token, H256},
 };
 
-use crate::api_server::{
-    tree::TreeApiClient,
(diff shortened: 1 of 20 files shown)


core/lib/zksync_core/src/api_server/web3/namespaces/zks.rs
```diff
@@ -499,20 +499,33 @@ impl ZksNamespace {
...
-            .context("get_proofs")?;
+            .as_deref()
+            .ok_or(Web3Error::TreeApiUnavailable)?;
+        let proofs_result = tree_api.get_proofs(l1_batch_number, hashed_keys).await;
+        let proofs = match proofs_result {
+            Ok(proofs) => proofs,
+            Err(TreeApiError::NotReady) => return Err(Web3Error::TreeApiUnavailable),
+            Err(TreeApiError::NoVersion(err)) => {
+                return if err.missing_version > err.version_count {
+                    // FIXME: should we distinguish the case when the tree is catching up?
```
slowli, line 519: We could fetch the latest L1 batch number from Postgres and compare it to `err.version_count`. If it's greater, then the Merkle tree is catching up. Not sure if it makes sense from the DevX perspective, or would just create an additional error that clients would need to handle.

popzxc: I think we can leave it as is and maybe revisit it later if someone complains.

core/node/node_framework/src/implementations/layers/web3_api/server.rs
```diff
@@ -113,15 +113,21 @@ impl WiringLayer for Web3ServerLayer {
         let sync_state = match context.get_resource::<SyncStateResource>().await {
             Ok(sync_state) => Some(sync_state.0),
             Err(WiringError::ResourceLacking(_)) => None,
-            Err(err) => {
-                return Err(err);
-            }
+            Err(err) => return Err(err),
+        };
+        let tree_api_client = match context.get_resource::<TreeApiClientResource>().await {
```
popzxc, line 118: ❤️

popzxc (approved):

popzxc (approved):

PR #1388, opened 2024-03-08
chore: Update readme for key generator

mm-zk:
## What ❔

* Fixing command line arguments in key-generator README file.

### DIFF
--- a/prover/vk_setup_data_generator_server_fri/README.md
+++ b/prover/vk_setup_data_generator_server_fri/README.md
@@ -12,7 +12,7 @@ The current set of verification keys is committed under 'data/' directory. If yo
 circuit changes), first please make sure that you have a CRS file (used for SNARK key), and then you can run:
 
 ```shell
-CRS_FILE=yyy ZKSYNC_HOME=xxx cargo run --release --bin key_generator generate-vk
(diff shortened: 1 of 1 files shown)


qingshui-btc:
Usage: keygen.py [OPTIONS]

Options:
  -h, --help           Show this message and exit.
  -n, --name TEXT      The name of the key holder.
  -p, --passphrase     Use a passphrase to encrypt the private key.
  -o, --output PATH    The output file path.

prover/vk_setup_data_generator_server_fri/README.md
```diff
@@ -30,23 +32,23 @@ You can run it with `dry-run`, to see the results, or set `dry-run` to false to
...
+ZKSYNC_HOME=xxx cargo run --release --bin key_generator update-commitments --dry-run=true --path vk_setup_data_generator_server_fri/data/
 ```
 
 ## Setup keys
 
 Setup keys are used when you run the actual prover. They are around 15GB for each circuit type, and we have different
 setup keys for GPU vs CPU prover.
 
-For example, the command below will generate the setup keys for the basic circuit of type 3.
+For example, the command below will generate the setup keys for the basic circuit of type 3 and put then in setup path directory.
```
popzxc, line 43: @mm-zk I think you should run `zk fmt` to make the CI pass.

PR #1393, opened 2024-03-08
feat(Validium): unit tests for `ensure_l1_batch_commit_data_generation_mode`

Oppen:
## What ❔

1. Add a set of unit tests for `ensure_l1_batch_commit_data_generation_mode`, making sure the right kind of errors or lack thereof are returned for a given set of parameters and L1 responses. It required a mock for `EthInterface` to return the needed values.
2. Annotate `ensure_l1_batch_commit_data_generation_mode` to make sure we use the result, otherwise we may end up in inconsistent state.

## Why ❔

QA

## Checklist

### DIFF
--- a/core/lib/zksync_core/src/utils/mod.rs
+++ b/core/lib/zksync_core/src/utils/mod.rs
@@ -215,6 +215,8 @@ pub async fn ensure_l1_batch_commit_data_generation_mode(
 
 #[cfg(test)]
 mod tests {
+    use std::sync::Mutex;
+
     use zksync_eth_client::{ContractCall, ExecutedTxStatus, FailureInfo, RawTransactionBytes};
     use zksync_types::{
         web3::types::{Block, BlockId, BlockNumber, Filter, Log, Transaction, TransactionReceipt},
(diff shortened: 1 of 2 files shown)


core/lib/zksync_core/src/utils/mod.rs
```diff
@@ -262,4 +269,264 @@ mod tests {
...
+            _: &'static str,
+        ) -> Result<Option<Block<H256>>, EthClientError> {
+            unimplemented!("Not needed");
+        }
+    }
+
+    #[tokio::test]
+    async fn ensure_l1_batch_commit_data_generation_mode_succeeds_when_both_match() {
+        let addr = Address::repeat_byte(0x01);
+        assert!(ensure_l1_batch_commit_data_generation_mode(
```
popzxc, line 423: Nit: here and in other places it'll probably be nicer if you'll use `assert_matches!` (it should already be in dev deps) to actually match on output.

Oppen: Sorry, I didn't check if we already had the dependency and didn't want to add it. I'll change it.

Oppen: OTOH, I'm not sure `anyhow` makes it easy to check which error was returned. I need to check that out first.

core/lib/zksync_core/src/utils/mod.rs
```diff
@@ -178,6 +178,7 @@ async fn get_pubdata_pricing_mode(
     eth_client.call_contract_function(args).await
 }
 
+#[must_use]
```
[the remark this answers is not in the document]
Oppen: I didn't realize that. It's also true for `Result<_>` types. I'll revert the commit.

core/lib/zksync_core/src/utils/mod.rs
```diff
@@ -262,4 +269,264 @@ mod tests {
             .unwrap();
         assert_eq!(l1_batch, None);
     }
+
+    #[derive(Debug)]
+    struct MockEthereumForCommitGenerationMode {
+        retval: Vec<ethabi::Token>,
+        // Can't copy `Error` so I need a different way to flag it
```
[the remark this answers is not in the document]
Oppen: I applied a variant of it. Because `EthInterface` requires the struct to be `Send` and `Sync` I used a `Mutex` and a `swap` instead.

core/lib/zksync_core/src/utils/mod.rs
Oppen: I applied it for the `Ok(())` cases, but it seems unfeasible for error cases because the fields seem to be private.

core/lib/zksync_core/src/utils/mod.rs
```diff
@@ -262,4 +271,271 @@ mod tests {
...
+        async fn eth_balance(&self, _: H160, _: &'static str) -> Result<U256, EthClientError> {
+            unimplemented!("Not needed");
+        }
+
+        async fn call_contract_function(
+            &self,
+            _: ContractCall,
+        ) -> Result<Vec<ethabi::Token>, EthClientError> {
+            let mut error = None;
+            core::mem::swap(&mut *self.error.lock().unwrap(), &mut error);
```
popzxc, line 401: Funnily that's roughly how `Option::take` is [implemented under the hood](https://doc.rust-lang.org/src/core/option.rs.html#1688). You should still be able to use `take` with a mutex, but nbd.

popzxc (approved):

core/lib/zksync_core/src/utils/mod.rs
Oppen: I tried to use `take` but it requires the inner type to implement `Default`.

Oppen: Though that was with `mem::take`. Maybe `Option::take` works.

PR #1398, opened 2024-03-11
feat(node_framework): Support for preconditions and oneshot tasks

popzxc:
This PR consists of three parts in one (sorry for that).

## Part 1. Preconditions

An alternative take on https://github.com/matter-labs/zksync-era/pull/1290

In short:

- Adds two new concepts to the framework: `Precondition` and `UnconstrainedTask`.
- Any precondition is a future that must successfully resolve before tasks can be started. An example of future precondition may be the readiness of storage after snapshot sync.
- Any `Task` will wait until all of the added preconditions successfully resolves.
- An `UnconstrainedTask` is a way to opt-out from waiting for preconditions to complete. These may be useful in two cases:
  * If you have a stateless component that should start ASAP (e.g. healthcheck server).
  * If your task is supposed to ensure that precondition is met (e.g. task that does state recovery).
 
The last point is important, btw: imagine that the node runs as a set of microservices on different machines. The precondition should be present on *all* machines (e.g. simply wait until it's ensured), but the corresponding unconstrained task should be run on *just one* machine (since it alters state).

In the current impl there is no way to subscribe to a certain set of preconditions or opt-out from a single precondition. You either have to wait for them all or may rely on none. This is done on purpose to not overcomplicate the publicly facing interface. The expectation is that `Task` is a working default that people may just pick up and it'll be guaranteed to work. You only need to meet extra complexity when a `Task` is not enough for you.

## Part 2. Oneshot tasks

Adds a notion of oneshot tasks and adds support of running the service in an oneshot mode (e.g. when we *only* add oneshot tasks and expect the node to exit as soon as *all of them* resolve).

It comes in two flavors:

- `OneshotTask`. An oneshot analog of `Task`.
- `UnconstrainedOneshotTask`. An oneshot analog of `UnconstrainedTask`. May be useful, for example, to enforce some `Precondition` and then exit.

This way the logical flow of the node is as follows:

1. Start all the unconstrained tasks, unconstrained oneshot tasks, and preconditions.
2. Wait for all the preconditions to resolve.
3. Start all the tasks and oneshot tasks.
4. Wait for any of long-running tasks (i.e. *not* oneshot) to exit.

## Part 3. Refactoring

There was a lot of new logic added, so the `ZkSyncService::run` became pretty messy.
The most notable part here is that now all flavors of tasks are managed by the new `Runnables` collection that encapsulates the logic of turning different tasks into futures that can be launched by the service.


core/node/node_framework/src/service/mod.rs
```diff
@@ -114,55 +124,121 @@ impl ZkStackService {
...
         }
+        tracing::info!("Wiring complete");
+
+        // Launch precondition checkers. Generally, if all of them succeed, we don't need to do anything.
+        // If any of them fails, the node have to shut down.
+        // So we run them in a detached task with an oneshot channel to report the error, if it occurs,
+        // and we'll use this channel in the same way that we use to check if any of the task exits.
+        // The difference is that preconditions, unlike tasks, are allowed to exit without causing the node
+        // to shut down.
+        let (precondition_fail_sender, precondition_fail_receiver) = oneshot::channel();
```
slowli, line 185: Dumb question: Wouldn't retaining the handle to the task spawned below work as well?

```rust
let preconditions_result = self.runtime.spawn(async move {
    // snipped
});
let precondition_system_task = Box::pin(async move {
    match preconditions_result.await {
        // ...
    }
});
```

core/node/node_framework/src/service/mod.rs
```diff
@@ -114,55 +124,121 @@ impl ZkStackService {
...
+                }
+            }
+            // Note that we don't have to `select` on the stop signal explicitly:
+            // Each prerequisite is given a stop signal, and if everyone respects it, this future
+            // will still resolve once the stop signal is received.
+        });
+        tasks.push(precondition_system_task);
 
         // Prepare tasks for running.
         let rt_handle = self.runtime.handle().clone();
```
slowli, line 224: Nit: `runtime_handle` is defined above, so this looks redundant.

slowli (commented):
Conceptually looks good. I wonder whether it could make sense to generalize preconditions to oneshot tasks that could be used for programmatic migrations etc. Can be done later, obviously.

core/node/node_framework/src/service/mod.rs
```diff
@@ -114,55 +124,121 @@ impl ZkStackService {
...
+        // to shut down.
+        let (precondition_fail_sender, precondition_fail_receiver) = oneshot::channel();
+        self.runtime.spawn(async move {
+            let preconditions = preconditions.into_iter().map(|fut| async move {
+                // Spawn each precondition as a separate task.
+                // This way we can handle the cases when a precondition task panics and propagate the message
+                // to the service.
+                // This is important since this task is detached: we do not store a handle for it and never await it explicitly.
+                let handle = tokio::runtime::Handle::current();
+                match handle.spawn(fut).await {
```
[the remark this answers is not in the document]
popzxc: On using a handle: over time, I kind of came to the conclusion that if spawning panics, it's a bit more obvious when it panics when you try obtain the handle, e.g. separation of concepts for getting the handle and actually spawning the future. It's a minor thing and mostly my preference, but can change if you think it's important.

On `catch_unwind` -- I've actually tried using it first, but unfortunately it requires an [UnwindSafe](https://doc.rust-lang.org/nightly/core/panic/unwind_safe/trait.UnwindSafe.html) bound on the futures. When I played with it, I met a few compiler errors when this condition was not met, and they were extremely unfriendly to non-async-native people, so adding an `UnwindSafe` bound to the precondition doesn't seem to be a good solution because of the complexity it bears.

popzxc:
> I wonder whether it could make sense to generalize preconditions to oneshot tasks that could be used for programmatic migrations etc. Can be done later, obviously.

@slowli Yup, that's exactly my plan, but indeed I want to tackle it in a separate PR.

core/node/node_framework/src/service/mod.rs
popzxc: Not dumb at all; it should work indeed, I'll check it.

popzxc: Actually, now that we're wrapping each precondition in its own tokio task, it seems that we don't need a detached task at all.

core/node/node_framework/src/task.rs
```diff
@@ -23,18 +55,97 @@ pub trait Task: 'static + Send {
...
+    fn name(&self) -> &'static str;
+
+    /// Runs the task without waiting for any precondition to be met.
+    async fn run_unconstrained(self: Box<Self>, stop_receiver: StopReceiver) -> anyhow::Result<()>;
+}
+
+/// An unconstrained analog of [`OneshotTask`].
+/// See [`UnconstrainedTask`] and [`OneshotTask`] for more details.
+#[async_trait::async_trait]
+pub trait UnconstrainedOneshotTask: 'static + Send {
```
slowli, line 142: Maybe a dumb suggestion: Could it make sense to have the base trait for all task types (potentially including preconditions) and use marker traits for each of the type? Looks like we have quite a few of task traits, but their signatures are almost identical. Yet another suggestion: add `fn task_type(&self) -> TaskType` to the task trait, where `TaskType` is defined like

```rust
enum TaskType {
    Normal,
    Oneshot,
    Unconstrained { oneshot: bool },
}
```

core/node/node_framework/src/service/mod.rs
```diff
@@ -187,37 +177,46 @@ impl ZkStackService {
...
+            // Spawn each oneshot task as a separate tokio task.
+            // This way we can handle the cases when such a task panics and propagate the message
+            // to the service.
+            let handle = tokio::runtime::Handle::current();
+            match handle.spawn(fut).await {
+                Ok(Ok(())) => Ok(()),
+                Ok(Err(err)) => Err(err),
+                Err(panic_err) => {
+                    let panic_msg = try_extract_panic_message(panic_err);
+                    Err(anyhow::format_err!("Precondition panicked: {panic_msg}"))
```
slowli, line 200: Nit: Misleading error message; AFAIU, the run tasks are not necessarily preconditions (maybe, it could make sense to pass around the task kind to make messages like this more informative?).

core/node/node_framework/src/task.rs
popzxc: Yeah, it was a deliberate decision. Reasons I had in mind for that:
1. I have heard several times already that people are concerned regarding having complex concepts in the framework. So _probably_ 4 simple concepts are actually easier to grok than one customizable (e.g. you get to choose one trait and then it's simple from there).
2. In fact, most of the time people want to use `Task`. So having it as a small trait with only mandatory fields feels like a plus: people are not distracted from what they want to do.
3. Base traits make the code harder to read, and I want (all) people to be able to read the framework code. E.g. write code in a "stupid" way where stupidity means easiness of access.
4. ...ultimately, it doesn't add too much complexity to the framework. Just a bit of boilerplate.

Or tl;dr I just thought that "stupid" way is more friendly for the people who will use the framework.

But indeed, I've spent some time thinking about it and not 100% decided, so will be happy to discuss.

core/node/node_framework/src/service/mod.rs
popzxc: Thanks, leftover from times when this code was only meant for preconditions.
Passing the task type maybe makes sense, but I think for the 1st iteration it should be sufficient that the panic backtrace will anyway be available (and for prod environment it'll anyway be the thing that we see first), so I'll leave it mostly as is.

AnastasiiaVashchuk (approved):

popzxc:
@slowli I'm going to merge this to not block any further progress, but if you have any comments -- I'll be happy to work on them in follow-up PRs.

PR #1401, opened 2024-03-11
feat(db): Configurable maximum number of open RocksDB files

vbar:
## What ❔

Added a field to RocksDBOptions, which is used as an argument to rocksdb::Options::set_max_open_files().

## Why ❔

This allows Merkle tree users to limit the number of open files required by RocksDB, which otherwise grows beyond normal OS limits (8192 files) and requires large amounts of memory (currently, beyond 32GB).

## Checklist

  - current (integration) tests run as before
  - external program requiring the change is now running
  - all new fields and methods are documented
  - no links added (nor modified)

### DIFF
--- a/core/lib/merkle_tree/src/storage/rocksdb.rs
+++ b/core/lib/merkle_tree/src/storage/rocksdb.rs
@@ -3,7 +3,7 @@
 use std::path::Path;
 
 use rayon::prelude::*;
-use zksync_storage::{db::NamedColumnFamily, rocksdb, rocksdb::DBPinnableSlice, RocksDB};
+use zksync_storage::{db::{NamedColumnFamily, RocksDBOptions}, rocksdb, rocksdb::DBPinnableSlice, RocksDB};
 
 use crate::{
     errors::{DeserializeError, ErrorContext},
--- a/core/lib/merkle_tree/src/storage/rocksdb.rs
+++ b/core/lib/merkle_tree/src/storage/rocksdb.rs
@@ -3,7 +3,7 @@
 use std::path::Path;
 
 use rayon::prelude::*;
-use zksync_storage::{db::{NamedColumnFamily, RocksDBOptions}, rocksdb, rocksdb::DBPinnableSlice, RocksDB};
+use zksync_storage::{db::NamedColumnFamily, rocksdb, rocksdb::DBPinnableSlice, RocksDB};
 
 use crate::{
     errors::{DeserializeError, ErrorContext},
--- a/core/lib/merkle_tree/src/storage/rocksdb.rs
+++ b/core/lib/merkle_tree/src/storage/rocksdb.rs
@@ -74,15 +74,6 @@ impl RocksDBWrapper {
         Ok(Self::from(RocksDB::new(path)?))
     }
 
-    /// Creates a new wrapper, initializing RocksDB with explicit options.
-    ///
-    /// # Errors
-    ///
-    /// Propagates RocksDB I/O errors.
-    pub fn with_options(path: &Path, options: RocksDBOptions) -> Result<Self, rocksdb::Error> {
-        Ok(Self::from(RocksDB::with_options(path, options)?))
-    }
-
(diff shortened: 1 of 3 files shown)


core/lib/merkle_tree/src/storage/rocksdb.rs
```diff
@@ -74,6 +74,15 @@ impl RocksDBWrapper {
         Ok(Self::from(RocksDB::new(path)?))
     }
 
+    /// Creates a new wrapper, initializing RocksDB with explicit options.
+    ///
+    /// # Errors
+    ///
+    /// Propagates RocksDB I/O errors.
+    pub fn with_options(path: &Path, options: RocksDBOptions) -> Result<Self, rocksdb::Error> {
```
slowli, line 82: Bikeshedding: This constructor looks unnecessary, given that its implementation is fully public, so you could use it instead.

core/lib/storage/src/db.rs
```diff
@@ -274,6 +274,8 @@ pub struct RocksDBOptions {
     /// Timeout to wait for the database to run compaction on stalled writes during startup or
     /// when the corresponding RocksDB error is encountered.
     pub stalled_writes_retries: StalledWritesRetries,
+    /// Number of open files that can be used by the DB. Default is -1, for no limit.
+    pub max_open_files: i32,
```
slowli, line 278: The fact that RocksDB uses `i32` for this param doesn't mean that it makes sense to use it in Rust 🙃 I'd rather use a type that makes sense from the Rust type system perspective, e.g. `Option<NonZeroU32>` (IMO, it would be OK to saturate values over `i32::MAX`, or you could error in the `RocksDB` constructor if such a value is provided.)

core/lib/storage/src/db.rs
```diff
@@ -304,7 +307,8 @@ impl<CF: NamedColumnFamily> RocksDB<CF> {
 
     pub fn with_options(path: &Path, options: RocksDBOptions) -> Result<Self, rocksdb::Error> {
         let caches = RocksDBCaches::new(options.block_cache_capacity);
-        let db_options = Self::rocksdb_options(None, None);
+        let mut db_options = Self::rocksdb_options(None, None);
+        db_options.set_max_open_files(options.max_open_files);
```
slowli, line 311: Just to be sure: is this setting enforced on the database level, rather than on the column family level (options for CFs are created below)?

core/lib/merkle_tree/src/storage/rocksdb.rs
vbar: OK

core/lib/storage/src/db.rs
vbar: Erroring out is apparently impossible w/o changing the signature (rocksdb::Error cannot be constructed), so saturating...

core/lib/storage/src/db.rs
vbar: Yes - actually I'd prefer it on process level (in case the process has multiple RocksDB instances), but that's probably too complex...

core/lib/storage/src/db.rs
```diff
@@ -304,7 +308,13 @@ impl<CF: NamedColumnFamily> RocksDB<CF> {
 
     pub fn with_options(path: &Path, options: RocksDBOptions) -> Result<Self, rocksdb::Error> {
         let caches = RocksDBCaches::new(options.block_cache_capacity);
-        let db_options = Self::rocksdb_options(None, None);
+        let mut db_options = Self::rocksdb_options(None, None);
+        let max_open_files = if let Some(non_zero) = options.max_open_files {
+            i32::try_from(non_zero.get()).ok().unwrap_or(i32::MAX)
```
slowli, line 313: Nit: You can just use `i32::try_from(non_zero.get()).unwrap_or(i32::MAX)`; converting `Result` to an `Option` isn't necessary.

slowli (dismissed):

PR #1406, opened 2024-03-11
test: add tests for `StorageTransaction` conversions

ischasny:
Add tests for conversion of StorageTransaction to different transaction types.


core/lib/dal/src/models/tests.rs
```diff
@@ -0,0 +1,356 @@
...
+use zksync_utils::bigdecimal_to_u256;
+
+use crate::{models::storage_transaction::StorageTransaction, BigDecimal};
+
+/// Creates an "empty" transaction with all values set to None / zeros / left empty etc.
+///
+/// Used by other tests as a template to create transactions of a specific type.
+fn empty_tx() -> StorageTransaction {
+    StorageTransaction {
+        gas_limit: None,
```
montekki, line 19: I suspect that you can just return `StorageTransaction::default()` here, it should be the same.

ischasny: There seem to be no `default` method there - it doesn't implement `Default` trait

montekki: I would suggest to add such a derive and it can even be gated under `#[cfg(test)]` since it looks like the crate is the same and it would work to save the boilerplate.

ischasny: Done

montekki (approved):

PR #1410, opened 2024-03-12
fix(reth): use reth instead of geth

Deniallugo:
## Checklist


core/lib/eth_client/src/clients/http/query.rs
```diff
@@ -129,8 +129,17 @@ impl EthInterface for QueryClient {
             .web3
             .eth()
             .block(BlockId::Number(BlockNumber::Pending))
-            .await?
-            .expect("Pending block should always exist");
+            .await?;
+        let block = if let Some(block) = block {
+            block
+        } else {
+            // fallback for local testing
```
popzxc, line 136: Pls elaborate what's the difference between local and non-local L1 envs.

popzxc (commented):
Nice!

docker-compose-gpu-runner-cuda-12-0.yml
```diff
@@ -1,14 +1,27 @@
...
+    image: "ghcr.io/paradigmxyz/reth:v0.2.0-beta.1"
+    volumes:
+      - type: bind
+        source: ./volumes/reth/data
+        target: /rethdata
+      - type: bind
+        source: ./volumes/reth/logs
+        target: /rethlogs
+      - type: bind
+        source: ./etc/test_config/
```
popzxc, line 14: IIRC `test_config` was introduced as a folder for some test configurations. Keeping reth configuration here is counterintuitive.

I suggest doing the following:
1. Create folder `/etc/reth`
2. Create `/etc/reth/README.md` with a couple of sentences on what this folder represents (e.g. data to run reth locally, including folders and files mounted to the reth container as read-only volumes). Please check it with grammarly and check that it doesn't look clunky😅 
3. Create `/etc/reth/chaindata` with the contents that are intended to be mounted to reth.
4. Here and in other docker-compose files use `/etc/reth/chaindata`.

docker-compose-gpu-runner-cuda-12-0.yml
```diff
@@ -1,14 +1,27 @@
...
+  reth:
+    restart: always
+    image: "ghcr.io/paradigmxyz/reth:v0.2.0-beta.1"
+    volumes:
+      - type: bind
+        source: ./volumes/reth/data
+        target: /rethdata
+      - type: bind
+        source: ./volumes/reth/logs
+        target: /rethlogs
```
popzxc, line 12: Do we need to bind it? IIRC for non-mounted directories, an ephemeral storage would be created that is dropped once the container exits. This seems to be fine for reth logs -- I doubt we'll have to check them often.

core/lib/eth_client/src/clients/http/query.rs
```diff
@@ -129,8 +129,18 @@ impl EthInterface for QueryClient {
...
             .eth()
             .block(BlockId::Number(BlockNumber::Pending))
-            .await?
-            .expect("Pending block should always exist");
+            .await?;
+        let block = if let Some(block) = block {
+            block
+        } else {
+            // Fallback for local reth. Because of artificial nature of producing blocks in local reth setup
+            // pending block could be not yet exists
```
popzxc, line 137: "there may be no pending block".

popzxc (commented):
Mostly LGTM

perekopskiy (approved):

PR #1411, opened 2024-03-12
feat(state-keeper): Refactor persistence in `StateKeeper`

slowli:
## What ❔

- Splits off miniblock and L1 batch persistence from `StateKeeperIO` into a separate trait.
- Moves tracking current miniblock / L1 batch number from `StateKeeperIO` to the state keeper. 

## Why ❔

- Simplifies further changes to the persistence logic, e.g. toggling saving protective reads.
- Can be used as a building block for generic retrospective L1 batch execution.

## Checklist


core/lib/zksync_core/src/state_keeper/io/mod.rs
```diff
@@ -96,209 +136,107 @@ pub trait StateKeeperIO: 'static + Send + IoSealCriteria {
...
-impl fmt::Debug for dyn StateKeeperIO {
-    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
-        formatter
-            .debug_struct("StateKeeperIO")
-            .field("current_l1_batch_number", &self.current_l1_batch_number())
-            .field("current_miniblock_number", &self.current_miniblock_number())
-            .finish()
+/// Handler for state keeper outputs (miniblocks and L1 batches).
+#[async_trait]
+pub trait HandleStateKeeperOutput: 'static + Send + fmt::Debug {
```
popzxc, line 159: Nit: I remember that there is a suggestion to [use imperatives for traits](https://github.com/rust-lang/api-guidelines/discussions/28), but it isn't universally followed even in std (e.g. `Iterator`, `Future`, etc). 

To me, it feels like an implicit convention is as follows:
- If there is a single method that majorly describes the behavior of a trait, it should be used as the name of a trait (`clone`, `from`, `into`, etc).
- Otherwise, the trait is treated as an interface in a more general sense and should be named as you would name an interface in any other language (excluding language-specific rules, like the `I` prefix).

In this particular case, the naming seems to be artificially affected and not really natural.  IMHO, `StateKeeperOutputHandler` reads _much_ more naturally.

Not a strong opinion tho.

core/lib/zksync_core/src/state_keeper/io/mod.rs
```diff
@@ -96,209 +136,107 @@ pub trait StateKeeperIO: 'static + Send + IoSealCriteria {
...
 }
 
-/// Handle for [`MiniblockSealer`] allowing to submit [`MiniblockSealCommand`]s.
 #[derive(Debug)]
-pub struct MiniblockSealerHandle {
-    commands_sender: mpsc::Sender<Completable<MiniblockSealCommand>>,
-    latest_completion_receiver: Option<oneshot::Receiver<()>>,
-    // If true, `submit()` will wait for the operation to complete.
-    is_sync: bool,
+pub struct CompoundOutputHandler {
```
popzxc, line 180: Nit: I like the concept, but I'm not sure if it's a good thing that you have two ways to achieve the same thing (e.g. by providing `Box<dyn HandleStateKeeperOutput>` directly or creating `CompoundOutputHandler` with no `with_aux_handler` invocation.

What if we will:
1. Rename this structure to simply `OutputHandler`.
2. Make `ZkSyncStateKeeper` accept `OutputHandler` instead of `Box<dyn HandleStateKeeperOutput>`

This way this structure will become a single _primary_ concept. Moreover, in the docs for it we'll be able to cover *both* the concept of `HandleStateKeeperOutput` and composability, without making people to find out about these separately.

(at the same time, internally, the state keeper can still have `Box<dyn HandleStateKeeperOutput>`)

popzxc (commented):
Only had time to look at a part of the PR.
Generally it looks very promising, will continue review on Monday (but please don't wait for me -- I may have a look retroactively).

core/lib/zksync_core/src/state_keeper/io/mod.rs
slowli: I've had similar thoughts, will do.

core/lib/zksync_core/src/state_keeper/io/mod.rs
```diff
@@ -59,12 +59,12 @@ pub struct MiniblockParams {
...
-#[derive(Debug)]
-pub struct L1BatchParams {
+#[derive(Debug, Clone)]
+pub struct L1BatchParams<H = H256> {
     /// Protocol version for the new L1 batch.
     pub(crate) protocol_version: ProtocolVersionId,
-    /// State hash of the previous L1 batch.
-    pub(crate) previous_batch_hash: H256,
+    /// State hash of the previous L1 batch. `()` means that it's not necessarily known at this point.
+    pub(crate) previous_batch_hash: H,
```
popzxc, line 67: Subjective -1 on this one. It seems to be a fancy way to avoid using `Option` here, but at the same time, this approach isn't common in the workspace. Given that the level of Rust expertise varies throughout the team significantly, I'm afraid that it may be confusing to some people, so personally I'd stick with a simpler approach.

core/lib/zksync_core/src/state_keeper/io/mod.rs
```diff
@@ -99,6 +99,19 @@ impl L1BatchParams {
     }
 }
 
+impl L1BatchParams<()> {
+    pub(crate) fn with_previous_batch_hash(self, previous_batch_hash: H256) -> L1BatchParams {
```
popzxc, line 103: ...to the point above: especially this transition may be perceived as weird (e.g. the fact that the type of `self` is not `L1BatchParams` (which is not the same as `L1BatchParams<()>`)).

core/lib/zksync_core/src/state_keeper/io/persistence.rs
```diff
@@ -0,0 +1,453 @@
...
+struct Completable<T> {
+    command: T,
+    completion_sender: oneshot::Sender<()>,
+}
+
+/// Canonical [`HandleStateKeeperOutput`] implementation that stores processed miniblocks and L1 batches to Postgres.
+#[derive(Debug)]
+pub struct StateKeeperPersistence {
+    pool: ConnectionPool,
+    object_store: Option<Arc<dyn ObjectStore>>, // FIXME: split off?
```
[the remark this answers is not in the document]
popzxc: Fully agree, cc @EmilLuta

core/lib/zksync_core/src/state_keeper/keeper.rs
```diff
@@ -297,25 +297,50 @@ impl ZkSyncStateKeeper {
...
+            .load_base_system_contracts(params.protocol_version, cursor)
+            .await
+            .with_context(|| {
+                format!(
+                    "failed loading system contracts for protocol version {:?}",
+                    params.protocol_version
+                )
+            })?;
+        Ok(params.into_env(self.io.chain_id(), contracts, cursor))
+    }
```
popzxc, line 333: This method feels a bit out of place for the state keeper: we only use `IoCursor` (which is passed to IO in other methods already) and `self.io` here.
Moreover, `wait_for_new_batch_params` already receives `cursor` as an argument.

Maybe it's better to implement this logic as a part of IO trait?

popzxc (commented):
Generally looks good, a few more minor comments.

perekopskiy (approved):

popzxc (dismissed):

PR #1414, opened 2024-03-12
updated keys

mm-zk:
Updated keys.


prover/vk_setup_data_generator_server_fri/data/finalization_hints_basic_7.bin
koalified-gachinator, line 1: is this expected to add new file here?

PR #1421, opened 2024-03-13
feat: Decide when to send proofs based on database and not based on GCS contents

mm-zk:
## What ❔

* Look at the database (`proof_generation_details` table), rather than on GCS contents when selecting proofs to send to l1
* Removed `proof_loading_mode` config option (no longer needed)

## Why ❔

* This helps during emergencies - if we load proofs based on database, we no longer have to worry about clearing the GCS directory - and we can focus only the database, which becomes source of truth.

### DIFF
--- a/core/lib/dal/src/proof_generation_dal.rs
+++ b/core/lib/dal/src/proof_generation_dal.rs
@@ -183,4 +183,34 @@ impl ProofGenerationDal<'_, '_> {
 
         result
     }
+
+    /// Checks if proof is ready and return proof's blob url.
+    /// There are 3 possible outcomes:
+    ///     None -> means that proof is not ready.
+    ///     Some(None) -> means that proof was skipped
+    ///     Some("path") -> means that proof was processed and is available at path.
+    pub async fn get_proof_blob_url(
+        &mut self,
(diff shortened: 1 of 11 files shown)


core/lib/dal/src/proof_generation_dal.rs
```diff
@@ -183,4 +183,34 @@ impl ProofGenerationDal<'_, '_> {
 
         result
     }
+
+    /// Checks if proof is ready and return proof's blob url.
+    /// There are 3 possible outcomes:
+    ///     None -> means that proof is not ready.
+    ///     Some(None) -> means that proof was skipped
+    ///     Some("path") -> means that proof was processed and is available at path.
```
RomanBrodetski, line 191: I would probably create a data type for this, but feel free to leave as is - the return type is clear from the comment

RomanBrodetski (commented):
Looks good!

core/lib/dal/src/proof_generation_dal.rs
perekopskiy: > I would probably create a data type for this

+1

PR #1423, opened 2024-03-13
feat(genesis): Using genesis config only during the genesis

Deniallugo:
## What ❔

During the genesis, we have some fixed params, and they are changing only during protocol upgrade. In this PR, I've not created an additional file that represents genesis state, but only loaded it from environmental variables.  

## Why ❔

For system config refactoring we have to split our config system to a few different files and it's the first step to it

## Checklist


montekki (approved):

core/lib/zksync_core/src/genesis.rs
```diff
@@ -32,67 +36,144 @@ use zksync_utils::{be_words_to_bytes, bytecode::hash_bytecode, h256_to_u256, u25
...
+            .recursion_circuits_set_vks_hash,
+        recursion_scheduler_level_vk_hash: first_l1_verifier_config
+            .recursion_scheduler_level_vk_hash,
     }
+}
 
+// This function is dangerous,
+// it doesn't enforce genesis correctness
+// Please, always use `ensure_genesis_state` instead
+pub async fn ensure_genesis_state_unchecked(
```
popzxc, line 154: I think the notion of the "dangerous function" comes from a fact that it isn't properly named.
It should be named `insert_genesis_batch` and then there will be no expectations for this function to ensure anything.

core/lib/zksync_core/src/genesis.rs
```diff
@@ -32,67 +36,144 @@ use zksync_utils::{be_words_to_bytes, bytecode::hash_bytecode, h256_to_u256, u25
...
-    storage: &mut StorageProcessor<'_>,
-    zksync_chain_id: L2ChainId,
-    genesis_params: &GenesisParams,
-) -> anyhow::Result<H256> {
-    let mut transaction = storage.start_transaction().await?;
+    pub fn protocol_version(&self) -> ProtocolVersionId {
+        // It's impossible to instantiate Genesis params with wrong protocol version
+        self.config.protocol_version.try_into().expect(
+            "Protocol version must be correctly \
+        initialized for genesis",
```
popzxc, line 119: Can be a single line.

popzxc (commented):
Conceptually I like the idea.

core/lib/dal/src/models/storage_block.rs
```diff
@@ -345,6 +345,8 @@ pub struct StorageL1BatchDetails {
     pub l2_fair_gas_price: i64,
     pub bootloader_code_hash: Option<Vec<u8>>,
     pub default_aa_code_hash: Option<Vec<u8>>,
+    pub commitment: Option<Vec<u8>>,
+    pub rollup_last_leaf_index: Option<i64>,
 }
```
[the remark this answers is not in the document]
Deniallugo: Because we need to have a proper genesis for external nodes and we have to expose the necessary data for it

core/lib/zksync_core/src/genesis.rs
```diff
@@ -32,67 +36,141 @@ use zksync_utils::{be_words_to_bytes, bytecode::hash_bytecode, h256_to_u256, u25
...
+    #[error("Base system contracts mismatched: From Config: {0:?}, Calculated {1:?}")]
+    BaseSystemContractsHashes(BaseSystemContractsHashes, BaseSystemContractsHashes),
+    #[error("Commitment mismatched: From Config: {0:?}, Calculated {1:?}")]
+    Commitment(H256, H256),
+    #[error("Wrong protocol version")]
+    ProtocolVersion,
+    #[error("Error: {0}")]
+    Other(#[from] anyhow::Error),
+    #[error("DB Error: {0}")]
+    DBError(#[from] SqlxError),
```
popzxc, line 54: I quickly checked and it looks like that it's only used for the DAL invocations, it's just in some places we add `context` and in some places we don't.
If that's true, then you can do the following:

1. Remove `Other`
2. Change `DBError(#[from] SqlxError)` to `DBError(anyhow::Error)` (note lack of `#[from]` -- we don't want an implicit conversion here).
3. Change all the functions in this module to return `GenesisError` and wrap DB errors to `GenesisError::DBError`.
4. For places where we still return `SqlxError` add `context`.

core/bin/zksync_server/src/main.rs
```diff
@@ -38,8 +38,7 @@ struct Cli {
     /// Generate genesis block for the first contract deployment using temporary DB.
     #[arg(long)]
     genesis: bool,
-    /// Wait for the `setChainId` event during genesis.
-    /// If `--genesis` is not set, this flag is ignored.
+    /// Set chain id (temporary will be moved to genesis config)
```
perekopskiy, line 41: Let's leave "Wait for the `setChainId` event during genesis." comment. Also, why have you changed the behavior, i.e. "If `--genesis` is not set, this flag is ignored." is not true anymore, is there any reason?

core/lib/types/src/api/mod.rs
```diff
@@ -714,6 +714,8 @@ pub struct BlockDetails {
 #[serde(rename_all = "camelCase")]
 pub struct L1BatchDetails {
     pub number: L1BatchNumber,
+    pub commitment: Option<H256>,
+    pub rollup_last_leaf_index: Option<u64>,
```
perekopskiy, line 718: IMO, we shouldn't add these fields to this endpoint only because of genesis changes. Maybe it's possible to get this data from L1 contract for genesis batch? If not maybe it's worth adding a separate endpoint

core/lib/zksync_core/src/genesis.rs
```diff
@@ -32,67 +36,141 @@ use zksync_utils::{be_words_to_bytes, bytecode::hash_bytecode, h256_to_u256, u25
...
+        let base_system_contracts_hashes = BaseSystemContractsHashes {
+            bootloader: config.bootloader_hash,
+            default_aa: config.default_aa_hash,
+        };
+        if base_system_contracts_hashes != base_system_contracts.hashes() {
+            return Err(GenesisError::BaseSystemContractsHashes(
+                base_system_contracts.hashes(),
+            ));
+        }
+        let _: ProtocolVersionId = config
```
perekopskiy, line 89: nit: please add comment

core/lib/zksync_core/src/genesis.rs
```diff
@@ -120,29 +202,56 @@ pub async fn ensure_genesis_state(
...
+    }
+
+    if genesis_params.config.rollup_last_leaf_index != rollup_last_leaf_index {
+        return Err(GenesisError::LeafIndexes(
+            genesis_params.config.rollup_last_leaf_index,
+            rollup_last_leaf_index,
+        ));
+    }
+
+    tracing::info!("operations_schema_genesis is complete");
```
perekopskiy, line 253: What's `operations_schema_genesis`?

etc/env/base/contracts.toml
```diff
@@ -2,52 +2,53 @@
...
+L2_WETH_TOKEN_PROXY_ADDR = "0x5E6D086F5eC079ADFF4FB3774CDf3e8D6a34F7E9"
+FRI_RECURSION_LEAF_LEVEL_VK_HASH = "0x400a4b532c6f072c00d1806ef299300d4c104f4ac55bd8698ade78894fcadc0a"
+FRI_RECURSION_NODE_LEVEL_VK_HASH = "0x5a3ef282b21e12fe1f4438e5bb158fc5060b160559c5158c6389d62d9fe3d080"
+FRI_RECURSION_SCHEDULER_LEVEL_VK_HASH = "0x1405880dc3317d635bddb0ab62bf5d013e5d1f462161c1f7ac3289c7fef956da"
+SNARK_WRAPPER_VK_HASH = "0x063c6fb5c70404c2867f413a8e35563ad3d040b1ad8c11786231bfdba7b472c7"
+BLOB_VERSIONED_HASH_RETRIEVER_ADDR = "0x0000000000000000000000000000000000000000"
 
 # Prover that should be used at genesis. 'fri' or 'snark'
-PROVER_AT_GENESIS="fri"
+PROVER_AT_GENESIS = "fri"
```
perekopskiy, line 49: Should be removed as well

core/lib/zksync_core/src/genesis.rs
```diff
@@ -120,29 +202,56 @@ pub async fn ensure_genesis_state(
...
+            .get_l1_batch_state_root(L1BatchNumber(0))
+            .await
+            .context("failed fetching state root hash for genesis L1 batch")?
+            .context("genesis L1 batch hash is empty")?);
+    }
+
+    tracing::info!("running regenesis");
+    let (genesis_root_hash, commitment, rollup_last_leaf_index) =
+        insert_genesis_batch(&mut transaction, genesis_params).await?;
+    if genesis_params.config.genesis_root_hash != genesis_root_hash {
```
perekopskiy, line 232: Why do we even need `genesis_params.config.genesis_root_hash`. Is it used anywhere else except this place? If not, do we need this check?

Deniallugo: Yes, it's used during contract deployment in this way we can deploy contracts without building the server

core/bin/zksync_server/src/main.rs
Deniallugo: For now, it just doesn't work properly. 
I'll implement it properly once we have shared bridges

core/lib/zksync_core/src/genesis.rs
perekopskiy: So, when genesis root hash/commitment changes (after another protocol upgrade) expected flow will be:
run zk init -> it fails -> change variable -> run zk init again? If so then it doesn't look good. I don't mind merging it because the current approach with printing is also bad but would be nice if we could improve it

core/lib/config/src/configs/chain.rs
```diff
@@ -116,6 +116,10 @@ pub struct StateKeeperConfig {
 
     /// Number of keys that is processed by enum_index migration in State Keeper each L1 batch.
     pub enum_index_migration_chunk_size: Option<usize>,
+
+    // Base system contract hash, required only for genesis file, it's temporary solution
+    pub bootloader_hash: Option<H256>,
+    pub default_aa_hash: Option<H256>,
```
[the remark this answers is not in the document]
Deniallugo: I kept the old names, for breaking as little code as possible.

core/lib/types/src/api/mod.rs
Deniallugo: Tbh, i was thinking that these information is useful and we can keep it in this method

core/lib/zksync_core/src/genesis.rs
Deniallugo: Yep, we can create a method for just updating Genesis config.

core/lib/zksync_core/src/genesis.rs
Deniallugo: I'd still keep `#[from]` otherwise it creates too much unnecessary boilerplate

Deniallugo:
And now I've realized i remove verifier field from api, I'll keep it as is, and remove only from configs

core/lib/zksync_core/src/genesis.rs
popzxc: Then any other error (including non-DB ones) would turn into `DBError`, which breaks the purpose of strong error typing (which I thought you're proponent of).

Deniallugo: and it's mostly the problem with using anyhow in libraries, that we can't distinguish errors. 

I could try to make sure that we use only methods with sqlx error from dal crate

perekopskiy (approved):

PR #1425, opened 2024-03-13
fix(api): Fix API server shutdown flow

slowli:
## What ❔

- API server now doesn't (gracefully) shut down immediately after receiving a signal. Instead, it waits for some time (until the inbound traffic reaches zero), and then starts the graceful shutdown.
- Some tasks run by the main node are now shutdown-aware.

## Why ❔

If the server starts shutting down immediately after receiving a signal (as it does now), this means that it will drop any new traffic, which is not what the K8s load balancer expects.

## Checklist


core/node/node_framework/src/implementations/layers/house_keeper.rs
```diff
@@ -196,8 +196,8 @@ impl Task for L1BatchMetricsReporterTask {
         "l1_batch_metrics_reporter"
     }
 
-    async fn run(self: Box<Self>, _stop_receiver: StopReceiver) -> anyhow::Result<()> {
-        self.l1_batch_metrics_reporter.run().await
+    async fn run(self: Box<Self>, stop_receiver: StopReceiver) -> anyhow::Result<()> {
+        self.l1_batch_metrics_reporter.run(stop_receiver.0).await
```
popzxc, line 200: Huh, looks like you've resolved [QIT-29](https://linear.app/matterlabs/issue/QIT-29). Could you please remove the TODO above? BTW, thanks for that :)

popzxc (approved):

Deniallugo (dismissed):

popzxc (approved):

popzxc (approved):

PR #1430, opened 2024-03-14
feat(storage): Make the storage caches task cancellation aware

popzxc:
## What ❔

Makes the support task for postgres storage caches aware of cancellation token.
Additionally, replaces the closure with a named type.

## Why ❔

Currently, it's being canceled implicitly (by dropping the command sender), which is not idiomatic in terms of the new framework.

## Checklist


popzxc:
@EmilLuta Nothing prevents us from extracting Postgres caches and the maintenance task into its own layer, though right now the benefits are not clear, given that the API is the sole consumer of these caches, and does it through `TxSender` -- which is already a resource with its own layer. So here we consider Postgres caches to be an implementation detail of `TxSender` to not pollute the public API too much.

PR #1445, opened 2024-03-16
fix: handle potential errors gracefully instead of unwrapping

amirf4:
## What ❔
This PR addresses the issue of using unwrap extensively, which could potentially cause panics in case of errors. The changes involve handling errors more gracefully by either returning a Result type or using match or if let statements to handle error cases explicitly.

## Why ❔
Using unwrap can lead to unexpected panics and application crashes if errors are not handled properly. By handling errors gracefully, we can ensure better application stability, provide meaningful error messages, and allow for proper error handling and recovery mechanisms.

## Checklist

- [x] Code has been formatted via zk fmt and zk lint.

- [x] Spellcheck has been run via zk spellcheck.

- [x] Linkcheck has been run via zk linkcheck.

### DIFF
--- a/core/lib/zksync_core/src/api_server/contract_verification/api_impl.rs
+++ b/core/lib/zksync_core/src/api_server/contract_verification/api_impl.rs
@@ -33,11 +33,14 @@ impl RestApi {
         if let Err(res) = Self::validate_contract_verification_query(&request) {
             return Ok(res);
         }
-        let mut storage = self_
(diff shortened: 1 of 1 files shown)


montekki:
Hi thanks for the proposition! Unfortunately this cannot be accepted since panicking in case of errors with the storage is one a necessity of operation of our system: if the storage is failing instead of trying to limp on the system should go down immediately.

PR #1452, opened 2024-03-18
feat(node-framework): Add circuit breaker checker layer to framework

AnastasiiaVashchuk:
Key thing to note - no more circuit breaker channel for signalling about check failure. Instead, make the circuit breaker a part of the other tasks and return an error instead of a signal throw channel. Motivation - framework design requires it.

## What ❔
Port circuit breaker checker to the framework

## Why ❔
these changes is a part of the system porting process to the framework

## Checklist


core/lib/circuit_breaker/src/lib.rs
```diff
@@ -21,50 +19,64 @@ pub enum CircuitBreakerError {
...
-        self,
-        circuit_breaker_sender: oneshot::Sender<CircuitBreakerError>,
-        stop_receiver: watch::Receiver<bool>,
-    ) -> anyhow::Result<()> {
+    pub async fn insert_breaker_if_not_exists(&self, circuit_breaker: Box<dyn CircuitBreaker>) {
+        let mut guard = self.circuit_breakers.lock().await;
+        if !guard.contains(&circuit_breaker) {
+            guard.push(circuit_breaker);
+        }
+    }
```
AnastasiiaVashchuk, line 66: So
Let's look into old logic:
https://github.com/matter-labs/zksync-era/blob/c072288a19511cb353e2895818178337b9a2d8ae/core/lib/zksync_core/src/lib.rs#L1337
as we can see we check if there is any from 3 API components and if yes - add `ReplicationLagChecker`.
In the framework logic there could be several `Web3ServerLayer`s...So to not add CB several times when we add few `Web3ServerLayer`s  we check whether there is already some concrete checker(in our case `ReplicationLagChecker`)

core/bin/zksync_server/src/main.rs
```diff
@@ -197,13 +197,6 @@ async fn main() -> anyhow::Result<()> {
         _ = sigint_receiver => {
             tracing::info!("Stop signal received, shutting down");
         },
-        error = cb_receiver => {
-            if let Ok(error_msg) = error {
-                let err = format!("Circuit breaker received, shutting down. Reason: {}", error_msg);
-                tracing::warn!("{err}");
-                vlog::capture_message(&err, vlog::AlertLevel::Warning);
-            }
-        },
```
AnastasiiaVashchuk, line 206: to add circuit breaker for the framework we have to change old logic.
lets avoid using channel and instead will send the error from the circuit breaker `check` method, in case the check fails. in fact the reaction of the system on the failed check won't change  since `wait_for_tasks()` runs circuit breaker task anyway.

core/lib/zksync_core/src/lib.rs
```diff
@@ -396,17 +396,19 @@ pub async fn initialize_components(
...
         let res = prometheus_task.await;
         drop(prometheus_health_updater);
         res
     });
 
+    let cb_stop_receiver = stop_receiver.clone();
     let mut task_futures: Vec<JoinHandle<anyhow::Result<()>>> = vec![
         prometheus_task,
-        tokio::spawn(circuit_breaker_checker.run(cb_sender, stop_receiver.clone())),
+        tokio::spawn(async move { circuit_breaker_checker.run(cb_stop_receiver).await }),
```
popzxc, line 411: `async move { x.await }` can be replaced by simply `x` (and then you don't need to create `cb_stop_receiver`, the `stop_receiver.clone()` will work (since there will be no `move`).

core/lib/zksync_core/src/lib.rs
```diff
@@ -396,17 +396,19 @@ pub async fn initialize_components(
     let (prometheus_health_check, prometheus_health_updater) =
         ReactiveHealthCheck::new("prometheus_exporter");
     app_health.insert_component(prometheus_health_check);
-    let prometheus_task = prom_config.run(stop_receiver.clone());
+    let pt_stop_receiver = stop_receiver.clone();
+    let prometheus_task = prom_config.run(pt_stop_receiver);
```
popzxc, line 400: Same here.

core/node/node_framework/src/implementations/layers/eth_sender.rs
```diff
@@ -103,6 +110,14 @@ impl WiringLayer for EthSenderLayer {
             eth_tx_manager_actor,
         }));
 
+        // Insert circuit breaker.
+        let CircuitBreakerCheckerResource(circuit_breaker_checker) = context.get_resource().await?;
```
popzxc, line 114: This won't work if we didn't add a circuit breaker layer. Circuit breaker component is optional -- it's a defensive measure, so we can perfectly run without it, if we ever want to.

For this to work, we need to split the circuit breaker into two parts:

1. Create structure `CircuitBreakers` which will hold the list of circuit breakers we have. It will have methods related to adding circuit breakers (like `insert_breaker_if_not_exists`).
2. Implement `Default` for `CircuitBreakers`. Change `CircuitBreakerCheckerResource` to `CircuitBreakersResource` and implement `Default` for it too. 
3. Here and in other places use `get_resource_or_default()` to get the resource. This way the resource will always be available, even if there is no circuit breaker layer.
4. Change the `CircuitBreakerChecker` so that it accepts `Arc<CircuitBreakers>` and access circuit breakers through this structure.
5. In the `CircuitBreakerCheckerLayer` you'll get `CircuitBreakersResource` and add a task for circuit breaker checker.

This way the code would work as follows:
- Any task can add a circuit breaker, it doesn't affect anything.
- *If* (and only if) we have a circuit breaker checker layer, we will fetch all the circuit breakers and instantiate a checker task. Otherwise adding these checkers would so nothing, but also (and it's important), would break nothing.

For inspiration, you may look at health check server, which is very similar. We have [AppHealthCheck](https://github.com/matter-labs/zksync-era/blob/main/core/lib/health_check/src/lib.rs#L123) there -- a container to which we can add health checks (for us `CircuitBreakers` would be such a container), and then _separately_ we have a way [to run the healthcheck server](https://github.com/matter-labs/zksync-era/blob/main/core/lib/zksync_core/src/api_server/healthcheck.rs#L52) that accepts `Arc<AppHealthCheck>` (for us it'll be `CircuitBreakerChecker::run(self, breakers: Arc<CircuitBreakers>, stop_receiver: watch::Receiver<bool>)`).

core/lib/circuit_breaker/src/lib.rs
```diff
@@ -21,50 +19,64 @@ pub enum CircuitBreakerError {
...
+
+    pub async fn run(self: Arc<Self>, stop_receiver: watch::Receiver<bool>) -> anyhow::Result<()> {
```
popzxc, line 68: Once you split the checker into two parts (see my other comment), you'll be able to get rid of `Arc<Self>` here and make it accept just `self` again.

core/lib/circuit_breaker/src/lib.rs
```diff
@@ -21,50 +19,64 @@ pub enum CircuitBreakerError {
...
         }
         Ok(())
     }
 
-    pub async fn run(
-        self,
-        circuit_breaker_sender: oneshot::Sender<CircuitBreakerError>,
-        stop_receiver: watch::Receiver<bool>,
-    ) -> anyhow::Result<()> {
+    pub async fn insert_breaker_if_not_exists(&self, circuit_breaker: Box<dyn CircuitBreaker>) {
```
popzxc, line 61: Nit: IMHO, `insert` is pretty self-explanatory (since this structure [has set semantics](https://doc.rust-lang.org/std/collections/struct.HashSet.html) anyways).

popzxc (commented):
Looks pretty good!

core/lib/circuit_breaker/src/lib.rs
AnastasiiaVashchuk: not sure I understand

AnastasiiaVashchuk: its Vec and not HashSet

popzxc: Vec is what you use internally, and I'm talking about _your_ container for circuit breakers, e.g. its API. It's OK for this method to be called `insert`, because it's not unexpected for your container to behave _like_ hashset.

core/node/node_framework/src/implementations/layers/eth_sender.rs
AnastasiiaVashchuk: Thanks! It fully makes sense.
changed

core/lib/circuit_breaker/src/lib.rs
```diff
@@ -1,15 +1,27 @@
...
-use tokio::sync::watch;
-use zksync_config::configs::chain::CircuitBreakerConfig;
+use tokio::sync::{watch, Mutex};
 
 pub mod l1_txs;
 pub mod replication_lag;
 pub mod utils;
 
+#[derive(Default, Debug)]
+pub struct CircuitBreakers(pub Mutex<Vec<Box<dyn CircuitBreaker>>>);
```
popzxc, line 11: I think the inner mutex shouldn't be `pub` (otherwise it's easy to break the internal invariant of unique elements).

core/lib/circuit_breaker/src/lib.rs
```diff
@@ -21,48 +33,42 @@ pub enum CircuitBreakerError {
...
         }
     }
 
     pub async fn check(&self) -> Result<(), CircuitBreakerError> {
-        for circuit_breaker in &self.circuit_breakers {
+        for circuit_breaker in self.circuit_breakers.0.lock().await.iter() {
             circuit_breaker.check().await?;
         }
         Ok(())
     }
```
popzxc, line 60: You can move this method to `CircuitBreakers` -- it probably logically belongs there (and it'll be consistent with the healthcheck module design).

core/node/node_framework/src/implementations/resources/circuit_breakers.rs
```diff
@@ -0,0 +1,23 @@
+use std::sync::Arc;
+
+use zksync_circuit_breaker::CircuitBreakers;
+
+use crate::resource::Resource;
+
+#[derive(Debug, Clone, Default)]
+pub struct CircuitBreakersResource {
+    pub replication_lag_limit_sec: Option<u32>,
```
popzxc, line 9: I think this may be designed better. Currently, you implicitly rely on the fact that some layer would fetch and create this resource, set this value, and then you will use it to instantiate `ReplicationLagChecker`. This isn't the best design for a couple of reasons:
1. You create more implicit relationships between components, which may be confusing during debugging.
2. You have several components aware of this configuration, which breaks the point of framework where we want to encapsulate the initialization as much as possible.

To improve it, you may add this field to `Web3ServerOptionalConfig`, and take it from there. This way the `Web3ServerLayer` will be the only layer aware of `ReplicationLagChecker`.

popzxc (approved):
Nice!

popzxc:
@AnastasiiaVashchuk seems like there are a few merge conflicts now, can you resolve?

AnastasiiaVashchuk:
> @AnastasiiaVashchuk seems like there are a few merge conflicts now, can you resolve?

moment

popzxc (approved):

PR #1457, opened 2024-03-19
feat(external-node): external node distributed operation mode

montekki:
## What ❔

Adds a way to run External Node in a "distributed setup": that if there is a need to launch different components of the EN on different machines for performance reasons. The currently supported component split is
* `core`
* `tree`
* `tree_api` (depends on `tree`)
* `api` (may be split into two components: `http_api` and `ws_api`)

Example setup:

```
zk external-node -- --components=core
```

```
 env EN_HEALTHCHECK_PORT=3082 EN_PROMETHEUS_PORT=3323 zk external-node -- --components="tree,tree_api"
```
## Checklist


core/bin/external_node/src/main.rs
```diff
@@ -458,6 +593,60 @@ struct Cli {
...
+    components: ComponentsToRun,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum Component {
+    HttpApi,
+    WsApi,
+    Tree,
+    TreeApi,
+    Core,
```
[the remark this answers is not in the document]
montekki: Yeah currently main node and EN init APIs in two separate places and I am not sure that they can be united. Ideally EN should reuse code for running API and other components from `./core/lib/zksync_core/src/lib.rs`, i will try to make it work but i am not sure that currently it is possible with main node configs vs. EN configs.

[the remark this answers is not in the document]
popzxc: Running EN/main node in a unified way will be handled by Legolizer, so I think it doesn't make much sense to create some kind of unification logic for the old system. An interim solution would do, IMO.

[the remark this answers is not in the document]
RomanBrodetski: got it - I definitely don't insist making it perfect, especially if it's temporary. 

But it's **really** important for me that we don't keep diverging EN and Main node further. I think this will make the future decentralization harder. Note how the EN and Main Node will converge once we properly integrate the consesnsus.

Seeing the @popzxc 's comment above I'm more optimistic. Will have a closer look at the PR later on.

core/bin/external_node/src/config/mod.rs
```diff
@@ -531,6 +553,14 @@ impl ExternalNodeConfig {
             .from_env::<OptionalENConfig>()
             .context("could not load external node config")?;
 
+        let api_component_config = envy::prefixed("EN_")
```
tomg10, line 556: wdyt about using different prefixes like EN_API and EN_TREE? It would be easier to tell where is certain env variable needed

core/bin/external_node/src/main.rs
```diff
@@ -359,44 +387,189 @@ async fn init_tasks(
...
+        .context("failed to build a tree_pool")?;
+
+    let fee_params_fetcher = Arc::new(MainNodeFeeParamsFetcher::new(main_node_client.clone()));
+    if !components.contains(&Component::Tree) {
+        anyhow::ensure!(
+            !components.contains(&Component::TreeApi),
+            "Merkle tree API cannot be started without a tree component"
+        );
+    }
+    let tree_reader: Option<Arc<dyn TreeApiClient>> = if components.contains(&Component::Tree) {
```
tomg10, line 503: I know you've described this nicely in config, but I think it could be really valuable to comment here as well that we either run tree API in that specific node or act as a proxy

core/bin/external_node/src/main.rs
```diff
@@ -359,44 +387,189 @@ async fn init_tasks(
...
+
+    let singleton_pool_builder = ConnectionPool::singleton(&config.postgres.database_url);
+
+    // Run the components.
+    let tree_pool = singleton_pool_builder
+        .build()
+        .await
+        .context("failed to build a tree_pool")?;
+
+    let fee_params_fetcher = Arc::new(MainNodeFeeParamsFetcher::new(main_node_client.clone()));
```
tomg10, line 496: nit: this line looks weirdly out of place here, in the middle of tree initialization doe

core/bin/external_node/src/config/mod.rs
```diff
@@ -244,6 +244,26 @@ pub struct OptionalENConfig {
...
+    /// local tree component running and in this case needs to send requests
+    /// to some external tree API.
+    pub tree_api_url: Option<String>,
+    pub tree_api_port: Option<u16>,
+    /// Proxy requests to `eth_syncing` to this node.
+    /// Used in case of running an API component of the EN together with a separate
+    /// instance of EN running `core` component. In that case it is the `core` component
+    /// that may do the actual syncing, not the API and so API has to forward these
+    /// requests to `core`.
+    pub ask_syncing_status_from: Option<String>,
```
tomg10, line 259: Won't this force such core instances to also have to run their own api ? If yes, were you considering using db for fetching current sync state or is something preventing us from using it that way?

tomg10 (commented):
Overall looks good, I had a comment about ask_syncing that I want addressed, other than that, I had a few small comments that are just small suggestions

core/bin/external_node/src/config/mod.rs
montekki: good idea

montekki: Yeah idk, currently this is an in-memory piece of data. So if we put it into DB then the core node may be down but the DB will still have "syncing" in it and API will return it

core/bin/external_node/src/main.rs
```diff
@@ -359,44 +387,200 @@ async fn init_tasks(
...
+                .with_tree_api(tree_reader)
+                .enable_api_namespaces(config.optional.api_namespaces());
+
+        if let Some(sync_state) = sync_state {
+            builder = builder.with_sync_state(sync_state);
+        }
+
+        if let Some(ask_sync_state_from) = ask_sync_state_from {
+            builder = builder.with_ask_sync_state(ask_sync_state_from);
+        }
```
perekopskiy, line 435: Why is it only checked for WS?

core/bin/external_node/src/main.rs
```diff
@@ -460,6 +630,59 @@ struct Cli {
...
+                Component::Tree,
+                Component::Core,
+            ])),
+            other => Err(format!("{} is not a valid component name", other)),
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+struct ComponentsToRun(Vec<Component>);
```
slowli, line 673: Bikeshedding: IMO, two-layer parsing looks confusing. IMO, it'd be better to have a single `Components` wrapper and refactor the inner parsing layer (i.e., one that happens after the input is split by `,`) as an function or an inherent static method in `Components`.

core/bin/external_node/src/config/mod.rs
RomanBrodetski: like @slowli 's approach the only thing is that this will result in more requests to the Main Node, and what's worse it'll scale linearly with the number of API servers.

On the other hand, the requests themselves are extremely cheap so maybe it's OK

montekki: I don't have a strong opinion however with the current way:
1. one source of truth
2. a simple proxying of a request to it
3. APIs will need outbound connectivity, not only within the "distributed setup"

replicate `SyncState`:
1. many unsynced "sources of truth"
2. still doing some requests but somewhere outside of EN "distributed setup"

core/bin/external_node/src/main.rs
montekki: yep, thanks, copypasta

core/bin/external_node/src/main.rs
```diff
@@ -359,44 +387,200 @@ async fn init_tasks(
...
+            let protocol_version = pool
+                .connection()
+                .await
+                .unwrap()
+                .protocol_versions_dal()
+                .last_used_version_id()
+                .await
+                .map(|version| version as u16);
+
+            EN_METRICS.version[&(format!("{}", version), protocol_version)].set(1);
```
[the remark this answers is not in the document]
montekki: yeah, this code is just copypaste from prev version we should probably do a re-export in vise that you are mentioning and then do a follow-up here

core/bin/external_node/src/config/mod.rs
montekki: Yep idk to me doing requests from API nodes to core nodes for this within an isolated distributed setup sounds more consistent then spamming O(n) requests to main node instead.

slowli: The main node can be queried using a background task with a reasonably large interval between queries, so that rate-limiting is not a concern. Regarding rate-limiting, [I'm working](https://github.com/matter-labs/zksync-era/pull/1500) on applying it globally (within a process, of course), so that it would be even less of a concern in the near future. What I don't like about exposing the API server for the core is that it's not the user might expect, and that the entire API server is exposed (which might even have security repercussions) while we only use a single method.

core/bin/external_node/src/main.rs
```diff
@@ -359,44 +386,204 @@ async fn init_tasks(
...
+    // we can get this tree's reader and use it right away. Otherwise, if configuration has
+    // specified address of another instance hosting tree API, create a tree reader to that
+    // remote API. A tree reader is necessary for `zks_getProof` method to work.
+    let tree_reader: Option<Arc<dyn TreeApiClient>> = if components.contains(&Component::Tree) {
+        let tree_api_config = if components.contains(&Component::TreeApi) {
+            Some(MerkleTreeApiConfig {
+                port: config
+                    .tree_component_config
+                    .api_port
+                    .ok_or(anyhow!("should contain tree api port"))?,
```
slowli, line 517: Nit: You can use `.context(_)` on an option with the same effect; it wouldn't require eagerly constructing an error (which could be avoided using `ok_or_else`, but `context` is shorter and arguably more idiomatic).

core/bin/external_node/src/main.rs
```diff
@@ -359,44 +386,204 @@ async fn init_tasks(
...
+                tree_api_config.as_ref(),
+                app_health,
+                stop_receiver.clone(),
+                tree_pool,
+            )
+            .await?,
+        )
+    } else if let Some(tree_api_url) = &config.api_component_config.tree_api_url {
+        Some(Arc::new(TreeApiHttpClient::new(tree_api_url)))
+    } else {
```
slowli, line 535: Bikshedding: I'd move this initialization to `run_api`. It would also make more transparent that specifying both `tree_api_url` and running a tree component is a misconfiguration (IMO, it'd make sense to log a warning in this case).

core/bin/external_node/src/config/mod.rs
montekki: @slowli done

core/bin/external_node/src/main.rs
```diff
@@ -55,6 +60,42 @@ mod version_sync_task;
 
 const RELEASE_MANIFEST: &str = include_str!("../../../../.github/release-please/manifest.json");
 
+async fn sync_state_updater(
+    sync_state: SyncState,
+    tree_pool: ConnectionPool<Core>,
+    main_node_client: HttpClient,
+    stop_receiver: watch::Receiver<bool>,
+) -> anyhow::Result<()> {
+    const UPDATE_INTERVAL: Duration = Duration::from_secs(60);
```
RomanBrodetski, line 69: isn't 60 seconds too much? we consider node out of sync after I think being t10 blocks behind, which takes 10 seconds on average. So I'd set 10 seconds interval

RomanBrodetski (commented):
Overall looks good! 

I assume we'll get rid of a new `struct` for `ComponentsToRun` once we migrate to legolizer

core/bin/external_node/src/main.rs
montekki:
```suggestion
    const UPDATE_INTERVAL: Duration = Duration::from_secs(10);
```

core/lib/zksync_core/src/sync_layer/sync_state.rs
```diff
@@ -73,6 +77,44 @@ impl SyncState {
...
+        connection_pool: ConnectionPool<Core>,
+        main_node_client: HttpClient,
+        mut stop_receiver: watch::Receiver<bool>,
+    ) -> anyhow::Result<()> {
+        const UPDATE_INTERVAL: Duration = Duration::from_secs(10);
+
+        loop {
+            if *stop_receiver.borrow() {
+                return Ok(());
+            }
```
slowli, line 92: Nit: You can simplify this and the other similar `loop` to

```rust
while !stop_receiver.borrow_and_update() {
    // ...
}
```

`borrow_and_update` is marginally preferable in this scenario because `borrow` doesn't update the latest received value of the `Receiver`; i.e., if hypothetically a no-op update was sent, `stop_receiver.changed()` below would spuriously resolve immediately. OTOH, a no-op update is a not a very realistic scenario.

core/bin/external_node/src/main.rs
```diff
@@ -461,6 +627,56 @@ struct Cli {
...
+            "tree" => Ok(&[Component::Tree]),
+            "tree_api" => Ok(&[Component::TreeApi]),
+            "core" => Ok(&[Component::Core]),
+            "all" => Ok(&[
+                Component::HttpApi,
+                Component::WsApi,
+                Component::Tree,
+                Component::Core,
+            ]),
+            other => Err(anyhow::anyhow!("{} is not a valid component name", other)),
```
slowli, line 659: Nit: You can specify `other` in an interpolated string: `{other} is not a valid component name`.

slowli (approved):

slowli (commented):
Other than an incorrect loop condition, looks good 👍 (The condition doesn't matter rn, but would make sense fix for the general correctness, IMO.)

core/lib/zksync_core/src/sync_layer/sync_state.rs
```diff
@@ -73,6 +77,41 @@ impl SyncState {
...
+
+    pub async fn run_updater(
+        self,
+        connection_pool: ConnectionPool<Core>,
+        main_node_client: HttpClient,
+        mut stop_receiver: watch::Receiver<bool>,
+    ) -> anyhow::Result<()> {
+        const UPDATE_INTERVAL: Duration = Duration::from_secs(10);
+
+        while *stop_receiver.borrow_and_update() {
```
montekki, line 89:
```suggestion
        while !*stop_receiver.borrow_and_update() {
```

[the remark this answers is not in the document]
montekki: done

slowli (approved):

PR #1460, opened 2024-03-19
feat(api): introduce mempool cache

RomanBrodetski:
## What ❔

This introduces an in-memory cache on ws API servers that caters for `eth_newPendingTransactionFilter` filter. Currently for each `eth_getFilterChanges` request to such filter, we do a DB query. This query dominates the DB load of ws DBs. 

The cache is updated preiodically, loading the new mempool transactions to a `Deque`. To return all the transactions that happened after a particular timestamp, we binsearch this data strucute - avoiding additional DB calls.

## Why ❔

to relieve some load from the read replica DB.


core/lib/state/src/cache/lru_cache.rs
```diff
@@ -0,0 +1,119 @@
+use std::hash::Hash;
```
RomanBrodetski, line 1: Note: this logic is moved from `cache/mod.rs` as-is

core/lib/state/src/cache/sequential_cache.rs
```diff
@@ -0,0 +1,150 @@
...
+
+/// A generic cache structure for storing key-value pairs in sequential order.
+/// It allows for non-unique keys and supports efficient retrieval of values based on a key
+/// threshold. The cache maintains a specified maximum capacity, removing the oldest entries
+/// as new ones are added.
+///
+/// Usage example: storing mempool transactions and querying them by the received_at field
+/// (clients maintain a cursor based on this key and poll the data structure for newer transactions)
+///
+///
```
popzxc, line 13: Nit: Unnecessary empty line.

core/lib/state/src/cache/sequential_cache.rs
```diff
@@ -0,0 +1,150 @@
...
+/// K: Type of the ordered, potentially non-unique key. Example: Transaction's received_at field.
+/// V: Type of the value associated with each key. Example: Transaction's hash.
+#[derive(Debug, Clone)]
+pub struct SequentialCache<K, V> {
+    name: &'static str,
+    data: VecDeque<(K, V)>,
+    capacity: usize,
+}
+
+impl<K: Ord + Copy, V: Clone> SequentialCache<K, V> {
```
popzxc, line 23: Nit: `Copy` requirement seems overly strict here, is it intentional or `Clone` would suffice? (Though I think even `Clone` can be avoided on both `K` and `V` if we change the signatures to work with references, but it'll alter the UX so dunno if that aligns with your point of view).

core/lib/state/src/cache/sequential_cache.rs
```diff
@@ -0,0 +1,150 @@
...
+    /// Panics when keys order is incorrect (a smaller key is inserted after a larger one)
+    pub(crate) fn insert<I>(&mut self, items: I)
+    where
+        I: IntoIterator<Item = (K, V)>,
+    {
+        for (key, value) in items {
+            let latency = METRICS.latency[&(self.name, Method::Insert)].start();
+            if let Some(last_key) = self.get_last_key() {
+                assert!(key >= last_key, "Keys must be inserted in sequential order");
+            }
```
popzxc, line 45: Nit: Given that we're (generally) try to avoid panicking right now, and given that it's not an internal invariant, probably it makes sense for this function to return `Result`. It's external code to guarantee the order of keys, so `cache.insert(items).expect("DAL guarantees keys to be ordered");` in theory makes more sense

popzxc: Also `Option` implements `Ord` for `T: Ord`, so it can be a one-liner, e.g. `anyhow::ensure!(Some(key) >= self.get_last_key(), "Keys must be inserted in sequential order");`

popzxc (approved):
Looks good. A few minor nits, feel free to ignore.

core/lib/state/src/cache/sequential_cache.rs
RomanBrodetski: > Option implements Ord for T: Ord

wow, really didn't expect it. Is surprising for me since `None < Some(1)` or `None > Some(1)` is essentially arbitrary (one can argue though that None can be thought of as `0`, hence it's less than anything. 

Found [this](https://www.reddit.com/r/rust/comments/1bdqocw/why_is_ord_implemented_on_option/) thread with many explanations, so I'm good :)

Will change to `result` as well.

core/lib/dal/sqlx-data.json
```diff
@@ -0,0 +1,3 @@
+{
```
popzxc, line 1: I guess you need to update sqlx-cli (prolly you still have 0.5 installed)

core/lib/state/src/cache/sequential_cache.rs
```diff
@@ -0,0 +1,153 @@
...
+        SequentialCache {
+            name,
+            data: VecDeque::with_capacity(capacity),
+            capacity,
+        }
+    }
+
+    /// Inserts multiple key-value pairs into the cache from an iterator. If adding these
+    /// items exceeds the cache's capacity, the oldest entries are removed. Keys can be non-unique.
+    /// Panics when keys order is incorrect (a smaller key is inserted after a larger one)
```
popzxc, line 37: Also the doc-comment is no longer valid.

core/lib/state/src/mempool_cache/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+use std::{future::Future, sync::Arc, time::Duration};
+
+use chrono::NaiveDateTime;
+use tokio::sync::{watch, RwLock};
+use zksync_dal::ConnectionPool;
+use zksync_types::H256;
+
+use crate::{cache::sequential_cache::SequentialCache, mempool_cache::metrics::METRICS};
+
+/// Used for `eth_getPendingTransactions` requests on API servers
```
perekopskiy, line 12: nit: there is no `eth_getPendingTransactions`. It's a `eth_newPendingTransactionFilter` + `eth_getFilterChanges`, but mentioning only latter is fine

core/lib/zksync_core/src/api_server/web3/namespaces/eth.rs
```diff
@@ -753,23 +753,41 @@ impl EthNamespace {
...
-                    .transactions_web3_dal()
-                    .get_pending_txs_hashes_after(
-                        *from_timestamp_excluded,
-                        Some(self.state.api_config.req_entities_limit),
-                    )
-                    .await
-                    .context("get_pending_txs_hashes_after")?;
+                    .mempool_cache
+                    .get_tx_hashes_after(*from_timestamp_excluded)
+                    .await;
```
perekopskiy, line 762: I think we should also truncate it to `req_entities_limit` length.

core/tests/ts-integration/tests/api/web3.test.ts
```diff
@@ -242,6 +242,42 @@ describe('web3 API compatibility tests', () => {
...
+    test('Should test getFilterChanges for pending transactions', async () => {
+        if (process.env.EN_MAIN_NODE_URL) {
+            // Pending transactions logic doesn't work on EN since we don't have proper mempool -
+            // transactions only appear in the DB after they are included in the block.
+            return;
+        }
+
+        // We will need to wait until the mempool cache on the server is updated.
+        // The default update period is 50 ms, so we will wait for 75 ms to be sure.
+        const mempool_cache_wait = 50 + 25;
```
perekopskiy, line 254: nit: camelCase

core/lib/dal/sqlx-data.json
popzxc: (Just in case: this file should be removed, it's a side effect of the old `sqlx`)

perekopskiy (approved):

PR #1466, opened 2024-03-20
feat(en): consistency checker persistent cursor

tomg10:
## What ❔

Makes the L1 batch number cursor used by the consistency checker (i.e., the last checked L1 batch) persistent.

## Why ❔

Consistency checker seems to be the only component that has a logical L1 batch number cursor influencing pruning without persistence. (E.g., metadata calc, commitment generator and batch status updater all have persistent cursors stored in Postgres.)


core/lib/dal/migrations/20240320173835_consistency-checker-cursor.up.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE consistency_checker_info
```
slowli, line 1: I think it could make sense to have a more general singleton table holding cursors for L1 batch (and miniblock?) processing; it could become useful if we split L1 batch processing as planned (e.g., separate initial writes / protective reads processing).

core/lib/dal/migrations/20240320173835_consistency-checker-cursor.up.sql
```diff
@@ -0,0 +1,9 @@
...
+(
+    last_processed_l1_batch BIGINT NOT NULL,
+    created_at              TIMESTAMP NOT NULL,
+    updated_at              TIMESTAMP NOT NULL
+);
+
+INSERT INTO consistency_checker_info(last_processed_l1_batch, created_at, updated_at)
+VALUES (1, NOW(), NOW());
```
slowli, line 9: Shouldn't the processed L1 batch number be zero rather than 1? At least it makes sense from the semantics perspective – one can say that the 0-th / genesis L1 batch is always checked, but this is not necessarily the case for the 1st batch. Also, personally, IMO making the `last_processed_l1_batch` column nullable would better reflect the domain (it'd allow not to insert an artificial value into the table).

core/lib/zksync_core/src/consistency_checker/mod.rs
```diff
@@ -504,11 +504,21 @@ impl ConsistencyChecker {
...
+            .get_consistency_checker_last_processed_l1_batch()
+            .await?;
+
         // We shouldn't check batches not present in the storage, and skip the genesis batch since
         // it's not committed on L1.
         let first_batch_to_check = first_batch_to_check
             .max(earliest_l1_batch_number)
-            .max(L1BatchNumber(1));
+            .max(L1BatchNumber(1))
+            .max(last_processed_batch);
```
slowli, line 521: It should be the next batch after the processed one, right? Would also allow skipping `max(L1BatchNumber(1))` altogether.

slowli (commented):
I considered, as an extension of this task, ability to skip to a later batch if the consistency checker lags too far behind (e.g., if it is out of a hard-coded or a configurable window like 1,000 latest batches). Could be done later, obviously.

core/lib/dal/migrations/20240320173835_consistency-checker-cursor.up.sql
tomg10: Nice catch! I was changing semantics of this table and missed this place

core/lib/zksync_core/src/consistency_checker/mod.rs
tomg10: Yes; corrected

core/lib/dal/migrations/20240320173835_consistency-checker-cursor.up.sql
tomg10: It could very well make sense, but I'd be for switching to such table only when we introduce new batch processing component as design of such table should be really well thought

core/lib/zksync_core/src/consistency_checker/mod.rs
```diff
@@ -504,11 +504,20 @@ impl ConsistencyChecker {
...
+            .blocks_dal()
+            .get_consistency_checker_last_processed_l1_batch()
+            .await?;
+
         // We shouldn't check batches not present in the storage, and skip the genesis batch since
         // it's not committed on L1.
         let first_batch_to_check = first_batch_to_check
             .max(earliest_l1_batch_number)
-            .max(L1BatchNumber(1));
+            .max(L1BatchNumber(last_processed_batch.0 + 1));
```
slowli, line 520: Nit: You can add `u32` to `L1BatchNumber` directly, so this can be simplified to `.max(last_processed_batch + 1)`.

slowli (approved):

core/lib/zksync_core/src/consistency_checker/mod.rs
tomg10: almost directly, I'd still have to use `*last_processed_batch + 1` so I think it's not worth the hassle

PR #1467, opened 2024-03-21
feat: Switch contract verification API to axum and get rid of actix-web usage

popzxc:
## What ❔

- Switches contract verification API to axum.
- Switches an eth-signer test to use axum.
- Adds support of contract verification API to the node framework.

## Why ❔

We use axum for web servers nowadays.

## Checklist


core/lib/eth_signer/src/json_rpc_signer.rs
```diff
@@ -380,43 +384,40 @@ mod tests {
...
+                server = Some(axum::serve(listener, app));
                 url = Some(new_url);
                 break;
             }
         }
 
         let server = server.expect("Could not bind to port from 9000 to 9999");
         let (abort_handle, abort_registration) = AbortHandle::new_pair();
-        let future = Abortable::new(server.run(), abort_registration);
+        let future = Abortable::new(server.into_future(), abort_registration);
```
popzxc, line 409: Probably not idiomatic, but it was quicker than trying to implement an idiomatic graceful shutdown, and probably it's not worth it given that it's a unit test for the code we don't even use (yet).

core/lib/zksync_core/src/api_server/contract_verification/api_impl.rs
```diff
@@ -1,24 +1,44 @@
...
+        .body(message.to_string())
+        .unwrap()
+}
+
+fn not_found() -> Response<String> {
+    Response::builder()
+        .status(axum::http::StatusCode::NOT_FOUND)
+        .body(String::new())
+        .unwrap()
 }
```
popzxc, line 33: Kinda clunky, but that's how API worked with actix, and the goal for this PR was to switch the used library, and not to fully refactor.

core/lib/utils/src/panic_notify.rs
popzxc, line 1: Yay! Less kludges!

montekki (approved):

PR #1471, opened 2024-03-21
feat(node_framework): Add a task to handle sigint

popzxc:
## What ❔

Adds a task that handles SIGINT (aka ctrl+c), propagating the signal to the `ZkStackService` instead of shutting the node down immediately.

## Why ❔

- Graceful shutdown.
- Replicating the current behavior.

## Checklist

### DIFF
--- a/core/node/node_framework/src/implementations/layers/sigint.rs
+++ b/core/node/node_framework/src/implementations/layers/sigint.rs
@@ -18,7 +18,7 @@ impl WiringLayer for SigintHandlerLayer {
     }
 
     async fn wire(self: Box<Self>, mut node: ServiceContext<'_>) -> Result<(), WiringError> {
-        // Sigint may happen at any time, so we must handle it as soon as it happens.
+        // SIGINT may happen at any time, so we must handle it as soon as it happens.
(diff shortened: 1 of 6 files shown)


core/node/node_framework/src/implementations/layers/sigint.rs
```diff
@@ -0,0 +1,60 @@
...
+struct SigintHandlerTask;
+
+#[async_trait::async_trait]
+impl UnconstrainedTask for SigintHandlerTask {
+    fn name(&self) -> &'static str {
+        "sigint_handler"
+    }
+
+    async fn run_unconstrained(
+        mut self: Box<Self>,
```
AnastasiiaVashchuk, line 37: do we need mut here?

popzxc: Good point, fixed. Thx.

RomanBrodetski (approved):

popzxc:
> I wonder if in the future we can add different timeouts as well

Basic support of timeouts is already present in the framework (e.g. after sending a stop signal, the service will wait 30 seconds (hardcoded timeout for now) for all the components to stop, then it'll exit forcefully).

PR #1478, opened 2024-03-21
feat(kl-factory): add base token tests

juan518munoz:
## What ❔

Add new tests for the base token implementation:
- Verifies deposits (TODO: check if they can be tightened).
- Verifies transfers.
- Verifies withdraws.

## Why ❔

Check base token behaves as expected.

## Checklist


core/tests/ts-integration/src/context-owner.ts
```diff
@@ -354,34 +367,33 @@ export class TestContextOwner {
...
-            );
-
-            // Add this promise to the list of L1 tx promises.
-            // l1TxPromises.push(depositHandle);
-            await depositHandle;
+            // const depositHandle = this.mainSyncWallet
+            //     .deposit({
+            //         token: zksync.utils.ETH_ADDRESS,
+            //         approveBaseERC20: true,
+            //         approveERC20: true,
```
ly0va, line 374: Should this `if` be removed or do we still need it?

core/tests/ts-integration/src/context-owner.ts
```diff
@@ -354,34 +367,33 @@ export class TestContextOwner {
...
+            //         amount: l2ETHAmountToDeposit as BigNumberish,
+            //         approveBaseOverrides: {
+            //             nonce: nonce,
+            //             gasPrice
+            //         },
```
ly0va, line 379: Does this deposit break the custom base token test flow? 
If it does, we probably need another flag to switch it off while testing networks with custom base token.
If not, we probably should uncomment this

juan518munoz: It breaks on the first `zk test i server` ran, but not in the next ones. If I'm not mistaken, this issue is present on `kl-factory` too, so we should try to see if it can be fixed on a separate PR, the better option might be to uncomment this for this PR so it doesn't make changes to `kl-factory`.

Raid5594:
Merged changes here:
https://github.com/matter-labs/zksync-era/pull/1492

PR #1491, opened 2024-03-25
refactor(api): Brush up after #1460

slowli:
## What ❔

Small changes after #1460:

- `MempoolCache` is moved to the core crate as per its domain.
- Some minor things like dropping DB connections earlier.

## Why ❔

Marginally simplifies maintenance.

## Checklist


RomanBrodetski:
oh thanks! Also thanks for reviewing the original PR (which I assume you did).

Re: move to core - makes sense, I created it in `state` because `postgres` caches were there.

RomanBrodetski (approved):

PR #1496, opened 2024-03-26
fix(prover): Remove redundant LoadingMode

EmilLuta:
This is PR 1 of many to move us to safe Prover Protocol Upgrades.

`ProtocolVersionLoadingMode` is a crutch that was added as part of transitioning to Boojum. This served as a manual intervention mechanism during protocol upgrades. It was a suboptimal choice (read as technical debt) done due to rushed migration to boojum.

A proper fix involves reading the information from database, which in turn is updated by information received from contracts. This change was done recently (1.4.2 migration), but the old cludge was left in place (you guessed it, due to rushed migration).

This PR removes this cludge and makes sure that nobody will try to use it assuming it will work (whilst it does work, it requires "specialized" knowledge on how to use it and may cause us submitting broken proofs to L1).

### DIFF
--- a/core/node/node_framework/src/implementations/layers/proof_data_handler.rs
+++ b/core/node/node_framework/src/implementations/layers/proof_data_handler.rs
@@ -1,6 +1,6 @@
 use std::sync::Arc;
 
-use zksync_config::{configs::ProofDataHandlerConfig, ContractsConfig};
+use zksync_config::configs::ProofDataHandlerConfig;
 use zksync_core::proof_data_handler;
 use zksync_dal::{ConnectionPool, Core};
 use zksync_object_store::ObjectStore;


core/lib/config/src/configs/proof_data_handler.rs
```diff
@@ -2,18 +2,10 @@ use std::time::Duration;
...
-#[derive(Debug, Deserialize, Clone, Copy, PartialEq)]
-pub enum ProtocolVersionLoadingMode {
-    FromDb,
-    FromEnvVar,
-}
-
 #[derive(Debug, Deserialize, Clone, PartialEq)]
 pub struct ProofDataHandlerConfig {
     pub http_port: u16,
     pub proof_generation_timeout_in_secs: u16,
```
mm-zk, line 8: BTW - what happens when someone still has PROOF_DATA_HANDLER_PROTOCOL_VERSION_LOADING_MODE set in their ENV variable ? it is silently ignored, or does the node crash ?

core/lib/protobuf_config/src/proto/proof_data_handler.proto
```diff
@@ -10,6 +10,4 @@ enum ProtocolVersionLoadingMode {
 message ProofDataHandler {
   optional uint32 http_port = 1; // required; u16
   optional uint32 proof_generation_timeout_in_secs = 2; // required; s
```
mm-zk, line 12: please leave the comment that 3 & 4 were used.

core/lib/zksync_core/src/proof_data_handler/request_processor.rs
```diff
@@ -105,44 +97,31 @@ impl RequestProcessor {
...
+            .unwrap()
+            .blocks_dal()
+            .get_l1_batch_header(l1_batch_number)
+            .await
+            .unwrap()
+            .expect(&format!("Missing header for {}", l1_batch_number));
 
-            }
-        };
+        let protocol_version = header.protocol_version.unwrap();
```
mm-zk, line 111: IMHO - we should merge fri protocol version and protocol version.

In theory, we might have protocol version changes that are not impacting the circuits, but that's rare enough and probably not worth the added complexity.

mm-zk (approved):

core/lib/protobuf_config/src/proto/proof_data_handler.proto
```diff
@@ -10,6 +10,4 @@ enum ProtocolVersionLoadingMode {
...
-  optional ProtocolVersionLoadingMode protocol_version_loading_mode = 3; // required
-  optional uint32 fri_protocol_version_id = 4; // required; u16
```
popzxc, line 14: @pompon0 is it safe to remove fields from proto repr? (I guess that it's probably OK here since it's not in use, but in general)

core/lib/config/src/configs/proof_data_handler.rs
RomanBrodetski: ignored..

core/lib/protobuf_config/src/proto/proof_data_handler.proto
pompon0: in general it is not. I haven't set up a CI compatibility check yet (https://linear.app/matterlabs/issue/BFT-446/enable-protobuf-compatibility-ci-for-zksync-era), so it was not detected automatically. You should add `reserved 3,4;` and `reserved "protocol_version_loading_mode","fri_protocol_version_id"` like here https://protobuf.dev/programming-guides/proto3/#fieldreserved. Until it is not in use though, it doesn't really matter.

pompon0 (dismissed):

core/lib/zksync_core/src/proof_data_handler/mod.rs
```diff
@@ -3,49 +3,24 @@ use std::{net::SocketAddr, sync::Arc};
...
-use zksync_types::{
-    protocol_version::{L1VerifierConfig, VerifierParams},
-    H256,
-};
 
 use crate::proof_data_handler::request_processor::RequestProcessor;
 
 mod request_processor;
 
-fn fri_l1_verifier_config(contracts_config: &ContractsConfig) -> L1VerifierConfig {
```
ilitteri, line 22: Why is this being removed?

Deniallugo (approved):

core/lib/config/src/configs/proof_data_handler.rs
Deniallugo: Remove as many config as you can.

Deniallugo (approved):

PR #1499, opened 2024-03-26
chore(vm): update contracts and bootloader for 6 blobs

koloz193:
## Checklist


core/lib/types/src/l2_to_l1_log.rs
```diff
@@ -61,16 +62,42 @@ pub fn l2_to_l1_logs_tree_size(protocol_version: ProtocolVersionId) -> usize {
...
     pub const VM_1_4_0_L2_L1_LOGS_TREE_SIZE: usize = 2048;
     pub const VM_1_4_2_L2_L1_LOGS_TREE_SIZE: usize = 4096;
+    pub const VM_1_5_0_L2_L1_LOGS_TREE_SIZE: usize = 16384;
 
     if protocol_version.is_pre_boojum() {
         PRE_BOOJUM_L2_L1_LOGS_TREE_SIZE
     } else if protocol_version.is_1_4_0() || protocol_version.is_1_4_1() {
         VM_1_4_0_L2_L1_LOGS_TREE_SIZE
-    } else {
+    } else if protocol_version.is_pre_1_5_0() {
```
StanislavBreadless, line 71: there is `is_1_4_2` function for protocol vesion

core/lib/types/src/l2_to_l1_log.rs
```diff
@@ -61,16 +62,42 @@ pub fn l2_to_l1_logs_tree_size(protocol_version: ProtocolVersionId) -> usize {
...
+    system_logs: &Vec<SystemL2ToL1Log>,
+) -> Vec<H256> {
+    let num_blobs = protocol_version.into_num_blobs() as u32;
+
+    let mut blob_hashes = system_logs
+        .iter()
+        .filter(|log| {
+            log.0.sender == PUBDATA_CHUNK_PUBLISHER_ADDRESS
+                && log.0.key >= H256::from_low_u64_be(BLOB1_LINEAR_HASH_KEY as u64)
+                && log.0.key < H256::from_low_u64_be((BLOB1_LINEAR_HASH_KEY + num_blobs) as u64)
```
StanislavBreadless, line 90: maybe better to have two functions for each protocol versions:

`protocol_version.into_num_blobs_created`
`protocol_version.into_num_blobs_required` ?

because here it works because we check for the `PUBDATA_CHUNK_PUBLISHER_ADDRESS` & it so happened that the only log (the upgrade one) that is lower than `BLOB1_LINEAR_HASH_KEY + 16` and not blob hash can be emitted by the bootloader only. In case we add a log to `PUBDATA_CHUNK_PUBLISHER_ADDRESS` this function will break

etc/env/base/chain.toml
```diff
@@ -87,9 +87,13 @@ save_call_traces = true
 virtual_blocks_interval = 1
 virtual_blocks_per_miniblock = 1
 
-bootloader_hash = "0x010008cb037f780742bcf34e05ff9b1a3140061de045a40192eb0424585521af"
-default_aa_hash = "0x01000563322a8444108dc763ebd80fb18ae14d2ec5c0fb5057a090023cd3d4a4"
+# WARNING! This slows down the statekeeper, forcing mempool to upload to GCS
+# It is meant as a validation flag to be used in STAGING only.
+# This variable should not be set to true in any customer facing environment.
+upload_witness_inputs_to_gcs = false
```
StanislavBreadless, line 93: ?

StanislavBreadless (approved):

PR #1500, opened 2024-03-26
feat(en): Rate-limit L2 client requests

slowli:
## What ❔

Implements rate-limiting for `HttpClient` (HTTP-based JSON-RPC client that we use in ENs to access main node Web3 APIs).

## Why ❔

Some request patterns made by ENs can be rate-limited (e.g., downloading L2 blocks during initial syncing).

## Checklist


core/lib/web3_decl/src/client.rs
slowli, line 1: Not sure whether this logic belongs to this crate, but I haven't found a better place for it.

core/lib/web3_decl/src/client.rs
```diff
@@ -0,0 +1,456 @@
...
+        L2ClientBuilder::default()
+    }
+
+    /// Sets the component operating this client. This is used in logging etc.
+    pub fn for_component(mut self, component_name: &'static str) -> Self {
+        self.component_name = component_name;
+        self
+    }
+
+    fn report_waiting(&self, origin: RateLimitOrigin, stats: &AcquireStats) {
```
slowli, line 66: Would it make sense to add a counter metric for slow requests (perhaps, with the method / component labels)?

core/node/node_framework/src/implementations/layers/web3_api/tx_sink.rs
```diff
@@ -33,9 +33,9 @@ impl WiringLayer for TxSinkLayer {
                 TxSinkResource(Arc::new(MasterPoolSink::new(pool)))
             }
             TxSinkLayer::ProxySink { main_node_url } => {
```
slowli, line 35: I think it makes sense to make `L2Client` a resource and get rid of `main_node_url` field in this variant. @popzxc WDYT?

core/bin/external_node/src/main.rs
```diff
@@ -128,22 +125,27 @@ async fn init_tasks(
...
                 .map(|version| version as u16);
 
-            EN_METRICS.version[&(format!("{}", version), protocol_version)].set(1);
+            EN_METRICS.version[&(version.to_string(), protocol_version)].set(1);
 
-            tokio::time::sleep(Duration::from_secs(10)).await;
+            // Error here corresponds to a timeout w/o `stop_receiver` changed; we're OK with this.
+            tokio::time::timeout(Duration::from_secs(10), stop_receiver_for_task.changed())
+                .await
+                .ok();
```
slowli, line 146: This and similar changes are sort of off-topic, but I got annoyed during local testing that the EN doesn't stop quickly after receiving a stop signal.

etc/env/ext-node.toml
```diff
@@ -48,6 +47,10 @@ mode = "FileBacked"
 file_backed_base_path = "artifacts"
 # ^ Intentionally set to coincide with main node's in order to read locally produced snapshots
 
+[en.main_node]
+url = "http://127.0.0.1:3050"
+rate_limit = { count = 10, window_ms = 30 }
```
tomg10, line 52: Why change defaults here?

tomg10: In general I'd say this rate limiting is more for us than for external partners

core/bin/external_node/src/main.rs
```diff
@@ -507,11 +503,17 @@ async fn main() -> anyhow::Result<()> {
         .main_node_url()
         .expect("Main node URL is incorrect");
     tracing::info!("Main node URL is: {main_node_url}");
-    let main_node_client = <dyn MainNodeClient>::json_rpc(&main_node_url)
+    let main_node_client = L2Client::builder()
+        .with_rate_limit(
```
tomg10, line 507: with my suggested changes, we could just have with_allowed_requests_per_second here

core/lib/web3_decl/src/client.rs
tomg10: I think it belongs here relatively well

core/lib/web3_decl/src/client.rs
```diff
@@ -0,0 +1,456 @@
...
+        }
+    }
+}
+
+/// JSON-RPC client for the main node with built-in middleware support.
+///
+/// The client should be used instead of `HttpClient` etc. A single instance of the client should be built
+/// and shared among all tasks run by the node in order to correctly rate-limit requests.
+#[derive(Debug, Clone)]
+pub struct L2Client {
```
tomg10, line 49: L2Client -> RateLimittedHttpClient as it's a wrapper for HttpClient?

core/lib/web3_decl/src/client.rs
```diff
@@ -0,0 +1,456 @@
...
+    where
+        R: DeserializeOwned,
+        Params: ToRpcParams + Send,
+    {
+        let stats = self.rate_limit.acquire(1).await;
+        self.report_waiting(RateLimitOrigin::Request(method), &stats);
+        self.inner.request(method, params).await
+    }
+
+    async fn batch_request<'a, R>(
```
tomg10, line 103: What is the use case for this method?

core/lib/web3_decl/src/client.rs
tomg10: Imo yes, or make rate limitting under warn, not debug (or both :P), now most of the partners won't be able to see it and so they won't tell us about it

core/lib/web3_decl/src/client.rs
```diff
@@ -0,0 +1,456 @@
...
+        &self,
+        batch: BatchRequestBuilder<'a>,
+    ) -> Result<BatchResponse<'a, R>, Error>
+    where
+        R: DeserializeOwned + fmt::Debug + 'a,
+    {
+        let stats = self.rate_limit.acquire(batch.iter().count()).await;
+        self.report_waiting(RateLimitOrigin::BatchRequest(&batch), &stats);
```
tomg10, line 111: I'd personally move those two lines (the same for other methods) into a method like throttle_if_needed (origin, requests_count)

core/lib/web3_decl/src/client.rs
```diff
@@ -0,0 +1,456 @@
...
+            rate_limit_window,
+            state,
+        }
+    }
+
+    /// Acquires the specified number of permits waiting if necessary. If the number of permits exceeds
+    /// the capacity of the limiter, it is saturated (i.e., this method cannot hang indefinitely or panic
+    /// in this case).
+    ///
+    /// This implementation is similar to `RateLimit` middleware in Tower, but is shared among client instances.
```
tomg10, line 232: I'd put an URL here to the Tower repo

tomg10 (changes requested):
Imo this implementation doesn't solve the problem that we were trying to solve, i.e. that some components may starve other components of requests.

core/bin/external_node/src/config/mod.rs
```diff
@@ -249,6 +250,13 @@ pub struct OptionalENConfig {
     // This is intentionally not a part of `RemoteENConfig` because fetching this info from the main node would defeat
     // its purpose; the consistency checker assumes that the main node may provide false information.
     pub contracts_diamond_proxy_addr: Option<Address>,
+    /// Number of permits in each rate limiting window applied to the main node HTTP client. Default is 10 permits.
+    #[serde(default = "OptionalENConfig::default_main_node_rate_limit_count")]
+    pub main_node_rate_limit_count: NonZeroUsize,
+    /// Duration of a rate limiting window applied to the main node HTTP client. Default is 50 ms.
+    /// Setting this duration to 0 will effectively disable rate limiting.
+    #[serde(default = "OptionalENConfig::default_main_node_rate_limit_window_ms")]
+    main_node_rate_limit_window_ms: u64,
```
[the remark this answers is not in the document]
slowli: A large window such as 1s could lead to stampeding requests, which we'd probably want to avoid. It's possible to hard-code a small window (IMO, it makes sense to do on the EN level, not in the `L2Client` definition, to allow for more flexibility) and re-compute the number of requests in this window based on the desired RPS. E.g., if a window is 50ms and the EN is configured with 100 rps limit, this would lead to 5 requests per window. Does this sound good?

etc/env/ext-node.toml
slowli: - Changing defaults is not strictly necessary; just did it to showcase the new params and test that they are parsed by the EN.
- Not sure I get your point about the intended audience. If *any* EN is switched off for some time for whatever reason, it would likely hit the rate limit in the block fetcher; so the issue doesn't look specific to our ENs.

core/lib/web3_decl/src/client.rs
slowli: If you mean why this method is here, it's a part of the `ClientT` trait, so I had no choice to *not* implement it. As to the use cases, JSON-RPC allows batching multiple requests in a single HTTP request, which is enabled by this method. We don't use it in the codebase because it's somewhat complex / less type-safe (but maybe we should 🙃). I've thought about implementing middleware that would automatically batch requests within a small time window (i.e., sort of debouncing), but it looks a complex task. And IMO it wouldn't remove the need of rate limiting in any case.

core/lib/web3_decl/src/client.rs
slowli: I've went with a more general name because we might want to introduce new middleware in the future and/or support multiple transports for the client (e.g., WS).

core/bin/external_node/src/config/mod.rs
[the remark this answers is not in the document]
RomanBrodetski: @slowli I think this is the best - it would not only make sure we don't do more than X requests per second, but distribute them (almost) uniformly over time. 

Re:  config vs no-config - probably indeed doing this the way you suggest and still configuring target RPS PER SECOND is a better interface.

core/bin/external_node/src/config/mod.rs
```diff
@@ -363,6 +367,13 @@ impl OptionalENConfig {
         10_000
     }
 
+    const fn default_main_node_rate_limit_rps() -> NonZeroUsize {
+        match NonZeroUsize::new(100) {
+            Some(value) => value,
+            None => unreachable!(),
+        }
```
popzxc, line 374: Bikeshedding: `NonZeroUsize::new(100).unwrap()` seems to be just enough and would make you think less about the construction. Subjective, of course.

core/lib/web3_decl/src/client/mod.rs
```diff
@@ -0,0 +1,357 @@
...
+/// JSON-RPC client for the main node with built-in middleware support.
+///
+/// The client should be used instead of `HttpClient` etc. A single instance of the client should be built
+/// and shared among all tasks run by the node in order to correctly rate-limit requests.
+#[derive(Debug, Clone)]
+pub struct L2Client {
+    inner: HttpClient,
+    rate_limit: SharedRateLimit,
+    component_name: &'static str,
+}
```
popzxc, line 78: Maybe stupid question, but still. Looking at jsonrpsee docs, `HttpClientBuilder` [allows setting a middleware](https://docs.rs/jsonrpsee-http-client/0.22.3/jsonrpsee_http_client/struct.HttpClientBuilder.html#method.set_http_middleware), won't it be possible to implement rate limiting through it to keep using `HttpClient` and avoid creating a wrapper client?

slowli: I've attempted to do this initially (see first commits of this PR), but have run into several issues:

- `HttpClient` allows to set HTTP-level middleware, not an RPC-level one, which I feel is more appropriate for rate limiting.
- There are hypothetical cases like automated "debouncing" batching of requests I've described [above](https://github.com/matter-labs/zksync-era/pull/1500#discussion_r1540952422), which wouldn't work with a HTTP-level middleware w/o significant hacks.
- Observability (logging, metrics) are also easier with RPC-level middleware.
- Naming the HTTP middleware type is problematic if middleware is complex. Tower provides [a boxed cloneable version of services](https://docs.rs/tower/latest/tower/util/struct.BoxCloneService.html), but it doesn't fit (it's unconditionally `!Sync`, and `Sync` is required for `HttpClient<_>` to implement RPC traits).

I'll probably add this explanation as a comment, so that it's not lost in the future.

tomg10 (dismissed):

RomanBrodetski (dismissed):

popzxc (approved):

PR #1501, opened 2024-03-26
feat(configs): Implement new format of configs and implement protobuf for it   

Deniallugo:
## What ❔

Loading new format of config from files. 
Only main node is fully moved to the ability to work from files.
Init process is intentionally has not be adapted for the file-based configurtaion
This format will be used with the new configuration 

## Why ❔

For having an opportunity not to rely on envs we can load files with necessary configuration.

## Checklist


core/bin/external_node/src/config/mod.rs
```diff
@@ -633,6 +633,9 @@ impl From<ExternalNodeConfig> for InternalApiConfig {
                 l2_weth_bridge: config.remote.l2_weth_bridge_addr,
             },
             bridgehub_proxy_addr: config.remote.bridgehub_proxy_addr,
+            // TODO set values
```
montekki, line 636: A `TODO` should not go as is into PR: it should be either fixed or a tracking issue should be added.

core/bin/zksync_server/src/main.rs
```diff
@@ -186,10 +187,17 @@ async fn main() -> anyhow::Result<()> {
     };
 
     // Run core actors.
-    let (core_task_handles, stop_sender, health_check_handle) =
-        initialize_components(&configs, &components, &secrets)
-            .await
-            .context("Unable to start Core actors")?;
+    let (core_task_handles, stop_sender, health_check_handle) = initialize_components(
```
montekki, line 190: Lets use a named fields struct and destruct it here; tuples with >= fields are easy to mess up.

core/lib/config/src/configs/contract_verifier.rs
```diff
@@ -20,4 +22,7 @@ impl ContractVerifierConfig {
     pub fn polling_interval(&self) -> Duration {
         Duration::from_millis(self.polling_interval.unwrap_or(1000))
     }
+    pub fn bind_addr(&self) -> SocketAddr {
+        SocketAddr::new("0.0.0.0".parse().unwrap(), self.port)
```
montekki, line 26:
```suggestion
        SocketAddr::new(IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)), self.port)
```

lets use a compile-time constant instead of parsing and unwrapping

core/lib/config/src/configs/wallets.rs
```diff
@@ -0,0 +1,55 @@
...
+            private_key: Some(private_key),
+        })
+    }
+
+    pub fn address(&self) -> Address {
+        self.address
+    }
+    pub fn private_key(&self) -> Option<H256> {
+        self.private_key
+    }
```
montekki, line 37: No need to do the copying, references can be returned.

core/lib/crypto_primitives/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+edition.workspace = true
+authors.workspace = true
+homepage.workspace = true
+repository.workspace = true
+license.workspace = true
+keywords.workspace = true
+categories.workspace = true
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
```
montekki, line 13:
```suggestion
```

core/lib/env_config/src/genesis.rs
```diff
@@ -1,19 +1,66 @@
...
     fn from_env() -> anyhow::Result<Self> {
         // Getting genesis from environmental variables is a temporary measure, that will be
         // re-implemented and for the sake of simplicity we combine values from different sources
         // #PLA-811
         let network_config = &NetworkConfig::from_env()?;
-        let contracts_config = &ContractsConfig::from_env()?;
+        let contracts_config = &ContractsForGenesis::from_env()?;
         let state_keeper = StateKeeperConfig::from_env()?;
+        let shared_bridge = if let Some(state_transition_proxy_addr) =
+            contracts_config.state_transition_proxy_addr
```
montekki, line 49: `if let Some(..) = { } else { none }` pattern can be simplified with a `.map()` i think here

core/lib/zksync_core/src/api_server/web3/namespaces/en.rs
```diff
@@ -91,6 +91,29 @@ impl EnNamespace {
             .await
             .context("genesis_config")?
             .context("Genesis not finished")?;
+
+        let shared_bridge = if self.state.api_config.state_transition_proxy_addr.is_some() {
+            Some(SharedBridge {
```
montekki, line 96: idk a `.map()` would probably be doable even with error handling.

core/lib/zksync_core/src/lib.rs
```diff
@@ -461,7 +448,11 @@ pub async fn initialize_components(
             task_futures.push(tokio::spawn(contract_verification::start_server(
                 connection_pool.clone(),
                 replica_connection_pool.clone(),
-                api_config.contract_verification.clone(),
+                configs
+                    .contract_verifier
+                    .clone()
+                    .context("Contract verifier")?
+                    .clone(),
```
montekki, line 455: these two consecutive clones are really necessary?

core/lib/env_config/src/genesis.rs
Deniallugo: yes, but we use `?` there and it doesn't work with map

core/bin/zksync_server/src/main.rs
Deniallugo: that's how it was before.. don't want to expand the scope of pr

core/lib/env_config/src/genesis.rs
montekki: you can [`.transpose()`](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=328425af65335aff22ad94831e5ccbe5) i think?

Deniallugo: Decided in DM that it doesn't worth it. 
Because we have to specify error type and the code won't be easier at the end

core/lib/protobuf_config/src/proto/contracts.proto
```diff
@@ -2,41 +2,31 @@ syntax = "proto3";
 
 package zksync.config.contracts;
 
+message L1 {
+  optional string governance_addr = 1; // required; H160
+  optional string verifier_addr = 6; // required; H160
```
pompon0, line 7: nit: fix the numbering of fields, given that it is not in use yet

core/lib/protobuf_config/src/proto/prover.proto
```diff
@@ -0,0 +1,99 @@
...
+  optional uint32 api_poll_duration_secs = 2; // required; s
+  optional uint32 prometheus_listener_port = 3; // required; u16
+  optional string prometheus_pushgateway_url = 4; // required
+  optional uint64 prometheus_push_interval_ms = 5; // optional; ms
+}
+
+
+message WitnessGenerator {
+  optional uint32 generation_timeout_in_secs = 1; // required;
+  optional uint32 basic_generation_timeout_in_secs = 8; // optional;
```
pompon0, line 73: numbering

core/lib/zksync_core/src/consensus/config.rs
```diff
@@ -69,7 +69,12 @@ impl Config {
             node_key,
             gossip_dynamic_inbound_limit: self.gossip_dynamic_inbound_limit,
             gossip_static_inbound: self.gossip_static_inbound.clone().into_iter().collect(),
-            gossip_static_outbound: self.gossip_static_outbound.clone().into_iter().collect(),
+            gossip_static_outbound: self
+                .gossip_static_outbound
+                .clone()
+                .into_iter()
+                .map(|(key, address)| (key, address))
```
pompon0, line 76: what happened here? `into_iter().collect()` doesn't compile any more?

pompon0 (approved):

montekki (approved):

core/bin/external_node/src/config/mod.rs
```diff
@@ -39,6 +41,8 @@ pub struct RemoteENConfig {
     pub l2_chain_id: L2ChainId,
     pub l1_chain_id: L1ChainId,
     pub max_pubdata_per_batch: u64,
+    pub l1_batch_commit_data_generator_mode: L1BatchCommitDataGeneratorMode,
+    pub dummy_prover: bool,
```
perekopskiy, line 45: I don't see any place where it's used, why was it added?

Deniallugo: I still want to keep it in genesis... because it shows the nature of the system

core/lib/config/src/configs/chain.rs
```diff
@@ -121,9 +121,9 @@ pub struct StateKeeperConfig {
     pub close_block_at_eth_params_percentage: f64,
     /// Denotes the percentage of L1 gas used in L2 block that triggers L2 block seal.
     pub close_block_at_gas_percentage: f64,
-
-    pub fee_account_addr: Address,
-
+    /// Fee account address. Value is deprecated and it's used only for generating wallets struct
+    #[deprecated]
+    pub fee_account_addr: Option<Address>,
```
popzxc, line 126: By the way, you can create a separate structure for deprecated one-time configs (e.g. in `zksync_env_config` crate), and deserialize them separately.
They'll have a comment explaining that these exist until we fully switch to a file-backed configuration scheme and aren't actually used by the core codebase otherwise.
This way you'll make sure that nobody actually uses these fields.

Not a strong requirement, though.

core/lib/config/src/configs/eth_sender.rs
```diff
@@ -3,21 +3,25 @@ use std::time::Duration;
 use serde::Deserialize;
 use zksync_basic_types::H256;
 
+use crate::ETHWatchConfig;
+
 /// Configuration for the Ethereum sender crate.
```
popzxc, line 8: Doc-comment is outdated.

core/lib/config/src/configs/eth_sender.rs
```diff
@@ -3,21 +3,25 @@ use std::time::Duration;
...
 #[derive(Debug, Deserialize, Clone, PartialEq)]
-pub struct ETHSenderConfig {
+pub struct ETHConfig {
     /// Options related to the Ethereum sender directly.
-    pub sender: SenderConfig,
+    pub sender: Option<SenderConfig>,
     /// Options related to the `GasAdjuster` submodule.
-    pub gas_adjuster: GasAdjusterConfig,
+    pub gas_adjuster: Option<GasAdjusterConfig>,
+    pub watcher: Option<ETHWatchConfig>,
```
popzxc, line 15: Not sure if mixing all of these is a good concept, but no big objections either.

core/lib/config/src/configs/general.rs
```diff
@@ -0,0 +1,35 @@
...
+        house_keeper::HouseKeeperConfig,
+        FriProofCompressorConfig, FriProverConfig, FriProverGatewayConfig,
+        FriWitnessGeneratorConfig, FriWitnessVectorGeneratorConfig, ObservabilityConfig,
+        PrometheusConfig, ProofDataHandlerConfig,
+    },
+    ApiConfig, ContractVerifierConfig, DBConfig, ETHConfig, PostgresConfig, SnapshotsCreatorConfig,
+};
+
+#[derive(Debug, PartialEq)]
+pub struct GeneralConfig {
```
popzxc, line 14: I don't really like this structure, given that it makes it possible to pass the whole `GeneralConfig` as an argument while expecting some (unknown at the call site) fields to be set. I would prefer to not have such a structure (e.g. we can deserialize each config we need from the same file).

core/lib/config/src/configs/wallets.rs
```diff
@@ -0,0 +1,71 @@
...
+#[derive(Debug, Clone)]
+pub struct StateKeeper {
+    pub fee_account: Wallet,
+}
+
+#[derive(Debug, Clone)]
+pub struct Wallets {
+    pub eth_sender: Option<EthSender>,
+    pub state_keeper: Option<StateKeeper>,
+}
```
popzxc, line 55: I'm not sure if I like the implications here.
First, the fee account private key is not needed by the core application. It's but an ability to express its presence sounds like a downside to me -- more ground to create problems out of nowhere.
Second, the semantics of "optional" private keys feel slightly off. It was done like this previously because it was a quickfix, but I'm not sure if we want to harden this kind of relationships. Two components may require a `Wallet` object: one may just need an address to read info, and one may be actually signing transactions, but you won't know from the constructor signature alone.
Third, configuration errors are not caught by this. If a component requires a private key to operate, but only address was provided, the config would successfully deserialized and the application would fail further in their lifecycle.
Finally, overall this kind of grouping seems natural on the first glance (these are accounts, accounts always have an address, but for some of them we don't know a private key), but I fail to see actual benefits. The PK and address may come from different sources in real life environment (public/secret configs), the information they expose have different levels of sensitivity (as stated above), some configurations can not be even expressed in theory (if an account is a smart wallet, we may have an address but it has no private key).

Probably I would much prefer keeping addresses and private keys separately, and ideally with clear indication of sensitivity (e.g. `EthSenderPrivateKey { key: H256 }` and `EthSender { /* other public configs */ }`.

popzxc: I also understand that in dev configs the PK and address would come together and it may be convenient to represent it this way from the dev environment POV, but we should design against the _operating_ environment.

RomanBrodetski: to add to this - we'll likely have the eth-sender as a separate service with separate deployments and configs. Please keep this in mind when designing the config system

Deniallugo: It's completely possible to use only part of the config

Deniallugo: Slightly changed the layout of protobuf

infrastructure/zk/src/run.ts
```diff
@@ -51,7 +51,7 @@ export async function revertReason(txHash: string, web3url?: string) {
 export async function catLogs(exitCode?: number) {
     utils.allowFailSync(() => {
         console.log('\nSERVER LOGS:\n', fs.readFileSync('server.log').toString());
-        console.log('\nPROVER LOGS:\n', fs.readFileSync('dummy_prover.log').toString());
+        console.log('\nPROVER LOGS:\n', fs.readFileSync('dummy_verifier.log').toString());
```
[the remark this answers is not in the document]
Deniallugo: no, but we will remove the whole zk all together, lets keep it as is for as long as we can

core/lib/config/src/configs/chain.rs
Deniallugo: It's already complicated, let's not overcomplicate it...

popzxc (dismissed):
I am still not quite fond of the new approach to the config separation, but after a private discussion with @Deniallugo it seems that he has a roadmap for this feature in mind.

Generally, given that the scope of the new functionality is huge, and given that it's a fairly important addition, I'm OK with merging as is, but hope that @Deniallugo will iterate on design. At any time I'll be happy to assist if this will be required.

etc/env/file_based/general.yaml
```diff
@@ -0,0 +1,315 @@
...
+  proof_compressor_stats_reporting_interval_ms: 10000
+  prover_job_archiver_reporting_interval_ms: 1800000
+  prover_job_archiver_archiving_interval_secs: 15
+
+prometheus:
+  listener_port: 3312
+  pushgateway_url: http://127.0.0.1:9091
+  push_interval_ms: 100
+
+observability:
```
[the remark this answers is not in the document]
Deniallugo: please add what we need to add here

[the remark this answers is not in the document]
Deniallugo: image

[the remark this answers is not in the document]
Deniallugo: this part is in place

popzxc (approved):

montekki (approved):

PR #1510, opened 2024-03-27
fix(prover): Remove FriProtocolVersionId

EmilLuta:
This is PR 2 of many to move us to safe Protocol Upgrades (phase 1).

`FriProtocolVersionId` was introduced as a cludge to serve as communication between core (which was saved from L1 straight to DB) and prover (which was manually configured from env vars). Whilst in theory it's possible to have an upgrade on core side without affecting prover, I don't envision this happening any time soon. Right now we maintain 2 different protocol numbers that cause confusion. This PR unifies the protocol ids.

Other relevant notes:
- this is not exactly backwards compatible -- we'll need to deploy both core and prover at the same time, otherwise provers will go in crash loop. We can make it backwards compatible, but at our track record of deployments (and speed of deployments), I'm concerned we will slow everything down for no real reason. I see very small risks here
- database tables in prover's end should be renamed to protocol_version as well (to be covered in next PR)
- all our code base assumes the old prover's interface to hashes which is confusing and unnecessary. I propose we address those with following PR. The points there are (to be done on both databases and configs): 
    - rename `recursion_scheduler_level_vk_hash` to `snark_wrapper_vk_hash`. This is what it represents today. I believe we can't change `recursion_scheduler_level_vk_hash` without changing `snark_wrapper_vk_hash`. 
    - delete `recursion_circuits_set_vks_hash`; it's all 0.

### DIFF
--- a/etc/env/base/proof_data_handler.toml
+++ b/etc/env/base/proof_data_handler.toml
@@ -1,5 +1,3 @@
 [proof_data_handler]
 http_port=3320
 proof_generation_timeout_in_secs=18000
-protocol_version_loading_mode="FromEnvVar"
-fri_protocol_version_id=2
--- a/prover/prover_fri_gateway/src/proof_gen_data_fetcher.rs
+++ b/prover/prover_fri_gateway/src/proof_gen_data_fetcher.rs
@@ -16,14 +16,14 @@ impl PeriodicApiStruct {
         let mut connection = self.pool.connection().await.unwrap();
         connection
             .fri_protocol_versions_dal()
-            .save_prover_protocol_version(data.fri_protocol_version_id, data.l1_verifier_config)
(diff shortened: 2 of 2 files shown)


core/lib/prover_interface/src/api.rs
```diff
@@ -14,7 +14,7 @@ use crate::{inputs::PrepareBasicCircuitsJob, outputs::L1BatchProofForL1};
 pub struct ProofGenerationData {
```
mm-zk, line 14: btw - where do we Serialize this struct?

mm-zk (approved):

mm-zk:
where does backwards incompatibility come from? (as AFAIK we actually used the same value for both)

Deniallugo (approved):

Flukena:
ok

PR #1515, opened 2024-03-28
feat(en): Rework storing and using protective reads

slowli:
## What ❔

- Does not persist protective reads if the node opts out of this.
- Handles absent protective reads in Metadata calculator (do not use them in the lightweight mode; warn in the full mode).

## Why ❔

Protective reads are only used by the full-mode tree to generate input for the prover. (Technically, they are also used to filter out inconsequential writes from storage slots touched in an L1 batch, but this logic can be removed / rewritten.)

## Checklist


perekopskiy (approved):

core/bin/external_node/src/config/mod.rs
```diff
@@ -237,13 +244,12 @@ pub struct OptionalENConfig {
...
-    pub mempool_cache_update_interval: u64,
-    /// Maximum number of transactions to be stored in the mempool cache. Default is 10000.
-    #[serde(default = "OptionalENConfig::default_mempool_cache_size")]
-    pub mempool_cache_size: usize,
+    /// Configures whether to persist protective reads when persisting L1 batches in the state keeper.
+    /// Protective reads are never required by full nodes so far, not until such a node runs a full Merkle tree
+    /// (presumably, to participate in L1 batch proving).
+    /// By default, set to `true` as a temporary safety measure.
+    #[serde(default = "OptionalENConfig::default_protective_reads_persistence_enabled")]
+    pub protective_reads_persistence_enabled: bool,
```
popzxc, line 252: Side-comment: we probably should revisit the EN docs to better describe possibilities for configuration. Doesn't have to be in this PR.

popzxc (approved):

core/bin/external_node/src/config/mod.rs
slowli: Yep, I think it makes sense to do separately. There's [a tracking task for this in Linear](https://linear.app/matterlabs/issue/PLA-850/update-en-docs).

PR #1521, opened 2024-03-29
fix(api): Change error code for Web3Error::NotImplemented

popzxc:
## What ❔

Changes the error code for `Web3Error::NotImplemented` from `InternalError` to `MethodNotFound`.

## Why ❔

Better representation of the nature of an error. Useful for better error-handling on the client side.

## Checklist


slowli (approved):
The relevant part of [the JSON-RPC spec](https://www.jsonrpc.org/specification#error_object) could be interpreted so that standard errors (including "Method not found") should (or must?) have specific `message`s, but I'm not sure. Personally, more specific error messages look fine.

popzxc:
I think that given that error code is standard, the clients should probably not rely on the contents of the message. Given the lack of explicit "MUST" there, it's probably fine to use a different error message.

PR #1522, opened 2024-03-29
feat(db): Wrap sqlx errors in DAL

slowli:
## What ❔

Wraps `sqlx::Error` in the core DAL crate using the existing instrumentation tools.

## Why ❔

This allows naturally providing additional context for errors w/o the need to manually specify it in all DAL call sites.

## Checklist


slowli:
Some general notes:

- This PR doesn't go for completeness; a significant part of DAL queries still returns `sqlx::Result<_>`. If the approach looks fine, I plan to convert these errors in one or more follow-up PRs.
- OTOH, I have converted *some* of DAL queries to understand which features are required from instrumentation (e.g., first-class `COPY` support; arg validation).
- I plan to add more docs in a follow-up PR as well.

tomg10 (approved):
This is exactly what I had in my mind when we talked about this, thanks for adding this and putting effort into porting significant part of the codebase, I don't even think we need to refactor the whole codebase now, if you don't have energy to, but big kudos if you do! 
Just one thing, I think it would be very useful, if you put a very short message on platform channel to advertise this change :) 
And other nit, i'd say it was quite risky to refactor that big part of the codebase and only then get review, you'd have wasted lots of time if any bigger changes to this approach were requested :P

slowli:
> i'd say it was quite risky to refactor that big part of the codebase and only then get review

Well yeah, but it gave me much better understanding at what kind of new APIs are required for DAL to support all functionality we want. Besides, a significant part of changes in this PR was mechanical (adding `instrument` / `with_arg` etc.), so it didn't require *that* much effort.

perekopskiy (dismissed):

core/lib/dal/src/consensus_dal.rs
```diff
@@ -26,15 +32,17 @@ impl ConsensusDal<'_, '_> {
...
         )
-        .fetch_optional(self.storage.conn())
+        .try_map(|row| {
+            row.genesis
+                .map(|genesis| {
+                    zksync_protobuf::serde::deserialize(genesis).decode_column("genesis")
+                })
+                .transpose()
+        })
+        .instrument("genesis")
```
pompon0, line 42: nit: why instrumenting a future, rather than transforming the result?

slowli: More opportunities for the instrumentation logic, some of which are used already (e.g., logic related to slow requests and measuring request latency).

perekopskiy (approved):

PR #1529, opened 2024-03-29
chore(genesis): add gas bound caller deployed

koloz193:
## Checklist


core/lib/constants/src/contracts.rs
```diff
@@ -125,6 +125,11 @@ pub const ERC20_TRANSFER_TOPIC: H256 = H256([
     99, 196, 161, 22, 40, 245, 90, 77, 245, 35, 179, 239,
 ]);
 
+pub const GAS_BOUND_CALLER_ADDRESS: Address = H160([
```
StanislavBreadless, line 128: please add a comment that the address is deliberately chosen to be a non-system one

PR #1530, opened 2024-03-31
refactor(prover): Idiomatic Eip4844 conversions

EmilLuta:
This PR introduces  idiomatic conversion for `Eip4844Blobs`. During the first implementation, `From` traits were used. `From` is not the right fit, given there are multiple not obvious conversions between blobs and vectors. Furthermore, the conversion can panic, which is against [guidelines](https://doc.rust-lang.org/std/convert/trait.From.html#when-to-implement-from).

This change moves to a more idiomatic function usage to avoid future confusion.

### DIFF
--- a/core/lib/basic_types/src/basic_fri_types.rs
+++ b/core/lib/basic_types/src/basic_fri_types.rs
@@ -31,10 +31,12 @@ impl Eip4844Blobs {
     pub fn blobs(self) -> Vec<Blob> {
         self.blobs
     }
-}
 
-impl From<Vec<u8>> for Eip4844Blobs {
-    fn from(blobs: Vec<u8>) -> Self {
+    pub fn encode(self) -> Vec<u8> {
+        self.blobs.iter().flatten().copied().collect()
+    }
+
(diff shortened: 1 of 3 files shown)


core/lib/basic_types/src/basic_fri_types.rs
```diff
@@ -31,10 +31,12 @@ impl Eip4844Blobs {
     pub fn blobs(self) -> Vec<Blob> {
         self.blobs
     }
-}
 
-impl From<Vec<u8>> for Eip4844Blobs {
-    fn from(blobs: Vec<u8>) -> Self {
+    pub fn encode(self) -> Vec<u8> {
+        self.blobs.iter().flatten().copied().collect()
```
montekki, line 36:
```suggestion
        let blobs_len = self.blobs.len() * EIP_4844_BLOB_SIZE;
        self.blobs
            .into_iter()
            .fold(Vec::with_capacity(blobs_len), |mut acc, v| {
                acc.extend(v);
                acc
            })
```

can save a bit of latency here by not reallocating.

PR #1532, opened 2024-03-31
feat(tree): Improved tree pruning

tomg10:
Changes in tree pruner to allow syncing pruning state with db pruning (pruning tree up to the same l1 batch as db)


core/lib/merkle_tree/src/pruning.rs
```diff
@@ -24,6 +24,28 @@ impl MerkleTreePrunerHandle {
     }
 }
 
+/// objects implementing this trait can be passed to pruner main loop, they act as a source of info about up to which version the tree should be pruned
+pub trait RetainedVersionSource {
```
slowli, line 28: This trait and its implementations look overly complicated. What I have in mind:

- Rework the pruning logic in the Merkle tree crate a bit more, so that it receives *commands* to prune past tree versions (probably via `MerkleTreePrunerHandle`). Commands are executed asynchronously, and we're OK if the pruner is shut down in the middle of a command; we'd just send a new command with a greater pruned version after a restart. To summarize, the difference with your approach is that `MerkleTreePruner` wouldn't drive pruning; external logic would.
- Use this logic straightforwardly in the metadata calculator: obtain a pruner handle, then send it to a Tokio task which would periodically query Postgres for new data to be pruned and send corresponding commands via the handle. Spawn a pruner itself in a blocking Tokio task. (So, 2 new Tokio tasks in total.) Manage both these tasks like other component tasks (i.e., make them cancel-aware etc.; `MerkleTreePruner::run()` should already be cancel-aware provided the handle is timely dropped). It could make sense to do this not in the `metadata_calculator` code, but rather in a higher-level glue code; compare with how `LazyAsyncTreeReader` can be obtained from `MetadataCalculator` via `tree_reader()` and then awaited for / used in other tasks.

core/lib/zksync_core/src/metadata_calculator/mod.rs
```diff
@@ -163,11 +166,28 @@ impl MetadataCalculator {
             "Merkle tree is initialized and ready to process L1 batches: {:?}",
             tree_reader.clone().info().await
         );
+        let (tree_pruner, pruner_handle) = tree.pruner();
+        let pruner_version_source = KeepPruningSyncedWithDbPruning {
+            pool: pool.clone(),
+            rt_handle: Handle::current(),
+        };
+        let pruner_thread = thread::spawn(move || tree_pruner.run(&pruner_version_source));
```
slowli, line 174: Would `tokio::task::spawn_blocking` not work here?

tomg10: corrected

core/lib/merkle_tree/src/pruning.rs
tomg10: corrected

core/lib/merkle_tree/src/pruning.rs
```diff
@@ -94,17 +111,24 @@ impl<DB: PruneDatabase> MerkleTreePruner<DB> {
...
+    pub fn prune_up_to(&mut self, target_retained_version: u64) -> Option<PruningStats> {
         let min_stale_key_version = self.db.min_stale_key_version()?;
+
+        //We must leave at least one version
+        let last_prunable_version = self.last_prunable_version();
+        if last_prunable_version.is_none() {
+            tracing::info!("Nothing to prune; skipping");
+            return None;
+        }
+        let target_retained_version = min(target_retained_version, last_prunable_version.unwrap());
```
slowli, line 131: Nit: You can use let-else or `?` on `last_prunable_version` for better code structuring and to avoid the `unwrap()` later.

core/lib/merkle_tree/src/pruning.rs
```diff
@@ -9,11 +14,11 @@ use crate::{
 
 /// Handle for a [`MerkleTreePruner`] allowing to abort its operation.
 ///
-/// The pruner is aborted once the handle is dropped.
-#[must_use = "Pruner is aborted once handle is dropped"]
+/// The pruner is aborted once the abort method on handle is called.
 #[derive(Debug)]
 pub struct MerkleTreePrunerHandle {
     aborted_sender: mpsc::Sender<()>,
+    target_retained_version: Arc<RwLock<Option<u64>>>,
```
slowli, line 21: As per access patterns, the lock is never held for a long amount of time and has an only reader, so using `Mutex` would be OK here. Alternatively, it's possible to use `AtomicU64` (target retained version set to 0 effectively means no pruning); would allow getting rid of poison checks.

core/lib/zksync_core/src/metadata_calculator/mod.rs
```diff
@@ -146,6 +151,34 @@ impl MetadataCalculator {
...
     }
 
+    async fn loop_pruning_tree(
+        &self,
+        pool: ConnectionPool<Core>,
+        tree_pruner: MerkleTreePruner<RocksDBWrapper>,
+        pruner_handle: MerkleTreePrunerHandle,
+        stop_receiver: watch::Receiver<bool>,
+    ) -> anyhow::Result<()> {
+        let pruner_thread = tokio::task::spawn_blocking(|| tree_pruner.run());
```
slowli, line 161: This looks too difficult and doesn't look like it'd fit well in the node framework. IMO, you should go for an approach in which all three tasks are publicly exposed and are supposed to be managed by the node framework:

```rust
// Metadata calculator constructor site, e.g. `main.rs` for the EN:
let mut metadata_calculator: MetadataCalculator = ..;
let (tree_pruner, tree_pruning_update_task) = metadata_calculator.enable_pruning(); 
// `tree_pruner` is an async / lazy wrapper for the pruner, similar to `LazyAsyncTreeReader`
task_futures.push(tokio::spawn(tree_pruner.run()));
// `tree_pruning_update_task` is a named wrapper for the logic in this method
task_futures.push(tokio::spawn(tree_pruning_update_task.run(stop_receiver.clone())));
```

Unlike with tree reader, pruner / pruning task are not cloneable, so they would probably need to be passed via oneshot channels (`enable_pruning()` would set the channel senders correspondingly).

core/lib/merkle_tree/src/domain.rs
```diff
@@ -61,6 +61,12 @@ impl ZkSyncTree {
             .expect("failed initializing `rayon` thread pool")
     }
 
+    /// Returns tree pruner and a handle to stop it
+    pub fn pruner(&self) -> (MerkleTreePruner<RocksDBWrapper>, MerkleTreePrunerHandle) {
```
slowli, line 65: - Bikeshedding: Please move this method below the constructor methods (i.e., `new_with_mode()`), so that getters are logically grouped.
- Don't quite like this interface because it easily allows creating multiple pruners for the same tree. It would be more idiomatic to integrate creating pruner into the tree constructor (probably via a builder since otherwise the constructor would become too messy). OTOH, it's certainly not the highest-priority issue; it's possible to clone the underlying `RocksDBWrapper` and misuse the tree APIs in this way.

Use of a hypothetical builder:

```rust
let builder = ZkSyncTree::builder(db);
let (pruner, pruner_handle) = builder.enable_pruning(); // takes &self
let tree = builder.build(); // or .build_lightweight() for the lightweight tree
```

...Then, you would stash `(pruner, pruner_handle)` in `AsyncTree` and allow to take them from it.

core/lib/merkle_tree/src/pruning.rs
tomg10: I deliberately wanted to avoid let-else as I generally try to keep indentation level as low as possible for bigger blocks of code, ? is a good suggestion

core/lib/merkle_tree/src/pruning.rs
tomg10: I agree mutex would also fit here, though I don't see it being better. I do see advantages of using AtomicU64 though, I completely haven't though about using it, nice!

tomg10: The only downside is that we can't store None in it, but there is a safe default, i.e. 0

core/lib/merkle_tree/src/domain.rs
tomg10: I don't like the use of builder pattern here, for me it looks awkward that ZkSyncTree builder can build different type of object. And how it's currently done is consistent with reader constructor. If we really want to make sure there is only one pruner, imo it makes more sense to leave the interface like it is, I've added a guardrail though

core/lib/zksync_core/src/metadata_calculator/pruning.rs
```diff
@@ -0,0 +1,241 @@
...
+            _ = stop_receiver.changed() => {
+                tracing::info!("Stop signal received before Merkle tree is initialized; shutting down tree pruning");
+                return Ok(());
+            }
+        }
+        tracing::info!("Obtained pruning handles; starting Merkle tree pruning");
+
+        // FIXME: is this good enough (vs a managed task)?
+        pruner.set_poll_interval(self.poll_interval);
+        let pruner_task_handle = tokio::task::spawn_blocking(|| pruner.run());
```
slowli, line 55: Subj. On one hand, `pruner` can panic, which won't be detected immediately by the current setup (it could be changed to do that; e.g., by checking the reference count of `target_retained_version` in `pruner_handle.set_target_retained_version` and returning an error if it's 1). On the other hand, since it's a blocking task, it cannot be stopped and fundamentally depends on `MerkleTreePruningTask` to be cancel-aware.

perekopskiy (dismissed):

perekopskiy (approved):

PR #1537, opened 2024-04-01
feat: Archiving of prover in gpu_prover_queue

Artemka374:
## Checklist

### DIFF
--- a/core/lib/zksync_core/src/house_keeper/fri_gpu_prover_archiver.rs
+++ b/core/lib/zksync_core/src/house_keeper/fri_gpu_prover_archiver.rs
@@ -1,0 +1,51 @@
+use prover_dal::{Prover, ProverDal};
+use zksync_db_connection::connection_pool::ConnectionPool;
+
+use crate::house_keeper::periodic_job::PeriodicJob;
+
+#[derive(Debug)]
+pub struct FriGpuProverArchiver {
+    pool: ConnectionPool<Prover>,
+    reporting_interval_ms: u64,
+    archiving_interval_secs: u64,
+}
+
+impl FriGpuProverArchiver {
+    pub fn new(
+        pool: ConnectionPool<Prover>,
+        reporting_interval_ms: u64,
+        archiving_interval_secs: u64,
+    ) -> Self {
+        Self {
+            pool,
+            reporting_interval_ms,
+            archiving_interval_secs,
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl PeriodicJob for FriGpuProverArchiver {
+    const SERVICE_NAME: &'static str = "FriGpuProverArchiver";
+
+    async fn run_routine_task(&mut self) -> anyhow::Result<()> {
+        let archived_provers = self
+            .pool
+            .connection()
+            .await
+            .unwrap()
+            .fri_gpu_prover_queue_dal()
+            .archive_old_provers(self.archiving_interval_secs)
+            .await;
+        tracing::info!("Archived {:?} fri gpu prover records", archived_provers);
+        metrics::counter!(
+            "server.prover_fri.archived_provers",
+            archived_provers as u64
+        );
+        Ok(())
+    }
+
+    fn polling_interval_ms(&self) -> u64 {
+        self.reporting_interval_ms
+    }
+}
--- a/core/lib/zksync_core/src/lib.rs
+++ b/core/lib/zksync_core/src/lib.rs
@@ -68,6 +68,7 @@ use crate::{
     genesis::GenesisParams,
     house_keeper::{
         blocks_state_reporter::L1BatchMetricsReporter,
+        fri_gpu_prover_archiver::FriGpuProverArchiver,
         fri_proof_compressor_job_retry_manager::FriProofCompressorJobRetryManager,
         fri_proof_compressor_queue_monitor::FriProofCompressorStatsReporter,
         fri_prover_job_retry_manager::FriProverJobRetryManager,
@@ -1087,6 +1088,20 @@ async fn add_house_keeper_to_task_futures(
                 .unwrap(),
         );
         let task = fri_prover_jobs_archiver.run(stop_receiver.clone());
+        task_futures.push(tokio::spawn(task));
+    }
+
+    if house_keeper_config.fri_gpu_prover_archiver_enabled() {
+        let fri_gpu_prover_jobs_archiver = FriGpuProverArchiver::new(
+            prover_connection_pool.clone(),
+            house_keeper_config
+                .fri_gpu_prover_archiver_reporting_interval_secs
+                .unwrap(),
+            house_keeper_config
+                .fri_gpu_prover_archiver_archiving_interval_secs
+                .unwrap(),
+        );
+        let task = fri_gpu_prover_jobs_archiver.run(stop_receiver.clone());
         task_futures.push(tokio::spawn(task));
     }
 
--- a/etc/env/base/house_keeper.toml
+++ b/etc/env/base/house_keeper.toml
@@ -14,3 +14,5 @@ fri_proof_compressor_job_retrying_interval_ms = 30000
 fri_proof_compressor_stats_reporting_interval_ms = 10000
 fri_prover_job_archiver_reporting_interval_ms = 1800000
 fri_prover_job_archiver_archiving_interval_ms = 172800
+fri_gpu_prover_archiver_reporting_interval_secs = 86400
+fri_gpu_prover_archiver_archiving_interval_secs = 172800
--- a/core/lib/config/src/configs/house_keeper.rs
+++ b/core/lib/config/src/configs/house_keeper.rs
@@ -19,6 +19,8 @@ pub struct HouseKeeperConfig {
     // TODO(PLA-862): Make these 2 variables required
     pub fri_prover_job_archiver_reporting_interval_ms: Option<u64>,
     pub fri_prover_job_archiver_archiving_interval_secs: Option<u64>,
+    pub fri_gpu_prover_archiver_reporting_interval_secs: Option<u64>,
+    pub fri_gpu_prover_archiver_archiving_interval_secs: Option<u64>,
 }
 
 impl HouseKeeperConfig {
@@ -26,6 +28,14 @@ impl HouseKeeperConfig {
         self.fri_prover_job_archiver_reporting_interval_ms.is_some()
             && self
                 .fri_prover_job_archiver_archiving_interval_secs
+                .is_some()
+    }
+
+    pub fn fri_gpu_prover_archiver_enabled(&self) -> bool {
+        self.fri_gpu_prover_archiver_reporting_interval_secs
+            .is_some()
+            && self
+                .fri_gpu_prover_archiver_archiving_interval_secs
                 .is_some()
     }
 }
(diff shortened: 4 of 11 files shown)


etc/env/base/house_keeper.toml
```diff
@@ -13,4 +13,6 @@ fri_prover_stats_reporting_interval_ms = 30000
 fri_proof_compressor_job_retrying_interval_ms = 30000
 fri_proof_compressor_stats_reporting_interval_ms = 10000
 fri_prover_job_archiver_reporting_interval_ms = 1800000
-fri_prover_job_archiver_archiving_interval_ms = 172800
\ No newline at end of file
+fri_prover_job_archiver_archiving_interval_ms = 172800
+fri_gpu_prover_archiver_reporting_interval_secs = 86400
+fri_gpu_prover_archiver_archiving_interval_secs = 172800
```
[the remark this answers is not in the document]
Artemka374: Hmm, I don't get why it can bring provers down. Could you elaborate?

prover/prover_fri/src/gpu_prover_availability_checker.rs
```diff
@@ -0,0 +1,62 @@
...
+                    .connection()
+                    .await
+                    .unwrap()
+                    .fri_gpu_prover_queue_dal()
+                    .get_prover_instance_status(self.address.clone(), self.zone.clone())
+                    .await;
+
+                if status.is_none() || status.unwrap() == GpuProverInstanceStatus::Dead {
+                    tracing::info!(
+                        "Prover instance at address {:?}, availability zone {} was found marked as dead while being alive, shutting down",
```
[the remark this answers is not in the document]
Artemka374: The obvious answer doesn't come to my head :) I think we can do as you suggest

etc/env/base/house_keeper.toml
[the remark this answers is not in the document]
Artemka374: But provers cannot just "resurrect" themselves. If they are dead - they are dead anyway. If they will be killed - new will be spawned. Or am I missing something?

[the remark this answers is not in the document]
Artemka374: But in that case we need to keep possible None status. Right, that just came to my mind! That's why we would need to check not only that status is dead, but also that it exists in the table. Since availability checker and archiver are not bound with each other - at some point archiver can archive the prover - after that the check will occur and it will panic. I don't think that it really depends on the value that we set - it's just possible.

core/lib/zksync_core/src/lib.rs
```diff
@@ -1122,6 +1123,20 @@ async fn add_house_keeper_to_task_futures(
...
 
+    if house_keeper_config.fri_gpu_prover_archiver_enabled() {
+        let fri_gpu_prover_jobs_archiver = FriGpuProverArchiver::new(
+            prover_connection_pool.clone(),
+            house_keeper_config
+                .fri_gpu_prover_archiver_reporting_interval_secs
+                .unwrap(),
+            house_keeper_config
+                .fri_gpu_prover_archiver_archiving_interval_secs
+                .unwrap(),
```
RomanBrodetski, line 1134: bikeshedding: here we check the fact that if `fri_gpu_prover_archiver_enabled` is true, then both fields are `Some` in runtime. We could organize it in a single method that would encapsulate this logic and avoid unwrtaps. We could do

```
if Some((reporting_interval_secs, archiving_interval_secs)) = house_keeper_config.fri_gpu_prover_archiver_params {
//component enabled
} else {
// component disabled
}
```

core/lib/zksync_core/src/metrics.rs
```diff
@@ -183,3 +183,13 @@ pub(crate) struct ExternalNodeMetrics {
 
 #[vise::register]
 pub(crate) static EN_METRICS: vise::Global<ExternalNodeMetrics> = vise::Global::new();
+
+#[derive(Debug, Metrics)]
+#[metrics(prefix = "house_keeper")]
+pub(crate) struct HouseKeeperMetrics {
+    pub prover_job_archived_amount: Counter,
```
RomanBrodetski, line 190: nit: IMO the `_amount` in the end doesn't improve readability since it's already in metrics/counter context. Note that this name also ends up in prometheus/grafana, where it's concatenated with all the prefixes - already long.

etc/env/base/house_keeper.toml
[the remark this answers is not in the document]
RomanBrodetski: side note: In additional to the potential comment I would really revisit the variable naming - see above. Currently it indeed reads like an interval on which something's happening, and it does feel innocent to reduce it. If we call it something like `max_age`, it's much more clear that this might have business logic reprecussions

prover/prover_dal/migrations/20240403070124_add_archived_at_column_to_prover_queue_archive.down.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE gpu_prover_queue_fri_archive
+    DROP COLUMN IF EXISTS archived_at;
```
RomanBrodetski, line 2: please configure your editor to add a new line in the end of file on save

prover/prover_fri/src/gpu_prover_availability_checker.rs
```diff
@@ -0,0 +1,77 @@
...
+                            self.address,
+                            self.zone
+                        );
+                        return Ok(());
+                    }
+                    Some(_) => (),
+                }
+
+                tokio::time::sleep(std::time::Duration::from_secs(
+                    self.polling_interval_secs as u64,
```
RomanBrodetski, line 67: why convert it to `Duration` on every iteration? Let's just store `Duration` in field

prover/prover_fri/src/gpu_prover_availability_checker.rs
```diff
@@ -0,0 +1,77 @@
...
+
+                match status {
+                    None => {
+                        METRICS.zombie_prover_instances_count[&KillingReason::Absent].inc();
+                        tracing::info!(
+                            "Prover instance at address {:?}, availability zone {} was not found in the database, shutting down",
+                            self.address,
+                            self.zone
+                        );
+                        return Ok(());
```
RomanBrodetski, line 52: I understand the process will shut down when any of the tasks returns, but it's not obvious from this code. From this code one just see the log message "shutting down" and then `return Ok()`. I would add a comment explaining the flow.

prover/prover_fri/src/metrics.rs
```diff
@@ -15,6 +17,14 @@ pub(crate) enum Layer {
     Base,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, EncodeLabelSet, EncodeLabelValue)]
+#[metrics(label = "reason", rename_all = "snake_case")]
+#[allow(dead_code)]
+pub(crate) enum KillingReason {
+    Dead,
+    Absent,
```
RomanBrodetski, line 25: I think some comments would be useful here - what each means?

RomanBrodetski (changes requested):
Also, please add comments on the logic of components that you introduce here

core/lib/zksync_core/src/house_keeper/fri_gpu_prover_archiver.rs
```diff
@@ -0,0 +1,52 @@
+use prover_dal::{Prover, ProverDal};
+use zksync_db_connection::connection_pool::ConnectionPool;
+
+use crate::{house_keeper::periodic_job::PeriodicJob, metrics::HOUSE_KEEPER_METRICS};
+
+/// FriGpuProverArchiver is a task that periodically archives old fri GPU prover records.
+/// The task will archive the `dead` prover records that have not been updated for a certain amount of time.
```
RomanBrodetski, line 7: This comment does not correspond to the logic defined in `dal`

prover/prover_fri/src/gpu_prover_availability_checker.rs
```diff
@@ -0,0 +1,81 @@
...
+                    .await
+                    .unwrap()
+                    .fri_gpu_prover_queue_dal()
+                    .get_prover_instance_status(self.address.clone(), self.zone.clone())
+                    .await;
+
+                // If the prover instance is not found in the database or marked as dead, we should shut down the prover
+                match status {
+                    None => {
+                        METRICS.zombie_prover_instances_count[&KillingReason::Absent].inc();
```
RomanBrodetski, line 51: what is a possible scenario that would result in this case? I only see bugs leading to this, so this should be a `warn`  or even `error`. Seeing how it is handled here makes me think I'm missing smt

Artemka374: It can appear in case the archiver have cleaned up dead instance first and only after that the check was executed, so it will be absent in the db

core/lib/zksync_core/src/house_keeper/fri_gpu_prover_archiver.rs
Artemka374: That is actually mistake of the dal implementation, thanks for noticing!

Artemka374: Should be okay now

prover/prover_fri/src/gpu_prover_availability_checker.rs
Artemka374: If thinking about it once again, it shouldn't just simply appear - it will depend on the env variables set. If we stick to what we have now(archiving_interval=24h, archive_after=48h, check_interval=30m) - checker will always kill the prover first(because the updated_at value in this prover will be too small), but if someone will change the variables not in the right way - we will be able to still handle this situation and the prover will be killed. I would say that if we will receive Absent status - it is anyway worse than just dead, so I agree that it would be good to have warn or error, probably the alert for such status should have higher priority as well

core/lib/protobuf_config/src/proto/prover.proto
```diff
@@ -1,99 +1,101 @@
...
-  optional uint32 compression_mode = 1; // required; u8
-  optional uint32 prometheus_listener_port = 2; // required; u16
-  optional string prometheus_pushgateway_url = 3; // required
-  optional uint64 prometheus_push_interval_ms = 4; // optional; ms
-  optional uint32 generation_timeout_in_secs = 5; // required; s
-  optional uint32 max_attempts = 6; // required
-  optional string universal_setup_path = 7; // required; fs path
-  optional string universal_setup_download_url = 8; // required
-  optional bool verify_wrapper_proof = 9; // required
+    optional uint32 compression_mode = 1; // required; u8
```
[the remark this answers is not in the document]
Artemka374: Ah, it is enforced on fmt/commiting, I tried to commit without it, but I need to fight with my IDE for it.

prover/prover_fri/src/gpu_prover_availability_checker.rs
RomanBrodetski: >  I'd defer to @RomanBrodetski if he wants more specific level.

I don't want any particular level, my question is more "if you want to notice that such events started to happen". E.g.  if it's indicattive of bug, you probably want to learn it sooner. Second Artem's message may suggest we do.

In the end 100% your decision - feel free to keep as is

RomanBrodetski (approved):

PR #1538, opened 2024-04-01
feat: Migrate gas limit to u64

StanislavBreadless:
## What ❔

⚠️ **Warning: this PR requires DB migration** ⚠️ 

In preparation for the 1.5.0 upgrade:

- We now use u64 to represent the `gasLimit`. In most places we already have bigdecimal / U256 there, so there are not many changes there. 
- Since the gas limit will be able to exceed u32 in 1.5.0, it means that the `block.gaslimit` should be able to exceed it to in the 1.5.0 upgrade too. Our values should return consistent block gas limit with the system contracts, so we have to return it here also. Due to this part, we will require a migration, to store the gas limit for each block.

Follow-up tasks needed:
- Add a seal criteria to prevent the sum of gas limits of transactions in a batch exceed the batch gas limit. Note, that in the first release of the 1.5.0 this property won't be enforced.

## Why ❔

To prepare for 1.5.0 upgrade, where gas limit can exceed u32::MAX

## Checklist


core/lib/dal/migrations/20240401105036_block-gas-limit.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE miniblocks ADD COLUMN gas_limit BIGINT;
```
perekopskiy, line 1: BIGINT is `i64`, it won't fit `u64::MAX`, you should use `NUMERIC`

Deniallugo: while we are here, should we move to u256?

StanislavBreadless: i64 is enough for our purposes, in reality the block gas limit we'll use is 2^50, which fits into i64

StanislavBreadless: >while we are here, should we move to u256?

I am fine with that, especially since we anyway use `U256` for the `Fee` struct, so if you are fine with migrating to numeric, I am fine also, but you can better assess the impact on performance

core/lib/multivm/src/interface/types/inputs/system_env.rs
```diff
@@ -10,7 +10,7 @@ pub struct SystemEnv {
     pub zk_porter_available: bool,
     pub version: ProtocolVersionId,
     pub base_system_smart_contracts: BaseSystemContracts,
-    pub gas_limit: u32,
+    pub computational_gas_limit: u32,
```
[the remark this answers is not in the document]
StanislavBreadless: Maybe call it `bootloader_gas_limit`? This field defines only how much will get into bootloader (for both old and the new versions). In theory, we could spend more compute than what is written here.

core/lib/multivm/src/versions/vm_m5/transaction_data.rs
```diff
@@ -229,14 +229,13 @@ impl TransactionData {
...
     gas_price_per_pubdata: u32,
     encoded_len: usize,
 ) -> u32 {
-    assert!(
-        gas_limit <= MAX_TX_ERGS_LIMIT,
-        "gas limit is larger than the maximal one"
-    );
+    // Even if the gas limit is greater than the `MAX_TX_ERGS_LIMIT`, we assume that everything beyond `MAX_TX_ERGS_LIMIT`
+    // will be spent entirely on publishing bytecodes and so we derive the overhead solely based on the capped value
+    let gas_limit = std::cmp::min(MAX_TX_ERGS_LIMIT as u64, gas_limit);
```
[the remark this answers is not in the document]
StanislavBreadless: this is a bit more robust approach, considering that we want to allow proving u64 there. During block reexecution it will behave the same (in the past it was:

```
if x > MAX_TX_ERGS_LIMIT {
  panic
} else {
  x
}
```

now it is:

```
return min(x, MAX_TX_ERGS_LIMIT) => 

if x > MAX_TX_ERGS_LIMIT {
  //This should've never happened
} else {
  x
}
```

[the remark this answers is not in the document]
perekopskiy: Makes sense

core/lib/multivm/src/interface/types/inputs/system_env.rs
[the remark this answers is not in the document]
perekopskiy: Agree, it's even better

core/lib/multivm/src/versions/vm_1_4_2/tests/tester/vm_tester.rs
```diff
@@ -176,8 +176,8 @@ impl<H: HistoryMode> VmTesterBuilder<H> {
         self
     }
 
-    pub(crate) fn with_gas_limit(mut self, gas_limit: u32) -> Self {
-        self.system_env.gas_limit = gas_limit;
+    pub(crate) fn with_computational_gas_limit(mut self, gas_limit: u32) -> Self {
```
perekopskiy, line 179: nit: with_bootloader_gas_limit

perekopskiy (dismissed):

perekopskiy (approved):

PR #1539, opened 2024-04-01
feat(sqlx): Use offline mode by default

Deniallugo:
## What ❔

* Introduce new zk command `zk db prepare`
* Add sqlx_offline=true to envs
* remove build.rs from prover  

## Why ❔

It allows us to make build faster and remove dependency from envs

## Checklist


etc/env/base/database.toml
```diff
@@ -1,16 +1,17 @@
+sqlx_offline = true
```
popzxc, line 1: Please add a comment mentioning that we intentionally want the workspace to use offline mode by default and need to run `zk db prepare` to update `sqlx` files.

perekopskiy (approved):

perekopskiy (approved):

popzxc (dismissed):

popzxc (approved):

PR #1545, opened 2024-04-02
feat(api): Add `tokens_whitelisted_for_paymaster`

perekopskiy:
## What ❔

- Adds `tokens_whitelisted_for_paymaster` config var to main node.
- Adds `en_tokensWhitelistedForPaymaster` endpoint.

## Why ❔

Required to whitelist some tokens to be used by paymaster in addition to natively bridged tokens.

## Checklist


core/bin/external_node/src/main.rs
```diff
@@ -424,19 +430,41 @@ async fn run_api(
...
+                        *tokens_whitelisted_for_paymaster_cache_clone.write().await = tokens;
+                    }
+                    Err(err) => {
+                        tracing::error!(
+                            "Failed to query `tokens_whitelisted_for_paymaster`, error: {err:?}"
+                        );
+                    }
+                }
+
+                tokio::time::sleep(Duration::from_secs(60)).await;
```
slowli, line 449: Please make the task cancel-aware [like here](https://github.com/matter-labs/zksync-era/blob/4f803711da6df5e75ef2c08a42b0029a9a84969a/core/bin/external_node/src/main.rs#L518) so that the EN doesn't hang up on it during shutdown.

core/bin/external_node/src/main.rs
```diff
@@ -424,19 +430,41 @@ async fn run_api(
...
+        let whitelisted_tokens_update_task = task::spawn(async move {
+            loop {
+                match main_node_client.tokens_whitelisted_for_paymaster().await {
+                    Ok(tokens) => {
+                        *tokens_whitelisted_for_paymaster_cache_clone.write().await = tokens;
+                    }
+                    Err(err) => {
+                        tracing::error!(
+                            "Failed to query `tokens_whitelisted_for_paymaster`, error: {err:?}"
+                        );
```
slowli, line 445: What would be the impact if the whitelisted tokens are desynced between the main node and an EN? Would the VM-related RPC methods work incorrectly? (I suppose we cannot avoid temporary desyncing in any case, unless we sync via an L2 smart contract or something like that.)

perekopskiy: If there are tokens that are in main node's list but not on EN, then EN will reject all transactions that call this token addresses during validation, the most popular scenario is when paymaster calls token contract.
If there are tokens that are in EN's list but not on main node, then EN can accept some transactions that will be rejected by main node.

core/bin/external_node/src/config/mod.rs
```diff
@@ -707,6 +707,8 @@ impl From<ExternalNodeConfig> for TxSenderConfig {
                 .optional
                 .l1_to_l2_transactions_compatibility_mode,
             max_pubdata_per_batch: config.remote.max_pubdata_per_batch,
+            // Does not matter for EN.
+            tokens_whitelisted_for_paymaster: None,
```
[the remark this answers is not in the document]
RomanBrodetski: In theoery each validator can have different parameters here, and decide for themself.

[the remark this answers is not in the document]
perekopskiy: > decide for themself

That's for sure. But I think most operators would like to use some default list. And it's an open question how to maintain this list.

core/node/node_framework/src/implementations/layers/web3_api/tx_sender.rs
```diff
@@ -92,6 +93,7 @@ impl WiringLayer for TxSenderLayer {
         }));
 
         // Build `TxSender`.
+        let whitelisted_tokens_for_aa = self.tx_sender_config.whitelisted_tokens_for_aa.clone();
         let mut tx_sender = TxSenderBuilder::new(self.tx_sender_config, replica_pool, tx_sink);
```
slowli, line 97: Nit: Since tokens are a part of `tx_sender_config`, maybe it makes sense to initialize tokens `TxSenderBuilder` using the config and allow to override this value?

```rust
impl TxSenderBuilder {
    pub fn new(config, ..) -> Self {
        // initializes tokens from `config`
    }

    pub fn with_whitelisted_tokens_cache(mut self, cache: Arc<RwLock<_>>) -> Self {
        // sets the cache
    }
}
```

core/lib/zksync_core/src/api_server/tx_sender/mod.rs
```diff
@@ -184,6 +185,7 @@ impl TxSenderBuilder {
         vm_concurrency_limiter: Arc<VmConcurrencyLimiter>,
         api_contracts: ApiContracts,
         storage_caches: PostgresStorageCaches,
+        whitelisted_tokens_for_aa_cache: Arc<RwLock<Vec<Address>>>,
```
slowli, line 188: I guess from the perspective of the node framework, we'd want an enterprise-y interface like

```rust
#[async_trait]
pub trait WhitelistedTokensProvider: 'static + Send + Sync + fmt::Debug {
    async fn whitelisted_tokens(&self) -> Vec<Address>;
}

// Implementation used by the main node
pub struct WhitelistedTokensProvider(Vec<Address>);

// Implementation used by the EN
#[derive(Debug)]
pub struct RpcWhitelistedTokensProvider(Arc<RwLock<Vec<Address>>>);

// Named task type for convenience.
#[derive(Debug)]
pub struct WhitelistedTokensUpdateTask {
    tokens: Arc<RwLock<Vec<Address>>>,
    main_node_client: _,
}

impl WhitelistedTokensUpdateTask {
    pub fn run(self, stop_receiver: watch::Receiver<bool>) -> anyhow::Result<()> {
        // ...
    }
}
```

Then again, I'm not sure how stable this solution will be (judging by previous discussion, it seems somewhat ad-hoc to me), so maybe it would be an overkill.

core/lib/zksync_core/src/api_server/web3/namespaces/en.rs
```diff
@@ -156,4 +156,16 @@ impl EnNamespace {
...
+    #[tracing::instrument(skip(self))]
+    pub async fn whitelisted_tokens_for_aa_impl(&self) -> Result<Vec<Address>, Web3Error> {
+        Ok(self
+            .state
+            .tx_sender
+            .0
+            .whitelisted_tokens_for_aa_cache
+            .read()
+            .await
+            .clone())
```
slowli, line 169: Nit: may make sense to extract this getter to a method, since it's used in several places.

core/lib/zksync_core/src/api_server/tx_sender/mod.rs
perekopskiy: Yeah, exactly. I decided not to integrate this task into node framework for now. In addition to solution being stable, I'm not sure about the best way to add such "small" tasks to the framework, there are bunch of them running in EN and they are not integrated into framework yet.

RomanBrodetski (dismissed):

RomanBrodetski (approved):

RomanBrodetski (approved):

PR #1546, opened 2024-04-02
feat(node-framework): Add consensus support

popzxc:
## What ❔

- Moves consensus-related logic to the consensus module (so that all the logic in its own dedicated task).
- Creates a layer for the consensus task.

## Why ❔

mhm

## Checklist

### DIFF
--- a/core/bin/external_node/src/main.rs
+++ b/core/bin/external_node/src/main.rs
@@ -219,7 +219,7 @@ async fn run_core(
                 anyhow::bail!("Consensus config is specified, but secrets are missing")
             }
             (None, _) => {
-                // Secrets may be unconditionally embedded in some envs, but they are unused
+                // Secrets may be unconditionally embedded in some environments, but they are unused
                 // unless a consensus config is provided.
                 None
             }
--- a/core/bin/external_node/src/main.rs
+++ b/core/bin/external_node/src/main.rs
@@ -8,7 +8,7 @@ use tokio::{
     sync::watch,
     task::{self, JoinHandle},
 };
-use zksync_concurrency::ctx;
+use zksync_concurrency::{ctx, scope};
 use zksync_config::configs::{
     api::MerkleTreeApiConfig, chain::L1BatchCommitDataGeneratorMode, database::MerkleTreeMode,
 };
@@ -228,19 +228,31 @@ async fn run_core(
         let pool = connection_pool.clone();
         let sync_state = sync_state.clone();
         let main_node_client = Arc::new(main_node_client.clone());
-        let stop_receiver = stop_receiver.clone();
+        let mut stop_receiver = stop_receiver.clone();
         async move {
+            // We instantiate the root context here, since the consensus task is the only user of the
+            // structured concurrency framework.
+            // Note, however, that awaiting for the `stop_receiver` is related to the root context behavior,
+            // not the consensus task itself. There may have been any number of tasks running in the root context,
+            // but we only need to wait for stop signal once, and it will be propagated to all child contexts.
             let ctx = ctx::root();
-            consensus::era::run_fetcher(
-                &ctx,
-                cfg,
-                pool,
-                sync_state,
-                main_node_client,
-                action_queue_sender,
-                stop_receiver,
-            )
+            scope::run!(&ctx, |ctx, s| async move {
+                s.spawn_bg(async move {
+                    consensus::era::run_fetcher(
+                        ctx,
+                        cfg,
+                        pool,
+                        sync_state,
+                        main_node_client,
+                        action_queue_sender,
+                    )
+                    .await
+                });
+                ctx.wait(stop_receiver.wait_for(|stop| *stop)).await??;
+                Ok(())
+            })
             .await
+            .context("consensus actor")
         }
     }));
 
--- a/core/bin/external_node/src/main.rs
+++ b/core/bin/external_node/src/main.rs
@@ -231,7 +231,7 @@ async fn run_core(
         let stop_receiver = stop_receiver.clone();
         async move {
             let ctx = ctx::root();
-            consensus::run_fetcher(
+            consensus::era::run_fetcher(
                 &ctx,
                 cfg,
                 pool,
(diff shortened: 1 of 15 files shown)


core/bin/external_node/src/main.rs
```diff
@@ -210,42 +210,37 @@ async fn run_core(
...
+            let ctx = ctx::root();
+            consensus::run_fetcher(
+                &ctx,
+                cfg,
+                pool,
+                sync_state,
+                main_node_client,
+                action_queue_sender,
+                stop_receiver,
+            )
```
pompon0, line 242: as I see it, you wrap Fetcher{..}.run() with run_fetcher(...) which just obscures which argument does what. What's the purpose?

core/lib/zksync_core/src/consensus/mod.rs
```diff
@@ -19,6 +25,71 @@ mod tests;
 
 pub use config::{Config, Secrets};
 
+/// Runs the consensus in the main node mode.
+pub async fn run_main_node(
+    ctx: &ctx::Ctx,
```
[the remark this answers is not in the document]
pompon0: it is misleading that you pass both ctx AND stop_receiver.

core/lib/zksync_core/src/consensus/mod.rs
```diff
@@ -19,6 +25,71 @@ mod tests;
...
+    scope::run!(ctx, |ctx, s| async {
+        s.spawn_bg(async {
+            // Consensus is a new component.
+            // For now in case of error we just log it and allow the server
+            // to continue running.
+            if let Err(err) = cfg.run(ctx, Store(pool)).await {
+                tracing::error!(%err, "Consensus actor failed");
+            } else {
+                tracing::info!("Consensus actor stopped");
+            }
```
pompon0, line 44: this doesn't belong to the consensus module, it should be up to the user to decide whether consensus actor failure is fatal.

core/bin/external_node/src/main.rs
popzxc: This is a sort of "default entrypoint". For some time now we'll have to maintain both framework and external node. If we'll have a copy-pasted implementation, it may run out of sync. With a single implementation, change one place does the thing for both EN and the framework.

All the public APIs are still available in the module, so if anything custom is required (though I doubt it will anytime soon), user is free to do so.

core/lib/zksync_core/src/consensus/mod.rs
popzxc: Same reasoning as above: have a single implementation for the time being. User that wants consensus actor failure to be fatal may easily write their own implementation of this function.

core/lib/zksync_core/src/consensus/mod.rs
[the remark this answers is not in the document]
popzxc: But they're both used here, no?
And even if we consider to fully switch to the framework, I still imagine stop receiver to be useful in some components (like API, where we want to have graceful shutdown), so I don't really consider them to be mutually exclusive.

AnastasiiaVashchuk (approved):

core/lib/zksync_core/src/consensus/mod.rs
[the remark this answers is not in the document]
pompon0: I mean, the context should be canceled when stop_receiver is triggered and that should happen outside of a function which accepts a context.

pompon0 (approved):

core/lib/zksync_core/src/consensus/era.rs
```diff
@@ -0,0 +1,64 @@
...
+) -> anyhow::Result<()> {
+    // Consensus is a new component.
+    // For now in case of error we just log it and allow the server
+    // to continue running.
+    if let Err(err) = cfg.run(ctx, Store(pool)).await {
+        tracing::error!(%err, "Consensus actor failed");
+    } else {
+        tracing::info!("Consensus actor stopped");
+    }
+    Ok(())
```
popzxc, line 33: I think that this wrapper is still convenient, even if that simple, because we intentionally silence the error here, while maintaining a signature that returns a `Result`. If we'll change this behavior here, we'll change it everywhere. Also, for symmetry with `run_fetcher`.

core/bin/external_node/src/main.rs
```diff
@@ -210,36 +210,43 @@ async fn run_core(
...
-                    tracing::info!("Consensus actor stopped");
-                    res
+            // We instantiate the root context here, since the consensus task is the only user of the
+            // structured concurrency framework.
+            // Note, however, that awaiting for the `stop_receiver` is related to the root context behavior,
+            // not the consensus task itself. There may have been any number of tasks running in the root context,
+            // but we only need to wait for stop signal once, and it will be propagated to all child contexts.
+            let ctx = ctx::root();
+            scope::run!(&ctx, |ctx, s| async move {
+                s.spawn_bg(async move {
```
pompon0, line 240: nit: wrapping in "async move" is unnecessary here. `s.spawn_bg(run_fetcher(...))` should work

core/node/node_framework/src/implementations/layers/consensus.rs
```diff
@@ -0,0 +1,176 @@
...
+                    &root_ctx,
+                    self.config,
+                    self.pool,
+                    self.sync_state,
+                    self.main_node_client,
+                    self.action_queue_sender,
+                )
+                .await
+            });
+            ctx.wait(stop_receiver.0.wait_for(|stop| *stop)).await??;
```
pompon0, line 167: fyi, if stop_receiver gets disconnected, then `FetcherTask::run()` will return an error. Is that intended?

pompon0 (approved):

core/node/node_framework/src/implementations/layers/consensus.rs
popzxc: I think it's OK. `stop_sender` per framework rules may only be dropped after either all tasks have exited (meaning that `FetcherTask` is not affected) or the task has ignored the stop signal (not the case here).
So to me, it looks like it can only return an error if there is a bug in the framework, which sounds reasonable.

PR #1547, opened 2024-04-02
fix(cache): use factory deps cache correctly

perekopskiy:
## What ❔

Factory deps should be known only for miniblocks when it was already published

## Checklist


core/lib/state/src/postgres/mod.rs
```diff
@@ -25,11 +25,13 @@ mod metrics;
 mod tests;
 
 /// Type alias for smart contract source code cache.
-type FactoryDepsCache = LruCache<H256, Vec<u8>>;
+type FactoryDepsCache = LruCache<H256, (Vec<u8>, MiniblockNumber)>;
```
slowli, line 28: Bikeshedding: IMO, it would be more readable to create a struct for cache values:

```rust
struct TimestampedFactoryDep {
    bytecode: Vec<u8>,
    inserted_at: MiniblockNumber,
}
```

core/lib/dal/src/storage_web3_dal.rs
```diff
@@ -237,17 +237,17 @@ impl StorageWeb3Dal<'_, '_> {
...
     }
 
-    /// This method doesn't check if block with number equals to `block_number`
-    /// is present in the database. For such blocks `None` will be returned.
-    pub async fn get_factory_dep_unchecked(
+    /// This method doesn't check if block with number equals to `block_number` is present in the database.
+    pub async fn get_factory_dep_and_miniblock_unchecked(
         &mut self,
         hash: H256,
         block_number: MiniblockNumber,
```
slowli, line 244: I wonder whether it would make sense to get rid of this arg, given that we filter by the output `MiniblockNumber` in any case. Not sure what the access patterns for these calls are, but theoretically this could allow to cache some deps earlier.

core/lib/state/src/postgres/mod.rs
```diff
@@ -556,23 +558,26 @@ impl ReadStorage for PostgresStorage<'_> {
...
+                if let Some((dep, miniblock)) = value.clone() {
+                    caches.factory_deps.insert(hash, (dep, miniblock));
                 }
             };
 
             value
         });
 
+        let result =
+            value.and_then(|(dep, miniblock)| (miniblock <= self.miniblock_number).then_some(dep));
```
slowli, line 579: Bikeshedding: IMO, using `Option::filter()` here would be more readable (especially if `TimestampedFactoryDep` is defined; see the other comment).

slowli (approved):

core/lib/dal/src/storage_web3_dal.rs
perekopskiy: Yeah, we can get rid of it. It will be cleaner but in terms of performance it won't change anything because I'm almost sure that in 99.99% `inserted_at <= storage.miniblock_number` is true. I will get rid of it anyway

core/lib/state/src/postgres/mod.rs
```diff
@@ -556,23 +563,32 @@ impl ReadStorage for PostgresStorage<'_> {
...
             };
 
             value
         });
 
+        let result = value
+            .filter(|timestamped_factory_dep| {
+                timestamped_factory_dep.inserted_at <= self.miniblock_number
+            })
+            .map(|timestamped_factory_dep| timestamped_factory_dep.bytecode);
```
slowli, line 591: I'll admit that `and_then` looked better; for some reason I thought that it'd be possible to shortcut a call using `?`: 

```rust
Some(value.filter(|dep| dep.inserted_at <= self.miniblock_number)?.bytecode)
```

Still possible if you move latency observation above 🙃

core/lib/state/src/postgres/mod.rs
```diff
@@ -24,12 +24,19 @@ mod metrics;
 #[cfg(test)]
 mod tests;
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct TimestampedFactoryDep {
+    bytecode: Vec<u8>,
+    inserted_at: MiniblockNumber,
+}
 /// Type alias for smart contract source code cache.
```
slowli, line 32: Nit: Missing empty line.

slowli (approved):

RomanBrodetski (approved):

PR #1550, opened 2024-04-02
perf(merkle tree): Manage indices / filters in RocksDB

slowli:
## What ❔

- Allows including indices and Bloom filters as a part of block cache in `RocksDB`. Adds the corresponding option for the Merkle tree on ENs.
- Allows profiling RocksDB operations and enables profiling for the Merkle tree operations.

## Why ❔

- Indices and Bloom filters are a major contributor to the EN RAM usage; by default, they are entirely loaded into RAM and are never unloaded. On the mainnet, Merkle tree indices / filters are >23 GiB currently.
- Operation profiling could allow to estimate the impact of the new RocksDB options more precisely.

## Checklist


slowli:
Some general notes:

- [RocksDB allows](https://github.com/facebook/rocksdb/wiki/Perf-Context-and-IO-Stats-Context) profiling timings of read operations. This allows to tell how much time was spent on seeking to / reading / decompressing / verifying blocks, [breakdown of get operations](https://github.com/facebook/rocksdb/wiki/Perf-Context-and-IO-Stats-Context#get-break-down) etc. However, enabling this kind of profiling led to a significant performance degradation (order of 3x) on [the Merkle tree load test](https://github.com/matter-labs/zksync-era/blob/main/core/lib/merkle_tree/examples/loadtest/main.rs) run locally. So if we want to profile these timings, it probably makes sense to sample them with a configurable rate in order to not tank the tree performance.
- On the Merkle tree load test, including RocksDB indices / filters into the block cache does make a difference (if `set_pin_l0_filter_and_index_blocks_in_cache` is not called – since the RocksDB is quite small during the load test, most of its data resides on level 0). Expectedly, it negatively affects performance if the block cache capacity is small.

popzxc (approved):

PR #1551, opened 2024-04-02
feat(node-framework): Add reorg detector 

AnastasiiaVashchuk:
## What ❔

+ Implement `ReorgDetectorCheckerLayer`, which adds a `CheckerPrecondition` precondition to monitor reorgs.
+ Implement `ReorgDetectorRunnerLayer`, which adds a `RunnerUnconstrainedOneshotTask` to detect and fix reorgs(reverting blocks).
+ Implement block reverter `Resource`.

## Why ❔

These changes are a part of the system porting process to the framework.

## Checklist


core/node/node_framework/src/implementations/layers/reorg_detector.rs
```diff
@@ -0,0 +1,93 @@
...
+            Err(reorg_detector::Error::ReorgDetected(last_correct_l1_batch)) => {
+                tracing::info!("Reverting to l1 batch number {last_correct_l1_batch}");
+                self.reverter.roll_back(last_correct_l1_batch).await?;
+                tracing::info!("Revert successfully completed");
+            }
+            Err(err) => return Err(err).context("reorg_detector.check_consistency()"),
+        }
+        Ok(())
+    }
+}
```
popzxc, line 75: I don't think it would work in general case, e.g. for EN running in a distributed mode (e.g. once we will be able to run EN on multiple machines, like we do with the main node). This precondition must be on every machine, since we don't want the components to start if there was a reorg, but at the same time if reorg will be detected, _every machine_ will attempt to do a rollback.

Instead, probably we need to change it as follows:

1. Precondition would simply check if there exists a reorg in a loop. If there is, the task would just sleep for a second. If not, the loop exits. This way every machine with this precondition would simply not start until the reorg is fixed.
2. You will also have an unconstrained oneshot task that will also check whether there is a reorg, and will perform a rollback if so. This task should be added by a separate layer, given that it will only run on one of the components, unlike the reorg detector.

Does it make sense?

AnastasiiaVashchuk: >Does it make sense?

1st - yes, 2nd - not really, to be more precise...this part
*given that it will only run on one of the components, unlike the reorg detector.*
particularly. Does it mean that in the distributed mode, only a single instance of EN has to revert?

popzxc: By distributed mode I mean that, say, API runs on one machine and state keeper + tree run on another one.

When you revert blocks, you make changes in Postgres and RocksDB, so you don't want two different machines to try to do it in parallel (moreover, API machines won't have RocksDB at all), so it has to be one machine that actually performs the rollback.

popzxc (approved):

PR #1552, opened 2024-04-02
fix(eth-client): Use local FeeHistory type

perekopskiy:
## What ❔

Uses local FeeHistory type that has optional `base_fee_per_gas`.

## Why ❔

See https://github.com/alloy-rs/alloy/blob/main/crates/rpc-types/src/eth/fee.rs#L52 (reth uses alloy-rpc-types)
Tests are flaky because of this, e.g. https://github.com/matter-labs/zksync-era/actions/runs/8521600793/job/23340115243#step:14:26

## Checklist

### DIFF
--- a/core/lib/eth_client/src/clients/http/query.rs
+++ b/core/lib/eth_client/src/clients/http/query.rs
@@ -16,7 +16,7 @@ use zksync_types::web3::{
 
 use crate::{
     clients::http::{Method, COUNTERS, LATENCIES},
-    types::{Error, ExecutedTxStatus, FailureInfo, RawTokens},
+    types::{Error, ExecutedTxStatus, FailureInfo, FeeHistory, RawTokens},
     Block, ContractCall, EthInterface, RawTransactionBytes,
 };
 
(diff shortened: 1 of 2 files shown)


core/lib/eth_client/src/clients/http/query.rs
```diff
@@ -104,14 +104,19 @@ impl EthInterface for QueryClient {
...
-            history.extend(chunk);
+            let block_count = helpers::serialize(&U256::from(chunk_size));
+            let newest_block = helpers::serialize(&web3::types::BlockNumber::from(chunk_end));
+            let reward_percentiles = helpers::serialize(&Option::<()>::None);
+
+            let fee_history: FeeHistory = CallFuture::new(self.web3.transport().execute(
+                "eth_feeHistory",
+                vec![block_count, newest_block, reward_percentiles],
+            ))
+            .await?;
```
Deniallugo, line 116: can you create a method for it?

popzxc (approved):

core/lib/eth_client/src/clients/http/query.rs
perekopskiy: If in future we'll need to reuse this code then we can easily extract it, at the moment I don't see any value in extracting

PR #1555, opened 2024-04-02
fix(configs): Make genesis fields optional

Deniallugo:
## What ❔

Make genesis fields optional. 

## Why ❔

For backward compatibility with env variables, we have to keep some genesis fields optional. We will replace it with required fields, once we move to the new config system 

## Checklist


core/lib/zksync_core/src/genesis.rs
```diff
@@ -111,8 +117,9 @@ impl GenesisParams {
         // if the version doesn't exist
         let _: ProtocolVersionId = config
             .protocol_version
+            .ok_or(GenesisError::MalformedConfig("default_aa_hash"))?
```
AnastasiiaVashchuk, line 120: typo in the error message here
should be `protocol_version`

AnastasiiaVashchuk (approved):

PR #1556, opened 2024-04-03
chore: Migrate more components including call tracer to u64

StanislavBreadless:
## What ❔

Migrate more parts to u64 in preparation to the v1.5.0.

Important note:
- While the internal `gas` used by the calls in the tracer could remain u32, the top level call should continue being equal to `gasLimit` requiring it to be moved to u64
- In our DB call traces are stored as `Vec<u8>` encoded by `bincode`, in other words no data about the structure is preserved and so we'll have to decode the trace based on the protocol version

## Why ❔

In preparation for v1.5.0, to reduce the diff on the v1.5.0 PR.

## Checklist


core/lib/basic_types/src/protocol_version.rs
```diff
@@ -121,6 +121,12 @@ impl ProtocolVersionId {
     pub fn is_pre_1_4_2(&self) -> bool {
         self < &ProtocolVersionId::Version21
     }
+
+    pub fn is_pre_1_5_0(&self) -> bool {
+        // In the current codebase all the protocol versions are pre-1.5.0.
+        // This method will be updated once the v1.5.0 is added to the server
+        true
+    }
```
perekopskiy, line 129: There is already `is_pre_shared_bridge` method, it should be merged with `is_pre_1_5_0` (I don't think it should be done but we shouldn't forget about it later)

core/lib/dal/src/blocks_web3_dal.rs
```diff
@@ -458,7 +458,17 @@ impl BlocksWeb3Dal<'_, '_> {
     pub async fn get_traces_for_miniblock(
         &mut self,
         block_number: MiniblockNumber,
-    ) -> sqlx::Result<Vec<Call>> {
+    ) -> anyhow::Result<Vec<Call>> {
+        let protocol_version = self
+            .storage
+            .blocks_dal()
+            .get_miniblock_protocol_version_id(block_number)
+            .await?;
```
perekopskiy, line 466: you should add unwrap_or_else ProtocolVersionId::last_potentially_undefined

core/lib/dal/src/transactions_dal.rs
```diff
@@ -500,6 +500,13 @@ impl TransactionsDal<'_, '_> {
     ) {
         {
             let mut transaction = self.storage.start_transaction().await.unwrap();
+            let protocol_version = transaction
+                .blocks_dal()
+                .get_miniblock_protocol_version_id(miniblock_number)
+                .await
+                .unwrap()
+                .unwrap();
```
perekopskiy, line 508: same here

core/lib/dal/src/transactions_dal.rs
```diff
@@ -1293,7 +1304,27 @@ impl TransactionsDal<'_, '_> {
...
+    pub async fn get_call_trace(&mut self, tx_hash: H256) -> anyhow::Result<Option<Call>> {
+        let protocol_version: Option<ProtocolVersionId> = sqlx::query!(
+            r#"
+            SELECT
+                protocol_version
+            FROM
+                miniblocks
+                LEFT JOIN transactions ON transactions.miniblock_number = miniblocks.number
+            WHERE
+                transactions.hash = $1
```
perekopskiy, line 1316: This query is formally correct but I'm not sure how postgres will execute it. Instead I suggest to do
`FROM transactions LEFT JOIN miniblocks ON transactions.miniblock_number = miniblocks.number`, this way postgres will 100% execute the query optimally.

core/lib/dal/src/transactions_dal.rs
```diff
@@ -1293,7 +1304,27 @@ impl TransactionsDal<'_, '_> {
...
+            "#,
+            tx_hash.as_bytes()
+        )
+        .fetch_optional(self.storage.conn())
+        .await?
+        .and_then(|row| row.protocol_version.map(|v| (v as u16).try_into().unwrap()));
+
+        let Some(protocol_version) = protocol_version else {
+            return Ok(None);
+        };
```
perekopskiy, line 1326: and same nit about unwrap_or_else

core/lib/basic_types/src/protocol_version.rs
StanislavBreadless: @perekopskiy thank you flagging this. Recently @mm-zk said that he is not 100% sure that 1.5.0 will go together with shared bridge, though 99% that these will, so let's probably merge those as part of the Kalman's branch in case these upgrades are mergeed

core/lib/dal/src/blocks_web3_dal.rs
StanislavBreadless: but here the error is from DB, shouldn't I propagate it?

StanislavBreadless: though for the `if` below, sure

perekopskiy (approved):
`Cargo license check / cargo-deny ` is failing, hopefully bump of `h2` should help

PR #1565, opened 2024-04-03
chore(types): add support for 6 blobs

koloz193:
## Checklist

### DIFF
--- a/core/lib/basic_types/src/protocol_version.rs
+++ b/core/lib/basic_types/src/protocol_version.rs
@@ -87,6 +87,26 @@ impl ProtocolVersionId {
         }
     }
 
+    /// Returns the number of blobs supported by each VM version
+    pub fn into_num_blobs_required(&self) -> usize {
+        if self.is_pre_1_5_0() {
+            2
+        } else {
+            16
+        }
+    }
+
+    /// Returns the number of blobs created within each VM version
+    pub fn into_num_blobs_created(&self) -> usize {
+        if self.is_pre_1_4_2() {
+            0
+        } else if self.is_pre_1_5_0() {
+            2
+        } else {
+            6
+        }
+    }
+
     // It is possible that some external nodes do not store protocol versions for versions below 9.
     // That's why we assume that whenever a protocol version is not present, version 9 is to be used.
     pub fn last_potentially_undefined() -> Self {
@@ -120,6 +140,20 @@ impl ProtocolVersionId {
 
     pub fn is_pre_1_4_2(&self) -> bool {
         self < &ProtocolVersionId::Version21
+    }
+
+    pub fn is_1_4_2(&self) -> bool {
+        self == &ProtocolVersionId::Version21 || self == &ProtocolVersionId::Version22
+    }
+
+    pub fn is_pre_1_5_0(&self) -> bool {
+        // TODO: review this when we actually deploy 1.5.0
+        true
+    }
+
+    pub fn is_post_1_5_0(&self) -> bool {
+        // TODO: review this when we actually deploy 1.5.0
+        false
     }
 }
 
--- a/core/lib/l1_contract_interface/src/i_executor/commit/kzg/mod.rs
+++ b/core/lib/l1_contract_interface/src/i_executor/commit/kzg/mod.rs
@@ -267,13 +267,13 @@ impl KzgInfo {
     }
 }
 
-pub fn pubdata_to_blob_commitments(pubdata_input: &[u8]) -> [H256; 2] {
+pub fn pubdata_to_blob_commitments(num_blobs: usize, pubdata_input: &[u8]) -> Vec<H256> {
     assert!(
-        pubdata_input.len() <= 2 * ZK_SYNC_BYTES_PER_BLOB,
-        "Pubdata length exceeds size of 2 blobs"
+        pubdata_input.len() <= num_blobs * ZK_SYNC_BYTES_PER_BLOB,
+        "Pubdata length exceeds size of blobs"
     );
 
-    let blob_commitments = pubdata_input
+    let mut blob_commitments = pubdata_input
         .chunks(ZK_SYNC_BYTES_PER_BLOB)
         .map(|blob| {
             let kzg_info = KzgInfo::new(blob);
@@ -283,8 +283,6 @@ pub fn pubdata_to_blob_commitments(pubdata_input: &[u8]) -> [H256; 2] {
 
     // If length of `pubdata_input` is less than or equal to `ZK_SYNC_BYTES_PER_BLOB` (126976)
     // then only one blob will be used and 32 zero bytes will be used as a commitment for the second blob.
-    [
-        blob_commitments.get(0).copied().unwrap_or_default(),
-        blob_commitments.get(1).copied().unwrap_or_default(),
-    ]
+    blob_commitments.resize(num_blobs, H256::zero());
+    blob_commitments
 }
--- a/core/lib/types/src/commitment/mod.rs
+++ b/core/lib/types/src/commitment/mod.rs
@@ -12,15 +12,17 @@ use serde::{Deserialize, Serialize};
 use zksync_contracts::BaseSystemContractsHashes;
 use zksync_mini_merkle_tree::MiniMerkleTree;
 use zksync_system_constants::{
-    BLOB1_LINEAR_HASH_KEY, BLOB2_LINEAR_HASH_KEY, KNOWN_CODES_STORAGE_ADDRESS,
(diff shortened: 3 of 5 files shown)


core/lib/types/src/l2_to_l1_log.rs
```diff
@@ -61,14 +62,46 @@ pub fn l2_to_l1_logs_tree_size(protocol_version: ProtocolVersionId) -> usize {
...
+            log.0.sender == PUBDATA_CHUNK_PUBLISHER_ADDRESS
+                && log.0.key >= H256::from_low_u64_be(BLOB1_LINEAR_HASH_KEY as u64)
+                && log.0.key
+                    < H256::from_low_u64_be((BLOB1_LINEAR_HASH_KEY + num_created_blobs) as u64)
+        })
+        .map(|log| (log.0.key, log.0.value))
+        .collect::<Vec<(H256, H256)>>();
+
+    blob_hashes.sort_unstable_by_key(|(k, _)| *k);
+    let mut blob_hashes = blob_hashes.iter().map(|(_, v)| *v).collect::<Vec<H256>>();
```
StanislavBreadless, line 102: please add an assert that the size is exactly `num_created_blobs`, just in case

StanislavBreadless (approved):
👍 , I left one small nit

core/lib/basic_types/src/protocol_version.rs
```diff
@@ -87,6 +87,26 @@ impl ProtocolVersionId {
...
+    pub fn into_num_blobs_required(&self) -> usize {
+        if self.is_pre_1_5_0() {
+            2
+        } else {
+            16
+        }
+    }
+
+    /// Returns the number of blobs created within each VM version
+    pub fn into_num_blobs_created(&self) -> usize {
```
Deniallugo, line 100: It's not into. 
And I'd prefer to have 

an additional function somewhere close to blobs creation, where are you based on version return the necessary value. 

Protocol version know nothing about blobs

core/lib/l1_contract_interface/src/i_executor/commit/kzg/mod.rs
```diff
@@ -283,8 +283,6 @@ pub fn pubdata_to_blob_commitments(pubdata_input: &[u8]) -> [H256; 2] {
 
     // If length of `pubdata_input` is less than or equal to `ZK_SYNC_BYTES_PER_BLOB` (126976)
     // then only one blob will be used and 32 zero bytes will be used as a commitment for the second blob.
```
perekopskiy, line 285: Please update comment

core/lib/types/src/commitment/mod.rs
```diff
@@ -606,7 +593,9 @@ impl L1BatchCommitment {
         let mut result = vec![];
         let pass_through_data_hash = self.pass_through_data.hash();
         result.extend_from_slice(pass_through_data_hash.as_bytes());
-        let metadata_hash = self.meta_parameters.hash();
+        let metadata_hash = self
+            .meta_parameters
+            .hash(&self.auxiliary_output.common().protocol_version);
```
perekopskiy, line 598: `meta_parameters` and `auxiliary_output` should be independent. I suggest adding `protocol_version` field to `L1BatchMetaParameters`

core/lib/types/src/l2_to_l1_log.rs
```diff
@@ -61,14 +66,46 @@ pub fn l2_to_l1_logs_tree_size(protocol_version: ProtocolVersionId) -> usize {
...
+pub fn parse_system_logs_for_blob_hashes(
+    protocol_version: &ProtocolVersionId,
+    system_logs: &[SystemL2ToL1Log],
+) -> Vec<H256> {
+    let num_required_blobs = num_blobs_required(protocol_version) as u32;
+    let num_created_blobs = num_blobs_created(protocol_version) as u32;
+
+    if num_created_blobs == 0 {
+        return vec![H256::zero(); num_required_blobs as usize];
     }
```
perekopskiy, line 92: I think this `if` not required

core/lib/types/src/blob.rs
```diff
@@ -0,0 +1,21 @@
...
+pub fn num_blobs_required(protocol_version: &ProtocolVersionId) -> usize {
+    if protocol_version.is_pre_1_5_0() {
+        2
+    } else {
+        16
+    }
+}
+
+/// Returns the number of blobs created within each VM version
+pub fn num_blobs_created(protocol_version: &ProtocolVersionId) -> usize {
```
perekopskiy, line 13: I think we can only have `num_blobs_created`. I also would rename method to `num_blobs_supported` or `max_num_blobs` and add a comment that this number represents max number of blobs supported by protocol for given protocol version. Add mention that this number also specifies number of system logs with blob linear hashes

core/lib/types/src/l2_to_l1_log.rs
koloz193: we need it because if we're in a protocol version that doesnt generate the linear hash logs, we can mistakenly use the expected system upgrade txn hash

core/lib/types/src/blob.rs
koloz193: we have different levels of blobs that are supported by the circuits (reflected in the batch commitment) and in the protocol

koloz193:
> We should have green upgrade test to be able to merge this, please have a look at the workaround we used to have for this test previous time: [9b773eb#diff-0bc88fd65c4d16fec636c414476a9ccf68a3974e74b3c1d039fc9a7a135b2204R337](https://github.com/matter-labs/zksync-era/commit/9b773ebb3c170dc3d503ec83fb33d8d463614b4f#diff-0bc88fd65c4d16fec636c414476a9ccf68a3974e74b3c1d039fc9a7a135b2204R337), I think you can reuse it.

@perekopskiy but should this matter if we haven't introduced a new protocol/vm version for 1.5.0?

core/lib/types/src/l2_to_l1_log.rs
perekopskiy: Hmm, I mean we won't use system upgrade log because this
```
log.0.key >= H256::from_low_u64_be(BLOB1_LINEAR_HASH_KEY as u64)
                && log.0.key
                    < H256::from_low_u64_be((BLOB1_LINEAR_HASH_KEY + num_created_blobs) as u64)
```
will never be true when `num_created_blobs == 0`. But I don't insist on removing the return, agree that in theory if we change method impl and it doesn't work for `num_created_blobs == 0` then the if will save us.

core/lib/types/src/blob.rs
perekopskiy: but if I'm not mistaken numbers are different only for 1.4.1 and it's explicitly covered with
```
if common.protocol_version.is_1_4_1() {
    // We are using zeroes as commitments to the KZG pubdata as per convention.
    result.extend(H256::zero().as_bytes());
    result.extend(H256::zero().as_bytes());
    result.extend(H256::zero().as_bytes());
    result.extend(H256::zero().as_bytes());
}
```

core/lib/types/src/commitment/tests/post_boojum_1_4_2_test.json
```diff
@@ -485,7 +485,8 @@
   "meta_parameters": {
     "zkporter_is_available": false,
     "bootloader_code_hash": "0x010007ed0e328b940e241f7666a6303b7ffd4e3fd7e8c154d6e7556befe6cd6d",
-    "default_aa_code_hash": "0x0100055b7a8be90522251be8be1a186464d056462973502ac8a0437c85e4d2a9"
+    "default_aa_code_hash": "0x0100055b7a8be90522251be8be1a186464d056462973502ac8a0437c85e4d2a9",
+    "protocol_version": "Version21"
```
perekopskiy, line 489: It would be good to have test for 1.5.0 commitment. You can write a test and change impl of `is_pre_1_5_0`, `is_post_1_5_0` for `#[cfg(test)]`.

core/lib/types/src/blob.rs
perekopskiy: Oh, sorry, I see, it's 6 vs 16 for 1.5.0. Then yeah, let's keep two functions, let's also remove the if in my comment above. And also, `num_blobs_required` should return 0 for pre-1.4.1 versions

core/lib/types/src/commitment/tests/post_boojum_1_4_2_test.json
StanislavBreadless: It was harder to do than anticipated. I talked to @perekopskiy in private, it was decided to include the test into the main 1-5-0-support PR

StanislavBreadless (dismissed):

perekopskiy (dismissed):

perekopskiy (approved):

PR #1566, opened 2024-04-03
feat(config): Initialize log config from files as well

Deniallugo:
## What ❔

Initialize  logs and backtrace for files 

## Why ❔
In a way to move all configs to files it's one the last parts

## Checklist


core/lib/vlog/src/lib.rs
```diff
@@ -301,6 +312,8 @@ impl ObservabilityBuilder {
             None
         };
 
+        // We use backtrace capture in `json_panic_handler` so we have to have it always enabled
+        std::env::set_var("RUST_BACKTRACE", "1");
```
itegulov, line 316: We only use `json_panic_handler` for JSON log format, no?

itegulov: Also, this overrides `RUST_BACKTRACE=full` which is probably an undesired side-effect

core/bin/external_node/src/config/observability.rs
```diff
@@ -30,12 +30,14 @@ pub fn observability_config_from_env() -> anyhow::Result<ObservabilityConfig> {
     } else {
         "plain".to_string()
     };
+    let log_directives = std::env::var("RUST_LOG").ok();
```
itegulov, line 33: Does this mean we are okay with leaving observability stuff to be configured based on env vars?

core/lib/vlog/src/lib.rs
Deniallugo: everything non zero is the same 
https://doc.rust-lang.org/std/backtrace/#environment-variables.

We use json logs as well

core/bin/external_node/src/config/observability.rs
RomanBrodetski: I also don't understand entirelly.. Do we have it somewhere written? Or can you explain how env and file initialization go along From this code it looks like we read this `log_directives` _only_ from env

core/lib/vlog/src/lib.rs
```diff
@@ -301,6 +318,10 @@ impl ObservabilityBuilder {
             None
         };
 
+        if let Some(rust_backtrace) = self.rust_backtrace {
+            std::env::set_var("RUST_BACKTRACE", rust_backtrace);
```
popzxc, line 322: Does this work or rather - does this _always_ work?

Quoting [the same doc you've linked](https://doc.rust-lang.org/std/backtrace/#environment-variables):

>Also note that the state of environment variables is cached once the first backtrace is created, so altering RUST_LIB_BACKTRACE or RUST_BACKTRACE at runtime might not actually change how backtraces are captured.

Deniallugo: hm.. I'll have a look

core/bin/external_node/src/config/observability.rs
Deniallugo: This method names: 
`observability_from_env` and it's used in external_node because we rename envs data there.

core/lib/vlog/src/lib.rs
Deniallugo: Based on documentation it's good enough. 
We may lose backtrace during the initialization.
But if initialization was successful we will catch the backtrace

core/lib/vlog/src/lib.rs
itegulov: Ah, apparently `full` is not a thing anymore indeed

> We use json logs as well

Yeah I just meant that `json_panic_handler` is not called 100% of the time we run our code, although I assume all real environments use JSON logs

core/lib/vlog/src/lib.rs
itegulov: > The documentation states that the value that is read at the first backtrace is cached, which means that overriding it afterwise doesn't work, no?

My assumption is that it works like this:
* We don't generate any backtraces during initialization => everything is fine, `RUST_BACKTRACE` is set before it was cached
* We generate a backtrace which is a fatal event during initialization as we don't have "try-catch" patterns there => `RUST_BACKTRACE` is not set, but we are panicking anyway

Still, this feels hacky and unnecessary given that we can just ensure to pass the correct `RUST_BACKTRACE` from upstream. We could maybe add a sanity check with a warning that `RUST_BACKTRACE` is not set and hence `json_panic_handler` is not going to work

Deniallugo: I have the same feeling as @itegulov 

But in the world, where we don't have `RUST_BACKTRACE` env, I see no other possibilities

Deniallugo: I had a look at the code of backtrace and yes. 
During `Backtrace::capture` they call method `enabled` that caches this value. 
So, everything should be fine if we set it a bit later and don't capture initialization backtraces

popzxc: If the only concern here is that `json_panic_handler` would work incorrectly, then one can always use [force_capture](https://doc.rust-lang.org/std/backtrace/struct.Backtrace.html#method.force_capture) there, no?

Deniallugo: But should we be able to remove this functionality?

popzxc: What do you mean? Which functionality?
I suggest using `force_capture` in json panic handler, and otherwise do not affect `RUST_BACKTRACE` in any way from the code (e.g. if it's set it's set).

Deniallugo: So, you suggest to remove possibility to install `RUST_BACKTRACE` am i right?

popzxc: ...from the codebase. Basically, do nothing about this env variable in our code.

Deniallugo: changed to force

core/bin/external_node/src/config/observability.rs
popzxc: To elaborate:
- One can construct `ObservabilityConfig` and use it with `vlog` in a way that they prefer, be it env variables or files or just hardcoding.
- This method is dedicated to fetch configs from `env` so we get this value from env in a way that we did before (e.g. backward compatibility).
- `vlog` still would fallback to env configuration if nothing wat provided, which makes sense given that `RUST_LOG` is an industry default.

popzxc (approved):

PR #1579, opened 2024-04-04
feat(block-reverter): only require private key for sending revert transactions

montekki:
## What ❔

Previosly whenever working with `block_reverter` you needed to specify the private key that is eneded to send the revert transaction. However this key is only required for one operation and makes no sense while the user just reverts a database or a rocksdb. This patch
* Makes operator private key required only for sending revert transactions
* Makes operator _address_ required to print suggested values

Thus limiting the usage of private key to a minimum.

## Checklist


Deniallugo (approved):

core/tests/revert-test/tests/revert-and-restart.test.ts
```diff
@@ -150,7 +150,7 @@ describe('Block reverting test', function () {
     step('revert blocks', async () => {
         const executedProcess = await utils.exec(
             'cd $ZKSYNC_HOME && ' +
-                'RUST_LOG=off cargo run --bin block_reverter --release -- print-suggested-values --json'
+                'RUST_LOG=off cargo run --bin block_reverter --release -- print-suggested-values --json --operator-address 0xde03a0B5963f75f1C8485B355fF6D30f3093BDE7'
```
Deniallugo, line 153: Probably it's better to get it from config

montekki: Does the address, not the key exist there ?

Deniallugo: for tests everything exists

montekki: done, takes it from env now

Deniallugo (dismissed):

Deniallugo (approved):

PR #1584, opened 2024-04-05
feat: Expose component configs as info metrics

slowli:
## What ❔

- Exposes configs of various components as info metrics [defined](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#info) in the Open Metrics standard.
- Reworks mempool cache to be shared between API servers.

## Why ❔

- More idiomatic and observable than printing these configs to logs (or not observing at all).
- Allows to use configs in Grafana dashboards, alerts etc.
- The mempool cache makes sense to share logically, and if multiple instances are used, reported metrics can be bogus.

## Checklist


core/bin/external_node/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "zksync_external_node"
-version = "0.1.0"
+version = "22.1.0" # x-release-please-version
```
slowli, line 3: Checked this annotation by running `release-please` locally [as per the troubleshooting guide](https://github.com/googleapis/release-please/blob/main/docs/troubleshooting.md); seems to work.

Cargo.toml
```diff
@@ -218,3 +219,7 @@ zksync_crypto_primitives = { path = "core/lib/crypto_primitives" }
 zksync_node_framework = { path = "core/node/node_framework" }
 zksync_eth_watch = { path = "core/node/eth_watch" }
 zksync_shared_metrics = { path = "core/node/shared_metrics" }
+
+# FIXME: remove after consensus crates are updated
+[replace]
```
slowli, line 224: The [`patch` section](https://doc.rust-lang.org/stable/cargo/reference/overriding-dependencies.html#the-patch-section) is a recommended way to do such overrides, but it doesn't seem to work in this case because it disallows the override to point at the same git repo as the original dep (not sure why).

core/bin/external_node/src/main.rs
```diff
@@ -539,36 +527,9 @@ async fn init_tasks(
...
-        while !*stop_receiver_for_task.borrow_and_update() {
-            let protocol_version = pool
-                .connection()
-                .await?
-                .protocol_versions_dal()
-                .last_used_version_id()
-                .await
-                .map(|version| version as u16);
-
-            EN_METRICS.version[&(version.to_string(), protocol_version)].set(1);
```
slowli, line 562: Were the old metrics used anywhere? Probably obvious, but this PR has changed their naming and split off protocol version into a gauge (since it can change during the program runtime).

core/bin/external_node/Cargo.toml
Deniallugo: We could configure release please to update all crates, but ill leave this decision to @popzxc

slowli: Would the versions of library crates necessarily correspond to EN / server versions? In general, release-please supports Cargo workspaces, but it looks somewhat convoluted judging by docs (and there may be other reasons why it's not used already 🙃).

perekopskiy (approved):

perekopskiy (approved):

PR #1589, opened 2024-04-05
chore: Remove initial_writes and repeated_writes from tree metadata

perekopskiy:
## What ❔

Removes initial_writes and repeated_writes from tree metadata. Witness object is used in tests to perform needed checks.

## Why ❔

initial_writes and repeated_writes weren't used except for testing and consistency check purposes

## Checklist

### DIFF
--- a/core/lib/merkle_tree/src/domain.rs
+++ b/core/lib/merkle_tree/src/domain.rs
@@ -3,10 +3,7 @@
 use rayon::{ThreadPool, ThreadPoolBuilder};
 use zksync_crypto::hasher::blake2::Blake2Hasher;
 use zksync_prover_interface::inputs::{PrepareBasicCircuitsJob, StorageLogMetadata};
-use zksync_types::{
-    writes::{InitialStorageWrite, RepeatedStorageWrite},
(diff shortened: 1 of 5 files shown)


itegulov (commented):
The changes look good to me and I love deleting code, but maybe a dumb question - is it not valuable to have a consistency check here? Is there a reason why we decided to delete it?

perekopskiy:
I'd say it's somewhat valuable but not required. It was only a piece of data that we checked for consistency, and e.g. repeated writes weren't covered. I think we shouldn't keep `initial_writes` only for the sake of this check

itegulov (approved):

slowli (approved):

slowli:
> is it not valuable to have a consistency check here

@itegulov As long as the Merkle tree is not responsible for enumeration index assignment, this check looks somewhat redundant, IMO. (When the check was initially implemented, the tree assigned enum indices independently of the state keeper.) I guess we could indicate in `TreeInstruction`s whether a write should be initial, so that the tree returns an error if it's mismatched with the tree state; IMO, it wouldn't have that much value in practice.

PR #1592, opened 2024-04-05
feat(genesis): add genesis config generator

Deniallugo:
## What ❔

Add binary that updates the genesis file. It's easier now to work during the new version generation

## Checklist


core/bin/genesis_generator/src/main.rs
```diff
@@ -0,0 +1,113 @@
+/// Each protocol upgrade required to update genesis config values.
+/// This tool generates the new correct genesis file that could be used for the new chain
+/// Please note, this tool update only yaml file, if you are still use env based configuration,
```
perekopskiy, line 3: nit
```suggestion
/// Please note, this tool update only yaml file, if you still use env based configuration,
```

core/bin/genesis_generator/src/main.rs
```diff
@@ -0,0 +1,113 @@
...
+use zksync_dal::{ConnectionPool, Core, CoreDal};
+use zksync_env_config::FromEnv;
+use zksync_protobuf::{
+    build::{prost_reflect, prost_reflect::ReflectMessage},
+    ProtoRepr,
+};
+use zksync_protobuf_config::proto::genesis::Genesis;
+use zksync_types::ProtocolVersionId;
+
+const DEFAULT_GENESIS_FILE_PATH: &'static str = "./etc/env/file_based//genesis.yaml";
```
perekopskiy, line 25: is double / intended?

core/bin/genesis_generator/src/main.rs
```diff
@@ -0,0 +1,113 @@
...
+
+#[derive(Debug, Parser)]
+#[command(author = "Matter Labs", version, about = "zkSync operator node", long_about = None)]
```
perekopskiy, line 28: please change about

core/bin/genesis_generator/src/main.rs
```diff
@@ -0,0 +1,113 @@
...
+        genesis_root_hash: None,
+        rollup_last_leaf_index: None,
+        genesis_commitment: None,
+        bootloader_hash: Some(base_system_contracts.bootloader),
+        default_aa_hash: Some(base_system_contracts.default_aa),
+        ..genesis_config
+    };
+
+    let params = GenesisParams::load_genesis_params(updated_genesis.clone())?;
+    let batch_params = insert_genesis_batch(&mut transaction, &params).await?;
```
perekopskiy, line 96: IMO, this tool shouldn't insert genesis batch, it should only generate config. I find it more handy, imagine you only want to regenerate config and check it but not modify DB. Anyway, if you disagree, you must at least mention somewhere in the description that this tool inserts genesis batch into DB.

Deniallugo: Yes, that's why i've not commit the tx. 
I wanted to reuse as much code as possible

Deniallugo: Added a comment about it

etc/env/file_based/genesis.yaml
```diff
@@ -1,19 +1,20 @@
...
+l1_chain_id: '9'
+l2_chain_id: '270'
+fee_account: '0x0000000000000000000000000000000000000001'
 prover:
-  recursion_leaf_level_vk_hash: 0x400a4b532c6f072c00d1806ef299300d4c104f4ac55bd8698ade78894fcadc0a
-  recursion_node_level_vk_hash: 0x5a3ef282b21e12fe1f4438e5bb158fc5060b160559c5158c6389d62d9fe3d080
   recursion_scheduler_level_vk_hash: 0x063c6fb5c70404c2867f413a8e35563ad3d040b1ad8c11786231bfdba7b472c7
+  recursion_node_level_vk_hash: 0x5a3ef282b21e12fe1f4438e5bb158fc5060b160559c5158c6389d62d9fe3d080
+  recursion_leaf_level_vk_hash: 0x400a4b532c6f072c00d1806ef299300d4c104f4ac55bd8698ade78894fcadc0a
+  recursion_circuits_set_vks_hash: '0x0000000000000000000000000000000000000000000000000000000000000000'
```
perekopskiy, line 14: Why some fields are in '' and some are not?

Deniallugo: so... it's a good question. 
It is how automatical tools work, if it could fit into u64 it must be in quotes, but if it's not it could be non-in quotes.

perekopskiy (approved):

PR #1594, opened 2024-04-05
refactor: moved consensus config to config crate

pompon0:
moved consensus config to config crate


core/lib/protobuf_config/src/proto/core/consensus.proto
Deniallugo, line 1: why is it the only in core?

Deniallugo: and should we add it to general?

pompon0: It is in core, because moving messages between files is incompatible wrt "FILE" rule set: https://buf.build/docs/breaking/rules .

pompon0: So I moved it between crates, without changing the proto file name, no problem.

pompon0: yes, I've done the moving of configs between crates that you asked for. You can add it to general now.

Deniallugo (dismissed):

Deniallugo (approved):

PR #1596, opened 2024-04-08
feat(eth-watch): Brush up Ethereum watcher component

slowli:
## What ❔

Brushes up the Ethereum Watcher component that was recently moved to its own crate. Wraps `sqlx` errors in the relevant DB queries.

## Why ❔

Simplifies maintenance.

## Checklist


core/bin/block_reverter/src/main.rs
```diff
@@ -86,7 +86,7 @@ async fn main() -> anyhow::Result<()> {
     }
     let _guard = builder.build();
 
-    let eth_sender = ETHConfig::from_env().context("ETHSenderConfig::from_env()")?;
+    let eth_sender = EthConfig::from_env().context("ETHSenderConfig::from_env()")?;
```
AnastasiiaVashchuk, line 89: context message seems to be outdated(there is no `ETHSenderConfig`)

core/lib/dal/src/protocol_versions_dal.rs
```diff
@@ -366,27 +383,34 @@ impl ProtocolVersionsDal<'_, '_> {
...
+        let maybe_row = instrumentation
+            .with(query)
+            .fetch_optional(self.storage)
+            .await?;
+        let Some(upgrade_tx_hash) = maybe_row.and_then(|row| row.upgrade_tx_hash) else {
+            return Ok(None);
+        };
+        let upgrade_tx_hash = H256::from_slice(&upgrade_tx_hash);
+
+        let instrumentation = Instrumented::new("get_protocol_upgrade_tx#get_tx")
```
AnastasiiaVashchuk, line 397: why `#get_tx`?

core/node/eth_watch/src/client.rs
```diff
@@ -125,13 +99,13 @@ impl EthClient for EthHttpQueryClient {
         from: BlockNumber,
         to: BlockNumber,
         retries_left: usize,
-    ) -> Result<Vec<Log>, Error> {
+    ) -> Result<Vec<Log>, EthClientError> {
         let latency = METRICS.get_priority_op_events.start();
```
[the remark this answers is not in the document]
perekopskiy: Agree, can be removed

core/node/eth_watch/src/client.rs
```diff
@@ -173,14 +147,9 @@ impl EthClient for EthHttpQueryClient {
 
                 // safety check to prevent infinite recursion (quite unlikely)
                 if from_number >= mid {
-                    return Err(Error::InfiniteRecursion);
+                    return result;
```
perekopskiy, line 150: If this code gets broken in future, won't having an explicit error help us to find it quickly? Even if so, I'm not sure if we should keep it.

core/node/eth_watch/src/event_processors/upgrades.rs
```diff
@@ -35,14 +34,13 @@ impl EventProcessor for UpgradesEventProcessor {
...
         events: Vec<Log>,
-    ) -> Result<(), Error> {
+    ) -> Result<(), EventProcessorError> {
         let mut upgrades = Vec::new();
-        for event in events
-            .into_iter()
-            .filter(|event| event.topics[0] == UPGRADE_PROPOSAL_SIGNATURE)
-        {
+        for event in events {
+            assert_eq!(event.topics[0], UPGRADE_PROPOSAL_SIGNATURE); // guaranteed by the watcher
```
perekopskiy, line 40: let's remove `UpgradesEventProcessor`, it was needed prior to some protocol version, now it's fully replaced by `GovernanceUpgradesEventProcessor`

core/node/eth_watch/README.md
```diff
@@ -2,3 +2,16 @@
...
+Internally, Eth Watcher contains _event processors_, each of which provides a relevant topic (i.e., a `bytes32` Solidity
+event selector) and is responsible for processing the corresponding events. Besides events, processors have access to
+the L1 client (to query more info) and to the node Postgres (to persist processing results). Examples of processors are:
+
+- [Priority operations processor](src/event_processors/priority_ops.rs): persists priority operations (aka L1
+  transactions)
+- [Upgrades processor](src/event_processors/upgrades.rs): persists protocol upgrades.
+
+Eth Watcher combines topics from the processors into a single filter and periodically queries L1 for the corresponding
+events. The fetched events are partitioned per processor and fed to them in succession.
```
AnastasiiaVashchuk, line 17: Grammarly says we need articles
```
This crate contains insert "an "an implementation of insert "the "the zkSync Era Eth Watcher component, insert "which is "which is responsible for fetching the changes from the corresponding L1 contract.
```

core/node/eth_watch/README.md
```diff
@@ -2,3 +2,16 @@
 
 This crate contains implementation of zkSync Era Eth Watcher component, responsible for fetching the changes from the
 corresponding L1 contract.
```
AnastasiiaVashchuk, line 4: nit:
I know you did not touch this, but  if you feel comfortable, you may change this so it looks grammarly better
```
This crate contains an implementation of the zkSync Era Eth Watcher component, which fetches the changes from the corresponding L1 contract.
```

AnastasiiaVashchuk (dismissed):

core/node/eth_watch/src/event_processors/upgrades.rs
slowli: Just to confirm: We'll never need to process historic events for the old protocol versions, right? This functionality is not used by ENs, and I think we'll never need to re-process protocol upgrades on the main node.

perekopskiy: Yes

core/node/eth_watch/src/client.rs
slowli: I've added a `WARN` log before returning the result. WDYT, is this a good way to deal with the issue?

slowli:
@pompon0 Do I understand correctly that `buf breaking` doesn't support renaming message types (i.e., `ETH` -> `Eth` and `ETHWatch` -> `EthWatch` in this case)? Seems like that judging [by docs](https://buf.build/docs/breaking/rules#field_wire_json_compatible_type); they say nothing about renaming message types 🙃

pompon0:
@slowli, in protobuf syntax renaming a type is indistinguishable from removing+adding a type. Changing between compatible types is allowed to some extent https://github.com/bufbuild/buf/pull/400 , but apparently message rename is sth that is not expected to be ever needed - message names are not used for encoding, they are treated just as globally unique identifiers of the types used for autogenerated code. Autogenerated code is immediately converted to our custom rust types (which can change names), so it doesn't affect the codebase at large. Having said that, I can imagine a need for changing field message type to another compatible type, so if the current situation significantly affects the UX, we may consider reimplementing protobuf compatibility checks (it is not that hard).

perekopskiy (approved):

core/node/eth_watch/src/client.rs
perekopskiy: warning should be enough, thanks

PR #1602, opened 2024-04-08
chore: weakened protobuf compatibility check

pompon0:
Weakened protobuf compatibility check from FILE (backward compatibility of generated code) to WIRE_JSON (backward/forward compatibility of generated JSON and binary messages). This will allow us for removing fields from the schema by [reserving ](https://protobuf.dev/programming-guides/proto3/#fieldreserved) the removed field and number.

### DIFF
--- a/.github/workflows/protobuf.yaml
+++ b/.github/workflows/protobuf.yaml
@@ -54,5 +54,7 @@ jobs:
           `find ./after/target/debug/build/*/output`
           | xargs cat > ./after.binpb
       - uses: bufbuild/buf-setup-action@v1
+        with:
+          github_token: ${{ github.token }}
       - name: buf breaking
         run: buf breaking './after.binpb' --against './before.binpb' --config '{"version":"v1","breaking":{"use":["WIRE_JSON"]}}' --error-format 'github-actions'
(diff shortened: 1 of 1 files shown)


perekopskiy (approved):

.github/workflows/protobuf.yaml
```diff
@@ -54,7 +54,5 @@ jobs:
           `find ./after/target/debug/build/*/output`
           | xargs cat > ./after.binpb
       - uses: bufbuild/buf-setup-action@v1
```
slowli, line 56: Off-topic: Would setting `github_token: ${{ github.token }}` work for this action in our setup? Seems that the check [fails with rate-limiting errors](https://github.com/matter-labs/zksync-era/actions/runs/8596293564/job/23552565599?pr=1584#step:9:20) sometimes.

slowli (dismissed):

.github/workflows/protobuf.yaml
pompon0: ack, added

perekopskiy (approved):

artmakh (approved):

PR #1608, opened 2024-04-08
fix: instruction count diff always N/A in VM perf comparison

joonazan:
Instruction count diff should now be correct. The problem wasn't found initially because counting instructions legitimately fails if the PR starts from a version where that wasn't supported yet.

Also found and fixed a bug that hid performance improvements.


joonazan:
Now it works based on the logs. It still shows N/A here but I know that I fixed the reason why it did that incorrectly earlier.

perekopskiy (approved):

artmakh (approved):

PR #1612, opened 2024-04-08
feat: EVM Equivalence in Yul

jrchatruc:
## What ❔

This PR adds an EVM Yul Interpreter on top of the already existing Solidity EVM Interpreter effort, along with units tests for it.

Because there is A LOT of duplicated code between the initCode and the runtime code of the contract, we split it up into different files and used a [preprocessor tool](https://www.npmjs.com/package/preprocess-cli-tool) to build the final interpreter file.

To preprocess and then compile the interpreter, first install the preprocess tool with

```
npm install -g preprocess-cli-tool
```

then run

```
./recompile_interpreter.sh
```

on the root of this repo. The resulting binary file will be on

```
contracts/system-contracts/contracts-preprocessed/artifacts/EvmInterpreterPreprocessed.yul.zbin
```

## Checklist


hex_to_binary.py
```diff
@@ -0,0 +1,14 @@
+file_path = "contracts/system-contracts/contracts-preprocessed/artifacts/EvmInterpreterPreprocessed.yul.zbin"
```
popzxc, line 1: I don't know if the PR is ready for review, but will mention just in case that we don't use python or bash for anything meaningful in the monorepo (and try to avoid them in general). In the monorepo we stick to either Rust or TypeScript.

PR #1614, opened 2024-04-08
feat(en): metric scraping and grafana for docker compose image

tomg10:
I've added prometheus and grafana do EN docker compose with a simple dashboard like this:
image


tomg10:
I didn't add snapshots applier metrics as it turned out prometheus server is only started after the recovery finishes. I will create a separate PR fixing that and adding additional panels if I have enough time before I'm OOO

slowli (approved):

PR #1617, opened 2024-04-09
feat: Finalize fee address migration

perekopskiy:
## What ❔

Removes code for fee address migration. Drops `l1_batches.fee_account_address`

## Why ❔

Finalize fee address migration, remove code that is not needed

## Checklist


core/lib/dal/migrations/20240409102923_drop_l1_batches_fee_account_address.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE l1_batches DROP COLUMN IF EXISTS fee_account_address;
```
slowli, line 1: There [are a couple of other columns](https://github.com/matter-labs/zksync-era/pull/688#discussion_r1426827324) in the same table that aren't used. Do we want to remove them as well, or should it be tackled separately?

slowli (approved):
Off-topic: Does this mean that the `storage` table can be hard-removed as well (in a separate PR, of course), or is it not time for that yet?

core/lib/dal/migrations/20240409102923_drop_l1_batches_fee_account_address.up.sql
perekopskiy: I'll do it separately today

PR #1619, opened 2024-04-09
fix(en): improved tree recovery logs

tomg10:


core/lib/zksync_core/src/metadata_calculator/recovery/mod.rs
```diff
@@ -261,9 +263,10 @@ impl AsyncTreeRecovery {
             snapshot.expected_root_hash
         );
         let tree = tree.finalize().await;
-        let finalize_latency = finalize_latency.observe();
+        finalize_latency.observe();
         tracing::info!(
-            "Finished tree recovery in {finalize_latency:?}; resuming normal tree operation"
+            "Tree recovery has finished, the recovery took {:?}! resuming normal tree operation",
```
slowli, line 268: Nit: I'd still include finalization latency in the log message (obviously, naming it in a more transparent way, e.g. "finalization took {:?}").

slowli (approved):

PR #1620, opened 2024-04-09
fix(kl-factory): first init failing

SantiagoPittella:
## What ❔

Alter the order of the commands in the init script.

## Why ❔

The first init ran in the repo was failing because the contracts were being compiled before downloading the dependencies.

## Checklist


SantiagoPittella:
@kelemeno can you take a look a this please? When you do a fresh clone of the repo the first init is failing because of the order of the steps.

PR #1627, opened 2024-04-10
feat(en): Add boxed L2 client and use it in DI

slowli:
## What ❔

- Adds boxed L2 client implementation and uses it in DI instead of a specific HTTP-based client.
- Adds a mock L2 client.
- Uses a mock L1 and L2 clients to test EN lifecycle in Rust, namely that it can start and timely terminates when sent a SIGINT.

## Why ❔

- L2 client abstraction is better for the node framework.
- We have a frequent issue of newly added managed tasks being non-responsive, which leads to unnecessary delays during EN shutdown (30s in the worst case).
- Using TypeScript integration tests for this purpose is slower and provides less control.

## Checklist

test(en): Add basic rust tests for EN


core/lib/zksync_core/src/consensus/era.rs
```diff
@@ -37,7 +36,7 @@ pub async fn run_fetcher(
     cfg: Option<(ConsensusConfig, ConsensusSecrets)>,
     pool: ConnectionPool<Core>,
     sync_state: SyncState,
-    main_node_client: Arc<dyn MainNodeClient>,
+    main_node_client: BoxedL2Client,
```
montekki, line 39: It seems that in a lot of places this code goes from `Arc` to `Box` thus signalling that it takes the ownership of the client. My question is if the bound to dynamic dispatch and boxing is even necessary in this case or the code can be generalized even further by
1. Making it generic over a type bound by the client trait
2. Folding a lot of manual code for this trait by using [`auto_impl`](https://docs.rs/auto_impl/latest/auto_impl/)

slowli: I should've probably added this as a comment somewhere, but generalizing JSON-RPC clients is tricky:

- The core trait is [`ClientT`](https://docs.rs/jsonrpsee-core/0.22.4/jsonrpsee_core/client/trait.ClientT.html) + [`SubscriptionClientT`](https://docs.rs/jsonrpsee-core/0.22.4/jsonrpsee_core/client/trait.SubscriptionClientT.html) for subscriptions. Both these traits are not object-safe.
- These traits are usually not used directly, but via extension traits (e.g., ones in our `zksync_web3_decl` crate). These traits are generally object-safe, but because they define a lot of methods, mocking them directly or implementing cross-cutting functionality on these traits directly looks unnecessarily painful.
- I have defined an object-safe version of `ClientT` in this PR (`ObjectSafeClient`), but for extension traits to work, we need to implement `ClientT` for it, hence the `BoxedL2Client` wrapper. (Cannot `impl ClientT for Arc<dyn ObjectSafeClient>` due to orphaning rules.)
- Because we have a wrapper anyway and want component tagging, I've folded `Clone` and tagging into `ObjectSafeClient` as well. Tagging alone makes using `Arc` impossible. And if we want tagging, I *think* it should implemented as a trait, so client implementations can access tags w/o the necessity to pass them around on each call.
- Regarding the `BoxedL2Client` name, I've used [`BoxCloneService`](https://docs.rs/tower/latest/tower/util/struct.BoxCloneService.html) from `tower` as a reference. IMO, client is implied to be cloneable, so having `Clone` specified in its name would be an overkill.
- Regarding object safety / dynamic dispatch: I don't see any alternatives in the long term if the node framework is a priority. I can think of multiple ways to customize the client by adding more middleware, changing the transport, etc.

montekki: My main concern is that our client code will need to go at some point anyway. At some point we will need to start using something like the [machinery from ethers/alloy](https://docs.rs/ethers/latest/ethers/providers/struct.Provider.html) for it and its [counterpart for l2](https://github.com/lambdaclass/zksync-web3-rs). And it sort of manages things as i've described, so my question is simply about making code more similar to what it would be in the future to make future refactorings less painful.

montekki (approved):

montekki (approved):

PR #1634, opened 2024-04-10
fix(vm): Fix storage oracle and estimation

StanislavBreadless:
## What ❔

- Makes "free" slots also free with regard to I/O costs. These are context / operator balance slots which are global and warm for everyone.
- Fixes a bug with setting the initial value inside storage oracle. This bug only affects 1.5.0.
- The gas used for pubdata is also multiplied by the `estimated_fee_scale_factor`. For "small" transactions, e.g. transfers this gives us the same estimation overhead as in the previous (v1.4.1) system. But this will produce a bit larger overhead for big transactions that involve a lot of pubdata, but this is the correct consistent behavior and it should not affect most of the users.

## Checklist


core/lib/zksync_core/src/api_server/tx_sender/mod.rs
```diff
@@ -850,12 +850,8 @@ impl TxSender {
...
             .observe(number_of_iterations);
 
-        let tx_body_gas_limit = cmp::min(
-            MAX_L2_TX_GAS_LIMIT,
-            ((upper_bound as f64) * estimated_fee_scale_factor) as u64,
-        );
-
-        let suggested_gas_limit = tx_body_gas_limit + additional_gas_for_pubdata;
+        let suggested_gas_limit =
+            ((upper_bound + additional_gas_for_pubdata) as f64 * estimated_fee_scale_factor) as u64;
```
perekopskiy, line 854: should we check for overflow here?

core/lib/zksync_core/src/api_server/tx_sender/mod.rs
```diff
@@ -897,30 +893,29 @@ impl TxSender {
...
+        };
 
         tracing::info!(
             initiator = ?tx.initiator_account(),
             nonce = %tx.nonce().unwrap_or(Nonce(0)),
-            "fee estimation: gas for pubdata: {}, tx body gas: {tx_body_gas_limit}, overhead gas: {overhead} \
+            "fee estimation: gas for pubdata: {}, suggested gas: {suggested_gas_limit}, overhead gas: {overhead} \
             (with params base_fee: {base_fee}, gas_per_pubdata_byte: {gas_per_pubdata_byte})",
             (tx_metrics.pubdata_published as u64) * gas_per_pubdata_byte,
         );
```
perekopskiy, line 921: Let's change what we log here:
- `gas for pubdata` should be `tx_metrics.pubdata_published * gas_per_pubdata_byte * estimated_fee_scale_factor`
- tx body gas (or maybe just call it computational gas?) should be `upper_bound * estimated_fee_scale_factor`
- let's print `estimated_fee_scale_factor` param as well

core/lib/zksync_core/src/api_server/tx_sender/mod.rs
StanislavBreadless: The overflow is never expected here, assuming adequate value of `estimated_fee_scale_factor`. There are also checks at the bottom of the function that the returned gas limit does not go beyond batch gas limit

core/lib/zksync_core/src/api_server/tx_sender/mod.rs
StanislavBreadless: > tx body gas (or maybe just call it computational gas?) should be upper_bound * estimated_fee_scale_factor

@perekopskiy the problem with this approach is that the `additional_gas_for_pubdata` may be higher than the "correct" published gas, since it is checked under the condition of `max_gas_limit` to avoid doing binary search over gas that can be spent on pubdata only (the large gas limit means larger balance change). This is a very real scenario which I observed locally when testing large gas prices when the `upper_bound` was near zero because of `additional_gas_for_pubdata` being large enough

-

perekopskiy: then maybe `suggested_gas_limit - tx_metrics.pubdata_published * gas_per_pubdata_byte * estimated_fee_scale_factor`?

perekopskiy (approved):

PR #1636, opened 2024-04-10
feat: Extract commitment generator into a separate crate

AnastasiiaVashchuk:
## What ❔

Extract `zksync_core::commitment_generator` to separate the crate in the node folder
Put `zksync_commitment_utils` to  `zksync_commitment_generator ` (reason: utils are used only by generator)

## Why ❔

## Checklist


montekki (approved):

core/node/commitment_generator/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+categories.workspace = true
+
+[dependencies]
+vise.workspace = true
+zksync_types.workspace = true
+zksync_dal.workspace = true
+zksync_health_check.workspace = true
+zksync_l1_contract_interface.workspace = true
+zksync_utils.workspace = true
+zksync_commitment_utils.workspace = true
```
slowli, line 19: Bikeshedding: Since it's the only crate using `zksync_commitment_utils`, which is a single file, maybe it makes sense to move utils functionality to the `zksync_commitment_generator` crate? Or is it used somewhere externally?

Cargo.toml
```diff
@@ -222,3 +222,4 @@ zksync_crypto_primitives = { path = "core/lib/crypto_primitives" }
 zksync_node_framework = { path = "core/node/node_framework" }
 zksync_eth_watch = { path = "core/node/eth_watch" }
 zksync_shared_metrics = { path = "core/node/shared_metrics" }
+zksync_commitment_generator = { path = "core/node/commitment_generator" }
```
slowli, line 225: Not specific to this crate: Shouldn't this crate and other crates the `core/node` directory be a part of the workspace (i.e., be mentioned in `workspace.members` list)?

core/node/commitment_generator/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+multivm.workspace = true
+
+tokio = { workspace = true, features = ["time"] }
+anyhow.workspace = true
+thiserror.workspace = true
+async-trait.workspace = true
```
slowli, line 25: These two deps and `metrics` are unused; please remove them.

core/node/commitment_generator/src/lib.rs
slowli, line 1: Please change `use metrics::{CommitmentStage, METRICS};` to `use crate::metrics::{..}` so that it's clear that these imports are local. (I guess that's a possible reason behind including an unnecessary `metrics` dependency 🙃)

core/node/commitment_generator/Cargo.toml
AnastasiiaVashchuk: 1. it makes sense
2. no

Cargo.toml
slowli: A separate workspace doesn't really make sense for now because of a two-way dependency with the `core` / main workspace. It [turns out](https://doc.rust-lang.org/stable/cargo/reference/workspaces.html#the-members-and-exclude-fields) that all crates in the `node` dir are already members of the main workspace:

> All [path dependencies](https://doc.rust-lang.org/stable/cargo/reference/specifying-dependencies.html#specifying-path-dependencies) residing in the workspace directory automatically become members. Additional members can be listed with the members key, which should be an array of strings containing directories with Cargo.toml files.

(Which is expected: how would `*.workspace = true` keys work in the crate manifests otherwise?) I'd suggest to add crates to the main workspace explicitly so that it's more clear and there's no discrepancy with the `core/node/node_framework` crate, which *is* included explicitly into `workspace.members`.

core/node/commitment_generator/Cargo.toml
```diff
@@ -10,13 +10,25 @@ keywords.workspace = true
...
+multivm.workspace = true
 circuit_sequencer_api_1_4_0.workspace = true
 circuit_sequencer_api_1_4_1.workspace = true
 circuit_sequencer_api_1_5_0.workspace = true
 
 zk_evm_1_5_0.workspace = true
 zk_evm_1_4_1.workspace = true
 zk_evm_1_3_3.workspace = true
-multivm.workspace = true
+
```
slowli, line 27: Nit: Unnecessary empty line here and at the end of the file.

slowli (approved):

Cargo.toml
AnastasiiaVashchuk: ah..yes..
did not know that 
> I'd suggest to add crates to the main workspace explicitly 

make sense

PR #1639, opened 2024-04-10
[EVM-Equivalent-Yul] Add mload, mstore and mstore8

ManuelBilbao:
## What ❔

This PR adds `mload`, `mstore` and `mstore8` tests.

## Checklist

### DIFF
--- a/core/lib/multivm/src/versions/vm_latest/tests/evm_simulator.rs
+++ b/core/lib/multivm/src/versions/vm_latest/tests/evm_simulator.rs
@@ -3,7 +3,7 @@ use std::{
     str::FromStr,
 };
 
-use ethabi::{encode, Contract, Token};
+use ethabi::{encode, ethereum_types::H264, Contract, Token};
 use itertools::Itertools;
 use tracing::{instrument::WithSubscriber, Instrument};
 // FIXME: 1.4.1 should not be imported from 1.5.0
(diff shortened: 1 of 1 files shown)


core/lib/multivm/src/versions/vm_latest/tests/evm_simulator.rs
```diff
@@ -627,6 +627,237 @@ fn test_basic_signextend_vectors() {
...
+                hex::decode("51").unwrap(),
+                // push0
+                hex::decode("5F").unwrap(),
+                // sstore
+                hex::decode("55").unwrap(),
+            ]
+            .into_iter()
+            .concat()
+        ),
+        16_711_680.into()
```
gianbelinche, line 706: Why this number?

contracts
gianbelinche, line 1: I think we should push this, its the git submodule

gianbelinche (changes requested):
Everything else is ok!

core/lib/multivm/src/versions/vm_latest/tests/evm_simulator.rs
fborello-lambda: This [playground](https://www.evm.codes/playground?unit=Wei&codeType=Mnemonic&code=%27PUSH3%20125985%5CnPOP%27_) shows why 16_711_680 is used. 16711680 == 0xFF0000

PR #1642, opened 2024-04-11
feat(api): Allow granular max response size config

slowli:
## What ❔

Allows configuring max response size per method for JSON-RPC web servers.

## Why ❔

Makes EN more operator-friendly. Some RPC methods (e.g., `eth_getTransactionReceipt`) may return large responses, while for others, smaller output size is OK.

## Checklist


Deniallugo:
Please add it to file based config as well. 
https://github.com/matter-labs/zksync-era/blob/main/etc/env/file_based/general.yaml

core/lib/config/src/configs/api.rs
```diff
@@ -18,6 +26,106 @@ pub struct ApiConfig {
     pub merkle_tree: MerkleTreeApiConfig,
 }
 
+/// Response size limits for specific RPC methods.
+///
+/// The unit of measurement for contained limits depends on the context. In [`MaxResponseSize`],
+/// limits are measured in bytes, but in configs, limits are specified in MiBs.
+#[derive(Debug, Clone, PartialEq)]
+pub struct MaxResponseSizeOverrides(HashMap<String, NonZeroUsize>);
```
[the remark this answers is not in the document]
Deniallugo: I'd say it's dangerous and ddos prone

core/bin/external_node/src/config/mod.rs
```diff
@@ -158,6 +159,9 @@ pub(crate) struct OptionalENConfig {
     /// Maximum response body size in MiBs. Default is 10 MiB.
     #[serde(default = "OptionalENConfig::default_max_response_body_size_mb")]
     pub max_response_body_size_mb: usize,
+    /// Method-specific overrides in MiBs for the maximum response body size.
+    #[serde(default = "MaxResponseSizeOverrides::empty")]
```
slowli, line 163: @tomg10 Should we leave the default overrides empty for the EN, or can we come up with a more intelligent default (e.g., `eth_getTransactionReceipt=None`)?

Deniallugo: Should we really do it? 
Let's disable non limit responses.

perekopskiy (approved):

RomanBrodetski (approved):
Sorry for the delay in review.

The code looks good. As for EN defaults , I don't have a strong opinion - but I don't mind keeping it unbound. Power users will configure it, but average users will only access the EN's API themselves, so this is not such a large concern.

slowli:
Merging as-is (i.e., retaining the default response limit of 10 MB); if necessary, can update in a follow-up PR.

PR #1644, opened 2024-04-11
chore(env): Remove `ZKSYNC_HOME` env var from server

Deniallugo:
## What ❔

ZKSYNC_HOME has been removed from server. It's required for future integration with file based configs
## Checklist


core/bin/contract-verifier/src/main.rs
```diff
@@ -24,9 +24,9 @@ async fn update_compiler_versions(connection_pool: &ConnectionPool<Core>) {
     let mut storage = connection_pool.connection().await.unwrap();
     let mut transaction = storage.start_transaction().await.unwrap();
 
-    let zksync_home = std::env::var("ZKSYNC_HOME").unwrap_or_else(|_| ".".into());
+    let zksync_home = locate_workspace().unwrap_or_else(|| ".".into());
```
slowli, line 27: I think in this case it would be expected to provide `zksolc_path` / `solc_path` / ... as a part of the binary configuration (passed as env variables or whatever way we pass the config to binaries). Having it defined in terms of the workspace is an unnecessary and weird complication.

core/lib/contracts/src/lib.rs
```diff
@@ -40,8 +40,12 @@ const LOADNEXT_CONTRACT_FILE: &str =
 const LOADNEXT_SIMPLE_CONTRACT_FILE: &str =
     "etc/contracts-test-data/artifacts-zk/contracts/loadnext/loadnext_contract.sol/Foo.json";
 
-fn read_file_to_json_value(path: impl AsRef<Path>) -> serde_json::Value {
-    let zksync_home = std::env::var("ZKSYNC_HOME").unwrap_or_else(|_| ".".into());
+fn home_path() -> PathBuf {
```
slowli, line 43: Like with contract verifier, relying on the workspace looks suboptimal. It basically makes this crate (and all crates depending on it) non-publishable. I guess we could use as a stopgap solution, but IMO a longer-term solution would be to include the contracts into the library itself in compile time via something like `include_bytes!`. Still leaves the open question of where to get contract bytecode from, of course; in an ideal world, it would be a part of the build script. IMO, it would be acceptable to cheat in this build script for now, assuming that it's invoked in the workspace with the contracts already built; compare this with specifying `ROCKSDB_LIB_DIR` when building `rocksdb` crate. Would still be a significant improvement over the current runtime trickery.

core/lib/utils/src/env.rs
```diff
@@ -0,0 +1,31 @@
+use std::{path::PathBuf, sync::Mutex};
+
+static WORKSPACE: Mutex<Option<PathBuf>> = Mutex::new(None);
+
+pub fn locate_workspace() -> Option<PathBuf> {
```
slowli, line 5: Please reference code in `insta` as an inspiration for this function.

core/lib/utils/src/env.rs
```diff
@@ -0,0 +1,31 @@
+use std::{path::PathBuf, sync::Mutex};
+
+static WORKSPACE: Mutex<Option<PathBuf>> = Mutex::new(None);
```
slowli, line 3: Nit: Could probably use a `OnceCell` here, using [`get_or_try_init`](https://docs.rs/once_cell/latest/once_cell/sync/struct.OnceCell.html#method.get_or_try_init) for synchronization (IIUC, these calls are linearized).

core/lib/utils/src/env.rs
```diff
@@ -0,0 +1,31 @@
...
+
+    let output = std::process::Command::new(
+        std::env::var("CARGO")
+            .ok()
+            .unwrap_or_else(|| "cargo".to_string()),
+    )
+    .arg("locate-project")
+    .arg("--workspace")
+    .output()
+    .ok()?;
```
slowli, line 19: I'm not sure I agree with swallowing errors here; it means that if something is misconfigured, we'd repeatedly run into the same errors. Note that `insta` turns errors into panics, which is the reasoning behind unwrapping mutex poison errors.

prover/vk_setup_data_generator_server_fri/src/keystore.rs
```diff
@@ -38,17 +39,15 @@ pub enum ProverServiceDataType {
...
-    let zksync_home = std::env::var("ZKSYNC_HOME").unwrap_or_else(|_| "/".into());
-    format!(
-        "{}/prover/vk_setup_data_generator_server_fri/data",
-        zksync_home
-    )
+fn get_base_path_from_env() -> PathBuf {
+    locate_workspace()
+        .unwrap_or_else(|| ".".into())
+        .join("/prover/vk_setup_data_generator_server_fri/data")
 }
```
slowli, line 51: Like with the contract verifier, this looks unnecessarily complicated and weird; it would be better to provide the relevant path via config to the binary instead.

slowli (commented):
On the technical level, the approach seems to make sense (other than error handling). Conceptually though, I think that each use should be handled differently (could require more effort though, especially for `zksync_contracts`).

core/bin/contract-verifier/src/main.rs
Deniallugo: i don't want to change current behaviour, but we can implement it cc: @perekopskiy

core/lib/utils/src/env.rs
Deniallugo: so, it's likely won't work with pure binary and that's why we have a fallback

Deniallugo:
@slowli I agree with your statement, that we can treat it differently, but it'll require more effort and we could create a task for it

core/lib/utils/src/env.rs
slowli: Ah, I see. It'd still make sense to log errors, IMO.

Deniallugo: I can log it on a trace level, because they doesn't make sense in production

core/bin/contract-verifier/src/main.rs
perekopskiy: I think it isn't worth an effort. Contract verifier will be reworked by devrel team soonish.

core/lib/utils/src/env.rs
```diff
@@ -0,0 +1,44 @@
...
+        .ok()?
+        .get("root")
+        .cloned()?;
+
+    let serde_json::Value::String(root) = root else {
+        return None;
+    };
+
+    let val = PathBuf::from(root).parent()?.to_path_buf();
+    WORKSPACE.set(val.clone()).unwrap();
```
slowli, line 42: Does [`OnceCell::get_or_try_init()`](https://docs.rs/once_cell/latest/once_cell/sync/struct.OnceCell.html#method.get_or_try_init) not work here for some reason? (Shortcutting e.g., if the output JSON can't be parsed or doesn't contain `root` isn't really such a reason; it's an error that should be logged.) `set().unwrap()` can panic if called from multiple threads in parallel.

core/lib/contracts/src/lib.rs
```diff
@@ -156,19 +158,17 @@ static DEFAULT_SYSTEM_CONTRACTS_REPO: Lazy<SystemContractsRepo> =
 
 /// Structure representing a system contract repository - that allows
 /// fetching contracts that are located there.
-/// As most of the static methods in this file, is loading data based on ZKSYNC_HOME environment variable.
+/// As most of the static methods in this file, is loading data based on workspace location environment variable.
```
perekopskiy, line 161: s/workspace location environment variable/workspace location

prover/vk_setup_data_generator_server_fri/src/vk_commitment_helper.rs
```diff
@@ -17,11 +18,10 @@ pub fn write_contract_toml(contract_doc: Document) -> anyhow::Result<()> {
...
-        .with_context(|| format!("contract.toml file does not exist on path {path}"))?;
+        .with_context(|| format!("contract.toml file does not exist on path {path:?}"))?;
     toml_data.parse::<Document>().context("invalid config file")
 }
 
-pub fn get_contract_toml_path() -> String {
-    let zksync_home = std::env::var("ZKSYNC_HOME").unwrap_or_else(|_| "/".into());
-    format!("{}/etc/env/base/contracts.toml", zksync_home)
+pub fn get_contract_toml_path() -> PathBuf {
+    workspace_dir_or_current_dir().join("../etc/env/base/contracts.toml")
```
perekopskiy, line 26: double-checking: does `join` work well with `..`?

slowli: Yes, it does, it's covered by tests in the crate.

perekopskiy (approved):

PR #1651, opened 2024-04-11
feat(vm-runner): implement VM runner storage layer

itegulov:
## What ❔

Adds an abstraction that can load unprocessed batches as well as implement `ReadStorageFactory`. Implementation largely reuses the existing primitives from state keeper and they had to be generalized slightly.

## Why ❔

One of the components for the upcoming VM runner

## Checklist


core/lib/zksync_core/src/vm_runner/storage.rs
```diff
@@ -0,0 +1,196 @@
...
+    /// re-execute. These are the transactions that are included in a sealed miniblock belonging
+    /// to a sealed L1 batch (with state keeper being the source of truth). The order of the
+    /// transactions is the same as it was when state keeper executed them.
+    ///
+    /// Can return `None` if there are no batches to be processed.
+    ///
+    /// # Errors
+    ///
+    /// Propagates DB errors.
+    pub async fn load_next_batch(&self) -> anyhow::Result<Option<BatchData>> {
```
perekopskiy, line 146: I was thinking that this component wouldn't only serve as a provider but would continuously load data for new batches (loader should provide info about max batch ready to be loaded) and drop data that is not needed. I think it will be better than having this provider struct and actual updater separately.

core/lib/zksync_core/src/vm_runner/storage.rs
```diff
@@ -0,0 +1,196 @@
...
+}
+
+/// Abstraction for VM runner's storage layer that provides two main features:
+///
+/// 1. A [`ReadStorageFactory`] implementation backed by either Postgres or RocksDB (if it's
+/// caught up). Always initialized as a `Postgres` variant and is then mutated into `Rocksdb`
+/// once RocksDB cache is caught up.
+/// 2. Loads data needed to re-execute the next unprocessed L1 batch.
+#[derive(Debug)]
+pub struct VmRunnerStorage<L: VmRunnerStorageLoader> {
```
perekopskiy, line 63: I think it doesn't work as intended. My understanding is that currently:
`VmRunnerStorage::new` spawns a catch up task up to `loader.latest_processed_batch`. Postgres is used before RocksDB catches up. Then RocksDB is used and for each access it synchronizes with postgres up to latest sealed batch. This doesn't fit the role of the component. 
- It should keep RocksDB synced at `VmRunnerStorageLoader::latest_processed_batch + 1`
- It should allow using RocksDB for batch range `[VmRunnerStorageLoader::latest_processed_batch + 1, VmRunnerStorageLoader::last_ready_to_be_loaded_batch]`. This should be achieved by keeping storage diff for each batch in that range. Also, you would need a new implementation of `ReadStorage`, that would use rocksDB and in-memory storage diff. 
e.g . latest_processed_batch=5, last_ready_to_be_loaded_batch=10
someone VmRunner (from outside) request storage for l1_batch 10. You take storage diff for batches [7; 10] and RocksDB (that has data for L1 batch 6), and create this new struct that implements `ReadStorage`. `read_value` implementation would check if requested key is in the storage diff, if it's then simply return it's value, otherwise read from RocksDB. Same for other methods: `is_write_initial`, `get_enumeration_index` -- storage diff should be enough for this; `load_factory_dep`, `is_bytecode_known` -- you would need to keep list of new factory deps additionally to storage diff

itegulov: I have implemented a proof of concept of how this should work, if you could take a look that it matches your expectations from the API perspective that would be great.

Just be aware that the implementation is still very rough, especially with the amount of clones and unnecessary iterating it currently has. I am working on improvements as we speak.

core/lib/state/src/storage_factory.rs
```diff
@@ -0,0 +1,188 @@
...
+    fn get_enumeration_index(&mut self, key: &StorageKey) -> Option<u64> {
+        match self {
+            Self::Postgres(postgres) => postgres.get_enumeration_index(key),
+            Self::Rocksdb(rocksdb) => rocksdb.get_enumeration_index(key),
+            Self::RocksdbWithMemory(rocksdb, diffs, _) => {
+                match rocksdb.get_enumeration_index(key) {
+                    None => diffs
+                        .iter()
+                        .find_map(|x| x.get(key).and_then(|sv| sv.enum_index)),
+                    Some(value) => Some(value),
```
perekopskiy, line 171: nit: I think you should firstly search for `key` in diffs and only then rocksdb

core/lib/zksync_core/src/vm_runner/storage.rs
```diff
@@ -0,0 +1,456 @@
...
+    /// Unique name of the VM runner instance.
+    fn name() -> &'static str;
+
+    /// Returns number of the next unprocessed L1 batch. Can return `None` if there are no batches
+    /// to be processed.
+    ///
+    /// # Errors
+    ///
+    /// Propagates DB errors.
+    async fn first_unprocessed_batch(
```
perekopskiy, line 54: I think `first_unprocessed_batch` is not needed here, it should always be equal to `latest_processed_batch` + 1.
And I suggested adding the method `last_ready_to_be_loaded_batch` which should be capable of controlling max batch that should be loaded by `StorageSyncTask`

core/lib/zksync_core/src/vm_runner/storage.rs
```diff
@@ -0,0 +1,456 @@
...
+                .pool
+                .connection_tagged(L::name())
+                .await
+                .context("Failed getting a Postgres connection")?;
+            let max_l1_batch = (state.l1_batch_number + self.max_batches_to_load - 1).min(
+                conn.blocks_dal()
+                    .get_sealed_l1_batch_number()
+                    .await?
+                    .unwrap_or_default(),
+            );
```
perekopskiy, line 155: I think max_l1_batch should be the last batch in `state`

core/lib/zksync_core/src/vm_runner/storage.rs
```diff
@@ -0,0 +1,456 @@
...
+                    l1_batch_number,
+                    state.l1_batch_number,
+                    state.l1_batch_number + self.max_batches_to_load - 1
+                );
+            }
+            let rocksdb_builder = RocksdbStorageBuilder::from_rocksdb(rocksdb.clone());
+            let rocksdb = rocksdb_builder
+                .synchronize(&mut conn, stop_receiver, Some(state.l1_batch_number))
+                .await
+                .context("Failed to catch up RocksDB storage to Postgres")?;
```
perekopskiy, line 168: RocksDB synchronization ideally shouldn't block access_storage. IIUC we can read and write to RocksDB in parallel.

core/lib/zksync_core/src/vm_runner/storage.rs
```diff
@@ -0,0 +1,456 @@
...
+            let latest_processed_batch = self.loader.latest_processed_batch(&mut conn).await?;
+            let next_batch = (latest_processed_batch + 1).min(
+                conn.blocks_dal()
+                    .get_sealed_l1_batch_number()
+                    .await?
+                    .unwrap_or_default(),
+            );
+            let rocksdb_builder = RocksdbStorageBuilder::from_rocksdb(rocksdb.clone());
+            let rocksdb = rocksdb_builder
+                .synchronize(&mut conn, &stop_receiver, Some(next_batch))
```
perekopskiy, line 346: I think you should sync storage up to `next_batch - 1`, because executing `next_batch` requires storage at the end of `next_batch - 1`.

core/lib/zksync_core/src/vm_runner/storage.rs
```diff
@@ -0,0 +1,456 @@
...
+            state.l1_batch_number = next_batch;
+            state
+                .storage
+                .retain(|l1_batch_number, _| l1_batch_number >= &next_batch);
+            let max_present = state
+                .storage
+                .last_entry()
+                .map(|e| *e.key())
+                .unwrap_or(next_batch - 1);
+            let max_desired = next_batch + self.max_batches_to_load;
```
perekopskiy, line 362: max_desired should be limited by `last_ready_to_be_loaded_batch` method

core/lib/zksync_core/src/vm_runner/storage.rs
perekopskiy: Yes, conceptually looks good! I left some comments though, please check

core/lib/state/src/storage_factory.rs
itegulov: This is the order in which RocksDB synchronization did it so I assumed there is some reasoning behind it, but I can change it for sure

core/lib/zksync_core/src/vm_runner/storage.rs
itegulov: So, I forgot to add the `loop` sequence that sleeps and blocks until the data is available here which is fixed now. I can do it your way too, but then this will mean that `access_storage` is going to become fallible because the data is not available yet. Which is going to shift the retry loop to the user which I think would be worse.

core/lib/zksync_core/src/vm_runner/storage.rs
itegulov: This isn't a "real" synchronization. Currently we have `RocksdbStorage` structured in a way that the only way to get an instance is through running `synchronize` on a builder. I will see if I can refactor this in a way that makes more sense though.

core/lib/zksync_core/src/vm_runner/storage.rs
itegulov: Ah yes fair point

core/lib/zksync_core/src/vm_runner/storage.rs
itegulov: I have improved the invariant by including `RocksdbStorage` in the state so it should be better now

core/lib/zksync_core/src/vm_runner/storage.rs
itegulov: Nevermind the above, I have decided to get rid of blocking logic entirely and now `max_l1_batch` is indeed just the last batch in `state`

core/lib/zksync_core/src/vm_runner/storage.rs
```diff
@@ -0,0 +1,397 @@
...
+            state.l1_batch_number = latest_processed_batch;
+            state
+                .storage
+                .retain(|l1_batch_number, _| l1_batch_number > &latest_processed_batch);
+            let max_present = state
+                .storage
+                .last_entry()
+                .map(|e| *e.key())
+                .unwrap_or(latest_processed_batch);
+            let max_desired = self.loader.last_ready_to_be_loaded_batch(&mut conn).await?;
```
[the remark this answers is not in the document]
perekopskiy: Makes sense, agree

core/lib/zksync_core/src/vm_runner/storage.rs
```diff
@@ -0,0 +1,196 @@
...
+    loader: L,
+}
+
+impl<L: VmRunnerStorageLoader> VmRunnerStorage<L> {
+    /// Creates a new VM runner storage using provided Postgres pool and RocksDB path.
+    pub async fn new(
+        pool: ConnectionPool<Core>,
+        rocksdb_path: String,
+        loader: L,
+        enum_index_migration_chunk_size: usize, // TODO: Remove
```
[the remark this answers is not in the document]
itegulov: See https://github.com/matter-labs/zksync-era/pull/1734

core/lib/zksync_core/src/vm_runner/storage.rs
```diff
@@ -0,0 +1,400 @@
...
+                tracing::info!("`StorageSyncTask` was interrupted");
+                return Ok(());
+            }
+            // It's important to hold a lock on `state` while we are updating RocksDB cache
+            // as otherwise `VmRunnerStorage` will have an inconsistent view on DB state.
+            let mut state = self.state.write().await;
+            let mut conn = self.pool.connection_tagged(L::name()).await?;
+            let latest_processed_batch = self.loader.latest_processed_batch(&mut conn).await?;
+            let rocksdb_builder = RocksdbStorageBuilder::from_rocksdb(rocksdb.clone());
+            if rocksdb_builder.l1_batch_number().await == Some(latest_processed_batch) {
```
perekopskiy, line 285: l1_batch_number() returns last batch persisted in rocksdb + 1, you should take in into account

core/lib/zksync_core/src/vm_runner/storage.rs
```diff
@@ -0,0 +1,400 @@
...
+            anyhow::anyhow!("Expected RocksDB to be initialized by `AsyncCatchupTask`")
+        })?;
+        loop {
+            if *stop_receiver.borrow() {
+                tracing::info!("`StorageSyncTask` was interrupted");
+                return Ok(());
+            }
+            // It's important to hold a lock on `state` while we are updating RocksDB cache
+            // as otherwise `VmRunnerStorage` will have an inconsistent view on DB state.
+            let mut state = self.state.write().await;
```
perekopskiy, line 281: hmm, if storage is used as intended, i.e. one never holds storage access to a batch with number `<= latest_processed_batch`, then it should be safe to write to RocksDB in parallel with data being read from it. I would prefer not to hold a lock here.
We can just say that implementation relies on the fact that one never uses storage pointer to a batch with number `<= latest_processed_batch`, otherwise -- UB

itegulov: Aren't both approaches pretty much equivalent to each other though?

If we expect the users to drop all storage access handles, it effectively means we expect them to drop all read locks on `state`. And they wouldn't be able to start reading again before we synchronize RocksDB to the next `latest_processed_batch ` anyway, effectively meaning they are waiting for write lock on `state` to drop.

core/lib/zksync_core/src/vm_runner/storage.rs
itegulov: Thanks, I keep forgetting about this quirk!

I will try to add more tests to cover various corner cases.

core/lib/zksync_core/src/vm_runner/tests/mod.rs
```diff
@@ -0,0 +1,424 @@
...
+                vm_runner_storage
+                    .access_storage_eventually(&receiver, L1BatchNumber(i))
+                    .await
+            })?;
+            // Check that both storages have identical key-value pairs written in them
+            for storage_log in &storage_logs {
+                let storage_key =
+                    StorageKey::new(AccountTreeId::new(storage_log.address), storage_log.key);
+                let expected = pg_storage.read_value(&storage_key);
+                let actual = vm_storage.read_value(&storage_key);
```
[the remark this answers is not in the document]
itegulov: Should be better now. While writing the tests I discovered that `PostgresStorage` has a bug where it reports wrong enum index when used with non-latest sealed batch. Should be fixed now.

core/lib/zksync_core/src/vm_runner/storage.rs
perekopskiy: it seems no, e.g. if latest_processed_batch=10, last_ready_to_be_loaded_batch=12. Batches 11 and 12 are being processed. Then 11th becomes processed and I suggest that we should update rocksDB with storage logs from 11th batch while 12th is still in-progress. 
Moreover, the main loop of `StorageSyncTask` firstly synchronizes RocksDB and then loads data for new batches to memory, so with the current approach, when batch 13 becomes ready it cannot be started before 12 finishes which contradicts the purpose of vm runner storage

itegulov: You are right, I just had a brain fart that we can't update `latest_processed_batch` to anything other than `last_ready_to_be_loaded_batch`.

itegulov: Should be done now, PTAL. Also, we should only prohibit access storage handles for batches `< latest_processed_batch`. Equivalent are still okay and even necessary to process the next batch.

perekopskiy (approved):
Looks good! Please fix lints

perekopskiy (approved):

PR #1652, opened 2024-04-11
feat(en): add manual vacuum step in db pruning

tomg10:


slowli (approved):
This approach works if `VACUUM` latency isn't too bad. Otherwise, it'll slow down DB pruning itself, so it could make sense to make it asynchronous w.r.t. the main pruning cycle.

PR #1658, opened 2024-04-11
chore(en): docker compose bumped version

tomg10:


core/bin/external_node/src/init.rs
```diff
@@ -77,8 +77,8 @@ pub(crate) async fn ensure_storage_initialized(
         InitDecision::SnapshotRecovery => {
             anyhow::ensure!(
                 consider_snapshot_recovery,
-                "Snapshot recovery is required to proceed, but it is not enabled. Enable by supplying \
-                 `--enable-snapshots-recovery` command-line arg to the node binary, or reset the node storage \
+                "Snapshot recovery is required to proceed, but it is not enabled. Enable by setting \
```
Deniallugo, line 80: Could we move back to `--enable-snapshots-recovery` ?
I'm doing my best to reduce dependencies from env variables

tomg10: yes, I can revert that

PR #1661, opened 2024-04-11
feat(main): gas oracle for base token impl

fkrause98:
## What ❔

Properly lint and restore the gas-oracle changes for kl-factory.


.github/workflows/ci-core-reusable.yml
```diff
@@ -110,8 +110,9 @@ jobs:
       - name: Deploy legacy era contracts
         run: ci_run zk contract setup-legacy-bridge-era
 
-      - name: Perform loadtest
-        run: ci_run zk run loadtest
+      - name: Perform loadtest - DISABLED (EVM-541)
```
popzxc, line 113: Are these changes expected? Disabling loadtest and revert test in the main branch doesn't seem to be a good idea.

Also the commit history in this PR looks odd, probably something got broken when working with multiple branches.

juan518munoz: This change is no longer present, ci workflows are the same than `main` (for now at least).

PR #1663, opened 2024-04-11
[EVM-Equivalent-YUL] Add tests for keccak256 opcode

fborello-lambda:
## What ❔

Add tests for `keccak256` opcode.

Note: The `MSTORE` opcode has to be implemented in order to pass the first `assert` of the test. The second `assert` passes because the `keccak256` function is applied to the memory with an offset of 0 and length of 4 bytes, which is equal to 0x0000_0000, since no value was stored in memory.

## Checklist


gianbelinche (approved):
LGTM

jrchatruc (approved):

PR #1666, opened 2024-04-12
fix(ci): Bump `release-please-action` to v4

slowli:
## What ❔

Fixes release-please config broken after #1584, by bumping the corresponding GitHub action.

## Why ❔

release-please shouldn't be broken.

## Checklist

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -17,13 +17,11 @@ jobs:
     steps:
       - name: Run release-please
         id: release
-        uses: google-github-actions/release-please-action@v3
+        uses: google-github-actions/release-please-action@v4
         with:
             token: ${{ secrets.RELEASE_TOKEN }}
-            command: manifest
             config-file: .github/release-please/config.json
             manifest-file: .github/release-please/manifest.json
-            monorepo-tags: true
 
       - name: Send Release Info
         if: ${{ steps.release.outputs.releases_created }}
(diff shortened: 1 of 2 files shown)


slowli:
I've changed the action config according to [the migration guide](https://github.com/google-github-actions/release-please-action#upgrading-from-v3-to-v4). Is there any way to test that it works w/o merging this PR?

.github/workflows/release-please.yml
```diff
@@ -17,13 +17,11 @@ jobs:
     steps:
       - name: Run release-please
         id: release
-        uses: google-github-actions/release-please-action@v3
+        uses: google-github-actions/release-please-action@v4
         with:
             token: ${{ secrets.RELEASE_TOKEN }}
-            command: manifest
```
Deniallugo, line 23: are you sure you have to change it?

slowli: `command` is not specified as [an action input](https://github.com/google-github-actions/release-please-action#action-inputs), so I think so. OTOH, [the migration guide](https://github.com/google-github-actions/release-please-action?tab=readme-ov-file#command) doesn't explicitly say to remove it 🙃

Deniallugo: ok, let's try, but please have a look at ci right after merge

Deniallugo (approved):

alexandrst88 (approved):

PR #1669, opened 2024-04-12
chore(gas_tracker): Split component into 2 parts and use it directly from eth_sender and state_keeper

AnastasiiaVashchuk:
Inline parts of `gas_tracker` into `eth_sender` and `state_keeper`  according to their use.
## Checklist


slowli (commented):
The `gas_tracker` module is not a component, and IMO, it doesn't merit a standalone crate. AFAICT, its public interface consists of two non-intersecting parts (which don't intersect in terms of the implementation either):

- `agg_l1_batch_base_cost` is used exclusively by `eth_sender`.
- Everything else is used exclusively by `state_keeper`.

If this is the case, I'd suggest to "inline" these parts into `eth_sender` and `state_keeper` and deal with the module in this way.

core/lib/zksync_core/src/state_keeper/utils.rs
```diff
@@ -1,24 +1,14 @@
...
-
-pub fn agg_l1_batch_base_cost(op: AggregatedActionType) -> u32 {
-    match op {
-        AggregatedActionType::Commit => AGGR_L1_BATCH_COMMIT_BASE_COST,
-        AggregatedActionType::PublishProofOnchain => AGGR_L1_BATCH_PROVE_BASE_COST,
-        AggregatedActionType::Execute => AGGR_L1_BATCH_EXECUTE_BASE_COST,
-    }
-}
-
 pub fn l1_batch_base_cost(op: AggregatedActionType) -> u32 {
```
slowli, line 12: Nit: `pub` functions in this module would probably work with `pub(super)` visibility (i.e., they can be made internal to the state keeper). Would allow to more clearly express the fact that they are only used internally by the state keeper.

core/lib/constants/src/gas.rs
```diff
@@ -0,0 +1,17 @@
+// Currently, every `AGGR_* cost` is overestimated,
+// so there are safety margins around `100_000 -- 200_000`
+
+pub const AGGR_L1_BATCH_COMMIT_BASE_COST: u32 = 242_000;
```
slowli, line 4: Just curious: Would these constants be shared with other crates in the future? If not, could make sense to define them in the modules that use them (this worked previously, right?).

core/lib/zksync_core/src/state_keeper/updates/l1_batch_updates.rs
```diff
@@ -56,9 +56,9 @@ mod tests {
     use zksync_types::{MiniblockNumber, ProtocolVersionId, H256};
 
     use super::*;
-    use crate::{
-        gas_tracker::new_block_gas_count,
-        state_keeper::tests::{create_execution_result, create_transaction},
+    use crate::state_keeper::{
+        tests::{create_execution_result, create_transaction},
+        utils::new_block_gas_count,
```
slowli, line 61: Nit: Looks like this import and similar imports in other test modules are redundant because of the `use super::*` wildcard import above.

AnastasiiaVashchuk: `super::*` give access only for `updates` mod in this case..so I think we need these imports from `crate::state_keeper::{`

PR #1674, opened 2024-04-12
fix(kl-factory): integration test fixes

juan518munoz:
## What ❔

- Fix first `zk test i server` for non eth based chains.
- Prevent base token tests from running when using an eth based chain.
- Tighten `ETH` check on base token deposit test.

## Checklist


juan518munoz:
`kl-factory` no longer has the issues that this PR aimed to solve, closing this.

PR #1683, opened 2024-04-15
fix: Fixing integration test for custom base token & temporarily disabling failing CI tests

mm-zk:
## What ❔

* Test was failing due to EVM-554 (wrong estimate of gas in the SDK).

For now, I've manually specified the gas limit.

* I've also disabled the parts of the CI (and created tickets for each one), to make the current CI pass.

### DIFF
--- a/core/tests/ts-integration/tests/ether.test.ts
+++ b/core/tests/ts-integration/tests/ether.test.ts
@@ -28,6 +28,10 @@ describe('ETH token checks', () => {
     });
 
     test('Can perform a deposit', async () => {
+        if (!isETHBasedChain) {
+            // TODO(EVM-555): Currently this test is not working for non-eth based chains.
+            return;
+        }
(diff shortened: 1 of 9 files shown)


core/tests/ts-integration/tests/ether.test.ts
```diff
@@ -28,6 +28,10 @@ describe('ETH token checks', () => {
     });
 
     test('Can perform a deposit', async () => {
+        if (!isETHBasedChain) {
+            // TODO(EVM-555): Currently this test is not working for non-eth based chains.
```
ly0va, line 32: From CI it looks like it's not working for eth-based chains as well, is that expected?

mm-zk: you're right. I'll disable it for all scenarios for now.

ly0va (approved):

PR #1684, opened 2024-04-15
feat: Extract BWIP into separate crate

AnastasiiaVashchuk:
## What ❔

Extract `zksync_core::basic_witness_input_producer` to separate crate in the node folder

## Why ❔

## Checklist


slowli:
@AnastasiiaVashchuk I thought that BWIP is basically broken at this point and is not run as a component on any environment. If this is the case, does it make sense to move it to the node framework? Maybe, just remove it from the codebase? 🙃

AnastasiiaVashchuk:
> @AnastasiiaVashchuk I thought that BWIP is basically broken at this point and is not run as a component on any environment. If this is the case, does it make sense to move it to the node framework? Maybe, just remove it from the codebase? 🙃

thanks for paging.
won't add it to framework..i think after vm runner integration we will remove it

PR #1685, opened 2024-04-15
feat: Extract house keeper into separate crate

AnastasiiaVashchuk:
## What ❔

Extract `zksync_core::house_keeper` to separate crate in the `node`  folder

## Why ❔

## Checklist


core/node/house_keeper/README.md
```diff
@@ -0,0 +1,11 @@
+# zkSync Era housekeeper
+
+This crate contains functionality for performing “administrative” work to keep the system flowing. It does:
+
+- reporting
+- archiving
+- retrying
+- monitoring
```
slowli, line 8: Nit: I feel that "monitoring" and "reporting" refer to roughly the same thing. Also, "retrying" and "queue cleaning" items are not clear; are they about about jobs (if so, it could make sense to have a single item: "Scheduling jobs, retrying failed jobs, and cleaning obsolete jobs").

core/node/house_keeper/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+license.workspace = true
+keywords.workspace = true
+categories.workspace = true
+
+[dependencies]
+vise.workspace = true
+zksync_dal.workspace = true
+zksync_shared_metrics.workspace = true
+zksync_utils.workspace = true
+zksync_db_connection.workspace = true
```
slowli, line 17: I don't think this dependency is necessary; AFAICT, house keeper only uses `ConnectionPool`, which is re-exported from `zksync_dal`.

core/node/house_keeper/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+repository.workspace = true
+license.workspace = true
+keywords.workspace = true
+categories.workspace = true
+
+[dependencies]
+vise.workspace = true
+zksync_dal.workspace = true
+zksync_shared_metrics.workspace = true
+zksync_utils.workspace = true
```
slowli, line 16: Nit: Looks like it's only used for `seconds_since_epoch` in a single place; IMO, it could make sense to inline this call.

core/node/house_keeper/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+zksync_db_connection.workspace = true
+prover_dal.workspace = true
+zksync_types.workspace = true
+zksync_config.workspace = true
+
+async-trait.workspace = true
+tokio = { workspace = true, features = ["time"] }
+anyhow.workspace = true
+tracing.workspace = true
+metrics.workspace = true
```
slowli, line 26: Would it be too much to rewrite the house keeper metrics to use `vise`? IIRC, it's essentially the only core component that still uses them. Could be done separately, of course.

AnastasiiaVashchuk: >IIRC, it's essentially the only core component that still uses them
does not looks so(photo1)
image


Taking into account that this is not the only one I would prefer to do this in the separate PR.

core/node/house_keeper/Cargo.toml
AnastasiiaVashchuk: do u want to say that if there is a single call place we have to inline?

core/node/house_keeper/Cargo.toml
AnastasiiaVashchuk: >which is re-exported from zksync_dal.

wdym?
from what I see its not - https://github.com/matter-labs/zksync-era/blob/9948187aa12bcd99ef512ec8cf322fefe2cd745b/core/lib/db_connection/src/connection_pool.rs#L249

slowli: `ConnectionPool` is defined in `zksync_db_connection`, yes, but it's re-exported from `zksync_dal` as well:

https://github.com/matter-labs/zksync-era/blob/9948187aa12bcd99ef512ec8cf322fefe2cd745b/core/lib/dal/src/lib.rs#L10

AnastasiiaVashchuk: ah i see, ok

core/node/house_keeper/README.md
AnastasiiaVashchuk: now should be correct

slowli (approved):

PR #1688, opened 2024-04-15
fix: Revert test

Raid5594:
## What ❔

Fixing the reverter test.
Fixes EVM-539

## Why ❔

Updates related to shared bridge.

## Checklist


core/lib/zksync_core/src/block_reverter/mod.rs
```diff
@@ -340,21 +340,27 @@ impl BlockReverter {
...
         let web3 = Web3::new(Http::new(&eth_config.eth_client_url).unwrap());
         let contract = state_transition_chain_contract();
-        let signer = PrivateKeySigner::new(
-            eth_config
-                .reverter_private_key
-                .expect("Private key is required to send revert transaction"),
-        );
+        // Your hexadecimal string
+        let hex_string = "e131bc3f481277a8f73d680d9ba404cc6f959e64296e0914dded403030d4f705";
+
```
Raid5594, line 345: Where's the best way to place the private key?

I was thinking of extending the `eth_sender` with something like this:

    // Don't load private key, if it's not required.
    #[deprecated]
    pub fn reverter_private_key(&self) -> Option<H256> {
        std::env::var("0xe131bc3f481277a8f73d680d9ba404cc6f959e64296e0914dded403030d4f705")
            .ok()
            .map(|pk| pk.parse().unwrap())
    }

core/lib/zksync_core/src/block_reverter/mod.rs
```diff
@@ -340,21 +340,27 @@ impl BlockReverter {
...
-            .expect(
-                "Either `revertBlocks` or `revertBatches` function must be present in contract",
-            );
+        let revert_function = contract.function("revertBatchesSharedBridge").expect(
+            "Either `revertBlocks` or `revertBatches` function must be present in contract",
+        );
         let data = revert_function
-            .encode_input(&[Token::Uint(last_l1_batch_to_keep.0.into())])
+            .encode_input(&[
+                Token::Uint(270.into()),
```
Raid5594, line 361: What would be the right way to fetch l2 chain id?

core/lib/zksync_core/src/block_reverter/mod.rs
mm-zk: normally we load the private keys from the environment variables.
(like it was here before: eth_config.reverter_private_key

mm-zk:
You also have to uncomment: https://github.com/matter-labs/zksync-era/blob/kl-factory/.github/workflows/ci-core-reusable.yml#L219

so that this revert test is actually run in the CI

etc/env/base/private.toml
```diff
@@ -9,6 +9,7 @@ test_database_prover_url = "postgres://postgres:notsecurepassword@localhost:5433
 [eth_sender.sender]
 # Set in env file for development, production, staging and testnet.
 operator_private_key = "0x27593fea79697e947890ecbecce7901b0008345e5d7259710d0dd5e500d040be"
+reverter_private_key = "0xe131bc3f481277a8f73d680d9ba404cc6f959e64296e0914dded403030d4f705"
```
mm-zk, line 12: comment how this differs. (and why do we even need a separate key for the reverter and cannot simply reuse the operator key)

core/tests/revert-test/tests/tester.ts
```diff
@@ -10,7 +10,8 @@ export class Tester {
         public ethProvider: ethers.providers.Provider,
         public ethWallet: ethers.Wallet,
         public syncWallet: zkweb3.Wallet,
-        public web3Provider: zkweb3.Provider
+        public web3Provider: zkweb3.Provider,
+        public hyperchainAdmin: ethers.Wallet
```
mm-zk, line 14: can you add a small comment on what is a hyperchain admin?

core/lib/zksync_core/src/block_reverter/mod.rs
```diff
@@ -333,6 +336,7 @@ impl BlockReverter {
         priority_fee_per_gas: U256,
         nonce: u64,
     ) {
+        const DEFAULT_ZKSYNC_CHAIN_ID: u128 = 270; // Default value if the environment variable is not set or cannot be parsed
```
mm-zk, line 339: too risky. If env variable is not set  - and AFAIK this is a separate tool, right? - then I'd prefer to crash (assert) mentioning that this is a wrong config.

mm-zk (changes requested):

etc/env/base/private.toml
Raid5594: You were right, we can use the same key

core/tests/revert-test/tests/tester.ts
Raid5594: Added, using the same wallet as ethWallet didn't work here, the server would fail after one of txs

core/tests/revert-test/tests/tester.ts
```diff
@@ -50,7 +57,10 @@ export class Tester {
             console.log(`Canceled ${cancellationTxs.length} pending transactions`);
         }
 
-        return new Tester(ethProvider, ethWallet, syncWallet, web3Provider);
+        const baseTokenAddress = process.env.CONTRACTS_BASE_TOKEN_ADDR!;
+        const isETHBasedChain = baseTokenAddress == zkweb3.utils.ETH_ADDRESS_IN_CONTRACTS;
+
+        return new Tester(ethProvider, ethWallet, syncWallet, web3Provider, hyperchainAdmin, isETHBasedChain, baseTokenAddress);
```
Raid5594, line 63: Added basetoken address & check for chain type in Tester

.github/workflows/ci-core-reusable.yml
```diff
@@ -216,13 +216,11 @@ jobs:
       - name: Fee projection tests
         run: ci_run zk test i fees
 
-      - name: Run revert test - DISABLED (EVM-539)
```
StanislavBreadless, line 219: there are two places in CI where it is disabled, you've turned on only one of those

Raid5594: Ah ok thanks!

PR #1690, opened 2024-04-15
refactor: removed polling from external io

pompon0:
ExternalIO contained unnecessary sleeps, which slowed down block processing.


RomanBrodetski (approved):
This looks strictly better - I might be missing something so let's also wait for @slowli 's review

core/lib/zksync_core/src/sync_layer/external_io.rs
```diff
@@ -220,34 +217,31 @@ impl StateKeeperIO for ExternalIO {
...
-                    anyhow::bail!("unexpected action in the action queue: {other:?}");
-                }
-                None => {
-                    tokio::time::sleep(POLL_INTERVAL).await;
-                }
+        let Some(action) = self.actions.recv_action(max_wait).await else {
+            return Ok(None);
+        };
+        match action {
+            SyncAction::OpenBatch {
```
slowli, line 224: Bikeshedding: I'd retain matching on the `Option`al action instead of a let-else above, IMO, it's a tiny bit more readable.

slowli (approved):
I wanted to do a similar refactoring myself 🙃 I *think* it's safe and equivalent to the previous implementation. Maybe, there was a reason @popzxc has organized waiting in a cycle, but I cannot think of one.

PR #1696, opened 2024-04-16
refactor: Rename miniblock to l2 block – base + state keeper

slowli:
## What ❔

Renames "miniblock" to "L2 block" in types, DAL and state keeper.

## Why ❔

- "Miniblock" is an internal term that is somewhat difficult to understand; "L2 block" is better in this regard.
- It makes sense for renaming to be complete (i.e., not only concern logs, but also the codebase), so that naming is reasonably consistent.

## Checklist


slowli:
May be obvious: if this approach is OK, I plan to create smaller follow-up PRs to change naming in other components.

perekopskiy (dismissed):
Nice!

perekopskiy (approved):

PR #1701, opened 2024-04-16
fix(en): Fix miscellaneous snapshot recovery nits

slowli:
## What ❔

- Allows to configure concurrency during DB recovery and restricts it to 10 by default.
- Sets the "affected" health status while the recovery is in progress.

## Why ❔

- DB recovery is very I/O-heavy, so having concurrency less than the DB pool size can utilize node resources more efficiently.
- "Affected" health status is more idiomatic; the node with such a status still passes readiness checks, but it signals that it isn't fully functional.

## Checklist


Cargo.toml
```diff
@@ -85,6 +85,7 @@ clap = "4.2.2"
 codegen = "0.2.0"
 criterion = "0.4.0"
 ctrlc = "3.1"
+derive_more = "=1.0.0-beta.6"
```
slowli, line 88: Unfortunately, the version of `derive_more` already used in the workspace (0.99.17) doesn't have good `Debug` derivation support.

slowli:
JFYI, this PR now has substantial intersection with #1532 in terms of metadata calculator. I'd prefer for this PR to be merged first.

RomanBrodetski (approved):

PR #1708, opened 2024-04-17
fix: Fixing local node

mm-zk:
## What ❔

* Fixing docker build and scripts for the local node/setup.

## Why ❔

* This will allow local node to work again - which is crucial before the bridgehub udpate.
* This PR is only fixing it in the 'main' branch - there will be a followup PR to the bridgehub branch coming later.

### DIFF
--- a/docker/local-node/Dockerfile
+++ b/docker/local-node/Dockerfile
@@ -1,11 +1,12 @@
 # Image is always built from the server image to reuse the common parts
 # This image is expected to be built locally beforehand (implemented in the `zk` tool)
-FROM matterlabs/server-v2:latest2.0
+ARG BASE_VERSION=latest2.0
+FROM matterlabs/server-v2:${BASE_VERSION}
 
 WORKDIR /
 
 # Install required dependencies
-RUN apt-get update; apt-get install -y make bash git openssl libssl-dev gcc g++ curl pkg-config software-properties-common jq wget
+RUN apt-get update; apt-get install -y make bash git openssl libssl-dev gcc g++ curl pkg-config software-properties-common jq wget vim-tiny
 RUN apt-get install -y curl gnupg libpq5 ca-certificates postgresql-client && rm -rf /var/lib/apt/lists/*
 
 # Install node and yarn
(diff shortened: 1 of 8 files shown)


docker/local-node/entrypoint.sh
```diff
@@ -1,46 +1,96 @@
...
+    update_config "/etc/env/base/database.toml" "state_keeper_db_path" "/var/lib/zksync/data/state_keeper"
+    update_config "/etc/env/base/database.toml" "backup_path" "/var/lib/zksync/data/backups"
 
-    # Switch zksolc compiler source from docker to binary
-    sed -i "s!'docker'!'binary'!" /contracts/l2-contracts/hardhat.config.ts
+    zk config compile
+  
+
+    zk db drop
+    zk db setup
```
ly0va, line 73: nit: equivalent to `zk db reset`

ly0va (dismissed):

docker/local-node/entrypoint.sh
mm-zk: done.

ly0va (approved):

RomanBrodetski (approved):
don't see anything suspicious

docker/local-node/Dockerfile
```diff
@@ -1,35 +1,37 @@
...
-# Copy compiler (both solc and zksolc) binaries
-# Obtain `solc` 0.8.12.
-RUN wget https://github.com/ethereum/solc-bin/raw/gh-pages/linux-amd64/solc-linux-amd64-v0.8.12%2Bcommit.f00d7308 \
-    && mv solc-linux-amd64-v0.8.12+commit.f00d7308 /usr/bin/solc \
-    && chmod +x /usr/bin/solc
+ENV RUSTUP_HOME=/usr/local/rustup \
+    CARGO_HOME=/usr/local/cargo \
+    PATH=/usr/local/cargo/bin:$PATH
+
+RUN curl https://sh.rustup.rs -sSf | bash -s -- -y && \
```
yorik, line 24: It would be better to do `apt-get install rust`

yorik (approved):

docker/local-node/Dockerfile
mm-zk: apt-get install rust fails - saying that there is no such package...

If you have other suggestion, I can fix it in the next PR.

PR #1715, opened 2024-04-17
feat: Adding local-node support for bridgehub

mm-zk:
## What ❔

* Fixing local node - now it will work with bridgehub
* Fixes EVM-564

## Why ❔

* This will allow easy testing.
* In the next PRs, I'll introduce ways to create local-node with custom base token and with validiums.


ly0va (approved):
Should we add some kind of CI for local node as well? Just to be sure that it builds & runs at least

benceharomi (approved):

PR #1717, opened 2024-04-17
fix(merkle_tree): don't panic in `BlockOutputWithProofs::verify_proofs`

haraldh:
## What ❔

don't panic in `BlockOutputWithProofs::verify_proofs` and rather return a `Result`

## Why ❔

So  `BlockOutputWithProofs::verify_proofs` can be used by other components.

## Checklist

### DIFF
--- a/core/lib/merkle_tree/src/hasher/proofs.rs
+++ b/core/lib/merkle_tree/src/hasher/proofs.rs
@@ -2,6 +2,8 @@
 
 use std::mem;
 
+use anyhow::ensure;
+
 use crate::{
     hasher::{HashTree, HasherWithStats},
     types::{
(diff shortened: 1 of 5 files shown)


core/lib/merkle_tree/src/hasher/proofs.rs
```diff
@@ -14,26 +16,38 @@ use crate::{
...
-    /// Panics if the proof doesn't verify.
+    #[must_use = "this method returns a boolean value, use it to check the result"]
     pub fn verify_proofs(
         &self,
         hasher: &dyn HashTree,
         old_root_hash: ValueHash,
         instructions: &[TreeInstruction],
-    ) {
-        assert_eq!(instructions.len(), self.logs.len());
+    ) -> bool {
```
slowli, line 25: Bikeshedding: This doesn't look idiomatic and creates unnecessary complications on the caller side. It'd be better to return a `Result`; then, you'd get `must_use` automatically, would be able to `unwrap()` in tests and handle errors as you'd like on the caller side (e.g., log them with `tracing::error!`). The crate generally defines domain-specific errors, but if you don't feel like defining them in this case, using `anyhow::Error` would be an adequate alternative. IMO, errors should also contain more context (e.g., the number of instructions and logs for the first error) to simplify debugging.

haraldh: Thinking about it, `anyhow::ensure!` with detailed information should be good enough... Errors here are not actionable anyway.

haraldh: changed

slowli (approved):

PR #1720, opened 2024-04-18
docs: fix the markdown link

haraldh:
## What ❔

fix a markdown link in documentation

## Why ❔

to make the linkchecker happy and nicely display the link

## Checklist

- [x] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [ ] Tests for the changes have been added / updated.
- [x] Documentation comments have been added / updated.
- [x] Code has been formatted via `zk fmt` and `zk lint`.
- [ ] Spellcheck has been run via `zk spellcheck`.
- [ ] Linkcheck has been run via `zk linkcheck`.

PR #1723, opened 2024-04-18
feat: make more structs serializable

haraldh:
## What ❔

Make more structs serializable.

## Why ❔

Needed for the TEE verifier as input. These structs will be serialized for intercloud data exchange.

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3763,6 +3763,7 @@ dependencies = [
  "hex",
  "itertools 0.10.5",
  "once_cell",
+ "serde",
  "thiserror",
  "tokio",
  "tracing",
--- a/core/lib/contracts/src/lib.rs
+++ b/core/lib/contracts/src/lib.rs
@@ -219,13 +219,13 @@ fn read_zbin_bytecode_from_path(bytecode_path: PathBuf) -> Vec<u8> {
         .unwrap_or_else(|err| panic!("Can't read .zbin bytecode at {:?}: {}", bytecode_path, err))
 }
 /// Hash of code and code which consists of 32 bytes words
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct SystemContractCode {
     pub code: Vec<U256>,
     pub hash: H256,
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct BaseSystemContracts {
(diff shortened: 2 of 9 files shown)


haraldh:
Failed action fixed with https://github.com/matter-labs/zksync-era/pull/1720

mm-zk:
is there any reason for this? 
Every struct that is serializable, makes it harder to do breaking changes to (for example, if I'm adding or removing the field, now i have to look to make sure that I didn't break the serialization)

haraldh:
> is there any reason for this? Every struct that is serializable, makes it harder to do breaking changes to (for example, if I'm adding or removing the field, now i have to look to make sure that I didn't break the serialization)

@mm-zk while I agree with you, it will be needed for an external running (SGX) verifier ... do you have any other suggestion on how to solve this? Either we will have to have a public constructor for those structs, or do it via serde.

haraldh:
Or we can state, that the serialization user is aware of struct changes and can cope with those by making sure the consumer is in lockstep with the producer.

mm-zk:
@haraldh - what do you mean by 'public constructor'? AFAIK in rust as long as all the fields are pub, you have implicit public constructor (Struct{field1: xx, field2: yy} )

haraldh:
> @haraldh - what do you mean by 'public constructor'? AFAIK in rust as long as all the fields are pub, you have implicit public constructor (Struct{field1: xx, field2: yy} )

True, in this case all the fields of the touched structs are public. So, we will have to make shadow structs for all of them instead, if this PR is not gonna make it.

Anyway, we are currently using `ciborium` for binary serialization and not `bincode` (which is not type safe), so we can recognize struct changes while deserializing and error out

mm-zk:
@haraldh - what do you mean by 'shadow structs'?

and do you actually serialize these objects as strings?

haraldh:
> @haraldh - what do you mean by 'shadow structs'?

as you could see in https://github.com/matter-labs/zksync-era/pull/1860

Adding Serialize, Deserialize results in removing the shadow structs, see the [core/lib/tee_verifier/src/lib.rs](https://github.com/matter-labs/zksync-era/pull/1860/commits/cd5c5aace3f129baab8c487f823e47d5d3422ef7#diff-714f108bff55c7ced6b8b97518daf0fa5d97cb2b63dc5ed2228020100eab02ee) changes in this commit


> 
> and do you actually serialize these objects as strings?

No... bincode

PR #1726, opened 2024-04-18
chore(db): Get rid of `storage` table (pt. 2)

slowli:
## What ❔

Remove mentions of the `storage` table in the codebase.

## Why ❔

The table has been deprecated for some time and should eventually be removed.

## Checklist


slowli:
Since the table was used in some circumstances (e.g., it was still updated on each L2 block and reverted together with storage logs), it's impossible to hard-remove it right away.

PR #1755, opened 2024-04-22
fix: ensure two connections for both executor and async catchup

itegulov:
## What ❔

Fixes recently discovered bug in v22.1.0 where executor and fee migration logic get starved out by async catch up task that holds a long-term connection.

## Checklist

### DIFF
--- a/core/lib/zksync_core/src/lib.rs
+++ b/core/lib/zksync_core/src/lib.rs
@@ -89,8 +89,8 @@ use crate::{
     },
     metadata_calculator::{MetadataCalculator, MetadataCalculatorConfig},
     state_keeper::{
-        create_state_keeper, MempoolFetcher, MempoolGuard, OutputHandler, SequencerSealer,
-        StateKeeperPersistence,
+        create_state_keeper, AsyncRocksdbCache, MempoolFetcher, MempoolGuard, OutputHandler,
+        SequencerSealer, StateKeeperPersistence,
     },
     utils::ensure_l1_batch_commit_data_generation_mode,
 };
@@ -828,11 +828,7 @@ async fn add_state_keeper_to_task_futures(
     batch_fee_input_provider: Arc<dyn BatchFeeModelInputProvider>,
     stop_receiver: watch::Receiver<bool>,
 ) -> anyhow::Result<()> {
-    // One (potentially held long-term) connection for async catch up task and another connection for
-    // batch executor.
-    let state_keeper_pool_builder =
-        ConnectionPool::<Core>::builder(postgres_config.master_url()?, 2);
-    let state_keeper_pool = state_keeper_pool_builder
+    let state_keeper_pool = ConnectionPool::<Core>::singleton(postgres_config.master_url()?)
         .build()
         .await
         .context("failed to build state_keeper_pool")?;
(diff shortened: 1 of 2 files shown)


core/lib/zksync_core/src/lib.rs
```diff
@@ -828,8 +828,11 @@ async fn add_state_keeper_to_task_futures(
     batch_fee_input_provider: Arc<dyn BatchFeeModelInputProvider>,
     stop_receiver: watch::Receiver<bool>,
 ) -> anyhow::Result<()> {
-    let pool_builder = ConnectionPool::<Core>::singleton(postgres_config.master_url()?);
-    let state_keeper_pool = pool_builder
+    // One (potentially held long-term) connection for async catch up task and another connection for
+    // batch executor.
+    let state_keeper_pool_builder =
+        ConnectionPool::<Core>::builder(postgres_config.master_url()?, 2);
```
perekopskiy, line 834: IMO, it would be better if we had 3 singleton pools for:
1. AsyncCatchupTask
2. AsyncRocksdbCache
3. MempoolIO

or at least allow 3 connections per shared pool

core/lib/zksync_core/src/lib.rs
```diff
@@ -828,8 +828,11 @@ async fn add_state_keeper_to_task_futures(
     batch_fee_input_provider: Arc<dyn BatchFeeModelInputProvider>,
     stop_receiver: watch::Receiver<bool>,
 ) -> anyhow::Result<()> {
-    let pool_builder = ConnectionPool::<Core>::singleton(postgres_config.master_url()?);
-    let state_keeper_pool = pool_builder
+    // One (potentially held long-term) connection for async catch up task and another connection for
+    // batch executor.
```
slowli, line 832: Are 2 connections enough? Besides the batch executor (which AFAIU holds a connection for an entire L1 batch), and the catch up task (which holds a connection while it's active), we have `MempoolIO` that also needs to acquire connections from time to time while the batch is executing. I think it could be safer to:

- Construct a separate two-connection pool specifically for `AsyncRocksdbCache` (for it, we know that 2 connections is enough)
- Refactor `create_state_keeper()` to accept `AsyncRocksdbCache` (or `Arc<dyn ReadStorageFactory>`) as an arg. Still provide a singleton pool for `MempoolIO`.

Note that some other state keeper dependencies (e.g., `OutputHandler`) are already provided as args to `create_state_keeper()`, and both miniblock sealer and mempool fetcher use separate singleton pools.

core/lib/zksync_core/src/lib.rs
itegulov: I split this pool into two as @slowli suggested: a two-sized pool for `AsyncCatchupTask/AsyncRocksdbCache ` and a separate singleton pool for MempoolIO

core/lib/zksync_core/src/lib.rs
itegulov: Done. My original thinking was that a singleton pool used to be enough for both batch executor and MempoolIO, but better safe than sorry indeed

perekopskiy (approved):

slowli (approved):

PR #1757, opened 2024-04-22
fix(snapshots_creator): Remove snapshots during reverts

slowli:
## What ❔

- Reverts snapshot metadata and GCS data for snapshots removed during reverts.
- Brushes up the reverter crate.

## Why ❔

- If the snapshots are not removed, they'll contain bogus data.
- Removing snapshot objects from GCS is not strictly necessary, but simplifies handling corner cases (e.g., creating a later snapshot with the same L1 batch number; removing a snapshot in the middle of EN recovery).

## Checklist


core/bin/block_reverter/src/main.rs
```diff
@@ -33,35 +32,35 @@ enum Command {
     /// Sends revert transaction to L1.
     #[command(name = "send-eth-transaction")]
     SendEthTransaction {
-        /// L1 batch number used to rollback to.
+        /// L1 batch number used to revert to.
```
AnastasiiaVashchuk, line 35: In my head, `rollback` is about the undoing of recent actions as if they never happened.
And `revert` is about undoing itself is recorded as a new action. 
AFAIU here is the case 1( e.g. undoing of recent actions as if they never happened)..so I am confused

core/lib/dal/src/snapshots_dal.rs
```diff
@@ -233,9 +262,36 @@ mod tests {
...
+        assert_eq!(
+            deleted_snapshots[0].storage_logs_filepaths,
+            snapshot_metadata.storage_logs_filepaths
+        );
+
+        let deleted_snapshot_metadata = dal.get_snapshot_metadata(l1_batch_number).await.unwrap();
+        assert!(
+            deleted_snapshot_metadata.is_none(),
+            "{deleted_snapshot_metadata:?}"
+        );
```
AnastasiiaVashchuk, line 294: shouldn't it be a separate test?

core/bin/block_reverter/src/main.rs
slowli: That's reasonable, but then we'd need to rename the reverter component to something like `BlockRollback` (or `BlockRollbackExecutor`?) because it doesn't do reverts. Would this make more sense to you?

AnastasiiaVashchuk: >Would this make more sense to you?

Yes. Also, there is a quick solution - I just left the doc with an explanation this part

slowli: Having [ubiquitous language](https://martinfowler.com/bliki/UbiquitousLanguage.html) (i.e., unified terminology in this case) is kind of important, that's why I've tried to eliminate a mix of "revert" / "rollback"; I think it's worse than using either of these terms consistently.

slowli: I've retained the name for now, and explained the difference between "roll back" and "revert" in the docs (and it's reflected in method naming).

core/node/block_reverter/Cargo.toml
```diff
@@ -14,12 +14,18 @@ zksync_types.workspace = true
...
+zksync_object_store.workspace = true
 zksync_storage.workspace = true
 zksync_eth_signer.workspace = true
 zksync_state.workspace = true
 zksync_merkle_tree.workspace = true
 
-tokio = { workspace = true, features = ["time"] }
-bitflags.workspace = true
+anyhow.workspace = true
+tokio = { workspace = true, features = ["time", "fs"] }
```
AnastasiiaVashchuk, line 24: seems `time` is not needed anymore

core/node/block_reverter/src/lib.rs
```diff
@@ -83,317 +77,443 @@ pub enum NodeRole {
...
+/// Reversion is a rare event of manual intervention, when the node operator
 /// decides to revert some of the not yet finalized batches for some reason
 /// (e.g. inability to generate a proof).
 ///
-/// It is also used to automatically perform a rollback on the external node
-/// after it is detected on the main node.
+/// It is also used to automatically roll back the external node state
+/// after it detects a reorg on the main node. Notice the difference in terminology; from the network
+/// perspective, *reversion* leaves a trace (L1 transactions etc.), while from the perspective of the node state, a *rollback*
+/// leaves the node in the same state as if the rolled back batches were never sealed in the first place.
```
AnastasiiaVashchuk, line 89: :+1:

core/node/block_reverter/Cargo.toml
slowli: This feature is still used [on this line](https://github.com/matter-labs/zksync-era/blob/fbd658652704d5bfba48e35873d5002a9280f9f4/core/node/block_reverter/src/lib.rs#L557).

AnastasiiaVashchuk (approved):

core/bin/block_reverter/src/main.rs
```diff
@@ -99,16 +101,16 @@ async fn main() -> anyhow::Result<()> {
             .default_priority_fee_per_gas,
     );
     let contracts = ContractsConfig::from_env().context("ContractsConfig::from_env()")?;
+    // FIXME: is it correct to parse the entire `NetworkConfig`?
```
slowli, line 104: Subj. The code in `main` reads a specific env var; is there a particular reason to it?

https://github.com/matter-labs/zksync-era/blob/884882039c54fd301be09796c8bdd783f2c866e2/core/node/block_reverter/src/lib.rs#L349-L352

core/bin/block_reverter/src/main.rs
```diff
@@ -99,16 +101,16 @@ async fn main() -> anyhow::Result<()> {
...
-        None
-    };
-    let config = BlockReverterEthConfig::new(eth_sender, contracts, operator_address.copied());
+    let era_chain_id = env::var("CONTRACTS_ERA_CHAIN_ID")
+        .context("`CONTRACTS_ERA_CHAIN_ID` env variable is not set")?
+        .parse()
+        .map_err(|err| {
+            anyhow::anyhow!("failed parsing `CONTRACTS_ERA_CHAIN_ID` env variable: {err}")
+        })?;
+    let config = BlockReverterEthConfig::new(eth_sender, &contracts, &network, era_chain_id)?;
```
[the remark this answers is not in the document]
montekki: I don't see any particular reason for it, it looks like it should have been a `L2ChainId`

core/bin/block_reverter/src/main.rs
montekki: It is a hack to manually pull a single var from what actually has to be `NetworkConfig`. The `NetworkConfig` has to be fully present on the node that we do reversion from, so it is ok to just parse it.

montekki (approved):

RomanBrodetski (approved):

PR #1760, opened 2024-04-22
perf: buffer IO for Keystore (de)serialization

Oppen:
## What ❔

Use `BufReader` and `BufWriter` rather than serializing to and
deserializing from memory, reducing one of the memory usage peaks caused
by temporaries.
For one particular case this avoids a 13GiB allocation.

## Why ❔

The CPU prover experiences high memory usage that can bring down some instances.
This is an ongoing effort to reduce sailing and peak memory usage.

## Checklist


Oppen:
> I tried testing the PR, but that's not possible as we're in the middle of 1.5.0 upgrade. Will follow up once the upgrade is done.

If possible, comment when that's done if you need any update to the code. I check Github notifications so I'll see it.

Re: warnings: of course, I'll fix those soon.

Oppen:
Rebased and fixed the warnings.

PR #1774, opened 2024-04-23
Kl factory optimise init

Raid5594:
## Checklist


core/tests/ts-integration/scripts/build.ts
```diff
@@ -0,0 +1,29 @@
+import path from 'path';
+import { exec } from 'child_process';
+import { needsRecompilation, setCompilationTime, isFolderEmpty } from './utils';
+
+const CONTRACTS_DIR = 'contracts';
+const OUTPUT_DIR = 'artifacts-zk';
+const TIMESTAMP_FILE = 'last_compilation.timestamp'; // File to store the last compilation time
```
kelemeno, line 7: I think is used in multiple places, maybe you can have them in utils and import them?

kelemeno:
Also kl-factory has been merged, please rebase on and merge main

infrastructure/zk/src/server.ts
```diff
@@ -46,20 +45,7 @@ export async function externalNode(reinit: boolean = false, args: string[]) {
 
 async function create_genesis(cmd: string) {
     await utils.confirmAction();
-    await utils.spawn(`${cmd} | tee genesis.log`);
-
-    const date = new Date();
```
kelemeno, line 51: Should we not keep this?

Raid5594: There was a discussion where it seems that we're not using these logs anymore

core/tests/ts-integration/scripts/build.ts
kelemeno: Yes, very good idea

infrastructure/zk/src/server.ts
Raid5594: @mm-zk is it fine if we remove these log from here?

mm-zk: AFAIK yes. @Deniallugo - FYI.

core/tests/ts-integration/scripts/build.ts
```diff
@@ -0,0 +1,32 @@
...
+import {
+    needsRecompilation,
+    setCompilationTime,
+    isFolderEmpty,
+    CONTRACTS_DIR,
+    OUTPUT_DIR,
+    TIMESTAMP_FILE
+} from './utils';
+
+async function main() {
```
mm-zk, line 12: please add comment - describing what we're doing here (using some .timestamp file to record the last time that files were changed)

core/tests/ts-integration/scripts/utils.ts
```diff
@@ -0,0 +1,87 @@
...
+
+// Write the current time to the timestamp file
+export function setCompilationTime(timestampFile: string) {
+    fs.writeFileSync(timestampFile, Date.now().toString());
+}
+
+// Determine if recompilation is needed
+export function needsRecompilation(folder: string, timestampFile: string): boolean {
+    const lastCompilationTime = getLastCompilationTime(timestampFile);
+    const latestModificationTime = getLatestModificationTime(folder);
```
mm-zk, line 56: do these sources have any dependencies on outside dirs?

core/tests/ts-integration/scripts/build.ts
Raid5594: +

core/tests/ts-integration/scripts/utils.ts
Raid5594: +

core/tests/ts-integration/package.json
```diff
@@ -9,8 +9,8 @@
     "fee-test": "RUN_FEE_TEST=1 zk f jest -- fees.test.ts",
     "api-test": "zk f jest -- api/web3.test.ts api/debug.test.ts",
     "contract-verification-test": "zk f jest -- api/contract-verification.test.ts",
-    "build": "hardhat compile",
-    "build-yul": "hardhat run scripts/compile-yul.ts"
+    "build": "yarn ts-node scripts/build.ts",
```
mm-zk, line 12: nit: do you actually have to say 'yarn' here? I thought that "ts-node XX" should be enough.

core/tests/ts-integration/scripts/build.ts
mm-zk: seems that it was not done.
I'm looking for a 2-line comment, where reader can quickly see what's special about this 'build' script, without having to read the details of the code.

etc/contracts-test-data/scripts/utils.ts
```diff
@@ -0,0 +1,159 @@
+import * as fs from 'fs';
```
mm-zk, line 1: you are repeating the same file twice, that's not good, as it will be harder to manage and update in the future.

Could it be some shared library instead?

etc/contracts-test-data/scripts/build.ts
```diff
@@ -0,0 +1,33 @@
...
+    setCompilationTime,
+    isFolderEmpty,
+    CONTRACTS_DIR,
+    OUTPUT_DIR,
+    TIMESTAMP_FILE
+} from './utils';
+import { exec } from 'child_process';
+
+async function main() {
+    const timestampFilePath = path.join(process.cwd(), TIMESTAMP_FILE); // File stores the timestamp of last compilation
```
mm-zk, line 13: you are repeating this twice (as it already exists in the other build.rs) - maybe move it to a shared util instead?

mm-zk (changes requested):

etc/contracts-test-data/scripts/build.ts
Raid5594: That's right, I was thinking of it, but as these are different workspaces, I was not sure if a shared util is a good idea

core/tests/ts-integration/scripts/build.ts
Raid5594: updated

Deniallugo: I'm not sure it's the right move. 
We could solve in two ways: 
1. Move l2 contracts compilation to foundry zksync 
2. Use the prebuild contracts from the repo. we do pulbish built contracts

infrastructure/zk/src/server.ts
Deniallugo: i don't think we need this function at all

PR #1782, opened 2024-04-23
feat: (txsink) add tx sink deny list implementation and add new server component for it 

JayT106:
## What ❔

- Add a new txsink implementation for rejecting the L2 transaction by filtering the sender's address in the API server. 
- New tx sink env config for the tx sink component read.
- Define new `TxSinkDenyList` component and can be added when launch the server.

For running server with the deny list tx sink, should execute
```
zk server --components api,tree,eth,state_keeper,housekeeper,commitment_generator,txsink_denylist
```
And in the env files (i.e. dev.env) should add `TX_SINK_DINY_LIST="deny addresses separated by comma"` 

## Why ❔

We need a mechanism to be able to block some addresses has potential attack/hack risks through L2 transactions, i.e. blocking fund withdraw from L2 to L1 or other factors.

## Checklist


core/lib/zksync_core/src/lib.rs
```diff
@@ -1294,11 +1330,17 @@ async fn run_http_api(
     with_debug_namespace: bool,
     storage_caches: PostgresStorageCaches,
     mempool_cache: MempoolCache,
+    tx_sink_config: Option<TxSinkConfig>,
 ) -> anyhow::Result<()> {
+    let config = TxSenderBuilderConfigs {
+        tx_sender_config: tx_sender_config.clone(),
```
devashishdxt, line 1336: Why `clone`?

JayT106: To construct the `TxSenderBuilderConfigs` takes the ownership of the value in `tx_sender_config` (it's a reference). So the compiler requires `clone` the value. 

If there any better way to construct the struct and assign it to `build_tx_sender`? I made this change  because of the linting error (too many arguments in build_tx_sender)

PR #1789, opened 2024-04-25
feat(Prover CLI): Batch proof compression job status

ilitteri:
## What ❔

Handles the batch proof compression status for the `status batch` command.

## Why ❔

A user that will do maintenance on the prover could find useful to see the batch proof compression job status.

## Example

```
cd prover/prover_cli
zk f cargo run --release  status batch -n 1
```

## Checklist

### DIFF
--- a/core/lib/basic_types/src/prover_dal.rs
+++ b/core/lib/basic_types/src/prover_dal.rs
@@ -1,7 +1,8 @@
 //! Types exposed by the prover DAL for general-purpose use.
 use std::{net::IpAddr, ops::Add, str::FromStr};
 
-use chrono::{DateTime, Duration, Utc};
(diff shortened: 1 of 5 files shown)


Artemka374 (commented):
LGTM, like the emojis, hehe

core/lib/basic_types/src/prover_dal.rs
```diff
@@ -229,3 +230,33 @@ impl FromStr for GpuProverInstanceStatus {
         }
     }
 }
+
+#[derive(Debug, EnumString, Display)]
+pub enum ProofCompressionJobStatus {
```
Artemka374, line 235: Isn't there a similar enum in types? May we reuse it here?

ilitteri: Yes! This is the one that you are talking about, I just moved it to the right module.

PR #1796, opened 2024-04-25
chore(validium): fix validium init and server

koloz193:
## Checklist

### DIFF
--- a/core/lib/zksync_core/src/utils/mod.rs
+++ b/core/lib/zksync_core/src/utils/mod.rs
@@ -172,7 +172,7 @@ async fn get_pubdata_pricing_mode(
 ) -> Result<Vec<ethabi::Token>, EthClientError> {
     let args = CallFunctionArgs::new("getPubdataPricingMode", ()).for_contract(
         diamond_proxy_address,
-        zksync_contracts::state_transition_manager_contract(),
+        zksync_contracts::hyperchain_contract(),
     );
     eth_client.call_contract_function(args).await
 }


.github/workflows/ci-core-reusable.yml
```diff
@@ -190,8 +191,14 @@ jobs:
...
-          ci_run zk config compile
-          ci_run zk init ${{ matrix.base_token == 'Custom' && '--base-token-name BAT' || ''}}
+          if [[ "${{ matrix.deployment_mode }}" == "Validium" ]]; then
+            ci_run zk env dev_validium_docker
+            ci_run zk config compile dev_validium_docker
+            ci_run zk init --validium-mode
+          else
+            ci_run zk config compile
+          fi
+          ci_run zk init ${{ matrix.base_token == 'Custom' && '--base-token-name BAT' || ''}} ${{ matrix.deployment_mode == 'Validium' && '--validium-mode' || ''}}
```
ilitteri, line 201: Careful here, we're performing `zk init` twice.

.github/workflows/ci-core-reusable.yml
```diff
@@ -301,8 +313,13 @@ jobs:
...
           ci_run zk
-          ci_run zk config compile
-          ci_run zk init ${{ matrix.base_token == 'Custom' && '--base-token-name BAT' || ''}}
+          if [[ "${{ matrix.deployment_mode }}" == "Rollup" ]]; then
+            ci_run zk config compile
+          elif [[ "${{ matrix.deployment_mode }}" == "Validium" ]]; then
+            ci_run zk env dev_validium_docker
+            ci_run zk config compile dev_validium_docker
+          fi
+          ci_run zk init ${{ matrix.base_token == 'Custom' && '--base-token-name BAT' || ''}} ${{ matrix.deployment_mode == 'Validium' && '--validium-mode' || ''}}
```
ilitteri, line 322: Ditto.

.github/workflows/ci-core-reusable.yml
```diff
@@ -124,14 +124,15 @@ jobs:
           ci_run cat /tmp/sccache_log.txt
 
   integration:
-    name: Integration (consensus=${{ matrix.consensus }}, base_token=${{ matrix.base_token }})
+    name: Integration (consensus=${{ matrix.consensus }}, base_token=${{ matrix.base_token }}), deployment_mode=${{ matrix.deployment_mode }})
```
ilitteri, line 127:
```suggestion
    name: Integration (consensus=${{ matrix.consensus }}, base_token=${{ matrix.base_token }}, deployment_mode=${{ matrix.deployment_mode }})
```

core/lib/l1_contract_interface/src/i_executor/structures/commit_batch_info.rs
ilitteri, line 1: Why do we stop sending the commitment to the pubdata to the L1?

ilitteri (commented):
The PR looks good in general. Left some comments with suggestions, nits, and spotted some errors, and questions.

I'm trying to run this locally like this
```
zk && zk clean --all && zk env dev_validium && zk init --validium mode
```
and I'm having these errors:

**zk init**
1. While running the server in genesis (the same errors is [in the CI](https://github.com/matter-labs/zksync-era/actions/runs/8850782925/job/24305894473?pr=1796#step:6:2457))
    ```
    Error: Failed to save SetChainId upgrade transaction
    Caused by:
        Expected a single set_chain_id event, got these 0: []
    ````
2. Error: TypeError: diamondProxy.setValidiumMode is not a function

**server**
`We can't operate after tx fail` -> `revert_reason: "revert: pl"`

I think that the server error is due to the error 2 I'm having during the `init`

ilitteri:
> I'm trying to run this locally like this
> 
> ```
> zk && zk clean --all && zk env dev_validium && zk init --validium mode
> ```
> 
> and I'm having these errors:
> 
> **zk init**
> 
>     1. While running the server in genesis (the same errors is [in the CI](https://github.com/matter-labs/zksync-era/actions/runs/8850782925/job/24305894473?pr=1796#step:6:2457))
>        ```
>        Error: Failed to save SetChainId upgrade transaction
>        Caused by:
>            Expected a single set_chain_id event, got these 0: []
>        ```
> 
>     2. Error: TypeError: diamondProxy.setValidiumMode is not a function
> 
> 
> **server** `We can't operate after tx fail` -> `revert_reason: "revert: pl"`
> 
> I think that the server error is due to the error 2 I'm having during the `init`

Update:

- `Error: Failed to save SetChainId upgrade transaction` is expected because there's a contract that is not deployed yet.
- `Error: TypeError: diamondProxy.setValidiumMode is not a function` and `We can't operate after tx fail` -> `revert_reason: "revert: pl"` are fixed now.

infrastructure/zk/src/contract.ts
```diff
@@ -130,15 +130,11 @@ export async function deployL2ThroughL1({
     );
 }
 
-async function _deployL1(onlyVerifier: boolean, deploymentMode: DeploymentMode): Promise<void> {
+async function _deployL1(onlyVerifier: boolean, _deploymentMode: DeploymentMode): Promise<void> {
```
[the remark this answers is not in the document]
koloz193: no we can actually remove it in most places

infrastructure/zk/src/contract.ts
```diff
@@ -130,15 +130,11 @@ export async function deployL2ThroughL1({
...
     await utils.confirmAction();
 
     const privateKey = process.env.DEPLOYER_PRIVATE_KEY;
-    const args = [
-        privateKey ? `--private-key ${privateKey}` : '',
-        onlyVerifier ? '--only-verifier' : '',
-        deploymentMode == DeploymentMode.Validium ? '--validium' : ''
-    ];
+    const args = [privateKey ? `--private-key ${privateKey}` : '', onlyVerifier ? '--only-verifier' : ''];
```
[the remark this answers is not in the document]
koloz193: validium mode is set as part of the register-hyperchain https://github.com/matter-labs/era-contracts/blob/b5314f657748bfa6313f1e43b5753475d59fa5ad/l1-contracts/src.ts/deploy.ts#L747

.github/workflows/ci-core-reusable.yml
koloz193: i dont think we init here again?

core/lib/l1_contract_interface/src/i_executor/structures/commit_batch_info.rs
koloz193: so that info is either the actual pubdata when we send via calldata or its the info needed for the blobs when using blobs. for validium, its sent using neither and so we don't need any info in there. the pubdata is already committed to by the system logs and will be proven inherently

.github/workflows/ci-core-reusable.yml
ilitteri: You're right!

ilitteri (approved):
LGTM. Validium seems to be working correctly, I've tested the changes locally. It can be merged as it is, but we'll need to track the addition of this `if` https://github.com/matter-labs/zksync-era/pull/1796/files#diff-8f8288bbb19c9bdae93f8ef9f09bbcf8b2d6a7c2141fd701b0d9e89d9c4cd226R11, as originally it was agreed with @popzxc and @RomanBrodetski not to use any boolean flag for Validium.

core/node/commitment_generator/src/input_generation.rs
```diff
@@ -0,0 +1,40 @@
...
+
+use zksync_types::{commitment::CommitmentInput, H256};
+
+#[derive(Debug)]
+pub struct ValidiumInputGenerator;
+
+#[derive(Debug)]
+pub struct RollupInputGenerator;
+
+pub trait InputGenerator: 'static + fmt::Debug + Send + Sync {
```
popzxc, line 11: Having some documentation explaining the purpose, e.g. what kind of input is being calculated here, and why do we both accept and return input (e.g. based on the signature alone it feels like input post-processor, idk), would be appreciated :)

core/bin/external_node/src/main.rs
```diff
@@ -337,14 +340,18 @@ async fn run_core(
...
+    ) = match config.optional.l1_batch_commit_data_generator_mode {
+        L1BatchCommitDataGeneratorMode::Rollup => (
+            Arc::new(RollupModeL1BatchCommitDataGenerator {}),
+            Box::new(RollupInputGenerator),
+        ),
+        L1BatchCommitDataGeneratorMode::Validium => (
+            Arc::new(ValidiumModeL1BatchCommitDataGenerator {}),
+            Box::new(ValidiumInputGenerator),
+        ),
     };
```
popzxc, line 355: Side comment: I'm starting to feel that we need to revisit all the differences we have between rollup and validium, and probably have a single abstraction that is capable of handling all the differences (right now I'm not sure if that's possible, just thinking out loud).

The current situation is still strictly better than having `if`s here and there, since we instantiate all the stuff in a single place and the underlying components don't have to think about it, but still it somewhat bugs me that we in theory can have rollup commit data generator with validium input generator.

No action required here.

core/node/node_framework/examples/main_node.rs
```diff
@@ -334,7 +337,17 @@ impl MainNodeBuilder {
...
-        self.node.add_layer(CommitmentGeneratorLayer);
+        let genesis = GenesisConfig::from_env()?;
+        let input_generator: Box<dyn InputGenerator> = if genesis
+            .l1_batch_commit_data_generator_mode
+            == L1BatchCommitDataGeneratorMode::Validium
+        {
+            Box::new(ValidiumInputGenerator)
+        } else {
+            Box::new(RollupInputGenerator)
+        };
```
popzxc, line 348: Strictly speaking, this is not idiomatic for the framework. An idiomatic way would be to have input generator as a resource, or at least pass `L1BatchCommitDataGeneratorMode` to the layer and instantiate the input generator inside (like it's done in the [eth_sender layer](https://github.com/matter-labs/zksync-era/blob/main/core/node/node_framework/src/implementations/layers/eth_sender.rs#L93)).

Overall, it's the layers who are responsible for interaction with concrete objects, and the point of the framework is that you don't do any kind of wiring _outside_ of layers.

core/node/commitment_generator/Cargo.toml
```diff
@@ -20,13 +20,15 @@ multivm.workspace = true
 circuit_sequencer_api_1_4_0.workspace = true
 circuit_sequencer_api_1_4_1.workspace = true
 circuit_sequencer_api_1_5_0.workspace = true
+zksync_env_config.workspace = true
```
popzxc, line 23: Is it used anywhere?

core/node/commitment_generator/src/lib.rs
```diff
@@ -1,6 +1,7 @@
 use std::time::Duration;
 
 use anyhow::Context;
+use input_generation::InputGenerator;
```
popzxc, line 4: Nit: Please use `crate::input_generation` here instead to clearly indicate that we're using local module and not external crate.

core/node/commitment_generator/Cargo.toml
```diff
@@ -20,13 +20,15 @@ multivm.workspace = true
...
 
 zk_evm_1_5_0.workspace = true
 zk_evm_1_4_1.workspace = true
 zk_evm_1_3_3.workspace = true
 
 tokio = { workspace = true, features = ["time"] }
 anyhow.workspace = true
+async-trait.workspace = true
```
popzxc, line 31: Is it used anywhere?

core/node/commitment_generator/src/input_generation.rs
koloz193: i figured there were 2 ways to go about it:
1. take the input generated and post process it
2. take the raw data and generate the input

i think in the long run we likely will want to do something closer to (2) but i wasn't sure if this needed to be done for the current use case. definitely will add some documentation though

core/bin/external_node/src/main.rs
koloz193: i agree, i thought about forking the commitment_generator folder for validiums but i didnt want to redo what we have for multivms (introduces bloat) and the even within validiums the difference will likely only change the single `pubdataCommitments` field on `CommitBatchInfo`.

If we changed the commitment_generator to be the base of the data and then the input generator took in the raw data and computed the actual input then maybe that seems like a better pipeline. wdyt?

core/node/commitment_generator/Cargo.toml
koloz193: it was originally, ill remove

core/node/commitment_generator/Cargo.toml
koloz193: it was originally, ill remove

StanislavBreadless (approved):

core/bin/external_node/src/main.rs
popzxc: I think it's fine to keep it as is for now and revisit a bit later.

popzxc (dismissed):

StanislavBreadless (approved):

popzxc (approved):

hatemosphere (approved):

PR #1797, opened 2024-04-25
feat(node): Extract genesis into separate crate

AnastasiiaVashchuk:
## What ❔

Extract utilities needed to perform genesis into a dedicated crate.

## Why ❔

Part of splitting `zksync_core` into a set of crates.

## Checklist


core/node/shared/src/genesis.rs
```diff
@@ -186,6 +184,44 @@ pub struct GenesisBatchParams {
     pub rollup_last_leaf_index: u64,
 }
 
+// TODO: add metrics to track latency of each stage
```
AnastasiiaVashchuk, line 187: Well, so (here)[https://github.com/matter-labs/zksync-era/blob/31db44f31316d1e838f87e6f8cf9412ce0d6c9e4/core/lib/zksync_core/src/metadata_calculator/helpers.rs#L436] we also record some metrics. Logically, they have to be here as well...but then there are two ways:
1) move `LoadChangesStage` somewhere else.
2) create another `genesis` metrics with a similar purpose(I do not like this idea)

Cargo.toml
```diff
@@ -230,3 +232,4 @@ zksync_shared_metrics = { path = "core/node/shared_metrics" }
 zksync_block_reverter = { path = "core/node/block_reverter" }
 zksync_commitment_generator = { path = "core/node/commitment_generator" }
 zksync_house_keeper = { path = "core/node/house_keeper" }
+zksync_shared = { path = "core/node/shared" }
```
slowli, line 235: Bikeshedding: IMO, `zksync_shared` is too generic; `zksync_node_shared` would be a better name. The crate directory could still be called `shared` to avoid duplication.

core/node/shared/src/genesis.rs
slowli: I believe it's acceptable to not track metrics for the genesis batch; it's a one-time action.

core/node/shared/src/genesis.rs
```diff
@@ -186,6 +184,44 @@ pub struct GenesisBatchParams {
...
+        .get_touched_slots_for_l1_batch(l1_batch_number)
+        .await
+        .context("cannot fetch touched slots")?;
+
+    let hashed_keys_for_writes: Vec<_> = touched_slots.keys().map(StorageKey::hashed_key).collect();
+    let l1_batches_for_initial_writes = storage
+        .storage_logs_dal()
+        .get_l1_batches_and_indices_for_initial_writes(&hashed_keys_for_writes)
+        .await
+        .context("cannot fetch initial writes batch numbers and indices")?;
```
slowli, line 204: I believe loading `l1_batches_for_initial_writes` and the filtering logic later is not necessary. All non-zero touched slots should be written. Come to think of it, we do get `deduplicated_writes` [during genesis batch initialization](https://github.com/matter-labs/zksync-era/blob/9c6ed838ffcfe6bd8fe157c69fbfe8823826849b/core/lib/zksync_core/src/genesis.rs#L435); it would be optimal just to pass these around and convert them to `TreeInstruction`s.

AnastasiiaVashchuk:
Key moments:
+ `zlsync_core/genesis.rs` content is chunked into 3  logical modules now: `utils.rs`, `tests.rs`, and `mod.rs`
+

Cargo.toml
```diff
@@ -16,6 +16,8 @@ members = [
     "core/node/block_reverter",
     "core/node/commitment_generator",
     "core/node/house_keeper",
+    "core/node/genesis",
```
popzxc, line 19: Initially this PR introduced `zksync_node_shared` crate, but I changed it to be simply genesis.
Reasoning:
1. We have a tracked history of turning non-specific "shared stuff" crates into a mess, I don't really want us to have one more point of concentration for random stuff.
2. Genesis is an important part of the ZK Stack node, so it seems to be justified to have its own crate.

Deniallugo (approved):
please run
`zk run genesis-config-generator --check`
if it's not failing i'm ok with this change

Deniallugo (approved):

PR #1799, opened 2024-04-25
chore: remove storage_logs field from ExecutionState

joonazan:
Before this change CurrentExecutionState contained deduplicated storage logs and the raw storage logs. However, the fast VM cannot produce the raw storage logs, as those are an implementation detail. Luckily it seems that they aren't needed, at least for out-of-circuit use.


joonazan:
I expect that compiling prover will fail but let's see.

joonazan:
CI for prover was skipped, so prover doesn't depend on multivm. In that case this change shouldn't cause any problems.

perekopskiy (approved):

perekopskiy:
@joonazan we need storage logs for each L2 block (deduplicated too but within each L2 block), is new VM capable of producing those?

PR #1801, opened 2024-04-25
feat(state-keeper): Parallel l2 block sealing

perekopskiy:
## What ❔

L2 block data (tx results, storage logs, events, l2 to l1 logs, factory deps, tokens) is sealed in parallel using concurrent DB connections. L2 block header is inserted after all l2 block data is inserted and serves as a marker that l2 block is sealed. DB queries are adjusted so L2 block is not accessed before block is sealed. Note, that some of the queries are only queried for resolved block range, i.e. caller never tries to access "pending" data. I haven't adjusted such queries on purpose to keep their performance the same.

## Why ❔

Improves L2 block sealing performance

## Checklist


core/lib/dal/src/models/storage_sync.rs
```diff
@@ -55,7 +55,7 @@ impl TryFrom<StorageSyncBlock> for SyncBlock {
                     .try_into()
                     .decode_column("l1_batch_number")?,
             ),
-            last_in_batch: block.last_batch_miniblock == Some(block.number),
+            last_in_batch: block.tx_count == 0,
```
perekopskiy, line 58: Fictive l2 block sealing and L1 batch sealing are not happening in the same transaction anymore, so the only way to determine if l2 block is the last in batch is to check if it has 0 txs.

core/lib/zksync_core/src/state_keeper/io/seal_logic/mod.rs
```diff
@@ -309,154 +319,74 @@ impl L2BlockSealCommand {
...
             total_tx_count = l1_tx_count + l2_tx_count,
-            event_count = self.l2_block.events.len()
+            event_count = this.l2_block.events.len()
         );
 
-        let definite_vm_version = self
+        if this.pre_insert_txs {
+            let mut connection = pool.connection_tagged("state_keeper").await?;
+            let progress = L2_BLOCK_METRICS.start(L2BlockSealStage::PreInsertTxs, is_fictive);
+            this.insert_transactions(&mut connection)
```
perekopskiy, line 345: I'm going to merge `pre_insert_txs` and `mark_txs_as_executed_in_l2_block` in a separate PR.

core/lib/dal/src/factory_deps_dal.rs
```diff
@@ -56,19 +56,26 @@ impl FactoryDepsDal<'_, '_> {
...
                 bytecode
             FROM
                 factory_deps
+                LEFT JOIN miniblocks ON miniblocks.number = factory_deps.miniblock_number
+                LEFT JOIN snapshot_recovery ON snapshot_recovery.miniblock_number = factory_deps.miniblock_number
             WHERE
                 bytecode_hash = $1
+                AND (
+                    miniblocks.number IS NOT NULL
+                    OR snapshot_recovery.miniblock_number IS NOT NULL
```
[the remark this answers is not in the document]
perekopskiy: Snapshot recovery saves all historical factory deps and storage logs with `miniblock_number = snapshot_recovery.miniblock_number`, they are all sealed

core/lib/dal/src/storage_web3_dal.rs
```diff
@@ -69,17 +69,37 @@ impl StorageWeb3Dal<'_, '_> {
...
     /// Gets the current value for the specified `key`.
     pub async fn get_value(&mut self, key: &StorageKey) -> DalResult<H256> {
-        self.get_historical_value_unchecked(key, L2BlockNumber(u32::MAX))
+        let Some(l2_block_number) = self
+            .storage
+            .blocks_dal()
+            .get_sealed_l2_block_number()
+            .await?
+        else {
+            return Ok(H256::zero());
```
slowli, line 78: This doesn't work with snapshot recovery, does it? Ditto for `get_values` below. I think this is acceptable because `*web3_dal` methods should be called once there's at least one local L1 batch, but still IMO this should be at least documented, and returning an error doesn't seem to have downsides. (Or you would use something like resolving `BlockNumber::Pending` and subtracting one, possibly at the call sites, so that these methods do work with snapshot recovery.)

core/lib/zksync_core/src/state_keeper/io/seal_logic/l2_block_seal_subtasks.rs
```diff
@@ -0,0 +1,584 @@
...
+    ) -> anyhow::Result<()> {
+        let subtasks = Self::all_subtasks();
+        let handles: Vec<_> = subtasks
+            .into_iter()
+            .map(|subtask| {
+                subtask
+                    .spawn(command.clone(), pool.clone())
+                    .map(|res| res.unwrap_or_else(|err| Err(err.into())))
+            })
+            .collect();
```
slowli, line 48: Nit: Since `try_join_all()` accepts iterator inputs, you don't need to `collect` handles.

core/lib/zksync_core/src/state_keeper/io/seal_logic/l2_block_seal_subtasks.rs
```diff
@@ -0,0 +1,584 @@
...
+}
+
+/// An abstraction that represents l2 block seal sub-task that can be run in parallel with other sub-tasks.
+#[async_trait::async_trait]
+pub(crate) trait L2BlockSealSubtask: Send + Sync + 'static {
+    /// Returns sub-task name.
+    fn name(&self) -> &'static str;
+
+    /// Spawns tokio task that runs the seal process.
+    fn spawn(
```
slowli, line 62: Dumb question: Why use "spawn" semantics instead of "run" ones (i.e., don't create a Tokio task, just have an async method)?

core/lib/zksync_core/src/state_keeper/io/seal_logic/l2_block_seal_subtasks.rs
```diff
@@ -0,0 +1,584 @@
...
+        tokio::spawn(async move {
+            let progress = L2_BLOCK_METRICS.start(
+                L2BlockSealStage::MarkTransactionsInL2Block,
+                command.is_l2_block_fictive(),
+            );
+
+            let mut connection = pool
+                .connection_tagged("state_keeper")
+                .await
+                .context(self.name())?;
```
slowli, line 99: Bikeshedding: IMO, it'd make more sense to add context to the call site of `spawn()` (or the equivalent `run()` method; see the other comment), so that it's clear that it'll be added to all errors uniformly, and requires less code.

core/lib/zksync_core/src/sync_layer/external_io.rs
```diff
@@ -141,6 +141,7 @@ impl StateKeeperIO for ExternalIO {
             cursor.next_l2_block,
         );
 
+        clear_pending_l2_block(&mut storage, cursor.next_l2_block - 1).await?;
```
slowli, line 144: Could this be placed in `StateKeeperPersistence` (namely, `StateKeeperOutputHandler::initialize()` impl)? If so, it would be more local (= easier to reason about) and wouldn't require adding to both `StateKeeperIO` impls and sort of breaking abstraction boundaries.

slowli (commented):
Looking at all additional filtering needed to beI wonder whether a potential alternative with primitive L2 block lifecycle would be easier. That is:

- Insert an L2 block header *before* other data (meaning that foreign keys don't need to be removed)
- ...but mark the L2 block as incomplete until all other data is inserted. Having a new `miniblocks` field looks like an overkill / impractical; another way is to record the latest finished L2 block in a singleton table.

On the one hand, this would allow retaining FK constraints and make filtering out incomplete L2 blocks more explicit. OTOH, it'd require filtering out throughout *all* queries touching L2 blocks, which is easy to screw up.

RomanBrodetski:
Didn't have a deep loop but wanted to respond to @slowli 's comment on

> Insert an L2 block header before other data (meaning that foreign keys don't need to be removed)

IMO it might be even beneficial to remote foreign keys - this will make insertion even faster.

perekopskiy:
My thoughts on alternative approach with inserting header first:
IMO foreign keys don't add much value. It's more important to keep data flow more convenient to use. As @slowli said, if we insert header first then we would need to add filters to (almost?) any query that reads from `miniblocks`. So more queries should be adjusted in addition to those touched in this PR

core/lib/dal/src/blocks_web3_dal.rs
```diff
@@ -501,6 +517,7 @@ impl BlocksWeb3Dal<'_, '_> {
             FROM
                 call_traces
                 INNER JOIN transactions ON tx_hash = transactions.hash
+                INNER JOIN miniblocks ON miniblocks.number = transactions.miniblock_number
```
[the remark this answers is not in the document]
perekopskiy: I've rechecked, in this particular query it's not needed as miniblock number is de-facto resolved above when querying protocol version. I will also check other queries and add comments if needed

core/lib/dal/src/storage_logs_dal.rs
```diff
@@ -298,17 +246,23 @@ impl StorageLogsDal<'_, '_> {
...
+                LEFT JOIN snapshot_recovery ON snapshot_recovery.miniblock_number = storage_logs.miniblock_number
             WHERE
-                hashed_key = ANY ($1)
-                AND miniblock_number <= $2
+                storage_logs.hashed_key = ANY ($1)
+                AND storage_logs.miniblock_number <= $2
+                AND (
+                    miniblocks.number IS NOT NULL
+                    OR snapshot_recovery.miniblock_number IS NOT NULL
+                )
```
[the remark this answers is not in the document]
perekopskiy: `storage_logs.miniblock_number <= (SELECT MAX(number) FROM miniblocks)` won't work because some queries that have this construction are used in state keeper and miniblocks table is empt right after snapshot recovery. However, I can change it to smth like 
```sql
storage_logs.miniblock_number <= (SELECT MAX(number) FROM miniblocks)
OR
storage_logs.miniblock_number <= (SELECT miniblock_number FROM snapshot_recovery)
```
Do you think it will be better?

core/lib/dal/src/storage_web3_dal.rs
perekopskiy: Agree, returning error makes sense

core/lib/zksync_core/src/state_keeper/io/seal_logic/l2_block_seal_subtasks.rs
perekopskiy: No reason, I agree that with "run" it will be cleaner

core/lib/zksync_core/src/sync_layer/external_io.rs
perekopskiy: So, I put it into IO::initialize because it's initialized first.
```rust
async fn run_inner(&mut self) -> Result<Infallible, Error> {
        let (cursor, pending_batch_params) = self.io.initialize().await?;
        self.output_handler.initialize(&cursor).await?;
        ...
```
Both IOs and StateKeeperPersistence in the current form require an absence of pending batch to work properly. We can change but that means adapting more SQL queries so I think current approach is better.
It doesn't look like a good solution for `StateKeeperOutputHandler` to rely on the fact that 1) IO is initialized first 2) IO cleans pending block. Maybe it's worth to run `clear_pending_l2_block` in both places, WDYT?

slowli: I see. IMO, adapting SQL queries is not worth the effort; I'm fine with the `clear_pending_l2_block()` in the output handler (or ensuring that a pending L2 block doesn't exist) in the short term. Abstractly, this issue makes me wonder whether distribution of responsibilities between `StateKeeperIO` and `StateKeeperOutputHandler` is chosen correctly; maybe, it would make sense to initialize the output handler first and make it return `IoCursor` that would then be provided to `StateKeeperIO`?

core/lib/dal/src/storage_logs_dal.rs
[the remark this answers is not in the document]
slowli: Slightly better, but still doesn't look very good. I guess it could be possible to use `COALESCE` to collapse both sub-`SELECT`s, but it still would look messy.

core/lib/zksync_core/src/sync_layer/external_io.rs
perekopskiy: hmm, I'm not sure about distribution. IMO it's ok to set one of these abstractions to be responsible for clean-up and initialize it first + document. For now will do clean-up in both

core/lib/zksync_core/src/api_server/tx_sender/proxy.rs
```diff
@@ -65,10 +67,28 @@ impl TxCache {
...
         pool: ConnectionPool<Core>,
-        stop_receiver: watch::Receiver<bool>,
+        mut stop_receiver: watch::Receiver<bool>,
     ) -> anyhow::Result<()> {
         const UPDATE_INTERVAL: Duration = Duration::from_secs(1);
 
+        tracing::info!(
+            "Waiting for at least one L1 batch in Postgres to start TxCache::run_updates"
+        );
+        // Starting the server before L1 batches are present in Postgres can lead to some invariants the server logic
```
slowli, line 77: Nit: It's not the "server" being started here, but rather transaction cache updater.

core/lib/zksync_core/src/state_keeper/io/common/mod.rs
```diff
@@ -107,3 +107,16 @@ pub(crate) async fn load_pending_batch(
         pending_l2_blocks,
     })
 }
+
+/// Clears pending l2 block data from the database.
+pub(crate) async fn clear_pending_l2_block(
```
slowli, line 112: Bikeshedding: I'd place this as a static method in `L2BlockSealProcess`; seems to logically belong there.

core/lib/zksync_core/src/state_keeper/io/seal_logic/l2_block_seal_subtasks.rs
```diff
@@ -0,0 +1,542 @@
...
+
+    pub async fn run_subtasks(
+        command: Arc<L2BlockSealCommand>,
+        pool: ConnectionPool<Core>,
+    ) -> anyhow::Result<()> {
+        let subtasks = Self::all_subtasks();
+        let handles = subtasks.into_iter().map(|subtask| {
+            let command = command.clone();
+            let pool = pool.clone();
+            tokio::spawn(async move {
```
slowli, line 43: Spawning tasks looks redundant; you can create non-`'static` futures in the iterator and join them:

```rust
pub async fn run_subtasks(
    command: &L2BlockSealCommand,
    pool: &ConnectionPool<Core>,
) -> anyhow::Result<()> {
    let subtasks = Self::all_subtasks();
    let handles = subtasks.into_iter().map(|subtask| {
        let subtask_name = subtask.name();
        async move { subtask.run(command, pool).await.context(subtask_name) }
    });
    futures::future::try_join_all(handles).await?;
    Ok(())
}
```

(This assumes that `L2BlockSealSubtask` is changed correspondingly to take `command` / `pool` by a reference.)

core/lib/zksync_core/src/state_keeper/io/seal_logic/mod.rs
```diff
@@ -61,7 +67,7 @@ impl UpdatesManager {
             false, // fictive L2 blocks don't have txs, so it's fine to pass `false` here.
         );
         l2_block_command
-            .seal_inner(&mut transaction, true)
+            .seal_inner(pool.clone(), true)
```
slowli, line 70: Could there be any issues because of non-atomicity of sealing a fictive L2 block and the corresponding L1 batch? E.g., would replaying pending L2 blocks still work correctly? Maybe, it makes sense to have a transactional version of sealing (e.g., by reworking `L2BlockSealSubtask::run()` to take `&mut Connection<'_, Core>` instead of a pool; in the concurrent version, a separate connection would be acquired for each task, and in the sequential / transactional version, tasks would be called sequentially with the same connection).

core/lib/zksync_core/src/state_keeper/io/seal_logic/l2_block_seal_subtasks.rs
perekopskiy: Stupid question: what's the difference between my and your pieces of code?

slowli: Your version spawns Tokio tasks, which provides indirection and true parallelism (if it's supported by the async runtime, which it is). On the downside, it requires all spawned tasks to be `'static` (note that in my approach joined futures are not `'static`; they reference `command` and `pool`) and complicates error handling. In contrast, my version would run all joined futures in the parent task (of course, they still would be polled concurrently, but if they need real parallelism to work efficiently, this won't be enough).

core/lib/zksync_core/src/state_keeper/io/seal_logic/mod.rs
```diff
@@ -269,10 +269,16 @@ impl UpdatesManager {
     }
 }
 
+#[derive(Debug)]
+pub(crate) enum SealStrategy<'a, 'b> {
```
slowli, line 273: Nit: To make dealing with this type simpler, you can take `Connection` by value; in the only place when the `Sequential` variant is constructed, you can obtain the connection later by destructuring the enum. This has a downside that it's possible to forget to commit the transactional connection at the end, but since it's used in a single place, IMO, it's manageable.

Here's another optimization that would simplify `SealStrategy` use. Since it's not used that extensively, feel free to ignore.

```rust
#[derive(Debug)]
pub(crate) enum SealStrategy<'pool> {
    Sequential(Connection<'pool, Core>),
    Parallel(&'pool ConnectionPool<Core>),
}

// As opposed to `Cow` from `std`; a union of an owned type and a mutable ref to it
enum Goat<'a, T> {
    Owned(T),
    Borrowed(&'a mut T),
}

impl<T> ops::Deref for Goat<'_, T> {
    type Target = T;

    fn deref(&self) -> &Self::Target {
        match self {
            Self::Owned(value) => value,
            Self::Borrowed(value) => value,
        }
    }
}

impl<T> ops::DerefMut for Goat<'_, T> {
    fn deref_mut(&mut self) -> &mut Self::Target {
        match self {
            Self::Owned(value) => value,
            Self::Borrowed(value) => value,
        }
    }
}

impl<'pool> SealStrategy<'pool> {
    async fn connection(&mut self) -> anyhow::Result<Goat<'_, Connection<'pool, Core>>> {
        Ok(match self {
            Self::Parallel(pool) => {
                Goat::Owned(pool.connection_tagged("state_keeper").await?)
            }
            Self::Sequential(conn) => Goat::Borrowed(conn),
        })
    }
}

// Then, you can simplify strategy uses like this:
if self.pre_insert_txs {
    let progress = L2_BLOCK_METRICS.start(L2BlockSealStage::PreInsertTxs, is_fictive);
    let mut connection = strategy.connection().await?;
    self.insert_transactions(&mut connection)
        .await
        .context("failed persisting transactions in L2 block")?;
    progress.observe(Some(self.l2_block.executed_transactions.len()));
}
```

The trick here is to have a meaningful lifetime in `SealStrategy` (namely, that of a connection pool); otherwise, it would be issues coecring `&mut Connection<'a, Core>` to `&mut Connection<'b, Core>` with `'a != 'b` because mutable references are invariant w.r.t. the reference type (which includes `'a` / `'b` in this case).

slowli (approved):

slowli (dismissed):

RomanBrodetski (approved):

PR #1808, opened 2024-04-26
fix: provide updated manual deposit for bot ETH and non-ETH chains

Raid5594:
## What ❔

Updated manual deposit for rich accounts to account for Bridgehub changes.
Added support for non-ETH based chains.

## Why ❔

Old API is no longer supported by the bridgehub.

## Checklist

### DIFF
--- a/infrastructure/local-setup-preparation/src/index.ts
+++ b/infrastructure/local-setup-preparation/src/index.ts
@@ -4,12 +4,15 @@ import { getEthersProvider, getWalletKeys } from './utils';
 
 // 10**12 ether
 const AMOUNT_TO_DEPOSIT = ethers.utils.parseEther('1000000000000');
+const OPERATOR = '0x4F9133D1d3F50011A6859807C837bdCB31Aaab13'; // We do not want to perform any L1 tx from operator address
+const CALLDATA = '0x';
 
 async function depositWithRichAccounts() {
     const ethProvider = getEthersProvider();
 
     const chainId = process.env.CHAIN_ETH_ZKSYNC_NETWORK_ID!;
     const wallets = getWalletKeys().map((wk) => new ethers.Wallet(wk.privateKey, ethProvider));
+    const isEthBasedChain = process.env.CONTRACTS_BASE_TOKEN_ADDR === utils.ETH_ADDRESS_IN_CONTRACTS;
 
     const handles: Promise<any>[] = [];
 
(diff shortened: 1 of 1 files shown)


infrastructure/local-setup-preparation/src/index.ts
```diff
@@ -31,34 +34,51 @@ async function depositWithRichAccounts() {
...
-                AMOUNT_TO_DEPOSIT,
-                '0x',
-                DEPOSIT_L2_GAS_LIMIT,
-                utils.REQUIRED_L1_TO_L2_GAS_PER_PUBDATA_LIMIT,
-                [],
-                wallet.address,
-                overrides
-            )
-        );
+            // TODO: Currently we're providing zero as an operator fee, which works right now,
```
mm-zk, line 61: let's be a little more strict about the TODOs.
Do you know why this will change in the future? if yes, do we have any linear ticket for this (and then we should reference this ticket here)

infrastructure/local-setup-preparation/src/index.ts
```diff
@@ -31,34 +34,51 @@ async function depositWithRichAccounts() {
     );
 
     for (const wallet of wallets) {
-        const contract = new ethers.Contract(process.env.CONTRACTS_BRIDGEHUB_PROXY_ADDR, utils.BRIDGEHUB_ABI, wallet);
+        if (!(wallet.address === OPERATOR)) {
```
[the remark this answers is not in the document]
kelemeno: If there a 'getOperateKey' is used, it should also  be used for the actual operator, this means: adding it to ValidatorTimelock in deploy.ts and adding it to the env vars, 
`ETH_SENDER_SENDER_OPERATOR_COMMIT_ETH_ADDR`
`ETH_SENDER_SENDER_OPERATOR_BLOBS_ETH_ADDR`

This might be a bit different for local-setup though.

infrastructure/local-setup-preparation/src/index.ts
Raid5594: double-checked with the team, it's an outdated comment. Operator fees are provided as expected

infrastructure/local-setup-preparation/src/utils.ts
```diff
@@ -7,6 +8,55 @@ interface WalletKey {
...
+        contractCache.contractMain = new ethers.Contract(
+            process.env.CONTRACTS_DIAMOND_PROXY_ADDR,
+            utils.ZKSYNC_MAIN_ABI,
+            ethProvider
+        );
+    }
+
+    if (!contractCache.stm) {
+        const stateTransitionManagerAddr = await contractCache.contractMain.getStateTransitionManager();
+        const stmABI = ['function validatorTimelock() view returns (address)'];
```
kelemeno, line 32: It might be better to use the full Abi, but it is probably ok for now

infrastructure/local-setup-preparation/src/index.ts
```diff
@@ -31,34 +33,49 @@ async function depositWithRichAccounts() {
...
+                        mintValue: overrides.value,
+                        l2Contract: wallet.address,
+                        l2Value: AMOUNT_TO_DEPOSIT,
+                        l2Calldata: CALLDATA,
+                        l2GasLimit: DEPOSIT_L2_GAS_LIMIT,
+                        l2GasPerPubdataByteLimit: utils.REQUIRED_L1_TO_L2_GAS_PER_PUBDATA_LIMIT,
+                        factoryDeps: [],
+                        refundRecipient: wallet.address
+                    },
+                    isEthBasedChain ? overrides : {}
```
kelemeno, line 75: This is actually a nice solution to the dance between the mintValue, overrides.value options. It works here and not in the SDK because you can set overrides to {}.

kelemeno (approved):

kelemeno (approved):

ly0va (approved):

PR #1812, opened 2024-04-26
feat(state-keeper): Remove `pre_insert_txs` step

perekopskiy:
## What ❔

Removes pre_insert_txs l2 block seal step. Instead `insert_txs` flag should be passed to `mark_txs_as_executed_in_l2_block`.

## Why ❔

Inserting txs with block data right away is faster then insert + updating block data.

## Checklist


core/lib/dal/src/transactions_dal.rs
```diff
@@ -511,6 +511,7 @@ impl TransactionsDal<'_, '_> {
         transactions: &[TransactionExecutionResult],
         block_base_fee_per_gas: U256,
         protocol_version: ProtocolVersionId,
+        insert_txs: bool,
```
RomanBrodetski, line 514:
```suggestion
        // On the main node, transactions are inserted into the DB by API servers.
        // However on the EN, they need to be inserted after they are executed by the state keeper
        insert_txs: bool,
```

core/lib/dal/src/transactions_dal.rs
```diff
@@ -524,19 +525,55 @@ impl TransactionsDal<'_, '_> {
...
+        if insert_txs {
+            let tx_hashes: Vec<_> = transactions.iter().map(|tx| tx.hash.as_bytes()).collect();
+            sqlx::query!(
+                r#"
+                DELETE FROM transactions
+                WHERE
+                    hash = ANY ($1)
+                "#,
+                &tx_hashes as &[&[u8]],
+            )
```
RomanBrodetski, line 537: hmmmm how can this happen?

core/lib/dal/src/transactions_dal.rs
```diff
@@ -571,7 +608,211 @@ impl TransactionsDal<'_, '_> {
...
+        l2_block_number: L2BlockNumber,
+        block_base_fee_per_gas: U256,
+        transactions: &[TransactionExecutionResult],
+    ) -> DalResult<()> {
+        let l2_txs_len = transactions
+            .iter()
+            .filter(|tx_res| {
+                matches!(
+                    tx_res.transaction.common_data,
+                    ExecuteTransactionCommon::L2(_)
```
RomanBrodetski, line 622: why do l1 transactions need special treatment? Please add a comment

RomanBrodetski (approved):

core/lib/dal/src/transactions_dal.rs
perekopskiy: Do you mean how EN can have txs that are not included in a block? If so then 2 cases:
1. recovery from main node postgres dump
2. block rollback

core/lib/dal/src/transactions_dal.rs
perekopskiy: Same as with updating: l2/l1/upgrade txs have different sets of fields that need to be inserted/updated

core/lib/dal/src/transactions_dal.rs
```diff
@@ -571,7 +613,211 @@ impl TransactionsDal<'_, '_> {
         }
     }
 
-    async fn handle_executed_l2_transactions(
+    async fn insert_executed_l2_transactions(
```
slowli, line 616: Dumb question: Would it be inefficient / impossible to rework queries updating executed transactions used by `mark_txs_as_executed_in_l2_block()` into insert-or-update queries? If this could be possible, it would lead to better code reuse and less queries to maintain.

perekopskiy: It's not possible with our schema. We have two unique constraints on `transactions`: hash and (initiator, nonce). Postgres allows upsert only on one unique constraint

RomanBrodetski:
@perekopskiy let's go for it! Or is there something that blocks this?

perekopskiy:
@RomanBrodetski I'll have to rebase it against `main` if we want to merge it sooner than https://github.com/matter-labs/zksync-era/pull/1801. It requires some effort, let me check, maybe 1801 will be merged really soon

PR #1816, opened 2024-04-28
Escaping the $ in services.postgres.healthcheck.test

breezytm:
## What ❔

The PR solves an error when starting the container 

**Error:** `invalid interpolation format for services.postgres.healthcheck.test: "psql -U postgres -c \"select exists (select * from pg_stat_activity where datname = '{{ database_name }}' and application_name = 'pg_restore')\" | grep -e \".f$\"". You may need to escape any $ with another $`


tomg10:
Hi @breezytm, to let me reproduce this issue, what version of docker compose are you using?

breezytm:
```Docker Compose version v2.4.1```

deniallugo-ml:

PR #1828, opened 2024-04-29
feat(config): Wrap sensitive urls

slowli:
## What ❔

Wraps URLs with potential sensitive parts (username / password) in a newtype. Uses these URLs in configs and elsewhere.

## Why ❔

Expresses sensitivity via type system (= easier to track and reason about uses). Makes it safe to `derive(Debug)` for types containing such URLs.

## Checklist


core/bin/external_node/src/config/mod.rs
```diff
@@ -742,70 +750,55 @@ pub(crate) struct ExternalNodeConfig {
...
-
-        let optional = envy::prefixed("EN_")
-            .from_env::<OptionalENConfig>()
-            .context("could not load external node config")?;
-
+    /// Loads config from the environment variables and fetches contracts addresses from the main node.
+    pub async fn new(
+        required: RequiredENConfig,
+        optional: OptionalENConfig,
+        main_node_client: &BoxedL2Client,
```
slowli, line 757: The clients are provided so that they are not initialized multiple times (at least the L2 client is supposed to be a singleton).

core/lib/basic_types/src/url.rs
```diff
@@ -0,0 +1,117 @@
+use std::{fmt, str::FromStr};
+
+use serde::{Deserialize, Deserializer};
+use url::Url;
+
+/// URL with potentially sensitive authentication info (user name and password). Has a specialized
+/// `Debug` implementation to hide sensitive parts and explicit API to expose the underlying URL.
+#[derive(Clone, PartialEq)]
+pub struct SensitiveUrl {
```
slowli, line 9: Not sure in which crate this type should be placed.

core/lib/db_connection/src/connection_pool.rs
```diff
@@ -24,25 +25,14 @@ use crate::{
 };
 
 /// Builder for [`ConnectionPool`]s.
-#[derive(Clone)]
+#[derive(derive_more::Debug, Clone)]
```
slowli, line 28: Not sure whether using `derive_more` makes sense here; it'd be relatively easy to implement `Debug` manually (although in this case it's easy for this impl to become stale).

core/lib/eth_client/src/clients/http/query.rs
```diff
@@ -22,29 +26,44 @@ use crate::{
...
+        Ok(Self {
+            web3: Web3::new(transport),
+            url,
+        })
     }
 }
 
 #[async_trait]
 impl EthInterface for QueryClient {
+    async fn fetch_chain_id(&self, component: &'static str) -> Result<L1ChainId, Error> {
```
slowli, line 57: I have a follow-up PR refactoring L1 clients almost prepared, but it depends on the code in this PR, so I'll publish it after this one is merged.

core/lib/eth_client/src/lib.rs
```diff
@@ -179,7 +183,7 @@ static_assertions::assert_obj_safe!(EthInterface);
 /// 2. Consider adding the "unbound" version to the `EthInterface` trait and create a default method
 /// implementation that invokes `contract` / `contract_addr` / `sender_account` methods.
 #[async_trait]
-pub trait BoundEthInterface: EthInterface {
+pub trait BoundEthInterface: AsRef<dyn EthInterface> + 'static + Sync + Send + fmt::Debug {
```
slowli, line 186: Not sure whether this is controversial; IMO, `EthInterface` and `BoundEthInterface` encapsulate "adjacent" functionality, so it doesn't really make sense to require one inherit from the other.

core/bin/external_node/src/config/mod.rs
```diff
@@ -621,26 +630,31 @@ pub(crate) struct RequiredENConfig {
     /// Port on which the healthcheck REST server is listening.
     pub healthcheck_port: u16,
     /// Address of the Ethereum node API.
-    /// Intentionally private: use getter method as it manages the missing port.
```
slowli, line 624: AFAIU, the underlying issue [was fixed in `jsonrpsee` v0.20](https://github.com/paritytech/jsonrpsee/blob/master/CHANGELOG.md#v0200---2023-08-11).

core/lib/basic_types/src/url.rs
popzxc, line 9: That's neat, although I somewhat feel like it may be overly specific. I see that we already have [`K256PrivateKey`](https://github.com/matter-labs/zksync-era/blob/main/core/lib/crypto_primitives/src/ecdsa_signature.rs#L47) that generally serves the similar purpose.

At this stage, wouldn't it be better to just use [secrecy](https://docs.rs/secrecy/latest/secrecy/) (I feel that both this PR and the interface of `K256PrivateKey` are kinda influenced by it)?

popzxc: If we are both to keep it and keep specific, then I think basic types is OK for time being.
If we are to generalize the approach for any kind of potentially sensitive data, then a separate dedicated crate may be better, IMHO.

slowli: Yeah, I was inspired by `secrecy` when naming method. IMO, it's sort of doesn't work in either case:

- Private keys from libsecp256k1 already implement most of necessary functionality (zeroize on drop, opaque `Debug`); it's just that their `Debug` impl doesn't work the best in our case (since it wasn't designed with Ethereum in mind).
- For URLs, we'd arguably want to output *some* parts of an URL in `Debug`; hence, a custom type.

popzxc (dismissed):
Still not fully sold that we need an ad-hoc approach rather than some generic wrapper that "Just Works" (mainly because of adoption potential), but certainly believe that it's an improvement over what we have now.

popzxc (approved):

popzxc (approved):

popzxc (approved):

PR #1843, opened 2024-05-01
refactor(node_framework): Refactor pools layer

popzxc:
## What ❔

Refactors pools layer:

- Introduces a single generic type instead of 3 mostly copy-pasted ones.
- Parameters for builder are now passed to the resource instead of cloning a builder.
- Fixes a problem where an unbound pool is created for each requester (now it's lazily initialized and cloned).

## Why ❔

a) less boilerplate
b) align the implementation to the expected behavior

## Checklist

### DIFF
--- a/core/node/node_framework/src/implementations/resources/pools.rs
+++ b/core/node/node_framework/src/implementations/resources/pools.rs
@@ -1,62 +1,103 @@
-use std::sync::{
-    atomic::{AtomicU32, Ordering},
-    Arc,
+use std::{
+    fmt,
+    sync::{
+        atomic::{AtomicU32, Ordering},
+        Arc,
+    },
+    time::Duration,
 };
 
 use prover_dal::Prover;
+use tokio::sync::Mutex;
 use zksync_dal::{ConnectionPool, Core};
 use zksync_db_connection::connection_pool::ConnectionPoolBuilder;
 
 use crate::resource::Resource;
 
-/// Represents a connection pool to the master database.
-#[derive(Debug, Clone)]
-pub struct MasterPoolResource {
+/// Represents a connection pool to a certain kind of database.
+#[derive(Clone)]
+pub struct PoolResource<P: PoolKind> {
     connections_count: Arc<AtomicU32>,
-    builder: ConnectionPoolBuilder<Core>,
+    url: String,
+    max_connections: u32,
+    statement_timeout: Option<Duration>,
+    unbound_pool: Arc<Mutex<Option<ConnectionPool<P::DbMarker>>>>,
+    _kind: std::marker::PhantomData<P>,
 }
 
-impl Resource for MasterPoolResource {
-    fn name() -> String {
-        "common/master_pool".into()
+impl<P: PoolKind> fmt::Debug for PoolResource<P> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("PoolResource")
+            .field("connections_count", &self.connections_count)
+            .field("url", &"...")
+            .field("max_connections", &self.max_connections)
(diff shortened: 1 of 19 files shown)


core/node/node_framework/src/implementations/resources/pools.rs
```diff
@@ -1,175 +1,148 @@
...
 
-/// Represents a connection pool to the master database.
-#[derive(Debug, Clone)]
-pub struct MasterPoolResource {
+/// Represents a connection pool to a certain kind of database.
+#[derive(Clone)]
+pub struct PoolResource<P: PoolKind> {
     connections_count: Arc<AtomicU32>,
-    builder: ConnectionPoolBuilder<Core>,
+    url: String,
```
popzxc, line 21: To be changed with a secure wrapper once the corresponding pr is merged.

core/node/node_framework/src/implementations/resources/pools.rs
```diff
@@ -1,175 +1,148 @@
...
         }
-
-        result
+        let pool = self.builder().build().await?;
+        *unbound_pool = Some(pool.clone());
+
+        let old_count = self
+            .connections_count
+            .fetch_add(self.max_connections, Ordering::Relaxed);
+        let total_connections = old_count + self.max_connections;
```
perekopskiy, line 79: nit: does it make sense to keep `total_connections` field? If I understand correctly it's equal to `(self.unbound_pool.is_some() as u32) * self.max_connections`

perekopskiy (approved):

AnastasiiaVashchuk (approved):

core/node/node_framework/src/implementations/resources/pools.rs
popzxc: Not really. `unbound_pool`, if created, has `max_connections`, but if it's created -- it is shared among all the requesters (e.g. if two components require `unbound_pool`, we're still at `max_connections`). However, we can go beyond `max_connections` if we create custom pools (e.g. singleton ones) -- and `total_connections` is normally `max_connections + how_many_custom_connections_have_been_requested`.

core/node/node_framework/src/implementations/resources/pools.rs
```diff
@@ -1,175 +1,148 @@
...
-            let old_count = self.connections_count.fetch_add(size, Ordering::Relaxed);
-            let total_connections = old_count + size;
-            tracing::info!(
-                "Created a new replica pool. Replica pool total connections count: {total_connections}"
-            );
-        }
-
-        result
-    }
+    fn kind_str() -> &'static str;
```
slowli, line 123: Bikeshedding: would a `const KIND_STR: &'static str` work here? IMO, it looks slightly more idiomatic (ofc, a function is more general, but it's not like implementations are expected to lazily initialize strings, leak them from a `Box<str>` etc.).

core/node/node_framework/src/implementations/resources/pools.rs
```diff
@@ -1,175 +1,148 @@
...
-            tracing::info!(
-                "Created a new replica pool. Replica pool total connections count: {total_connections}"
-            );
-        }
-
-        result
-    }
+#[derive(Debug, Clone)]
+#[non_exhaustive]
+pub struct ReplicaPool {}
```
slowli, line 114: Bikeshedding: IMO, a struct as a newtype wrapping `()` is a more conventional way to declare marker types. Obviously, doesn't matter much.

slowli (approved):

core/node/node_framework/src/implementations/resources/pools.rs
popzxc: Makes sense, will do in the next PR.

PR #1844, opened 2024-05-01
chore(node_framework): Fix flaky test

popzxc:
## What ❔

Due to the fact that the task wrapper has a `tokio::select`  [waiting for either ](https://github.com/matter-labs/zksync-era/blob/main/core/node/node_framework/src/task.rs#L69) barrier or stop signal, the second task in the test may not be started at all. This is highly unlikely, but it happened at least 2 times on CI.

This PR fixes it by introducing an additional barrier that makes sure that both tasks have started.

The fix is somewhat clunky, but it should do the job, and the tests there warrant a refactoring anyways, so let's keep it like that for the time being.

## Why ❔

Flaky tests bad. 

## Checklist

### DIFF
--- a/core/node/node_framework/src/service/tests.rs
+++ b/core/node/node_framework/src/service/tests.rs
@@ -2,7 +2,7 @@ use std::sync::{Arc, Mutex};
 
 use anyhow::anyhow;
 use assert_matches::assert_matches;
-use tokio::runtime::Runtime;
+use tokio::{runtime::Runtime, sync::Barrier};
 
 use crate::{
     service::{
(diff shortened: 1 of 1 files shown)


AnastasiiaVashchuk (approved):

pompon0 (approved):
LG. In general though it is a nice property for concurrency frameworks to execute all tasks that have been constructed, even if the stop signal has been already raised.

popzxc:
> In general though it is a nice property for concurrency frameworks to execute all tasks that have been constructed, even if the stop signal has been already raised.

That's a good point, and I thought so as well. There is, however, a caveat that cancellation may happen while not every precondition has exited, so I have no good answer right now on how to improve it. I've added this to my list of things to think about though.

PR #1848, opened 2024-05-02
feat(en): Brush up EN config

slowli:
## What ❔

- Records configuration metadata using a derive proc macro.
- Captures this metadata for EN and prints it if `--config-list` arg is passed to the EN binary.

## Why ❔

Improves EN UX.

## Checklist


slowli:
I've started implementing schema-assisted layered config parsing 
(roughly speaking, [`config`](https://crates.io/crates/config) + `envy` + custom pre-processing), but realized that it'll be too large to review. Still, I'm not satisfied with the EN config structure, which this PR doesn't address.

An excerpt of the output produced by `zksync_external_node --config-list`:

```text
# Optional configuration
This part of the external node config is completely optional to provide. It can tweak limits of the API, delay intervals of certain components, etc. If any of the fields are not provided, the default values will be used.

filters_limit
    Type: integer [Rust: usize], default: 10000
    Max possible limit of filters to be in the API state at once.

subscriptions_limit
    Type: integer [Rust: usize], default: 10000
    Max possible limit of subscriptions to be in the API state at once.

req_entities_limit
    Type: integer [Rust: usize], default: 1024
    Max possible limit of entities to be requested via API at once.

max_tx_size_bytes
max_tx_size
    Type: integer [Rust: usize], default: 1000000 [unit: bytes]
    Max possible size of an ABI-encoded transaction supplied to `eth_sendRawTransaction`.

vm_execution_cache_misses_limit
    Type: integer [Rust: Option<usize>], default: None
    Max number of cache misses during one VM execution. If the number of cache misses exceeds this value, the API server panics. This is a temporary solution to mitigate API request resulting in thousands of DB queries.
```

core/lib/config/src/metadata.rs
slowli, line 1: Not sure where to place this module. If we go with the home-grown config management framework, it could make sense to place it in a separate crate or even repo (like `vise`).

PR #1850, opened 2024-05-02
fix(pruning): Don't require metadata to exist for first L1 batches to be pruned

tomg10:


core/node/db_pruner/src/prune_conditions.rs
```diff
@@ -76,6 +76,15 @@ impl PruneCondition for NextL1BatchHasMetadataCondition {
     async fn is_batch_prunable(&self, l1_batch_number: L1BatchNumber) -> anyhow::Result<bool> {
         let mut storage = self.conn.connection().await?;
         let next_l1_batch_number = L1BatchNumber(l1_batch_number.0 + 1);
+        let protocol_version = storage
+            .blocks_dal()
+            .get_batch_protocol_version_id(next_l1_batch_number)
+            .await?;
+        // That old l1 batches must have been processed and those old batches are problematic
+        // as they have metadata that is not easily retrievable(misses some fields in db)
+        if protocol_version.is_none() || !protocol_version.unwrap().is_post_1_4_1() {
```
slowli, line 85: Nit: To be more idiomatic / avoid `unwrap()`s, you can use `map_or` here:

```rust
let old_protocol_version = protocol_version.map_or(true, |ver| !ver.is_post_1_4_1());
```

Also, there's `is_pre_1_4_1()` method that would allow to get rid of a negation.

slowli (approved):

core/node/db_pruner/src/prune_conditions.rs
tomg10: fixed

slowli (approved):

PR #1853, opened 2024-05-03
refactor(eth-client): Refactor L1 client interfaces

slowli:
## What ❔

Refactors L1 client interfaces so that they work similarly to L2 client ones.

## Why ❔

DevEx and maintainability.

## Checklist

### DIFF
--- a/core/bin/external_node/src/config/mod.rs
+++ b/core/bin/external_node/src/config/mod.rs
@@ -841,7 +841,7 @@ impl ExternalNodeConfig {
         // We can query them from main node, but it's better to set them explicitly
         // as well to avoid connecting to wrong environment variables unintentionally.
         let eth_chain_id = eth_client
-            .fetch_chain_id("en")
+            .fetch_chain_id()
             .await
             .context("Unable to check L1 chain ID through the configured L1 client")?;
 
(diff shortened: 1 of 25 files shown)


slowli:
The elephant in the room is: Why do we use different backends for L1 and L2 clients? L2 client uses `jsonrpsee`; L1 client uses `web3` and `jsonrpc-core` (both look barely maintained or unmaintained). Would it make sense to switch to `jsonrpsee` for the L1 client? Is it just a question of vendoring the appropriate types from `web3` and writing the `jsonrpsee` client trait(s) for the relevant subset of Web3 methods we use?

popzxc:
> L1 client uses web3 and jsonrpc-core (both look barely maintained or unmaintained). Would it make sense to switch to jsonrpsee for the L1 client?

L1 client is much older than L2 one, it [was initially written ](https://github.com/matter-labs/zksync/tree/master/core/lib/eth_client) for Lite :) 

So, if we can get rid of `jsonrpc`, that's great.

popzxc (approved):

perekopskiy (approved):

perekopskiy:
In addition to L1 client being much older, we use jsonrpsee client only for L2 because it's autogenerated from API declaration trait so it's very convenient to just use it. For L1 it requires some work to generate the client. As you have said it shouldn't be a big amount of work to switch to jsonrpsee.
As for `web3` and `jsonrpc-core` dependencies, the latter is only used because `web3` depends on it. We use types from `web3` a lot and it will be difficult to get rid of it. If I understand correctly the current state is that we are waiting for `alloy` to mature and then we switch

PR #1856, opened 2024-05-06
feat(vm-runner): implement output handler for VM runner

itegulov:
## What ❔

This PR creates a new abstraction `OutputHandlerFactory` that can orchestrate output handlers in a non-trivial manner. Specifically, for the purposes of VM runner we create an implementation `ConcurrentOutputHandlerFactory` that can produce delegated output handlers that internally use another output handler (i.e. delegatee is specific to the VM runner implementation). The main benefit of this implementation is that `handle_l1_batch` is non-blocking and finishes almost immediately. The actual work is spawned as a future and is then handled asynchronously enabling us to create many output handlers in parallel.

## Why ❔

One of the components for the upcoming VM runner

## Checklist


core/lib/zksync_core/src/vm_runner/output_handler.rs
```diff
@@ -0,0 +1,204 @@
...
+        &mut self,
+        l1_batch_number: L1BatchNumber,
+    ) -> anyhow::Result<Box<dyn StateKeeperOutputHandler>>;
+}
+
+pub struct ConcurrentOutputHandlerFactory<L: VmRunnerStorageLoader, F: OutputHandlerFactory> {
+    pool: ConnectionPool<Core>,
+    state: Arc<DashMap<L1BatchNumber, oneshot::Receiver<BoxFuture<'static, anyhow::Result<()>>>>>,
+    loader: L,
+    factory: F,
```
perekopskiy, line 32: Do you think it's better to keep factory here instead of `StateKeeperOutputHandler`? From the top of my head I don't see a use-case where factory is required, do you have any? Maybe I miss something and factory is required here then please explain

core/lib/zksync_core/src/vm_runner/output_handler.rs
```diff
@@ -0,0 +1,204 @@
...
+        let mut latest_processed_batch = self.loader.latest_processed_batch(&mut conn).await?;
+        drop(conn);
+        loop {
+            if *stop_receiver.borrow() {
+                tracing::info!("`ConcurrentOutputHandlerFactoryTask` was interrupted");
+                return Ok(());
+            }
+            match self.state.remove(&(latest_processed_batch + 1)) {
+                None => {
+                    tokio::time::sleep(Duration::from_millis(50)).await;
```
perekopskiy, line 189: nit: let's make it a constant

core/lib/zksync_core/src/vm_runner/output_handler.rs
```diff
@@ -0,0 +1,204 @@
...
+                }
+                Some((_, receiver)) => {
+                    let future = receiver.await?;
+                    future.await?;
```
perekopskiy, line 193: nit: let's add comment explaining what's going here

perekopskiy (commented):
Looks really good! I've left some nits. Also, let's add a unit test

core/lib/zksync_core/src/vm_runner/output_handler.rs
itegulov: Hmm, so my thinking was that each output handler instance would be short-lived. Meaning it gets created when the user asks for a new handler and then gets destroyed when `handle_l1_batch` is called. Thus every `AsyncOutputHandler` would own its own internal handler that gets created through the factory and is destroyed once `AsyncOutputHandler` is dropped.

The other option would be to share one underlying handler for all `AsyncOutputHandler`s, but then we would have to wrap it in `Arc<RwLock<...>>` and then acquire the lock every time we call `handle_l2_block()` etc which would create thread competition where it is not strictly needed.

itegulov:
Still working on the documentation but functionally I think the PR is ready for a review

core/node/vm_runner/src/output_handler.rs
```diff
@@ -0,0 +1,541 @@
...
+                    );
+                    tokio::time::sleep(SLEEP_INTERVAL).await;
+                }
+                Some((_, receiver)) => {
+                    // Wait until the future is sent through the receiver, happens when
+                    // `handle_l1_batch` is called on the corresponding output handler
+                    let future = receiver.await?;
+                    // Wait until the future is completed, meaning that the `handle_l1_batch`
+                    // computation has finished, and we can consider this batch to be completed
+                    future.await?;
```
slowli, line 239: If I understand correctly, this will make `handle_l1_batch()` not concurrent; futures don't do work unless polled, and you poll these futures one-by-one here. Is this intentional? To achieve concurrent processing, one way is spawning Tokio tasks and using their handles instead of futures; tasks progress independently of whether their handles are polled.

core/node/vm_runner/src/output_handler.rs
```diff
@@ -0,0 +1,541 @@
...
+                    latest_processed_batch += 1;
+                    let mut conn = self.pool.connection_tagged(Io::name()).await?;
+                    self.io
+                        .mark_l1_batch_as_completed(&mut conn, latest_processed_batch)
+                        .await?;
+                    drop(conn);
```
slowli, line 245: Nit: This is useless; `conn` is dropped at the end of the scope anyway.

core/node/vm_runner/src/output_handler.rs
```diff
@@ -0,0 +1,541 @@
...
+                    tracing::debug!(
+                        "Output handler for batch #{} has not been created yet",
+                        latest_processed_batch + 1
+                    );
+                    tokio::time::sleep(SLEEP_INTERVAL).await;
+                }
+                Some((_, receiver)) => {
+                    // Wait until the future is sent through the receiver, happens when
+                    // `handle_l1_batch` is called on the corresponding output handler
+                    let future = receiver.await?;
```
slowli, line 236: Nit: Could make sense to add `context` for an error here and below; esp. here because `RecvError` on its own is not very informative. (In this case, it means that the batch handler was dropped before fully processing a batch.)

core/node/vm_runner/src/output_handler.rs
```diff
@@ -0,0 +1,541 @@
...
+
+impl Debug for OutputHandlerState {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("OutputHandlerState").finish()
+    }
+}
+
+#[derive(Debug)]
+struct AsyncOutputHandler {
+    internal: Option<OutputHandlerState>,
```
slowli, line 143: Using an `Option` looks redundant here. You can use `mem::replace(&mut self.internal, OutputHandlerState::Finished)` in `handle_l1_batch()` instead. This would allow to unite `OutputHandlerState` and `AsyncOutputHandler` in a single type.

core/node/vm_runner/src/output_handler.rs
```diff
@@ -0,0 +1,541 @@
...
+/// `handle_l1_batch` implementation (where the heaviest work is expected to happen).
+///
+/// Once the asynchronous work done in `handle_l1_batch` finishes it is also guaranteed to mark the
+/// batch is processed by `Io`. It is guaranteed, however, that for any processed batch all batches
+/// preceding it are also processed. No guarantees about subsequent batches. For example, if
+/// batches #1, #2, #3, #5, #9, #100 are processed then only batches #{1-3} will be marked as
+/// processed and #3 would be the latest processed batch as defined in [`VmRunnerIo`].
+pub struct ConcurrentOutputHandlerFactory<Io: VmRunnerIo, F: OutputHandlerFactory> {
+    pool: ConnectionPool<Core>,
+    state: Arc<DashMap<L1BatchNumber, oneshot::Receiver<BoxFuture<'static, anyhow::Result<()>>>>>,
```
slowli, line 49: - Nit: Could make sense to define a type alias for `oneshot::Receiver<_>` so that it's easier to read.
- Is using a concurrent map (vs `Mutex<HashMap<_>>`) really warranted here? As long as we don't need references from the map, looks like a `Mutex` would do fine.

core/node/vm_runner/src/output_handler.rs
```diff
@@ -0,0 +1,541 @@
...
+///
+/// The idea behind this trait is that often handling output data is independent of the order of the
+/// batch that data belongs to. In other words, one could be handling output of batch #100 and #1000
+/// simultaneously. Implementing this trait signifies that this property is held for the data the
+/// implementation is responsible for.
+#[async_trait]
+pub trait OutputHandlerFactory: Debug + Send {
+    /// Creates a [`StateKeeperOutputHandler`] implementation for the provided L1 batch. Only
+    /// supposed to be used for the L1 batch data it was created against. Using it for anything else
+    /// is undefined behavior.
```
slowli, line 27: UB has specific meaning in Rust 🙃 AFAIU, in this case the use will lead to an error, which is much better than UB (which could e.g. mean a segfault).

core/node/vm_runner/Cargo.toml
```diff
@@ -16,13 +16,16 @@ zksync_dal.workspace = true
...
 zksync_storage.workspace = true
+zksync_core.workspace = true
 vm_utils.workspace = true
 
 tokio = { workspace = true, features = ["time"] }
 anyhow.workspace = true
 async-trait.workspace = true
 once_cell.workspace = true
 tracing.workspace = true
+futures = { workspace = true, features = ["compat"] }
```
slowli, line 27: Nit: Using `futures` just for `BoxFuture` is redundant; it's just a type alias which you can define yourself (also see the comment about using Tokio's `JoinHandle`s).

core/node/vm_runner/Cargo.toml
```diff
@@ -16,13 +16,16 @@ zksync_dal.workspace = true
 zksync_contracts.workspace = true
 zksync_state.workspace = true
 zksync_storage.workspace = true
+zksync_core.workspace = true
```
slowli, line 19: Is this a temporary dependency? I'd expect the dependency to be inverted eventually (i.e., core crate to depend on this library).

core/node/vm_runner/src/output_handler.rs
itegulov: 🤦 

I had a feeling the tests were not running concurrently, thanks for the great catch

core/node/vm_runner/src/output_handler.rs
itegulov: > Is using a concurrent map (vs Mutex<HashMap<_>>) really warranted here? As long as we don't need references from the map, looks like a Mutex would do fine.

Just seemed natural to use a concurrent map given that we never touch the same entries at the same time. Also, `dashmap` is already a transitive dependency in the system so just felt it would be appropriate to depend on it.

core/node/vm_runner/Cargo.toml
itegulov: Yep, no need for `BoxFuture` anymore anyway

core/node/vm_runner/Cargo.toml
itegulov: Yeah, so I think `zksync_core` as a crate is going to be gone soon anyway (see https://github.com/matter-labs/zksync-era/pull/1940). Really I just need to depend on state keeper to reuse some of its abstractions. Will change once it has its own crate.

itegulov: The PR is merged now so I changed the dependency to state keeper

perekopskiy (approved):

core/node/vm_runner/src/output_handler.rs
itegulov: Resolving this but feel free to reopen

PR #1859, opened 2024-05-06
feat: Extract several crates from zksync_core

popzxc:
## What ❔

Extracts the following crates from `zksync_core`:

- `db_pruner`
- `fee_model`
- `eth_sender`
- `vm_runner`
- `test_utils`

Most of the crates are extracted as-is to minimize diff (except for paths). Refactoring of these crates is considered to be out of scope.

## Why ❔

Part of effort of splitting `zksync_core` into a set of independent crates.

## Checklist


core/node/fee_model/src/l1_gas_price/gas_adjuster/mod.rs
```diff
@@ -134,7 +133,6 @@ impl GasAdjuster {
                 "Effective gas price is too high: {gas_price}, using max allowed: {}",
                 max_l1_gas_price
             );
-            KEEPER_METRICS.gas_price_too_high.inc();
```
popzxc, line 137: We've discussed in the past that this metric isn't used anywhere and it's OK to remove it (until we feel that we need it).

core/node/vm_runner/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+license.workspace = true
+keywords.workspace = true
+categories.workspace = true
+
+[dependencies]
+multivm.workspace = true
+vise.workspace = true
+zksync_types.workspace = true
+zksync_dal.workspace = true
+zksync_config.workspace = true
```
AnastasiiaVashchuk, line 17: `zksync_config` is redundant here

core/node/vm_runner/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+tokio = { workspace = true, features = ["time"] }
+anyhow.workspace = true
+async-trait.workspace = true
+once_cell.workspace = true
+itertools.workspace = true
+thiserror.workspace = true
+tracing.workspace = true
+
+[dev-dependencies]
+test-casing.workspace = true
```
AnastasiiaVashchuk, line 34: `itertools`, `thiserror`, `test-casing` as well are not used here

core/node/vm_runner/Cargo.toml
popzxc: Huh, funny that [udeps](https://github.com/est31/cargo-udeps) didn't discover that.
Good catch!

AnastasiiaVashchuk: @popzxc  same for other `Cargo.toml`s(for example `core/node/db_pruner/Cargo.toml`)
please check them all

core/node/db_pruner/README.md
```diff
@@ -1,4 +1,4 @@
-# Db pruner
+# `zksync_node_db_pruner`
```
AnastasiiaVashchuk, line 1: why do we change style  here?
It's a `.md` file and not code docs, so it looks weird to me, or I miss smth?

core/node/vm_runner/Cargo.toml
popzxc: Of course

core/node/db_pruner/README.md
popzxc: Just a matter of style. I think given that we have a lot of crate names, it'll be simpler to just use crate names as headers in these files rather than attempt to invent something unique in each case.

Generally, I think it's not that important, at least for the first iteration.

Deniallugo (approved):

PR #1864, opened 2024-05-06
feat(loadtest): Support shared bridge interface in loadtest EVM-571

ly0va:
## What ❔

Adds support for the contracts' new post-shared bridge interface in loadtest.

## Why ❔

To be able to run loadtest against custom hyperchains (including with custom base token), and not only Era.

## Checklist


core/tests/loadnext/src/executor.rs
```diff
@@ -85,11 +108,32 @@ impl Executor {
...
+                .unwrap_or("ETH".to_string())
+        );
+
         self.check_onchain_balance().await?;
-        self.mint().await?;
-        self.deposit_to_master().await?;
 
-        self.deposit_eth_to_paymaster().await?;
+        tracing::info!("Master Account: Minting Test ERC20 token...");
+        self.mint(self.config.main_token).await?;
```
StanislavBreadless, line 129: nit: it might be sometimes confusing to the reader on what is the difference between "main" and "base" tokens. I would suggest renaming the "main" token field

core/tests/loadnext/src/executor.rs
```diff
@@ -280,23 +324,36 @@ impl Executor {
             .get_balance(paymaster_address, None)
             .await?;
 
+        // TODO: decimals may differ from eth
```
StanislavBreadless, line 327: Better to create a task (if you believe it is to be resolved) or just state that in comment that the formatting supports only 18 decimals for now

core/tests/loadnext/src/sdk/ethereum/mod.rs
```diff
@@ -524,29 +566,43 @@ impl<S: EthereumSigner> EthereumProvider<S> {
...
-                l1_token_address,
-                amount,
+                U256::from(self.l2_chain_id),
+                base_cost + operator_tip,
+                U256::zero(),
                 l2_gas_limit,
                 U256::from(L1_TO_L2_GAS_PER_PUBDATA),
+                Address::default(),
+                bridge_address,
+                U256::zero(),
```
StanislavBreadless, line 588: maybe worth adding a comment that we do not support depositing ETH to non-eth-based chains

core/tests/loadnext/src/executor.rs
ly0va: It's used in workflows (e.g. [here](https://github.com/matter-labs/zksync-era-workflows/blob/5d959660d8a9c439b45a4f4d9c27af7993c6e035/.github/workflows/loadtest-reusable.yml#L123)) so probably easier to add a doc comment?

core/tests/loadnext/src/sdk/ethereum/mod.rs
```diff
@@ -185,7 +213,6 @@ impl<S: EthereumSigner> EthereumProvider<S> {
         l1_token_address: Address,
         bridge: Option<Address>,
     ) -> Result<Address, ClientError> {
-        // TODO(EVM-571): This should be moved to the shared bridge, which does not have `l2_token_address` on L1. Use L2 contracts instead.
         let bridge = bridge.unwrap_or(self.default_bridges.l1_erc20_default_bridge.unwrap());
```
kelemeno, line 216: Would it be possible to go via the L2SharedBridge? That would be a better long term solution. The L1 erc20 bridge is deprecated, we should avoid using it

core/tests/loadnext/src/sdk/ethereum/mod.rs
```diff
@@ -439,6 +479,7 @@ impl<S: EthereumSigner> EthereumProvider<S> {
 
     /// Performs a deposit in zkSync network.
     /// For ERC20 tokens, a deposit must be approved beforehand via the `EthereumProvider::approve_erc20_token_deposits` method.
+    /// Cannot deposit ETH onto custom base token chains.
```
kelemeno, line 482: We can deposit Eth to custom base token chains though. I am not sure what you meant here?

core/tests/loadnext/src/sdk/ethereum/mod.rs
ly0va: we don't use this method anymore, I can remove it

core/tests/loadnext/src/sdk/ethereum/mod.rs
ly0va: I meant this method can't.

kelemeno (dismissed):

deniallugo-ml:

PR #1869, opened 2024-05-07
refactor(en): Remove chain IDs and unused field from remote EN config

slowli:
## What ❔

- Removes `max_pubdata_per_batch` from the remote EN config.
- Moves L1 and L2 chain IDs from the remote config to the required one. Adds a managed task asynchronously validating these IDs.

## Why ❔

- `max_pubdata_per_batch` is unused.
- L1 and L2 chain IDs are required to start a node anyway. Async validation means that the EN can now start w/o connecting to L1 / L2 nodes (well, other than to fetch contract addresses etc., but those will be dealt with separately).

## Checklist

### DIFF
--- a/core/bin/external_node/src/config/mod.rs
+++ b/core/bin/external_node/src/config/mod.rs
@@ -31,7 +31,7 @@ use zksync_web3_decl::{
     client::BoxedL2Client,
     error::ClientRpcContext,
     jsonrpsee::{core::ClientError, types::error::ErrorCode},
-    namespaces::{EnNamespaceClient, EthNamespaceClient, ZksNamespaceClient},
+    namespaces::{EnNamespaceClient, ZksNamespaceClient},
 };
 
 pub(crate) mod observability;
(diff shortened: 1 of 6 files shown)


core/bin/external_node/src/config/mod.rs
```diff
@@ -639,6 +608,13 @@ impl OptionalENConfig {
 /// This part of the external node config is required for its operation.
 #[derive(Debug, Deserialize, Clone, PartialEq)]
 pub(crate) struct RequiredENConfig {
+    /// L1 chain ID (e.g., 9 for Ethereum mainnet). This ID will be checked against the `eth_client_url` RPC provider on initialization
+    /// to ensure that there's no mismatch between the expected and actual L1 network.
+    pub l1_chain_id: L1ChainId,
+    /// L2 chain ID (e.g., 270 for zkSync Era mainnet). This ID will be checked against the `main_node_url` RPC provider on initialization
+    /// to ensure that there's no mismatch between the expected and actual L2 network.
+    pub l2_chain_id: L2ChainId,
```
slowli, line 616: An alternative that would still preserve offline property (i.e., node being able to start / function w/o a stable connection to L1 or L2) is storing chain IDs in the node storage (as a part of `GenesisConfig`?). @Deniallugo Is there a plan to store `GenesisConfig` in the node storage eventually, or will it still be a (required?) part of the config? By @popzxc's classification of config params, `GenesisConfig` (or at least chain IDs) are immutable and set during genesis, so L1 is an authoritative source for the config; but if we want to support offline mode, we'd want to cache it locally.

popzxc: Yes, caching genesis configuration locally in the node storage matches my vision.

popzxc (approved):

PR #1873, opened 2024-05-07
feat(main): Gas oracle for custom token implementation

juan518munoz:
## What ❔

Implement a `zk server` component that fetches the conversion rate between the token used as base an ETH, in order to calculate gas prices on l2 correctly.

This PR also adds a component that returns a sample response (for developing purposes).

## Why ❔

Current implementation assumes gas costs on L2 are always as they would be when using an ETH based chain.

## Checklist


.init.env
popzxc, line 1: Empty file

Cargo.toml
```diff
@@ -235,6 +235,7 @@ zksync_block_reverter = { path = "core/node/block_reverter" }
 zksync_commitment_generator = { path = "core/node/commitment_generator" }
 zksync_house_keeper = { path = "core/node/house_keeper" }
 zksync_node_genesis = { path = "core/node/genesis" }
+zksync_base_token_fetcher = { path = "core/node/base_token_fetcher" }
```
popzxc, line 238: Thanks for making it a separate crate!

core/lib/zksync_core/src/lib.rs
```diff
@@ -335,6 +338,47 @@ pub async fn initialize_components(
...
+        });
+        task_futures.push(conversion_rate_task);
+    };
+
+    let base_token_fetcher = if components.contains(&Component::BaseTokenFetcher) {
+        if components.contains(&Component::DevConversionRateApi) {
+            // If we are using the dev conversion rate API,
+            // we need to wait for it to start before initializing the base token fetcher.
+            // a more deterministic way should be used
+            tokio::time::sleep(Duration::from_secs(1)).await;
```
popzxc, line 359: I don't think it's a very robust approach. Maybe you can provide something like:

```
impl BaseTokenFetcher {
// Will retry creation of the fetcher for at most `timeout`.
fn new_with_timeout(config: BaseTokenFetcherConfig, timeout: Duration) -> anyhow::Result<Self> {
 // ... try to create in a loop with small `sleep` until `timeout` passes or creation is successful
}
}
```

Still not pretty and shows fundamental problems in the design, but IMHO would suite as a better short-term approach.

core/node/base_token_fetcher/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+use hex::ToHex;
+use metrics::atomics::AtomicU64;
+use tokio::sync::Mutex;
+use zksync_config::configs::BaseTokenFetcherConfig;
+
+/// Trait used to query the stack's native token conversion rate. Used to properly
+/// determine gas prices, as they partially depend on L1 gas prices, denominated in `eth`.
+#[async_trait]
+pub trait ConversionRateFetcher: 'static + std::fmt::Debug + Send + Sync {
+    fn conversion_rate(&self) -> anyhow::Result<u64>;
```
popzxc, line 14: IIRC last time we've discussed it we came to a conclusion that `u64` probably won't work well here.
Is it going to be tackled in some other PR?

infrastructure/zk/src/run.ts
```diff
@@ -28,6 +28,18 @@ export async function deployERC20AndWeth({
                 { "name": "DAI",  "symbol": "DAI",  "decimals": 18 },
                 { "name": "wBTC", "symbol": "wBTC", "decimals":  8, "implementation": "RevertTransferERC20" },
                 { "name": "BAT",  "symbol": "BAT",  "decimals": 18 },
+                { "name": "GNT",  "symbol": "GNT",  "decimals": 18 },
```
popzxc, line 31: I think these changes are here by accident, IIRC these lines were removed recently. Please check what's the relevant part of the diff here and make sure to remove everything else.

Same in other files.

juan518munoz: Changed for the tokens used in main.

.init.env
juan518munoz: removed from PR

core/lib/zksync_core/src/lib.rs
juan518munoz: Done! Here's the implementation:
```rs
    /// Attemps to create a new `BaseTokenFetcher` instance with a timeout for getting the initial conversion rate
    pub async fn new_with_timeout(config: BaseTokenFetcherConfig) -> anyhow::Result<Self> {
        let http_client = reqwest::Client::new();

        let mut conversion_rate = None;
        let mut tries = 0;
        while conversion_rate.is_none() && tries < MAX_CONVERSION_RATE_FETCH_RETRIES {
            match http_client
                .get(format!(
                    "{}/conversion_rate/0x{}",
                    config.host,
                    config.token_address.encode_hex::<String>()
                ))
                .send()
                .await
            {
                Ok(res) => {
                    conversion_rate = Some(res.json::<u64>().await.context(
                        "Unable to parse the response of the native token conversion rate server",
                    )?);
                }
                Err(_err) => {
                    tries += 1;
                    sleep(Duration::from_secs(1)).await;
                }
            }
        }
```

core/node/base_token_fetcher/src/lib.rs
jrchatruc: The work for that was started on [this PR](https://github.com/matter-labs/zksync-era/pull/1324) but then got frozen after some discussions.

shahar4: @popzxc @jrchatruc we'll 
1. release this with the current limitations but should be very clear about them in docs and comments
2. go with `U256` as a follow-up

For now, I'll take care of docs and @jrchatruc please add comment/s where you see fit.

popzxc: I think `u64/U256` issue was about the type used for gas prices (and potential of overflow).
Here, however, we're discussing the fact that we cannot adequately express token prices (e.g. if the token price is $2000, and ETH price is $3000, we cannot express it using `u64`. Similarly, we cannot _precisely_ express token costing $0.7, creating some rounding error, and I don't know if it's OK).

jrchatruc: You're right, I got things mixed up, but we did make a PR which ultimately got merged into the `U256` one addressing the fractionary conversion rate issue ([here's a comment about it](https://github.com/matter-labs/zksync-era/pull/1324#issuecomment-2000339914) on the `U256` PR). I think in all the different PRs it got lost

docs/guides/launch.md
```diff
@@ -94,6 +94,52 @@ template.
...
+
+Additionally, you can run the server with the `dev_conversion_rate_api` component to enable the development conversion
+rate API, this component is only meant for development purposes and should not be used in production and will provide a
+fixed conversion rate for all tokens:
+
+```
+zk server --components="api,tree,eth,state_keeper,housekeeper,basic_witness_input_producer,commitment_generator,base_token_fetcher,dev_conversion_rate_api"
+```
+
+This server will use the host setted in the environment variable previously mentioned.
```
shahar4, line 128: `setted` -> `set`

core/node/fee_model/src/l1_gas_price/gas_adjuster/mod.rs
```diff
@@ -166,8 +177,9 @@ impl GasAdjuster {
         let calculated_price =
             (self.config.internal_l1_pricing_multiplier * effective_gas_price as f64) as u64;
 
-        // Bound the price if it's too high.
-        self.bound_gas_price(calculated_price)
+        let conversion_rate = self.base_token_fetcher.conversion_rate().unwrap_or(1);
```
shahar4, line 180: What about blobs? Seems like the gas prices there aren't determined using this.

core/lib/zksync_core/src/dev_api_conversion_rate/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+            tracing::info!("Stop signal received, conversion rate server is shutting down");
+        })
+        .await
+        .expect("Conversion rate server failed");
+    tracing::info!("Conversion rate server shut down");
+    Ok(())
+}
+
+// basic handler that responds with a static string
+async fn get_conversion_rate(extract::Path(token_address): extract::Path<String>) -> Json<u64> {
```
shahar4, line 32: what's going on here?

core/node/base_token_fetcher/src/lib.rs
shahar4: @popzxc Good point apologies for the confusion. I see what you're saying given this is used simply as `self.bound_gas_price(calculated_price) * conversion_rate`, and the linked PR also has nothing to do with this comment.

popzxc: @jrchatruc So just to be clear, it will be changed to correctly represent a ratio in this PR, right?

jrchatruc: We could definitely do that; for that we would need to merge [this PR](https://github.com/lambdaclass/zksync-era/pull/219) into this branch first before merging this one into `main`.

core/node/base_token_fetcher/README.md
```diff
@@ -0,0 +1,3 @@
+# zkSync Era Base Token Fetcher
+
+This crate contains functionality for the component responsible for fetching the base token price.
```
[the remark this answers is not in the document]
jrchatruc: Addressed here [791e7a5](https://github.com/matter-labs/zksync-era/pull/1873/commits/791e7a5d04f9d547fca529c638f5beca0a2147d6)

infrastructure/zk/src/up.ts
```diff
@@ -9,6 +9,7 @@ export function createVolumes() {
 }
 
 export async function up(runObservability: boolean, composeFile?: string) {
```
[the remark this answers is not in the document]
jrchatruc: what do you mean here?

core/tests/ts-integration/src/context-owner.ts
```diff
@@ -226,6 +226,7 @@ export class TestContextOwner {
         this.reporter.message(`Operator address is ${this.mainEthersWallet.address}`);
 
         const requiredL2ETHAmount = this.requiredL2ETHPerAccount().mul(accountsAmount);
+        // kl todo, for non eth based chains, we need to check different
```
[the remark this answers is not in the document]
jrchatruc: This was a leftover from the `kl-factory` branch, removed it here [ca05bcc](https://github.com/matter-labs/zksync-era/pull/1873/commits/ca05bccfd106c07f52d9d7dbe822bdf504e6561f)

docs/guides/launch.md
jrchatruc: Fixed here [1e0a0d4](https://github.com/matter-labs/zksync-era/pull/1873/commits/1e0a0d4cf6b74ffc87b8e044a3b258a6410eb4f5)

core/lib/zksync_core/src/dev_api_conversion_rate/mod.rs
jrchatruc: If the question is about the `if` statements below, those are in place to check for the `eth` as native token case, where the conversion rate is trivially `1`. The reason for the `0x000` and `0x001` addresses is that the address that identifies the `eth` native token in L2 is now `0x001` instead of zero, after some changes introduced by the shared bridge feature.

core/node/fee_model/src/l1_gas_price/gas_adjuster/mod.rs
jrchatruc: The `estimate_effective_gas_price` function is not concerned with the whole blob/pubdata pricing stuff, that's done by the `estimate_effective_pubdata_price` function. That being said, because this PR was done before blob functionality was introduced, that last function seems to be outdated and the price it returns should also be multiplied by the conversion rate on the `PubdataSendingMode::Blobs` case (see [here](https://github.com/matter-labs/zksync-era/blob/1e0a0d4cf6b74ffc87b8e044a3b258a6410eb4f5/core/node/fee_model/src/l1_gas_price/gas_adjuster/mod.rs#L185))

cc @popzxc for visiblity

infrastructure/zk/src/run.ts
```diff
@@ -178,3 +194,107 @@ command
     });
 
 command.command('snapshots-creator').action(snapshots_creator);
+
+command
```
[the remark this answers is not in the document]
jrchatruc: This was also a leftover from previous commits of the `kl-factory` branch, removed them here 
5671088

infrastructure/zk/src/test/test.ts
```diff
@@ -39,7 +39,6 @@ export async function l1Contracts() {
 export const command = new Command('test').description('run test suites').addCommand(integration.command);
 
 command.command('prover').description('run unit-tests for the prover').action(prover);
-command.command('l1-contracts').description('run unit-tests for the layer 1 smart contracts').action(l1Contracts);
```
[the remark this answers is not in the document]
jrchatruc: No most likely a leftover from previous branches; reintroduced here [ac715fb](https://github.com/matter-labs/zksync-era/pull/1873/commits/ac715fba81fbe2ff0d82fabc7b4f0530d391938d)

core/node/base_token_fetcher/src/lib.rs
popzxc: @jrchatruc looking at the PR title alone (the diff is broken there so I cannot see exact changes), it seems to be off-topic. This PR seems to be again about porting gas prices from `u64` to `U256` which AFAIK we decided not to do for now, since it has some issues. What we need to do is to change the data type returned by `conversion_rate` from `u64` to, for example, [Ratio](https://docs.rs/num/latest/num/rational/struct.Ratio.html) in order to precisely represent relations between ETH and native token prices.

shahar4: +1 to @popzxc

infrastructure/zk/src/up.ts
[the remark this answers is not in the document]
shahar4: why were these changes made?

core/node/fee_model/src/l1_gas_price/gas_adjuster/mod.rs
shahar4: Exactly. This has to be a part of this PR otherwise it doesn't make sense. @jrchatruc can you please fix that?

core/lib/zksync_core/src/dev_api_conversion_rate/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+            }
+            tracing::info!("Stop signal received, conversion rate server is shutting down");
+        })
+        .await
+        .expect("Conversion rate server failed");
+    tracing::info!("Conversion rate server shut down");
+    Ok(())
+}
+
+// basic handler that responds with a static string
```
shahar4, line 31: nit: to me, this comment doesn't add any clarity to the reader

core/lib/zksync_core/src/dev_api_conversion_rate/mod.rs
shahar4: 1. This is not clear without a comment
2. Why the 42?
3. This isn't the biggest deal because of the nature of this file as a local dev file, but this isn't robust. is there any config or more reliable than a hardcoded string value that can be used here?

core/node/base_token_fetcher/src/lib.rs
juan518munoz: We have creates a [PR](https://github.com/lambdaclass/zksync-era/pull/220) that moves the data type returned by `conversion_rate` to `BigDecimal`, we think it's the best approach, yet we also experimented a bit with `Ratio` and got it to work too.

core/node/fee_model/src/l1_gas_price/gas_adjuster/mod.rs
juan518munoz: Fixed it
```rs
    ...
    let conversion_rate = 
    self.base_token_fetcher.conversion_rate().unwrap_or(1);

    self.bound_blob_base_fee(calculated_price) * conversion_rate
}
```

core/lib/zksync_core/src/dev_api_conversion_rate/mod.rs
juan518munoz: 1. We have improved the function documentation a bit, do you think there's still some things worth adding/clarifying?
2. We picked 42 just because. It is beetwen the range of working conversion rates, and this being used for a dev component means that it doesn't affect other parts of the codebase.
3. It may be possible to move the value to an `.env` file if believe it's a better approach.

core/node/fee_model/src/l1_gas_price/gas_adjuster/mod.rs
shahar4: thanks! there's going to be a collision now with the other PR so just note that this is kept it with the new Decimal

core/lib/zksync_core/src/dev_api_conversion_rate/mod.rs
juan518munoz: Do you think the new comment gives enough clarity? Should we add anything elese?
```rs
/// Basic handler that responds with a static string with the format expected by the token fetcher component
/// Returns 1 in the case of using ETH as the base token, a hardcoded 42 otherwise
```

etc/env/base/base_token_fetcher.toml
Deniallugo, line 1: please add this config to file_based configuration as well  https://github.com/matter-labs/zksync-era/blob/main/etc/env/file_based/general.yaml

juan518munoz: Done!

popzxc:
I will close as stale; @shahar4 feel free to reopen if needed.

PR #1885, opened 2024-05-08
feat(node_framework): Add tree api server & client to the metadata calculator

popzxc:
## What ❔

Adds ability to spawn the tree API from the metadata calculator layer.
Note: this functionality isn't utilized by the main node example, I expect it to be done separately (likely when we'll be porting the main node itself to the framework).

## Why ❔

Previously it wasn't possible, so the implementation was incomplete.

## Checklist


core/node/node_framework/src/implementations/layers/metadata_calculator.rs
```diff
@@ -93,3 +132,24 @@ impl Task for MetadataCalculatorTask {
...
+
+#[async_trait::async_trait]
+impl Task for TreeApiTask {
+    fn name(&self) -> &'static str {
+        "tree_api"
+    }
+
+    async fn run(self: Box<Self>, stop_receiver: StopReceiver) -> anyhow::Result<()> {
+        self.tree_reader
+            .wait()
```
slowli, line 150: For additional responsiveness, it could make sense to select between this future and `stop_receiver`; `tree_reader.wait()` never resolves if the metadata calculator is dropped. Or you can change `tree_reader.wait()` signature to make it fallible.

slowli (dismissed):

core/lib/zksync_core/src/metadata_calculator/helpers.rs
```diff
@@ -355,15 +354,15 @@ impl LazyAsyncTreeReader {
     }
 
     /// Waits until the tree is initialized and returns a reader for it.
-    pub async fn wait(mut self) -> AsyncTreeReader {
+    pub async fn wait(mut self) -> anyhow::Result<AsyncTreeReader> {
```
slowli, line 357: Maybe an `Option` would be better? What I don't quite like in the current approach is that it'll produce errors during shutdown whereas logically they are not errors (the corresponding task should early-return `Ok(())` if it encounters such an "error"). OTOH, this is significantly less fluent than using `?`. One way to improve it (albeit fragile) is to create a special error type for such early returns and detect it in [the error chain](https://docs.rs/anyhow/latest/anyhow/struct.Error.html#method.chain) of task errors (if it's present, the task hasn't *really* errored).

popzxc: I thought about option too, but then I thought that maybe it _is_ an error to wait for a tree if there never will be one? This way we'll get an error report about potentially useless operation -- which may be a good thing.

AnastasiiaVashchuk (approved):

core/node/node_framework/src/implementations/layers/metadata_calculator.rs
```diff
@@ -25,11 +32,23 @@ use crate::{
 /// - Adds `tree_health_check` to the `ResourceCollection<HealthCheckResource>`.
 /// - Adds `metadata_calculator` to the node.
```
AnastasiiaVashchuk, line 33: worth to say that it adds TreeApiTask as well?

core/node/node_framework/src/implementations/layers/metadata_calculator.rs
```diff
@@ -49,15 +68,16 @@ impl WiringLayer for MetadataCalculatorLayer {
             .get_custom(10)
             .await?;
 
-        let object_store = context.get_resource::<ObjectStoreResource>().await.ok(); // OK to be None.
-        if object_store.is_none() {
-            tracing::info!(
-                "Object store is not provided, metadata calculator will run without it."
-            );
-        }
+        let object_store = match self.config.mode {
```
AnastasiiaVashchuk, line 71: will be fair to say it is `object_store_resource`

core/node/node_framework/src/implementations/layers/metadata_calculator.rs
popzxc: Doc-comments still will be tackled separately

core/node/node_framework/src/implementations/layers/metadata_calculator.rs
popzxc: I hope this can be inferred from the context 😅

PR #1888, opened 2024-05-08
feat(CI): contracts compilers cached in a docker image

tomg10:


mm-zk:
i see that in this pr you are simply copying binaries from docker to others.
my original idea was to mount this docker as a volume in our docker compose that we run when we setup tests.

slowli:
Just to understand: Is using [the GitHub `cache` action](https://github.com/actions/cache) not an option? I know that we run everything in Docker Compose, but as @mm-zk said, it should be possible to mount cache as a rw volume in the appropriate place. If this work, it'd be beneficial compared to the implemented approach because it wouldn't be necessary to manually synchronize compiler versions. OTOH, for the cache to work well, its key should reflect its content (so, it should include compiler versions somehow, perhaps in a hashed form).

infrastructure/zk/src/compiler.ts
```diff
@@ -13,13 +13,24 @@ export async function compileSystemContracts() {
...
 
+export async function prepareCompilersCache() {
+    // TODO(tomek) remove line below as soon as there is at least one pushed docker image
+    await utils.spawn('zk docker build compilers');
+    await utils.spawn('docker create --name temp matterlabs/compilers:latest2.0');
+    await utils.spawn('mkdir -p /root/.cache/hardhat-nodejs/compilers-v2');
+    await utils.spawn('docker cp temp:./.cache/hardhat-nodejs/compilers-v2/ /root/.cache/hardhat-nodejs/');
+    await utils.spawn('chmod -R +x /root/.cache/hardhat-nodejs/');
+    await utils.spawn('docker rm temp');
+}
```
itegulov, line 24: This seems very environment-specific. Also, can we at least reuse whatever hardhat is using to derive this directories? `env-paths` AFAIR

docker/compilers/Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+# Currently we only use 0.3.10 release of vyper compiler (path taken from the list.json above)
+RUN wget -nv -O $COMPILER_DIR/vyper/linux/0.3.10 https://github.com/vyperlang/vyper/releases/download/v0.3.10/vyper.0.3.10%2Bcommit.91361694.linux
+
+
+# This matches COMPILER_REPOSITORY_URL from hardhat-core.
+RUN wget -nv -O $COMPILER_DIR/linux-amd64/list.json https://binaries.soliditylang.org/linux-amd64/list.json
+
+RUN (for ver in solc-linux-amd64-v0.8.20+commit.a1b79de6 solc-linux-amd64-v0.8.23+commit.f704f362 solc-linux-amd64-v0.8.24+commit.e11b9ed9; do \
+    wget -nv -O $COMPILER_DIR/linux-amd64/$ver https://binaries.soliditylang.org/linux-amd64/$ver; \
+    done)
```
itegulov, line 49: We also have some arm runners, do you plan to cross-build this to support them too?

itegulov (commented):
I share @slowli's concern that we will need to manually update the images with this approach. Moreover, it will be very easy to upgrade solc version and forget about this Dockerfile as having the compiler here is not a hard requirement for CI.

Ideally I am envisioning some way to share this `env-paths`-based directory between builds so kind of what @slowli is talking about above.

tomg10:
Closing this for now until we switch from foundry to hardhat

PR #1892, opened 2024-05-09
feat(api): Add zeppelinos well-known slots

perekopskiy:
## What ❔

Add zeppelinos well-known slots

## Why ❔

Makes possible to use trusted zeppelinos proxy during tx validation

## Checklist


core/lib/constants/src/trusted_slots.rs
```diff
@@ -31,12 +31,24 @@ const INITIALIZER_INITIALING_SLOT: H256 = H256([
     0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01,
 ]);
 
+const ZEPPELINOS_ADMIN_SLOT: H256 = H256([
```
vladbochok, line 34: ZEPPELINOS?

perekopskiy: Yes, I didn't find an ERC for that, but it's used here:
https://github.com/zeppelinos/zos-lib/blob/8a16ef3ad17ec7430e3a9d2b5e3f39b8204f8c8d/contracts/upgradeability/UpgradeabilityProxy.sol#L24
https://github.com/zeppelinos/zos-lib/blob/8a16ef3ad17ec7430e3a9d2b5e3f39b8204f8c8d/contracts/upgradeability/AdminUpgradeabilityProxy.sol#L26
and circle's FiatTokenProxy is almost the same. If you think another name is more suitable please suggest

vladbochok: ZEPPELINOS is the legacy name for OPEN_ZEPPELIN library, I think we can change the name to OPEN_ZEPPELIN_ADMIN_SLOT

vladbochok: Uh, I see, the slot is also legacy one

vladbochok (approved):

PR #1897, opened 2024-05-09
feat(en): Brush up EN observability config

slowli:
## What ❔

Brushes up observability config and allows to configure Prometheus exporter to push changes to a gateway.

## Why ❔

Support of Prometheus push gateway is desired for some use cases (e.g., running an EN in Docker to measure its performance).

## Checklist


core/bin/external_node/src/config/observability.rs
```diff
@@ -1,43 +1,99 @@
-use zksync_config::configs::ObservabilityConfig;
-
-pub fn observability_config_from_env() -> anyhow::Result<ObservabilityConfig> {
-    // The logic in this method mimics the historical logic of loading observability options
-    // This is left intact, since some of the existing deployments may rely on the this behavior.
-    let sentry_url = if let Ok(sentry_url) = std::env::var("MISC_SENTRY_URL") {
-        if sentry_url == "unset" {
```
slowli, line 7: Is it reasonable to remove `unset` logic? Maybe I'm missing something obvious, but it seems that not having the corresponding env variable set would suffice.

core/bin/external_node/src/config/observability.rs
```diff
@@ -1,43 +1,99 @@
...
+
+/// Observability part of the node configuration.
+#[derive(Debug, Default, Deserialize)]
+pub(crate) struct ObservabilityENConfig {
+    /// Port to bind the Prometheus exporter server to. If not specified, the server will not be launched.
+    /// If the push gateway URL is specified, it will prevail.
+    pub prometheus_port: Option<u16>,
+    /// Prometheus push gateway to push metrics to. Overrides `prometheus_port`. A full URL must be specified
+    /// including `job_id` and other path segments; it will be used verbatim as the URL to push data to.
+    pub prometheus_pushgateway_url: Option<String>,
```
slowli, line 18: Is this approach reasonable? I don't really like how the gateway URL is opaquely constructed in `PrometheusConfig`.

core/bin/external_node/src/config/observability.rs
```diff
@@ -1,43 +1,99 @@
...
+            if let Some(value) = source.var(old_name) {
+                vars.insert(new_name.to_owned(), value);
+            }
         }
-    } else {
-        None
-    };
-    let sentry_environment = std::env::var("EN_SENTRY_ENVIRONMENT").ok().or_else(|| {
-        let l1_network = std::env::var("CHAIN_ETH_NETWORK").ok();
-        let l2_network = std::env::var("CHAIN_ETH_ZKSYNC_NETWORK").ok();
```
slowli, line 17: AFAICT, these vars are never set for our EN nodes; `EN_SENTRY_ENVIRONMENT` is always used instead; so I've removed this fallback.

core/bin/external_node/src/config/observability.rs
popzxc: I think people (myself included) are afraid to remove it to not accidentally put something in crashloop if some ancient component has `unset` in its configmap 😅
So to remove it, we must ensure that nobody (including external parties) has it in configmap.

slowli: Returned it back just to be safe.

RomanBrodetski: We have some `unset`s in [infra](https://github.com/matter-labs/gitops-kubernetes/blob/main/apps/common/zksync-v2-config/zksync-v2-config.yaml#L76)

PR #1903, opened 2024-05-10
snark proof is already verified inside wrap_proof function

olesHolem:
## What ❔

This PR avoids using useless era-zkevm-test-harness branch

## Why ❔

The only reason this dependency was used is verification of the wrapper proof that was already verified inside `wrap_proof` function.


Artemka374 (dismissed):

RomanBrodetski:
@olesHolem is this still relevant? If so, please resolve the conflict and let's merge!

PR #1906, opened 2024-05-10
fix(en): Fix reorg detector logic for dealing with last L1 batch

slowli:
## What ❔

Fixes a reorg detector false positive error if the latest remote L1 batch doesn't have its state hash computed.

## Why ❔

This error was observed in CI and can theoretically appear on real envs.

## Checklist


core/lib/zksync_core/src/reorg_detector/mod.rs
```diff
@@ -86,8 +93,7 @@ trait MainNodeClient: fmt::Debug + Send + Sync {
 
     async fn l2_block_hash(&self, number: L2BlockNumber) -> EnrichedClientResult<Option<H256>>;
 
-    async fn l1_batch_root_hash(&self, number: L1BatchNumber)
-        -> EnrichedClientResult<Option<H256>>;
+    async fn l1_batch_root_hash(&self, number: L1BatchNumber) -> EnrichedClientResult<StateHash>;
```
pompon0, line 96: why not Result<H256,SomeErr>?

core/lib/zksync_core/src/reorg_detector/mod.rs
```diff
@@ -346,9 +356,19 @@ impl ReorgDetector {
             .with_context(|| format!("Root hash does not exist for local batch #{l1_batch}"))?;
         drop(storage);
 
-        let Some(remote_hash) = self.client.l1_batch_root_hash(l1_batch).await? else {
-            tracing::info!("Remote L1 batch #{l1_batch} is missing");
-            return Err(HashMatchError::RemoteHashMissing);
+        let remote_hash = loop {
```
pompon0, line 359: how about having this retry loop within l1_batch_root_hash  call? Or how about  treating an l1 batch as nonexistent (for the purpose of reorg detector) until it has root_hash populated?

tomg10 (approved):

core/lib/zksync_core/src/reorg_detector/mod.rs
slowli: `EnrichedClientResult` is an alias for `Result<_, EnrichedClientError>` in the same spirit as `anyhow::Result<_>` is an alias of `Result<_, anyhow::Error>`. IMO, the `Result` suffix in the type name makes it easy to parse.

core/lib/zksync_core/src/reorg_detector/mod.rs
slowli: > how about having this retry loop within l1_batch_root_hash call?

This would work, but IMO, that'd break abstraction boundary; it's not client's responsibility to determine the course of action if the batch exists, but its state root hash doesn't. OTOH, the client trait is specific to the reorg detector, so that's a contentious point.

> how about treating an l1 batch as nonexistent (for the purpose of reorg detector) until it has root_hash populated?

Do you mean making multiple calls in `sealed_l1_batch_number()`? Again, could work, but it'd be less efficient, and in a common case when the remote L1 batch is the one to check, the obtained state root hash will be wasted. (Of course, it's possible to change `sealed_l1_batch_number()` signature so that it'd return the state root hash as well, but that looks clunky.)

core/lib/zksync_core/src/reorg_detector/mod.rs
pompon0: I know that. I meant  EnrichedClientResult<Result<H256,SomeError>> or embed the MissingBatch, MissingHash in the EnrichedClientError.

core/lib/zksync_core/src/reorg_detector/mod.rs
slowli: Good catch! IMO, it's reasonable to refactor `check_consistency()` so that it's cancel-aware. Since it does mutate the reorg detector state, this looks preferable to `select!`ing between it and the stop signal future.

core/lib/zksync_core/src/reorg_detector/mod.rs
slowli: Sorry for misunderstanding. Subjectively, Non-`H256` values don't feel like an error from the client perspective (more like an extended `Option`), and changing `EnrichedClientError` for this specific purpose looks awkward because it's defined in an upstream crate and is intended for general-purpose errors.

pompon0: Ok, I feel the opposite:
* anything different than root hash is not a happy path, therefore an error is a reasonable result
* Result IS an extended Option, since Option<X> is isomorphic to Result<X,()>. And it is the intended rust interpretation, given the obvious correspondence between Option and Result APIs

core/bin/external_node/src/main.rs
```diff
@@ -954,6 +953,16 @@ async fn run_node(
     )
     .await?;
     let sigint_receiver = env.setup_sigint_handler();
+    // Spawn reacting to signals in a separate task so that the node is responsive to signals right away
+    // (e.g., during the initial reorg detection).
+    tokio::spawn({
+        let stop_sender = stop_sender.clone();
+        async move {
```
pompon0, line 960: how about adding it to the task set? It would eliminate tokio::select entirely.

core/bin/external_node/src/tests.rs
```diff
@@ -94,6 +99,34 @@ fn expected_health_components(components: &ComponentsToRun) -> Vec<&'static str>
...
 }
 
+fn mock_eth_client(diamond_proxy_addr: Address) -> MockEthereum {
+    MockEthereum::default().with_call_handler(move |call, _| {
+        tracing::info!("L1 call: {call:?}");
+        if call.to == Some(diamond_proxy_addr) {
+            let call_signature = &call.data.as_ref().unwrap().0[..4];
+            let contract = zksync_contracts::hyperchain_contract();
+            let pricing_mode_sig = contract
+                .function("getPubdataPricingMode")
```
pompon0, line 109: is there no way of hardcoding those method names here (just for tests)? Perhaps there are some already defined constants that can be used instead?

core/lib/zksync_core/src/reorg_detector/mod.rs
pompon0: I also feel that it would be clearer to make l1_batch_root_hash cancel aware and move the retries to the implementation. This would simplify the control flow.

pompon0 (dismissed):

core/bin/external_node/src/main.rs
slowli: This is clever! However, [logging in `ManagedTasks`](https://github.com/matter-labs/zksync-era/blob/76564c54777e244925798952257107d8b9d76bfc/core/lib/utils/src/wait_for_tasks.rs#L37) assumes that a task terminating is an error. I don't really want to change `ManagedTasks` to account for this additional task given that we'll (hopefully) switch to the node framework in the midterm.

pompon0 (approved):

PR #1910, opened 2024-05-11
fix(core/prover): Changes to support Validium

EmilLuta:


popzxc (approved):
it seriously bothers me that we have more and more abstractions implemented for either rollup or validium (ideally we should have exactly 1, switching the rollup to either mode, or at the very least a factory), but otherwise lgtm

perekopskiy (approved):
`CommitmentPostProcessor` and `InputGenerator` look a bit hacky. Probably it would be better if the whole logic of commitment generation was abstracted -- but it can require some code duplication between Rollup and Validium impls.

koloz193:
> `CommitmentPostProcessor` and `InputGenerator` look a bit hacky. Probably it would be better if the whole logic of commitment generation was abstracted -- but it can require some code duplication between Rollup and Validium impls.

agreed, i debated between code duplication and having a specialized sub-component. the thing i wanted to avoid was ending up in a scenario where we make an update to `CommitmentGenerator` and needed to update all of the impls

yorik (approved):
CI approved

PR #1916, opened 2024-05-13
feat(local-node): Added support for validium and custom base token

mm-zk:
## What ❔

* Added support for validium and custom base token for local node.


docker/local-node/entrypoint.sh
```diff
@@ -146,8 +146,22 @@ else
     zk contract initialize-validator
   fi
 
-  ## init hyperchain
-  zk contract register-hyperchain
+
+  if [ -z "$CUSTOM_BASE_TOKEN" ]; then
+    echo "Starting chain with ETH"
```
shahar4, line 151: nit: might be a confusing log if it doesn't specify it's a gas token (also below)

"Starting chain with ETH as a gas token" instead?

docker/local-node/entrypoint.sh
```diff
@@ -146,8 +146,22 @@ else
     zk contract initialize-validator
   fi
 
-  ## init hyperchain
-  zk contract register-hyperchain
```
shahar4, line 150: Is there a chance something else is relying on this? The new CLI perhaps?

docker/local-node/entrypoint.sh
mm-zk: done.

docker/local-node/entrypoint.sh
mm-zk: it is still executed - in line 155.

'the -z' option means that a given env variable is not set.

shahar4 (approved):

PR #1918, opened 2024-05-13
chore(docs): external node documentation final tweaks

tomg10:


docs/guides/external-node/00_quick_start.md
```diff
@@ -52,17 +52,21 @@ The HTTP JSON-RPC API can be accessed on port `3060` and WebSocket API can be ac
 
 ### System Requirements
 
-This configuration is approximate and should be considered as **minimal** requirements.
+> [!NOTE]
+>
+> This configuration is only for nodes that use snapshots recovery, for requirements for nodes running from DB dump, see
```
RomanBrodetski, line 57: this is confusing for people with no background. We should mention something along the lines of

> This configuration is only for nodes that use snapshot recovery - they will not have historical data accessible via API. For Archival nodes, please refer to ...

tomg10: I've added some explainations in this "note"

PR #1923, opened 2024-05-14
feat(en): Fetch L1 batch root hashes from main node

slowli:
## What ❔

Implements a PoC for the treeless node mode, i.e., a mode in which a node doesn't maintain a Merkle tree locally, but rather downloads necessary data from a trusted peer (i.e., the main node). This extends to the case when a node is configured to start a Merkle tree after some time or immediately; the root hashes produced by the tree (which are expected to be slower than the tree data fetcher) will be used to double-check fetched hashes. Also, these hashes are checked by the consistency checker against L1 commitment transactions. 

## Why ❔

- Merkle tree recovery takes more than 50% of overall recovery process on the mainnet, so by introducing this mode, a recovering node can start operating a couple of hours earlier.
- Merkle tree introduces non-trivial operational requirements (~2.5 TB disk space / ~16 GB RAM on the mainnet), so by not running the tree, these requirements can be reduced.

## Checklist

fix(state-keeper): Fix hanging up during shutdown if L1 batch root hashes are not being computed


core/lib/dal/src/blocks_dal.rs
```diff
@@ -804,33 +804,12 @@ impl BlocksDal<'_, '_> {
         if update_result.rows_affected() == 0 {
             tracing::debug!("L1 batch #{number}: tree data wasn't updated as it's already present");
 
-            // Batch was already processed. Verify that existing hash matches
-            let matched: i64 = sqlx::query!(
```
slowli, line 808: Is there a particular reason the tree data was checked like this? I've simplified it just to use `get_l1_batch_tree_data()`.

core/lib/zksync_core/src/state_keeper/keeper.rs
```diff
@@ -274,23 +274,50 @@ impl ZkSyncStateKeeper {
...
+                .await?
             {
-                return Ok(envs);
+                return Ok(params);
             }
         }
         Err(Error::Canceled)
     }
 
+    async fn wait_for_new_batch_env(
```
slowli, line 293: This fixes state keeper hanging up during node shutdown if the requested L1 batch root hash is not being computed. It was technically an issue even without introducing tree data fetcher; I've just encountered this problem during local testing and decided it makes sense to fix.

slowli (commented):
To keep this PR reasonably small, I'd prefer to add integration testing in a separate PR.

core/bin/external_node/src/main.rs
```diff
@@ -641,6 +642,16 @@ async fn init_tasks(
         None
     };
 
+    if components.contains(&Component::TreeFetcher) {
```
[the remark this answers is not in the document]
popzxc: Nah, I think it's fine.

core/lib/dal/src/blocks_dal.rs
popzxc: Probably historical reasons(tm).

popzxc (dismissed):

popzxc (approved):

PR #1928, opened 2024-05-14
fix(basic_types): bincode deserialization for `web3::Bytes`

haraldh:
## What ❔

Only use custom format, if human readable serializer is used.

## Why ❔

Don't assume we only use serde_json... Make bincode ser/deser work

## Checklist

### DIFF
--- a/core/lib/basic_types/src/web3/mod.rs
+++ b/core/lib/basic_types/src/web3/mod.rs
@@ -119,6 +119,20 @@ impl<'a> Visitor<'a> for BytesVisitor {
     {
         self.visit_str(value.as_ref())
     }
+
+    fn visit_bytes<E>(self, value: &[u8]) -> Result<Self::Value, E>
+    where
+        E: Error,
+    {
+        Ok(Bytes(value.to_vec()))
+    }
+
+    fn visit_byte_buf<E>(self, value: Vec<u8>) -> Result<Self::Value, E>
+    where
+        E: Error,
+    {
+        Ok(Bytes(value))
+    }
 }
 
 // `Log`: from `web3::types::log`
--- a/core/lib/basic_types/src/web3/mod.rs
+++ b/core/lib/basic_types/src/web3/mod.rs
@@ -66,7 +66,7 @@ impl Serialize for Bytes {
             serialized.push_str(&hex::encode(&self.0));
             serializer.serialize_str(serialized.as_ref())
         } else {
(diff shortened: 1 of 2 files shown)


pbeza:
You might consider pulling in my simple test for your fix: https://github.com/matter-labs/zksync-era/pull/1933 (it properly fails if you revert your fix).

popzxc (dismissed):
LGTM, but would suggest to wait for a review from @slowli just in case.
Also, a unit test would be appreciated.

popzxc:
@haraldh Removed from the queue due to my comment above. Feel free to re-add if you disagree :)

haraldh:
Added the unit test from https://github.com/matter-labs/zksync-era/pull/1933 and extended it

popzxc (approved):

haraldh:
ok, waiting for a review from @slowli

core/lib/basic_types/src/web3/mod.rs
```diff
@@ -72,7 +76,11 @@ impl<'a> Deserialize<'a> for Bytes {
     where
         D: Deserializer<'a>,
     {
-        deserializer.deserialize_identifier(BytesVisitor)
+        if deserializer.is_human_readable() {
+            deserializer.deserialize_identifier(BytesVisitor)
+        } else {
+            Vec::<u8>::deserialize(deserializer).map(Bytes)
+        }
```
slowli, line 83: I recall there being a gotcha with this approach if the deserialized field is present in a `#[serde(flatten)]` group; I've encountered this in [my crate that implements similar functionality](https://github.com/slowli/hex-buffer-serde/blob/d203ab86d3e1772412728df19e3dca3d0eb6f447/src/var_len.rs#L305). Essentially, `serde` uses (or at least used in the past) a surrogate deserializer for such fields with `is_human_readable()` always set to true. This necessitates the original `BytesVisitor` supporting deserialization from bytes, even if it doesn't make logical sense. I guess in this case we can say that this field isn't supposed to be present in `#[serde(flatten)]` groups, but for completeness it could make sense to extend `BytesVisitor`.

popzxc: IIRC `bincode` doesn't support flattening anyways.

haraldh: anyways... added some `BytesVisitor` code ... plz re-review

popzxc (approved):

PR #1933, opened 2024-05-14
Add a [de]serialization unit test for the `Bytes`

pbeza:


core/lib/basic_types/src/web3/tests.rs
```diff
@@ -83,3 +83,12 @@ fn block_can_be_deserialized() {
     let block: Block<H256> = serde_json::from_str(pre_dencun).unwrap();
     assert!(block.excess_blob_gas.is_none());
 }
+
+#[test]
+fn test_bytes_serde() {
+    use bincode::{deserialize, serialize};
```
popzxc, line 89: Bikeshedding: probably using fully qualified paths here would be more expressive.

core/lib/basic_types/src/web3/tests.rs
```diff
@@ -83,3 +83,12 @@ fn block_can_be_deserialized() {
     let block: Block<H256> = serde_json::from_str(pre_dencun).unwrap();
     assert!(block.excess_blob_gas.is_none());
 }
+
+#[test]
+fn test_bytes_serde() {
```
popzxc, line 88: If we are to test differences between serializer, it might also make sense to add:
- Tests for human-readable (de)serialization, e.g. `serde_json` (though I guess it's kinda covered by the test above).
- Snapshot tests for bincode serialization (we don't want to accidentally break repr there)

haraldh:
added this to https://github.com/matter-labs/zksync-era/pull/1928

haraldh:
added this to https://github.com/matter-labs/zksync-era/pull/1928

PR #1938, opened 2024-05-15
feat(en): Improve tree snapshot recovery

slowli:
## What ❔

- Adds more logs that will allow to track tree recovery progress more clearly.
- Uses tagged DB connections.
- Makes chunk size configurable.

## Why ❔

- Logs and tagged connections improve observability.
- Configuring chunk size during tree recovery allows to fine-tune its performance (theoretically; needs to be tested on E2E tests).

## Checklist


slowli:
> us fine-tuning this config value will take time we should spend working on tree-less mode.

Come on, it's changing a value once per day in the E2E snapshot recovery workflow config and looking at results (this second part can be deferred). I don't think it would require much time if we set up it this way.

tomg10:
It's still much lower priority work, this sounds like P3 at best, while tree-less mode is P0/P1. The situation would look totally different if treeless progress was blocked by some external factors and this task was done as a "time filler", but as I understand this is not the case right now.

popzxc (approved):

PR #1943, opened 2024-05-15
fix(prover): bump cargo.lock version to include 1.5.0 crypto hotfix

RomanBrodetski:
## What ❔

New version of circuits

## Why ❔

There is a divergency between in-circuit and out-of-circuit logic


Artemka374 (approved):
LGTM

PR #1946, opened 2024-05-15
fix: parentHash in websocket blocks subscription is shown as 0x0

AnastasiiaVashchuk:
## Checklist


RomanBrodetski (approved):

core/lib/dal/src/blocks_web3_dal.rs
```diff
@@ -187,7 +189,10 @@ impl BlocksWeb3Dal<'_, '_> {
 
         let blocks = rows.into_iter().map(|row| BlockHeader {
             hash: Some(H256::from_slice(&row.hash)),
-            parent_hash: H256::zero(),
+            parent_hash: row
```
perekopskiy, line 192: There are a few more fields that are not set `here` but are set in `get_api_block`: `gas_used`, `gas_limit`, `base_fee_per_gas`. Please set them as well. Also, would be nice to cover `get_block_headers_after` with unit test

perekopskiy: But feel free to merge this and do in a separate PR

AnastasiiaVashchuk: will be covered separately

PR #1947, opened 2024-05-15
fix(utils): bincode ser-/deserialization for `BytesToHexSerde`

haraldh:
## What ❔

Only do a hex string conversion if serializer/deserializer `.is_human_readable()`.

## Why ❔

bincode does not need a hex string presentation.

## Checklist

### DIFF
--- a/core/lib/utils/src/serde_wrappers.rs
+++ b/core/lib/utils/src/serde_wrappers.rs
@@ -25,28 +25,70 @@ impl<P: Prefix> BytesToHexSerde<P> {
     where
         S: Serializer,
     {
-        // First, serialize to hexadecimal string.
-        let hex_value = format!("{}{}", P::prefix(), hex::encode(value));
+        if serializer.is_human_readable() {
+            // First, serialize to hexadecimal string.
+            let hex_value = format!("{}{}", P::prefix(), hex::encode(value));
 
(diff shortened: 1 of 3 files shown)


core/lib/utils/src/serde_wrappers.rs
```diff
@@ -25,28 +25,70 @@ impl<P: Prefix> BytesToHexSerde<P> {
...
+        }
     }
 
     pub fn deserialize<'de, D>(deserializer: D) -> Result<Vec<u8>, D::Error>
     where
         D: Deserializer<'de>,
     {
-        let deserialized_string = String::deserialize(deserializer)?;
+        if deserializer.is_human_readable() {
+            let deserialized_string = String::deserialize(deserializer)?;
```
slowli, line 44: This place could theoretically trigger the same problem with `is_human_readable()` set for a surrogate deserializer used for `#[serde(flatten)]` groups, but if it works as is, it can be dealt with later.

slowli (approved):

core/lib/utils/src/serde_wrappers.rs
haraldh: I don't want to pull in another deserializer just for testing.. if the need for one arises, we can revisit. bincode fails with flatten: https://github.com/bincode-org/bincode/issues/245

PR #1949, opened 2024-05-15
feat(vm): Fast VM integration

joonazan:
## What ❔

Integrates the new VM into server codebase:

- Allows to run the new VM in the state keeper / batch executor (but not the API server sandbox). The new VM is disabled by default everywhere and requires an explicit opt-in.
- Introduces the new component, VM playground, to test the new VM.
- Runs the load test with the new VM in CI.

## Why ❔

New VM is several times more efficient as per preliminary benchmarks and is easier to maintain.

## Checklist


slowli (approved):
I'm obviously biased here, but this PR looks ready to merge for me. Risks are (probably) mitigated by making the new VM strictly opt-in.

yorik (approved):
Approve for CI

PR #1951, opened 2024-05-15
fix: Disallow non null updates for transactions

Artemka374:
## What ❔

Add check to `set_tx_id` query, that doesn't allow to update tx_id if it is not NULL.

## Why ❔

To prevent issues when 1 batch can be commited 2 times.

## Checklist


core/lib/dal/src/blocks_dal.rs
```diff
@@ -402,17 +402,18 @@ impl BlocksDal<'_, '_> {
         number_range: ops::RangeInclusive<L1BatchNumber>,
         eth_tx_id: u32,
         aggregation_type: AggregatedActionType,
-    ) -> DalResult<()> {
+    ) -> anyhow::Result<()> {
```
Deniallugo, line 405: please keep dal error

Deniallugo (changes requested):
Please add tests for this feature

core/lib/dal/src/blocks_dal.rs
Artemka374: Well... I can't. There is no possibility to customize the DalResult. So it's either anyhow or panic

Deniallugo: @slowli what is the best approach for it? 
Probably we could add custom errors to dal error?

RomanBrodetski (dismissed):
very nice!

core/lib/dal/src/blocks_dal.rs
montekki: maybe it makes sense panicking here as we've discussed it should panic anyway.

Deniallugo: Just for me, it's better to panic, than using anyhow... 
But i think we have an ability to customize the errors

perekopskiy: `Instrumented::constraint_error` fits here

Deniallugo: Oh, thank you @perekopskiy! That's what I wanted

core/lib/dal/src/blocks_dal.rs
```diff
@@ -423,16 +424,30 @@ impl BlocksDal<'_, '_> {
                 .with_arg("eth_tx_id", &eth_tx_id)
                 .execute(self.storage)
                 .await?;
+
+                if result.rows_affected() == 0 {
+                    let err = Instrumented::new("set_eth_tx_id#commit")
+                        .with_arg("number_range", &number_range)
+                        .with_arg("eth_tx_id", &eth_tx_id)
+                        .arg_error(
```
slowli, line 432: Nit: This is a constraint error rather an arg one (an arg error should be possible to produce without running DB queries).

core/lib/dal/src/blocks_dal.rs
```diff
@@ -423,16 +424,30 @@ impl BlocksDal<'_, '_> {
                 .with_arg("eth_tx_id", &eth_tx_id)
                 .execute(self.storage)
                 .await?;
+
+                if result.rows_affected() == 0 {
+                    let err = Instrumented::new("set_eth_tx_id#commit")
```
slowli, line 429: Nit: You can define this `Instrumented` instance before the query and share it both for the query (via `Instrumented::with()`) and the following check. Would require minor changes to the `db_connection` crate (namely, adding `Clone` for `Instumented<'_, ()>`).

Deniallugo (approved):

slowli (approved):

montekki (approved):

PR #1954, opened 2024-05-15
fix(pruning): run vacuum exactly on midnight, so that deploys don't interrupt vacuum schedule

tomg10:


slowli (approved):
This looks more and more like a cron job rather than something that should be defined in app code 🙃 The downside of a cron job would be that we cannot control how it runs on "external" nodes (or whether it runs at all).

PR #1955, opened 2024-05-16
feat(vm-runner): implement VM runner main body

itegulov:
## What ❔

Main body of VM runner that combines all previously implemented components into a reusable framework that re-executes blocks in VM.

## Checklist


core/node/vm_runner/src/storage.rs
```diff
@@ -22,6 +21,24 @@ use zksync_types::{block::L2BlockExecutionData, L1BatchNumber, L2ChainId};
...
 use crate::VmRunnerIo;
 
+#[async_trait]
+pub trait StorageLoader: ReadStorageFactory {
+    /// Loads next unprocessed L1 batch along with all transactions that VM runner needs to
+    /// re-execute. These are the transactions that are included in a sealed L2 block belonging
+    /// to a sealed L1 batch (with state keeper being the source of truth). The order of the
+    /// transactions is the same as it was when state keeper executed them.
+    ///
+    /// Can return `None` if there are no batches to be processed.
```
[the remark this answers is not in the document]
itegulov: Good catch, the comment was from the previous iteration. I rephrased this to say that `None` is returned when the batch is temporarily unavailable as you said.

core/node/vm_runner/src/process.rs
```diff
@@ -0,0 +1,153 @@
...
+            let Some(batch_data) = self.storage.load_batch(next_batch).await? else {
+                // Next batch has not been loaded yet
+                tokio::time::sleep(SLEEP_INTERVAL).await;
+                continue;
+            };
+            let mut updates_manager =
+                UpdatesManager::new(&batch_data.l1_batch_env, &batch_data.system_env);
+            let Some(handler) = self
+                .batch_processor
+                .init_batch(
```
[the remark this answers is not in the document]
itegulov: Hmm, from what I see, errors in `init_batch` can only happen in the command receiver task. Or do you mean that we should represent interruption as a `Result` rather than `Option`?

core/node/vm_runner/src/process.rs
```diff
@@ -0,0 +1,153 @@
...
+    ///
+    /// Caller is expected to provide a component-specific implementation of [`VmRunnerIo`] and
+    /// an underlying implementation of [`OutputHandlerFactory`].
+    pub fn new(
+        pool: ConnectionPool<Core>,
+        io: Box<dyn VmRunnerIo>,
+        storage: Arc<dyn StorageLoader>,
+        output_handler_factory: Box<dyn OutputHandlerFactory>,
+        batch_processor: Box<dyn BatchExecutor>,
+    ) -> Self {
```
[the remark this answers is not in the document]
itegulov: Fair point. This introduced some awkwardness as now we have to store the same object as both `Arc<dyn ReadStorageFactory>` and `Arc<dyn StorageLoader>` but probably not a big deal (see the comment I left in the code for more context).

core/node/vm_runner/src/process.rs
```diff
@@ -0,0 +1,177 @@
...
+                if handle.is_finished() {
+                    handle
+                        .await
+                        .context("Failed to join a task to process a batch")?
+                        .context("Failed to process a batch")?;
+                } else {
+                    retained_handles.push(handle);
+                }
+            }
+            handles = retained_handles;
```
perekopskiy, line 86: nit: please add a comment explaining what's going on here

perekopskiy (approved):

core/node/state_keeper/src/testonly/test_batch_executor.rs
```diff
@@ -833,3 +837,18 @@ impl BatchExecutor for MockBatchExecutor {
...
+pub struct MockReadStorageFactory;
+
+#[async_trait]
+impl ReadStorageFactory for MockReadStorageFactory {
+    async fn access_storage(
+        &self,
+        _stop_receiver: &Receiver<bool>,
+        _l1_batch_number: L1BatchNumber,
+    ) -> anyhow::Result<Option<PgOrRocksdbStorage<'_>>> {
+        // Presume that the storage is never accessed so doesn't matter what we return here
```
slowli, line 851: Nit: If the storage is indeed never accessed, it makes sense to use `unimplemented!()` here.

core/node/vm_runner/src/process.rs
```diff
@@ -0,0 +1,177 @@
...
+    /// [`ConcurrentOutputHandlerFactory`], [`VmRunnerStorage`].
+    ///
+    /// Caller is expected to provide a component-specific implementation of [`VmRunnerIo`] and
+    /// an underlying implementation of [`OutputHandlerFactory`].
+    pub fn new(
+        pool: ConnectionPool<Core>,
+        io: Box<dyn VmRunnerIo>,
+        // Normally `storage` and `loader` would be the same object, but it is awkward to deal with
+        // them as such due to Rust's limitations on upcasting coercion. See
+        // https://github.com/rust-lang/rust/issues/65991.
```
slowli, line 47: If I'm not missing anything, this can be circumvented by adding a specific upcasting method to `StorageLoader`:

```rust
#[async_trait]
pub trait StorageLoader: ReadStorageFactory {
    fn upcast(self: Arc<Self>) -> Arc<dyn ReadStorageFactory>;
    // ...
}
```

The implementation is trivial:

```rust
#[async_trait]
impl<Io: VmRunnerIo> StorageLoader for VmRunnerStorage<Io> {
    fn upcast(self: Arc<Self>) -> Arc<dyn ReadStorageFactory> {
        self
    }
    // ...
}
```

core/node/vm_runner/src/process.rs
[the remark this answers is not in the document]
slowli: Ah, sorry; I've looked at `init_batch()` code not noticing that smelling `expect`s are performed in a separate task. It would make sense to propagate errors in that task though, but it definitely looks like something worthy a separate PR.

core/node/vm_runner/src/process.rs
```diff
@@ -0,0 +1,177 @@
...
+            else {
+                tracing::info!("VM runner was interrupted");
+                break;
+            };
+            let mut output_handler = self
+                .output_handler_factory
+                .create_handler(next_batch)
+                .await?;
+
+            let handle = tokio::task::spawn(async move {
```
slowli, line 121: Bikeshedding: I'd try to extract the spawned task code in a function.

core/node/vm_runner/src/process.rs
```diff
@@ -0,0 +1,177 @@
...
+            .latest_processed_batch(&mut self.pool.connection().await?)
+            .await?
+            + 1;
+        loop {
+            let mut retained_handles = Vec::new();
+            for handle in handles {
+                if handle.is_finished() {
+                    handle
+                        .await
+                        .context("Failed to join a task to process a batch")?
```
slowli, line 80: - Nit: The outer error is most probably a panic (it could also be a task getting aborted, but unless you call `JoinHandle::abort()`, this shouldn't happen). Thus, it could make sense to rephrase to "Processing a batch panicked".
- Nit: Is the batch number obvious from the source error? Otherwise, it'd be relatively easy to add it by using `Vec<(L1BatchNumber, JoinHandle<_>)>` or `HashMap<L1BatchNumber, _>`.

core/node/vm_runner/src/process.rs
itegulov: Nice trick, worked like a charm!

perekopskiy (approved):

slowli (approved):

PR #1959, opened 2024-05-16
fix(prover): Bump Cargo.lock and update VKs

Artemka374:
## What ❔

Bump versions of repos in Cargo.lock and update circuits.

## Why ❔

There is a divergency between in-circuit and out-of-circuit logic

## Checklist


prover/prover_fri/tests/basic_test.rs
```diff
@@ -79,6 +79,7 @@ async fn prover_and_assert_base_layer(
 //     prover_and_assert_base_layer(5176866, 1, L1BatchNumber(128623), 1086).await;
 // }
 
+#[ignore]
```
mm-zk, line 82: wait, why?

Artemka374: Because of changed circuit, the test is not passing, will need to update the proof later

mm-zk (approved):

PR #1964, opened 2024-05-16
test(en): Integration test for treeless mode

slowli:
## What ❔

Adds integration test coverage for the treeless EN mode by extending the snapshot recovery test and introducing a genesis recovery test (self-explanatory).

## Why ❔

Ensures that the treeless mode works as expected.

## Checklist


core/tests/recovery-test/tests/genesis-recovery.test.ts
```diff
@@ -0,0 +1,271 @@
...
+    externalNodeArgs,
+    getExternalNodeHealth,
+    killExternalNode,
+    NodeComponents,
+    sleep,
+    stopExternalNode,
+    waitForProcess
+} from '../src';
+
+// FIXME: check consistency checker health once it has acceptable speed
```
slowli, line 18: I've initially added consistency checker health checks to the new test; however, they've turned out to make the test exceedingly slow (order of 3 min) because of the slow commitment generator. I've created a task to tackle this separately.

tomg10 (changes requested):
I know this test does a lot and you did great work developing it, but I think some additional work should be put into making the code cleaner, extracting even more methods could make sense, separating low-level methods such as spawn and handling files and processes from high level test logic etc.

core/tests/recovery-test/tests/genesis-recovery.test.ts
```diff
@@ -0,0 +1,271 @@
...
+            env: externalNodeEnv
+        });
+        try {
+            await waitForProcess(childProcess);
+        } finally {
+            childProcess.kill();
+        }
+    });
+
+    step('drop external node storage', async () => {
```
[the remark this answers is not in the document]
slowli: It is, but I feel it will spaghettify the test code even more; as you've written, there's a problem with mixing different levels of abstraction in the test already.

core/tests/recovery-test/tests/genesis-recovery.test.ts
```diff
@@ -0,0 +1,271 @@
...
+
+// FIXME: check consistency checker health once it has acceptable speed
+
+/**
+ * Tests recovery of an external node from scratch.
+ *
+ * Assumptions:
+ *
+ * - Main node is run for the duration of the test.
+ * - "Rich wallet" 0x36615Cf349d7F6344891B1e7CA7C72883F5dc049 is funded on L2 (e.g., as a result of previous integration or load tests).
```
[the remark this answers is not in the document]
slowli: Fixed it so that only funding this address on L1 is required (which is always true if the network was set up using `zk init`).

.github/workflows/ci-core-reusable.yml
```diff
@@ -217,12 +217,28 @@ jobs:
         # We use `yarn` directly because the test launches `zk` commands in both server and EN envs.
         # An empty topmost environment helps avoid a mess when redefining env vars shared between both envs
         # (e.g., DATABASE_URL).
+        #
+        # Since `base_token` doesn't meaningfully influence the test, we use it as a flag for
+        # enabling / disabling tree during pruning.
```
popzxc, line 222: This feels _extremely_ hacky, but I guess it's better than adding even more workflows.

popzxc (approved):

tomg10 (approved):

.github/workflows/ci-core-reusable.yml
slowli: As an alternative: introduce a new matrix variable, but then you'd need to specify all matrix combinations manually (at least I haven't found a way to e.g. tie one matrix variable to another). Would you like it better?

artmakh (approved):

PR #1965, opened 2024-05-16
feat: save writes needed for tree in state keeper

perekopskiy:
## What ❔

State keeper saves data about L1 batch write logs into DB that is needed for tree. Then metadata calculator loads this data.

## Why ❔

Currently, data loading in metadata calculator takes significant time, mostly because it needs enumeration indices for repeated writes and to get those it needs to perform a lot random reads against Postgres. But VM already knows these indices (they are needed to construct pubdata input and are loaded from RocksDB), so it can pass it to state_keeper and data can be persisted improving metadata calculator data load time.

## Checklist


core/lib/dal/src/blocks_dal.rs
```diff
@@ -2159,6 +2172,40 @@ impl BlocksDal<'_, '_> {
...
+
+        let tree_writes = tree_writes
+            .map(|data| {
+                bincode::deserialize(&data).map_err(|err| {
+                    Instrumented::new("get_tree_writes")
+                        .with_arg("l1_batch_number", &l1_batch_number)
+                        .constraint_error(err.into())
+                })
+            })
+            .transpose()?;
```
slowli, line 2205: This can be done more fluently and idiomatically by using `.try_map()` like [e.g. here](https://github.com/matter-labs/zksync-era/blob/33000475b47831fc3791dac338aae4d0e7db25b0/core/lib/dal/src/blocks_dal.rs#L2014).

core/lib/dal/src/blocks_dal.rs
```diff
@@ -2241,6 +2289,31 @@ impl BlocksDal<'_, '_> {
             .context("storage contains neither L2 blocks, nor snapshot recovery info")?;
         Ok(snapshot_recovery.protocol_version)
     }
+
+    pub async fn set_tree_writes(
+        &mut self,
+        l1_batch_number: L1BatchNumber,
+        tree_writes: Vec<TreeWrite>,
+    ) -> DalResult<()> {
```
slowli, line 2297: Dumb question: Why two ways to insert tree writes for an L1 batch? I'd try to retain this one and use it atomically with inserting a batch header if necessary.

core/lib/types/src/storage/writes/mod.rs
```diff
@@ -188,6 +188,19 @@ fn prepend_header(compressed_state_diffs: Vec<u8>) -> Vec<u8> {
     res.to_vec()
 }
 
+/// Struct for storing tree writes in DB.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct TreeWrite {
```
slowli, line 193: Dumb question: This struct looks close to `StateDiffRecord`; correspondingly, the `tree_writes` field looks close to `compressed_state_diffs`. Could the logic implemented in this PR be implemented by saving `compressed_state_diffs` early if it's possible to extract from the VM execution, and deserializing `compressed_state_diffs` in metadata calculator? If so, this would solve the issue of data duplication, and would eliminate potential problems associated with using `bincode` serialization (can unexpectedly break in the future).

core/lib/dal/src/blocks_dal.rs
perekopskiy: `set_tree_writes` was added at the last moment for the test. I agree, it would look better to use `set_tree_writes` everywhere, I will change code

core/lib/types/src/storage/writes/mod.rs
perekopskiy: So, using `compressed_state_diffs` won't work because the compression is lossy, it doesn't keep hashed keys for repeated writes. Uncompression can only be performed if having additional context (leaf index => hashed key mapping) and querying this context from DB is exactly what we want to prevent.
As for the bincode, I'm not sure, maybe it makes sense to add explicit methods like `to_bytes` and `from_bytes` and use them instead of bincode, this way it will be easier to catch breaking changes in the future, WDYT?

slowli: Ah, that's a pity. Explicit byte conversions can subjectively be better (provided the requirement of format stability is explained somewhere), but since we cannot piggyback the existing serialization logic, it could be painful to implement.

core/lib/types/src/storage/writes/mod.rs
```diff
@@ -188,6 +188,78 @@ fn prepend_header(compressed_state_diffs: Vec<u8>) -> Vec<u8> {
...
+    pub fn vec_to_bytes(vec: Vec<Self>) -> Vec<u8> {
+        let size = Self::TREE_WRITE_BYTES_SIZE * vec.len();
+        let mut result = Vec::with_capacity(size);
+
+        for element in vec {
+            element.extend_with_bytes(&mut result);
+        }
+
+        result
+    }
```
itegulov, line 240: I saw that you switched the implementation to use manual (de)serialization but the cause is still unclear to me. I assumed that `bincode` would product efficient implementations for `fixed-hash` types, is that not the case in your experience? Also, even if you can produce more efficient implementation, is there a reason not to implement `Serialize`/`Deserialize` traits instead of ad-hoc functions (there would be a miniscule overhead for vec length but otherwise I expect the behaviour to be identical)?

core/lib/types/src/storage/writes/mod.rs
```diff
@@ -188,6 +188,78 @@ fn prepend_header(compressed_state_diffs: Vec<u8>) -> Vec<u8> {
...
+
+    pub fn vec_from_slice(slice: &[u8]) -> anyhow::Result<Vec<Self>> {
+        if slice.len() % Self::TREE_WRITE_BYTES_SIZE == 0 {
+            let mut result = Vec::with_capacity(slice.len() / Self::TREE_WRITE_BYTES_SIZE);
+            for i in 0..(slice.len() / Self::TREE_WRITE_BYTES_SIZE) {
+                result.push(Self::from_slice(
+                    &slice
+                        [i * Self::TREE_WRITE_BYTES_SIZE..((i + 1) * Self::TREE_WRITE_BYTES_SIZE)],
```
slowli, line 248: Bikeshedding: I'd extract the range start / end to local vars so that it's more readable.

core/lib/types/src/storage/writes/mod.rs
```diff
@@ -188,6 +188,78 @@ fn prepend_header(compressed_state_diffs: Vec<u8>) -> Vec<u8> {
...
+    const TREE_WRITE_BYTES_SIZE: usize = 20 + 32 + 32 + 8;
+
+    fn extend_with_bytes(&self, vec: &mut Vec<u8>) {
+        vec.extend_from_slice(&self.address.0);
+        vec.extend_from_slice(&self.key.0);
+        vec.extend_from_slice(&self.value.0);
+        vec.extend_from_slice(&self.leaf_index.to_be_bytes());
+    }
+
+    pub fn from_slice(slice: &[u8]) -> anyhow::Result<Self> {
```
slowli, line 214: Nit: Does this method need to be public? It seems to be only used from `vec_from_slice`.

core/lib/types/src/storage/writes/mod.rs
```diff
@@ -188,6 +188,78 @@ fn prepend_header(compressed_state_diffs: Vec<u8>) -> Vec<u8> {
...
+        } else {
+            anyhow::bail!(
+                "Incorrect slice len, expected: {}, got {}",
+                Self::TREE_WRITE_BYTES_SIZE,
+                slice.len()
+            );
+        }
+    }
+
+    pub fn vec_to_bytes(vec: Vec<Self>) -> Vec<u8> {
```
slowli, line 231: Nit: You don't need ownership here; using `&[Self]` should suffice.

core/lib/types/src/storage/writes/mod.rs
```diff
@@ -188,6 +188,78 @@ fn prepend_header(compressed_state_diffs: Vec<u8>) -> Vec<u8> {
...
+    /// Value written.
+    pub value: H256,
+    /// Leaf index of the slot.
+    pub leaf_index: u64,
+}
+
+impl TreeWrite {
+    const TREE_WRITE_BYTES_SIZE: usize = 20 + 32 + 32 + 8;
+
+    fn extend_with_bytes(&self, vec: &mut Vec<u8>) {
```
slowli, line 207: Nit: Could make sense to add a roundtrip unit test.

core/node/metadata_calculator/src/helpers.rs
```diff
@@ -550,8 +550,43 @@ impl L1BatchWithLogs {
...
+
+        let storage_logs = if let Some(tree_writes) = tree_writes {
+            // If tree writes are present in DB then simply use them.
+            let writes = tree_writes.into_iter().map(|tree_write| {
+                let storage_key =
+                    StorageKey::new(AccountTreeId::new(tree_write.address), tree_write.key);
+                TreeInstruction::write(storage_key, tree_write.leaf_index, tree_write.value)
+            });
+            let reads = protective_reads.into_iter().map(TreeInstruction::Read);
+            writes.chain(reads).collect::<Vec<_>>()
```
slowli, line 568: Are tree instructions properly sorted in this case? AFAIU, they should be sorted by `StorageKey`, which doesn't look to be true if `protective_reads` aren't empty.

core/node/state_keeper/src/io/seal_logic/mod.rs
```diff
@@ -109,6 +110,96 @@ impl UpdatesManager {
             current_l1_batch_number = self.l1_batch.number
         );
 
+        let progress = L1_BATCH_METRICS.start(L1BatchSealStage::FilterWrittenSlots);
```
slowli, line 113: Dumb question: Can this be implemented as a separate `StateKeeperOutputHandler`? IMO, if it would work, it'd be cleaner. Or would it be too complex to reason about non-atomicity w/ L1 batch header insertion? (AFAIU, batch sealing is synchronous now, and `StateKeeperOutputHandler`s are executed in `OutputHandler` one by one, so there shouldn't be an issue of not having an expected `l1_batches` table entry. OTOH, non-atomicity could mean that the metadata calculator would see an L1 batch w/o tree writes yet added.)

core/lib/dal/src/blocks_dal.rs
```diff
@@ -469,6 +470,7 @@ impl BlocksDal<'_, '_> {
         Ok(())
     }
 
+    #[allow(clippy::too_many_arguments)]
```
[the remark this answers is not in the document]
perekopskiy: I added 8th parameter but reverted later. I will remove this `allow`

core/lib/types/src/storage/writes/mod.rs
perekopskiy: Yes, default serialization for fixed-hash types always serializes to strings, meaning 2x bytes then needed, see https://github.com/paritytech/parity-common/blob/master/primitive-types/impls/serde/src/serialize.rs.

As for `Serialize/Deserialize` impls, makes sense, I will try that

core/node/metadata_calculator/src/helpers.rs
perekopskiy: They are not sorted indeed, nice catch!

core/node/state_keeper/src/io/seal_logic/mod.rs
perekopskiy: In general, I like the idea
> OTOH, non-atomicity could mean that the metadata calculator would see an L1 batch w/o tree writes yet added

But I'm not sure how to overcome this, for old batches tree writes won't be present, so we can poll tree_writes in metadata with some reasonable timeout if the batch is the last sealed one, WDYT?

itegulov (approved):
Looks good on my side

core/node/state_keeper/src/io/seal_logic/mod.rs
perekopskiy: I've opened a separate PR with new output handler, please check https://github.com/matter-labs/zksync-era/pull/2018. Part with `wait_for_tree_writes` looks a bit clunky to me but I don't know how to improve it

core/node/metadata_calculator/src/helpers.rs
```diff
@@ -550,8 +553,83 @@ impl L1BatchWithLogs {
...
+        }))
+    }
+
+    #[allow(clippy::needless_pass_by_ref_mut)]
+    async fn wait_for_tree_writes(
+        connection: &mut Connection<'_, Core>,
+        l1_batch_number: L1BatchNumber,
+    ) -> anyhow::Result<Option<Vec<TreeWrite>>> {
+        let interval = Duration::from_millis(50);
+        let timeout = Duration::from_secs(5);
```
slowli, line 607: Bikeshedding: IMO, these values would work better as constants.

core/lib/dal/src/storage_logs_dedup_dal.rs
```diff
@@ -172,28 +172,45 @@ impl StorageLogsDedupDal<'_, '_> {
         .map(|max| max as u64))
     }
 
-    /// Returns the maximum enumeration index assigned in a specific L1 batch.
-    pub async fn max_enumeration_index_for_l1_batch(
+    pub async fn max_enumeration_index_in_l1_batch(
```
slowli, line 175: Nit: I'd document what this new version of the method does. Also, if I understand its semantics correctly ("get the max enumeration index by a certain L1 batch"), IMO, `max_enumeration_index_by_l1_batch` or `*_at_l1_batch` would be a better name.

core/lib/dal/src/storage_logs_dedup_dal.rs
```diff
@@ -172,28 +172,45 @@ impl StorageLogsDedupDal<'_, '_> {
...
         &mut self,
         l1_batch_number: L1BatchNumber,
     ) -> DalResult<Option<u64>> {
-        let row = sqlx::query!(
+        let Some(max_l1_batch_number) = sqlx::query!(
```
slowli, line 179: Dumb question: Can't this be a single query, e.g. using a subquery? AFAIU, using a "plain" query won't do because it would involve a full `initial_writes` scan at the worst case.

core/node/metadata_calculator/src/helpers.rs
```diff
@@ -550,8 +553,83 @@ impl L1BatchWithLogs {
...
+        load_changes_latency.observe();
+
+        Ok(Some(Self {
+            header,
+            storage_logs,
+            mode,
+        }))
+    }
+
+    #[allow(clippy::needless_pass_by_ref_mut)]
```
slowli, line 601: Just to understand: Is this a false positive by the Rust compiler? Could make sense to add a comment if so.

slowli (commented):
Generally, looks really good 👍

core/node/metadata_calculator/src/helpers.rs
itegulov: AFAIU it might because of https://github.com/rust-lang/rust-clippy/issues/11179

Sort of off-topic but we should consider bumping rustc version one of these days if there is no strong reason to keep it at `2023-08-20`

core/lib/dal/src/storage_logs_dedup_dal.rs
perekopskiy: Yes, all correct, I will refactor to use subquery

core/node/metadata_calculator/src/helpers.rs
perekopskiy: Yes, exactly. I will add a comment so it will look like `#[allow(clippy::needless_pass_by_ref_mut)] // false positive`, it's the same for 3 more functions in our code.

PR #1973, opened 2024-05-17
fix(merkle-tree): Fix tree API health check status

slowli:
## What ❔

- Makes health status "affected" if the tree API is not available.
- Distinguishes between "tree API not enabled on the node" and "tree API temporarily unavailable" errors.

## Why ❔

Right now, the status is "not ready", meaning that the entire app health status is "not ready" as well. This makes API servers excluded from serving traffic, which is not desirable (they can still serve all of requests other than `zks_getProof`).

## Checklist

### DIFF
--- a/core/lib/web3_decl/src/error.rs
+++ b/core/lib/web3_decl/src/error.rs
@@ -39,11 +39,12 @@ pub enum Web3Error {
     LogsLimitExceeded(usize, u32, u32),
     #[error("invalid filter: if blockHash is supplied fromBlock and toBlock must not be")]
     InvalidFilterBlockHash,
-    #[error("Not implemented")]
-    NotImplemented,
-    /// This covers both transient unavailability (e.g., due to connection issues) and unavailability
-    /// caused by node configuration.
-    #[error("Tree API is not available")]
+    /// Weaker form of a "method not found" error; the method implementation is technically present,
+    /// but the node configuration prevents the method from functioning.
+    #[error("Method not implemented")]
+    MethodNotImplemented,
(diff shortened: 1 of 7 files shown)


core/lib/web3_decl/src/error.rs
```diff
@@ -39,10 +39,12 @@ pub enum Web3Error {
     LogsLimitExceeded(usize, u32, u32),
     #[error("invalid filter: if blockHash is supplied fromBlock and toBlock must not be")]
     InvalidFilterBlockHash,
-    #[error("Not implemented")]
-    NotImplemented,
-
-    #[error("Tree API is not available")]
+    /// Weaker form of a "method not found" error; the method implementation is technically present,
+    /// but the node configuration prevents the method from functioning.
+    #[error("Method not implemented")]
```
popzxc, line 44: Will this change be breaking if someone depends on the exact error message?

popzxc (approved):

perekopskiy (approved):

core/lib/web3_decl/src/error.rs
slowli: Yep, I wanted to comment on this as well. Do we know whether anyone depends on exact error messages returned by the RPC server? Obviously, changing the message isn't critical, so if this is a concern, I can change it back.

popzxc: Well, I think generally we would expect people _not_ to depend on the error message (especially given that there are error codes), so probably it's fine to do it this time.

PR #1978, opened 2024-05-17
fix(en): Minor node fixes

slowli:
## What ❔

Fixes minor issues with the EN:

- Transient error detection for RPC clients is suboptimal, leading to what really is a transient error crashing consistency checker.
- Since refactoring signal handling, `ManagedTasks::wait_single()` on EN may inappropriately log task termination errors even after a signal is being received.
- Since recent refactoring (?), block fetcher doesn't tag its L2 client.
- Block fetcher may produce many rate limiting logs, which currently have WARN level.
- Initializing L2 client multiple times (e.g., in a load test) produces many WARN logs.

## Why ❔

Improves EN UX.

## Checklist

### DIFF
--- a/core/lib/web3_decl/src/client/mod.rs
+++ b/core/lib/web3_decl/src/client/mod.rs
@@ -196,7 +196,7 @@ impl<Net: Network, C: ClientBase> Client<Net, C> {
             origin,
             &stats,
         );
-        tracing::warn!(
+        tracing::debug!(
             network = network_label,
             component = self.component_name,
             %origin,
--- a/core/bin/external_node/src/main.rs
+++ b/core/bin/external_node/src/main.rs
@@ -1019,9 +1019,12 @@ async fn run_node(
 
     let mut tasks = ManagedTasks::new(task_handles);
     tokio::select! {
-        () = tasks.wait_single() => {},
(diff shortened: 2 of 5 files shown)


core/lib/web3_decl/src/client/mod.rs
```diff
@@ -196,7 +196,7 @@ impl<Net: Network, C: ClientBase> Client<Net, C> {
             origin,
             &stats,
         );
-        tracing::warn!(
+        tracing::debug!(
```
slowli, line 199: @tomg10 You were previously advocating for the WARN level here. The thing is: as currently implemented (i.e., when this is used as the only rate limiter), this will spam a lot of warnings when a node is catching up using JSON-RPC communication. Hence, this change.

popzxc (approved):

core/lib/web3_decl/src/client/mod.rs
tomg10: I wanted to have it as warn as I don't think any component should rely on the fact that there exists a rate limitter, i.e. be implemented/configured in a way that doesn't send too many requests

tomg10: for instance as this may throttle other components as far as I remember

PR #1984, opened 2024-05-17
perf(commitment-generator): Run commitment generation for multiple batches in parallel

slowli:
## What ❔

Generates commitments with configurable parallelism.

## Why ❔

- As long as produced commitments are persisted in the same order as previously (not even atomically), there's still a cursor to be used by the commitment generator.
- Can provide speed up during node recovery (esp. if treeless mode is used). Also can speed up some integration tests.

## Checklist


core/lib/dal/src/blocks_dal.rs
```diff
@@ -190,6 +192,34 @@ impl BlocksDal<'_, '_> {
         Ok(row.map(|row| L1BatchNumber(row.number as u32)))
     }
 
+    /// Gets a number of the last L1 batch that is ready for commitment generation (i.e., doesn't have commitment
+    /// yet, and has tree data).
+    pub async fn get_last_l1_batch_ready_for_commitment_generation(
```
slowli, line 197: Does a separate query make sense, or would it make sense to extend `get_next_l1_batch_ready_for_commitment_generation()` to return `ops::RangeInclusive` and use a single DB query?

core/bin/external_node/src/main.rs
```diff
@@ -363,14 +363,13 @@ async fn run_core(
...
     app_health.insert_component(batch_status_updater.health_check())?;
 
-    let commitment_generator_pool = singleton_pool_builder
-        .build()
-        .await
-        .context("failed to build a commitment_generator_pool")?;
-    let commitment_generator = CommitmentGenerator::new(
-        commitment_generator_pool,
+    let mut commitment_generator = CommitmentGenerator::new(
+        connection_pool.clone(),
```
slowli, line 367: Does using a common pool make sense, or would it make sense to create a separate pool for the generator? Realistically, it wouldn't require many connections unless a node is catching up.

core/node/commitment_generator/src/lib.rs
```diff
@@ -21,34 +21,53 @@ use zksync_utils::h256_to_u256;
...
             connection_pool,
             health_updater: ReactiveHealthCheck::new("commitment_generator").1,
             commitment_mode,
+            parallelism: Self::default_parallelism(),
         }
     }
 
+    fn default_parallelism() -> NonZeroU32 {
+        // Leave at least one core free to handle other blocking tasks. `unwrap()`s are safe by design.
+        let cpus = u32::try_from(num_cpus::get().saturating_sub(1).clamp(1, 16)).unwrap();
```
perekopskiy, line 62: `process_batch` spawns 2 blocking tasks: for `events_queue` and `bootloader_memory`. I'm wondering how that will work when parallelism is set to `num_cpus::get().saturating_sub(1)`. I think it should be fine, especially given that events task usually takes up to 1s, while bootloader task takes around 3s for v22 and 6s for v24; but still want to double-check there won't be issues because of that

perekopskiy (commented):
Looks good!

core/lib/dal/src/blocks_dal.rs
perekopskiy: IMO, separate query makes sense

core/bin/external_node/src/main.rs
perekopskiy: I think it's ok. IIRC currently in external node we use common pool for almost everything

core/node/commitment_generator/src/lib.rs
```diff
@@ -335,29 +381,72 @@ impl CommitmentGenerator {
...
+
+    /// Runs this commitment generator indefinitely. It will process L1 batches added to the database
+    /// processed by the Merkle tree (or a tree fetcher), with a previously configured max parallelism.
     pub async fn run(self, stop_receiver: watch::Receiver<bool>) -> anyhow::Result<()> {
+        tracing::info!(
+            "Starting commitment generator with mode {:?} and parallelism {}",
+            self.commitment_mode,
+            self.parallelism
+        );
+        if self.connection_pool.max_size() < self.parallelism.get() {
```
perekopskiy, line 422: Main node runs CommitmentGenerator using a singleton connection pool. While it shouldn't affect anything as under normal operation it will process batches one by one, the log will be emitted. Does it make sense to provide larger pool or set parallelism=1?

slowli: IMO, providing a larger pool would be a preferred option. Is this permissible from the infra perspective (i.e., is there a risk that this makes the DB to run out of connections)?

core/node/commitment_generator/src/lib.rs
slowli: What kind of issues do you have in mind? Since parallelism only controls the max number of L1 batches processed at the same time (not, say, the number of blocking threads for Tokio), it doesn't look possible to deadlock generation if the parallelism factor is too large.

perekopskiy (dismissed):

perekopskiy (approved):

PR #1993, opened 2024-05-20
feat(proof_data_handler): add new endpoints to the TEE prover interface API

pbeza:
## What ❔

This PR introduces three new endpoints to the prover interface API:
1. `/tee/proof_inputs` - for fetching input data for the TEE verifier. It is intended for TEE workers to obtain a batch to process.
2. `/tee/submit_proofs/<l1_batch_number>` - for submitting TEE proof.
3. `/tee/register_attestation` - for registering TEE attestation.

The first two introduced API endpoints correspond to the existing, analogous `/proof_generation_data` and `/submit_proof/<l1_batch_number>` endpoints used for the ZK proofs.

The state of batches (e.g., _proven_, _taken_, etc.) is tracked in the database. The `TeeVerifierInputProducer` generates serialized TEE prover inputs, which are then stored in the object store.

To run the unit tests, you need to use the following command: `zk test rust --package zksync_proof_data_handler --lib tests`. Running `cargo test` directly fails because the `zk` command sets up an additional database for testing purposes.

To test it manually, run the ZK server with the command:
```
zk server --components proof_data_handler --use-node-framework
```
and then send an HTTP request:
- to get TEE verifier input data:
  ```
  curl -X POST -H "Content-Type: application/json" --data-raw "{}" -vvv http://127.0.0.1:3320/tee/proof_inputs
  ```
  To inspect the database for the TEE verifier input data jobs, run:
  ```
  $ PGPASSWORD='notsecurepassword' psql -h 127.0.0.1 -p 5432 -U postgres
  # \c zksync_local
  # SELECT * FROM tee_verifier_input_producer_jobs;
  ```
- register TEE attestation:
  ```
  curl -X POST -H "Content-Type: application/json" --data-raw '{ "attestation": [ 4, 3, 2, 1, 0 ], "pubkey": [ 5, 6, 7, 8, 9 ] }' -vvv http://127.0.0.1:3320/tee/register_attestation
  ```
  To inspect the database for the TEE attestations, run:
  ```
  $ PGPASSWORD='notsecurepassword' psql -h 127.0.0.1 -p 5432 -U postgres
  # \c zksync_local
  # SELECT * FROM tee_attestations;
  ```
- to submit TEE proof:
  ```
  curl -X POST -H "Content-Type: application/json" --data-raw '{ "signature": [ 0, 1, 2, 3, 4 ], "pubkey": [ 5, 6, 7, 8, 9 ], "proof": [ 10, 11, 12, 13, 14 ] }' -vvv http://127.0.0.1:3320/tee/submit_proofs/1
  ```
  To inspect the database for the TEE proofs, run:
  ```
  $ PGPASSWORD='notsecurepassword' psql -h 127.0.0.1 -p 5432 -U postgres
  # \c zksync_local
  # SELECT * FROM tee_proof_generation_details;
  ```

## Why ❔

This PR contributes to the effort outlined in the docs:
- https://www.notion.so/matterlabs/2FA-for-zk-rollups-with-TEEs-a2266138bd554fda8846e898fef75131?pvs=4
- https://www.notion.so/matterlabs/Proof-2F-verification-with-SGX-5fca2c619dd147938971cc00ae53e2b0?pvs=4

## Checklist


pbeza:
@thomasknauth @haraldh @RomanBrodetski This PR is ready for the first round of reviews. You can find the details on how to run and test it manually or automatically [here](https://github.com/matter-labs/zksync-era/pull/1993#issue-2306384443). Feel free to comment and add other reviewers. Thanks.

pbeza:
Just came across [a doc](https://www.notion.so/matterlabs/External-Proving-Partner-2d34463ee0da4f0fb31dd008e3855870?pvs=4#fb56f3a9ac0a46ebad883d85553c483e) suggesting we will introduce the `/proof_generation_data/verified` endpoint to provide proofs that Matter Labs has already created. AFAIK, we haven't decided whether to add this endpoint yet.

core/node/proof_data_handler/src/tee_request_processor.rs
```diff
@@ -0,0 +1,105 @@
...
+        match payload {
+            SubmitTeeProofRequest::Proof(proof) => {
+                println!(
+                    "Received proof {:?} for block number: {:?}",
+                    proof, l1_batch_number
+                );
+                dal.save_proof_artifacts_metadata(
+                    l1_batch_number,
+                    &proof.signature,
+                    &proof.signature,
```
thomasknauth, line 89: This should be `pubkey`, I guess.

core/node/proof_data_handler/src/lib.rs
```diff
@@ -46,18 +81,25 @@ pub async fn run_server(
...
-            }
-            tracing::info!("Stop signal received, proof data handler server is shutting down");
-        })
-        .await
-        .context("Proof data handler server failed")?;
-    tracing::info!("Proof data handler server shut down");
-    Ok(())
+        )
+        .route(
+            "/tee_proof_generation_data",
```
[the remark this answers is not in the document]
pbeza: I named it this way based on the [existing](https://github.com/matter-labs/zksync-era/blob/1c5229cdab5dce07308452aa9cea1b0f53f7a4b9/core/node/proof_data_handler/src/lib.rs#L29) `/proof_generation_data` ZK-specific endpoint, but I prefer your shortened version more. It's fixed now.

core/node/proof_data_handler/src/tee_request_processor.rs
pbeza: Fixed.

core/lib/dal/migrations/20240523085604_add_tee_proof_generation_details_table.up.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE IF NOT EXISTS tee_proof_generation_details
+(
+    l1_batch_number         BIGINT PRIMARY KEY REFERENCES tee_verifier_input_producer_jobs (l1_batch_number) ON DELETE CASCADE,
+    status                  TEXT      NOT NULL,
+    signature               BYTEA,
+    pubkey                  BYTEA,
+    attestation             BYTEA,
```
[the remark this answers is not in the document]
pbeza: [Done](https://github.com/matter-labs/zksync-era/blob/056ba0e2299af622c34c3e984c918b7306b4b8ae/core/lib/dal/migrations/20240523085604_add_tee_proof_generation_details_table.up.sql#L1-L6). ✅

core/node/proof_data_handler/src/lib.rs
```diff
@@ -46,18 +82,35 @@ pub async fn run_server(
...
+                move |payload: Json<TeeProofGenerationDataRequest>| async move {
+                    get_tee_proof_gen_processor
+                        .get_proof_generation_data(payload)
+                        .await
+                },
+            ),
+        )
+        .route(
+            "/submit_tee_proof/:l1_batch_number",
+            post(
```
popzxc, line 98: Given that TEE is optional, I think it probably makes sense to:

a) Rework the routing scheme so that instead of `/tee_proof_inputs`, `/submit_tee_proof`, and `/register_tee_attestation` we have `/tee/proof_inputs`, `/tee/submit_proofs` and `/tee/register_attestation` (e.g. we introduce a `/tee` namespace instead of maintaining a flat structure). This also feels more RESTful.
b) Make it configurable (e.g. it should be possible to run proof data handler _without_ tee endpoints, and tee endpoints should be opt-in).

popzxc: (If we go this way the next logical thing would be to introduce something like this for the ZK proofs, e.g. `/zk/proof_generation_data` etc), but it's certainly a bigger task and out of scope here.
cc @RomanBrodetski @EmilLuta

core/node/proof_data_handler/Cargo.toml
```diff
@@ -10,13 +10,26 @@ keywords.workspace = true
...
+zksync_utils.workspace = true
 anyhow.workspace = true
 axum.workspace = true
 tokio.workspace = true
+tracing.workspace = true
+
+[dev-dependencies]
+http-body-util = "0.1.0"
+hyper = { version = "0.14", features = ["full"] }
+mime = "0.3.16"
```
popzxc, line 29: Please only use workspace deps (e.g. add crates there first, and refer to them as `.workspace = true`).

core/lib/prover_interface/tests/job_serialization.rs
```diff
@@ -76,3 +78,105 @@ async fn test_final_proof_deserialization() {
...
+                    "x": [ 0, 0, 0, 0 ],
+                    "y": [ 1, 0, 0, 0 ],
+                    "infinity": true
+                }
+            },
+            "protocol_version": "0.25.10"
+        }
+    }"#;
+    let decoded_obj: SubmitProofRequest = serde_json::from_str(&encoded_obj).unwrap();
+    println!("{}", encoded_obj);
```
popzxc, line 153: We should not use `println` in tests.

core/node/proof_data_handler/src/tee_request_processor.rs
```diff
@@ -0,0 +1,122 @@
...
+        &self,
+        request: Json<TeeProofGenerationDataRequest>,
+    ) -> Result<Json<TeeProofGenerationDataResponse>, RequestProcessorError> {
+        tracing::info!("Received request for proof generation data: {:?}", request);
+
+        let l1_batch_number_result = self
+            .pool
+            .connection()
+            .await
+            .unwrap()
```
popzxc, line 49: You should be careful with unwraps, you can simply propagate the error here (same below).

core/node/proof_data_handler/Cargo.toml
```diff
@@ -10,13 +10,26 @@ keywords.workspace = true
...
-
-tracing.workspace = true
+zksync_utils.workspace = true
 anyhow.workspace = true
 axum.workspace = true
 tokio.workspace = true
+tracing.workspace = true
+
+[dev-dependencies]
+http-body-util = "0.1.0"
```
popzxc, line 27: Also this dependency seems to be unused.

core/node/proof_data_handler/src/tee_request_processor.rs
```diff
@@ -0,0 +1,122 @@
...
+        let l1_batch_number = L1BatchNumber(l1_batch_number);
+        let mut connection = self.pool.connection().await.unwrap();
+        let mut dal = connection.tee_proof_generation_dal();
+
+        match payload {
+            SubmitTeeProofRequest::Proof(proof) => {
+                println!(
+                    "Received proof {:?} for block number: {:?}",
+                    proof, l1_batch_number
+                );
```
popzxc, line 86: No `println` in production code either, please use `tracing`.

core/node/proof_data_handler/src/tests.rs
```diff
@@ -0,0 +1,247 @@
...
+                .header(http::header::CONTENT_TYPE, mime::APPLICATION_JSON.as_ref())
+                .body(req_body)
+                .unwrap(),
+        )
+        .await
+        .unwrap();
+
+    assert_eq!(response.status(), StatusCode::OK);
+
+    let body = hyper::body::to_bytes(response.into_body()).await.unwrap();
```
popzxc, line 109: This function [seems to be deprecated](https://docs.rs/hyper/0.14.29/src/hyper/body/to_bytes.rs.html#50), and it suggests to use `.collect().await?.to_bytes()` instead. Probably would also make it possible to avoid direct dependency.

core/node/proof_data_handler/src/tests.rs
```diff
@@ -0,0 +1,247 @@
...
+        },
+        L1BatchCommitmentMode::Rollup,
+    );
+    let req_body = Body::from(serde_json::to_vec(&json!({})).unwrap());
+    let response = app
+        .oneshot(
+            Request::builder()
+                .method(Method::POST)
+                .uri("/tee_proof_inputs")
+                .header(http::header::CONTENT_TYPE, mime::APPLICATION_JSON.as_ref())
```
popzxc, line 100: I think a string `"application/json"` would be as declarative, and would let us avoid a direct dependency on `mime`.

core/node/proof_data_handler/src/tests.rs
```diff
@@ -0,0 +1,247 @@
...
+
+    // this should fail because we haven't saved the attestation for the pubkey yet
+
+    let response = send_submit_tee_proof_request(&app, &uri, &tee_proof_request).await;
+    assert_eq!(response.status(), StatusCode::BAD_GATEWAY);
+
+    // save the attestation for the pubkey
+
+    if let SubmitTeeProofRequest::Proof(ref proof) = tee_proof_request {
+        let valid_until = Utc::now() + Duration::days(42);
```
haraldh, line 160: IMHO, we agreed on not getting the collateral in the SGX worker and therefore `valid_until` is unknown and not needed.

haraldh (changes requested):

core/node/proof_data_handler/src/tests.rs
pbeza: How did you know? Was it recommended by a linter that you are using? If so, can you recommend one?

popzxc: Nope, I just _really_ do not like adding new top-level dependencies, so I went to the source to check whether the implementation is trivial and can be vendored instead 😅

core/node/proof_data_handler/src/tee_request_processor.rs
pbeza: Added accidentally (perhaps I just allowed GitHub Copilot to autocomplete). I think there may be no place in the code where `println!` makes sense. If so, does it make sense to add a linter rule to flag `println!` as an error/warning?

core/node/proof_data_handler/Cargo.toml
popzxc: Brain it is 🤖 
Whenever I see a new dep, I check a) how (if) it's used, and b) whether we can avoid that.

popzxc: OTOH [cargo-udeps](https://github.com/est31/cargo-udeps) exists, but last time I used it it had plenty of false negatives. It doesn't handle workspace deps well, I guess.

core/node/proof_data_handler/src/tee_request_processor.rs
popzxc: We use `println` in CLI-like tools (e.g. where we do not expect logging to be configured) -- you may check the codebase for usages.

core/node/proof_data_handler/src/tests.rs
pbeza: [Fixed](https://github.com/matter-labs/zksync-era/pull/1993/commits/79611757437dfa02095ff7fede8251bd6efa6572) ✅

core/node/proof_data_handler/src/tee_request_processor.rs
pbeza: [Fixed](https://github.com/matter-labs/zksync-era/pull/1993/commits/054cdcf8ff3ba6e9b1868a5788a40cfcfff3d78d) ✅

core/lib/prover_interface/tests/job_serialization.rs
pbeza: [Fixed](https://github.com/matter-labs/zksync-era/pull/1993/commits/4103ae01a0237930818b2cc88b1164613bf3f611) ✅

core/node/proof_data_handler/src/tee_request_processor.rs
pbeza: [Fixed](https://github.com/matter-labs/zksync-era/pull/1993/commits/1a03ce0b4a6eb7cc9ab82c9492fd4835e8d63e29) ✅

I'm not fully satisfied with the way I implemented it:
https://github.com/matter-labs/zksync-era/blob/1a03ce0b4a6eb7cc9ab82c9492fd4835e8d63e29/core/node/proof_data_handler/src/tee_request_processor.rs#L45-L49

This would probably be better:
```rust
        let mut connection = self
            .pool
            .connection()
            .await
            .map_err(|e| RequestProcessorError::Sqlx(*e.inner()))?;
```
but:
```
.map_err(|e| RequestProcessorError::Sqlx(*e.inner()))?;
                                         ^^^^^^^^^^ move occurs because value has type `zksync_dal::SqlxError`, which does not implement the `Copy` traitdoes not implement the `Copy` trait
```

If you have a better idea on how to handle it, I welcome your suggestions.

pbeza:
@slowli @popzxc @RomanBrodetski I put quite an effort into implementing tests using `ObjectStoreFactory::mock()`. I just rebased my branch on the latest `origin/main` and noticed that you removed `mock()` @slowli. 😢 Quote from https://github.com/matter-labs/zksync-era/commit/5c0396441ea5383da322670a57ca0392d2a647fa git message:
```
(...)
- Expose GCS, file-based and mock object stores directly and remove
ability to create mock object stores from `ObjectStoreFactory`
(...)
```
Any recommendation?

popzxc:
@pbeza as commit message mentions, mock storage factory is now exposed directly, e.g. as `MockObjectStore`, you can see [sample usage](https://github.com/matter-labs/zksync-era/pull/2138/files#diff-038b6d667b931521d1903ce8990b3af7dc04f83dce380bdfaaa69f2d48ed9787R14) in the same PR as you linked.

core/node/proof_data_handler/Cargo.toml
pbeza: [Fixed](https://github.com/matter-labs/zksync-era/pull/1993/commits/a96bf714791796307aab5fe2733267af78a138d5) ✅

core/node/proof_data_handler/Cargo.toml
pbeza: Does it make sense to add a linter rule to explicitly disallow adding dependencies other than workspace dependencies? You did some heavy lifting in #1485, and I'm afraid new developers (like me) might unknowingly undo the effort unless there is an automated way to prevent them from doing so.

popzxc: If you know the way to do it -- please do! Any PRs that improve the maintainability of the codebase are always welcome.

core/node/proof_data_handler/src/tests.rs
pbeza: This turned out to be tricky to resolve ([here](https://github.com/matter-labs/zksync-era/pull/1993/commits/581efcdac4fa1eea7dddc5ef231f385cfb17d258) is my attempt to fix this).

The problem boils down to the fact that we are using an old version of the `http` crate. Is there any particular reason for this?
https://github.com/matter-labs/zksync-era/blob/bad5a6c0ec2e166235418a2796b6ccf6f8b3b05f/Cargo.toml#L118
That is why I end up with the following error:
```
error[E0599]: `http_body::combinators::box_body::UnsyncBoxBody<axum::body::Bytes, axum::Error>` is not an iterator
   --> core/node/proof_data_handler/src/tests.rs:109:37
    |
109 |     let body = response.into_body().collect().await.unwrap().to_bytes();
    |                                     ^^^^^^^ `http_body::combinators::box_body::UnsyncBoxBody<axum::body::Bytes, axum::Error>` is not an iterator
    |
   ::: /home/patryk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/http-body-0.4.5/src/combinators/box_body.rs:15:1
    |
15  | pub struct UnsyncBoxBody<D, E> {
    | ------------------------------ doesn't satisfy `_: Iterator`
    |
    = note: the following trait bounds were not satisfied:
            `http_body::combinators::box_body::UnsyncBoxBody<axum::body::Bytes, axum::Error>: Iterator`
            which is required by `&mut http_body::combinators::box_body::UnsyncBoxBody<axum::body::Bytes, axum::Error>: Iterator`

For more information about this error, try `rustc --explain E0599`.
```

So, we need to either leave it as-is, using a deprecated function, or upgrade the `http` crate. I'm not sure if there are any other alternatives.

popzxc: OK, here's how to solve this:

1. You added `hyper 1.3.1` to deps, while we transitively depend on `0.14.29` (can be checked in `Cargo.lock`) => Change version in `Cargo.toml`.
2. The method `collect` comes from `HttpBody` trait, not from `Iterator`, as linked above => import said trait via `use hyper::body::HttpBody`.
3. The method was added to `http-body` only in 0.4.6, while we currently depend on 0.4.5 => granularly update via `cargo update http-body --precise 0.4.6`.
4. The code I've copy-pasted from suggestion used `?` to propagate error, while your test doesn't return `Result` => change to `unwrap`.

After doing this I've managed to get your code to compile.

pbeza: That was really helpful! Thank you! [Fixed](https://github.com/matter-labs/zksync-era/pull/1993/commits/6870aa49f858c830874e8c53e1a2b3488dc3a38b). ✅

core/node/proof_data_handler/src/tests.rs
pbeza: [Fixed](https://github.com/matter-labs/zksync-era/pull/1993/commits/c0414f269f79cc8b67931b4d79c9b3b40820187f) ✅

core/node/proof_data_handler/src/lib.rs
pbeza: @popzxc Do you think TEE endpoints/code should be optional at the compilation (`#[cfg(feature = "tee")]`), or at the runtime level? (I presume runtime)

popzxc: Runtime is fine.

pbeza: [Fixed](https://github.com/matter-labs/zksync-era/commit/dd719e16f0bfe6f605ef939c9566642514df3f57)

pbeza:
@popzxc All the issues resolved and CI passing. 🫡 Ready to merge?

core/node/proof_data_handler/src/tee_request_processor.rs
```diff
@@ -0,0 +1,134 @@
...
+                    l1_batch_number,
+                    &proof.signature,
+                    &proof.pubkey,
+                    &proof.proof,
+                    TeeType::Sgx,
+                )
+                .await
+                .map_err(RequestProcessorError::Sqlx)?;
+            }
+            SubmitTeeProofRequest::SkippedProofGeneration => {
```
RomanBrodetski, line 101: Is this logic actually used? I was thinking we don't even create jobs when TEEs are not enabled.

core/node/metadata_calculator/src/updater.rs
```diff
@@ -151,6 +151,10 @@ impl TreeUpdater {
                     .proof_generation_dal()
                     .insert_proof_generation_details(l1_batch_number, object_key)
                     .await;
+                storage
+                    .tee_proof_generation_dal()
+                    .insert_tee_proof_generation_details(l1_batch_number)
+                    .await;
```
popzxc, line 157: I'm not really aware of TEE lifecycle, but it looks a bit suspicious that we both call `create_tee_verifier_input_producer_job` and `insert_tee_proof_generation_details` here.
A few questions:
1. What details are being inserted? I assume that it's a copy-paste from proof generation dal, but there we provide an object key for each batch. Here we only seem to create a job -- so a name like `insert_tee_proof_generation_job` would probably reflect the nature of action better.
2. Would TEE proof generation be blocked until tee verifier input is produced? If so, wouldn't it be better for the tee verifier input producer to create a job once the input is produced?

core/node/proof_data_handler/Cargo.toml
```diff
@@ -10,13 +10,24 @@ keywords.workspace = true
 categories.workspace = true
 
 [dependencies]
+vm_utils.workspace = true
```
popzxc, line 13: Is it used anywhere?

core/node/proof_data_handler/Cargo.toml
```diff
@@ -10,13 +10,24 @@ keywords.workspace = true
...
 zksync_config.workspace = true
 zksync_dal.workspace = true
 zksync_object_store.workspace = true
 zksync_prover_interface.workspace = true
+zksync_tee_verifier.workspace = true
 zksync_types.workspace = true
-
-tracing.workspace = true
+zksync_utils.workspace = true
```
popzxc, line 20: Same

core/node/proof_data_handler/src/lib.rs
```diff
@@ -2,29 +2,65 @@ use std::{net::SocketAddr, sync::Arc};
...
+    connection_pool: ConnectionPool<Core>,
+    config: ProofDataHandlerConfig,
+    commitment_mode: L1BatchCommitmentMode,
+) -> Router {
+    let get_tee_proof_gen_processor =
+        TeeRequestProcessor::new(blob_store.clone(), connection_pool.clone(), config.clone());
+    let submit_tee_proof_processor = get_tee_proof_gen_processor.clone();
+    let register_tee_attestation_processor = get_tee_proof_gen_processor.clone();
+    let get_proof_gen_processor =
+        RequestProcessor::new(blob_store, connection_pool, config.clone(), commitment_mode);
```
popzxc, line 60: TEE objects are being constructed even if `config.tee_support` is `false`. Is this intentional?

core/node/proof_data_handler/src/lib.rs
```diff
@@ -48,16 +84,38 @@ pub async fn run_server(
...
-                tracing::warn!("Stop signal sender for proof data handler server was dropped without sending a signal");
-            }
-            tracing::info!("Stop signal received, proof data handler server is shutting down");
-        })
-        .await
-        .context("Proof data handler server failed")?;
-    tracing::info!("Proof data handler server shut down");
-    Ok(())
+    if config.tee_support {
+        return router.route(
```
popzxc, line 88: Probably a more idiomatic way would be to make `router` mutable and here simply modify it (e.g. `router = router.route(..)`), so that we have a single return point. The benefit here would be that this approach would allow the conditional composition of several modules out of box.

core/node/proof_data_handler/src/tee_request_processor.rs
```diff
@@ -0,0 +1,134 @@
...
+        &self,
+        request: Json<TeeProofGenerationDataRequest>,
+    ) -> Result<Json<TeeProofGenerationDataResponse>, RequestProcessorError> {
+        tracing::info!("Received request for proof generation data: {:?}", request);
+
+        let mut connection = self
+            .pool
+            .connection()
+            .await
+            .map_err(|_| RequestProcessorError::Sqlx(SqlxError::PoolClosed))?;
```
popzxc, line 49: Will leave it as a separate comment to not mix new and old discussions. I think the right approach here would be to use `DalError` instead of `SqlxError` in `RequestProcessorError` (I believe that `RequestProcessorError` simply predates `DalError` which is why it wasn't integrated). Then you'll be able to use `inner` inside of `into_response` for the purpose of pattern matching.

core/lib/prover_interface/src/outputs.rs
```diff
@@ -5,15 +5,21 @@ use serde::{Deserialize, Serialize};
...
-/// to the L1 contract.
+/// A "final" proof that can be sent to the L1 contract.
 #[derive(Clone, Serialize, Deserialize)]
 pub struct L1BatchProofForL1 {
     pub aggregation_result_coords: [[u8; 32]; 4],
     pub scheduler_proof: FinalProof,
     pub protocol_version: ProtocolSemanticVersion,
 }
 
+#[derive(Clone, PartialEq, Serialize, Deserialize)]
```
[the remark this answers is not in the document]
pbeza: [Added](https://github.com/matter-labs/zksync-era/pull/1993/commits/2aa36eff03d21f4766d1d4e0388e99fb859ebc0d) for this particular `struct` only. ✅

core/node/proof_data_handler/Cargo.toml
pbeza: No, I missed it. [Fixed](https://github.com/matter-labs/zksync-era/pull/1993/commits/0114a3cdf5a36650d4cc26cbf59e66c5b3e5014b). ✅

core/node/proof_data_handler/Cargo.toml
pbeza: Missed it. [Fixed](https://github.com/matter-labs/zksync-era/pull/1993/commits/0114a3cdf5a36650d4cc26cbf59e66c5b3e5014b). ✅

core/node/proof_data_handler/src/lib.rs
pbeza: Nope. [Fixed](https://github.com/matter-labs/zksync-era/pull/1993/commits/c4e4074a7135e3ff391ef0a315b939d079a13a0b). ✅

core/node/proof_data_handler/src/lib.rs
pbeza: [Fixed](https://github.com/matter-labs/zksync-era/pull/1993/commits/c4e4074a7135e3ff391ef0a315b939d079a13a0b). ✅

core/node/proof_data_handler/src/tee_request_processor.rs
pbeza: @RomanBrodetski I'm not sure if I understand, can you pls elaborate on that? I tried to grasp [the reasoning](https://matter-labs-workspace.slack.com/archives/C05ANUCGCKV/p1694763650123479) behind using `SkippedProofGeneration` for rate limiting and thought we could apply the same logic to TEEs if we want to rate limit for any reason. I'm not sure why you're linking it to the TEE enable/disable toggle switch. 🤔

core/node/proof_data_handler/src/tee_request_processor.rs
pbeza: [Fixed](https://github.com/matter-labs/zksync-era/pull/1993/commits/2b296b3d0c397e2ba72cce49eacb866e917e4d0d). ✅

core/node/metadata_calculator/src/updater.rs
pbeza: [Fixed](https://github.com/matter-labs/zksync-era/pull/1993/commits/79e1d07b6ba6e98a0bb3a9adce9ca38829657e6a). ✅

Fun fact - I used to have the implementation you suggested, but later I arbitrarily switched to a different implementation:
https://github.com/matter-labs/zksync-era/blob/daaf6147ca353753d077b1cdafc41d417c0333e1/core/node/tee_verifier_input_producer/src/lib.rs#L257-L266

pbeza:
@popzxc @RomanBrodetski CI checks are passing, and manual tests have been successfully redone. From my perspective, it's ready to merge. ✅

There is just [one outstanding comment](https://github.com/matter-labs/zksync-era/pull/1993#discussion_r1632764685) that I need to clarify if you think it's critical to resolve before merging.

core/node/proof_data_handler/src/tee_request_processor.rs
popzxc: We can skip proof generation on testnet/local envs. Locally it makes sense as we don't run a prover. On testnet we do run a prover, but proving is expensive, so we apply sampling there -- we only generate _some_ proves, not for every batch.
With TEE, ability to skip proofs doesn't seem to make sense -- since we may simply work without it. If we don't need logic for skipping jobs, maybe it makes sense to remove this functionality completely?

haraldh (approved):

core/lib/dal/src/tee_proof_generation_dal.rs
```diff
@@ -0,0 +1,211 @@
...
+            .with_arg("pubkey", &pubkey)
+            .with_arg("attestation", &attestation);
+        let result = instrumentation
+            .clone()
+            .with(query)
+            .execute(self.storage)
+            .await?;
+        if result.rows_affected() == 0 {
+            let err = instrumentation.constraint_error(anyhow::anyhow!(
+                "Unable to insert TEE attestation for a non-existent batch number"
```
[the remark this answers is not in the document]
pbeza: [Fixed](https://github.com/matter-labs/zksync-era/pull/1993/files#diff-46d00b49fb68d6da13c9ecdc5ab3dab6adf502ec216349721586e96553652444R204) ✅

core/node/proof_data_handler/src/tee_request_processor.rs
pbeza: [Fixed](https://github.com/matter-labs/zksync-era/pull/1993/commits/eb145f4ec07c9404d97321d4e304e3da45e5d886). ✅

pbeza:
@popzxc @RomanBrodetski [Again](https://github.com/matter-labs/zksync-era/pull/1993#issuecomment-2160596423): CI checks are passing, and manual tests have been successfully redone. From my perspective, it's ready to merge. ✅

Sorry for the mess with rebasing and force pushing. I didn't expect it to make incremental code review harder. I will merge instead of rebase next time.

popzxc (approved):
👍

pbeza:
@popzxc @RomanBrodetski are we good to merge? This is my first PR to be merged, so I'm not sure about the process. How many reviewers need to approve it before we can merge?

haraldh (approved):

popzxc:
@pbeza generally up to the reviewee. If your PR is approved, you are free to merge, but it's not uncommon
- either for the reviewer to suggest waiting for a 2nd review (e.g. if reviewer doesn't feel confident about some part of the diff), or
- for the reviewee to ask for more reviews (e.g. if the change is complex and they want more eyes to at least skim through it).

So overall, we trust people to decide what would be the safest and most productive approach.

pbeza:
Got it, I didn't have permission to merge, which is why I asked. Thanks for the clarification.

PR #1994, opened 2024-05-20
fix(API): polish web3 api  block-related types

AnastasiiaVashchuk:
Follow-up PR for https://github.com/matter-labs/zksync-era/pull/1946/files
context: https://github.com/matter-labs/zksync-era/pull/1946/files#r1602029812

## What ❔
+ return correct `gas_limit` and `gas_used`
+ Improve the `ws` unit test to check the full response 

## Checklist


core/lib/basic_types/src/web3/mod.rs
```diff
@@ -354,8 +354,8 @@ pub struct BlockHeader {
     #[serde(rename = "gasLimit", default)]
     pub gas_limit: U256,
     /// Base fee per unit of gas (if past London)
-    #[serde(rename = "baseFeePerGas", skip_serializing_if = "Option::is_none")]
-    pub base_fee_per_gas: Option<U256>,
```
AnastasiiaVashchuk, line 358: Era started to worked *after* London fork so `base_fee_per_gas` should be present.

perekopskiy: Types in this file are used by L1 eth client. As you've stated all era networks (testnet, mainnet) started to work after London hard fork, but I would still leave `Option` to make it possible to query pre-London blocks -- can be required for some cases, I don't think we should limit eth client functionality.

core/lib/dal/src/blocks_web3_dal.rs
```diff
@@ -200,8 +202,11 @@ impl BlocksWeb3Dal<'_, '_> {
             receipts_root: H256::zero(),
             number: Some(U64::from(row.number)),
             gas_used: U256::zero(),
```
perekopskiy, line 204: Let's return meaningful gas_used as well (see `get_api_block` impl)

perekopskiy (commented):
Please mark PR as fix, as it's a fix and it should be reflected in changelog

core/lib/dal/src/blocks_web3_dal.rs
```diff
@@ -199,9 +203,20 @@ impl BlocksWeb3Dal<'_, '_> {
...
-            base_fee_per_gas: None,
+            gas_used: {
+                if let (Some(gas_limit), Some(refunded_gas)) =
+                    (row.block_gas_limit, row.refunded_gas)
+                {
+                    bigdecimal_to_u256(gas_limit.into()) - U256::from(refunded_gas as u64)
+                } else {
+                    U256::zero()
+                }
+            },
```
perekopskiy, line 214: This is incorrect, you should do `transaction_gas_limit - refunded_gas`.

Also, you should take into account that you now join `transactions` table, number of rows returned will be equal to number of transactions, not number of blocks, so doing `let blocks = rows.into_iter().map(..` is incorrect, check how it's done in `get_api_block` with `fold` and reuse. Alternatively, you can avoid joining and select aggregated `transaction_gas_limit - refunded_gas` in SQL with subquery

perekopskiy (approved):

PR #1997, opened 2024-05-21
feat(node_framework): Migrate main node to the framework

popzxc:
## What ❔

Makes it possible to launch main node using node framework.
Also does some necessary things that were required for migration (e.g. splits eth sender layer into two, adds a layer for tee component, fixes a few minor bugs, etc).

## Why ❔

Indeed.

## Checklist


core/bin/zksync_server/src/main.rs
```diff
@@ -209,7 +214,30 @@ async fn main() -> anyhow::Result<()> {
         opt.components.0
     };
 
+    // If the node framework is used, run the node.
+    if opt.use_node_framework {
+        // We run the node from a different thread, since the current thread is in tokio context.
```
popzxc, line 219: This is to be fixed once we fully migrate to the framework.

core/node/node_framework/examples/main_node.rs
```diff
@@ -356,47 +354,6 @@ impl MainNodeBuilder {
         Ok(self)
     }
 
-    fn add_consensus_layer(mut self) -> anyhow::Result<Self> {
```
popzxc, line 359: This example is incomplete, and since the configs are loaded differently here, I decided to remove it instead of reworking.
Probably once we migrate the main node to the framework fully, this example is to be removed anyways.

core/node/node_framework/src/implementations/layers/web3_api/tree_api_client.rs
```diff
@@ -10,6 +10,10 @@ use crate::{
     wiring_layer::{WiringError, WiringLayer},
 };
 
+/// Layer that inserts the `TreeApiHttpClient` into the `ServiceContext` resources, if there is no
```
popzxc, line 13: AFAIU the HTTP client is currently used as a "fallback" (sort of), so I decided to embed this logic into layer directly, rather than dealing with `resource already exists` outside.

.github/workflows/ci-core-reusable.yml
```diff
@@ -202,7 +202,7 @@ jobs:
       # `sleep 5` because we need to wait until server started properly
       - name: Run server
         run: |
-          ci_run zk server --components=$SERVER_COMPONENTS &>server.log &
+          ci_run zk server --use-node-framework --components=$SERVER_COMPONENTS &>server.log &
```
popzxc, line 205: This means that for each workflow in the matrix,
1) we will run integration tests with the server initialized via framework, but
2) when we run EN integration tests, the server will be initialized the old way.

So, we will cover all the feature sets without having any more workflows.

perekopskiy (dismissed):
Looks good!

Deniallugo (dismissed):

slowli (approved):

artmakh (approved):

artmakh (approved):

PR #2000, opened 2024-05-21
feat(en): Fetch old L1 batch hashes from L1

slowli:
## What ❔

Fetches commitment hashes for old L1 batches from L1 instead of using a trusted L2 node.

## Why ❔

This requires slightly less trust than a trusted L2 node. While the main node can theoretically write arbitrary hash to L1, it cannot equivocate, so a "fake" hash will be discovered soon enough, and it would be impossible to prove.

## Checklist


core/lib/eth_client/src/lib.rs
```diff
@@ -76,14 +79,14 @@ impl Options {
 pub trait EthInterface: Sync + Send {
     /// Fetches the L1 chain ID (in contrast to [`BoundEthInterface::chain_id()`] which returns
     /// the *expected* L1 chain ID).
-    async fn fetch_chain_id(&self) -> Result<L1ChainId, Error>;
+    async fn fetch_chain_id(&self) -> EnrichedClientResult<L1ChainId>;
```
slowli, line 82: If necessary, I can extract these changes to `eth_client` ([refactoring its error types](https://github.com/matter-labs/zksync-era/pull/2000/commits/26241de8eb3cc47fdba2069547b130724e973da6)) in a separate PR. It allowed to simplify error handling for the L1 data fetcher.

core/node/node_sync/src/tree_data_fetcher/provider/mod.rs
```diff
@@ -0,0 +1,321 @@
...
+            let threshold_timestamp = info.l1_commit_block_timestamp + Self::L1_BLOCK_RANGE.as_u64() / 2;
+            if U256::from(l1_batch_seal_timestamp) > threshold_timestamp {
+                tracing::debug!(
+                    number = number.0,
+                    "L1 batch #{number} seal timestamp ({l1_batch_seal_timestamp}) is too far ahead \
+                     of the previous processed L1 batch ({info:?}); not using L1 batch info"
+                );
+                None
+            } else {
+                // This is an exact lower boundary: L1 batches are committed in order
```
slowli, line 208: Just to double-check: Is this always true? Seems so after reading L1 contracts, but I may be missing something.

core/node/node_sync/src/tree_data_fetcher/provider/mod.rs
```diff
@@ -0,0 +1,321 @@
...
+
+/// Data provider combining [`L1DataProvider`] with a fallback provider.
+#[derive(Debug)]
+pub(super) struct CombinedDataProvider {
+    l1: Option<L1DataProvider>,
+    fallback: Box<dyn TreeDataProvider>,
+}
+
+#[async_trait]
+impl TreeDataProvider for CombinedDataProvider {
```
slowli, line 286: A nice property is that any errors emitted by `L1DataProvider` are not treated as fatal; they either lead to switching it off, or to using the L2 provider as a fallback (in case of transient errors). I.e., `L1DataProvider` is only required to not return bogus root hashes, which I *think* is the case.

core/node/node_sync/src/tree_data_fetcher/provider/mod.rs
perekopskiy: Correct

perekopskiy (approved):

popzxc (approved):
LGTM, though probably it makes sense to add metrics (e.g. how often do we have to use fallback provider, how many guesses we do in the binary search, etc).

slowli:
@popzxc I was sure I've added metrics (thought about them as well) 🙃 Let me do that in a follow-up PR.

PR #2001, opened 2024-05-21
feat(toolbox): add zk toolbox

Deniallugo:
## Checklist


zk_toolbox/crates/zk_inception/src/commands/ecosystem/create.rs
```diff
@@ -91,6 +91,7 @@ fn create(args: EcosystemCreateArgs, shell: &Shell) -> anyhow::Result<()> {
 
     if args.start_containers {
         let spinner = Spinner::new("Starting containers...");
+        initialize_docker(shell, &ecosystem_config)?;
```
Deniallugo, line 94: would it be better to use run from containers?

aon: Tried this approach, but logger breaks (runs `logger::outro` when run hasn't finished yet).

PR #2006, opened 2024-05-21
refactor(config): Remove `proof_loading_mode`  variable

AnastasiiaVashchuk:
## What ❔
Clear code base from `proof_loading_mode` variable

## Why ❔

We do not use `OldProofFromDb ` option anymore. It is expected to always be `FriProofFromGcs `

## Checklist

### DIFF
--- a/core/lib/protobuf_config/src/proto/config/eth_sender.proto
+++ b/core/lib/protobuf_config/src/proto/config/eth_sender.proto
@@ -43,7 +43,6 @@ message Sender {
   optional uint64 l1_batch_min_age_before_execute_seconds = 15; // optional; s
   optional uint64 max_acceptable_priority_fee_in_gwei = 16; // required; gwei
   optional PubdataSendingMode pubdata_sending_mode = 18; // required
-  optional ProofLoadingMode proof_loading_mode = 19;
 }
 
 message GasAdjuster {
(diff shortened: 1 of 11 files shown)


RomanBrodetski:
@AnastasiiaVashchuk please also don't forget to remove corresponding entries from gitops. Please evaluate if it's possible to do it now or we need to wait for deploy on every env.

AnastasiiaVashchuk:
> @AnastasiiaVashchuk please also don't forget to remove corresponding entries from gitops. Please evaluate if it's possible to do it now or we need to wait for deploy on every env.

imho it make sense after we will deploy this change on all envs

core/lib/protobuf_config/src/proto/config/eth_sender.proto
```diff
@@ -43,7 +43,6 @@ message Sender {
   optional uint64 l1_batch_min_age_before_execute_seconds = 15; // optional; s
   optional uint64 max_acceptable_priority_fee_in_gwei = 16; // required; gwei
   optional PubdataSendingMode pubdata_sending_mode = 18; // required
-  optional ProofLoadingMode proof_loading_mode = 19;
```
Deniallugo, line 46: you have to mark is as reserved

Deniallugo (approved):

PR #2008, opened 2024-05-22
chore: Moved kzg related code to test_harness

mm-zk:
## What ❔

* Moved kzg / blob related code (commitment computation etc) to zkevm_test_harness

## Why ❔

* It depended on internal circuit methods - and this way, the interface between sequencer and VM is smaller and cleaner, with only 3 methods being exposed.


koloz193 (approved):
These are the best types of PRs:
Screenshot 2024-05-22 at 8 25 30 AM

My only comment is do we still need this file or was this done to not leak the dependency on crypto into other crates?

mm-zk:
@koloz193 - we can decide either way. In this PR, I wanted to minimise the impact on zksync-era, so I did the 're-export'.

PR #2016, opened 2024-05-22
feat(zk-toolbox): add balance check

Deniallugo:
## Checklist

### DIFF
--- a/zk_toolbox/crates/common/src/forge.rs
+++ b/zk_toolbox/crates/common/src/forge.rs
@@ -1,12 +1,17 @@
 use std::path::{Path, PathBuf};
+use std::str::FromStr;
 
 use clap::Parser;
+use ethers::abi::Address;
+use ethers::middleware::Middleware;
+use ethers::prelude::U256;
 use ethers::{abi::AbiEncode, types::H256};
 use serde::{Deserialize, Serialize};
 use strum_macros::Display;
 use xshell::{cmd, Shell};
 
 use crate::cmd::Cmd;
+use crate::ethereum::create_ethers_client;
 
 /// Forge is a wrapper around the forge binary.
 pub struct Forge {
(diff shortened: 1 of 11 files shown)


zk_toolbox/crates/zk_inception/src/forge_utils.rs
```diff
@@ -12,3 +13,21 @@ pub fn fill_forge_private_key(
     }
     Ok(forge)
 }
+
+pub async fn check_the_balance(forge: &ForgeScript) -> anyhow::Result<()> {
+    let mut not_enough_money = true;
+    while not_enough_money {
```
aon, line 19: It seems we don't really need a while loop here.

zk_toolbox/crates/common/src/forge.rs
```diff
@@ -93,6 +98,39 @@ impl ForgeScript {
...
+        self.args.args.iter().find_map(|a| {
+            if let ForgeScriptArg::RpcUrl { url } = a {
+                Some(url.clone())
+            } else {
+                None
+            }
+        })
+    }
+
+    pub async fn check_the_balance(&self, minimum_value: U256) -> anyhow::Result<(bool, Address)> {
```
aon, line 122:
```suggestion
    pub async fn check_deployer_balance(&self, minimum_value: U256) -> anyhow::Result<(bool, Address)> {
```

zk_toolbox/crates/common/src/forge.rs
```diff
@@ -93,6 +98,39 @@ impl ForgeScript {
...
+        let Some(rpc_url) = self.rpc_url() else {
+            return Ok((true, Address::zero()));
+        };
+        let Some(private_key) = self.private_key() else {
+            return Ok((true, Address::zero()));
+        };
```
aon, line 128: In this case, shouldn't we be returning `false` if parameters are not correctly set?

zk_toolbox/crates/zk_inception/src/forge_utils.rs
```diff
@@ -12,3 +13,21 @@ pub fn fill_forge_private_key(
...
+            .check_the_balance(MINIMUM_BALANCE_FOR_WALLET.into())
+            .await?;
+
+        not_enough_money = !success;
+
+        if not_enough_money {
+            if common::PromptConfirm::new(format!("Address {address:?} doesn't have enough money to deploy contracts do you want to continue?")).ask() {
+                break;
+            }
+        }
```
aon, line 30: I think if `not_enough_money == true` and the user prompts to quit, then we should raise an error from there so that program execution is stopped.

zk_toolbox/crates/zk_inception/src/forge_utils.rs
```diff
@@ -12,3 +13,21 @@ pub fn fill_forge_private_key(
     }
     Ok(forge)
 }
+
+pub async fn check_the_balance(forge: &ForgeScript) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
Deniallugo: it's not only deployer

zk_toolbox/crates/zk_inception/src/forge_utils.rs
Deniallugo: is there any other way, how to check it every time? 
we have two ways to close the loop, have enough money of forcefully continue

zk_toolbox/crates/zk_inception/src/forge_utils.rs
Deniallugo: but it's not quit.. 
It's yes or no, if no it will check again , if yes it will proceed. 

I think it's common knowledge that you can cancel everything using Ctrl + c

zk_toolbox/crates/common/src/forge.rs
aon: Could this ever happen that there's no rpc_url or private_key set?

Deniallugo: no idea...probably if we want to have it without broadcasting

aon (approved):

PR #2017, opened 2024-05-22
feat(vm-runner): shadow protective reads using VM runner

itegulov:
## What ❔

Adds a new component `vm_runner_protective_reads` that computes protective reads independently and asynchronously from state keeper.

For now, the component does not actually save anything; instead, it computes protective reads and compares them against what state keeper has already written to the DB. So, in short, this is just the first stepping stone aka a sanity check that the VM runner mechanism works as intended.

## Why ❔

In the future, we want to be able to save protective reads asynchronously thus saving time on L1 batch sealing.

## Checklist


.github/workflows/ci-core-reusable.yml
```diff
@@ -134,7 +134,7 @@ jobs:
         base_token: ["Eth", "Custom"]
         deployment_mode: ["Rollup", "Validium"]
     env:
-      SERVER_COMPONENTS: "api,tree,eth,state_keeper,housekeeper,commitment_generator${{ matrix.consensus && ',consensus' || '' }}"
+      SERVER_COMPONENTS: "api,tree,eth,state_keeper,housekeeper,commitment_generator,vm_runner_protective_reads${{ matrix.consensus && ',consensus' || '' }}"
```
perekopskiy, line 137: Let's add it to EN int tests and loadtest

core/node/node_framework/src/implementations/layers/vm_runner/protective_reads.rs
```diff
@@ -0,0 +1,71 @@
...
+impl WiringLayer for ProtectiveReadsWriterLayer {
+    fn layer_name(&self) -> &'static str {
+        "vm_runner_protective_reads"
+    }
+
+    async fn wire(self: Box<Self>, mut context: ServiceContext<'_>) -> Result<(), WiringError> {
+        let master_pool = context.get_resource::<PoolResource<MasterPool>>().await?;
+
+        let (protective_reads_writer, tasks) = ProtectiveReadsWriter::new(
+            master_pool.get_custom(3).await?,
```
perekopskiy, line 40: please add comment on why 3 connections

etc/env/base/vm_runner.toml
```diff
@@ -0,0 +1,9 @@
+# Configuration for the VM runner crate
```
perekopskiy, line 1: please update yaml config as well (etc/env/file_based)

perekopskiy (commented):
Nice!

core/node/node_framework/src/implementations/layers/vm_runner/protective_reads.rs
itegulov: Actually, thinking about it one more time, it seems like we need `window_size + 2` size pool here. Please verify my logic I left in the comment

etc/env/base/vm_runner.toml
itegulov: I think I have added it but I am struggling to find a way to test it locally.

Also, I tried to figure out the scheme in which way the naming hierarchy is structured (i.e. if `protective_reads_writer.protective_reads_db_path` is the correct path). My current guess is that it's based on the field names in `GeneralConfig` but I haven't found any relevant logic in the code.

perekopskiy (approved):

etc/env/base/vm_runner.toml
itegulov: Nevermind, talked to @Deniallugo and apparently it's just based off of protobuf

amakhortov (approved):

artmakh (approved):

PR #2018, opened 2024-05-22
refactor: Add tree writes persistence

perekopskiy:
## What ❔

Refactors state keeper, so tree writes are saved in a separate output handler.

## Why ❔

Code quality: split `seal_l1_batch` method which is already huge.

## Checklist

### DIFF
--- a/core/node/metadata_calculator/src/helpers.rs
+++ b/core/node/metadata_calculator/src/helpers.rs
@@ -514,6 +514,7 @@ impl L1BatchWithLogs {
         storage: &mut Connection<'_, Core>,
         l1_batch_number: L1BatchNumber,
         mode: MerkleTreeMode,
+        wait_for_tree_writes: bool,
     ) -> anyhow::Result<Option<Self>> {
         tracing::debug!(
             "Loading storage logs data for L1 batch #{l1_batch_number} for {mode:?} tree"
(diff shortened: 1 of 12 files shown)


core/node/metadata_calculator/src/helpers.rs
```diff
@@ -552,10 +553,34 @@ impl L1BatchWithLogs {
         };
 
         let load_tree_writes_latency = METRICS.start_load_stage(LoadChangesStage::LoadTreeWrites);
-        let tree_writes = storage
-            .blocks_dal()
-            .get_tree_writes(l1_batch_number)
-            .await?;
+
+        let tree_writes = if wait_for_tree_writes {
+            // If `wait_for_tree_writes` flag is set then we expect that waiting for tree writes
```
slowli, line 558: Bikeshedding: This looks awkward; I'd structure it as a timeout with a polling loop inside:

```rust
let tree_writes = tokio::time::timeout(_, async {
    loop {
        if let Some(tree_writes) = storage
            .blocks_dal()
            .get_tree_writes(l1_batch_number)
            .await?
        {
            break anyhow::Ok(tree_writes); 
        }
        tokio::time::sleep(_).await;
    }
})
.await
.ok()
.transpose()?;
```

... and maybe extract it to a method.

core/node/metadata_calculator/src/helpers.rs
```diff
@@ -552,10 +553,34 @@ impl L1BatchWithLogs {
         };
 
         let load_tree_writes_latency = METRICS.start_load_stage(LoadChangesStage::LoadTreeWrites);
-        let tree_writes = storage
-            .blocks_dal()
-            .get_tree_writes(l1_batch_number)
-            .await?;
+
+        let tree_writes = if wait_for_tree_writes {
```
slowli, line 557: This flag certainly doesn't feel like a right approach. Intuitively, I'd try logic like this:

- Try getting tree writes immediately. If they are there, use them.
- Check whether tree writes are present for the previous batch (a potential alternative: look at last N batches, N > 1 but still small so that we don't need to do anything resembling a full scan of `l1_batches`). If they are there, wait for writes of the current batch; if not, use a fallback.

So it's similar to what you've implemented *in practice*, but IMO significantly more tractable because of locality; we're not given a parameter from outside whether to wait for tree writes, but decide on our own given reasoning that can ostensibly hold regardless of how tree writes generation is handled.

perekopskiy:
I'm merging it for now, @slowli please review it as a part of main PR

PR #2019, opened 2024-05-22
feat: verification of L1Batch witness (BFT-471)

pompon0:
Verification of an L1 batch, based on StoredBatchInfo. We extract the hash of the last block of the transaction via merkle path from the root_state (for batch n and n-1) and we recompute the rolling block hash based on transactions in the payload. The verification is not perfect as there are still some fields in payload for which we don't have a commitment - we should address that later.

Verification will be used for syncing L1Batches over p2p network (not implemented yet).

Also removed serde serialization from TransactionRequest which is unused - we don't need to maintain encoding compatibility for it.


core/node/consensus/src/batch.rs
```diff
@@ -0,0 +1,266 @@
...
+                .batch_number
+                .checked_add(1)
+                .context("batch_number overflow")?
+                == u64::from(comm.number.0)
+        );
+        anyhow::ensure!(self.this_batch.info.batch_number == u64::from(comm.number.0));
+        for (i, b) in self.blocks.iter().enumerate() {
+            anyhow::ensure!(b.l1_batch_number == comm.number);
+            anyhow::ensure!(b.protocol_version == self.this_batch.protocol_version);
+            anyhow::ensure!(b.last_in_batch == (i + 1 == self.blocks.len()));
```
StanislavBreadless, line 246: do we also want to add `anyhow::ensure` for the intermediate results of hashes, etc? It does not give any additional security, but allows to fail fast

core/node/consensus/src/tests.rs
```diff
@@ -515,3 +517,45 @@ async fn test_centralized_fetcher(from_snapshot: bool) {
     .await
     .unwrap();
 }
+
+/// Tests that generated L1 batch witnesses can be verified successfully.
+/// TODO: add tests for verification failures.
```
StanislavBreadless, line 522: will it be resolved within this PR?

core/node/consensus/src/batch.rs
```diff
@@ -0,0 +1,266 @@
...
+    block::{unpack_block_info, L2BlockHasher},
+    AccountTreeId, L1BatchNumber, L2BlockNumber, ProtocolVersionId, StorageKey, Transaction, H256,
+    U256,
+};
+
+use crate::ConnectionPool;
+
+/// Commitment to the last block of a batch.
+pub(crate) struct LastBlockCommit {
+    /// Hash of the `StoredBatchInfo`.
```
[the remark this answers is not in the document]
pompon0: This is what is stored on L1, so hash of the StoredBatchInfo, which contains a root_hash of the L2 state, which contains state of the SystemContext contract, which contains the L2 block hash.

core/node/consensus/src/tests.rs
pompon0: no, but will be resolved before we start using this logic. I don't have time right now to work on it.

core/node/consensus/src/batch.rs
pompon0: sg

core/node/consensus/src/batch.rs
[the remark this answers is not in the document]
StanislavBreadless: I would probably add a comment about it, from the outside it looks a bit confusing that the commitment for an L2 block is the batch hash

core/lib/dal/src/sync_dal.rs
```diff
@@ -53,35 +53,44 @@ impl SyncDal<'_, '_> {
             FROM
                 miniblocks
             WHERE
-                miniblocks.number = $1
+                miniblocks.number BETWEEN $1 AND $2
             "#,
-            i64::from(block_number.0)
+            i64::from(numbers.start.0),
+            i64::from(numbers.end.0 - 1),
```
popzxc, line 59: I'm not an expert in SQL, but AFAIU `BETWEEN` operator is not ordered (e.g. arguments can come in either order), and so it looks like this query will work incorrectly if called with an empty range. Maybe bail if `numbers.is_empty()`?

core/lib/dal/src/transactions_web3_dal.rs
```diff
@@ -392,18 +398,46 @@ impl TransactionsWeb3Dal<'_, '_> {
...
             WHERE
-                miniblocks.number = $1
+                miniblocks.number BETWEEN $1 AND $2
             ORDER BY
+                miniblock_number,
                 index_in_block
             "#,
-            i64::from(l2_block.0)
+            i64::from(blocks.start.0),
+            i64::from(blocks.end.0 - 1),
```
popzxc, line 407: Same

core/node/consensus/src/batch.rs
```diff
@@ -0,0 +1,266 @@
...
+                .context("L2BlockNumber overflow")?,
+        );
+
+        // Verify merkle paths.
+        self.current_l2_block_info
+            .verify(Self::current_l2_block_info_key(), self.info.batch_hash)?;
+        self.tx_rolling_hash
+            .verify(Self::tx_rolling_hash_key(), self.info.batch_hash)?;
+        self.l2_block_hash_entry
+            .verify(Self::l2_block_hash_entry_key(prev), self.info.batch_hash)?;
```
popzxc, line 168: Nit: Maybe add `context` explaining which proof verification failed exactly? Would probably speed up debugging if ever occurs in a real env.

core/node/consensus/Cargo.toml
```diff
@@ -21,20 +21,27 @@ zksync_consensus_bft.workspace = true
...
 async-trait.workspace = true
 secrecy.workspace = true
+tempfile.workspace = true
 tracing.workspace = true
 
 [dev-dependencies]
 zksync_node_genesis.workspace = true
 zksync_node_test_utils.workspace = true
 zksync_node_api_server.workspace = true
+zksync_test_account.workspace = true
```
popzxc, line 44: Same

core/node/consensus/src/batch.rs
```diff
@@ -0,0 +1,266 @@
...
+/// verified against `L1BatchCommit1`.
+pub struct L1BatchWithWitness {
+    pub(crate) blocks: Vec<Payload>,
+    pub(crate) this_batch: LastBlockWitness,
+    pub(crate) prev_batch: LastBlockWitness,
+}
+
+impl LastBlockWitness {
+    /// Address of the SystemContext contract.
+    fn addr() -> AccountTreeId {
```
popzxc, line 52: Bikeshedding: maybe `system_context_addr`? I find `LastBlockWitness::addr()` a bit confusing.

popzxc (commented):
Generally looks good, a few minor things.

core/lib/types/src/l2/mod.rs
```diff
@@ -206,10 +206,19 @@ impl L2Tx {
...
             paymaster_params,
         );
-
-        let data = res.get_signed_bytes(chain_id);
-        res.set_signature(PackedEthSignature::sign_raw(private_key, &data).context("sign_raw")?);
-        Ok(res)
+        // We do a whole dance to reconstruct missing data: RLP encoding, hash and signature.
+        let mut req: TransactionRequest = tx.into();
+        req.chain_id = Some(chain_id.as_u64());
+        let data = req
```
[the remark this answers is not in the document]
pompon0: not exactly, because we need to call `req.get_signed_bytes(&sig)` afterward anyway

core/lib/dal/src/sync_dal.rs
pompon0: thx, good catch

core/node/consensus/src/batch.rs
pompon0: done

core/node/consensus/src/batch.rs
pompon0: done

popzxc (approved):
LGTM, but would probably still suggest to wait for an approval from Stas

StanislavBreadless (approved):

PR #2020, opened 2024-05-22
chore: removed consensus column from the miniblocks table

pompon0:
unused for ~4 months
removal requested by @RomanBrodetski


RomanBrodetski:
@Deniallugo are you sure we should block this? The column is not used for a while, and the new version is on every env apart from mainnet. IMO let's merge

RomanBrodetski (approved):

Deniallugo:
I just don't feel super confident on removing columns before the upgrade

PR #2021, opened 2024-05-22
feat: Add metrics for transaction execution result in state keeper

AnastasiiaVashchuk:
## Checklist


core/node/state_keeper/src/batch_executor/main_executor.rs
```diff
@@ -207,6 +210,9 @@ impl CommandReceiver {
         // There is some post-processing work that the VM needs to do before the block is fully processed.
         let result = vm.finish_batch();
         if result.block_tip_execution_result.result.is_failed() {
+            KEEPER_METRICS.tx_execution_result
+                [&metrics::TxExecutionResult::from(&result.block_tip_execution_result.result)]
+                .inc();
```
perekopskiy, line 215: I think we shouldn't increment metric here as batch tip is not a transaction and it's "successful" in 100% of cases (if it's not then server panics in the next line)

core/node/state_keeper/src/metrics.rs
```diff
@@ -30,6 +30,40 @@ pub enum TxExecutionType {
     L2,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, EncodeLabelValue)]
+#[metrics(rename_all = "snake_case")]
+pub enum TxExecutionStatus {
+    Success,
+    Rejected,
+    Reverted,
+    Halted,
```
perekopskiy, line 39: If you apply my suggestion from the previous comment, then `Halted` variant won't be needed:
- all txs that result into `Halted` are either rejected or rolled back
- all txs that are added to `UpdatesManager` must not have `Halted` status, I suggest asserting it

core/node/state_keeper/src/updates/l2_block_updates.rs
```diff
@@ -109,6 +111,9 @@ impl L2BlockUpdates {
             ExecutionResult::Halt { reason } => Some(reason.to_string()),
         };
 
+        KEEPER_METRICS.tx_execution_result[&TxExecutionResult::from(&tx_execution_result.result)]
```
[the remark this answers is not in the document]
AnastasiiaVashchuk: Good idea!
thanks!

core/node/state_keeper/src/metrics.rs
```diff
@@ -30,6 +30,40 @@ pub enum TxExecutionType {
...
+impl From<&ExecutionResult> for TxExecutionResult {
+    fn from(result: &ExecutionResult) -> Self {
+        match result {
+            ExecutionResult::Success { output: _ } => TxExecutionResult {
+                status: TxExecutionStatus::Success,
+                reason: None,
+            },
+            ExecutionResult::Revert { output } => TxExecutionResult {
+                status: TxExecutionStatus::Reverted,
+                reason: Some(output.to_user_friendly_string()),
```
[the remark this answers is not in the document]
AnastasiiaVashchuk: make sense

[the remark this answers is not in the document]
AnastasiiaVashchuk: i have checked `enum Halt` and `enum VmRevertReason`. IMHO it makes sense to just record only  outer enum option for non-plain options, if you understand what I mean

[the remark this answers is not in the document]
slowli: I don't know which level of detail we need; I've (implicitly) asked you the same thing 🙃 My guess would be that recording variant discriminants for `Halt` and `VmRevertReason` would indeed be sufficient, but please clarify this with someone more knowledgeable.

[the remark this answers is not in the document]
perekopskiy: `VmRevertReason` has high cardinality, also I think recording outer enum is useless for it. For `Reject` (`Halt` should be removed as per my suggestion above) I think it makes sense to record hardcoded messages that come from seal criterions and `Halt` variants

AnastasiiaVashchuk:
@slowli  @perekopskiy, I listened to your comments and thought it would be better & sufficient to track `TxExecutionResult`. PTAL

core/node/state_keeper/src/io/mod.rs
```diff
@@ -136,7 +136,7 @@ pub trait StateKeeperIO: 'static + Send + Sync + fmt::Debug + IoSealCriteria {
     /// Marks the transaction as "not executed", so it can be retrieved from the IO again.
     async fn rollback(&mut self, tx: Transaction) -> anyhow::Result<()>;
     /// Marks the transaction as "rejected", e.g. one that is not correct and can't be executed.
-    async fn reject(&mut self, tx: &Transaction, error: &str) -> anyhow::Result<()>;
+    async fn reject(&mut self, tx: &Transaction, error: UnexecutableReason) -> anyhow::Result<()>;
```
slowli, line 139: Bikeshedding: I'd rename the argument to `reason: UnexecutableReason`, so that its name is more expected.

core/node/state_keeper/src/keeper.rs
```diff
@@ -22,6 +22,7 @@ use super::{
     updates::UpdatesManager,
     utils::gas_count_from_writes,
 };
+use crate::seal_criteria::UnexecutableReason;
```
slowli, line 25: Nit: You can add this to the `seal_criteria` imports above 🙃

core/lib/multivm/src/interface/types/errors/halt.rs
```diff
@@ -44,6 +44,34 @@ pub enum Halt {
     FailedToPublishCompressedBytecodes,
 }
 
+impl Halt {
+    pub fn to_metrics_friendly_string(&self) -> String {
```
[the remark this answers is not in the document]
AnastasiiaVashchuk: did not know about 'api-guidelines'
thanks!

core/node/state_keeper/src/testonly/test_batch_executor.rs
```diff
@@ -756,15 +756,15 @@ impl StateKeeperIO for TestIO {
...
     }
 
-    async fn reject(&mut self, tx: &Transaction, error: &str) -> anyhow::Result<()> {
+    async fn reject(&mut self, tx: &Transaction, error: UnexecutableReason) -> anyhow::Result<()> {
         let action = self.pop_next_item("reject");
         let ScenarioItem::Reject(_, expected_tx, expected_err) = action else {
             panic!("Unexpected action: {:?}", action);
         };
         assert_eq!(tx, &expected_tx, "Incorrect transaction has been rejected");
         if let Some(expected_err) = expected_err {
```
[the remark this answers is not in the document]
AnastasiiaVashchuk: why this will be better?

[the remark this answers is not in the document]
slowli: Should reiterate again that this makes most sense if `UnexecutableReason` has well-defined granular variants. In this case, comparing `UnexecutableReason`s is better because it's more strongly typed, meaning:

- More difficult to screw up (you can't accidentally mistype an expected error message; it cannot get unnoticeably outdated etc.)
- Easier to read the intention behind the test case
- Easier to debug in case of an error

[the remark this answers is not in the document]
AnastasiiaVashchuk: understood and accepted
thanks

core/node/state_keeper/src/io/mempool.rs
```diff
@@ -265,7 +268,11 @@ impl StateKeeperIO for MempoolIO {
         Ok(())
     }
 
-    async fn reject(&mut self, rejected: &Transaction, error: &str) -> anyhow::Result<()> {
+    async fn reject(
+        &mut self,
+        rejected: &Transaction,
+        error: UnexecutableReason,
```
slowli, line 274: Nit: Renaming the arg in the trait definition sort of implies that it should be renamed in implementations as well 🙃

core/node/state_keeper/src/seal_criteria/mod.rs
```diff
@@ -33,6 +33,114 @@ use super::{
...
+        ErrorMessage::NotEnoughGasProvided => "NotEnoughGasProvided",
+    }
+}
+/// Represents the reason variants for why a transaction was considered unexecutable.
+///
+/// This enum can capture either a hardcoded textual message or a more structured
+/// reason encapsulated in a `Halt` type.
+#[derive(Debug, Clone, PartialEq)]
+pub enum UnexecutableReason {
+    Text(ErrorMessage),
```
slowli, line 110: TBH, I thought that variants you've defined in `ErrorMessage` would be included directly into `UnexecutableReason`:

```rust
pub enum UnexecutableReason {
    Halt(Halt),
    TxEncodingSize,
    LargeEncodingSize,
    PubdataLimit,
    ProofWillFail,
    TooMuchGas,
    OutOfGasForBatchTip,
    BootloaderOutOfGas,
    NotEnoughGasProvided,
}
```

Would it not work like that? This looks simpler; you don't require an additional enum and some of conversions.

core/node/state_keeper/src/testonly/test_batch_executor.rs
```diff
@@ -756,20 +756,19 @@ impl StateKeeperIO for TestIO {
...
-                expected_err,
-                error
-            );
-        }
+        assert!(
+            reason == expected_err,
+            "Transaction was rejected with an unexpected error. Expected part was {}, but the actual error was {}",
+            expected_err,
+            reason
+        );
```
slowli, line 770: Nit: If you use `assert_eq!` macro, you wouldn't need to add the second sentence to the message, since it would show non-equal arguments by default as a part of the panic message.

core/node/state_keeper/src/updates/l2_block_updates.rs
```diff
@@ -104,9 +106,17 @@ impl L2BlockUpdates {
...
         let revert_reason = match &tx_execution_result.result {
-            ExecutionResult::Success { .. } => None,
-            ExecutionResult::Revert { output } => Some(output.to_string()),
-            ExecutionResult::Halt { reason } => Some(reason.to_string()),
+            ExecutionResult::Success { .. } => {
+                KEEPER_METRICS.inc_succeeded_txs();
+                None
+            }
+            ExecutionResult::Revert { output } => {
+                KEEPER_METRICS.inc_reverted_txs(vm_revert_reason_as_metric_label(output));
```
slowli, line 114: Nit: It would help encapsulation to pass `&VmRevertReason` to `inc_reverted_txs()` and convert it to a label there; labels are responsibility of metrics.

core/node/state_keeper/src/metrics.rs
```diff
@@ -77,6 +91,32 @@ pub struct StateKeeperMetrics {
     pub blob_base_fee_too_high: Counter,
 }
 
+impl StateKeeperMetrics {
+    pub fn inc_rejected_txs(&self, reason: &'static str) {
+        self.tx_execution_result[&TxExecutionResult {
+            status: TxExecutionStatus::Rejected,
+            reason: Some(reason),
+        }]
+            .inc();
```
slowli, line 100: Nit: Weird formatting here and below. I'd suggest to extract `TxExecutionResult` to a local var so that this issue doesn't repeat in the future.

core/node/state_keeper/src/keeper.rs
```diff
@@ -670,23 +671,23 @@ impl ZkSyncStateKeeper {
...
+                let criterion = error_message.as_static_str();
+
                 let resolution = if is_first_tx {
-                    SealResolution::Unexecutable(error_message.to_string())
+                    error_message.into()
                 } else {
                     SealResolution::ExcludeAndSeal
                 };
-                AGGREGATION_METRICS.inc(error_message, &resolution);
+                AGGREGATION_METRICS.inc(criterion, &resolution);
```
slowli, line 687: Nit: Here as well, from the encapsulation standpoint it would make sense to refactor `inc()` to accept `&ErrorMessage` (or `&UnexecutableReason`; see the other comment), so that mapping to labels is encapsulated in the `metrics` module.

slowli (commented):
Sorry for comments. The PR looks fine, but I feel it could be made even better.

core/node/state_keeper/src/seal_criteria/mod.rs
AnastasiiaVashchuk: >This looks simpler

I do not feel so. There are two categories of UnexecutableReason: halt and hardcoded string. 
I think it is logical to single them out.
@perekopskiy?

core/node/state_keeper/src/keeper.rs
AnastasiiaVashchuk: @slowli , if I will incapsulate mapping, then I could not get `ErrorMessage` that is needed to create `resolution`, that has to be return.

core/node/state_keeper/src/seal_criteria/mod.rs
perekopskiy: both options seem fine to me

slowli: > There are two categories of UnexecutableReason: halt and hardcoded string

That's retroactive reasoning which I can't agree with. This is how it was implemented previously, yes, but this doesn't mean that this is the best approach.

core/node/state_keeper/src/keeper.rs
slowli: I think I understand what the problem is. `TxAggregationMetrics::inc()` is used in two places with somewhat different semantics for the criterion argument. I think it's definitely inappropriate to use a human-readable label produced by `as_static_str()` here; but even if you make `error_message_as_metric_label` a method in `ErrorMessage`, it would still produce non-backward compatible labels. (Is backward compatibility a concern?) IDK, it may make sense to initialize `criterion` manually together with `error_message` in the `match` above.

perekopskiy (approved):

slowli (approved):
Nice 👍

PR #2050, opened 2024-05-24
perf(en): Parallelize persistence and chunk processing during tree recovery

slowli:
## What ❔

Persists chunks during tree recovery in parallel to processing subsequent chunks.

## Why ❔

- Could speed up tree recovery ~2x on the mainnet (both persistence and processing of a single chunk of 200k entries take ~3s).
- Significantly easier to implement and reason about than alternatives.
- May be used together with alternatives.

## Checklist


slowli:
I've tested the new functionality locally by restoring a testnet node from a snapshot, and it did help with tree recovery time (~50s vs ~65s, i.e. around -25%). The improvement should be more prominent on the mainnet.

perekopskiy (approved):

PR #2052, opened 2024-05-24
feat: added debug_proof to prover_cli

mm-zk:
## What ❔

* Prover_cli has a new command - `debug-proof`, that prints a lot more detailed information about any basic proof file.

## Why ❔

* This will speed up debugging  of the failed proofs.
* Small caveat - the cli has to be compiled with `verbose_circuits` feature - which will include the whole zkevm_test_harness and all the other crypto libraries.

How to use:
* Simply download the .bin file with the proof, and run `./prover_cli  debug-proof file.bin`

Part of EVM-639


ilitteri (dismissed):
LGTM. There's just a missing new line at the end of file in `prover/prover_cli/Cargo.toml`.

prover/Cargo.lock
```diff
@@ -7880,6 +7882,7 @@ dependencies = [
  "lazy_static",
  "rand 0.4.6",
  "rayon",
+ "regex",
  "reqwest",
  "serde",
  "serde_json",
```
RomanBrodetski, line 7888: I understand you need to update dependencies to be able to use this `debug_basic_circuit` feature? Did you just use newest versions from respective branch (that include bug fixes)? If so, we'd need to regenerate VKs etc?

mm-zk: yes I did. And yes, we'll have to regenerate VKs. 

I can put this PR on hold - and we can do the one with new VKs & updated deps first.

mm-zk: we should be able to merge it now.

PR #2058, opened 2024-05-27
fix(pruning): Fix DB pruner responsiveness during shutdown

slowli:
## What ❔

Makes DB pruner more responsive during node shutdown.

## Why ❔

## Checklist

### DIFF
--- a/core/node/db_pruner/src/lib.rs
+++ b/core/node/db_pruner/src/lib.rs
@@ -1,9 +1,8 @@
 //! Postgres pruning component.
 
-use std::{fmt, sync::Arc, time::Duration};
+use std::{sync::Arc, time::Duration};
 
 use anyhow::Context as _;
-use async_trait::async_trait;
 use serde::Serialize;
 use tokio::sync::watch;
 use zksync_dal::{pruning_dal::PruningInfo, Connection, ConnectionPool, Core, CoreDal};
@@ -14,7 +13,7 @@ use self::{
     metrics::{MetricPruneType, METRICS},
     prune_conditions::{
         ConsistencyCheckerProcessedBatch, L1BatchExistsCondition, L1BatchOlderThanPruneCondition,
-        NextL1BatchHasMetadataCondition, NextL1BatchWasExecutedCondition,
+        NextL1BatchHasMetadataCondition, NextL1BatchWasExecutedCondition, PruneCondition,
     },
 };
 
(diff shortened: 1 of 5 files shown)


core/node/db_pruner/src/lib.rs
```diff
@@ -221,10 +229,21 @@ impl DbPruner {
...
+        let mut dal = transaction.pruning_dal();
+        let stats = tokio::select! {
+            result = dal.hard_prune_batches_range(
+                last_soft_pruned_l1_batch,
+                last_soft_pruned_l2_block,
+            ) => result?,
+
+            _ = stop_receiver.changed() => {
+                // `hard_prune_batches_range()` can take a long time. It looks better to roll back it explicitly here if a node is getting shut down
+                // rather than waiting a node to force-exit after a timeout, which would interrupt the DB connection and will lead to an implicit rollback.
```
slowli, line 241: Is this line of reasoning OK, or would it better to implicitly roll back the transaction when a node shuts down once it exceeds the graceful shutdown timeout?

core/node/db_pruner/src/lib.rs
```diff
@@ -290,7 +317,9 @@ impl DbPruner {
                     self.health_updater.update(health);
                     true
                 }
-                Ok(pruning_done) => !pruning_done,
+                Ok(PruningIterationOutcome::Interrupted) => break,
```
tomg10, line 320: wouldn't the pruner have stopped anyway even without this additional logic of PruningIterationOutcome?

slowli: It would, but IMO returning `PruningIterationOutcome` (i.e., a specialized type vs `bool`) makes it cleaner.

tomg10: well, but you didn't just add a type for "pruning_done", you also added third outcome and assert for specific outcome in tests.

core/node/db_pruner/src/lib.rs
tomg10: +1

core/node/db_pruner/src/lib.rs
popzxc: IMHO it's either "be more explicit" or "have more branching". Either option is fine, for me.

core/node/db_pruner/src/tests.rs
```diff
@@ -324,3 +355,59 @@ async fn pruner_is_resistant_to_errors() {
...
+    insert_l2_blocks(&mut conn, 10, 2).await;
+
+    let pruner = DbPruner::with_conditions(
+        DbPrunerConfig {
+            removal_delay: Duration::MAX, // intentionally chosen so that pruning iterations stuck
+            pruned_batch_chunk_size: 3,
+            minimum_l1_batch_age: Duration::ZERO,
+        },
+        pool.clone(),
+        vec![], //No checks, so every batch is prunable
```
popzxc, line 372: Well, now things got serious. In the first comment, you should start with a capital letter, and in the second one you skipped a space between `//` and `N`. ❗ 😠 ❗ 

(jk)

popzxc (dismissed):
Nice.

tomg10 (approved):

popzxc (approved):

PR #2059, opened 2024-05-27
feat: Protocol semantic version

perekopskiy:
## What ❔

Adds support for protocol semantic versions. Major version is always 0. Minor version has the same meaning as the previous concept of protocol version we used. Patch version can be bumped only for vks upgrades.
Adds new DB table `protocol_vk_patches` that keeps track of patch versions.
- eth watcher was updated correspondingly to L1 contracts changes, now it also saves data to `protocol_vk_patches`
- proof data handler passes and accepts protocol semantic version for each batch/final proof
- eth sender determines patch version for batch proofs from L1 and waits for the proof generated for the patch to be present in GCS

## Why ❔

It makes it possible to upgrade vks without bumping minor part of protocol version.

## Checklist

Part of EVM-648


core/bin/genesis_generator/src/main.rs
```diff
@@ -80,7 +82,10 @@ async fn generate_new_config(
 
     let base_system_contracts = BaseSystemContracts::load_from_disk().hashes();
     let mut updated_genesis = GenesisConfig {
-        protocol_version: Some(ProtocolVersionId::latest() as u16),
+        protocol_version: Some(ProtocolSemanticVersion {
+            minor: ProtocolVersionId::latest(),
+            patch: 0.into(),
```
StanislavBreadless, line 87: Probably worth adding a comment on why we don't care about the patch version being correct

core/lib/basic_types/src/protocol_version.rs
```diff
@@ -255,3 +278,106 @@ impl From<ProtocolVersionId> for VmVersion {
         }
     }
 }
+
+basic_type!(
+    /// Patch part of semantic protocol version.
+    VkPatch,
+    u16
```
StanislavBreadless, line 285: why not u32? The packing of the type allows to have u32 patch type, so not sure why not use it all

core/lib/basic_types/src/protocol_version.rs
```diff
@@ -255,3 +278,106 @@ impl From<ProtocolVersionId> for VmVersion {
         }
     }
 }
+
+basic_type!(
+    /// Patch part of semantic protocol version.
+    VkPatch,
```
StanislavBreadless, line 284: Also, why tie specifically to `VkPatch`? I think that a "patch" is generally "a version change that should not affect user-facing behavior". One of the examples of how we could do a patch without changing vks is if we had to somehow amend the priority queue in an emergency

core/lib/dal/src/protocol_versions_dal.rs
```diff
@@ -29,32 +29,63 @@ pub struct ProtocolVersionsDal<'a, 'c> {
...
     ) -> DalResult<()> {
+        let mut db_transaction = self.storage.start_transaction().await?;
+
         sqlx::query!(
             r#"
             INSERT INTO
                 protocol_versions (
                     id,
                     timestamp,
-                    recursion_scheduler_level_vk_hash,
```
StanislavBreadless, line 44: just in case, will the migration cause a small downtime? I.e. the old SQL does not work with the previous DB state and the new one does not work with the old one

core/lib/dal/src/protocol_versions_dal.rs
```diff
@@ -313,20 +322,83 @@ impl ProtocolVersionsDal<'_, '_> {
...
+            minor_version as i32,
+            recursion_scheduler_level_vk_hash.as_bytes()
+        )
+        .instrument("get_patch_version_for_vk")
+        .fetch_optional(self.storage)
+        .await?;
+        Ok(row.map(|row| VkPatch(row.patch as u16)))
+    }
+
+    pub async fn first_vk_patch_for_version(
```
StanislavBreadless, line 353: I think it is better to add a comment on why it is not always zero

core/node/api_server/src/web3/namespaces/en.rs
```diff
@@ -95,12 +98,21 @@ impl EnNamespace {
...
             .context("Genesis batch doesn't exist")?;
 
-        let protocol_version = genesis_batch
+        let minor = genesis_batch
             .protocol_version
             .context("Genesis is not finished")? as u16;
+        let minor = ProtocolVersionId::try_from(minor).context("Malformed protocol version")?;
+        let patch = storage
+            .protocol_versions_dal()
+            .first_vk_patch_for_version(minor)
```
StanislavBreadless, line 107: why first and not last?

infrastructure/zk/src/config.ts
```diff
@@ -180,9 +181,10 @@ export function pushConfig(environment?: string, diff?: string) {
     env.modify('DATABASE_MERKLE_TREE_BACKUP_PATH', `./db/${environment}/backups`, l2InitFile, false);
 
     if (process.env.CONTRACTS_DEV_PROTOCOL_VERSION) {
+        const minor = unpackStringSemVer(process.env.CONTRACTS_DEV_PROTOCOL_VERSION)[1];
         env.modify(
             'CONTRACTS_GENESIS_PROTOCOL_VERSION',
-            (parseInt(process.env.CONTRACTS_GENESIS_PROTOCOL_VERSION!) + 1).toString(),
+            `0.${minor + 1}.0`, // The major version is always 0 for now
```
StanislavBreadless, line 187: It is better to parse the version, since after some upgrades are done the "true" genesis version may become a version with some patch number.

prover/prover_fri_gateway/src/proof_gen_data_fetcher.rs
```diff
@@ -14,19 +14,7 @@ impl PeriodicApiStruct {
...
-        connection
-            .fri_witness_generator_dal()
-            .save_witness_inputs(
-                data.l1_batch_number,
-                &blob_url,
-                data.protocol_version_id,
-                data.eip_4844_blobs,
-            )
-            .await;
+        todo!() // Will be fixed in prover PR.
```
StanislavBreadless, line 17: is it okay to merge the PR in the current state?

core/bin/genesis_generator/src/main.rs
perekopskiy: in genesis generator there is no "correct" patch version, but I will add a comment

core/lib/basic_types/src/protocol_version.rs
perekopskiy: I wanted to be consistent with ProtocolVersionId. Agree, better to use u32

core/lib/basic_types/src/protocol_version.rs
perekopskiy: On a server side patch implies only vk changes, so I came up with this name. Do you think it makes sense to bump patch if the upgrade doesn't change server/prover params? I don't see reasons for that, but if you think so I can amend strcut name and eth sender behavior a bit which currently assumes that each patch changes verification keys

core/lib/dal/src/protocol_versions_dal.rs
perekopskiy: If I haven't missed anything, then the only query that breaks is the one that inserts protocol versions. Given that we won't have upgrades in between applying migrations and updating server it won't cause downtime. Please raise if you think some other query may break and I missed it

core/node/api_server/src/web3/namespaces/en.rs
perekopskiy: In this method we return patch for genesis batch. We don't store mapping (batch number => patch) anywhere so I don't see better option here, note that it only works for **genesis** batch

prover/prover_fri_gateway/src/proof_gen_data_fetcher.rs
perekopskiy: Not okay, the plan is to merge server and prover changes to main in a single PR

mm-zk: can we merge the prover PR to this branch, then?

perekopskiy: Do you mean now? I would prefer to do it at the end when tests pass there and all review comment are resolved. Then we can merge and do final round of review

core/lib/basic_types/src/protocol_version.rs
```diff
@@ -255,3 +278,103 @@ impl From<ProtocolVersionId> for VmVersion {
...
+        Ok(Self {
+            minor,
+            patch: VkPatch(patch),
+        })
+    }
+
+    pub fn pack(&self) -> U256 {
+        (U256::from(self.minor as u16) << U256::from(PACKED_SEMVER_MINOR_OFFSET))
+            | U256::from(self.patch.0)
+    }
```
mm-zk, line 315: can you add a small unittest for pack & parse?

core/lib/basic_types/src/protocol_version.rs
```diff
@@ -255,3 +278,103 @@ impl From<ProtocolVersionId> for VmVersion {
...
+        }
+
+        let minor = parts[1]
+            .parse::<u16>()
+            .map_err(ParseProtocolSemanticVersionError::ParseIntError)?;
+        let minor = ProtocolVersionId::try_from(minor)
+            .map_err(|_| ParseProtocolSemanticVersionError::InvalidFormat)?;
+
+        let patch = parts[2]
+            .parse::<u32>()
```
mm-zk, line 363: I wonder if we also should not allow 'minor' to be 0..

as this generally means that we're reading data from before the semver upgrade

core/lib/basic_types/src/protocol_version.rs
StanislavBreadless: If not hard, I would leave our options open just in case. I agree that 99% of the time those will be VKs, but one of the reasons we started the whole patch thing is so that we could always reuse the same `DefaultUpgrade` contract that requires that the protocol version number goes up

core/node/api_server/src/web3/namespaces/en.rs
StanislavBreadless: okay

core/lib/dal/src/protocol_versions_web3_dal.rs
```diff
@@ -17,11 +17,25 @@ impl ProtocolVersionsWeb3Dal<'_, '_> {
             StorageProtocolVersion,
```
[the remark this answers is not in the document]
perekopskiy: I wasn't changing API method in this PR to avoid breaking changes. Will be done separately

core/node/eth_sender/src/eth_tx_aggregator.rs
```diff
@@ -310,9 +310,15 @@ impl EthTxAggregator {
                     )),
                 ));
             }
-            let protocol_version_id = U256::from_big_endian(&multicall3_protocol_version)
-                .try_into()
-                .unwrap();
+
+            let protocol_version = U256::from_big_endian(&multicall3_protocol_version);
+            // In case the protocol version is smaller than `PACKED_SEMVER_MINOR_MASK`, it will mean that it is
+            // equal to the `protocol_version_id` value, since it the interface from before the semver was supported.
```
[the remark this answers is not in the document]
perekopskiy: I don't think so, `ProtocolVersionId` shoudn't be aware of packing

core/lib/basic_types/src/protocol_version.rs
perekopskiy: This method is only used for parsing newly added config variable. It won't read from old variables and they have different format anyway (not a semver but just a number)

mm-zk: ok

core/lib/dal/src/protocol_versions_web3_dal.rs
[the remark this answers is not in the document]
mm-zk: ack.

core/node/eth_sender/src/eth_tx_aggregator.rs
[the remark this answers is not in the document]
mm-zk: depending in how many places we have such code. (as we are exposing the 'details' - like PACKED_SEMVER_MINOR_MASK).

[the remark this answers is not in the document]
perekopskiy: This should be the only place

core/lib/dal/src/models/storage_protocol_version.rs
```diff
@@ -63,7 +63,7 @@ impl From<StorageProtocolVersion> for api::ProtocolVersion {
             .as_ref()
             .map(|hash| H256::from_slice(hash));
         api::ProtocolVersion {
-            version_id: storage_protocol_version.id as u16,
+            version_id: storage_protocol_version.minor as u16,
```
mm-zk, line 66: I wonder if we shouldn't add the 'patch' info to api::ProtocolVersion too.

core/lib/dal/src/protocol_versions_dal.rs
```diff
@@ -313,20 +322,86 @@ impl ProtocolVersionsDal<'_, '_> {
         })
     }
 
-    pub async fn last_version_id(&mut self) -> DalResult<Option<ProtocolVersionId>> {
-        Ok(sqlx::query!(
+    pub async fn get_patch_versions_for_vk(
```
mm-zk, line 325: what should be the behavior when we query for 'old' minor protocols? should we return `vec![0]` then?

core/lib/dal/src/protocol_versions_dal.rs
```diff
@@ -313,20 +322,86 @@ impl ProtocolVersionsDal<'_, '_> {
...
+            "#,
+            version_id as i32,
+        )
+        .instrument("first_patch_for_version")
+        .fetch_optional(self.storage)
+        .await?;
+        Ok(row.map(|row| VersionPatch(row.patch as u32)))
+    }
+
+    pub async fn latest_semantic_version(&mut self) -> DalResult<Option<ProtocolSemanticVersion>> {
```
mm-zk, line 381: when people deploy this binary version on their existing chain - what will be present in protocol_patches? would that table be empty? (if yes - then this will return an error / None, right?)

core/lib/dal/src/protocol_versions_dal.rs
```diff
@@ -350,20 +425,24 @@ impl ProtocolVersionsDal<'_, '_> {
         Some((id as u16).try_into().unwrap())
     }
 
-    pub async fn all_version_ids(&mut self) -> Vec<ProtocolVersionId> {
+    pub async fn all_versions(&mut self) -> Vec<ProtocolSemanticVersion> {
```
mm-zk, line 428: would allversions return also versions older than 24?

mm-zk: oh - this seems to be used only in tests..

core/node/eth_sender/src/aggregator.rs
```diff
@@ -320,30 +320,42 @@ impl Aggregator {
...
-                .await
-                .unwrap();
-            if verifier_config_for_next_batch != l1_verifier_config {
-                return None;
-            }
-        }
-        let proofs =
-            load_wrapped_fri_proofs_for_range(batch_to_prove, batch_to_prove, blob_store).await;
+            .unwrap();
+        let allowed_patch_versions = storage
```
mm-zk, line 329: please add comment explaining what we're doing here (that there might be multiple verification keys, and that l1_verifier_config tells us what VK L1 can accept)

core/node/eth_sender/src/aggregator.rs
```diff
@@ -320,30 +320,42 @@ impl Aggregator {
...
+            .protocol_versions_dal()
+            .get_patch_versions_for_vk(
+                minor_version,
+                l1_verifier_config.recursion_scheduler_level_vk_hash,
```
mm-zk, line 333: btw - I'm surprised that we do it in '2' rounds (first get minor, and then query with minor).
as AFAIK the single verification key will never be used for 2 different semvers, right?

mm-zk (approved):

core/lib/dal/src/models/storage_protocol_version.rs
perekopskiy: As I've said, I wasn't changing API types for now

core/lib/dal/src/protocol_versions_dal.rs
perekopskiy: yes, check migration, it adds patch 0 for all present protocol version, so the method will return vec![0]

core/lib/dal/src/protocol_versions_dal.rs
perekopskiy: table will have same number of rows as `protocol_versions`, there will patch 0 for each minor version

core/node/eth_sender/src/aggregator.rs
perekopskiy: > as AFAIK the single verification key will never be used for 2 different semvers, right?

no, upgrades might not touch vks, we should take it into account

mm-zk (approved):

core/node/node_sync/src/external_io.rs
```diff
@@ -341,10 +342,13 @@ impl StateKeeperIO for ExternalIO {
...
-                protocol_version
-                    .version_id
-                    .try_into()
-                    .context("cannot convert protocol version")?,
+                ProtocolSemanticVersion {
+                    minor: protocol_version
+                        .version_id
+                        .try_into()
+                        .context("cannot convert protocol version")?,
+                    patch: VersionPatch(0),
```
StanislavBreadless, line 350: Just in case, what is the behavior when on stage we release 0.25.0, but then we realize that it was bad and use 0.25.1 rightaway on testnet. It will mean that the data will be different in ENs and will believe that the provided data is actually for 0.25.0, while it was for 0.25.1

perekopskiy: so ENs will just have some vks and only one patch per minor version, they don't use them so it should be fine

Artemka374 (dismissed):

Artemka374 (dismissed):

Artemka374 (approved):

PR #2060, opened 2024-05-27
fix(tee-verifier): better debug output

thomasknauth:
TEE verifier input producer fails on stage; need better visibility into why.

## Checklist

### DIFF
--- a/core/lib/tee_verifier/src/lib.rs
+++ b/core/lib/tee_verifier/src/lib.rs
@@ -252,7 +252,11 @@ impl TeeVerifierInput {
             }
             (false, TreeLogEntry::ReadMissingKey { .. }) => TreeInstruction::Read(key),
             _ => {
-                tracing::error!("Failed to map LogQuery to TreeInstruction");
+                tracing::error!(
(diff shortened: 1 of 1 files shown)


pbeza (dismissed):

core/lib/tee_verifier/src/lib.rs
```diff
@@ -253,6 +253,7 @@ impl TeeVerifierInput {
             (false, TreeLogEntry::ReadMissingKey { .. }) => TreeInstruction::Read(key),
             _ => {
                 tracing::error!("Failed to map LogQuery to TreeInstruction");
+                tracing::error!("{:?} {:?}", log_query.rw_flag, *tree_log_entry);
```
RomanBrodetski, line 256: nitpick: it would probably fir better directly to the line above - also this log is not temporary so we might forget it here.. I'll approve PR to avoid blocking you, but please resolve in one of the follow-ups

RomanBrodetski (dismissed):

RomanBrodetski (approved):

pbeza (approved):

deniallugo-ml:

PR #2062, opened 2024-05-27
feat: Make house keeper emit correct protocol version

Artemka374:
## What ❔

Emit protocol version from database in metrics.

## Why ❔

Before we were emitting hardcoded protocol version, which is not accurate

## Checklist


slowli (approved):
Nice 👍

PR #2064, opened 2024-05-28
fix(block-reverter): Fix reverting snapshot files

slowli:
## What ❔

- Does not retry "not found" errors when removing objects.
- Makes rolling back snapshot files optional.

## Why ❔

The current implementation leads to very slow reverts because of retries.

## Checklist

### DIFF
--- a/core/lib/object_store/src/gcs.rs
+++ b/core/lib/object_store/src/gcs.rs
@@ -55,7 +55,7 @@ where
     }
 }
 
-pub struct GoogleCloudStorage {
+pub(crate) struct GoogleCloudStorage {
     bucket_prefix: String,
     max_retries: u16,
     client: Client,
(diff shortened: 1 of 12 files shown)


core/lib/object_store/src/gcs.rs
```diff
@@ -17,31 +17,39 @@ use google_cloud_storage::{
...
 {
     let mut retries = 1;
-    let mut backoff = 1;
+    let mut backoff_secs = 1;
     loop {
         match f().await {
             Ok(result) => return Ok(result),
-            Err(err) => {
-                tracing::warn!(%err, "Failed GCS request {retries}/{max_retries}, retrying.");
+            Err(err) if err.is_transient() => {
```
slowli, line 37: This change can lead to issues where `ObjectStore` previously relied on retries for logically non-transient errors (e.g., "not found" errors). Do we want to change it, or use a safer approach (e.g., consider all errors transient / retryable unless removal is concerned)?

core/node/block_reverter/src/lib.rs
```diff
@@ -380,6 +380,8 @@ impl BlockReverter {
         object_store: &dyn ObjectStore,
         deleted_snapshots: &[SnapshotMetadata],
     ) -> anyhow::Result<()> {
+        const CONCURRENT_REMOVE_REQUESTS: usize = 20;
```
slowli, line 383: Is this a reasonable value? Unfortunately, I haven't managed to find "native" batching for GCS API (it looks like classic REST API, so it's not quite unexpected).

core/node/block_reverter/src/lib.rs
```diff
@@ -419,18 +421,46 @@ impl BlockReverter {
...
+                        if !is_incomplete_snapshot {
+                            is_incomplete_snapshot = true;
+                            tracing::warn!(
+                                "Snapshot for L1 batch #{} is incomplete (misses al least storage logs chunk ID {chunk_id}). \
+                                 It is probable that it's currently being created, in which case you'll need to clean up produced files \
+                                 manually afterwards",
+                                snapshot.l1_batch_number
+                            );
+                        }
+                        return None;
```
slowli, line 438: Maybe, it does make sense to attempt removing "future" storage log chunks?

core/lib/object_store/src/gcs.rs
popzxc: I think the change makes total sense. If it occurs that it breaks something (I _really_ hope it does not), we can simply revert the commit and reimplement the change with the newly obtained knowledge.

core/node/block_reverter/src/lib.rs
RomanBrodetski: I think it's reasonable

core/node/block_reverter/src/lib.rs
RomanBrodetski: I think it doesn't - snapshot creator shouldn't be running while a rollback is taking place. Let's keep as is and remember/revisit when we revamping the block revertion next wave

RomanBrodetski (approved):

PR #2068, opened 2024-05-28
feat(sync-layer): adapt MiniMerkleTree to manage priority queue

ly0va:
## What ❔

Enables the MiniMerkleTree to be used for priority queue to efficiently calculate merkle proofs for priority transactions.

## Why ❔

As part of the preparation for the priority queue migration to sync layer.

## Checklist


core/lib/mini_merkle_tree/src/lib.rs
```diff
@@ -86,67 +98,164 @@ where
...
+        if self.head_index + self.hashes.len() > self.binary_tree_size {
+            self.binary_tree_size *= 2;
+        }
+    }
+
+    /// Caches the rightmost `count` leaves.
+    /// Does not affect the root hash, but makes it impossible to get the paths to the cached leaves.
+    /// # Panics
+    /// Panics if `count` is greater than the number of non-cached leaves in the tree.
+    pub fn cache(&mut self, count: usize) {
```
slowli, line 164: Bikeshedding: I think this naming is confusing; it doesn't *just* cache data, but also trims some elements from the tree representation. I'd call it something like `trim_start()`.

core/lib/mini_merkle_tree/src/lib.rs
```diff
@@ -86,67 +98,164 @@ where
...
+        assert!(self.hashes.len() >= count, "not enough leaves to cache");
+        let mut new_cache = vec![];
+        self.compute_merkle_root_and_path(count - 1, None, Some(&mut new_cache));
+        self.hashes.drain(0..count);
+        self.head_index += count;
+        self.left_cache = new_cache;
     }
 
     fn compute_merkle_root_and_path(
```
slowli, line 173: This code looks quite confusing and difficult to maintain.

- I'd suggest to push the elements from the cached path to `hashes` as the first element once you've reached an appropriate level. In this way, you eliminate quite a few corner cases; e.g., `hashes` will always start with an even index.
- I may misunderstand what `cache` is, but isn't it essentially a Merkle path for the first retained tree element? (Technically, it can be a subset of this path to the left of the retained element.) If so, why do you need a separate argument / computation for this path, rather than getting it as a left path for the first retained element?

core/lib/mini_merkle_tree/src/lib.rs
```diff
@@ -19,16 +19,26 @@ use zksync_crypto::hasher::{keccak::KeccakHasher, Hasher};
 /// we unlikely to ever hit.
 const MAX_TREE_DEPTH: usize = 32;
 
-/// In-memory Merkle tree of bounded depth (no more than 10).
+/// In-memory Merkle tree of bounded depth (no more than 32).
```
slowli, line 22: Bikeshedding: Technically, it's now not a full Merkle tree, but rather a partial tree, potentially with the tree start trimmed. I'd rename the type `PartialMerkleTree` or `TrimmedMerkleTree` correspondingly.

core/lib/mini_merkle_tree/src/lib.rs
```diff
@@ -86,67 +98,164 @@ where
...
-    pub fn merkle_root_and_path(self, index: usize) -> (H256, Vec<H256>) {
-        let mut merkle_path = vec![];
-        let root_hash = self.compute_merkle_root_and_path(index, Some(&mut merkle_path));
-        (root_hash, merkle_path)
+    /// `index` is relative to the leftmost uncached leaf.
+    pub fn merkle_root_and_path(&self, index: usize) -> (H256, Vec<H256>) {
+        let (mut left_path, mut right_path) = (vec![], vec![]);
+        let root_hash =
+            self.compute_merkle_root_and_path(index, Some((&mut left_path, &mut right_path)), None);
+        (root_hash, right_path)
```
slowli, line 131: - `left_path` looks unused here. It may make sense to pass `left_path` and `right_path` as separate `Option<&mut _>` values.
- Bikeshedding: IMO, "start path" / "end path" naming would be better because "left" can also be interpreted as a subset of some Merkle path to the left of something.

core/lib/mini_merkle_tree/src/lib.rs
```diff
@@ -86,67 +98,164 @@ where
...
+    }
+
+    /// Returns the root hash and the Merkle proofs for an interval of leafs.
+    /// The interval is [0, `length`), where `0` is the leftmost uncached leaf.
+    pub fn merkle_root_and_paths_for_interval(
```
slowli, line 136: Bikeshedding: Subjectively, "range" feels like a better term to use here rather than "interval"; "interval" often refers to a time interval in our codebase.

core/lib/mini_merkle_tree/src/lib.rs
```diff
@@ -86,67 +98,164 @@ where
...
+        (root_hash, left_path, right_path)
+    }
+
+    /// Adds a new leaf to the tree (replaces leftmost empty leaf).
+    /// If the tree is full, its size is doubled.
+    /// Note: empty leaves != zero leaves.
+    pub fn push(&mut self, leaf: [u8; LEAF_SIZE]) {
+        let leaf_hash = self.hasher.hash_bytes(&leaf);
+        self.hashes.push_back(leaf_hash);
+        if self.head_index + self.hashes.len() > self.binary_tree_size {
```
slowli, line 155: Bikeshedding: IMO, `start_index` would be a better name; "head" could be interpreted as something related to the root of the tree.

core/lib/mini_merkle_tree/src/tests.rs
```diff
@@ -156,24 +156,81 @@ fn verify_merkle_proof(
...
+    right_path: &[H256],
+    merkle_root: H256,
+) {
+    assert_eq!(left_path.len(), right_path.len());
+
+    let hasher = KeccakHasher;
+    let mut hashes: Vec<_> = items.iter().map(|item| hasher.hash_bytes(item)).collect();
+    let mut level_len = hashes.len();
+
+    for (left_item, right_item) in left_path.iter().zip(right_path.iter()) {
```
slowli, line 172: This can be simplified using the same approach I've mentioned above, especially since performance is not a concern for tests: Construct hashes on a certain level as `left_path[level]? || hashes || right_path[level]?`, so that elements start from an even index and contain all data to create the next level by zipping hashes pairwise.

core/lib/mini_merkle_tree/src/lib.rs
```diff
@@ -19,16 +19,26 @@ use zksync_crypto::hasher::{keccak::KeccakHasher, Hasher};
...
 /// can be specified larger than the number of provided leaves. In this case, the remaining leaves
 /// will be considered to equal `[0_u8; LEAF_SIZE]`.
+///
+/// The tree has dynamic size, meaning that it can grow by a factor of 2 when the number of leaves
+/// exceeds the current tree size. It does not shrink.
+///
+/// The tree is optimized for the case when the queries are performed on the rightmost leaves
+/// and the leftmost leaves are cached. Caching enables the merkle roots and paths to be computed
+/// in `O(n)` time, where `n` is the number of uncached leaves (in contrast to the total number of
+/// leaves). Cache itself only takes up `O(depth)` space.
```
perekopskiy, line 34: nit: I think it should be `O(max(n, height))` instead of `O(n)` time

core/lib/mini_merkle_tree/src/lib.rs
```diff
@@ -86,67 +118,135 @@ where
             hasher,
             hashes,
             binary_tree_size,
+            start_index: 0,
+            cache: vec![],
```
slowli, line 122: Don't like this. As it's implemented right now, `cache` is the the Merkle path for the first retained element in the partial tree, unless the tree was never trimmed, or after it was entirely trimmed (AFAIU, in the second case, `cache` will contain garbage). These two corner case provide potential for bugs. IMO, it would make sense to:

- Either consistently compute `cache` so that it's *always* correctly defined. I.e., recompute it in the constructor. Regarding complete trimming, it's more tricky; you can compute the cache right away if you define it as a *subset* of a Merkle path to the left of the next tree element. In this case, I'd strongly suggest to signal this explicitly by defining a Merkle path subset as `Vec<Option<H256>>`. (For symmetry, the right path would be a subset of a Merkle path as well.) This would also make defining `cache` in the tree constructor trivial; it would be `vec![None; tree_depth]`.
- Or, make `cache` optional and recompute it on demand (probably would require some sort of interior mutability).

IMO, the first approach would be more appropriate here.

core/lib/mini_merkle_tree/src/lib.rs
```diff
@@ -86,67 +118,135 @@ where
...
+            left_path.reserve(depth);
+        }
+        if let Some(right_path) = end_path.as_deref_mut() {
+            right_path.reserve(depth);
         }
 
-        let mut hashes = self.hashes;
-        let mut level_len = hashes.len();
+        let mut hashes = self.hashes.clone();
+        let mut start_index = self.start_index;
```
slowli, line 215: Bikeshedding: This is confusing given that `end_index` is relative and `start_index` is absolute. I'd suggest to call the variable `absolute_start_index`.

core/lib/mini_merkle_tree/src/lib.rs
```diff
@@ -86,67 +118,135 @@ where
...
+        self.cache = new_cache;
     }
 
     fn compute_merkle_root_and_path(
-        self,
-        mut index: usize,
-        mut merkle_path: Option<&mut Vec<H256>>,
+        &self,
+        mut end_index: usize,
+        mut start_path: Option<&mut Vec<H256>>,
```
slowli, line 203: Dumb question: Since the relative start index is always zero, doesn't the `start_path` always *conceptually* equal to `cache` (except for the root hash at the end of `cache`)? See also the comment about partial Merkle paths; the paths are not exactly equal right now due to the differing right subset of the path (which can change when new elements are inserted to the tree after `cache` is computed), but the right subset should not be used in proof verification anyway.

core/lib/mini_merkle_tree/src/lib.rs
ly0va: I'm not sure this will help the understanding of what it does.

- It can still be used as a full Merkle tree
- I believe `Mini` is still a good enough descriptor since it highlights a property with a more immediate consequence
- To understand what is meant by `Partial` it will anyway be necessary to read the docs

core/lib/mini_merkle_tree/src/lib.rs
```diff
@@ -86,67 +118,135 @@ where
...
-                merkle_path.push(adjacent_hash);
+            if start_index % 2 == 1 {
+                hashes.push_front(self.cache[level]);
+            }
+            if hashes.len() % 2 == 1 {
+                hashes.push_back(empty_hash_at_level);
             }
 
-            for i in 0..(level_len / 2) {
+            let push_sibling_hash = |path: Option<&mut Vec<H256>>, index: usize| {
```
slowli, line 226: Bikeshedding (sort of): IMO, this is still quite complicated. I've come up with this alternative (doesn't include changes discussed in other line comments):

```rust
let mut hashes = self.hashes.clone();
let mut absolute_start_index = self.start_index;

for level in 0..depth {
    let empty_hash_at_level = self.hasher.empty_subtree_hash(level);
    let mut start_index = 0; // relative index in `hashes`
    if absolute_start_index % 2 == 1 {
        hashes.push_front(self.cache[level]);
        start_index += 1;
        end_index += 1;
    }
    if hashes.len() % 2 == 1 {
        hashes.push_back(empty_hash_at_level);
    }

    // Probably redundant (see other comments)
    if let Some(path) = start_path.as_deref_mut() {
        path.push(hashes[start_index ^ 1]);
    }
    if let Some(path) = end_path.as_deref_mut() {
        let hash = hashes.get(end_index ^ 1).copied().unwrap_or_default(); // looks bad; see the other comment
        path.push(hash);
    }

    let level_len = hashes.len() / 2;
    for i in 0..level_len {
        hashes[i] = self.hasher.compress(&hashes[2 * i], &hashes[2 * i + 1]);
    }

    hashes.truncate(level_len);
    end_index /= 2;
    absolute_start_index /= 2;
}
```

The idea is to use relative indexing everywhere, other than when pushing `cache`d entries.

core/lib/mini_merkle_tree/src/lib.rs
```diff
@@ -86,67 +118,135 @@ where
...
+                // `index` is relative to `head_index`
+                if let Some(path) = path {
+                    let sibling = ((start_index + index) ^ 1) - start_index + start_index % 2;
+                    // When trimming all existing leaves,
+                    // the first untrimmed leaf and its sibling might not exist.
+                    let hash = hashes.get(sibling).copied().unwrap_or_default();
```
slowli, line 232: Dumb question: Why is using a default hash (rather than `empty_hash_at_level`) OK here? See also the other comment regarding Merkle path subsets. IIUC, missing `sibling` here means that a sibling would be to the right; that is, it wouldn't be used anyway. But without additional explanation and signaling via types, this is quite obscure.

core/lib/mini_merkle_tree/src/lib.rs
```diff
@@ -86,67 +118,135 @@ where
...
             }
-            if level_len % 2 == 1 {
-                hashes[level_len / 2] = self
-                    .hasher
-                    .compress(&hashes[level_len - 1], &empty_hash_at_level);
-            }
 
-            index /= 2;
-            level_len = level_len / 2 + level_len % 2;
+            hashes.drain(level_len..);
```
slowli, line 245: Nit: IIUC, you can use [`VecDeque::truncate()`](https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.truncate) here and in similar places.

core/lib/mini_merkle_tree/src/lib.rs
ly0va: Good catch

core/lib/mini_merkle_tree/benches/tree.rs
```diff
@@ -10,11 +10,7 @@ const TREE_SIZES: &[usize] = &[32, 64, 128, 256, 512, 1_024];
 fn compute_merkle_root(bencher: &mut Bencher<'_>, tree_size: usize) {
     let leaves = (0..tree_size).map(|i| [i as u8; 88]);
     let tree = MiniMerkleTree::new(leaves, None);
-    bencher.iter_batched(
-        || tree.clone(),
-        MiniMerkleTree::merkle_root,
-        BatchSize::SmallInput,
-    );
+    bencher.iter_batched(|| &tree, MiniMerkleTree::merkle_root, BatchSize::SmallInput);
```
slowli, line 13: Since `merkle_root()` (and `merkle_root_and_path()` below) no longer consume the tree, you can just use `iter()` here and below:

```rust
bencher.iter(|| tree.merkle_root());
```

core/lib/mini_merkle_tree/src/lib.rs
```diff
@@ -86,67 +120,146 @@ where
...
+    /// Returns `true` if the tree is empty.
+    pub fn is_empty(&self) -> bool {
+        self.start_index == 0 && self.hashes.is_empty()
+    }
+
     /// Returns the root hash of this tree.
     /// # Panics
-    /// Will panic if the constant below is invalid.
-    pub fn merkle_root(self) -> H256 {
+    /// Should never panic, unless there is a bug.
```
slowli, line 135: Nit: No need to document a panic then; instead, put `#[allow(clippy::missing_panics_doc)]` on the method with a this comment.

core/lib/mini_merkle_tree/src/lib.rs
```diff
@@ -86,67 +120,146 @@ where
...
+    /// `index` is relative to the leftmost uncached leaf.
+    /// # Panics
+    /// Panics if `index` is >= than the number of leaves in the tree.
+    pub fn merkle_root_and_path(&self, index: usize) -> (H256, Vec<H256>) {
+        assert!(index < self.hashes.len(), "leaf index out of bounds");
+        let mut end_path = vec![];
+        let root_hash = self.compute_merkle_root_and_path(index, Some(&mut end_path), None);
+        (
+            root_hash,
+            end_path.into_iter().map(Option::unwrap).collect(),
```
slowli, line 158: This awkwardness with `unwrap()`s makes me think that an alternative with filtering a full path after getting it from `compute_merkle_root_and_path()` would be better. I.e., `compute_merkle_root_and_path()` would have two args and would always set the full path; and then it could be filtered based on the side and index of the element that the path corresponds to.

core/lib/mini_merkle_tree/src/lib.rs
```diff
@@ -86,67 +120,146 @@ where
...
+    }
+
+    /// Returns the root hash and the Merkle proofs for a range of leafs.
+    /// The range is 0..length, where `0` is the leftmost untrimmed leaf.
+    /// # Panics
+    /// Panics if `length` is 0 or greater than the number of leaves in the tree.
+    pub fn merkle_root_and_paths_for_range(
+        &self,
+        length: usize,
+    ) -> (H256, Vec<Option<H256>>, Vec<Option<H256>>) {
```
slowli, line 169: As a further suggestion to the previous comment: maybe, it makes sense to create a newtype for Merkle paths returned here?

```rust
pub struct MerklePath(Vec<Option<H256>>);

impl MerklePath {
    fn new(full_path: Vec<H256>, side: Side) -> Self {
        // ...
    }

    // needs some sort of a getter as well; not sure how exactly it should look
}
```

core/lib/mini_merkle_tree/src/lib.rs
```diff
@@ -86,67 +120,146 @@ where
...
+        let mut right_path = vec![];
+        let root_hash =
+            self.compute_merkle_root_and_path(length - 1, Some(&mut right_path), Some(Side::Right));
+        (root_hash, self.cache.clone(), right_path)
+    }
+
+    /// Adds a raw hash to the tree (replaces leftmost empty leaf).
+    /// If the tree is full, its size is doubled.
+    /// Note: empty leaves != zero leaves.
+    pub fn push_hash(&mut self, leaf_hash: H256) {
```
slowli, line 181: Sorry for not asking this previously: Is it intentional that adding hashes to the tree via this method and `from_hashes` is public? IMO, it's slightly confusing that there are multiple ways to add data to the tree.

core/lib/mini_merkle_tree/src/lib.rs
```diff
@@ -86,67 +120,146 @@ where
...
+    pub fn trim_start(&mut self, count: usize) {
+        assert!(self.hashes.len() >= count, "not enough leaves to cache");
+        let mut new_cache = vec![];
+        // Cache is a left subset of the path to the first untrimmed leaf.
+        let root = self.compute_merkle_root_and_path(count, Some(&mut new_cache), Some(Side::Left));
+        self.hashes.drain(..count);
+        self.start_index += count;
+        if self.start_index == self.binary_tree_size {
+            // In order to be able to get the root of a completely trimmed tree,
+            // we need to cache it.
```
slowli, line 212: Bikeshedding: IMO, this comment is confusing. The root is added not in any case when the tree is completely trimmed, but if it is completely trimmed *and* pushing a next element will lead to tree depth growth (since `binary_tree_size` is always a power of 2). IMO, it could be useful to document an invariant what `cache` is, at the corresponding field (instead of in `trim_start()`), so that it's more clear:

> Left subset of the Merkle path to the first untrimmed leaf (i.e., a leaf with index `self.start_index`). Used to fill in data for trimmed tree leaves when computing Merkle paths and the root hash. Because only the left subset of the path is used, the cache is not invalidated when new leaves are pushed into the tree. If all leaves are trimmed, `cache` is the left subset of the Merkle path to the next leaf to be inserted, which still has index `self.start_index`.

core/lib/mini_merkle_tree/src/lib.rs
```diff
@@ -86,67 +120,146 @@ where
...
+        let leaf_hash = self.hasher.hash_bytes(&leaf);
+        self.push_hash(leaf_hash);
+    }
+
+    /// Caches the rightmost `count` leaves.
+    /// Does not affect the root hash, but makes it impossible to get the paths to the cached leaves.
+    /// # Panics
+    /// Panics if `count` is greater than the number of non-cached leaves in the tree.
+    pub fn trim_start(&mut self, count: usize) {
+        assert!(self.hashes.len() >= count, "not enough leaves to cache");
```
slowli, line 204: Nit: The panic message here is confusing; `cache` is an implementation detail. IMO, it'd make sense to specify here that it's impossible to prune more leaves than the pruned tree has.

Ditto for the method doc comment; the method is (IMO, quite appropriately) called `trim_start` rather than `cache_start`. Additionally, isn't it the *leftmost* `count` leaves (in the trimmed tree) that are getting cached?

core/lib/mini_merkle_tree/src/lib.rs
```diff
@@ -86,67 +120,146 @@ where
...
+        if let Some(right_path) = end_path.as_deref_mut() {
+            right_path.reserve(depth);
+        }
+
+        let mut hashes = self.hashes.clone();
+        let mut absolute_start_index = self.start_index;
+
+        if hashes.is_empty() {
+            hashes.push_back(self.hasher.empty_subtree_hash(0));
         }
```
slowli, line 234: This doesn't look necessary. If there's a cached leaf hash at 0th level, then the 0th level will be constructed from 2 entries correctly. If there's no cached leaf hash at 0th level, then pushing an empty tree hash will compute `self.hasher.empty_subtree_hash(1)`, which would be used on level 1 anyway.

core/lib/mini_merkle_tree/src/tests.rs
```diff
@@ -156,24 +158,79 @@ fn verify_merkle_proof(
...
+        verify_range_merkle_proof(&leaves[..i], 0, &start_path, &end_path, merkle_root);
+    }
+
+    tree.trim_start(25);
+    leaves.drain(..25);
+
+    for i in 1..=25 {
+        let (merkle_root, start_path, end_path) = tree.merkle_root_and_paths_for_range(i);
+        verify_range_merkle_proof(&leaves[..i], 25, &start_path, &end_path, merkle_root);
+    }
```
slowli, line 224: I'd advise to test more thoroughly, e.g. like this

```rust
let mut start_index = 0;
for trimmed_count in 1..10 {
    tree.trim_start(trimmed_count);
    leaves.drain(..trimmed_count);
    start_index += trimmed_count;
    let tree_len = tree.hashes.len();

    for i in 1..=tree_len {
        let (merkle_root, start_path, end_path) = tree.merkle_root_and_paths_for_range(i);
        verify_range_merkle_proof(
            &leaves[..i],
            start_index,
            &start_path,
            &end_path,
            merkle_root,
        );
    }
}
```

I.e., remove 1, 2, 3, ..., 9 leaves at a time and test Merkle proofs for all remaining leaves in the tree.

core/lib/mini_merkle_tree/src/tests.rs
```diff
@@ -205,15 +262,127 @@ fn merkle_proofs_are_valid_in_very_small_trees() {
...
+        .unwrap();
+    assert_eq!(tree.merkle_root(), expected_root);
+}
+
+#[test]
+fn trim_all_and_check_root() {
+    let mut tree = MiniMerkleTree::new(iter::repeat([1; 88]).take(4), None);
+    let root = tree.merkle_root();
+    tree.trim_start(4);
+    assert_eq!(tree.merkle_root(), root);
```
slowli, line 382: Nit: Similar to the test above, you can test this more thoroughly, e.g. by testing trees with varying number of leaves:

```rust
for len in 1..=100 {
    let mut tree = MiniMerkleTree::new(iter::repeat([1; 88]).take(len), None);
    let root = tree.merkle_root();
    tree.trim_start(len);
    assert_eq!(tree.merkle_root(), root);
}
```

core/lib/mini_merkle_tree/src/lib.rs
```diff
@@ -19,16 +19,27 @@ use zksync_crypto::hasher::{keccak::KeccakHasher, Hasher};
...
+/// leaves). Cache itself only takes up `O(depth)` space. However, caching prevents the retrieval of paths to the
+/// cached leaves.
 #[derive(Debug, Clone)]
 pub struct MiniMerkleTree<const LEAF_SIZE: usize, H = KeccakHasher> {
     hasher: H,
-    hashes: Box<[H256]>,
+    hashes: VecDeque<H256>,
     binary_tree_size: usize,
+    start_index: usize,
+    cache: Vec<Option<H256>>,
```
StanislavBreadless, line 42: Better to include what does `cache[i]` contains. At the start I thought that `cache[0]` is for root, but then when reading I realized that it is not

core/lib/mini_merkle_tree/src/lib.rs
```diff
@@ -86,67 +120,146 @@ where
...
     }
 
     fn compute_merkle_root_and_path(
-        self,
-        mut index: usize,
-        mut merkle_path: Option<&mut Vec<H256>>,
+        &self,
+        mut end_index: usize,
+        mut end_path: Option<&mut Vec<Option<H256>>>,
+        side: Option<Side>,
```
StanislavBreadless, line 222: Could you please add comments on the meaning of these params

core/lib/mini_merkle_tree/src/lib.rs
```diff
@@ -86,67 +120,146 @@ where
...
         for level in 0..depth {
-            let empty_hash_at_level = self.hasher.empty_subtree_hash(level);
-
-            if let Some(merkle_path) = merkle_path.as_deref_mut() {
-                let adjacent_idx = index ^ 1;
-                let adjacent_hash = if adjacent_idx < level_len {
-                    hashes[adjacent_idx]
-                } else {
-                    empty_hash_at_level
+            if absolute_start_index % 2 == 1 {
```
StanislavBreadless, line 237: better to add a comment on this step

core/lib/mini_merkle_tree/src/lib.rs
```diff
@@ -86,67 +120,146 @@ where
...
+                hashes.push_front(self.cache[level].expect("cache is invalid"));
+                end_index += 1;
+            }
+            if hashes.len() % 2 == 1 {
```
StanislavBreadless, line 241: Could you please add a comment about this step

core/lib/mini_merkle_tree/src/lib.rs
```diff
@@ -86,67 +120,146 @@ where
...
+        self.cache = new_cache;
     }
 
     fn compute_merkle_root_and_path(
-        self,
-        mut index: usize,
-        mut merkle_path: Option<&mut Vec<H256>>,
+        &self,
+        mut end_index: usize,
+        mut end_path: Option<&mut Vec<Option<H256>>>,
```
StanislavBreadless, line 221: better to add a comment on what does `None` mean in a merkle path

core/lib/mini_merkle_tree/src/lib.rs
```diff
@@ -156,6 +269,12 @@ fn tree_depth_by_size(tree_size: usize) -> usize {
     tree_size.trailing_zeros() as usize
 }
 
+#[derive(Debug, Clone, Copy)]
+enum Side {
```
StanislavBreadless, line 273: Add a comment on what `Side` means

core/lib/mini_merkle_tree/src/lib.rs
ly0va: It is, since without them the only interface we have supports adding constant-sized arrays, whereas priority operations can be dynamically sized. 

Perhaps it would be better for `push` and `with_hasher` to accept slices instead?

slowli: Probably yes, or make the tree parametric by the value type instead of using `const LEAF_SIZE` parameter (it doesn't makes sense for priority ops, does it?). Previously, it only accepted fixed-size arrays because we didn't need anything else 🙃 Just to clarify, what would be an empty hash for a priority op? The hash of an empty byte slice? You didn't refactor `HashEmptySubtree` or added more impls for it; and it feels arbitrary to use a hash of `[0_u8; 88]` for priority ops.

core/lib/mini_merkle_tree/src/lib.rs
ly0va: It won't always be the full path anyway, since if we try to fully trim a tree, the path is bound to have some `None`s anyway. And we agreed that using garbage/empty values instead of `None` is worse than explicitly signaling it with types.

Also, I think having `end_path.into_iter().map(Option::unwrap).collect()`
is not that bad and much more readable than having filtering like

```rust
fn filter_merkle_path_by_side(
    path: Vec<Option<H256>>,
    index: usize,
    side: Side,
) -> Vec<Option<H256>> {
    path.into_iter()
        .enumerate()
        .map(|(i, hash)| {
            if (index & (1 << i)) >> i == side as usize {
                None
            } else {
                hash
            }
        })
        .collect()
}
```

core/lib/mini_merkle_tree/src/lib.rs
```diff
@@ -86,67 +120,146 @@ where
...
-
         let depth = tree_depth_by_size(self.binary_tree_size);
-        if let Some(merkle_path) = merkle_path.as_deref_mut() {
-            merkle_path.reserve(depth);
+        if let Some(right_path) = end_path.as_deref_mut() {
+            right_path.reserve(depth);
+        }
+
+        let mut hashes = self.hashes.clone();
+        let mut absolute_start_index = self.start_index;
```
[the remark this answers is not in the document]
ly0va: no, it's relative to `absolute_start_index`

core/lib/mini_merkle_tree/src/lib.rs
ly0va: This is a good point. I suggest we keep the `push_hash` and `from_hashes` since initializing the tree can be faster that way (since hashes are stored in Postgres, so no need to rehash when initializing). 

Re: `HashEmptySubtree` you are right, I'll need to have a separate impl for priority ops. I would rather come back to it in the next PR where I would integrate the tree into `eth_watch` and `eth_sender`. 

I like the idea of making the tree generic over leaf type, instead of over `LEAF_SIZE`, done in the last commit

core/lib/mini_merkle_tree/src/lib.rs
```diff
@@ -156,24 +287,34 @@ fn tree_depth_by_size(tree_size: usize) -> usize {
...
+    /// Implementations are encouraged to cache the returned values.
     fn empty_subtree_hash(&self, depth: usize) -> H256 {
-        static EMPTY_TREE_HASHES: Lazy<Vec<H256>> = Lazy::new(compute_empty_tree_hashes::<88>);
-        EMPTY_TREE_HASHES[depth]
+        static EMPTY_TREE_HASHES: OnceCell<Vec<H256>> = OnceCell::new();
+        EMPTY_TREE_HASHES.get_or_init(|| compute_empty_tree_hashes(self.empty_hash()))[depth]
+    }
+
+    /// Returns an empty hash
+    fn empty_hash(&self) -> H256;
```
slowli, line 308: Bikeshedding: `empty_leaf_hash`?

core/lib/mini_merkle_tree/src/lib.rs
slowli: > if we try to fully trim a tree, the path is bound to have some Nones anyway

Do you mean that the path would have values that cannot be cached (since they would be invalidated when new leaves are inserted into the tree)? If so, this doesn't contradict my proposal; in it, `compute_merkle_root_and_path` returns a current full path to the specified leaf in the tree, and it is caller's responsibility to decide what to do with it. So e.g., if the path is intended for caching, its right side needs to be removed.

Of course, it's subjective, but I like `filter_merkle_path_by_side()` better (maybe together with a proposal to have a special type for it); the only thing in your snippet I don't think makes sense is `path` type; should be `Vec<H256>` (i.e., a *full* path). IIUC, `compute_merkle_root_and_path()` can create full paths just by always inserting `hashes[index ^ 1]` to the path on each level; or am I missing something?

slowli (approved):
Still think that an approach with filtering a full Merkle path returned from `compute_merkle_root_and_path` would be better, but the current implementation is good enough™.

PR #2072, opened 2024-05-28
fix(api): correct default fee data in eth call

StanislavBreadless:
## What ❔

Previously, if no `max_fee_per_gas` was provided, it would mean that we would use 0 as the base fee. This would also require the gas per pubdata to be 0.

With this PR we will now use the current gas per pubdata. This PR also provides a large eth call gas limit to ensure that it will work fine even under very high L1 gas price

## Checklist


core/tests/ts-integration/package.json
```diff
@@ -4,7 +4,7 @@
   "license": "MIT",
   "private": true,
   "scripts": {
-    "test": "zk f jest --forceExit --testTimeout 60000",
+    "test": "zk f jest --forceExit --testTimeout 60000 -- system.test.ts",
```
Deniallugo, line 7: seems you shouldn't specify it

core/node/api_server/src/execution_sandbox/execute.rs
```diff
@@ -170,23 +170,34 @@ impl TransactionExecutor {
...
+            .await
+            .context("failed to resolve block info")?
+            .protocol_version;
+        drop(connection);
+
         // Protection against infinite-loop eth_calls and alike:
         // limiting the amount of gas the call can use.
         // We can't use `BLOCK_ERGS_LIMIT` here since the VM itself has some overhead.
-        tx.common_data.fee.gas_limit = ETH_CALL_GAS_LIMIT.into();
+        tx.common_data.fee.gas_limit = get_eth_call_gas_limit(protocol_version.into()).into();
```
perekopskiy, line 200: - Comment above says we can't use `BLOCK_ERGS_LIMIT` but it's used for the latest VM, is comment misleading?
- Do I understand correctly that bootloader will limit computational gas for tx anyway?

core/tests/ts-integration/tests/system.test.ts
```diff
@@ -311,6 +311,12 @@ describe('System behavior checks', () => {
         ).toBeAccepted();
     });
 
+    test('Gas per pubdata byte getter should work', async () => {
+        const systemContextArtifact = getTestContract('ISystemContext');
+        const systemContext = new ethers.Contract(SYSTEM_CONTEXT_ADDRESS, systemContextArtifact.abi, alice.provider);
+        console.log(await systemContext.gasPerPubdataByte());
```
perekopskiy, line 317: Can the result be somehow asserted instead of being printed?

core/node/api_server/src/execution_sandbox/execute.rs
Deniallugo, line 200: i'm not sure it's the correct place for setting it. Probably, it's better to pass an argument, `use_max_limit` or so

StanislavBreadless: 1. The comment is outdated
2. Yes

core/tests/ts-integration/tests/system.test.ts
StanislavBreadless: definitely, sorry

core/node/api_server/src/execution_sandbox/execute.rs
```diff
@@ -170,23 +170,35 @@ impl TransactionExecutor {
...
+        let mut connection = connection_pool
+            .connection_tagged("api")
+            .await
+            .context("failed acquiring DB connection")?;
+        let protocol_version = block_args
+            .resolve_block_info(&mut connection)
+            .await
+            .context("failed to resolve block info")?
+            .protocol_version;
+        drop(connection);
```
Deniallugo, line 197: i really don't fill it's correct to get protocol version here, we get it later one more time

core/node/api_server/src/web3/namespaces/debug.rs
```diff
@@ -173,6 +175,7 @@ impl DebugNamespace {
                 vm_permit,
                 shared_args,
                 self.state.connection_pool.clone(),
+                call_overrides,
                 tx.clone(),
```
slowli, line 179: Dumb question: I see that `tx.common_data.fee.gas_limit.as_u64()` is used below to create a `Call`. IIUC, it won't use `call_overrides` to patch this value. Is this expected?

core/lib/multivm/src/utils.rs
```diff
@@ -446,6 +446,33 @@ pub fn get_max_batch_gas_limit(version: VmVersion) -> u64 {
...
+        VmVersion::VmVirtualBlocksRefundsEnhancement => {
+            crate::vm_refunds_enhancement::constants::ETH_CALL_GAS_LIMIT as u64
+        }
+        VmVersion::VmBoojumIntegration => {
+            crate::vm_boojum_integration::constants::ETH_CALL_GAS_LIMIT as u64
+        }
+        VmVersion::Vm1_4_1 => crate::vm_1_4_1::constants::ETH_CALL_GAS_LIMIT as u64,
+        VmVersion::Vm1_4_2 => crate::vm_1_4_2::constants::ETH_CALL_GAS_LIMIT as u64,
+        VmVersion::Vm1_5_0SmallBootloaderMemory => crate::vm_latest::constants::ETH_CALL_GAS_LIMIT,
+        VmVersion::Vm1_5_0IncreasedBootloaderMemory => {
```
slowli, line 470: Nit: Can't these two variants be united using `|`? IIUC, they return the same value.

core/node/api_server/src/web3/namespaces/debug.rs
StanislavBreadless: good point, this issue was there even before my PR

core/tests/ts-integration/tests/system.test.ts
```diff
@@ -311,6 +311,16 @@ describe('System behavior checks', () => {
...
     });
 
+    test('Gas per pubdata byte getter should work', async () => {
+        const systemContextArtifact = getTestContract('ISystemContext');
+        const systemContext = new ethers.Contract(SYSTEM_CONTEXT_ADDRESS, systemContextArtifact.abi, alice.provider);
+        const currentGasPerPubdata = await systemContext.gasPerPubdataByte();
+
+        // The current gas per pubdata depends on a lot of factors, so it wouldn't be sustainable to check the exact value.
+        // We'll just check that it is greater than zero.
+        expect(currentGasPerPubdata.toNumber()).toBeGreaterThan(0);
```
perekopskiy, line 321: test fails for validium

core/node/api_server/src/tx_sender/mod.rs
```diff
@@ -1036,4 +1039,19 @@ impl TxSender {
...
+        &self,
+        block_args: BlockArgs,
+    ) -> anyhow::Result<u64> {
+        let mut connection = self.acquire_replica_connection().await?;
+
+        let protocol_version = block_args
+            .resolve_block_info(&mut connection)
+            .await
+            .context("failed to resolve block info")?
+            .protocol_version;
```
slowli, line 1053: Nit / future work: It may make sense to create a dedicated method for resolving the protocol version since `resolve_block_info` loads a lot of unnecessary stuff from the DB.

slowli (commented):
Mostly looks good. I feel that a better approach would be to refactor the executor so that it accepts `CallRequest` for calls, so that it can transform it internally, and `CallOverrides` is not required. This would certainly require more in-depth refactoring.

core/node/api_server/src/tx_sender/mod.rs
StanislavBreadless: I would leave it for future work

slowli (dismissed):

perekopskiy (dismissed):

perekopskiy (approved):

PR #2077, opened 2024-05-28
fix(en): chunk factory deps 

tomg10:


core/lib/snapshots_applier/src/lib.rs
```diff
@@ -588,18 +588,27 @@ impl<'a> SnapshotsApplier<'a> {
...
-            .await?;
+        // we cannot insert all factory deps because of field size limit triggered by UNNEST
+        // in underlying query, see `https://www.postgresql.org/docs/current/limits.html`
+        // there were around 100 thousand contracts on mainnet, where this issue first manifested
+        let chunk_size = 1000;
+        let chunks_count = factory_deps.factory_deps.len().div_ceil(chunk_size);
+        for chunk_id in 0..chunks_count {
+            let chunk_start = chunk_id * chunk_size;
+            let chunk_end = min(factory_deps.factory_deps.len(), (chunk_id + 1) * chunk_size);
+            let chunk = &factory_deps.factory_deps[chunk_start..chunk_end];
```
slowli, line 599: I think it's possible to use [standard slice chunking](https://doc.rust-lang.org/std/primitive.slice.html#method.chunks) here; would be shorter and more idiomatic.

core/lib/snapshots_applier/src/lib.rs
```diff
@@ -588,18 +588,27 @@ impl<'a> SnapshotsApplier<'a> {
...
-            .map(|dep| (hash_bytecode(&dep.bytecode.0), dep.bytecode.0))
-            .collect();
-        storage
-            .factory_deps_dal()
-            .insert_factory_deps(
-                self.applied_snapshot_status.l2_block_number,
-                &all_deps_hashmap,
-            )
-            .await?;
+        // we cannot insert all factory deps because of field size limit triggered by UNNEST
```
slowli, line 591: Nit: You can enclose "UNNEST" in backticks (i.e., format it as code) so that it won't trigger a spellcheck error.

core/lib/snapshots_applier/src/lib.rs
tomg10: nice!

core/lib/snapshots_applier/src/lib.rs
tomg10: thanks, will use this trick next time

slowli (approved):

PR #2080, opened 2024-05-29
fix(eth-watch): Do not track for stm, only for diamond proxy

StanislavBreadless:
## What ❔

We plan to generally have all upgrades executed via a direct call to diamond proxy. 

While in principle it is possible that an upgrade comes via a direct call through STM, this is a very rare circumstance and for now out of scope

## Checklist


perekopskiy (commented):
Looks good! Please update comments "Listens to operation events coming from the governance contract and saves new protocol upgrade proposals to the database.", "zkSync diamond proxy if pre-shared bridge; state transition manager if post shared bridge." near `pub struct GovernanceUpgradesEventProcessor`, IIUC they are outdated

StanislavBreadless:
@perekopskiy the first one is still correct as only ops from the governance are tracked

kelemeno (approved):

PR #2085, opened 2024-05-29
chore(eth-sender): refactor of eth-sender gas fees

tomg10:
I've added more detailed logs.. I've also generally split the calculations and metrics for fees for blob and non-blob transactions as they are radically different.


montekki (approved):

core/node/eth_sender/src/metrics.rs
```diff
@@ -86,6 +86,12 @@ pub(super) struct EthSenderMetrics {
     pub used_base_fee_per_gas: Histogram<u64>,
     #[metrics(buckets = FEE_BUCKETS)]
     pub used_priority_fee_per_gas: Histogram<u64>,
+    #[metrics(buckets = FEE_BUCKETS)]
+    pub blob_tx_used_blob_fee_per_gas: Histogram<u64>,
+    #[metrics(buckets = FEE_BUCKETS)]
+    pub blob_tx_used_base_fee_per_gas: Histogram<u64>,
+    #[metrics(buckets = FEE_BUCKETS)]
+    pub blob_tx_used_priority_fee_per_gas: Histogram<u64>,
```
RomanBrodetski, line 94: not very important, but still - I'd make blob/non-blob a label, not different metrics. This way it's easier to show on a grafana graph (`    pub tx_used_priority_fee_per_gas: Family<TxType, Histogram<Duration>>,`) where TxType - unless already defined somehwere - can be defined in the same file and only used for these metrics

tomg10: good idea! corrected

montekki (dismissed):

PR #2086, opened 2024-05-29
fix(zk_toolbox): update readme to make make more clear

zk-Lumi:
## What ❔

- Refactored 

## Why ❔

- Testing partners were getting confused

## Checklist


zk_toolbox/README.md
```diff
@@ -42,16 +42,18 @@ that connects all ZK chains, like the BridgeHub, the shared bridges, and state t
 To create a ZK Stack project, you must first create an ecosystem:
 
 ```bash
+# Skip this step if you are manually building from a local copy of the ZkSync repository (which is an existing ecosystem)
```
Deniallugo, line 45: It may not be the correct wording.

They could think that they don't need to create it anyway..
but they could create it somewhere outside zksync repository

zk_toolbox/README.md
```diff
@@ -42,16 +42,18 @@ that connects all ZK chains, like the BridgeHub, the shared bridges, and state t
...
 zk_inception ecosystem create
 ```
 
-If you chose to not start database & L1 containers after creating the ecosystem, you can later run
-`zk_inception containers`
+If you chose to not start database & L1 containers after creating the ecosystem, or are manually building from a local
```
Deniallugo, line 49: I don't agree with this statement

zk_toolbox/README.md
zk-Lumi: True that. Let me think about how to improve this.

zk_toolbox/README.md
zk-Lumi: Whole statement or when manually building from local copy?

How would one run the database and containers if they don't create the ecosystem and opt in at that step?

PR #2087, opened 2024-05-29
fix(metadata-calculator): protective reads sort

perekopskiy:
## What ❔

Makes protective reads sorted

## Why ❔

Fix bug

## Checklist

### DIFF
--- a/core/node/metadata_calculator/src/helpers.rs
+++ b/core/node/metadata_calculator/src/helpers.rs
@@ -634,15 +634,12 @@ impl L1BatchWithLogs {
                     StorageKey::new(AccountTreeId::new(tree_write.address), tree_write.key);
                 TreeInstruction::write(storage_key, tree_write.leaf_index, tree_write.value)
             });
-            let reads = protective_reads
-                .into_iter()
-                .sorted()
-                .map(TreeInstruction::Read);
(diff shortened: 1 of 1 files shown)


core/node/metadata_calculator/src/helpers.rs
```diff
@@ -634,7 +634,10 @@ impl L1BatchWithLogs {
                     StorageKey::new(AccountTreeId::new(tree_write.address), tree_write.key);
                 TreeInstruction::write(storage_key, tree_write.leaf_index, tree_write.value)
             });
-            let reads = protective_reads.into_iter().map(TreeInstruction::Read);
+            let reads = protective_reads
+                .into_iter()
+                .sorted()
+                .map(TreeInstruction::Read);
```
slowli, line 640: TBH, I think it would be safer to first collect protective reads and writes to a `Vec` and sort it then. It would require less non-local reasoning, and it looks like overhead would be reasonably small.

slowli (approved):
Ah, I got fooled by a comment that protective reads are already sorted 🙃

core/node/metadata_calculator/src/helpers.rs
perekopskiy: Yeah, agree. Then comment won't be needed

Artemka374 (approved):

PR #2110, opened 2024-05-31
feat(en): file based configs for en

Deniallugo:
## What ❔

Allow EN works with file-based config system. 
Also it brings new functionality based on config system to the main node, because it was lacking before, such as custom api namespaces

## Why ❔

Part of the refactoring to the new config system

## Checklist


core/bin/external_node/src/config/mod.rs
```diff
@@ -874,6 +1201,65 @@ impl ExternalNodeConfig<()> {
         })
     }
 
+    pub fn from_files(
+        general_config_path: PathBuf,
+        external_node_config_patch: PathBuf,
```
slowli, line 1206: Typo (?): "patch" -> "path".

core/bin/external_node/src/config/mod.rs
```diff
@@ -874,6 +1201,65 @@ impl ExternalNodeConfig<()> {
...
+            .context("failed decoding general YAML config")?;
+        let external_node_config =
+            read_yaml_repr::<proto::en::ExternalNode>(external_node_config_patch)
+                .context("failed decoding external node YAML config")?;
+        let secrets_config = read_yaml_repr::<proto::secrets::Secrets>(secrets_configs)
+            .context("failed decoding secrets YAML config")?;
+
+        let consensus = consensus
+            .map(read_yaml_repr::<proto::consensus::Config>)
+            .transpose()
```
slowli, line 1220: Why are these separate files required? I guess for secrets the reasoning is that we want to store secrets in a K8s secret (which is a moderately bogus reasoning because a node is not necessarily run on K8s), but what about other two? Why can't they be parsed from sections in the general configuration file?

core/bin/external_node/src/config/mod.rs
```diff
@@ -874,6 +1201,65 @@ impl ExternalNodeConfig<()> {
         })
     }
 
+    pub fn from_files(
+        general_config_path: PathBuf,
+        external_node_config_patch: PathBuf,
+        secrets_configs: PathBuf,
```
slowli, line 1207: Nit: Why not `secrets_config_path`?

core/bin/external_node/src/config/observability.rs
```diff
@@ -98,4 +104,40 @@ impl ObservabilityENConfig {
...
+                )
+            } else {
+                (None, None, LogFormat::default(), None)
+            };
+        let (prometheus_port, prometheus_pushgateway_url, prometheus_push_interval_ms) =
+            if let Some(api) = general_config.api_config.as_ref() {
+                (
+                    Some(api.prometheus.listener_port),
+                    Some(api.prometheus.pushgateway_url.clone()),
+                    api.prometheus.push_interval_ms.unwrap_or_default(),
```
slowli, line 128: Do I understand correctly that `prometheus` cannot be moved to `observability` config mentioned above (where it clearly belongs) because of compatibility concerns? That's a pity.

core/bin/external_node/src/main.rs
```diff
@@ -756,7 +768,26 @@ async fn main() -> anyhow::Result<()> {
     // Initial setup.
     let opt = Cli::parse();
 
-    let mut config = ExternalNodeConfig::new().context("Failed to load node configuration")?;
+    let mut config = if let Some(config_path) = opt.config_path.clone() {
+        let secrets_path = opt
+            .secrets_path
+            .clone()
+            .context("File based config can't be used partially please provide all paths")?;
```
slowli, line 775: IIUC, this relationship can be enforced on the `clap` level using [`requires`](https://docs.rs/clap/latest/clap/struct.Arg.html#method.requires) attribute.

core/lib/config/src/configs/en_config.rs
```diff
@@ -0,0 +1,40 @@
...
+}
+
+#[derive(Debug, Clone, PartialEq, Deserialize)]
+pub struct SnapshotRecovery {
+    pub enabled: bool,
+    pub postgres_max_concurrency: Option<NonZeroUsize>,
+    pub tree_chunk_size: Option<u64>,
+}
+
+/// Temporary config for initializing external node, will be completely replaced by consensus config later
```
slowli, line 23: This comment looks bogus. From the fields below, only first 3 look like something agreed upon by the network.

core/lib/config/src/configs/en_config.rs
```diff
@@ -0,0 +1,40 @@
...
+
+    // Main node configuration
+    pub main_node_url: SensitiveUrl,
+    pub main_node_rate_limit_rps: Option<NonZeroUsize>,
+
+    // En specific components
+    pub commitment_generator_max_parallelism: Option<NonZeroU32>,
+    pub tree_api_remote_url: Option<String>,
+    pub snapshot_recovery: Option<SnapshotRecovery>,
+    pub pruning: Option<Pruning>,
```
slowli, line 39: Since `SnapshotRecovery` and `Pruning` both have default values, it could make sense making these fields non-optional and use `#[serde(default)]` / corresponding glue for Protobuf. There's a similar problem with [`Secrets`](https://github.com/matter-labs/zksync-era/blob/49198f695a93d24a5e2d37a24b2c5e1b6c70b9c5/core/lib/config/src/configs/secrets.rs#L19) and `DatabaseSecrets`.

core/lib/config/src/configs/experimental.rs
```diff
@@ -20,6 +35,9 @@ impl Default for ExperimentalDBConfig {
             state_keeper_db_block_cache_capacity_mb:
                 Self::default_state_keeper_db_block_cache_capacity_mb(),
             state_keeper_db_max_open_files: None,
+            reads_persistence_enabled: Self::default_protective_reads_persistence_enabled(),
+            processing_delay_ms: 0,
```
slowli, line 39: `Self::default_merkle_tree_processing_delay_ms()` should be used here.

core/lib/config/src/configs/en_config.rs
```diff
@@ -0,0 +1,40 @@
...
+    // Genesis
+    pub l2_chain_id: L2ChainId,
+    pub l1_chain_id: L1ChainId,
+    pub l1_batch_commit_data_generator_mode: L1BatchCommitmentMode,
+
+    // Main node configuration
+    pub main_node_url: SensitiveUrl,
+    pub main_node_rate_limit_rps: Option<NonZeroUsize>,
+
+    // En specific components
```
slowli, line 35: Ironically, none of the following args are specific to EN.

- `commitment_generator_max_parallelism` can be readily used on the main node
- `tree_api_remote_url` is already used on the main node as `Web3JsonRpcConfig.tree_api_url` (so it can be removed, right?)
- Snapshot recovery and pruning are not sufficiently mature to use for the main node, but theoretically, they can be used for it.

core/lib/config/src/configs/en_config.rs
```diff
@@ -0,0 +1,40 @@
...
+    pub tree_chunk_size: Option<u64>,
+}
+
+/// Temporary config for initializing external node, will be completely replaced by consensus config later
+#[derive(Debug, Clone, PartialEq, Deserialize)]
+pub struct ENConfig {
+    // Genesis
+    pub l2_chain_id: L2ChainId,
+    pub l1_chain_id: L1ChainId,
+    pub l1_batch_commit_data_generator_mode: L1BatchCommitmentMode,
```
slowli, line 29: Just to understand better: There's intersection in these 3 params with `GenesisConfig`. Is there nothing that can be done to share these 3 params?

core/bin/external_node/src/config/mod.rs
```diff
@@ -413,6 +442,229 @@ pub(crate) struct OptionalENConfig {
...
+                let result: Result<Vec<_>, _> = a.iter().map(|a| serde_json::from_str(a)).collect();
+                result
+            })
+            .transpose()?;
+
+        Ok(OptionalENConfig {
+            filters_limit: load_optional_config_or_default!(
+                general_config.api_config,
+                web3_json_rpc.filters_limit,
+                default_filters_limit
```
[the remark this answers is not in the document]
Deniallugo: I was trying to mimic previous behaviour

core/bin/external_node/src/config/mod.rs
Deniallugo: Once consensus is fully ready we move the genesis part of external config there and the rest to the general. 
I think for this transition period we could tolerate it

core/bin/external_node/src/config/observability.rs
Deniallugo: while we are here, I think we can to move it for all configs

core/bin/external_node/src/config/mod.rs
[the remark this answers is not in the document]
slowli: Do defaults differ between the main node and EN? If not, IMO, it would make sense to refactor it in both places. Even if you don't do that, you can use existing getters in the config.

core/bin/external_node/src/config/mod.rs
slowli: But why wait to move it there? I think it should be possible and relatively easy to parse configs from a certain object path in a YAML document, and it would result in better UX at once.

core/lib/config/src/configs/en_config.rs
Deniallugo: So, yes they are part of the genesis config, I've mentioned it here, just because with consensus we will remove this file and we don't need to have full genesis

core/lib/config/src/configs/en_config.rs
Deniallugo: moved it to general or removed, thanks for the clarification

core/bin/external_node/src/config/mod.rs
Deniallugo: they have completely different semantic, don't want to mess them up

core/lib/config/src/configs/en_config.rs
Deniallugo: So, the first 3 will be moved to the genesis over l1 or consensus and the next 2 fields are gonna move to the nodes in consensus config.

I'd like to keep it as is for the sake of simplicity

core/bin/external_node/src/config/mod.rs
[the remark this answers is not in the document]
Deniallugo: I had one another thought and we have defaults only for EN, just to keep backward compatibility with envs and have them the same if not provided. 

We can make these fields mandatory for file based configuration and that's it

core/bin/external_node/src/config/observability.rs
slowli: Yep, this is what I meant. However, AFAIU, we'd hit Protobuf incompatibility in this case; or is simply moving a config not a concern at this point? (I.e., not caring that it could be specified in the old location somewhere.)

core/lib/config/src/configs/experimental.rs
```diff
@@ -32,4 +50,10 @@ impl ExperimentalDBConfig {
     pub fn state_keeper_db_block_cache_capacity(&self) -> usize {
         self.state_keeper_db_block_cache_capacity_mb * super::BYTES_IN_MEGABYTE
     }
+    const fn default_protective_reads_persistence_enabled() -> bool {
+        true
+    }
+    const fn default_merkle_tree_processing_delay_ms() -> u64 {
```
slowli, line 56: Micro-nit: missing an empty line between methods.

core/lib/config/src/configs/pruning.rs
```diff
@@ -0,0 +1,11 @@
+use std::num::NonZeroU64;
+
+use serde::Deserialize;
+
+#[derive(Debug, Clone, PartialEq, Deserialize)]
+pub struct PruningConfig {
+    pub enabled: bool,
+    pub chunk_size: Option<u32>,
+    pub removal_delay_sec: Option<NonZeroU64>,
+    pub data_retention_sec: Option<u64>,
```
slowli, line 10: Could you please move comments for these params from their previous declaration? Ditto for `SnapshotRecoveryConfig`.

core/lib/zksync_core_leftovers/src/temp_config_store/mod.rs
```diff
@@ -28,6 +32,12 @@ pub fn decode_yaml_repr<T: ProtoRepr>(yaml: &str) -> anyhow::Result<T::Type> {
     let this: T = zksync_protobuf::serde::deserialize_proto_with_options(d, false)?;
     this.read()
 }
+
+pub fn read_yaml_repr<T: ProtoRepr>(path_buf: PathBuf) -> anyhow::Result<T::Type> {
+    let yaml = std::fs::read_to_string(path_buf).context("failed decoding  YAML config")?;
```
slowli, line 37: Micro-nit: Extra spacing between "decoding" and "YAML".

core/lib/zksync_core_leftovers/src/temp_config_store/mod.rs
```diff
@@ -28,6 +32,12 @@ pub fn decode_yaml_repr<T: ProtoRepr>(yaml: &str) -> anyhow::Result<T::Type> {
...
+    decode_yaml_repr::<T>(&yaml)
+}
+
 //
 // TODO (QIT-22): This structure is going to be removed when components will be responsible for their own configs.
```
slowli, line 42: Off-topic (?): Is this comment still relevant? With the god config approach, I don't think we'll ever get to this point. Or will this TODO be naturally resolved when `core_leftovers` are removed?

core/lib/protobuf_config/src/proto/config/snapshot_recovery.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto3";
+
+package zksync.config.snapshot_recovery;
+
+message SnapshotRecovery {
+  optional bool enabled = 1;
+  optional uint64 postgres_max_concurrency = 2;
+  optional uint64 tree_chunk_size = 3;
+  optional uint32 l1_batch = 4;
+  optional uint64 tree_parallel_persistence_buffer = 5;
```
slowli, line 10: Bikeshedding: Abstractly, I'd prefer `postgres` and `tree` to be extracted to separate sub-config objects so that the config is more structured:

```yaml
snapshot_recovery:
  enabled: true
  l1_batch: 9112
  postgres:
    max_concurrency: 10
  tree:
    chunk_size: 200000
    parallel_persistence_buffer: 4
```

Not a strong preference though, so feel free to leave as-is if you disagree.

core/bin/external_node/src/main.rs
```diff
@@ -702,6 +702,17 @@ struct Cli {
     /// Comma-separated list of components to launch.
     #[arg(long, default_value = "all")]
     components: ComponentsToRun,
+    /// Path to the yaml config. If set, it will be used instead of env vars.
+    #[arg(long, requires = "secrets_path")]
+    config_path: Option<std::path::PathBuf>,
+    /// Path to the yaml with secrets. If set, it will be used instead of env vars.
+    #[arg(long, requires = "external_node_config_path")]
```
slowli, line 709: The `requires` attributes look confusing. Can't `config_path` specify that it requires all other paths via multiple `require` attrs?

core/bin/external_node/src/config/mod.rs
```diff
@@ -794,6 +1075,35 @@ impl ExperimentalENConfig {
     pub fn state_keeper_db_block_cache_capacity(&self) -> usize {
         self.state_keeper_db_block_cache_capacity_mb * BYTES_IN_MEGABYTE
     }
+
+    pub fn from_configs(general_config: &GeneralConfig) -> anyhow::Result<Self> {
```
slowli, line 1079: IMO, smaller config structs like `ExperimentalENConfig` can be removed immediately in favor of using `GeneralConfig` directly; not insisting though.

core/bin/external_node/src/config/mod.rs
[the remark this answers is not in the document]
slowli: > We can make these fields mandatory for file based configuration

This doesn't sound right; these fields should have reasonable defaults. This doesn't mean that types of the corresponding fields should be `Option`al; I'd argue that they shouldn't *because* of the reasonable defaults.

> we have defaults only for EN

That's not true. E.g., for `filters_limit` we have a default value 10_000 exposed via `Web3JsonRpcConfig::filters_limit()` getter. The fact that `Web3JsonRpcConfig.filters_limit` is optional is IMO a flaw (see above).

core/bin/external_node/src/config/mod.rs
```diff
@@ -874,6 +1205,64 @@ impl ExternalNodeConfig<()> {
         })
     }
 
+    pub fn from_files(
```
slowli, line 1208: So here's how embedding configs in a single file could work:

```rust
// Could make sense to eventually add to `zksync_protobuf::serde`
#[derive(Debug)]
pub struct SerdeRepr<T: ProtoRepr>(pub T::Type);

impl<'de, T: ProtoRepr> Deserialize<'de> for SerdeRepr<T> {
    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
        let repr: T = deserialize_proto(deserializer)?;
        repr.read().map(Self).map_err(D::Error::custom)
    }
}

#[derive(Debug, Deserialize)]
struct WrappingConfig {
    external_node: SerdeRepr<proto::en::ExternalNode>,
    consensus: Option<SerdeRepr<proto::consensus::Config>>,
    #[serde(flatten)]
    general: SerdeRepr<proto::general::GeneralConfig>,
}

// usage
let yaml = std::fs::read_to_string(&path).with_context(|| path.display().to_string())?;
let config: WrappingConfig = serde_yaml::from_str(&yaml)?;
```

core/bin/external_node/src/config/mod.rs
Deniallugo: I agree, i don't want to do this in this pr, just for making less mistakes

core/bin/external_node/src/config/mod.rs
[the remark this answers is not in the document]
Deniallugo: I'd left it for the next iteration, where I'd probably remove these smaller structs in favor of properly compatible ones. 
 
We will have no `OptionalEnConfig` `RequiedEnConfig` and so on , rather general and ExternalNode

core/bin/external_node/src/config/mod.rs
Deniallugo: I generally like it, but we don't need to implement it.

core/bin/external_node/src/config/observability.rs
Deniallugo: it makes it slightly harder to work with protobuf configs. 
Probably we can address it in the next step

Deniallugo: I've found that i have prometheus filed in non api.. 
but seems i have to revisit all Prometheus configs

core/lib/config/src/configs/en_config.rs
Deniallugo: it makes it slightly harder to work with protobuf configs.
Probably we can address it in the next step

core/lib/config/src/configs/snapshot_recovery.rs
```diff
@@ -0,0 +1,41 @@
+use std::num::NonZeroUsize;
+
+use serde::Deserialize;
+use zksync_basic_types::L1BatchNumber;
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Default)]
+pub struct Tree {
```
slowli, line 7: Bikeshedding: IMO, it would make sense to name these structs `TreeRecoveryConfig` / `PostgresRecoveryConfig`, so that they are less ambiguous if used in isolation.

core/lib/protobuf_config/src/proto/config/snapshot_recovery.proto
```diff
@@ -0,0 +1,20 @@
...
+  optional uint64 max_concurrency = 1;
+}
+
+message SnapshotRecovery {
+  optional bool enabled = 1;
+  optional Postgres postgres = 2;
+  optional Tree tree = 3;
+  optional uint32 l1_batch = 4;
+}
+
```
slowli, line 20: Micro-nit: unnecessary newline.

core/lib/config/src/configs/api.rs
```diff
@@ -213,6 +213,13 @@ pub struct Web3JsonRpcConfig {
     /// (additionally to natively bridged tokens).
     #[serde(default)]
     pub whitelisted_tokens_for_aa: Vec<Address>,
+    /// Enabled JSON RPC API namespaces. If not set, all namespaces will be available
+    #[serde(default)]
+    pub api_namespaces: Option<Vec<String>>,
```
slowli, line 218: IMO, we can afford to be more typesafe here, i.e. declare a namespace enum (or move `Namespace` from the API server crate) and use `Option<HashSet<Namespace>>`. Or do you think we should allow for custom namespaces be defined by the node?

core/bin/zksync_server/src/node_builder.rs
```diff
@@ -292,7 +292,11 @@ impl MainNodeBuilder {
         let circuit_breaker_config = try_load_config!(self.configs.circuit_breaker_config);
         let with_debug_namespace = state_keeper_config.save_call_traces;
 
-        let mut namespaces = Namespace::DEFAULT.to_vec();
+        let mut namespaces = if let Some(namespaces) = &rpc_config.api_namespaces {
+            serde_json::from_str(&namespaces.join(","))?
```
slowli, line 296: Will this work? If you want to parse data from a JSON array, you need to wrap an array in `[]` and quote strings (I think `impl Debug for str` does this, but not 100% sure it's always JSON-compatible – it probably is at least for ASCII strings). This looks quite awkward in any case; isn't it possible to e.g. move `Namespace` to `zksync_config` and use it in the config declaration directly?

core/lib/config/src/configs/pruning.rs
```diff
@@ -0,0 +1,20 @@
+use std::num::NonZeroU64;
+
+use serde::Deserialize;
+
+#[derive(Debug, Clone, PartialEq, Deserialize)]
+pub struct PruningConfig {
+    pub enabled: bool,
+    /// Chunk size for multi-get operations. Can speed up loading data for the Merkle tree on some environments,
+    /// but the effects vary wildly depending on the setup (e.g., the filesystem used).
```
slowli, line 9: That's a wrong doc comment; this one belongs to the Merkle tree configuration.

core/lib/config/src/configs/api.rs
Deniallugo: I was trying to do it. But it doesn't seem like a proper solution, it will be typed safety a bit later. I'd prefer to keep it string here and convert later

core/bin/zksync_server/src/node_builder.rs
Deniallugo: Yes it works.
I don't think we have to move namespaces to the config folder.

It doesn't belong there

slowli: > Yes it works.

I've run it locally, and it really doesn't.

```text
$ API_WEB3_JSON_RPC_API_NAMESPACES=eth,web3,net,pubsub,zks,en,debug \
  zk server --components api --use-node-framework
2024-06-18T13:53:06.216021Z  INFO zksync_server: No sentry URL was provided
2024-06-18T13:53:06.263693Z  INFO zksync_db_connection::connection_pool: Created DB pool with parameters ConnectionPoolBuilder { database_url: "postgres://***:***@localhost/zksync_local", max_size: 1, acquire_timeout: 30s, statement_timeout: None, db: "zksync_dal::Core" }
Error: expected value at line 1 column 1

Stack backtrace:
   0: std::backtrace::Backtrace::create
   1: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from
   2: zksync_server::node_builder::MainNodeBuilder::add_ws_web3_api_layer
   3: zksync_server::node_builder::MainNodeBuilder::build
   4: std::sys_common::backtrace::__rust_begin_short_backtrace
   5: core::ops::function::FnOnce::call_once{{vtable.shim}}
   6: std::sys::pal::unix::thread::Thread::new::thread_start
   7: __pthread_joiner_wake
Error: Child process exited with code 1
```

(These are the default namespaces from [EN config](https://github.com/matter-labs/zksync-era/blob/26f2010ea2edd1cb79d80852c626051afc473c48/etc/env/configs/ext-node.toml#L27).) What am I doing wrong? Am I supposed to ~SQL~ JSON-inject array brackets and quotes into names like this: `API_WEB3_JSON_RPC_API_NAMESPACES='["eth","zks"]'` (that's namespaces `["eth"` and `"zks"]`)? This looks weird and semantically invalid.

Deniallugo: Decided to switch to strum. Seems it's much better approach

core/bin/external_node/src/main.rs
```diff
@@ -704,6 +704,21 @@ struct Cli {
...
+    )]
+    config_path: Option<std::path::PathBuf>,
+    /// Path to the yaml with secrets. If set, it will be used instead of env vars.
+    #[arg(long, requires = "config_path", requires = "external_node_config_path")]
+    secrets_path: Option<std::path::PathBuf>,
+    /// Path to the yaml with genesis. If set, it will be used instead of env vars.
+    #[arg(long, requires = "config_path", requires = "secrets_path")]
+    external_node_config_path: Option<std::path::PathBuf>,
+    /// Path to the yaml with consensus.
+    consensus_path: Option<std::path::PathBuf>,
```
slowli, line 721: Just to discuss: Would it make sense to specify a path to a directory that would contain all these config files with predefined names? This would allow to change the set of these files in the future w/o changing command-line args here. Also, this approach (configuration by convention) is commonly used in various frameworks.

slowli (commented):
Mostly looks fine.

core/node/api_server/src/web3/mod.rs
```diff
@@ -86,8 +86,9 @@ enum ApiTransport {
     Http(SocketAddr),
 }
 
-#[derive(Debug, Deserialize, Clone, PartialEq)]
+#[derive(Debug, Deserialize, Clone, PartialEq, strum_macros::EnumString)]
```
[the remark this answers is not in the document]
Deniallugo: it's already in our dependencies, so i decided that it's ok

core/bin/external_node/src/main.rs
Deniallugo: I was thinking about it and probably it's not that bad, as i thought before, i can touch it later

core/lib/config/src/configs/api.rs
Deniallugo: Agree, but i really don't want to move namespaces to the places where are they are not belong it's part of the api

core/lib/config/src/configs/snapshot_recovery.rs
```diff
@@ -0,0 +1,41 @@
...
+    /// Reasonable values are order of 100,000 (meaning an iteration takes several seconds).
+    ///
+    /// **Important.** This value cannot be changed in the middle of tree recovery (i.e., if a node is stopped in the middle
+    /// of recovery and then restarted with a different config).
+    pub chunk_size: Option<u64>,
+    /// Buffer capacity for parallel persistence operations. Should be reasonably small since larger buffer means more RAM usage;
+    /// buffer elements are persisted tree chunks. OTOH, small buffer can lead to persistence parallelization being inefficient.
+    ///
+    /// If not set, parallel persistence will be disabled.
+    pub parallel_persistence_buffer: Option<NonZeroUsize>,
```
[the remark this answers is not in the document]
slowli: Got it. It's still an experimental parameter though, and if we e.g. decide to re-parameterize it, Protobuf-based backward compatibility checks can get in the way. (To be fair, just removing the param would work fine, although it would produce some visual noise in the form of a reserved Protobuf field.)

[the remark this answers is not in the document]
Deniallugo: What is your suggestion here? 
We can't just mark it as experimental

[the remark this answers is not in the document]
Deniallugo: also, i think as well as we do with database, we can cleanup all reserved after the upgrade

core/lib/config/src/configs/api.rs
slowli: Right. As discussed previously, we could decentralize configs (so that API server defines its config in its crate), but of course this would require reworking config approach entirely.

core/lib/config/src/configs/snapshot_recovery.rs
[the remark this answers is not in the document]
slowli: One option would be to define it in a separate struct (e.g., `ExperimentalSnapshotRecoveryConfig` with a `tree_ parallel_persistence_buffer` field) in the `experimental` module, attach it as a field to `SnapshotRecoveryConfig`, and correspondingly define it in the `zksync.config.experimental` package in Protobuf. This package [is excluded](https://github.com/matter-labs/zksync-era/blob/dd2c941e94b5fde9064d9ff4394a00a4b0707fbf/.github/workflows/protobuf.yaml#L77) from backward compatibility checks.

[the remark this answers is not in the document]
Deniallugo: It's a bit painful, but i've done it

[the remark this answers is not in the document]
slowli: I feel it's painful because you've decided to make it painful 🙃 IMO, it'd be much simpler if you had the same config structure in Protobuf and here, as I've suggested. Well, it'd be even simpler if we didn't have Protobuf config altogether, but we've been over this.

Cargo.toml
```diff
@@ -159,6 +159,7 @@ sqlx = "0.7.3"
 static_assertions = "1.1"
 structopt = "0.3.20"
 strum = "0.24"
+strum_macros = "0.24"
```
slowli, line 162: I don't think importing this crate directly is necessary; other `strum` uses in the workspace import it via

```toml
strum = { workspace = true, features = ["derive"] }
```

core/lib/config/src/configs/snapshot_recovery.rs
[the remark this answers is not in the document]
Deniallugo: i think, having the same structure will make everything even harder. 
Let's stick to this approach..

core/lib/zksync_core_leftovers/src/temp_config_store/mod.rs
```diff
@@ -23,6 +27,11 @@ pub fn decode_yaml_repr<T: ProtoRepr>(yaml: &str) -> anyhow::Result<T::Type> {
     this.read()
 }
 
+pub fn read_yaml_repr<T: ProtoRepr>(path_buf: PathBuf) -> anyhow::Result<T::Type> {
+    let yaml = std::fs::read_to_string(path_buf).context("failed decoding YAML config")?;
```
perekopskiy, line 31: nit: reading from file fails here, not decoding

PR #2119, opened 2024-05-31
fix(prover_cli): Fix delete command

ColoCarletti:
Changes: 

- Get database url from the `prover_cli` config.
 - Remove the deletion of archived jobs.


Oppen (approved):

ilitteri (changes requested):
Looking good. We also need to delete the `delete_gpu_prover_queue_fri_archive` method from `FriGpuProverQueueDal`

ilitteri (approved):
LGTM. There are some things failing in the CI

PR #2124, opened 2024-06-03
refactor(node-framework): use owned type to identify `Task`s

itegulov:
## What ❔

This PR refactors `Task`-related traits to use an owned type `TaskId` instead of `&'static str`.

## Why ❔

There are use cases where a task's id needs to be generated dynamically (e.g. broad trait implementations) which this PR accommodates for.

## Checklist

### DIFF
--- a/core/node/node_framework/src/task.rs
+++ b/core/node/node_framework/src/task.rs
@@ -28,11 +28,24 @@
 //! - A task that must be started as soon as possible, e.g. healthcheck server.
 //! - A task that may be a driving force for some precondition to be met.
 
-use std::sync::Arc;
+use std::{
+    fmt::{Display, Formatter},
+    sync::Arc,
+};
 
 use tokio::sync::Barrier;
 
 use crate::service::StopReceiver;
+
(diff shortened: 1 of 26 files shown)


core/node/node_framework/src/task.rs
```diff
@@ -28,20 +28,33 @@
...
+    sync::Arc,
+};
 
 use tokio::sync::Barrier;
 
 use crate::service::StopReceiver;
 
+/// A unique human-readable identifier of a task.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct TaskId(pub String);
```
popzxc, line 42: I think you may:
- Add `impl From<&str> for TaskId` (will simplify all `id` impls).
- Add `impl Deref for TaskId` with `Target = str` (will simplify usage I guess)
- Add a constructor and make internal representation private to make it a bit more future-proof.

itegulov: Great points, done!

popzxc (approved):

PR #2126, opened 2024-06-03
feat!: updated boojum and nightly rust compiler

mm-zk:
## What ❔

* Updated boojum version, which allows us to update nightly rust compiler.

## Why ❔

* Our rust nightly is quite old (around 1 year)


tomg10:
I believe you should also bump nightly version in zk-environment

popzxc (commented):
It comes with a bunch of clippy issues (as clippy is updated too) -- unfortunately, it has to be fixed as well.

popzxc:
Huh, I guess generic Poseidon implementation does have an impact on performance 👀

itegulov (approved):
I think this should also allow us to delete all false positive occurrences of `#[allow(clippy::needless_pass_by_ref_mut)]` in the code. Not a big deal if you don't want to do it in this PR though.

core/lib/dal/src/events_dal.rs
```diff
@@ -84,7 +84,7 @@ impl EventsDal<'_, '_> {
                 write_str!(
                     &mut buffer,
                     r"\\x{topic0:x}|\\x{topic1:x}|\\x{topic2:x}|\\x{topic3:x}|",
-                    topic0 = EventTopic(event.indexed_topics.get(0)),
+                    topic0 = EventTopic(event.indexed_topics.first()),
                     topic1 = EventTopic(event.indexed_topics.get(1)),
                     topic2 = EventTopic(event.indexed_topics.get(2)),
                     topic3 = EventTopic(event.indexed_topics.get(3))
```
popzxc, line 90: Damn clippy, you're ruining all the aesthetics!

popzxc (dismissed):

core/lib/dal/src/events_dal.rs
mm-zk: yeah...

.dockerignore
```diff
@@ -46,3 +46,4 @@ contracts/.git
 !etc/env/dev.toml
 !etc/env/consensus_secrets.yaml
 !etc/env/consensus_config.yaml
+!rust-toolchain
```
hatemosphere, line 49: please re-add newline

hatemosphere (changes requested):

.dockerignore
mm-zk: done

hatemosphere (approved):

popzxc (approved):

PR #2137, opened 2024-06-04
feat(en): Allow recovery from specific snapshot

slowli:
## What ❔

Allows recovering a node from a specific snapshot specified at the start of recovery.

## Why ❔

Useful at least for testing recovery and pruning end-to-end on the testnet. There, L1 batches are produced very slowly, so it makes sense to recover from an earlier snapshot in order to meaningfully test pruning.

## Checklist

### DIFF
--- a/core/bin/external_node/src/init.rs
+++ b/core/bin/external_node/src/init.rs
@@ -12,7 +12,13 @@ use zksync_snapshots_applier::{SnapshotsApplierConfig, SnapshotsApplierTask};
 use zksync_types::{L1BatchNumber, L2ChainId};
 use zksync_web3_decl::client::{DynClient, L2};
 
-use crate::config::SnapshotsRecoveryConfig;
+use crate::config::snapshot_recovery_object_store_config;
+
+#[derive(Debug)]
+pub(crate) struct SnapshotRecoveryConfig {
+    /// If not specified, the latest snapshot will be used.
+    pub snapshot_l1_batch: Option<L1BatchNumber>,
+}
 
 #[derive(Debug)]
 enum InitDecision {
@@ -27,7 +33,7 @@ pub(crate) async fn ensure_storage_initialized(
     main_node_client: Box<DynClient<L2>>,
     app_health: &AppHealthCheck,
     l2_chain_id: L2ChainId,
-    consider_snapshot_recovery: bool,
+    recovery_config: Option<SnapshotRecoveryConfig>,
 ) -> anyhow::Result<()> {
     let mut storage = pool.connection_tagged("en").await?;
     let genesis_l1_batch = storage
@@ -57,7 +63,7 @@ pub(crate) async fn ensure_storage_initialized(
         }
         (None, None) => {
             tracing::info!("Node has neither genesis L1 batch, nor snapshot recovery info");
-            if consider_snapshot_recovery {
+            if recovery_config.is_some() {
                 InitDecision::SnapshotRecovery
             } else {
                 InitDecision::Genesis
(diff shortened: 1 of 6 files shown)


core/lib/snapshots_applier/src/lib.rs
```diff
@@ -246,6 +261,12 @@ impl SnapshotsApplierTask {
         }
     }
 
+    /// Specifies the L1 batch to recover from. This setting is ignored if recovery is complete, but is checked
+    /// if recovery is in progress (so if a node started recovering from another snapshot, it will error).
```
tomg10, line 265: I'd expect such setting to only impact the recovery initialization, the same way the snapshots recovery behaves if the newest snapshot changes

core/bin/external_node/src/init.rs
```diff
@@ -78,25 +84,31 @@ pub(crate) async fn ensure_storage_initialized(
...
 
             let config = SnapshotsApplierConfig::default();
-            let snapshots_applier_task = SnapshotsApplierTask::new(
+            let mut snapshots_applier_task = SnapshotsApplierTask::new(
                 config,
                 pool,
                 Box::new(main_node_client.for_component("snapshot_recovery")),
                 blob_store,
             );
+            if let Some(snapshot_l1_batch) = recovery_config.snapshot_l1_batch {
```
tomg10, line 105: "too old or non-existent" is not very helpful, we keep snapshots in gcs for 28 days, let's put info that you shouldn't use a snapshot older than 21 days

core/bin/external_node/src/init.rs
```diff
@@ -12,7 +12,13 @@ use zksync_snapshots_applier::{SnapshotsApplierConfig, SnapshotsApplierTask};
 use zksync_types::{L1BatchNumber, L2ChainId};
 use zksync_web3_decl::client::{DynClient, L2};
 
-use crate::config::SnapshotsRecoveryConfig;
+use crate::config::snapshot_recovery_object_store_config;
+
+#[derive(Debug)]
+pub(crate) struct SnapshotRecoveryConfig {
+    /// If not specified, the latest snapshot will be used.
+    pub snapshot_l1_batch: Option<L1BatchNumber>,
```
tomg10, line 20: I'd put "override" into the name to indicate that this is not a normal mode of operation.

core/lib/snapshots_applier/src/tests/mod.rs
```diff
@@ -142,6 +143,163 @@ async fn snapshots_creator_can_successfully_recover_db(
...
+
+    let all_storage_logs = storage
+        .storage_logs_dal()
+        .dump_all_storage_logs_for_tests()
+        .await;
+    assert_eq!(all_storage_logs.len(), storage_logs.len());
+}
+
+#[tokio::test]
+async fn applier_errors_if_snapshot_l1_batch_differs_from_previous_one() {
```
tomg10, line 272: This imo should not be an error as mentioned in previous comments. Lets make this as simple as possible. This is a feature to use mostly for us for some debugging, if partners ever need to use it, they will rather have our guidance and not use two versions

tomg10 (changes requested):

core/bin/external_node/src/init.rs
slowli: That's an internal policy that can be changed at any point, and we'll probably forget to update the log here if it does.

core/lib/snapshots_applier/src/lib.rs
```diff
@@ -373,16 +409,34 @@ impl SnapshotRecoveryStrategy {
 
     async fn create_fresh_recovery_status(
         main_node_client: &dyn SnapshotsApplierMainNodeClient,
+        snapshot_l1_batch: Option<L1BatchNumber>,
     ) -> Result<SnapshotRecoveryStatus, SnapshotsApplierError> {
-        let snapshot_response = main_node_client.fetch_newest_snapshot().await?;
+        let (l1_batch_number, is_latest) = match snapshot_l1_batch {
```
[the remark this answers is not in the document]
slowli: `is_latest` is also used for error context, which I consider more important, because the context of the error IMO *is* substantially different depending on whether the user has specified a snapshot (if they did, it's probably user's fault that a snapshot is not found; if they didn't, it's probably a bug on the main node).

core/lib/snapshots_applier/src/tests/mod.rs
```diff
@@ -142,6 +143,163 @@ async fn snapshots_creator_can_successfully_recover_db(
...
+
+    let err = task.run().await.unwrap_err();
+    assert!(
+        format!("{err:#}").contains("not present on main node"),
+        "{err:#}"
+    );
+}
+
+#[tokio::test]
+async fn snapshot_applier_recovers_after_stopping() {
```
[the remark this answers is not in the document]
tomg10: weird; it's good that you've spotted this

tomg10 (approved):

PR #2140, opened 2024-06-04
fix(en): Fix transient error detection in consistency checker

slowli:
## What ❔

Considers a subset of contract call errors as transient.

## Why ❔

Currently, consistency checker considers all contract call errors fatal, which leads to EN terminating when it shouldn't.

## Checklist

### DIFF
--- a/core/node/consistency_checker/src/lib.rs
+++ b/core/node/consistency_checker/src/lib.rs
@@ -42,10 +42,12 @@ enum CheckError {
 
 impl CheckError {
     fn is_transient(&self) -> bool {
-        matches!(
-            self,
-            Self::Web3(err) if err.is_transient()
-        )
+        match self {
+            Self::Web3(err) | Self::ContractCall(ContractCallError::EthereumGateway(err)) => {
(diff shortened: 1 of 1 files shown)


core/node/consistency_checker/src/lib.rs
```diff
@@ -532,7 +534,10 @@ impl ConsistencyChecker {
 
         while let Err(err) = self.sanity_check_diamond_proxy_addr().await {
             if err.is_transient() {
-                tracing::warn!("Transient error checking diamond proxy contract; will retry after a delay: {err}");
+                tracing::warn!(
+                    "Transient error checking diamond proxy contract; will retry after a delay: {:#}",
+                    anyhow::Error::from(err)
+                );
```
slowli, line 540: There is certainly a better way to output the chain of error sources (e.g., it doesn't need to consume the error), but I'm not sure where to place the corresponding helper; converting an error to `anyhow::Error` looks like an OK-ish alternative.

popzxc (approved):

PR #2143, opened 2024-06-05
refactor: Added intermediate representations for rlp structs

pompon0:
Added a bunch of intermediate representations for rlp structs. I need these to be able to verify L1 batches for https://linear.app/matterlabs/issue/BFT-471/implement-stateless-l1-batch-verification


core/lib/types/src/abi.rs
```diff
@@ -0,0 +1,343 @@
...
+            Token::FixedArray(self.reserved.iter().map(|x| Token::Uint(*x)).collect()),
+            Token::Bytes(self.data.clone()),
+            Token::Bytes(self.signature.clone()),
+            Token::Array(self.factory_deps.iter().map(|x| Token::Uint(*x)).collect()),
+            Token::Bytes(self.paymaster_input.clone()),
+            Token::Bytes(self.reserved_dynamic.clone()),
+        ])
+    }
+
+    pub fn hash(&self) -> H256 {
```
StanislavBreadless, line 118: just in case, this hash is correct only for L1->L2 transactions. The rest have special functions to calculate the correct hash of the transaction, so better to at least add a comment about it

core/lib/types/src/l1/mod.rs
```diff
@@ -265,159 +264,116 @@ impl L1Tx {
...
-            gas_per_pubdata_limit,
+            gas_limit: req.transaction.gas_limit,
+            max_fee_per_gas: req.transaction.max_fee_per_gas,
+            gas_per_pubdata_limit: req.transaction.gas_per_pubdata_byte_limit,
             op_processing_type: OpProcessingType::Common,
             priority_queue_type: PriorityQueueType::Deque,
             // DEPRECATED.
-            // TODO (PLA-962): start setting it to 0 for all new transactions.
-            eth_block,
+            eth_block: 0,
```
StanislavBreadless, line 346: i guess eth_watch still relies on these numbers being non-zero. if the output of this method is not intended for eth watcher, then better to add a comment about it

core/lib/types/src/l1/mod.rs
```diff
@@ -265,159 +264,116 @@ impl L1Tx {
...
+            contract_address: u256_to_account_address(&req.transaction.to),
+            calldata: req.transaction.data,
+            factory_deps: Some(req.factory_deps),
+            value: req.transaction.value,
         };
         Ok(Self {
             common_data,
             execute,
-            received_timestamp_ms: unix_timestamp_ms(),
+            received_timestamp_ms: 0,
```
StanislavBreadless, line 358: probably better to indeed a comment that describes that not all the data will be correct in this method

core/lib/types/src/transaction_request.rs
```diff
@@ -691,8 +671,21 @@ impl TransactionRequest {
...
+            .unwrap();
+        Ok((tx, hash))
+    }
 
+    pub fn from_bytes(
+        bytes: &[u8],
+        chain_id: L2ChainId,
+    ) -> Result<(Self, H256), SerializationTransactionError> {
+        let (tx, hash) = Self::from_bytes_unverified(bytes)?;
+        if tx.chain_id.is_some() && tx.chain_id != Some(chain_id.as_u64()) {
```
StanislavBreadless, line 686: Note, that for some types having chain id is required. It is actually required for all types except for the legacy one

core/lib/l1_contract_interface/src/i_executor/structures/stored_batch_info.rs
```diff
@@ -8,6 +12,43 @@ use crate::Tokenizable;
 #[derive(Debug)]
 pub struct StoredBatchInfo<'a>(pub &'a L1BatchWithMetadata);
 
+/// Compact representation the the `StoredBatchInfo` from `IExecutor.sol`.
```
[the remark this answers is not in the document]
pompon0: done, replaced StorageBatchInfo with StorageBatchInfoCompact (and renamed)

core/lib/types/src/l1/mod.rs
pompon0: Yes, that's why it is being set in the TryFrom<Log> below.

core/lib/types/src/l1/mod.rs
pompon0: Done, added a comment.

core/lib/types/src/transaction_request.rs
pompon0: made chain_id required for every transaction type except for legacy.

core/lib/types/src/abi.rs
```diff
@@ -0,0 +1,368 @@
+use anyhow::Context as _;
```
popzxc, line 1: I understand that it's probably out of scope, but would it be hard to move this file to the `l1_contract_interface` crate (as we've moved some L1 ABI stuff there already, and generally it feels more appropriate)?
From a quick glance, it doesn't look too hard, but may require some wrestling with the orphan rule (constructors instead of traits, I guess).

core/lib/types/Cargo.toml
```diff
@@ -33,6 +33,7 @@ num_enum.workspace = true
 hex.workspace = true
 prost.workspace = true
 itertools.workspace = true
+tracing.workspace = true
```
popzxc, line 36: It doesn't seem to be used anywhere.

popzxc (commented):
Mostly looks good.

core/lib/types/Cargo.toml
pompon0: removed

core/node/eth_watch/Cargo.toml
```diff
@@ -17,6 +17,7 @@ zksync_contracts.workspace = true
 zksync_system_constants.workspace = true
 zksync_eth_client.workspace = true
 zksync_shared_metrics.workspace = true
+zksync_concurrency.workspace = true
```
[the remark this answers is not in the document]
pompon0: done

core/lib/types/src/abi.rs
pompon0: for example conversion from Log (zksync_basic_types) to L1Tx (zksync_types) goes through `abi::NewPriorityRequest`. zksync_l1_contract_interface depends on zksync_types, so moving abi.rs there is a bit problematic.

popzxc: I see, no big deal then. It looks like the stuff from `zksync_types` the crate depends is either exported from `basic_types` or belongs to `l1_contract_interface` domain, but de-entangling this mess indeed may take a while.

popzxc (approved):

core/lib/types/src/abi.rs
pompon0: yup

PR #2146, opened 2024-06-05
feat(api): Rework zks_getProtocolVersion

perekopskiy:
## What ❔

- renames fields in method's response
- removes `verification_keys_hashes` completely

## Why ❔

- more meaningfully field names and camelCase for consistency
- `verification_keys_hashes` doesn't make sense in the context of minor protocol version

## Checklist


core/lib/types/src/api/mod.rs
```diff
@@ -642,15 +642,28 @@ impl From<Call> for DebugCall {
 
 #[derive(Default, Serialize, Deserialize, Clone, Debug)]
 pub struct ProtocolVersion {
```
popzxc, line 644: Probably it makes sense to add a comment explaining that this structure is in "migration" phase, and what should be changed and when.
Otherwise we risk finding ourselves asking "Why was it done this way?" X month later (not for the first time).

core/lib/types/src/api/mod.rs
```diff
@@ -751,3 +764,36 @@ pub struct ApiStorageLog {
...
     pub written_value: U256,
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[allow(deprecated)]
+    #[test]
+    fn check_protocol_version_type_compatibility() {
```
popzxc, line 774: Also worth a comment on when this test will no longer be needed.

core/node/node_sync/src/external_io.rs
```diff
@@ -337,35 +337,57 @@ impl StateKeeperIO for ExternalIO {
...
+                })
+                .context("Missing bootloader code hash")?;
+            let default_account_code_hash = protocol_version
+                .default_account_code_hash
+                .or_else(|| {
+                    protocol_version
+                        .base_system_contracts
+                        .map(|hashes| hashes.default_aa)
+                })
+                .context("Missing default account code hash")?;
```
popzxc, line 361: To avoid `allow(deprecated)` maybe it makes sense to add getters on `ProtocolVersion` that would encapsulate this logic?

mm-zk (approved):

popzxc (approved):

PR #2153, opened 2024-06-05
feat(object-store): Allow mirroring object store objects locally

slowli:
## What ❔

Implements an optional mirroring layer for object store that persists objects fetched from an external store (e.g., a GCS one) to a configurable local directory.

## Why ❔

Allows to speed up repeated snapshot recovery w/o sacrificing generality (i.e. switching to a file-backed store and downloading all snapshot data manually).

## Checklist


danielCalbonero12345 (approved):

core/lib/object_store/src/factory.rs
```diff
@@ -54,25 +55,22 @@ impl ObjectStoreFactory {
...
-                );
                 let store = StoreWithRetries::try_new(config.max_retries, || {
                     GoogleCloudStore::new(
                         GoogleCloudStoreAuthMode::Authenticated,
                         bucket_base_url.clone(),
                     )
                 })
                 .await?;
-                Ok(Arc::new(store))
+                Self::wrap_cache(store, config.cache_path.as_ref()).await
```
popzxc, line 68: AFAIU we rely on the GCS TTL and do not perform any kind of garbage collection ourselves. Couldn't it cause all the blobs fetched by main node to be (uselessly) persisted, if this configuration option is provided?

slowli: That's a good point! It would still be a problem even if garbage collection was performed because it's not like the cached store somehow has knowledge that an object was removed in some other workflow. The simplest approach would be to say that it's safe to remove cache entries at any point (it is) and it's the node operator responsibility to do this; would it be fine? Alternatively, it could be possible to come up with some bounded-capacity cache and the corresponding removal policy, but I feel that is overcomplicating things.

popzxc: > it's the node operator responsibility to do this; would it be fine?

I think that it would've been fine for Era, but with ZK Stack and RaaS providers in mind, it feels like we should minimize the amount of manual interventions we expect from any chain operator.

> but I feel that is overcomplicating things

Another approach would be to not have a cache. Do you have an estimate on how often this cache would be used and how much speedup (assuming that for GCS operator has 1gbps connection, and for local FS it uses NAS within k8s cluster -- e.g. a "default" setup we use now) will it provide?

slowli: > Another approach would be to not have a cache. 

Well, that's certainly an option if the operator has high-speed Internet connection, and cache is opt-in for this reason. As I've written in the Linear task thread, a motivating use case for it was me testing pruning / snapshot recovery end-to-end locally (I don't have 1 Gbps connection at home 🙃). Do you think this use case is obscure enough to not waste time on it? (A little bit of a stretch, but I think package caching Cargo / npm / etc. do is conceptually similar.)

popzxc: Ah, sorry, I didn't read the task description -- only the PR one.
For that use case _I think_ the implementation that implies the node operator responsibility makes sense, but I'm still a bit afraid that this feature may be misused (roughly _cache is never bad, lemme get some caching_).


What if we will ship this feature as-is, but slightly change semantics to make it look, ugh, less universally useful? For example, not `cache_path`, but `local_mirror_path`? And then in docs we can elaborate that "If provided, fetched artifacts will be stored in provided location to avoid fetching the same artifact twice. Mostly useful for situations like local development or testing".

popzxc (approved):

PR #2158, opened 2024-06-06
fix(vm-runner): add config value for the first processed batch

itegulov:
## What ❔

Adds a new config value that regulate which batch to consider as the last processed on a fresh start.

Also renames config values to be more idiomatic due to several from DevOps and other people :)

## Why ❔

We don't want to wait half a year for VM to process all testnet/mainnet batches.

## Checklist


perekopskiy (approved):
`protobuf_compatibility` is failing but it's expected. It's not required job so PR can be merged

PR #2167, opened 2024-06-06
refactor(prover): Add config package

matias-gonz:
## What ❔

Add config package

## Checklist


Deniallugo (approved):

Deniallugo (approved):

Deniallugo:
> Most prover components load the config the same way as the ones updated. What would happen if we don't update them? My assumption is that they'll crash just like the rest of the components, but I could be wrong.

We are updating them one by one

PR #2168, opened 2024-06-06
fix(api): Fix transaction methods for pruned transactions

slowli:
## What ❔

Reworks `TxCache` for full nodes to look more like a mempool:

- Retains transactions from it when they are sent to the main node
- Gets rid of querying transactions from the main node in `TxProxy`.

## Why ❔

Right now, two transaction-related methods (`eth_getTransactionByHash` and `zks_getTransactionDetails`) return data for pruned transactions. Some other methods (e.g., `eth_getTransactionReceipt`) do not return data for pruned transactions (i.e., return `null`). This looks inconsistent; also may be wasteful w.r.t. calls to the main node.

## Checklist


core/node/api_server/src/tx_sender/proxy.rs
```diff
@@ -1,42 +1,101 @@
...
+///   We don't want to return such transactions if they are already included in an L2 block or replaced locally, but `Self::run_updates()`
+///   hasn't run yet.
 #[derive(Debug, Clone, Default)]
 pub(crate) struct TxCache {
     inner: Arc<RwLock<TxCacheInner>>,
 }
 
 #[derive(Debug, Default)]
 struct TxCacheInner {
+    // FIXME: implement expiration / LRU?
```
slowli, line 44: Should some kind of expiration and/or LRU logic be implemented right now? What would be an acceptable approach?

core/node/api_server/src/tx_sender/proxy.rs
```diff
@@ -158,12 +210,34 @@ impl TxProxy {
...
+        tx_hash: H256,
+    ) -> Result<Option<L2Tx>, Web3Error> {
+        let Some(tx) = self.tx_cache.get_tx(tx_hash).await else {
+            return Ok(None);
+        };
 
-    async fn forget_tx(&self, tx_hash: H256) {
-        self.tx_cache.remove_tx(tx_hash).await;
+        let initiator_address = tx.initiator_account();
+        let nonce_map = storage
```
slowli, line 223: Is getting this info from the DB on each transaction fetch acceptable?

core/node/api_server/src/tx_sender/proxy.rs
```diff
@@ -245,11 +280,7 @@ impl TxSink for TxProxy {
         // But before we do that, save the tx to cache in case someone will request it
         // Before it reaches the main node.
         self.save_tx(tx.clone()).await;
-        self.submit_tx_impl(tx).await?;
-        // Now, after we are sure that the tx is on the main node, remove it from cache
-        // since we don't want to store txs that might have been replaced or otherwise removed
-        // from the mempool.
-        self.forget_tx(tx.hash()).await;
+        self.submit_tx_impl(tx).await?; // FIXME: shouldn't tx be removed from cache on error?
```
slowli, line 283: Subj. Unless I'm missing something, if submitting a transaction fails, currently (including in `main`) the transaction will remain in `TxCache` indefinitely.

popzxc: I guess it should

slowli: It could make sense to retain it locally if there's some logic pushing such txs to the main node asynchronously and the original EN RPC call doesn't return an error. Neither of these 2 conditions hold (in the `main` branch as well); thus, it looks weird to retain the transaction in the cache even after returning an error to the client and having no way to remove this transaction from the cache later (esp. in `main`; with this PR, it could get removed eventually if it gets to the main node and is included into a block, or if we implement some sort of expiration / LRU logic for the cache).

popzxc: Sorry, I was referring to the comment in the code -- e.g. I think that tx should be removed from cache on error.

slowli: Got it. I'm just thinking aloud about why it would make sense to remove, and the corresponding pros / cons.

popzxc (dismissed):

popzxc (approved):

PR #2175, opened 2024-06-06
refactor(dal):  Deprecate l1_batches.l2_to_l1_logs

AnastasiiaVashchuk:
**review notes**: 
- changes in `core/lib/types/src/commitment/tests/**.json`s  - only formatting.
- as a follow-up to this PR there should be another PR that adds drop `l1_batches.l2_to_l1_logs` column migration.

## What ❔

- New dal method `get_l2_to_l1_logs_by_number`  that fetches logs for a specific number from `l2_to_l1_logs` table.
- More complicated way to construct `L1BatchHeader`. Now we  fetch data from the `l1_batches` table as `StorageL1BatchHeader `..then fetch l2l1 logs using  `get_l2_to_l1_logs_by_number` and after, use the helper  `into_l1_batch_header_with_logs ` method to construct `L1BatchHeader`.

## Why ❔

The main motivation for this change is to avoid keeping info about l2l1 logs in 2 tables.

## Checklist


perekopskiy (commented):
tests are failing

AnastasiiaVashchuk:
> tests are failing

yes, but looks like not related to the changes

core/lib/dal/src/blocks_web3_dal.rs
```diff
@@ -420,32 +420,45 @@ impl BlocksWeb3Dal<'_, '_> {
...
+                value
             FROM
-                l1_batches
+                l2_to_l1_logs
+                JOIN miniblocks ON l2_to_l1_logs.miniblock_number = miniblocks.number
             WHERE
-                number = $1
+                l1_batch_number = $1
+            ORDER BY
+                (tx_index_in_l1_batch, log_index_in_tx)
```
perekopskiy, line 452: please remove parenthesis here, I remember postgres working strange with them

core/lib/dal/src/blocks_web3_dal.rs
```diff
@@ -420,32 +420,45 @@ impl BlocksWeb3Dal<'_, '_> {
         Ok(hash)
     }
 
-    pub async fn get_l2_to_l1_logs(
+    pub async fn get_l2_to_l1_logs_by_number<L>(
```
perekopskiy, line 423: nit: IMO `get_l2_to_l1_logs_for_batch` sounds better

core/lib/dal/src/models/storage_block.rs
```diff
@@ -55,38 +54,41 @@ pub(crate) struct StorageL1BatchHeader {
     pub pubdata_input: Option<Vec<u8>>,
 }
 
-impl From<StorageL1BatchHeader> for L1BatchHeader {
-    fn from(l1_batch: StorageL1BatchHeader) -> Self {
-        let priority_ops_onchain_data: Vec<_> = l1_batch
+pub trait IntoL1BatchHeader {
```
slowli, line 57: Bikeshedding: IMO, this trait is redundant; you can define `into_l1_batch_header_with_logs()` methods as inherent in `StorageL1BatchHeader` / `StorageL1Batch`. It would be slightly simpler.

core/lib/dal/src/blocks_web3_dal.rs
slowli, line 423: IMO, it would make sense to make the templated method crate-private and expose its non-templated public variant with the same signature as before:

```rust
pub(crate) async fn get_l2_to_l1_logs_by_number<L>(
    &mut self,
) -> DalResult<Vec<L>>
where
    L: From<StorageL2ToL1Log>,
{
    // ...
}

pub async fn get_l2_to_l1_logs(
    &mut self,
) -> DalResult<Vec<api::L2ToL1Log>> {
    self.get_l2_to_l1_logs_by_number(l1_batch_number).await
}
```

It may make sense to place the crate-private method in `BlocksDal` since it is used outside of Web3-related methods.

core/lib/dal/src/blocks_dal.rs
```diff
@@ -281,7 +283,18 @@ impl BlocksDal<'_, '_> {
...
-        Ok(l1_batches.into_iter().map(Into::into).collect())
+        let mut l1_batch_headers = Vec::with_capacity(storage_l1_batch_headers.len());
+
+        for batch in storage_l1_batch_headers {
+            let l2_to_l1_logs = self
+                .storage
+                .blocks_web3_dal()
+                .get_l2_to_l1_logs_by_number::<UserL2ToL1Log>(L1BatchNumber(batch.number as u32))
+                .await?;
+            l1_batch_headers.push(batch.into_l1_batch_header_with_logs(l2_to_l1_logs));
```
slowli, line 294: Are there supposed to be a few L1 batches requested here? Otherwise, it may make sense to generalize `get_l2_to_l1_logs_by_number()` to support a set (range?) of batches. OTOH, `get_l1_batches_for_eth_tx_id()` seems to only be used in [metrics for `eth_sender`](https://github.com/matter-labs/zksync-era/blob/298a97e800b4c156628050789de7a490a7565d60/core/node/eth_sender/src/metrics.rs#L135) that doesn't really needs logs or most of other L1 batch header fields (it only needs timestamps and transaction counts), so it may make sense to refactor it and remove this method altogether.

core/lib/dal/src/models/storage_event.rs
```diff
@@ -1,5 +1,6 @@
 use zksync_types::{
-    api::{L2ToL1Log, Log},
+    api::{self},
```
slowli, line 2: Nit: Just `api` would do fine.

core/lib/dal/src/models/storage_event.rs
```diff
@@ -67,10 +67,10 @@ pub struct StorageL2ToL1Log {
...
 }
 
-impl From<StorageL2ToL1Log> for L2ToL1Log {
-    fn from(log: StorageL2ToL1Log) -> L2ToL1Log {
-        L2ToL1Log {
-            block_hash: log.block_hash.map(|hash| H256::from_slice(&hash)),
+impl From<StorageL2ToL1Log> for api::L2ToL1Log {
+    fn from(log: StorageL2ToL1Log) -> api::L2ToL1Log {
+        api::L2ToL1Log {
+            block_hash: None,
```
slowli, line 73: Dumb question: Is this field supposed to not be filled unless logs are are returned as a part of [transaction receipts](https://github.com/matter-labs/zksync-era/blob/298a97e800b4c156628050789de7a490a7565d60/core/lib/dal/src/transactions_web3_dal.rs#L132)?

core/lib/types/src/commitment/tests/post_boojum_1_4_1_test.json
slowli, line 1: Dumb question: What's the reason to remove these fixtures in this PR?

slowli (commented):
Organizational nit: This looks like a "refactor" or "feat" change, not a "fix"; IIUC, it doesn't fix incorrect behavior.

core/lib/dal/src/blocks_web3_dal.rs
AnastasiiaVashchuk: fair

AnastasiiaVashchuk:
> Organizational nit: This looks like a "refactor" or "feat" change, not a "fix"; IIUC, it doesn't fix incorrect behavior.

Yeah..probably `refactor` is next here.

core/lib/dal/src/models/storage_event.rs
AnastasiiaVashchuk: AFAIU - yes.

AnastasiiaVashchuk: Looking into this specific case...`l2_to_l1_logs` table does not have `block_hash` column. and in `dal` we do `NULL::bytea AS "block_hash"` for it

core/lib/dal/src/blocks_web3_dal.rs
AnastasiiaVashchuk: fair. Thanks for the suggestion.

core/lib/types/src/commitment/tests/post_boojum_1_4_1_test.json
AnastasiiaVashchuk: they are refactored(i mentioned this in PR header)
image

core/lib/dal/src/models/storage_block.rs
AnastasiiaVashchuk: won't it be confucing?

core/lib/dal/src/blocks_dal.rs
AnastasiiaVashchuk: >, so it may make sense to refactor 

By looking quickly into [metrics for eth_sender](https://github.com/matter-labs/zksync-era/blob/298a97e800b4c156628050789de7a490a7565d60/core/node/eth_sender/src/metrics.rs#L135)..the only quick way for refactoring I see is to *modify* `get_l1_batches_for_eth_tx_id()`, by asking db only for timestamps, l1_tx_count and l2_tx_count(Idk it will improve smth)
To get L1 batches' related info(timestamps, l1_tx_count and l2_tx_count), having only EthTx will be at least not easy/quick.

core/lib/dal/src/blocks_dal.rs
```diff
@@ -365,8 +375,20 @@ impl BlocksDal<'_, '_> {
...
         .fetch_optional(self.storage)
-        .await?
-        .map(Into::into))
+        .await?;
+
+        if let Some(storage_l1_batch_header) = storage_l1_batch_header {
+            let l2_to_l1_logs = self
+                .get_l2_to_l1_logs_for_batch::<UserL2ToL1Log>(L1BatchNumber(
+                    storage_l1_batch_header.number as u32,
+                ))
```
slowli, line 384: Nit: You can use the original `number` argument here.

core/lib/dal/src/blocks_dal.rs
```diff
@@ -1228,16 +1240,16 @@ impl BlocksDal<'_, '_> {
...
+        let mut l1_batches_with_metadata = Vec::with_capacity(raw_batches.len());
         for raw_batch in raw_batches {
             let block = self
                 .map_storage_l1_batch(raw_batch)
                 .await
-                .context("get_l1_batch_with_metadata()")?
+                .context("map_storage_l1_batch()")?
                 .context("Block should be complete")?;
-            l1_batches.push(block);
+            l1_batches_with_metadata.push(block);
```
slowli, line 1250: Nit / slightly off-topic: "block" should be renamed to "batch" in the local var naming and in `context()` message.

core/lib/dal/src/blocks_dal.rs
```diff
@@ -1774,7 +1787,16 @@ impl BlocksDal<'_, '_> {
...
+        let l2_to_l1_logs = self
+            .get_l2_to_l1_logs_for_batch::<UserL2ToL1Log>(L1BatchNumber(
+                storage_batch.number as u32,
+            ))
+            .await?;
+
+        let header: L1BatchHeader = storage_batch
+            .clone()
+            .into_l1_batch_header_with_logs(l2_to_l1_logs);
         let Ok(metadata) = storage_batch.try_into() else {
```
slowli, line 1800: Nit: Makes sense to do this conversion before loading logs so that an additional DB query isn't required if the batch metadata is incomplete.

core/lib/dal/src/blocks_dal.rs
```diff
@@ -2273,6 +2295,48 @@ impl BlocksDal<'_, '_> {
...
+                key,
+                value
+            FROM
+                l2_to_l1_logs
+                JOIN miniblocks ON l2_to_l1_logs.miniblock_number = miniblocks.number
+            WHERE
+                l1_batch_number = $1
+            ORDER BY
+                tx_index_in_l1_batch,
+                log_index_in_tx
```
slowli, line 2329: Dumb question: Is ordering by `miniblock_number, log_index_in_miniblock` (i.e., the primary key of `l2_to_l1_logs`) the same or different? Intuitively, if they are the same, ordering by the primary key could be more efficient.

core/lib/dal/src/models/storage_block.rs
slowli: Why would it be confusing? Different structs can define methods with the same name especially if the have the same semantics.

core/lib/dal/src/models/storage_event.rs
slowli: Yeah, I get that the current implementation in `main` sets `block_hash` to `null`; my question is whether it's a value that API clients expect. A potential alternative here would be to set it to an L2 block hash that the log is a part of.

core/lib/dal/src/blocks_dal.rs
slowli: Sorry, that's what I meant, but I wasn't specific enough: create a method like this

```rust
pub struct L1BatchStatistics {
    pub number: L1BatchNumber,
    pub timestamp: u64,
    pub transaction_count: u32,
}

pub async fn get_l1_batch_statistics_for_eth_tx_id(
    &mut self,
    eth_tx_id: u32,
) -> DalResult<Vec<L1BatchStatistics>> {
    // ...
}
```

AnastasiiaVashchuk: I see! Now makes much more sense for me!

core/lib/dal/src/models/storage_event.rs
AnastasiiaVashchuk: @perekopskiy  wdyt?

core/lib/dal/src/blocks_dal.rs
AnastasiiaVashchuk: sure.

core/lib/dal/src/blocks_dal.rs
AnastasiiaVashchuk: I have tested locally. Ordering by `key` does not work.

AnastasiiaVashchuk: ah nice!
changed

core/lib/dal/src/models/storage_event.rs
AnastasiiaVashchuk: Ah, just checked. `api::L2toL1Log` is used only in `get_l2_to_l1_logs_by_hashes`.  In current `get_l2_to_l1_logs_by_hashes` impl  we can not get `block_hash`(we can not do join with miniblocks table but it will be suboptimal) so it is fine to leave everything as is.

slowli (approved):

PR #2184, opened 2024-06-07
feat(prover): Add file based config for prover fri

matias-gonz:
## What ❔
Add file based config for prover fri

## Checklist


prover/prover_fri/src/main.rs
```diff
@@ -104,23 +107,31 @@ async fn main() -> anyhow::Result<()> {
...
+            .context("object_store config")?,
+    );
     let object_store_factory = ObjectStoreFactory::new(object_store_config.0);
-    let public_object_store_config =
-        PublicObjectStoreConfig::from_env().context("PublicObjectStoreConfig::from_env()")?;
+    let public_object_store_config = general_config
+        .snapshot_creator
+        .context("snapshot_creator config")?
+        .object_store
+        .context("object store config")?;
```
Deniallugo, line 121: It seems we have to add this object store as a `general.prover.public_object_store`
So, you will need to change the https://github.com/matter-labs/zksync-era/blob/8468716d6ef9ffcdd19a2625a41f14823547a0c9/core/lib/config/src/configs/fri_prover.rs

and add a new object store there. 

Try it yourself and ping me if you will need more guidance

Deniallugo (approved):

PR #2195, opened 2024-06-10
feat(zk_toolbox): Use docker compose instead of docker-compose

matias-gonz:
## What ❔

Use `docker compose` instead of `docker-compose`

## Why ❔

The recommended command-line syntax is docker compose:
https://docs.docker.com/compose/migrate/#docker-compose-vs-docker-compose


aon (dismissed):

zk_toolbox/crates/common/src/prerequisites.rs
```diff
@@ -11,10 +11,6 @@ const PREREQUISITES: [Prerequisite; 6] = [
         name: "docker",
         download_link: "https://docs.docker.com/get-docker/",
```
Deniallugo, line 12: should we check the version then? I do remember we had a lot of trouble with it a year ago

Deniallugo (approved):

PR #2196, opened 2024-06-10
feat(node): Port (most of) Node to the Node Framework

popzxc:
## What ❔

Ports most of the EN parts to the framework, except:
- Reorg detector
- Genesis / snapshot recovery
- Unique metrics stuff

## Why ❔

This forms the "main" body of the framework-based EN. The binary can already run, it can proxy and re-execute transactions, etc.
The remaining work may be somewhat complex, so I would like to ship it separately -- so that it receives thorough review.

There are some TODOs without a task number in this PR, expect them either to be fixed in the next one, or otherwise an issue to be added.

## Checklist


core/bin/external_node/src/tests.rs
```diff
@@ -264,6 +265,7 @@ async fn node_reacts_to_stop_signal_during_initial_reorg_detection() {
     let opt = Cli {
         enable_consensus: false,
         components: "core".parse().unwrap(),
+        use_node_framework: false,
```
AnastasiiaVashchuk, line 268: Do we plan to add some test with  `use_node_framework: true`?

popzxc: We plan to get rid of this option and make the node framework to be used by default, so, ugh, in a way - yes.

core/node/node_framework/src/implementations/layers/postgres_metrics.rs
```diff
@@ -0,0 +1,54 @@
...
+    }
+}
+
+#[derive(Debug)]
+struct PostgresMetricsScrapingTask {
+    pool_for_metrics: ConnectionPool<Core>,
+}
+
+#[async_trait::async_trait]
+impl Task for PostgresMetricsScrapingTask {
```
AnastasiiaVashchuk, line 38: `// Start scraping Postgres metrics before store initialisation as well.` comment in `core/bin/external_node/src/main.rs` makes me wonder why we  do `impl Task`  and not  `UnconstrainedTask`

core/node/node_framework/src/implementations/layers/state_keeper/mod.rs
```diff
@@ -32,12 +33,22 @@ use crate::{
 ///
 #[derive(Debug)]
 pub struct StateKeeperLayer {
-    db_config: DBConfig,
+    state_keeper_db_path: String,
+    state_keeper_db_block_cache_capacity: usize,
+    state_keeper_db_max_open_files: Option<NonZero<u32>>,
```
slowli, line 38: Bikeshedding: IMO, for forward compatibility it would be beneficial to use `RocksdbStorageOptions` here instead and accept it as an arg for `StateKeeperLayer::new()`.

core/node/node_framework/src/implementations/layers/postgres_metrics.rs
popzxc: That's a good catch -- it should.

core/bin/external_node/src/node_builder.rs
```diff
@@ -0,0 +1,511 @@
...
+            self.config.optional.l2_block_seal_queue_capacity,
+        )
+        .with_pre_insert_txs(true) // EN requires txs to be pre-inserted.
+        .with_protective_reads_persistence_enabled(
+            self.config.optional.protective_reads_persistence_enabled,
+        );
+
+        let io_layer = ExternalIOLayer::new(self.config.required.l2_chain_id);
+
+        // We only need call traces on the external node if the `debug_` namespace is enabled.
```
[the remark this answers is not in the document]
popzxc: Agree, but I think it deserves a separate discussion -- feel free to raise in chat.

etc/env/configs/ext-node.toml
```diff
@@ -55,6 +55,7 @@ url = "http://127.0.0.1:3050"
 # Here we use TOML multiline strings: newlines will be trimmed.
 log = """\
 warn,\
+zksync_node_framework=info,\
```
AnastasiiaVashchuk, line 58: also please add `zksync_node_consensus=info,\`

core/node/node_framework/src/implementations/layers/metadata_calculator.rs
```diff
@@ -154,3 +170,14 @@ impl Task for TreeApiTask {
             .await
     }
 }
+
+#[async_trait::async_trait]
+impl Task for MerkleTreePruningTask {
+    fn id(&self) -> TaskId {
+        "merkle_tree_pruning_task".into()
+    }
```
AnastasiiaVashchuk, line 178: nit: we don't mention `task` in other implementations of `fn id(&self) -> TaskId {`

core/node/node_framework/src/implementations/layers/validate_chain_ids.rs
```diff
@@ -0,0 +1,61 @@
...
+}
+
+#[async_trait::async_trait]
+impl Precondition for ValidateChainIdsTask {
+    fn id(&self) -> TaskId {
+        "validate_chain_ids".into()
+    }
+
+    async fn check(self: Box<Self>, stop_receiver: StopReceiver) -> anyhow::Result<()> {
+        (*self).run_once(stop_receiver.0).await
```
AnastasiiaVashchuk, line 59: why here it's run as a precondition(means only once), meanwile in `core/bin/external_node/src/main.rs` validate_chain_ids_task uses `run` instead of `run_once `?

core/bin/external_node/src/node_builder.rs
```diff
@@ -0,0 +1,511 @@
...
+                    // Do nothing, will be handled by the `Tree` component.
+                }
+                Component::TreeFetcher => {
+                    self = self.add_tree_data_fetcher_layer()?;
+                }
+                Component::Core => {
+                    // Core is a singleton & mandatory component,
+                    // so until we have a dedicated component for "auxiliary" tasks,
+                    // it's responsible for things like metrics.
+                    self = self.add_postgres_metrics_layer()?;
```
AnastasiiaVashchuk, line 495: Not fully understand why we decided to use `add_postgres_metrics_layer` only if `Component::Core` is enabled?
Orienting on the current initialization(`core/bin/external_node/src/main.rs  line:913`), we add `metrics_task` before `init_tasks`. Also, I see that the pool is used not only by the `Core` component.

core/node/node_framework/src/implementations/layers/tree_data_fetcher.rs
```diff
@@ -0,0 +1,68 @@
...
+
+#[async_trait::async_trait]
+impl Task for TreeDataFetcher {
+    fn id(&self) -> TaskId {
+        "tree_data_fetcher".into()
+    }
+
+    async fn run(self: Box<Self>, stop_receiver: StopReceiver) -> anyhow::Result<()> {
+        (*self).run(stop_receiver.0).await?;
+        Ok(())
```
AnastasiiaVashchuk, line 66: nit: `(*self).run(stop_receiver.0).await`

core/node/node_sync/src/validate_chain_ids_task.rs
```diff
@@ -138,6 +138,23 @@ impl ValidateChainIdsTask {
...
+        let main_node_l2_check =
+            Self::check_l2_chain_using_main_node(self.main_node_client, self.l2_chain_id);
+        let joined_futures =
+            futures::future::try_join3(eth_client_check, main_node_l1_check, main_node_l2_check)
+                .fuse();
+        tokio::select! {
+            res = joined_futures => res.map(drop),
+            _ = stop_receiver.changed() =>  Ok(()),
+        }
+    }
```
AnastasiiaVashchuk, line 155: Do I understand correctly that after the complete porting, the method below(`run`) will be removed?

core/bin/external_node/src/node_builder.rs
```diff
@@ -0,0 +1,511 @@
...
+                .config
+                .optional
+                .slow_query_threshold()
+                .map(|d| d.as_millis() as u64),
+            test_server_url: None,
+            test_prover_url: None,
+        };
+        let secrets = DatabaseSecrets {
+            server_url: Some(self.config.postgres.database_url()),
+            server_replica_url: Some(self.config.postgres.database_url()),
```
AnastasiiaVashchuk, line 96: do not understand why to set `server_replica_url` here

core/node/node_framework/src/implementations/layers/postgres_metrics.rs
AnastasiiaVashchuk: I guess same for `PrometheusExporterLayer`.

core/bin/external_node/src/node_builder.rs
```diff
@@ -0,0 +1,511 @@
...
+            self.config.required.l1_chain_id,
+            self.config.required.eth_client_url.clone(),
+        );
+        self.node.add_layer(query_eth_client_layer);
+        Ok(self)
+    }
+
+    fn add_state_keeper_layer(mut self) -> anyhow::Result<Self> {
+        // While optional bytecode compression may be disabled on the main node, there are batches where
+        // bytecode compression was enabled. To process these batches (and also for the case where
```
AnastasiiaVashchuk, line 160:
```suggestion
        // optional bytecode compression was enabled. To process these batches (and also for the case where
```

core/node/node_framework/src/implementations/layers/l1_batch_commitment_mode_validation.rs
```diff
@@ -0,0 +1,59 @@
...
+    implementations::resources::eth_interface::EthInterfaceResource,
+    precondition::Precondition,
+    service::{ServiceContext, StopReceiver},
+    task::TaskId,
+    wiring_layer::{WiringError, WiringLayer},
+};
+
+#[derive(Debug)]
+pub struct L1BatchCommitmentModeValidationLayer {
+    duamond_proxy_addr: Address,
```
AnastasiiaVashchuk, line 14: duamond_proxy_addr -> diamond_proxy_addr

core/bin/external_node/src/node_builder.rs
```diff
@@ -0,0 +1,511 @@
...
+            acquire_timeout_sec: None,
+            statement_timeout_sec: None,
+            long_connection_threshold_ms: None,
+            slow_query_threshold_ms: self
+                .config
+                .optional
+                .slow_query_threshold()
+                .map(|d| d.as_millis() as u64),
+            test_server_url: None,
+            test_prover_url: None,
```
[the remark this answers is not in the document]
popzxc: Uh oh hehehe. Looks like this one is _messy_. 
We use these fields in `from_env` to (I guess?) check [that they're deserializable](https://github.com/matter-labs/zksync-era/blob/main/core/lib/env_config/src/database.rs#L59), while actually [we access it via `env::var`](https://github.com/matter-labs/zksync-era/blob/main/core/lib/db_connection/src/connection_pool.rs#L154).

I guess we can safely remove it from the config -- if the value is not present, the error message should be informative enough, but it's out of scope for this PR indeed.

core/node/node_framework/src/implementations/layers/metadata_calculator.rs
popzxc: IMO `merkle_tree_pruning` would sound weird. E.g. here it's a task not because it's `impl Task for <...>`, but it's _really_ a part of its name.

core/bin/external_node/src/node_builder.rs
popzxc: Old EN codebase is not fully adapted for working in a distributed way. Postgres metrics task scrapes the metrics from postgres -- they will be the same for any component that is connected to postgres -- and exports them to prometheus. We only need a single component exporting these metrics, because otherwise the exported metrics will be duplicated.

So 2 parts of the answer for your question:
1. We need to run it with core, because we need it to run only with a single component, and `Core` is a mandatory singleton componenet of EN. On main node, we have house keeper for such stuff.
2. The old code was written before EN could run in a distributed mode, and it is not suitable for distributed mode.

core/node/node_sync/src/validate_chain_ids_task.rs
popzxc: Most likely.

core/bin/external_node/src/node_builder.rs
popzxc: Because api layer uses replica pool. In the old code, we could pass two copies of main pool to API, but the framework considers main pool and different pool as different resources -- so we configure the pools layer to provide both, currently pointing to the same (main) pool in fact.
In the future, we will probably want to support replica pool configuration for EN too, since we now support running in a distributed mode.

core/bin/external_node/src/node_builder.rs
```diff
@@ -0,0 +1,511 @@
...
+    fn add_query_eth_client_layer(mut self) -> anyhow::Result<Self> {
+        let query_eth_client_layer = QueryEthClientLayer::new(
+            self.config.required.l1_chain_id,
+            self.config.required.eth_client_url.clone(),
+        );
+        self.node.add_layer(query_eth_client_layer);
+        Ok(self)
+    }
+
+    fn add_state_keeper_layer(mut self) -> anyhow::Result<Self> {
```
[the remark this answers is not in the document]
AnastasiiaVashchuk: yes. because it is not very logical to call method `add_state_keeper_layer` and add a bunch of layers(not only state_keeper"

core/bin/external_node/src/node_builder.rs
AnastasiiaVashchuk: that answers my question. thanks!

core/bin/external_node/src/node_builder.rs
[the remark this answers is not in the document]
popzxc: These layers are parts of state keeper -- e.g. they fill the resources that are requested by the state keeper and state keeper _only_ (in other words, there is no value in these layers without state keeper, and there is no state keeper without these layers (or their alternatives) added). On the other hand, sequencer is not defined solely by the state keeeper -- for example, Merkle Tree is also part of the sequencer, but it isn't added here.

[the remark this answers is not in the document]
AnastasiiaVashchuk: I see now. Thanks.

AnastasiiaVashchuk (approved):

PR #2197, opened 2024-06-10
fix(en): Fix reorg detection in presence of tree data fetcher

slowli:
## What ❔

Fixes reorg detection logic so that it accounts for the tree data fetcher:

- **In tree data fetcher:** Tries to detect reorgs, so that root hashes are not written for diverging L1 batches.
- **In reorg detector:** Checks last L2 block correspondence during binary searching a diverging L1 batch.

## Why ❔

Reorg detection may be broken if tree data fetcher is enabled:

- The tree data fetcher doesn't check that fetched L1 batch root hashes correspond to local L1 batches, i.e. it can fetch a root hash after a revert.
- Hence, the logic in reorg detector which binary-searches the diverged L1 batch is broken because the latest L1 batch isn't guaranteed to diverge if there's a divergence.

## Checklist

### DIFF
--- a/core/node/node_sync/src/tree_data_fetcher/metrics.rs
+++ b/core/node/node_sync/src/tree_data_fetcher/metrics.rs
@@ -40,6 +40,7 @@ pub(super) enum StepOutcomeLabel {
     UpdatedBatch,
     NoProgress,
     RemoteHashMissing,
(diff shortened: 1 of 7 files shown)


popzxc (approved):
Nice!

PR #2204, opened 2024-06-11
feat: Base Token Fundamentals

shahar4:
## What ❔

This PR puts in place the major pieces needed in the Base Token flow as described in the [design doc](https://www.notion.so/matterlabs/Custom-Base-Token-Tech-Design-74e158112f24463c82d96415bed76d04). 

This PR adds a preliminary flow of:
- Periodically fetch the new price from an external API
- Update the conversion ration in the DB
- Have the Fee Model use the latest stored conversion ratio

In the scope of this PR:
- Introducing BaseTokenAdjuster to the Node Framework
- Adding the DB migration and the associated DAL
- Adding the proper configs for this layer to work
- Conditionally updating the gas price with a BaseToken<->ETH conversion ratio such that it's used in charging the gas in the base token and estimating gas for txs.
- Unit + Integration tests

Out of scope of this PR:
- Protocol contract changes needed for a base token chain to work (these are already merged)
- Using an external API to fetch prices
- Updating the L1 with the conversion ratio
- Observability, API redundancy,

## Why ❔
Multiple ZK Chains are ready to go on mainnet with a custom token used as the gas token.

## Checklist


juan518munoz:
Hi, we've been working on this implementation ourselves in this [PR](https://github.com/matter-labs/zksync-era/pull/2141/files). It's close to being finished, but considering you are already developing it, we are leaving it as is and closing the pull request. Feel free make use of any of it's parts if you find it useful.

core/lib/dal/migrations/20240611121747_add_base_token_price_table.up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE base_token_prices (
+    id SERIAL PRIMARY KEY,
+    created_at TIMESTAMP NOT NULL,
+    updated_at TIMESTAMP NOT NULL,
+
+    ratio_timestamp TIMESTAMP NOT NULL,
```
thomas-nguy, line 6: why do we need the timestamp?

From the DAL, I don't see any use case where we need the base_token_price other than the latest.

The table will grow infinitely and fetching price will get slower over time

core/node/base_token_adjuster/src/base_token_adjuster.rs
```diff
@@ -0,0 +1,121 @@
...
+                &api_price.denominator,
+                &api_price.ratio_timestamp.naive_utc(),
+            )
+            .await?;
+        drop(conn);
+
+        Ok(())
+    }
+
+    async fn maybe_update_l1(
```
thomas-nguy, line 93: to be more clean: create two functions 

- should_update_l1(). ->  custom logics that could be overided by hyperchains?
- update_l1() -> do the update operation, I don't think it require customisation

and replace the block by 
```
if (should_update_l1()) {
update_l1()
} 
```

core/node/fee_model/src/lib.rs
```diff
@@ -209,10 +222,19 @@ fn compute_batch_fee_model_input_v2(
...
 
-    PubdataIndependentBatchFeeModelInput {
-        l1_gas_price,
-        fair_l2_gas_price,
-        fair_pubdata_price,
+    // if there is base token quote available - multiplying the final results onto it
+    // if not - ETH is used as a base token
+    match base_token_price {
+        Some(x) => PubdataIndependentBatchFeeModelInput {
+            l1_gas_price: (x * l1_gas_price as f64) as u64,
```
thomas-nguy, line 229: how safe is it?

wouldnt it be safer to do

```
l1_gas_price / denominator) * nominator
```

core/lib/dal/migrations/20240611121747_add_base_token_price_table.up.sql
shahar4: To the very least we're going to need to persist the 
- Latest ratio
- Latest ratio used in the L1
Besides that, we were thinking that multiple records might be needed for debugging.

We were planning on adding a cleaning process for this table. Our plan was to roll it out without one, see how this is used, and create one accordingly. Perhaps we should already go ahead and persist only some days worth of history.

core/node/base_token_adjuster/src/base_token_adjuster.rs
shahar4: Good point. You can ignore this as updating the L1 is out of scope for this PR. This is here mostly to show the flow.

core/lib/dal/migrations/20240611121747_add_base_token_price_table.up.sql
thomas-nguy: gotcha

I wonder if keeping the record is necessary as it will be persisted on L1 anyways ?

The latest L1 ratio can also be fetched on L1, no need to duplicate it in DB.

I think only the boolean `used_in_l1` really matter here so that we know if the component crash midway, then we need to start by sending to L1 before fetching the new rate

core/lib/dal/src/base_token_dal.rs
```diff
@@ -0,0 +1,56 @@
...
+    pub async fn insert_token_price(
+        &mut self,
+        numerator: &BigDecimal,
+        denominator: &BigDecimal,
+        ratio_timestamp: &chrono::NaiveDateTime,
+    ) -> DalResult<()> {
+        sqlx::query!(
+            r#"
+            INSERT INTO
+                base_token_prices (numerator, denominator, ratio_timestamp)
```
thomas-nguy, line 21: there are few other values that require to be not null and without default values if you refer to the schema such as created_at or updated_at

core/node/base_token_adjuster/src/base_token_adjuster.rs
```diff
@@ -165,21 +163,27 @@ impl MainNodeBaseTokenAdjuster {
 
         tokio::time::sleep(sleep_duration).await;
     }
+
+    /// If no base token L1 address is set then the base token is ETH.
+    fn is_base_token_eth(&self) -> bool {
+        self.base_token_l1_address.is_none()
```
thomas-nguy, line 169:
```suggestion
        !self.base_token_l1_address.map_or(true, |v| v == L1_ETH_ADDRESS);
```

You also need to check if its the L1_ETH_ADDRESS

core/lib/config/src/configs/contracts.rs
```diff
@@ -37,6 +37,7 @@ pub struct ContractsConfig {
     pub l2_testnet_paymaster_addr: Option<Address>,
     pub l1_multicall3_addr: Address,
     pub ecosystem_contracts: Option<EcosystemContracts>,
+    // Used for the RPC API and by the BaseTokenAdjuster. If not set, ETH is the chain's base token.
```
thomas-nguy, line 40: for ETH based chain, it is not None but defaulted to “0x0..01”

core/node/base_token_adjuster/src/base_token_adjuster.rs
```diff
@@ -0,0 +1,429 @@
...
+/// Converts the fee parameters to the base token using the latest ratio fetched from the DB.
+fn convert_to_base_token(params: FeeParamsV2, base_token_to_eth: BigDecimal) -> FeeParamsV2 {
+    let FeeParamsV2 {
+        config,
+        l1_gas_price,
+        l1_pubdata_price,
+    } = params;
+
+    let convert_price = |price_in_wei: u64| -> u64 {
+        let converted_price_bd = BigDecimal::from(price_in_wei) * base_token_to_eth.clone();
```
XinyuCRO, line 207: the conversion rate should be `eth_to_base_token` ? 
according to doc: https://github.com/matter-labs/zksync-era/blob/627aab9703c47795247f8b6d21533520498ed025/docs/guides/advanced/fee_model.md#custom-base-token-configurations

core/lib/types/src/fee_model.rs
```diff
@@ -236,9 +237,76 @@ pub struct FeeParamsV1 {
...
+            l1_gas_price,
+            l1_pubdata_price,
+            conversion_ratio,
+        }
+    }
+
+    /// Returns the fee model config with the minimal L2 gas price denominated in the chain's base token (WEI or equivalent).
+    pub fn config(&self) -> FeeModelConfigV2 {
+        FeeModelConfigV2 {
+            minimal_l2_gas_price: self.convert_to_base_token(self.config.minimal_l2_gas_price),
```
thomas-nguy, line 264: If you do that, we need also to change the logic in the smart contract

https://github.com/matter-labs/era-contracts/blob/db9387690502937de081a959b164db5a5262ce0a/l1-contracts/contracts/state-transition/chain-deps/facets/Mailbox.sol#L180

otherwise it will become very confusing

core/node/base_token_adjuster/src/base_token_adjuster.rs
```diff
@@ -0,0 +1,216 @@
...
+            if *stop_receiver.borrow() {
+                tracing::info!("Stop signal received, base_token_adjuster is shutting down");
+                break;
+            }
+
+            let start_time = Instant::now();
+
+            match self.fetch_new_ratio().await {
+                Ok(new_ratio) => {
+                    match self.persist_ratio(&new_ratio, &pool).await {
```
thomas-nguy, line 59: This logic is will most likely be overridden by hyper chains operator as one may want to persist the ratio only under some conditions 

 Better to move the altogether with the `run` function under `external_price_component`

core/node/node_framework/src/implementations/layers/base_token_adjuster.rs
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    async fn run(self: Box<Self>, stop_receiver: StopReceiver) -> anyhow::Result<()> {
+        let mut adjuster = zksync_base_token_adjuster::MainNodeBaseTokenAdjuster::new(
+            self.main_pool,
+            self.config,
+            self.base_token_l1_address,
+        );
+
+        adjuster.run(stop_receiver.0).await
```
thomas-nguy, line 69: No need to run if `base_token_l1_address` is `ETH` 

Could print a warning if its the case

core/node/base_token_adjuster/src/base_token_adjuster.rs
shahar4: 100% -- made the token address non optional and comparing with the ETH shared bridge "address"

core/lib/dal/migrations/20240611121747_add_base_token_price_table.up.sql
shahar4: Why we persist the ratio:
- If the external price APIs crash, you don't want the server to halt
- Be able to know what the last ratio stored on the L1 and be prudent with frequent L1 updating
- Possible debugging
- For the operator to potentially look back and collect data on prices to analyze their off-chain BaseToken<->ETH conversions

Also, there's a ticket to add to `HouseKeeper` the ability to clear rows over 7 days old.

core/lib/config/src/configs/contracts.rs
shahar4: thanks! changed :)

core/node/base_token_adjuster/src/base_token_adjuster.rs
shahar4: @XinyuCRO I'm not sure I'm following. The PR was in draft mode and this was changed. If this is still relevant please comment again.

core/lib/types/src/fee_model.rs
shahar4: @thomas-nguy I think we covered this offline that no SC changes are needed. If you need more context or disagree please unresolve

core/node/base_token_adjuster/src/base_token_adjuster.rs
shahar4: A few thoughts:
- I think this comment conflates persisting the most recent ratio with updating the L1
- I believe operators will want to customize the behavior, but not override it. They will be able to decide how often new price quotes are fetched, and under what conditions the L1 price is updated
- I also believe that operators might want to override what external APIs they're reaching out to, and so this logic is encapsulated outside of the `BaseTokenAdjuster`
- Lastly, this is injected into the node via the node framework, so operators will be able to override whatever this entire layer if any of the core logic isn't fulfilling their needs

Please lmk if you see things differently!

core/node/node_framework/src/implementations/layers/base_token_adjuster.rs
shahar4: This would introduce many more conditionals to the flow. Running this component with ETH as base token would hold a 1:1 conversion ratio.

thomas-nguy: `This would introduce many more conditionals to the flow.`
Could you explain that part?

This component will run in the background and will insert the same entry every 30s (default) in DB when none of that is required for ETH-based chain. Not to mention the house keeper that also need to purge the records. 

At least please add some documentation to make sure operators does not run it when it is not necessary

core/node/node_framework/src/implementations/layers/l1_gas.rs
```diff
@@ -75,8 +84,18 @@ impl WiringLayer for SequencerL1GasLayer {
...
         let gas_adjuster = Arc::new(adjuster);
 
+        let pool_resource = context.get_resource::<PoolResource<ReplicaPool>>().await?;
+        let replica_pool = pool_resource.get().await?;
+
+        let base_token_adjuster = MainNodeBaseTokenAdjuster::new(
+            replica_pool.clone(),
+            self.base_token_adjuster_config,
+            self.base_token_l1_address,
+        );
```
popzxc, line 94: I thought `Arc<dyn BaseTokenAdjuster>` would be a resource.
Then here we may fetch this resource, and if it's not provided, use a `NoOpBaseTokenAdjuster` which will always return 1.
Then separately, we will have `CustomBaseTokenAdjusterLayer` which will provide a resource and spawn the required maintenance task.

This way if we add `CustomBaseTokenAdjusterLayer`, gas prices will be adjusted. If we don't add, the system will work just fine for ETH.

core/lib/types/src/fee_model.rs
```diff
@@ -236,9 +237,76 @@ pub struct FeeParamsV1 {
...
+                u64::MAX
+            }
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+pub struct BaseTokenConversionRatio {
+    pub numerator: u64,
+    pub denominator: u64,
```
popzxc, line 309: Should it be [`NonZeroU64`](https://doc.rust-lang.org/std/num/type.NonZeroU64.html)?

core/node/base_token_adjuster/src/base_token_adjuster.rs
```diff
@@ -0,0 +1,214 @@
...
+            }
+
+            self.sleep_until_next_fetch(start_time).await;
+        }
+        Ok(())
+    }
+
+    // TODO (PE-135): Use real API client to fetch new ratio through self.PriceAPIClient & mock for tests.
+    //  For now, these are hard coded dummy values.
+    async fn fetch_new_ratio(&self) -> anyhow::Result<BaseTokenAPIRatio> {
```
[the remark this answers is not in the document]
shahar4: I agreed in principle, but to deliver this PR and be able to work with this as a foundation I left it out. There's already [follow-up draft PR](https://github.com/matter-labs/zksync-era/pull/2315) on the `PriceAPIClient` resource.

core/node/base_token_adjuster/src/base_token_adjuster.rs
```diff
@@ -0,0 +1,214 @@
...
+    async fn get_conversion_ratio(&self) -> anyhow::Result<BaseTokenConversionRatio> {
+        if self.is_base_token_eth() {
+            return Ok(BaseTokenConversionRatio {
+                numerator: 1,
+                denominator: 1,
+            });
+        }
+
+        // Retries are necessary for the initial setup, where prices may not yet be persisted.
+        self.retry_get_latest_price().await
```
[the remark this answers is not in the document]
shahar4: Makes sense. Changed to a caching flow with `get_conversion_ratio` simply taking the cached value.

core/node/base_token_adjuster/src/base_token_adjuster.rs
```diff
@@ -0,0 +1,214 @@
...
+
+#[async_trait]
+impl BaseTokenAdjuster for MainNodeBaseTokenAdjuster {
+    async fn get_conversion_ratio(&self) -> anyhow::Result<BaseTokenConversionRatio> {
+        if self.is_base_token_eth() {
+            return Ok(BaseTokenConversionRatio {
+                numerator: 1,
+                denominator: 1,
+            });
+        }
```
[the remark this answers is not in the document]
shahar4: Resolving as this isn't relevant anymore with the new flow.

core/node/base_token_adjuster/src/base_token_adjuster.rs
```diff
@@ -0,0 +1,214 @@
...
+        &self,
+        api_price: &BaseTokenAPIRatio,
+        pool: &ConnectionPool<Core>,
+    ) -> anyhow::Result<usize> {
+        let mut conn = pool
+            .connection_tagged("base_token_adjuster")
+            .await
+            .context("Failed to obtain connection to the database")?;
+
+        let id = conn
```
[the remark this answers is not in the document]
shahar4: It's for the L1-updating flow which I decided to keep out of this PR and will happen as an immediate follow-up. We will use this ID to update the DB entry that a specific conversion ratio was used in the L1.

core/node/node_framework/src/implementations/layers/base_token_adjuster.rs
shahar4: Resolving this as with the new flow the persisting layer isn't added if the operated chain has ETH as a base token and the added resource will be a NOOP resource in this case.

core/node/fee_model/src/l1_gas_price/gas_adjuster/mod.rs
```diff
@@ -19,6 +20,11 @@ mod metrics;
 #[cfg(test)]
 mod tests;
 
+pub trait L1GasAdjuster: Debug + Send + Sync {
+    fn estimate_effective_gas_price(&self) -> u64;
+    fn estimate_effective_pubdata_price(&self) -> u64;
+}
```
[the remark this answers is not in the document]
shahar4: Good point. Was using it for unit tests but could do without it. Reverted this change.

core/node/fee_model/src/l1_gas_price/gas_adjuster/mod.rs
```diff
@@ -431,3 +437,29 @@ impl<T: Ord + Copy + Default> GasStatistics<T> {
         self.0.read().unwrap().last_processed_block
     }
 }
+
+#[derive(Debug)]
+#[allow(dead_code)]
+pub struct MockGasAdjuster {
```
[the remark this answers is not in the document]
shahar4: It was but no longer is.

core/node/node_framework/src/implementations/layers/base_token_fetcher.rs
```diff
@@ -0,0 +1,72 @@
...
+#[async_trait::async_trait]
+impl Task for BaseTokenFetcher {
+    fn id(&self) -> TaskId {
+        "base_token_fetcher".into()
+    }
+
+    async fn run(mut self: Box<Self>, stop_receiver: StopReceiver) -> anyhow::Result<()> {
+        (*self).run(stop_receiver.0).await?;
+        Ok(())
+    }
```
popzxc, line 71: Now that I thought about it more -- it looks like the task approach won't work.
GasAdjuster task [has a check](https://github.com/matter-labs/zksync-era/blob/main/core/node/node_framework/src/implementations/layers/l1_gas.rs#L103) that only actually runs the task if someone uses the adjuster. However, here we can't do the same -- because the only user of the resource is the `GasAdjuster`, and it always has a reference to the base token fetcher.

Probably we have no choice but to add an `update` method to the `BaseTokenFetcher`, and invoke it from `GasAdjuster::keep_updated`

core/node/base_token_adjuster/src/base_token_fetcher.rs
```diff
@@ -0,0 +1,123 @@
...
+    pub latest_ratio: BaseTokenConversionRatio,
+    pub config: BaseTokenAdjusterConfig,
+}
+
+impl BaseTokenFetcher {
+    pub fn new(pool: Option<ConnectionPool<Core>>, config: BaseTokenAdjusterConfig) -> Self {
+        let latest_ratio = BaseTokenConversionRatio {
+            numerator: config.initial_numerator,
+            denominator: config.initial_denominator,
+        };
```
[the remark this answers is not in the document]
shahar4: Removed initial config & added the DB fetching to the async constructor.

core/node/base_token_adjuster/src/base_token_fetcher.rs
```diff
@@ -0,0 +1,123 @@
...
+use tokio::{sync::watch, time::sleep};
+use zksync_config::BaseTokenAdjusterConfig;
+use zksync_dal::{ConnectionPool, Core, CoreDal};
+use zksync_types::{base_token_ratio::BaseTokenRatio, fee_model::BaseTokenConversionRatio};
+
+const CACHE_UPDATE_INTERVAL: Duration = Duration::from_millis(500);
+
+#[derive(Debug, Clone)]
+pub struct BaseTokenFetcher {
+    pub pool: Option<ConnectionPool<Core>>,
```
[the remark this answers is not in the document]
shahar4: Added the `BaseTokenFetcher` and a separate NoOp impl to avoid this.

core/lib/dal/migrations/20240611121747_add_base_token_ratio_table.up.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE base_token_ratios (
+    id SERIAL PRIMARY KEY,
+    created_at TIMESTAMP NOT NULL,
+    updated_at TIMESTAMP NOT NULL,
+
+    ratio_timestamp TIMESTAMP NOT NULL,
+    numerator NUMERIC NOT NULL,
+    denominator NUMERIC NOT NULL,
```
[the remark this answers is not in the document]
shahar4: For consistency. When communicating with the L1, we cannot store a float ratio in the contract so have to have `numerator`/`denominator` in integers.

core/node/node_framework/src/implementations/layers/base_token_fetcher.rs
shahar4: Leaving this as is as discussed offline.

core/node/fee_model/src/lib.rs
```diff
@@ -130,8 +143,9 @@ impl BatchFeeModelInputProvider for ApiFeeInputProvider {
     }
 
     /// Returns the fee model parameters.
-    fn get_fee_model_params(&self) -> FeeParams {
-        self.inner.get_fee_model_params()
+    async fn get_fee_model_params(&self) -> anyhow::Result<FeeParams> {
+        self.inner.get_fee_model_params().await
+        // TODO(PE-137): impl APIBaseTokenAdjuster and use it in the ApiFeeInputProvider.
```
popzxc, line 148: Do we need to? I think that `get_batch_fee_input_scaled` already covers everything we need (e.g. we either chose current params or the params from the last block, whichever is stricter).

core/lib/types/src/fee_model.rs
```diff
@@ -236,9 +239,85 @@ pub struct FeeParamsV1 {
...
+            None => {
+                if converted_price_bd > BigDecimal::from(u64::MAX) {
+                    tracing::warn!(
+                        "Conversion to base token price failed: converted price is too large: {}. Using u64::MAX instead.",
+                        converted_price_bd
+                    );
+                } else {
+                    tracing::error!(
+                    "Conversion to base token price failed: converted price is not a valid u64: {}. Using u64::MAX instead.",
+                    converted_price_bd
```
popzxc, line 299: This shouldn't be possible, right? The numerator, denominator, and gas price are all positive, and we've checked that the value is less than `u64::MAX`. So reaching this point would mean a big problem in the implementation. I'd say it's adequate to panic here, because we don't know what's going on and it's not safe to continue.

core/lib/types/src/fee_model.rs
```diff
@@ -236,9 +239,85 @@ pub struct FeeParamsV1 {
...
+        self.convert_to_base_token(self.l1_gas_price)
+    }
+
+    /// Returns the l1 pubdata price denominated in the chain's base token (WEI or equivalent).
+    pub fn l1_pubdata_price(&self) -> u64 {
+        self.convert_to_base_token(self.l1_pubdata_price)
+    }
+
+    /// Converts the fee param to the base token.
+    fn convert_to_base_token(&self, price_in_wei: u64) -> u64 {
```
popzxc, line 282: We probably want metrics for stuff like base token price, price ratio, converted price for observability; without this information it may be hard to debug issues in production.
It doesn't have to be added in this PR, but please create a task and don't forget to add it (both in code and in grafana once we have at least one deployment).

core/node/base_token_adjuster/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "zksync_base_token_adjuster"
```
popzxc, line 2: Naming suggestion (for crates, structures, components, layers, etc):
`base_token_adjuster` -> `base_token_ratio_perserver`
`base_token_fetcher` -> `base_token_ratio_provider`

Still clunky, but, ugh, IMHO, less clunky.

core/node/base_token_adjuster/src/base_token_fetcher.rs
```diff
@@ -0,0 +1,119 @@
...
+                denominator: latest_storage_ratio.denominator,
+            };
+        }
+
+        tracing::info!("Stop signal received, base_token_fetcher is shutting down");
+        Ok(())
+    }
+}
+
+async fn get_latest_price(pool: ConnectionPool<Core>) -> anyhow::Result<BaseTokenConversionRatio> {
```
popzxc, line 58: Can be make into a method to avoid cloning pool (and to keep structural integrity).

core/node/base_token_adjuster/README.md
```diff
@@ -0,0 +1,13 @@
...
+This crate contains all the logic to handle ZK Chain with custom base tokens. It is used by other node layers to adjust
+the fees to be denominated in the chain's base token.
+
+## Overview
+
+The Base Token Adjuster:
+
+- Connect with external APIs to get the current price of the base token.
+- Persist the price of the base token in the database.
+- Upon request, adjust the fees to be denominated in the base token.
```
popzxc, line 12: This is not true, right? The component doesn't seem to do that.

core/node/fee_model/src/lib.rs
shahar4: We don't. This is a leftover. Removed!

core/lib/types/src/fee_model.rs
shahar4: 100%. There's already a task for all of this.

core/node/base_token_adjuster/README.md
shahar4: Correct. Updated the readme!

core/lib/types/src/fee_model.rs
```diff
@@ -236,9 +239,85 @@ pub struct FeeParamsV1 {
...
+    /// Returns the fee model config with the minimal L2 gas price denominated in the chain's base token (WEI or equivalent).
+    pub fn config(&self) -> FeeModelConfigV2 {
+        FeeModelConfigV2 {
+            minimal_l2_gas_price: self.convert_to_base_token(self.config.minimal_l2_gas_price),
+            ..self.config
+        }
+    }
+
+    /// Returns the l1 gas price denominated in the chain's base token (WEI or equivalent).
+    pub fn l1_gas_price(&self) -> u64 {
```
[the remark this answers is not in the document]
shahar4: Added a ticket for this!

core/lib/dal/migrations/20240611121747_add_base_token_ratio_table.up.sql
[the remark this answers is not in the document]
shahar4: It seems like `NUMERIC` is the best choice here as `BIGINT` doesn't cover the full range as it's signed. I added precision and scale to represent a `u64 better`.

core/lib/config/src/configs/contracts.rs
```diff
@@ -37,6 +37,8 @@ pub struct ContractsConfig {
     pub l2_testnet_paymaster_addr: Option<Address>,
     pub l1_multicall3_addr: Address,
     pub ecosystem_contracts: Option<EcosystemContracts>,
+    // Used by the RPC API and by the node builder in wiring the BaseTokenRatioProvider layer. If not set,
+    // ETH is the chain's base token and the "ETH address" is SHARED_BRIDGE_ETHER_TOKEN_ADDRESS.
     pub base_token_addr: Option<Address>,
```
popzxc, line 42: Nit: you may add a `pub fn base_token_addr() -> Address { self.base_token_addr.unwrap_or(SHARED_BRIDGE_ETHER_TOKEN_ADDRESS) }` to simplify usage.

core/lib/dal/src/base_token_dal.rs
```diff
@@ -0,0 +1,63 @@
...
+            ratio_timestamp,
+        )
+        .instrument("insert_token_ratio")
+        .fetch_one(self.storage)
+        .await?;
+
+        Ok(row.id as usize)
+    }
+
+    // TODO (PE-128): pub async fn mark_l1_update()
```
popzxc, line 41: Nit: It's better to leave TODO comments on something that is already implemented but misses some functionality. If nothing has been done yet, it's better to just rely on the task; otherwise such comments look confusing.

core/node/base_token_adjuster/src/base_token_ratio_provider.rs
```diff
@@ -0,0 +1,124 @@
+use std::{fmt::Debug, num::NonZeroU64, time::Duration};
+
+use anyhow::Context;
+use async_trait::async_trait;
+use tokio::sync::watch;
+use zksync_dal::{ConnectionPool, Core, CoreDal};
+use zksync_types::fee_model::BaseTokenConversionRatio;
+
+const CACHE_UPDATE_INTERVAL: Duration = Duration::from_millis(500);
```
popzxc, line 9: Does it have to be this small? IIRC we use a single value for the whole batch, so updating it very often may result in unpredictable behavior. I don't have any concrete suggestions at this point, but please thing through the whole flow to make sure that both UX and operator's predictability are good.

core/node/fee_model/src/lib.rs
```diff
@@ -53,34 +55,45 @@ impl dyn BatchFeeModelInputProvider {
...
+/// This struct gets the L1 gas price directly from the provider rather than from another node, as is the
+/// case with the external node.
 #[derive(Debug)]
 pub struct MainNodeFeeInputProvider {
     provider: Arc<GasAdjuster>,
+    base_token_ratio_provider: Arc<dyn BaseTokenRatioProvider>,
     config: FeeModelConfig,
 }
 
+#[async_trait]
```
popzxc, line 68: Not needed. Actually looks like the whole trait doesn't need `async_trait`, so it can be removed completely.

core/node/node_framework/src/implementations/layers/base_token_ratio_provider.rs
```diff
@@ -0,0 +1,65 @@
...
+
+/// Wiring layer for `BaseTokenRatioProvider`
+///
+/// Responsible for serving the latest ETH<->BaseToken conversion ratio. This layer is only wired if
+/// the base token is not ETH. If wired, this layer inserts the BaseTokenRatioProviderResource and kicks
+/// off a task to poll the DB for the latest ratio and cache it.
+///
+/// If the base token is ETH, a default, no-op impl of the BaseTokenRatioProviderResource is used by other
+/// layers to always return a conversion ratio of 1.
+
```
popzxc, line 23: Nit: empty line should be `///`

popzxc (commented):
Mostly looks good. Left a few nits.

core/lib/config/src/configs/base_token_adjuster.rs
```diff
@@ -0,0 +1,25 @@
...
+        Self {
+            price_polling_interval_ms: Self::default_interval(),
+        }
+    }
+}
+
+impl BaseTokenAdjusterConfig {
+    pub fn default_interval() -> u64 {
+        DEFAULT_INTERVAL_MS
+    }
```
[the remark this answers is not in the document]
shahar4: Not sure I agree
- The default should only be used in this file. In fact I shouldn't have made it public (changing this)
- Otherwise, the actual pub field could be used to fetch the actual value, not the default
- Even creating a getter here that returns the `Duration` and removes the need to convert to `Duration` seems overkill to me as this field is only used in one place as of now

WDYT?

core/lib/dal/src/base_token_dal.rs
shahar4: Agreed as a general principle, but the purpose of this here, and other TODOs around this PR is to not only place the TODO but also indicate to the reader the intended full flow.

core/node/base_token_adjuster/src/base_token_ratio_provider.rs
shahar4: I have thought through this. There's a funky thing with our fee model that we
- take one value for the entire batch
- upon estimateGas, take the higher of (prev block and current estimation)
So half a second, to me, captures the tradeoff of reflecting current behavior while not caching excessively. The only thing it doesn't capture is that in reality, operators are unlikely to fetch new ratios from external APIs every half a second. We will keep this value in mind as we observe patterns of external fetching frequency

core/node/fee_model/src/lib.rs
shahar4: One of the impls of `get_batch_fee_input_scaled` is async so keeping this as is (this was also the case prior to this PR, I just imported `use async_trait::async_trait;`.

core/lib/config/src/configs/contracts.rs
shahar4: I'm not so convinced. The `base_token_addr` being set to ETH address happens during chain registration. If the chain registration logic changes and we keep this field optional we may hide a bug. This was why I wanted to remove the optionality. I think that my clarification comment has done more to confuse than clarify so removing it!

core/lib/config/src/configs/base_token_adjuster.rs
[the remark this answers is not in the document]
popzxc: Oh, I didn't mean `fn default_interval() -> Duration`, I meant `fn price_polling_interval(&self) -> Duration`. I still think it makes sense to add it, even for a single use.
No objections on publicity of methods.

core/lib/config/src/configs/contracts.rs
popzxc: To not hide a bug, we may add a cross-check to make sure that the variable has the same value as contract (like [here](https://github.com/matter-labs/zksync-era/blob/main/core/node/node_framework/src/implementations/layers/l1_batch_commitment_mode_validation.rs)), and it would be a better guarantee.
We're only talking about convenience here, not safety.

core/lib/dal/src/base_token_dal.rs
popzxc: The problem with such TODOs is that pretty often people either forget to remove them or priorities change and these TODOs are left hanging. I think the intended full flow can be implied from the already existing implementation, so idk.

core/node/base_token_adjuster/src/base_token_ratio_provider.rs
popzxc: Up to you. You may even make it an optional configuration parameter if you want (as you did [here](https://github.com/matter-labs/zksync-era/blob/cfdeb10f0fa8518e74790ec7099923f7546fa2a4/core/lib/config/src/configs/base_token_adjuster.rs#L10)).

core/node/fee_model/src/lib.rs
popzxc: Ah, true. Didn't notice that, sorry.

core/node/base_token_adjuster/src/base_token_ratio_provider.rs
shahar4: Oh yes, maybe that's the way to go indeed! I added an explicit ticket to tackle this as a small add once we have more info.

popzxc (approved):
Looks mergeable, with the expectation of the follow-up work.

hatemosphere (approved):

PR #2213, opened 2024-06-11
chore(eth-sender): extact abstract l1 interface and all fee-related code

tomg10:
This PR deliberately only moves the code around without refactoring the logic itself to decrease the risk of error. More PRs will most likely follow

### DIFF
--- a/core/node/eth_sender/src/eth_fees_oracle.rs
+++ b/core/node/eth_sender/src/eth_fees_oracle.rs
@@ -1,0 +1,148 @@
+use std::{
+    cmp::{max, min},
+    fmt,
+    sync::Arc,
+};
+
+use zksync_eth_client::{ClientError, EnrichedClientError};
+use zksync_node_fee_model::l1_gas_price::L1TxParamsProvider;
+use zksync_types::eth_sender::TxHistory;
+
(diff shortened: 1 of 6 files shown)


core/node/eth_sender/src/eth_fees_oracle.rs
```diff
@@ -0,0 +1,148 @@
...
+    fn calculate_fees(
+        &self,
+        previous_sent_tx: &Option<TxHistory>,
+        has_blob_sidecar: bool,
+        time_in_mempool: u32,
+    ) -> Result<EthFees, EthSenderError> {
+        match has_blob_sidecar {
+            true => self.calculate_fees_with_blob_sidecar(previous_sent_tx),
+            false => self.calculate_fees_no_blob_sidecar(previous_sent_tx, time_in_mempool),
+        }
```
perekopskiy, line 146: nit: IMO using `if` is more appropriate than matching boolean

perekopskiy (approved):

core/node/eth_sender/src/eth_fees_oracle.rs
tomg10: indeed, corrected

PR #2223, opened 2024-06-12
fix(dal): Fix contract_address field in getTransactionRecipt

AnastasiiaVashchuk:
## What ❔

+ Change `get_transaction_receipts`  DAL method to fetch the contract address from the `events` table. 
+ Add integration test that checks we return the correct contract address in the receipt of the transaction that deployed inner-outer contract(e.g. contract that initializes another contract in its constructor).

## Why ❔
 The current approach(going to the `storage_logs` table to get the contract address) is error-prone. The problem here is that we deduplicate storage logs, which renders operation_number invalid. So, the ordering is not correct, and we might return the incorrect deployed address. This is important when one tx deploys a contract that initializes another contract in the constructor(which means we deploy several contracts here) - the user expects the "outer" contract address but gets the inner.

## Checklist


core/tests/ts-integration/tests/api/web3.test.ts
```diff
@@ -932,6 +934,21 @@ describe('web3 API compatibility tests', () => {
...
+        let innerContractBytecode = contracts.inner.bytecode;
+        let outerContractOverrides = {
+            customData: {
+                factoryDeps: [innerContractBytecode]
+            }
+        };
+        const outerContract = await deployContract(alice, contracts.outer, [1], undefined, outerContractOverrides);
+        let receipt = await outerContract.deployTransaction.wait();
+
+        expect(outerContract.address).toEqual(receipt.contractAddress);
```
perekopskiy, line 949: I'm not sure how address which is returned from `deployContract` is calculated. If it's taken from `receipt.contractAddress` then the assert doesn't check what we want to check. Let's add another `expect` that checks that bytecode hash of `receipt.contractAddress` is indeed bytecode hash of outer

AnastasiiaVashchuk: Hmmm....I thought once again, and I do not believe that checking bytecode hashes will increase test robustness.
I have decided to check receipt contract address with precalculated address throw `utils.createAddress`

perekopskiy: Both works

core/tests/ts-integration/tests/api/web3.test.ts
```diff
@@ -932,6 +934,24 @@ describe('web3 API compatibility tests', () => {
         expect(txFromApi.v! <= 1).toEqual(true);
     });
 
+    // We want to be sure that correct(outer) contract address is return in the transaction receipt,
+    // when there is a contract that initializa another contract in the constructor
+    test('Should check inner-outer contract address in the receipt of the deploy tx', async () => {
+        const nonce = await alice.getNonce();
+        const expectedAddress = zksync.utils.createAddress(alice.address, nonce);
```
perekopskiy, line 941: You should use "deploy nonce", not the usual one. Tbh it will be easier if you simply check bytecode/bytecodeHash for `receipt.contractAddress` to match expected one

AnastasiiaVashchuk: 1. what do you mean by "deploy nonce"?
2. why to check bytecodes if the goal is to check the fact we extract correct contract addresses from the DB?

perekopskiy: 1. Check docs https://docs.zksync.io/build/developer-reference/ethereum-differences/nonces
2. Yes, but the correct address should be the only one with the bytecode of `outerContract`

AnastasiiaVashchuk: I see, thanks!

AnastasiiaVashchuk: feel free to check

perekopskiy (approved):

PR #2232, opened 2024-06-13
feat: verification of L1Batch witness (BFT-471) - attempt 2

pompon0:
Rollforward matter-labs/zksync-era#2019 which was reverted by matter-labs/zksync-era#2230

Made Execute json encoding super backward compatible.
The only change from the previous attempt is in core/lib/types/src/execute.rs


core/lib/types/src/tx/execute.rs
```diff
@@ -4,30 +4,53 @@ use zksync_utils::ZeroPrefixHexSerde;
 
 use crate::{ethabi, Address, EIP712TypedStructure, StructBuilder, H256, U256};
 
-/// `Execute` transaction executes a previously deployed smart contract in the L2 rollup.
-#[derive(Clone, Default, Serialize, Deserialize, PartialEq)]
+#[derive(Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
+struct ExecuteSerde {
```
popzxc, line 9: A comment on why do we need this structure (and why we don't just slap `derive` on `Execute`) would be helpful

pompon0: done

popzxc (approved):

popzxc (approved):

PR #2233, opened 2024-06-13
fix(prover): Disallow state changes from successful

EmilLuta:
This PR is done as a fix for boojnet outage.

TL;DR; of outage -- race condition caused by prover jobs moving from 'successful` state to `in_progress`/`in_gpu_proving`.

The PR addresses: 
- no job can move from successful state (considered final state)
- fix local development (contracts were pointing to 0.24.0 instead of 0.24.1) -- can be split to a different PR, if this is problematic.
- add table constraint -- again, can be split in different PR
- add checks for recursion_tip number of jobs (post outage check, should not happen ever, but better to verify)


prover/prover_dal/src/fri_proof_compressor_dal.rs
```diff
@@ -177,10 +160,14 @@ impl FriProofCompressorDal<'_, '_> {
                 updated_at = NOW()
             WHERE
                 l1_batch_number = $3
+                AND status != $4
+                OR status != $5
```
perekopskiy, line 164: Maybe AND?

prover/witness_generator/src/recursion_tip.rs
```diff
@@ -156,6 +156,8 @@ impl JobProcessor for RecursionTipWitnessGenerator {
             .get_final_node_proof_job_ids_for(l1_batch_number)
             .await;
 
+        assert_eq!(final_node_proof_job_ids.len(), number_of_final_node_jobs as usize, "recursion tip witness job was scheduled without all final node jobs being completed; expected {}, got {}", number_of_final_node_jobs, final_node_proof_job_ids.len());
```
perekopskiy, line 159: not sure how lints passed, but the line is way too long

perekopskiy (approved):

prover/prover_dal/src/fri_proof_compressor_dal.rs
```diff
@@ -177,10 +160,14 @@ impl FriProofCompressorDal<'_, '_> {
...
                 l1_batch_number = $3
+                AND status != $4
+                AND status != $5
             "#,
             ProofCompressionJobStatus::Failed.to_string(),
             error,
-            i64::from(block_number.0)
+            i64::from(block_number.0),
+            ProofCompressionJobStatus::Successful.to_string(),
+            ProofCompressionJobStatus::SentToServer.to_string(),
```
RomanBrodetski, line 170: IMO it would be worth it to log such occurances. This would make it easier to debug potential additional occurances. 
You can make `update` return the number of affected rows and then compare with `1`.

PR #2234, opened 2024-06-13
perf(db): Improve storage switching for state keeper cache

slowli:
## What ❔

Improves switching logic between Postgres and RocksDB for SK cache. With these changes, RocksDB is guaranteed to be used if it's up to date when it's opened.

## Why ❔

Previously, Postgres could be used after node start (primarily if there's a pending L1 batch) even if RocksDB is up to date. This is caused by potential delays when opening RocksDB. This behavior was observed in the wild, e.g. for a mainnet full node with pruning enabled.

## Checklist

### DIFF
--- a/core/lib/state/src/catchup.rs
+++ b/core/lib/state/src/catchup.rs
@@ -105,23 +105,33 @@ pub struct AsyncCatchupTask {
 impl AsyncCatchupTask {
     /// Create a new catch-up task with the provided Postgres and RocksDB instances. Optionally
     /// accepts the last L1 batch number to catch up to (defaults to latest if not specified).
-    pub fn new(
-        pool: ConnectionPool<Core>,
-        state_keeper_db_path: String,
-        state_keeper_db_options: RocksdbStorageOptions,
-        to_l1_batch_number: Option<L1BatchNumber>,
-    ) -> (Self, RocksdbCell) {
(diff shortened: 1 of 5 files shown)


core/lib/state/src/catchup.rs
```diff
@@ -49,7 +141,10 @@ impl AsyncCatchupTask {
     /// Propagates RocksDB and Postgres errors.
     pub async fn run(self, stop_receiver: watch::Receiver<bool>) -> anyhow::Result<()> {
         let started_at = Instant::now();
-        tracing::debug!("Catching up RocksDB asynchronously");
+        tracing::debug!(
+            "Catching up RocksDB asynchronously to L1 batch #{:?}",
+            self.to_l1_batch_number
+        );
```
itegulov, line 147: nit: This is kind of opinionated but wouldn't this be cleaner (structured logs and all that good stuff)?

```suggestion
        tracing::debug!(
            target_l1_batch = ?self.to_l1_batch_number,
            "Catching up RocksDB asynchronously",
        );
```

popzxc (dismissed):

itegulov (approved):

core/lib/state/src/catchup.rs
slowli: Implemented it in a slightly different way, via `tracing::instrument` spans.

itegulov (approved):

PR #2235, opened 2024-06-13
feat(eth-sender): add eth_sender_lock table i.e exclusive locks in eth-sender

tomg10:


RomanBrodetski:
is the sole purpose of this to avoid running two instances of eth-sedner?

PR #2240, opened 2024-06-14
chore: move contract verifier logic to lib

AntonD3:
## What ❔

Move the contract verifier struct to the separate lib crate.

## Why ❔

This logic can be used as a library. At lest, for example, now it's needed for some compiler analysis tool.

## Checklist


AntonD3:
I didn't change the name of the bin crate(it's still `zksync_contract_verifier`) to be sure that build/run scripts will not be broken, so I named the lib crate - `zksync_contract_verifier_lib` for now.

PR #2244, opened 2024-06-14
feat: Expose fair_pubdata_price for blocks and batches

mm-zk:
## What ❔

* Exposing fair_pubdata_price in our block & batches API

## Why ❔

* This is crucial information for projects that need to replay our transactions (for example era_test_node).
* Without this, they cannot compute the correct gas costs.


perekopskiy:
Hmm, It's exposed in `en_syncBlock` which is a preferred endpoint for re-execution purposes. But it seems it makes sense to add it to `zks_getBlockDetails` as well as it already returns `fair_l2_gas_price`, right?

mm-zk:
@perekopskiy - good point. The alternative is for me to switch era-test-node to use en_SyncBlock. But yes, as zks_getBlockDetails already returns other gas pricing, it still makes sense for it to return all of it.

PR #2067, opened 2024-05-28
DONT MERGE: Change store chain-id for eth-* compoenents

StanislavBreadless:
## What ❔

1. Make eth watch track its state in a separate table
2. Make eth_sender track chain id for transactions so that nonce is fetched on per-chain-id basis

**Note that it is breaking for ENs due to some fields being removed and breaking serde encoding. Thus, [this](https://github.com/matter-labs/zksync-era/pull/2091) PR should be merged first**

## Checklist

- [ ] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [ ] Tests for the changes have been added / updated.
- [ ] Documentation comments have been added / updated.
- [ ] Code has been formatted via `zk fmt` and `zk lint`.
- [ ] Spellcheck has been run via `zk spellcheck`.

PR #2245, opened 2024-06-14
feat: upgraded encoding of transactions in consensus Payload.

pompon0:
Currently the encoded transaction is up to 3x larger than its rlp encoding. This is caused by data duplication between: raw, input and factory_deps fields. In the new encoding we use the rlp directly. It will be used starting with protocol version 25.


core/lib/dal/src/consensus/proto/mod.proto
```diff
@@ -13,10 +13,30 @@ message Payload {
...
   optional uint32 virtual_blocks = 6; // required
   optional bytes operator_address = 7; // required; H160
+  // Set for protocol_version < 25.
   repeated Transaction transactions = 8;
+  // Set for protocol_version >= 25.
+  repeated TransactionV25 transactions_v25 = 12;
   optional bool last_in_batch = 10; // required
 }
 
+message L1Transaction {
```
moshababo, line 23: naming: L1Transaction**V25** / L2Transaction**V25** ?

core/lib/dal/src/consensus/mod.rs
```diff
@@ -100,6 +128,50 @@ impl Payload {
...
+
+    fn build(tx: &Self::Type) -> Self {
+        let tx = abi::Transaction::try_from(tx.clone()).unwrap();
+        use proto::transaction_v25::T;
+        Self {
+            t: Some(match tx {
+                abi::Transaction::L1 {
+                    tx, factory_deps, ..
+                } => T::L1(proto::L1Transaction {
+                    rlp: Some(ethabi::encode(&[tx.encode()])),
```
moshababo, line 166: What’s the rationale for using ABI/RLP encoding here?

core/lib/dal/src/consensus/mod.rs
```diff
@@ -100,6 +128,50 @@ impl Payload {
...
+                        // Unwrap is safe because `ethabi::decode` does the verification.
+                        let tx =
+                            abi::L2CanonicalTransaction::decode(tokens.into_iter().next().unwrap())
+                                .context("L2CanonicalTransaction::decode()")?;
+                        Ok(tx)
+                    })
+                    .context("rlp")?
+                    .into(),
+                factory_deps: l1.factory_deps.clone(),
+                eth_block: 0,
```
moshababo, line 151: If `eth_block` still defined if deprecated and unused?

core/lib/dal/src/consensus/mod.rs
```diff
@@ -38,38 +39,59 @@ pub struct Payload {
...
-        for (i, tx) in message.transactions.iter().enumerate() {
-            transactions.push(tx.read().with_context(|| format!("transactions[{i}]"))?)
+    fn read(r: &Self::Proto) -> anyhow::Result<Self> {
+        let protocol_version = required(&r.protocol_version)
+            .and_then(|x| Ok(ProtocolVersionId::try_from(u16::try_from(*x)?)?))
+            .context("protocol_version")?;
+        let mut transactions = vec![];
+
+        match protocol_version {
+            v if v >= ProtocolVersionId::Version25 => {
```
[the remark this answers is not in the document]
pompon0: Migration between versions is a gradual process. Currently the envs are using V24, and at the moment they transition to V25, the binary will need to support both encodings. Only once all envs move to V25 we will be able to remove support for V24.

core/lib/dal/src/consensus/mod.rs
pompon0: the rationale is that RLP encoding is a public standard, therefore very stable. There is no value in adding an alternative proto encoding for rlp structs.

core/lib/dal/src/consensus/mod.rs
pompon0: eth_block will start to be populated with 0s after https://github.com/matter-labs/zksync-era/pull/2067 is merged and deployed. Until then eth_block is non-zero. Afterwards we will still have non-zero eth_block values for the old certificates (hopefully only for V24 or older). It is not like I can remove it from abi::L1Transaction just like that.

popzxc (approved):

core/lib/dal/src/consensus/proto/mod.proto
pompon0: ack, however this is supposed to be new standard encoding and it is the role of the top level message to indicate which protocol version it should be used with, so I'd prefer to avoid spamming V25 suffix too much.

core/lib/dal/src/consensus/tests.rs
```diff
@@ -1,21 +1,75 @@
...
+
+use super::{proto, Payload};
+use crate::tests::mock_protocol_upgrade_transaction;
+
+fn execute(rng: &mut impl Rng) -> Execute {
+    Execute {
+        contract_address: rng.gen(),
+        value: rng.gen::<u128>().into(),
+        calldata: (0..10 * 32).map(|_| rng.gen()).collect(),
+        // TODO: find a way to generate valid random bytecode.
```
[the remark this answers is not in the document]
pompon0: ack

core/lib/dal/Cargo.toml
```diff
@@ -49,5 +49,9 @@ strum = { workspace = true, features = ["derive"] }
 tracing.workspace = true
 chrono = { workspace = true, features = ["serde"] }
 
+[dev-dependencies]
+zksync_test_account.workspace = true
+zksync_concurrency.workspace = true
```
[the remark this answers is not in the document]
pompon0: My standpoint on this aspect didn't change - I don't want the dal crate to expose public api which can invalidate storage invariants - it means that input data needs to be validated before modifying storage. If you have a concrete proposal on how to that split, then let's discuss.

Regarding those 2 extra dev-dependencies - neither of them is consensus-specific: test_account just provides a way of generating test data for types from zksync_types, and zksync_concurrency is general purpose concurrency library, just with limited usage atm.

PR #2247, opened 2024-06-14
feat(eth-sender): separate gas calculations for blobs transactions

tomg10:
This PR implements point (1) from
https://www.notion.so/matterlabs/Eth-sender-blob-fees-problem-b84e1715248944559a0a656a6c9da320

### DIFF
--- a/core/node/fee_model/src/l1_gas_price/gas_adjuster/mod.rs
+++ b/core/node/fee_model/src/l1_gas_price/gas_adjuster/mod.rs
@@ -335,6 +335,10 @@ impl L1TxParamsProvider for GasAdjuster {
         self.config.default_priority_fee_per_gas
     }
 
+    // The idea is that when we finally decide to send blob tx, we want to offer gas fees high
+    // enough to "almost be certain" that the transaction gets included. To never have to double
+    // the gas prices as then we have very little control how much we pay in the end. This strategy
+    // works as no matter if we double or triple such price, we pay the same block base fees.
     fn get_blob_tx_base_fee(&self) -> u64 {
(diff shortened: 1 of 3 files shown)


core/node/fee_model/src/l1_gas_price/gas_adjuster/mod.rs
```diff
@@ -334,6 +334,18 @@ impl L1TxParamsProvider for GasAdjuster {
...
+    fn get_blob_tx_base_fee(&self) -> u64 {
+        self.base_fee_statistics.last_added_value() * 2
+    }
+
+    fn get_blob_tx_blob_base_fee(&self) -> u64 {
+        self.blob_base_fee_statistics.last_added_value().as_u64() * 2
+    }
+
+    fn get_blob_tx_priority_fee(&self) -> u64 {
+        self.get_priority_fee() * 2
```
montekki, line 347: Can you explain this part, why are the values doubled from the getgo? Am i understanding correctly that for the first submission of the transaction here this multiplication also works?

tomg10: This is correct, I definitely need to add comments to this piece of code. The idea is that when we finally decide to send blob tx, we want to offer gas fees high enough to "almost be certain" that the transaction gets included. To never have to double the gas prices as then we have very little control how much we pay in the end. All our competitors use this strategy. This PR is only a "half" of the solution, what is missing is delaying sending blob transactions until the prices are acceptable, you can see it here: https://github.com/matter-labs/zksync-era/pull/2305

montekki (approved):

PR #2250, opened 2024-06-14
chore: add details to state diff hash mismatch error

dimazhornyk:
## What ❔

Add details to state diff hash mismatch error.

## Why ❔

The error in this assert is not informative and doesn't help much during debugging.

## Checklist


core/lib/types/src/commitment/mod.rs
```diff
@@ -355,7 +357,7 @@ impl L1BatchAuxiliaryOutput {
 
                 let state_diffs_packed = serialize_commitments(&state_diffs);
                 let state_diffs_hash = H256::from(keccak256(&(state_diffs_packed)));
-                let state_diffs_compressed = compress_state_diffs(state_diffs);
+                let state_diffs_compressed = compress_state_diffs(state_diffs.clone());
```
joonazan, line 360: This clone seems unnecessary. Could compress_state_diffs take a reference instead?

joonazan: &mut would work. It's a bit dirty but we really don't want to clone that huge array.

deniallugo-ml:

PR #2252, opened 2024-06-17
chore: use simple StorageLog struct instead of LogQuery when possible

joonazan:
There are many fields in LogQuery on the core side (it also exists in zk_evm) that are unused and often just set to some garbage value but that is very confusing.

This PR changes the uses of LogQuery in the VmInterface to StorageLog, which only has the necessary fields.


core/node/state_keeper/src/io/persistence.rs
```diff
@@ -482,27 +474,18 @@ mod tests {
...
-                        initial_value: log.log_query.read_value,
-                        final_value: log.log_query.written_value,
-                    }
+                .filter(|&log| log.kind == StorageLogKind::InitialWrite)
+                .map(|log| StateDiffRecord {
+                    address: *log.key.address(),
+                    key: h256_to_u256(*log.key.key()),
+                    derived_key: log.key.hashed_key().0,
+                    enumeration_index: 0,
+                    initial_value: U256::zero(), // TODO not sure about this one
```
joonazan, line 487: This is the only change that I'm not confident about. This test might actually need an initial value that it now doesn't get anymore.

core/node/genesis/src/utils.rs
```diff
@@ -41,78 +40,60 @@ pub(super) async fn add_eth_token(transaction: &mut Connection<'_, Core>) -> any
...
-                        rollback: false,
-                        is_service: false,
-                    }
-                })
-                .collect::<Vec<_>>()
+        .map(move |storage_log| {
+            MultiVmLogQuery {
+                // Monotonically increasing Timestamp. Normally it's generated by the VM, but we don't have a VM in the genesis block.
+                timestamp: MultiVMTimestamp(0),
+                tx_number_in_block: 0,
```
perekopskiy, line 86: Nit: let's change the comment explaining that it's fine to set 0 for timestamp and tx_number_in_block here

perekopskiy (approved):

PR #2256, opened 2024-06-18
feat(en): Unify snapshot recovery and recovery from L1

slowli:
## What ❔

- Replaces `storage_logs.{address, key}` fields in the snapshot with `storage_logs.hashed_key`.
- Makes `storage_logs.{address, key}` fields in Postgres optional and adapts the codebase for it.

## Why ❔

- This will make snapshot data equivalent to data obtainable from L1, and thus would allow to unify snapshot recovery and recovery from L1 data.
- Decreases the snapshot size somewhat.

## Checklist


core/tests/recovery-test/tests/snapshot-recovery.test.ts
```diff
@@ -160,17 +162,23 @@ describe('snapshot recovery', () => {
...
+                    const snapshotValue = '0x' + storageLog.storageValue.toString('hex');
+                    const valueOnBlockchain = await mainNode.getStorageAt(
+                        snapshotAccountAddress,
+                        snapshotKey,
+                        l2BlockNumber
+                    );
+                    expect(snapshotValue).to.equal(valueOnBlockchain);
+                } else {
+                    // TODO: requires adding a method like `zks_resolveStorageAt(hashedKey)` returning
+                    //   { address, key, enumerationIndex }
```
slowli, line 180: Would such a method be useful?

slowli (commented):
I feel that backward compatibility is clunky, but I'm not sure how to make it better.

- The snapshot creator now supports creating both v0 and v1 snapshots (configured via an env variable). If we start creating *only* v1 snapshots, the old EN versions will fail to recover from a snapshot.
- Snapshot creator logic is largely unchanged, meaning that it's impossible to create v0 and v1 snapshots for the same L1 batch, or to start creating a v1 snapshot while v0 is in progress (or vice versa). The latter looks like an arbitrary restriction, but it looks moderately easy to fix. Having an L1 batch number as the snapshot ID is more deep-rooted in the codebase / API, so there could be issues changing it.
- New EN versions support recovery from both v0 and v1 snapshots. However, there's currently no way to command the EN to recover from a snapshot with a specific version (which could be useful for E2E testing), other than manually calling RPC methods and finding an L1 batch number with the appropriate version, and then specifying recovery from this specific snapshot.

tomg10:
Well... I told you as soon as we release the Node, this unification would be painful. I don't think we should add support for having multiple versions for the same l1 batch, I'm also against changing API just for executing this migration. Can you write down migration plan to the new format as you envision it now?

tomg10:
Can't we just stop supporting v0 as soon as we release a breaking version of EN? Instead of having two snapshots creators, just switch snapshots-creator to generate v1 as soon as that happens?

slowli:
That's related to how much we trust the v1 format correctness. If we trust it completely, we can not bother with the EN and switch the snapshot creator from v0 to v1 as you've described. IMO, it's moderately risky. Otherwise, we want to run v1 snapshot creation in parallel with v0 for some time for E2E tests, and have a switch in EN to recover from v1 / v0 snapshots for the same reason.

tomg10:
We can always rollback to v0 if we encounter problems

tomg10:
Though we should probably locally at least once generate mainnet v1 snapshot and recover from it

slowli:
I guess we theoretically can, but I'm not a fan of testing features in production 🙃 If I correctly understand your proposal (switching the creator from v0 to v1 once when a node version with v1 support is released), we won't be able to test recovery end-to-end until it's in production *at all*. Whereas with the gradual rollout we can, and it won't impact external node operators. Also, to clarify your proposal: Do you think it makes sense to change the `snapshots` RPC as I've described above, or is it OK for old node versions to stop being able to recover after the snapshot creator switch from v0 to v1? (Unless the node operator manually finds and specifies an L1 batch to recover from, I guess.)

slowli:
As long as we don't start producing v1 snapshots right away, missing functionality I've described above (API changes; EN opting in to v1 snapshots) looks OK to implement separately; this PR is large as is. We should agree on the general migration approach though 🙃

tomg10:
> we won't be able to test recovery end-to-end until it's in production at all.
I think this is not true, but would require some temporary hacking, i.e. forcing the node to use locally generated snapshot instead of a one from mainnet

tomg10:
> is it OK for old node versions to stop being able to recover after the snapshot creator switch from v0 to v1

I think so; It's okay for old versions to stop working after releasing a major version

tomg10:
> Unless the node operator manually finds and specifies an L1 batch to recover from, I guess.

Old snapshots are kept only for a month

slowli:
I think I've figured out how migration to v1 snapshots can be made less painful, at least initially, and implemented it in the recent commits. Basically, allow the EN to drop storage key preimages if recovering from snapshot v0; in this way, we can test soundness of not having key preimages without ever creating snapshots v1. We can set this flag for E2E recovery test and for some of our ENs. I can theoretically now remove v1 snapshots altogether from this PR; but I'd prefer not to because they are implemented already 🙃

RomanBrodetski:
Great work! 

I don't have capacity to do a detailed review, but I had a quick look and it looks good. I also think there is a lot of value in delivering this - even though it won't be used that much, it's important to have this functionality - it's important for the chain liveness in case of a catastrophic event

core/lib/snapshots_applier/src/tests/mod.rs
```diff
@@ -1,5 +1,7 @@
 //! Snapshot applier tests.
 
+// FIXME: test dropping preimages for version 0 snapshot
```
popzxc, line 3: Will this be addressed in this PR?

core/tests/recovery-test/tests/snapshot-recovery.test.ts
popzxc: I think it's better not to add any new methods unless there is a strong business or technical need for it.
Right now I don't think this method will bring much value.

popzxc (dismissed):
I can't say I did a very thorough review (not enough time, unfortunately), but I've read through the whole diff and didn't find any significant issues.
The change makes sense, and the migration process seems to be adequate (though we probably will need to involve DevRell to communicate it).

core/lib/snapshots_applier/src/tests/mod.rs
slowli: It is already addressed (`applier_recovers_v0_snapshot` test with `drop_storage_key_preimages` set), I just forgot to remove the FIXME.

popzxc (dismissed):

popzxc (dismissed):
The third time's the charm

tomg10 (approved):

popzxc (approved):
OK, here goes the fourth

artmakh (approved):

PR #2262, opened 2024-06-18
[DO NOT MERGE] new eth-sender design

tomg10:
Note that this PR is intentionally not complete, I've included only API, schema and a readme to make reviewing and discussing it easier


core/lib/dal/migrations/20240612152419_create_l1_transactions_status.up.sql
```diff
@@ -0,0 +1,25 @@
...
+    blob_sidecar     BYTEA,
+    tx_hash          TEXT,
+    tx_type          transaction_type   NOT NULL,
+    status           transaction_status NOT NULL,
+
+    created_at       TIMESTAMP          NOT NULL,
+    updated_at       TIMESTAMP          NOT NULL
+);
+
+-- for some time, we will need to fetch both old and new transactions
```
tomg10, line 21: This will be the first step of the migration. After I'm sure that on every env we have started using l1_transactions table, I will backfill l1_transactions table and stop using eth_txs table in core code.

core/node/eth_sender/README.md
```diff
@@ -0,0 +1,34 @@
...
+where `idempotency_key` can be just any String and status be one of `Pending`, `Confirmed`, `Failed`
+
+Attempting to send tx for the same `idempotency_key`, but with different parameters returns in an error.
+
+Nonces are determined via order of calling send_tx.
+
+## Integration with server-v2
+
+Whenever we want to send an L1 transaction, we create an entry in `l1_transactions` table with status `Created`. We have
+eth_sender_sync that periodically picks up all entries from l1_transactions. It finds all that have status `Created` and
```
tomg10, line 30: this new 'component' will belong to server-v2, either as part of housekeeper or as a separate component. @RomanBrodetski do you have preference whether it needs to be separate? It will be relatively small piece of code with practically no config

core/node/eth_sender/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+Eth-sender is the component that executes scheduled transactions on Ethereum. As of 06.2024 it is only used by core to
+execute Commit, Prove and Execute L1 transactions.
+
+It was designed with fire-and-forget ideology in mind, as soon as you send transaction to be executed, eth-sender will
+make sure it is mined, handling stuff as bumping gas prices and resending transaction if needed.
+
+## API
+
+eth-sender API exposes only two methods:
```
tomg10, line 11: There will be also a second, Admin API usable via CLI to make common operational tasks less error-prone, but it will be designed separately

PR #2271, opened 2024-06-19
feat(tee_verifier_input_producer): use `FactoryDepsDal::get_factory_deps()

haraldh:
## What ❔

Use `FactoryDepsDal::get_factory_deps()` in the tee_verifier_input_producer crate.

## Why ❔

This optimizes getting the system contracts and gets rid of a couple of workarounds.

## Checklist


core/node/tee_verifier_input_producer/src/lib.rs
```diff
@@ -6,21 +6,20 @@
 //! For testing purposes, the L1 batch is re-executed immediately for now.
 //! Eventually, this component will only extract the inputs and send them to another
 //! machine over a "to be defined" channel, e.g., save them to an object store.
+#![feature(slice_flatten)]
```
popzxc, line 9: No nightly features are allowed in the core, we're only using nightly compiler because we are forced by our dependencies.

haraldh: so `get_factory_deps()` returns the bytecode in chunks of `Vec<[u8; 32]>`, but `pub fn store_factory_dep(&mut self, hash: H256, bytecode: Vec<u8>)` in `InMemoryStorage` wants flat byte vecs... maybe bytecode should get its own type... also the conversion between U256 and H256 all the time is really annoying.

popzxc: That's true, there are many places that can be improved in the codebase, it's mostly a question of having (no) resources to work on low-priority stuff. Not that it's not important, more that it works and we've got plenty of other areas where there are "hotter" issues.

haraldh: "fixed"

core/node/tee_verifier_input_producer/src/lib.rs
```diff
@@ -107,19 +102,43 @@ impl TeeVerifierInputProducer {
...
+                unsafe { (len.unchecked_mul(N), cap.unchecked_mul(N)) }
+            };
+            // SAFETY:
+            // - `ptr` was allocated by `data`
+            // - `ptr` is well-aligned because `[T; N]` has the same alignment as `T`.
+            // - `new_cap` refers to the same sized allocation as `cap` because
+            // `new_cap * size_of::<T>()` == `cap * size_of::<[T; N]>()`
+            // - `len` <= `cap`, so `len * N` <= `cap * N`.
+            unsafe { Vec::<T>::from_raw_parts(ptr.cast(), new_len, new_cap) }
+        }
```
popzxc, line 133: Ugh, sorry to still be that guy, but 😅 
We do not _fully_ forbid `unsafe`, but we try to use it in places where it would have a noticeable impact.
Here we have a function that is invoked once per batch (a few minutes on the mainnet currently), so the benefits of having `unsafe` instead of creating a second `Vec` and recollecting are not really clear.

popzxc: Though it's a nice and well-documented function that uses unsafe, if I've seen one 👍

haraldh: copy and pasted from the std lib :)

haraldh: fixed, but it hurts my feelings :)

popzxc (approved):
LGTM. cargo deny is failing for all the PRs right now, but it doesn't block merge.

PR #2276, opened 2024-06-19
fix: VM performance diff: don't show 0 as N/A

joonazan:
No difference in number of instructions executed now shows as 0 and N/A is only shown when data is missing for some reason.


core/tests/vm-benchmark/src/compare_iai_results.rs
```diff
@@ -32,35 +32,25 @@ fn main() {
...
-    for name in perf_changes
-        .keys()
-        .collect::<HashSet<_>>()
-        .union(&duration_changes.keys().collect())
-    {
+    for name in perf_changes.keys().collect::<HashSet<_>>().union(
+        &duration_changes
+            .iter()
+            .filter_map(|(key, value)| (*value != 0).then_some(key))
+            .collect(),
```
slowli, line 52: Nit: Can be extracted to a local var for readability.

slowli (approved):

PR #2283, opened 2024-06-19
feat: migrate ethers to alloy

aon:
## What ❔

- Migrates ethers to alloy, given new alloy 0.1 release

## Why ❔

- Ethers is deprecated

## Checklist


deniallugo-ml:

PR #2290, opened 2024-06-20
feat(eth-sender): Metrics of how many times transactions were resend / were in mempool 

tomg10:


RomanBrodetski:
didn't read the full PR, but can you please explain why? In theoery the number of records in history table is the same no?

perekopskiy (commented):

PR #2293, opened 2024-06-21
feat(node_framework): Support shutdown hooks + more

popzxc:
## What ❔

- Adds support for shutdown hooks. These are to be executed sequentially after all the tasks are either completed or dropped.
  - Note: I didn't spend too much time designing it, went with a "well enough" approach, as there are a ton of other things to do rn. We can revisit the design later, unless there are critical issues here.
  - One known caveat is that the hooks are not very reusable, and two tasks can add the same hook, and it will be executed twice. Not sure if it's an issue, given that the second execution would be a no-op.
- Moves waiting of rocksdb termination from state keeper and metadata calculator tasks to the hooks.
- Increases the amount of logs we emit.
- Adds task names to many service logs where they were missing.
- Collects all the errors that occurred in the framework.
- Improves handling of stop signals in queued job processor.

## Why ❔

- Performing the shutdown routine in the task itself is deadlock-prone.
- Added logs would've helped with identifying the issues we've already met.

## Checklist


core/node/node_framework/src/service/runnables.rs
```diff
@@ -7,9 +7,50 @@ use tokio::sync::Barrier;
 use super::StopReceiver;
 use crate::{
     precondition::Precondition,
-    task::{OneshotTask, Task, UnconstrainedOneshotTask, UnconstrainedTask},
+    task::{OneshotTask, Task, TaskId, UnconstrainedOneshotTask, UnconstrainedTask},
 };
 
+/// Alias for a shutdown hook function type.
+pub trait ShutdownHookFn:
```
slowli, line 14: Dumb question: Since Rust futures do nothing unless polled, wouldn't it suffice to use `BoxFuture<'static, anyhow::Result<()>>` as a shutdown hook? What benefit does enclosing it in a closure provide?

core/node/node_framework/src/service/mod.rs
```diff
@@ -190,13 +197,33 @@ impl ZkStackService {
...
+            // Limit each shutdown hook to the same timeout as the tasks.
+            let hook_with_timeout =
+                async move { tokio::time::timeout(TASK_SHUTDOWN_TIMEOUT, hook.invoke()).await };
+            match self.runtime.block_on(hook_with_timeout) {
+                Ok(Ok(())) => {
+                    tracing::info!("Shutdown hook {name} completed");
+                }
+                Ok(Err(err)) => {
+                    tracing::error!("Shutdown hook {name} failed: {err}");
+                    // We still have to invoke all the remaining hooks, so we don't return early.
```
slowli, line 219: Are errors in shutdown hooks not propagated intentionally? Before the node framework, e.g. failing to terminate RocksDB instances was propagated.

core/node/node_framework/src/service/runnables.rs
popzxc: It probably would, I just thought that expressing it as a closure provides less cognitive burden, e.g. it's easier to understand that the hook will be invoked later. Storing `Future` until some point in time is not a common operation, so I thought it might be less intuitive.

core/node/node_framework/src/service/mod.rs
popzxc: That's a good point. Initially when I wrote this code I thought "we're exiting anyways, so who cares", but indeed propagating the error would probably be a better approach. I guess I'll do it similarly to how task work -- the first occurred error will be propagated.

core/node/node_framework/src/service/runnables.rs
slowli: IMO, the interface should be as simple as possible, so a `Future` would work better than a closure returning a future. The fact that the future would be polled eventually can be documented; IMO, they are moderately easy to understand.

core/node/node_framework/src/service/mod.rs
slowli: I think it should make sense to return an error (one of the errors if several are returned, I suppose) provided that the main task logic didn't return an error.

popzxc: Yup, already in works.

core/node/node_framework/src/service/runnables.rs
popzxc: No objections.

core/lib/queued_job_processor/src/lib.rs
```diff
@@ -86,15 +86,18 @@ pub trait JobProcessor: Sync + Send {
...
             } else if iterations_left.is_some() {
                 tracing::info!("No more jobs to process. Server can stop now.");
                 return Ok(());
             } else {
                 tracing::trace!("Backing off for {} ms", backoff);
-                sleep(Duration::from_millis(backoff)).await;
+                // Error here corresponds to a timeout w/o `stop_receiver` changed; we're OK with this.
+                tokio::time::timeout(Duration::from_millis(backoff), stop_receiver.changed())
+                    .await
+                    .ok();
```
popzxc, line 100: Once timeouts became errors, it turned out that job processor doesn't really respect stop signals.
While this is not a full solution, I think it's 90% of solution, which is good enough for now.

core/node/node_framework/src/service/mod.rs
```diff
@@ -159,67 +163,128 @@ impl ZkStackService {
...
             .runtime
             .block_on(futures::future::select_all(join_handles));
-        let result = match resolved {
-            Ok(Ok(())) => Ok(()),
-            Ok(Err(err)) => Err(err).context("Task failed"),
+        // Extract the result and report it to logs early, before waiting for any other task to shutdown.
+        // We will also collect the errors from the remaining tasks, hence a vector.
+        let mut errors = Vec::new();
+        let task_name = tasks_names.swap_remove(resolved_idx);
+        match resolved {
```
popzxc, line 183: Here and below some refactoring is certainly required, but I would like to postpone it a bit.

core/node/node_framework/src/service/named_future.rs
```diff
@@ -0,0 +1,51 @@
...
+use crate::task::TaskId;
+
+pin_project! {
+    /// Implements a future with the name tag attached.
+    pub struct NamedFuture<F> {
+        #[pin]
+        inner: F,
+        name: TaskId,
+    }
+}
```
popzxc, line 14: Probably name propagation can be implemented in a cleaner way, it was the simplest option I could do quickly (and IMO still better than what we had before).

core/node/node_framework/src/service/mod.rs
```diff
@@ -159,67 +163,128 @@ impl ZkStackService {
...
+            Ok(Err(err)) => {
+                tracing::error!("Task {task_name} failed: {err}");
+                errors.push(TaskError::TaskFailed(task_name, err));
+            }
             Err(panic_err) => {
                 let panic_msg = try_extract_panic_message(panic_err);
-                Err(anyhow::format_err!(
-                    "One of the tasks panicked: {panic_msg}"
-                ))
+                tracing::error!("Task {task_name}: {panic_msg}");
```
slowli, line 193: Nit: `Task {task_name} panicked`?

core/node/node_framework/src/service/mod.rs
```diff
@@ -159,67 +163,128 @@ impl ZkStackService {
...
-            tracing::warn!(
-                "{execution_timeouts_count} tasks didn't finish in {TASK_SHUTDOWN_TIMEOUT:?} and were dropped"
-            );
-        } else {
-            tracing::info!("Remaining tasks finished without reaching timeouts");
+
+        // Report the results of the remaining tasks.
+        for (name, result) in remaining_tasks_names
+            .into_iter()
+            .zip(execution_results.into_iter())
```
slowli, line 216: Nit: `.into_iter()` is redundant; `zip()` [accepts](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.zip) `IntoIterator` argument.

core/node/node_framework/src/service/mod.rs
```diff
@@ -159,67 +163,128 @@ impl ZkStackService {
...
+                }
+                Ok(Ok(Err(err))) => {
+                    tracing::error!("Task {name} failed: {err}");
+                    errors.push(TaskError::TaskFailed(name, err));
+                }
+                Ok(Err(err)) => {
+                    let panic_msg = try_extract_panic_message(err);
+                    tracing::error!("Task {name} panicked: {panic_msg}");
+                    errors.push(TaskError::TaskPanicked(name, panic_msg));
+                }
```
slowli, line 230: This looks copy-pasted from similar task output handling above. Could it be possible to extract it into a function / method?

core/node/node_framework/src/service/named_future.rs
```diff
@@ -0,0 +1,51 @@
...
+
+impl<T, F> NamedFuture<F>
```
slowli, line 16: Dumb question: Why do you need `T` param here? It looks like it's possible to define `F: Future` w/o specifying its output type. Ditto below (you can use `type Output = F::Output`).

core/node/node_framework/src/service/named_future.rs
```diff
@@ -0,0 +1,51 @@
...
+
+impl<T, F> Future for NamedFuture<F>
+where
+    F: Future<Output = T>,
+{
+    type Output = T;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut task::Context<'_>) -> task::Poll<Self::Output> {
+        self.project().inner.poll(cx)
+    }
```
slowli, line 42: One thing possible to do here is to automatically instrument `NamedFuture`s with a tracing span (which will also be present in logs). Not sure whether it would add much value though.

core/node/node_framework/src/service/named_future.rs
```diff
@@ -0,0 +1,52 @@
...
+        Self { inner, name }
+    }
+
+    pub fn id(&self) -> TaskId {
+        self.name.clone()
+    }
+
+    pub fn into_inner(self) -> F {
+        self.inner
+    }
```
slowli, line 31: Bikeshedding: Since this is only used to spawn hooks onto a Tokio runtime, it could make sense to define a dedicated method instead:

```rust
impl<F> NamedFuture<F>
where
    F: Future + Send + 'static,
    F::Output: Send + 'static,
{
    pub fn spawn(self, handle: &tokio::runtime::Handle) -> NamedFuture<JoinHandle<F::Output>> {
        NamedFuture {
            name: self.name,
            inner: handle.spawn(self.inner),
        }
    }
}
```

slowli (approved):

core/node/node_framework/src/service/named_future.rs
popzxc: That makes sense, will do it in the next PR!

core/node/node_framework/src/service/mod.rs
```diff
@@ -1,7 +1,9 @@
 use std::{collections::HashMap, time::Duration};
 
 use anyhow::Context;
-use futures::{future::BoxFuture, FutureExt};
+use error::TaskError;
+use futures::FutureExt;
+use runnables::NamedBoxFuture;
```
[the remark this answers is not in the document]
popzxc: One more reason why IDE auto-imports are chaotic evil.

PR #2303, opened 2024-06-22
feat(zk_toolbox): Add chain name check

matias-gonz:
## What ❔

Add chain name check.

Chain name cannot contain '-' otherwise database migration will fail due to that special character using sqlx.


matias-gonz:
Closed in favor of #2309

PR #2304, opened 2024-06-23
feat(): tidy up base token oracle

thomas-nguy:
Changes: 

- Complete the flow by allowing to run the base-token-adjuster component to zk core builder if BaseTokenAdjuster component is specified in the component list. This will effectively run a component that update the exchange rate in DB at polling interval.

- Remove unnecessary `base_token` in config. we will use the common config `base_token_addr` in contract configs instead to avoid introducing redundancy

- fix DB query

### DIFF
--- a/core/lib/dal/src/base_token_dal.rs
+++ b/core/lib/dal/src/base_token_dal.rs
@@ -18,9 +18,9 @@ impl BaseTokenDal<'_, '_> {
         let row = sqlx::query!(
             r#"
             INSERT INTO
-                base_token_prices (numerator, denominator, ratio_timestamp)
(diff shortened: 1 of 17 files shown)


core/lib/dal/src/base_token_dal.rs
```diff
@@ -18,9 +18,9 @@ impl BaseTokenDal<'_, '_> {
         let row = sqlx::query!(
             r#"
             INSERT INTO
-                base_token_prices (numerator, denominator, ratio_timestamp)
+                base_token_prices (numerator, denominator, ratio_timestamp, created_at, updated_at)
```
thomas-nguy, line 21: Need update the .sqlx directory. Any idea how to do it @shahar4 ?

PR #2306, opened 2024-06-24
feat(): tidy up base token bare bones

thomas-nguy:
- Remove unnecessary base_token in config. we will use the common config base_token_addr in contract configs instead to avoid introducing redundancy 

- Rename some methods

- add some missing config file to make it work with zk init


etc/env/configs/dev.toml
```diff
@@ -1,3 +1,4 @@
 __imports__ = [ "base", "l1-inits/.init.env", "l2-inits/dev.init.env" ]
 
 ETH_SENDER_SENDER_PUBDATA_SENDING_MODE = "Blobs"
+CHAIN_STATE_KEEPER_FEE_MODEL_VERSION="V2"
```
thomas-nguy, line 4: no reason to use V1 anymore right?

PR #2311, opened 2024-06-24
docs(pruning): Improve pruning and snapshot recovery docs

slowli:
## What ❔

Improves various pruning and snapshot recovery docs.

## Why ❔

Makes docs more thorough and clearer.

## Checklist

### DIFF
--- a/docs/guides/external-node/08_pruning.md
+++ b/docs/guides/external-node/08_pruning.md
@@ -2,7 +2,7 @@
 
 It is possible to configure a ZKsync node to periodically prune all data from L1 batches older than a configurable
 threshold. Data is pruned both from Postgres and from tree (RocksDB). Pruning happens continuously (i.e., does not
-require stopping the node) in the background during normal node operation. It generally does not significantly influence
+require stopping the node) in the background during normal node operation. It is designed to not significantly impact
 node performance.
 
 Types of pruned data in Postgres include:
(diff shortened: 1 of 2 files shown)


docs/guides/external-node/08_pruning.md
```diff
@@ -1,13 +1,37 @@
...
+threshold. Data is pruned both from Postgres and from tree (RocksDB). Pruning happens continuously (i.e., does not
+require stopping the node) in the background during normal node operation. It generally does not significantly influence
+node performance.
 
-> [!NOTE]
->
-> If you need a node with data retention period of up to a few days, please set up a node from a
-> [_snapshot_](07_snapshots_recovery.md) and wait for it to have enough data. Pruning an archival node can take
-> unpractical amount of time. In the future we will be offering pre-pruned DB snapshots with a few months of data.
+Types of pruned data in Postgres include:
```
tomg10, line 8: I'd rephrase this so that we don't enumerate all types of entities that are pruned, do you think describing that no data from pruned batches is available is not enough?

docs/guides/external-node/08_pruning.md
```diff
@@ -1,13 +1,37 @@
 # Pruning
 
-It is possible to configure ZKsync Node to periodically remove all data from batches older than a configurable
-threshold. Data is pruned both from Postgres and from tree (RocksDB).
+It is possible to configure a ZKsync node to periodically prune all data from L1 batches older than a configurable
+threshold. Data is pruned both from Postgres and from tree (RocksDB). Pruning happens continuously (i.e., does not
+require stopping the node) in the background during normal node operation. It generally does not significantly influence
```
tomg10, line 5: I'd rephrase this to something like "It's designed to generally not influence node performance"

tomg10 (dismissed):
Good work! I left two minor comments

docs/guides/external-node/08_pruning.md
slowli: I think it makes sense to list main kinds of data that is getting pruned, because readers may not have good intuition regarding this.

tomg10: then at least write a TLDR of what is pruned, my first reaction to this list was something like "there are some kinds of data that are pruned and some that aren't" We'd also decrease the risk for this list to get out-of-date

tomg10 (approved):
You can probably dismiss my other comment, after longer thought it's fine as it is!

PR #2315, opened 2024-06-24
[DRAFT]: Add token price fetcher from coingecko

ischasny:
Add a component that fetches price using CoinGecko's [SimplePrice API](https://docs.coingecko.com/reference/simple-token-price). 

New price fetcher will be connected to the base token adjuster, once its implementation is stable.


deniallugo-ml:

PR #2316, opened 2024-06-24
feat(eth-sender): Make eth-sender tests use blob txs + refactor of eth-sender tests

tomg10:


core/lib/eth_client/src/clients/mock.rs
```diff
@@ -28,7 +28,16 @@ struct MockTx {
 }
 
 impl From<Vec<u8>> for MockTx {
-    fn from(tx: Vec<u8>) -> Self {
+    fn from(raw_tx: Vec<u8>) -> Self {
+        let is_eip4844 = raw_tx[0] == EIP_4844_TX_TYPE;
+        let tx: Vec<u8> = if is_eip4844 {
+            // decoding rlp-encoded length
+            let len = raw_tx[2] as usize * 256 + raw_tx[3] as usize - 2;
```
tomg10, line 35: I've tried to use rlp library for this, but couldn't make it work. This is an rlp-encoded list with 4 elements and I want to extract the first one of it. Do you know how we can decode this instead of hacking it this way?

core/node/eth_sender/src/eth_tx_manager.rs
```diff
@@ -377,54 +398,6 @@ impl EthTxManager {
         Ok(None)
     }
 
-    async fn send_unsent_txs(
```
tomg10, line 380: This method has been planned for removal for some time (get_unsent_txs doesn't return new transactions anymore because of setting sent_at_block unconditionally)

core/node/eth_sender/src/abstract_l1_interface.rs
```diff
@@ -50,9 +50,14 @@ pub(super) trait AbstractL1Interface: 'static + Sync + Send + fmt::Debug {
     async fn get_tx_status(
         &self,
         tx_hash: H256,
+        operator_type: OperatorType,
     ) -> Result<Option<ExecutedTxStatus>, EthSenderError>;
 
-    async fn send_raw_tx(&self, tx_bytes: RawTransactionBytes) -> EnrichedClientResult<H256>;
+    async fn send_raw_tx(
```
tomg10, line 56: new argument is added here to make tests easier to write, but overall this change would have been made when introducing gateway support anyway (gateway will be modelled as a new operator type)

core/lib/dal/src/eth_sender_dal.rs
```diff
@@ -56,6 +56,55 @@ impl EthSenderDal<'_, '_> {
         Ok(txs.into_iter().map(|tx| tx.into()).collect())
     }
 
+    pub async fn get_eth_txs_history_entries_max_id(&mut self) -> usize {
```
perekopskiy, line 59: nit: I think you should somehow mark `get_eth_txs_history_entries_max_id` and  `get_last_sent_eth_tx_hash` as methods that should be used only for tests, if I'm not mistaken we just do a separate `impl {}` block in other dals for this purpose with a comment "use these methods only for tests"

perekopskiy (approved):

core/lib/eth_client/src/clients/mock.rs
perekopskiy: Unfortunately I haven't used rlp lib. I think it's ok to leave it this way for mock client

core/lib/dal/src/eth_sender_dal.rs
tomg10: nice! I've moved it to a separate impl

perekopskiy (approved):

PR #2317, opened 2024-06-25
feat: add revert tests to zk_toolbox

aon:
## What ❔

- Adds revert tests to zk_toolbox

## Checklist


zk_toolbox/crates/zk_supervisor/src/commands/test/revert.rs
```diff
@@ -0,0 +1,40 @@
...
+fn run_test(
+    shell: &Shell,
+    args: &RevertAndRestartArgs,
+    ecosystem_config: &EcosystemConfig,
+) -> anyhow::Result<()> {
+    Spinner::new("Running test...").freeze();
+
+    let mut cmd = Cmd::new(cmd!(shell, "yarn mocha tests/revert-and-restart.test.ts"))
+        .env("CHAIN_NAME", &ecosystem_config.default_chain)
+        .env("ZKSYNC_HOME", &ecosystem_config.link_to_code);
```
Deniallugo, line 33: why do we need it? i've almost everywhere removed it

core/tests/revert-test/tests/revert-and-restart.test.ts
```diff
@@ -232,8 +279,8 @@ describe('Block reverting test', function () {
         await killServerAndWaitForShutdown(tester);
 
         // Run again.
-        utils.background(`zk server --components=${components}`, [null, logs, logs]);
-        await utils.sleep(10);
+        runServerInBackground({ components: [components], stdio: [null, logs, logs], cwd: pathToHome });
+        await utils.sleep(30);
```
Deniallugo, line 283: why?

zk_toolbox/crates/zk_inception/src/commands/args/run_server.rs
```diff
@@ -14,4 +15,6 @@ pub struct RunServerArgs {
     #[clap(long, short)]
     #[arg(trailing_var_arg = true, allow_hyphen_values = true, hide = false, help = MSG_SERVER_ADDITIONAL_ARGS_HELP)]
     additional_args: Vec<String>,
+    #[clap(long, help = MSG_SERVER_BUILD_HELP)]
+    pub build: bool,
```
Deniallugo, line 19: why do we need it? 
just cargo build normally enough

Deniallugo:
could we add the test to ci?

core/lib/protobuf_config/src/chain.rs
```diff
@@ -165,3 +166,60 @@ impl ProtoRepr for proto::Mempool {
...
+    }
+
+    fn build(this: &Self::Type) -> Self {
+        Self {
+            network: Some(proto::Network::new(&this.network).into()),
+            zksync_network: Some(this.zksync_network.clone()),
+            zksync_network_id: Some(this.zksync_network_id.as_u64()),
+        }
+    }
+}
```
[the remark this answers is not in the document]
aon: Are those the same values from genesis? If so we should deprecate the `NetworkConfig` struct.

zk_toolbox/crates/zk_supervisor/src/commands/test/revert.rs
aon: `CHAIN_NAME` allows the test to be run file based configs instead of env variables. Agreed that `ZKSYNC_HOME` can be removed.

core/tests/revert-test/tests/revert-and-restart.test.ts
aon: Sleep time was increased to make tests more reliable. With only 10 s of sleep time, tests sometime failed.

core/lib/protobuf_config/src/chain.rs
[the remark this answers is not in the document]
Deniallugo: yep. 
no problem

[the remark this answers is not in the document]
Deniallugo: network is `l1_chain_id`
`zksync_network` and `zksync_network_id` are equivalent and we could use `l2_chain_id`

zk_toolbox/crates/zk_inception/src/commands/args/run_server.rs
aon: The user could need to build the server before running it, in order to e.g. prepare for a CI run or just to have it ready for running. Therefore `zk_inception server --build` is useful. Although I'm now realizing that the code that actually built the server is not there, I'll add it again.

Deniallugo: Do we need it? 
I'm not strongly against but probably it's a bit of redundant

core/tests/revert-test/tests/revert-and-restart.test.ts
Deniallugo: hm, weird.. but it will increase the time of the test by it self

zk_toolbox/crates/zk_inception/src/commands/args/run_server.rs
aon: We do, otherwise revert tests fail due to timeout waiting for server to build.

core/lib/protobuf_config/src/chain.rs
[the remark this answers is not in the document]
aon: Changed, although I set `zksync_network` as `zksync_network_id.to_string()` because I couldn't find where else to get this information from. Would you like me to add it to any specific config file?

.github/workflows/ci-zk-toolbox-reusable.yml
```diff
@@ -95,8 +95,57 @@ jobs:
...
+          echo ZKSYNC_HOME=$(pwd) >> $GITHUB_ENV
+          echo $(pwd)/bin >> $GITHUB_PATH
+          echo IN_DOCKER=1 >> .env
+
+      - name: Start services
+        run: |
+          ci_localnet_up
+          ci_run sccache --start-server
+
+      - name: Initialize ecosystem
```
Deniallugo, line 135: I think we can just execute at the end of integration test

core/bin/block_reverter/src/main.rs
```diff
@@ -103,35 +122,111 @@ async fn main() -> anyhow::Result<()> {
...
+    let network_config = match &genesis_config {
+        Some(genesis_config) => NetworkConfig {
+            network: Network::from_chain_id(genesis_config.l1_chain_id),
+            zksync_network: genesis_config.l2_chain_id.as_u64().to_string(),
+            zksync_network_id: genesis_config.l2_chain_id,
+        },
+        None => NetworkConfig::from_env().context("NetworkConfig::from_env()")?,
+    };
+
+    let config = BlockReverterEthConfig::new(&eth_sender, &contracts, &network_config)?;
```
Deniallugo, line 218: could we reduce this dependency? we need only `zksync_network_id` in `BlockReverterEthConfig`

core/tests/revert-test/tests/revert-and-restart.test.ts
```diff
@@ -169,25 +223,27 @@ describe('Block reverting test', function () {
...
+            `cd ${pathToHome} && cargo run --bin block_reverter --release -- rollback-db --l1-batch-number ${lastL1BatchNumber} --rollback-postgres --rollback-tree --rollback-sk-cache ${fileConfigFlags}`
         );
 
         let blocksCommitted = await mainContract.getTotalBatchesCommitted();
         expect(blocksCommitted.eq(lastL1BatchNumber), 'Revert on contract was unsuccessful').to.be.true;
     });
 
     step('execute transaction after revert', async () => {
-        // Set 1 second deadline for `ExecuteBlocks` operation.
-        process.env.ETH_SENDER_SENDER_AGGREGATED_BLOCK_EXECUTE_DEADLINE = '1';
```
Deniallugo, line 186: you are not changing the files, right?

core/bin/block_reverter/src/main.rs
aon: In order to reduce the dependency, I have to manually try to load the `zksync_network_id` from env, which doesn't seem like the correct approach. With this criteria `NetworkConfig` was removed from `BlockReverterEthConfig` but not as a dependency for `block_reverter` bin:

https://github.com/matter-labs/zksync-era/blob/f3447ff358e1c0afe2eacd6fea1baff3cd7b55aa/core/bin/block_reverter/src/main.rs#L209-L218

core/tests/revert-test/tests/revert-and-restart.test.ts
Deniallugo: Looks like we should, but seems we don't have problems

Deniallugo (approved):

artmakh (approved):

PR #2318, opened 2024-06-25
feat(metadata-calculator): option to use VM runner for protective reads

itegulov:
## Checklist

### DIFF
--- a/core/bin/external_node/src/main.rs
+++ b/core/bin/external_node/src/main.rs
@@ -139,6 +139,7 @@ async fn run_tree(
             .merkle_tree_include_indices_and_filters_in_block_cache,
         memtable_capacity: config.optional.merkle_tree_memtable_capacity(),
         stalled_writes_timeout: config.optional.merkle_tree_stalled_writes_timeout(),
+        sealed_batches_have_protective_reads: config.optional.protective_reads_persistence_enabled,
         recovery: MetadataCalculatorRecoveryConfig {
             desired_chunk_size: config.experimental.snapshots_recovery_tree_chunk_size,
             parallel_persistence_buffer: config
--- a/core/bin/zksync_server/src/node_builder.rs
+++ b/core/bin/zksync_server/src/node_builder.rs
@@ -175,9 +175,11 @@ impl MainNodeBuilder {
         let merkle_tree_env_config = try_load_config!(self.configs.db_config).merkle_tree;
         let operations_manager_env_config =
             try_load_config!(self.configs.operations_manager_config);
+        let state_keeper_config = try_load_config!(self.configs.state_keeper_config);
         let metadata_calculator_config = MetadataCalculatorConfig::for_main_node(
             &merkle_tree_env_config,
             &operations_manager_env_config,
+            &state_keeper_config,
         );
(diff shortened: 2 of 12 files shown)


core/bin/external_node/src/main.rs
```diff
@@ -139,6 +139,7 @@ async fn run_tree(
             .merkle_tree_include_indices_and_filters_in_block_cache,
         memtable_capacity: config.optional.merkle_tree_memtable_capacity(),
         stalled_writes_timeout: config.optional.merkle_tree_stalled_writes_timeout(),
+        sealed_batches_have_protective_reads: config.optional.protective_reads_persistence_enabled,
```
itegulov, line 142: I am not 100% sure about this. My thinking was:
case 1) if EN is not persisting protective reads then it is also not running tree
case 2) if EN is persisting protective reads then we can run metadata calculator with `sealed_batches_have_protective_reads=true`

@perekopskiy could you double check this please?

core/node/metadata_calculator/src/updater.rs
```diff
@@ -184,28 +187,38 @@ impl TreeUpdater {
...
     ) -> anyhow::Result<()> {
-        let Some(last_sealed_l1_batch) = storage
-            .blocks_dal()
-            .get_sealed_l1_batch_number()
-            .await
-            .context("failed loading sealed L1 batch number")?
-        else {
-            tracing::trace!("No L1 batches to seal: Postgres storage is empty");
-            return Ok(());
+        let last_l1_batch_with_protective_reads = if self.sealed_batches_have_protective_reads {
```
perekopskiy, line 192: if `self.tree.mode()` is "light" then it should use `get_sealed_l1_batch_number` regardless config var

core/node/node_framework/examples/main_node.rs
```diff
@@ -135,9 +135,11 @@ impl MainNodeBuilder {
     fn add_metadata_calculator_layer(mut self) -> anyhow::Result<Self> {
         let merkle_tree_env_config = DBConfig::from_env()?.merkle_tree;
         let operations_manager_env_config = OperationsManagerConfig::from_env()?;
+        let state_keeper_config = StateKeeperConfig::from_env()?;
```
perekopskiy, line 138: nit: state_keeper_**env**_config for consistency

core/bin/zksync_server/src/node_builder.rs
```diff
@@ -175,9 +175,11 @@ impl MainNodeBuilder {
         let merkle_tree_env_config = try_load_config!(self.configs.db_config).merkle_tree;
         let operations_manager_env_config =
             try_load_config!(self.configs.operations_manager_config);
+        let state_keeper_config = try_load_config!(self.configs.state_keeper_config);
```
perekopskiy, line 178: nit: state_keeper_**env**_config for consistency

core/bin/external_node/src/main.rs
perekopskiy: Yes, correct. More precisely -- EN never runs full tree, so it doesn't matter what we set in `sealed_batches_have_protective_reads`

perekopskiy (approved):

PR #2319, opened 2024-06-25
feat(node_framework): Document implementations

popzxc:
## What ❔

Provides minimal documentation for the already present framework implementations.

## Why ❔

Accessibility.

## Checklist


Deniallugo (approved):

core/node/node_framework/src/implementations/resources/state_keeper.rs
```diff
@@ -24,6 +28,8 @@ impl Resource for BatchExecutorResource {
     }
 }
 
+/// A resource that provides [`OutputHandler`] implementation to the service.
```
AnastasiiaVashchuk, line 31: `OutputHandler ` is a struct and not a trait so it won't be correct to speak about implementation here

popzxc: Technically true, thought `OutputHandler` is just a wrapper over a set of `dyn StateKeeperOutputHandler`, so I guess this comment isn't very misleading.

PR #2322, opened 2024-06-25
feat(gas_adjuster): Use eth_feeHistory for both base fee and blobs

popzxc:
## What ❔

Updated the codebase to support blob information in the  `eth_feeHistory` RPC method.
Changes GasAdjuster so that it only uses this method to retrieve info.

## Why ❔

Use dedicated RPC method for getting info instead of custom implementation.
Less requests to L1.
Less code to maintain.

## Checklist

### DIFF
--- a/core/lib/basic_types/src/web3/mod.rs
+++ b/core/lib/basic_types/src/web3/mod.rs
@@ -827,6 +827,7 @@ pub enum TransactionCondition {
 }
 
 // `FeeHistory`: from `web3::types::fee_history`
+// Adapted to support blobs.
 
 /// The fee history type returned from `eth_feeHistory` call.
 #[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
@@ -834,14 +835,25 @@ pub enum TransactionCondition {
 pub struct FeeHistory {
     /// Lowest number block of the returned range.
     pub oldest_block: BlockNumber,
-    /// A vector of block base fees per gas. This includes the next block after the newest of the returned range, because this value can be derived from the newest block. Zeroes are returned for pre-EIP-1559 blocks.
+    /// A vector of block base fees per gas. This includes the next block after the newest of the returned range,
+    /// because this value can be derived from the newest block. Zeroes are returned for pre-EIP-1559 blocks.
     #[serde(default)] // some node implementations skip empty lists
     pub base_fee_per_gas: Vec<U256>,
     /// A vector of block gas used ratios. These are calculated as the ratio of gas used and gas limit.
     #[serde(default)] // some node implementations skip empty lists
     pub gas_used_ratio: Vec<f64>,
-    /// A vector of effective priority fee per gas data points from a single block. All zeroes are returned if the block is empty. Returned only if requested.
+    /// A vector of effective priority fee per gas data points from a single block. All zeroes are returned if
+    /// the block is empty. Returned only if requested.
     pub reward: Option<Vec<Vec<U256>>>,
+    /// An array of base fees per blob gas for blocks. This includes the next block following the newest in the
+    /// returned range, as this value can be derived from the latest block. For blocks before EIP-4844, zeroes
+    /// are returned.
+    #[serde(default)] // some node implementations skip empty lists
+    pub base_fee_per_blob_gas: Vec<U256>,
+    /// An array showing the ratios of blob gas used in blocks. These ratios are calculated by dividing blobGasUsed
+    /// by the maximum blob gas per block.
+    #[serde(default)] // some node implementations skip empty lists
+    pub blob_gas_used_ratio: Vec<f64>,
 }
 
 // `SyncInfo`, `SyncState`: from `web3::types::sync_state`
--- a/core/lib/eth_client/src/clients/http/query.rs
+++ b/core/lib/eth_client/src/clients/http/query.rs
@@ -117,7 +117,7 @@ where
             // prior to EIP-4844.
             // https://ethereum.github.io/execution-apis/api-documentation/
             if fee_history.base_fee_per_gas.len() != fee_history.base_fee_per_blob_gas.len() {
-                tracing::warn!(
+                tracing::error!(
                     "base_fee_per_gas and base_fee_per_blob_gas have different lengths: {} and {}",
                     fee_history.base_fee_per_gas.len(),
                     fee_history.base_fee_per_blob_gas.len()
--- a/core/lib/eth_client/src/clients/http/query.rs
+++ b/core/lib/eth_client/src/clients/http/query.rs
@@ -8,7 +8,7 @@ use zksync_web3_decl::error::{ClientRpcContext, EnrichedClientError, EnrichedCli
 use super::{decl::L1EthNamespaceClient, Method, COUNTERS, LATENCIES};
 use crate::{
     types::{ExecutedTxStatus, FailureInfo},
-    EthInterface, RawTransactionBytes,
+    BaseFees, EthInterface, RawTransactionBytes,
 };
 
 #[async_trait]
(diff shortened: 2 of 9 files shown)


core/node/api_server/src/web3/namespaces/eth.rs
```diff
@@ -695,6 +695,8 @@ impl EthNamespace {
             base_fee_per_gas,
             gas_used_ratio,
             reward,
+            base_fee_per_blob_gas: vec![],
+            blob_gas_used_ratio: vec![],
```
popzxc, line 699: We are adding two new fields to the `eth_feeHistory` response on our API as well.
This is an implication of using the same types both internally and for API.

Is this OK? I would think so, given that adding new fields is not breaking, and this actually makes our API more compatible with L1.

core/lib/basic_types/src/web3/mod.rs
```diff
@@ -827,21 +827,33 @@ pub enum TransactionCondition {
...
     pub gas_used_ratio: Vec<f64>,
-    /// A vector of effective priority fee per gas data points from a single block. All zeroes are returned if the block is empty. Returned only if requested.
+    /// A vector of effective priority fee per gas data points from a single block. All zeroes are returned if
+    /// the block is empty. Returned only if requested.
     pub reward: Option<Vec<Vec<U256>>>,
+    /// An array of base fees per blob gas for blocks. This includes the next block following the newest in the
+    /// returned range, as this value can be derived from the latest block. For blocks before EIP-4844, zeroes
+    /// are returned.
+    #[serde(default)] // some node implementations skip empty lists
+    pub base_fee_per_blob_gas: Vec<U256>,
```
Deniallugo, line 852: probably we have to skip serialization

popzxc: Do we? I touched the [same topic above](https://github.com/matter-labs/zksync-era/pull/2322#discussion_r1652527714). IMHO, it increases our compatibility with L1.

core/lib/eth_client/src/clients/mock.rs
```diff
@@ -591,10 +575,25 @@ mod tests {
     use super::*;
     use crate::{CallFunctionArgs, EthInterface};
 
+    macro_rules! base_fees {
```
slowli, line 578: Nit: Why can't this be a function? Functions are more maintainable than macros.

core/lib/eth_client/src/clients/http/query.rs
```diff
@@ -78,14 +78,22 @@ where
...
+    ) -> EnrichedClientResult<Vec<BaseFees>> {
         const MAX_REQUEST_CHUNK: usize = 1024;
 
         COUNTERS.call[&(Method::BaseFeeHistory, self.component())].inc();
         let latency = LATENCIES.direct[&Method::BaseFeeHistory].start();
         let mut history = Vec::with_capacity(block_count);
         let from_block = upto_block.saturating_sub(block_count);
 
+        // Non-panicking conversion to u64.
+        fn cast_to_u64(value: U256, tag: &str) -> EnrichedClientResult<u64> {
```
slowli, line 90: Bikeshedding: Function definitions in the middle of a function look not very readable. IMO, it would make sense to move it at the start of the method body.

core/lib/eth_client/src/clients/http/query.rs
```diff
@@ -103,11 +111,22 @@ where
                 .with_arg("chunk_size", &chunk_size)
                 .with_arg("block", &chunk_end)
                 .await?;
-            history.extend(fee_history.base_fee_per_gas);
+
+            for (base, blob) in fee_history
+                .base_fee_per_gas
+                .into_iter()
+                .zip(fee_history.base_fee_per_blob_gas)
```
slowli, line 118: Dumb question: Do all Web3 providers support these new fields? If they are not returned, this field here will be empty, so that `history` will not be filled either, even if `fee_history.base_fee_per_gas` is filled.

core/node/fee_model/src/l1_gas_price/gas_adjuster/tests.rs
```diff
@@ -32,97 +32,92 @@ fn samples_queue() {
...
-            393216,
-            393216 * 2,
-            393216,
-            393216 * 2,
-            393216 * 3,
-            393216 * 4,
-        ])
-        .build();
+    // Helper macro to read a value from adjuster
+    macro_rules! read {
```
slowli, line 36: Nit: Similar to the other line comment, I *think* this can be expressed as a generic function as well.

core/node/state_keeper/src/io/tests/tester.rs
```diff
@@ -47,9 +47,16 @@ impl Tester {
...
-        let eth_client = MockEthereum::builder()
-            .with_fee_history(vec![0, 4, 6, 8, 7, 5, 5, 8, 10, 9])
-            .build();
+        let block_fees = vec![0, 4, 6, 8, 7, 5, 5, 8, 10, 9];
+        let base_fees = block_fees
+            .into_iter()
+            .zip(std::iter::repeat(1))
+            .map(|(block, blob)| BaseFees {
+                base_fee_per_gas: block,
+                base_fee_per_blob_gas: blob.into(),
```
slowli, line 56: Nit: Can't `1.into()` be inlined here?

core/lib/eth_client/src/clients/http/query.rs
popzxc: I've checked Infura, Alchemy and `reth` -- all seem to support it.

core/lib/eth_client/src/clients/mock.rs
popzxc: Personal preference, IMHO short macros in tests are perfectly readable.
No strong opinion, will change.

perekopskiy (dismissed):

core/node/api_server/src/web3/namespaces/eth.rs
Deniallugo: that's what I meant in my initial comment, we could skip the serialization for them as a result we will not change the behaviour of our api

Deniallugo (dismissed):

core/lib/eth_client/src/clients/http/query.rs
popzxc: Actually, it seems that it's [a part of specification now](https://ethereum.github.io/execution-apis/api-documentation/).

slowli: Got it. IMO, it'd still be safer to fill these value with zeros at least if `base_fee_per_blob_gas` is empty (don't know what to do if it's non-empty, but shorter than `base_fee_per_gas`; maybe return an error). OTOH, this may be an unnecessary complication, so feel free to ignore.

core/lib/eth_client/src/lib.rs
```diff
@@ -65,6 +65,13 @@ impl Options {
     }
 }
 
+/// Information about the base fees provided by the L1 client.
+#[derive(Debug, Clone, PartialEq, Eq, Default)]
+pub struct BaseFees {
+    pub base_fee_per_gas: u64,
+    pub base_fee_per_blob_gas: U256,
```
slowli, line 72: Bikeshedding: Discrepancy between types of these fields looks weird; it could make sense to define both as `U256`.

core/lib/eth_client/src/clients/http/query.rs
popzxc: I've added a `tracing::error` for cases where the lengths of returned arrays are not equal.
Filling with zeros doesn't sound right -- we rely on these values in production, and using zeros silently will cause blob transactions to stuck probably.
Returning an error doesn't feel right either -- if occurs, it will probably either stop or crash many components.

That said, I've checked a few more providers and they do seem to support it; moreover they specify that the values returned must be 0 for pre EIP-4844 blocks (I've also adapted our API to match this behavior).

core/node/fee_model/src/l1_gas_price/gas_adjuster/mod.rs
```diff
@@ -95,25 +93,29 @@ impl GasAdjuster {
...
-            if let Some(current_blob_base_fee) = blob_base_fee_history.last() {
+            if let Some(current_blob_base_fee) =
+                base_fees.last().map(|fee| fee.base_fee_per_blob_gas)
+            {
                 // Blob base fee overflows `u64` only in very extreme cases.
                 // It doesn't worth to observe exact value with metric because anyway values that can be used
                 // are capped by `self.config.max_blob_base_fee()` of `u64` type.
-                if current_blob_base_fee > &U256::from(u64::MAX) {
+                if current_blob_base_fee > U256::from(u64::MAX) {
                     tracing::error!("Failed to report current_blob_base_fee = {current_blob_base_fee}, it exceeds u64::MAX");
```
slowli, line 119: I mean, it's possible to redefine the metric as `f64` if we really want to. Not sure whether it makes sense from the semantic standpoint though.

slowli (approved):

core/lib/eth_client/src/lib.rs
popzxc: My reasoning here was that `BaseFees` is an internal interface, and internally we only treat `base_fee_per_gas` as `u64`, and `base_fee_per_blob_gas` as `U256`.
AFAIR treating gas as `u64` is considered a problem (it has issues, for example, with custom base tokens), but fixing it is not scheduled just yet -- so I decided to be consistent here and not introduce more type variance across components.

core/node/fee_model/src/l1_gas_price/gas_adjuster/mod.rs
popzxc: Up to @RomanBrodetski 
I personally think that if the blob price will reach `u64::MAX`, we'll know it anyway 😅

PR #2326, opened 2024-06-25
fix: limit validation gas using the normal out of gas mechanism

joonazan:
Account validation has a gas limit that behaves differently from normal gas limits.
That is overly tricky and very hard to replicate in vm2. It also already has a bug because new functionality that uses gas wasn't added to the gas counting code.
This PR utilizes the normal gas limit by temporarily taking away gas and giving it back after validation has ended.


joonazan:
We decided to implement this kind of solution for vm2 but keep vm_latest as-is.

PR #2330, opened 2024-06-26
feat(node_framework): Implement FromContext and IntoContext derive macro

popzxc:
## What ❔

- Improves context interfaces: removes unneded `async` when fetching resources, and made `add_task` accept `T` instead of `Box`.
- Adds previously proposed `FromContext` and `IntoContext` macro that will allow to redefine the wiring layer interface so that it doesn't have direct access to the context.

⚠️ I didn't port the whole framework to using macros, since there may be changes in the macro itself during this review. Once we merge this, I will simultaneously rework the `WiringLayer` interface and will port layers. For now, I've used new macros in the `BatchStatusUpdater` layer to see how it works. Seems that it works fine 😅 

## Why ❔

Ergonomics.

## Checklist


core/lib/node_framework_derive/src/lib.rs
```diff
@@ -0,0 +1,129 @@
...
+/// #     fn from_context(context: &mut service::ServiceContext<'_>) -> Result<Self, WiringError>;
+/// # }
+/// # struct MandatoryResource;
+/// # struct OptionalResource;
+/// # struct ResourceWithDefault;
+/// #[derive(FromContext)]
+/// # #[ctx(local)]
+/// struct MyWiringLayerInput {
+///     // The following field _must_ be present in the context.
+///     mandatory_resource: MandatoryResource,
```
slowli, line 45: I think this approach can be tweaked as follows:

- Require that all fields for derived `FromContext` / `IntoContext` implementations implement the corresponding trait itself. I.e., make implementations just delegate to this impls similar to `serde` traits.
- Implement `FromContext` / `IntoContext` for the relevant types.

```rust
impl<T: Resource + Clone> FromContext for T {
    fn from_context(context: &mut ServiceContext<'_>) -> Result<Self, WiringError> {
        context.get_resource::<T>()
    }
}

impl<T: FromContext> FromContext for Option<T> {
    fn from_context(context: &mut ServiceContext<'_>) -> Result<Self, WiringError> {
        Ok(match T::from_context(context) {
            Ok(resource) => Some(resource),
            Err(WiringError::ResourceLacking { .. }) => None,
            Err(err) => return Err(err),
        })
    }
}

mpl<T: Resource> IntoContext for T {
    fn into_context(self, context: &mut ServiceContext<'_>) -> Result<(), WiringError> {
        context.insert_resource(self)
    }
}

impl<T: IntoContext> IntoContext for Option<T> {
    fn into_context(self, context: &mut ServiceContext<'_>) -> Result<(), WiringError> {
        if let Some(this) = self {
            this.into_context(context)
        } else {
            Ok(())
        }
    }
}

// Wrapper to distinguish tasks from resources.
#[derive(Debug)]
pub struct ContextTask<T>(pub T);

impl<T: Task> IntoContext for ContextTask<T> {
    fn into_context(self, context: &mut ServiceContext<'_>) -> Result<(), WiringError> {
        let Self(task) = self;
        context.add_task(task);
        Ok(())
    }
}
```

The only thing that it doesn't cover is the "or default" extractor for `FromContext`. It is possible to implement via a wrapper like this:

```rust
#[derive(Debug)]
pub struct OrDefault<T>(pub T);

impl<T: Resource + Clone + Default> FromContext for OrDefault<T> {
    fn from_context(context: &mut ServiceContext<'_>) -> Result<Self, WiringError> {
        Ok(Self(context.get_resource_or_default::<T>()))
    }
}
```

...but I feel it would be unnecessarily verbose. If we assume that wrappers always have form `Wrapper<T>(pub T)`, then it's possible to introduce their support in derive macros like this:

```rust
#[derive(FromContext)]
struct MyContext {
    #[with = OrDefault]
    app_health: AppHealthCheckResource,
}

// expands to
impl FromContext for MyContext {
    fn from_context(context: &mut ServiceContext<'_>) -> Result<Self, WiringError> {
        Ok(Self {
            app_health: <OrDefault<_> as FromContext>::from_context(context)?.0,
        })
    }
}
```

Alternatively, use `serde`-inspired functional extractors:

```rust
pub fn get_or_default<T>(context: &mut ServiceContext<'_>) -> Result<T, WiringError>
where
    T: Resource + Clone + Default,
{
    Ok(context.get_resource_or_default::<T>())
}

#[derive(FromContext)]
struct MyContext {
    #[with = get_or_default]
    // ^ accepts a path to a function w/ corresponding signature (may be generic, as in this case)
    app_health: AppHealthCheckResource,
}

// expands into
impl FromContext for MyContext {
    fn from_context(context: &mut ServiceContext<'_>) -> Result<Self, WiringError> {
        Ok(Self {
            app_health: or_default(context)?,
        })
    }
}

```

I'm eager to do this in a follow-up PR to not block your efforts / to expedite, but IMO it would be more idiomatic / modular.

core/lib/node_framework_derive/tests/ui.rs
```diff
@@ -0,0 +1,11 @@
+#[test]
+fn ui_pass() {
```
slowli, line 2: Bikeshedding: Could make sense to place UI tests in the `node_framework` crate to test that they work with real traits. Also, wouldn't require to keep mock definitions of these traits for every test, which could potentially get outdated.

slowli (commented):
Future work (?): One of the benefits of derive macros is being able to add context to errors in a uniform way (i.e., on a wiring error, provide information in which field for which struct it happened). Can be dealt with later, of course.

core/lib/node_framework_derive/tests/ui.rs
popzxc: My concern here is that AFAIU `trybuild` needs to build & link every test, and `zksync_node_framework` currently depends on most of our dependency tree, so it would be pretty slow.
It certainly makes sense to move tests there, but maybe it would be better to do once we move the implementations out of the framework crate (going to happen right after we rework `WiringLayer` & finish EN support)? I can create a task for that.

core/lib/node_framework_derive/src/lib.rs
popzxc: Thanks for the thoughtful comment!
Funnily, I initially started with an approach close to [what you propose](https://github.com/matter-labs/zksync-era/pull/2330/commits/103139859ba09803175832d024c85370f41db90d), but then met the following problems:

First, we still will need wrapper types exposed to people `ContextTask`, in your case, and it feels unnecessary. I don't want the system to be designed around the procedural macro, I want it to be a mere tool to reduce boilerplate. And it leads to some clunkiness, where `Option<ContextTask<T>>` works, but `ContextTask<Option<T>>` in a compile error related to the trait implementations that may confuse inexperienced developers.

Then, we're building a somewhat complex infrastructure around these traits, allowing extensibility, but at the same time making it harder to understand. My point of view is that it's easy to slap either `task` or `resource` on fields, and it's clear what they do. Supporting `Option` _feels_ natural, IMHO. Same for `default` parameter. On the other hand, generic `with` support allows developer to do `#[with = my_custom_getter]`, which again will increase the complexity, because now you have to go and check what this function does, and it can in theory get very messy if two such modifiers will be applied and the behavior will become dependent on the order of fields in the structure.

What I wanted to achieve here is an explicitly restricted macro, that is easy to use, but _at the same time_ as soon as macro does not work for you anymore, you either redesign your code (getting and putting stuff from context shouldn't normally be complex), you just go and implement a trait manually.

I see the point regarding inconsistency, though, and we may discuss it further to try to find an optimal solution.

popzxc: tl;dr the problem domain feels very narrow to me, and I want to avoid overgeneralization just for the sake of overgeneralization.

core/lib/node_framework_derive/tests/ui.rs
popzxc: Yup, the first run is very slow, but re-runs are fast. I guess we can live with it for a while.
Moved it there.

core/lib/node_framework_derive/src/lib.rs
popzxc: We discussed the design in DMs a bit, and I reworked the macro according to the results of the discussion.

core/lib/node_framework_derive/src/macro_impl.rs
```diff
@@ -0,0 +1,190 @@
...
+    fn render_from_context(self) -> Result<proc_macro2::TokenStream> {
+        let crate_path = self.crate_path();
+        let ident = self.ident;
+        let mut fields = Vec::new();
+        for field in self.fields {
+            let ty = field.ty;
+            let ident = field.ident;
+            let default = field.label.default;
+
+            if field.label.krate.is_some() {
```
slowli, line 101: Bikeshedding: IMO, it would be beneficial to have separate struct or at least parsers for differing contexts (despite them all using `#[context(_)]` wrapper. Using a single parser and post-checking fields like this looks not idiomatic.

core/lib/node_framework_derive/src/labels.rs
```diff
@@ -0,0 +1,98 @@
...
+                }
+                attr.parse_nested_meta(|meta| {
+                    let mut added = false;
+                    for &label in Self::LABELS {
+                        if meta.path.is_ident(label) {
+                            match label {
+                                Self::CRATE_LABEL => {
+                                    let value = meta.value()?;
+                                    let lit: syn::LitStr = value.parse()?; // Check that the value is a path.
+                                    self_.krate = Some(syn::parse_str(&lit.value())?);
```
slowli, line 68: Bikeshedding: You can parse `syn::Path` here directly. `serde` and some other crates use `LitStr` wrapper because of legacy reasons (AFAIU, really old Rust version had very limited attribute values support).

core/node/node_framework/src/service/context_traits.rs
```diff
@@ -0,0 +1,133 @@
...
+    }
+}
+
+impl FromContext for () {
+    fn from_context(_context: &mut ServiceContext<'_>) -> Result<Self, WiringError> {
+        Ok(())
+    }
+}
+
+impl<T: Resource + Clone> FromContext for Option<T> {
```
slowli, line 50: Bikeshedding: Wouldn't it be possible to generalize this implementation to `T: FromContext` (ditto for `impl IntoContext for Option<T>` below)? It would make these traits more composable.

core/lib/node_framework_derive/src/labels.rs
```diff
@@ -0,0 +1,98 @@
...
+    pub(crate) krate: Option<syn::Path>, // `crate` is a reserved keyword and cannot be a raw identifier.
+    pub(crate) span: Option<proc_macro2::Span>,
+    pub(crate) task: bool,
+    pub(crate) default: bool,
+}
+
+impl fmt::Debug for CtxLabel {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // For some weird reason, doc tests fail with the derived impl, stating that
+        // `syn::Path` does not implement `Debug`.
```
slowli, line 19: That's because `Debug` is implemented for `syn::Path` only if [the `extra-traits` crate feature](https://docs.rs/syn/latest/syn/#optional-features) is enabled. Most probably, this feature is enabled indirectly in the workspace, but not when compiling doc tests for a specific crate.

PR #2331, opened 2024-06-26
fix(object-store): Consider some token source errors transient

slowli:
## What ❔

Considers some token source GCS errors transient.

## Why ❔

Considering errors as transient leads to less abnormal application terminations.

## Checklist


popzxc (approved):
Nice

PR #2332, opened 2024-06-26
feat: Adding unstable RPC endpoint to return the execution_info

mm-zk:
## What ❔

* Execution info shows some details from the VM execution (like circutits used, pubdata etc)
* This data was only stored in DB and not accessible outside - after this PR, it is available under `nstable_getTransactionExecutionInfo`

## Why ❔

* This allows us to do more advanced debugging of issues - especially for cases where we might not have access to the underlying database.
* In the future, some parts of this might be migrated into a 'stable' RPC.

## Evidence
image


core/lib/dal/src/models/storage_transaction.rs
```diff
@@ -396,6 +397,13 @@ impl From<StorageTransactionReceipt> for TransactionReceipt {
     }
 }
 
+// Details of the transaction execution.
+#[derive(Debug, Clone, sqlx::FromRow)]
+pub struct StorageTransactionExecutionInfo {
+    // This is an opaque JSON field, with VM version specific contents.
```
popzxc, line 403: Nit: these should be doc-comments (`///`).

core/lib/dal/src/transactions_web3_dal.rs
```diff
@@ -550,6 +573,19 @@ mod tests {
             .get_transaction_by_hash(H256::zero(), L2ChainId::from(270))
             .await;
         assert!(web3_tx.unwrap().is_none());
+
+        let execution_info = conn
+            .transactions_web3_dal()
+            .get_unstable_transaction_execution_info(tx_hash)
+            .await
+            .unwrap()
+            .unwrap();
```
popzxc, line 582: I guess this `unwrap` should be `.expect("Transaction execution info is missing in the DAL")` (normally we `unwrap` only in the places that are assumed to always hold, while here it's a part of the test).

core/lib/dal/src/transactions_web3_dal.rs
```diff
@@ -550,6 +573,19 @@ mod tests {
...
+
+        // Check that execution info has at least the circuit statistics field.
```
popzxc, line 584: Maybe add: "If this assertion fails because the transaction execution info format has changed, replace `circuit_statistic` with any other valid field".

core/lib/types/src/api/mod.rs
```diff
@@ -821,6 +822,13 @@ pub struct ApiStorageLog {
     pub written_value: U256,
 }
 
+// Coming from TransactionExecutionMetrics
```
popzxc, line 825:
```suggestion
/// Raw transaction execution data.
/// Data is taken from `TransactionExecutionMetrics`.
```

popzxc (approved):
Mostly looks good.
I'm not sure who currently maintains our docs, but probably we should mention there that while `unstable` namespace exists, no applications should rely on it, as the methods there may be changed or removed at any time.

core/lib/dal/src/transactions_web3_dal.rs
```diff
@@ -151,6 +151,29 @@ impl TransactionsWeb3Dal<'_, '_> {
             .await
     }
 
+    pub async fn get_unstable_transaction_execution_info(
+        &mut self,
+        hash: H256,
+    ) -> DalResult<Option<StorageTransactionExecutionInfo>> {
```
slowli, line 157: Bikeshedding: I'd try not to return storage models from public API (a bit questionable here given that this API is marked as unstable). Instead, it could make sense to return an `api` type (most of the methods in `_Web3Dal` do that), or `serde_json::Value`.

slowli (commented):
Overall looks good 👍

core/node/api_server/src/web3/mod.rs
```diff
@@ -108,6 +110,7 @@ impl Namespace {
         Self::Zks,
         Self::En,
         Self::Pubsub,
+        Self::Unstable,
```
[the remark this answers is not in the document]
mm-zk: which config parameter?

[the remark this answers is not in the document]
popzxc: [This one](https://github.com/matter-labs/zksync-era/blob/287958db6ca54959fd56c04d4a7a3cbfc9baa877/core/lib/config/src/configs/api.rs#L218)

popzxc (approved):

PR #2338, opened 2024-06-27
feat(snapshots_applier): Add a method to check whether snapshot recovery is done

popzxc:
## What ❔

subj

## Why ❔

Such a method is useful, for example, to understand whether it's safe to start the API server in distributed mode.
Will be a part of EN check for storage initialization.

## Checklist


core/lib/snapshots_applier/src/lib.rs
```diff
@@ -261,6 +261,41 @@ impl SnapshotsApplierTask {
...
+        if applied_snapshot_status.storage_logs_chunks_left_to_process() != 0 {
+            return Ok(Some(false));
+        }
+        // Currently, migrating tokens is the last step of the recovery.
+        // The number of tokens is not a part of the snapshot header, so we have to re-query the main node.
+        let added_tokens = conn.tokens_dal().get_all_l2_token_addresses().await?.len();
+        let tokens_on_main_node = client
+            .fetch_tokens(applied_snapshot_status.l2_block_number)
+            .await?
+            .len();
```
popzxc, line 290: It's a bit wasteful, but I think it's OK, given that we will only reach this point _after_ we made sure that all the storage logs have been applied.
I don't feel like using a static `OnceCell` for caching here, and introducing a dedicated structure feels a bit like overkill.

popzxc: Also, this check is to be invoked if there are no blocks in the storage, so realistically we will invoke this method 1-2 times per actual snapshot recovery.

core/lib/snapshots_applier/src/lib.rs
```diff
@@ -261,6 +261,41 @@ impl SnapshotsApplierTask {
         }
     }
 
+    /// Checks whether the snapshot recovery is already completed.
+    ///
+    /// Returns `None` if no snapshot recovery information is detected in the DB.
+    /// Returns `Some(true)` if the recovery is completed.
+    /// Returns `Some(false)` if the recovery is not completed.
```
slowli, line 268: Bikeshedding: IMO, it could make sense to introduce a dedicated enum with 3 variants instead here.

core/lib/snapshots_applier/src/lib.rs
```diff
@@ -261,6 +261,41 @@ impl SnapshotsApplierTask {
...
+        else {
+            return Ok(None);
+        };
+        // If there are unprocessed storage logs chunks, the recovery is not complete.
+        if applied_snapshot_status.storage_logs_chunks_left_to_process() != 0 {
+            return Ok(Some(false));
+        }
+        // Currently, migrating tokens is the last step of the recovery.
+        // The number of tokens is not a part of the snapshot header, so we have to re-query the main node.
+        let added_tokens = conn.tokens_dal().get_all_l2_token_addresses().await?.len();
```
slowli, line 286: Can't this number be greater than the number of recovered tokens if recovery was completed and the node has processed some blocks that introduce new tokens? Could make sense to use `TokensWeb3Dal::get_all_tokens()`, which is used in the `fetch_tokens` server handler.

core/lib/snapshots_applier/src/tests/mod.rs
```diff
@@ -215,6 +238,12 @@ async fn snapshot_applier_recovers_after_stopping() {
     assert!(!task_handle.is_finished());
     task_handle.abort();
 
+    assert_eq!(
+        is_recovery_completed(&pool, &client).await,
+        Some(false),
+        "Recovery has been abrupted"
```
slowli, line 244: Grammar: can "abrupt" be used as a verb? 🤔 Maybe you've meant "aborted"?

core/lib/snapshots_applier/src/tests/utils.rs
```diff
@@ -18,16 +22,44 @@ use zksync_types::{
...
+    pub(super) fn set_token_response_error(&self, error: EnrichedClientError) {
+        *self.tokens_response_error.write().unwrap() = Some(error);
+    }
+
+    fn take_token_response_error(&self) -> Option<EnrichedClientError> {
+        self.tokens_response_error.write().unwrap().take()
+    }
+}
+
+impl fmt::Debug for MockMainNodeClient {
```
slowli, line 50: Dumb question: Can't `Debug` still be derived?

core/lib/snapshots_applier/src/lib.rs
popzxc: I thought about it too, but IMHO this is not ambiguous and given that the method has very limited usage, it's OK to leave it as is.

core/lib/snapshots_applier/src/lib.rs
popzxc: Great catch, thanks!

core/lib/snapshots_applier/src/tests/utils.rs
popzxc: It can! Artifact from initial iterations.

core/lib/snapshots_applier/src/lib.rs
popzxc: OTOH no strong opinion, so let's add it, why not.

tomg10 (approved):

PR #2340, opened 2024-06-27
feat: L1 batch QC database (BFT-476)

aakoshh:
## What ❔

- [x] Add an `l1_batches_consensus` table to hold [L1 batch Quorum Certificates](https://github.com/matter-labs/era-consensus/blob/177881457f392fca990dbb3df1695737d90fd0c7/node/libs/roles/src/attester/messages/batch.rs#L67) from Attesters
- [x] Add attesters to the config
- [x] Implement methods in `PersistentBatchStore` 
  - [x] `persisted`
  - [x] `last_batch`
  - [x] `last_batch_qc`
  - [x] `get_batch`
  - [x] `get_batch_qc`
  - [x] `store_qc`
  - [ ] `queue_next_batch` - _not going to implement for now_
  - [ ] assign `SyncBatch::proof` - _not going to implement for now_
- [x] Add tests for all new methods in `ConsensusDal` and the `PersistentBatchStore`

### Caveat

Implemented the updating of `persisted` with a loop that polls the database for newly available `SyncBatch` records, even if they have no proof. This inevitably triggers the gossiping of batch statuses and the pulling of `SyncBatch` between peers. For this reason `queue_next_batch` just drop the data, since we can't do anything with it without the proof yet. Returning an error or panicking would stop the consensus tasks. 

I ended up disabling the `persisted` by leaving its dummy implementation in place because when enabled the full node tests keep going on forever, printing the following logs in a loop:

```console
❯ RUST_LOG=info zk test rust test_full_nodes --no-capture
...
2024-07-03T14:22:57.882784Z  INFO in{addr=[::1]:53082}: zksync_consensus_network: 191: new connection
2024-07-03T14:22:57.883457Z  INFO in{addr=[::1]:53082}:gossip: zksync_consensus_network::gossip::runner: 383: peer = node:public:ed25519:068ffa0b3fedbbe5c2a6da3defd26e0d084248f12bfe98db85f7785b0b08b63e
2024-07-03T14:22:57.883764Z  INFO out{addr="[::1]:52998"}:gossip: zksync_consensus_network::gossip::runner: 416: peer = node:public:ed25519:7710ed90aad9f5859dfba06e13fb4e6fb0fe4d686f81f9d819464ad1fdc371bd
2024-07-03T14:22:57.886204Z  INFO in{addr=[::1]:53082}:gossip: zksync_consensus_network::rpc: 222: message too large: max = 10240B, got 13773B
2024-07-03T14:22:57.886280Z  INFO out{addr="[::1]:52998"}:gossip: zksync_consensus_network::rpc: 222: message too large: max = 10240B, got 13773B
2024-07-03T14:22:57.886633Z  INFO in{addr=[::1]:53082}:gossip: zksync_consensus_network::rpc: 222: canceled
...
2024-07-03T14:22:57.888143Z  INFO out{addr="[::1]:52998"}:gossip: zksync_consensus_network::rpc: 222: disconnected
...
2024-07-03T14:22:57.888390Z  INFO zksync_consensus_network: 216: [::1]:53082: gossip.run_inbound_stream(): push_batch_store_state.: end of stream
2024-07-03T14:22:57.888446Z  INFO zksync_consensus_network: 158: gossip.run_outbound_stream("[::1]:52998"): push_batch_store_state.: end of stream
```

So in the tests the message size exceeds the maximum. I think it's [hardcoded here](https://github.com/matter-labs/era-consensus/blob/decb988eb9e1a45fd5171d2cc540a360d9ca5f1f/node/actors/network/src/gossip/runner.rs#L109). Since this functionality isn't expected to work, I think we can disable it for now.

## Why ❔

The workflow of signing and submitting L1 batch certificates will be like this:
1. Data is inserted into the `l1_batches` table. 
2. If the node is one of the Attesters it picks up the batch, signs and sends it to the gossip layer via https://github.com/matter-labs/era-consensus/pull/137
3. The consensus collects votes about the L1 batch, and when the threshold is reached it saves the quorum certificate into Postgres
4. The node monitors Main Node (later L1) for new batch QCs and upserts them into the database (the QC can be different than what a particular node inserted based on gossip). This way a node which has been down for a period of time can backfill any QCs it missed. It is assumed that the Main Node API only serves QCs that have no gaps following them, ie. they are final - if it was L1 it wouldn't allow submissions with gaps, and this simulates that semantic.
5. The last height that doesn't have any gaps following it is used as a floor for what needs to be (re)signed and gossiped

This PR supports the above workflow up to step 3.

## Checklist


core/lib/dal/migrations/20240627142548_l1_batches_consensus.up.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE l1_batches_consensus (
+  l1_batch_number BIGINT PRIMARY KEY REFERENCES l1_batches (number) ON DELETE CASCADE,
+  certificate JSONB NOT NULL,
+  is_submitted BOOLEAN NOT NULL DEFAULT false,
```
RomanBrodetski, line 4: you mean submitted to l1 right? So it will be just `false` for a while - shall we maybe ditch it for now and introduce when we actually use it?

aakoshh: Yes, L1 or main node - a synonym to "finalized". It starts as `false` and then turns `true` when we observe the QC on L1 or the main node API. 

No problem I'll remove it and add it as a migration in BFT-480 👍

core/node/consensus/src/storage/connection.rs
```diff
@@ -239,17 +250,152 @@ impl<'a> Connection<'a> {
...
+        // It is supposed to be a Merkle proof that the rolling hash of the batch has been included
+        // in the L1 state tree. The state root hash of L1 won't be available in the DB, it requires
+        // an API client.
+        let batch = attester::SyncBatch {
+            number,
+            payloads,
+            proof: Vec::new(),
+        };
+
+        Ok(Some(batch))
```
RomanBrodetski, line 355: it's not clear for me why we need to load the whole payload for this feature. It should be enough to sign the commitement/root hash - why does consensus side request the whole payload? cc @brunoffranca

core/node/consensus/src/storage/connection.rs
```diff
@@ -239,17 +250,152 @@ impl<'a> Connection<'a> {
...
+            state.last = Some(
+                self.get_batch(ctx, last)
+                    .await
+                    .context("get_batch()")?
+                    .context("last batch not available")?,
+            );
+        }
+
+        Ok(state)
+    }
```
RomanBrodetski, line 400: nit: Is there any specific reason why it's done in imperative way? Could we use `unwrap_or` or similar to achieve the same behaviour?

RomanBrodetski (commented):
just some minor comments

core/lib/dal/src/consensus_dal.rs
```diff
@@ -370,13 +396,81 @@ impl ConsensusDal<'_, '_> {
...
+        use InsertCertificateError as E;
+        let mut txn = self.storage.start_transaction().await?;
+
+        let _l1_batch_header = txn
+            .blocks_dal()
+            .get_l1_batch_header(zksync_types::L1BatchNumber(cert.message.number.0 as u32))
+            .await?
+            .ok_or(E::MissingPayload)?;
+
+        // TODO: Verify that the certificate matches the stored batch:
```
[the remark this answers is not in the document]
aakoshh: Ah okay, thanks, I'll leave a note about this. I was going to follow the [insert_block_certificate](https://github.com/matter-labs/zksync-era/blob/9673bb14ecabc36d77a2bfa4765e9abe33d04650/core/lib/dal/src/consensus_dal.rs#L386-L388) pattern here. It would be a simple enough sanity check, but the other case of validating `SyncBatch::proof` requires API clients that mean it's definitely not something I'd put in here.

Might be worth considering some kind of wrapper to indicate that validation has happened already.

core/node/consensus/src/storage/connection.rs
aakoshh: My understanding from yesterday was that this is the way gossip propagates the batches through the network:
* an RPC server is registered that can [serve batches](https://github.com/matter-labs/era-consensus/blob/decb988eb9e1a45fd5171d2cc540a360d9ca5f1f/node/actors/network/src/gossip/runner.rs#L146) upon request
* this is [called](https://github.com/matter-labs/era-consensus/blob/decb988eb9e1a45fd5171d2cc540a360d9ca5f1f/node/actors/network/src/gossip/runner.rs#L322) whenever the other party indicates it has a new batch
* the `BatchStore` will [look](https://github.com/matter-labs/era-consensus/blob/decb988eb9e1a45fd5171d2cc540a360d9ca5f1f/node/libs/storage/src/batch_store.rs#L231) either in the cache or the database
* the first time `BatchStore` [receives](https://github.com/matter-labs/era-consensus/blob/decb988eb9e1a45fd5171d2cc540a360d9ca5f1f/node/libs/storage/src/batch_store.rs#L111) a batch it will queue it in the cache
* the `BatchStoreRunner` [forwards](https://github.com/matter-labs/era-consensus/blob/decb988eb9e1a45fd5171d2cc540a360d9ca5f1f/node/libs/storage/src/batch_store.rs#L175) anything queued in the cache to the persistent store

So, the payload is for peers that don't have them because they missed all the L2 blocks. Now they get them in a batch, not with a Quorum Certificate but a proof of inclusion on L1.

At the moment this gossiping of availability and pulling of batches seems to happen whether you need it or not.

core/node/consensus/src/storage/connection.rs
aakoshh: The reason was that it seemed to me that the `first` and `last` fields can independently be empty:
* Both can be empty if we haven't got any batch at all
* `first` can be determined but the inclusion on L1 means we might not be able to construct the proof yet, so `last` is empty
* eventually we'll have `last` for something, although it might not be what is currently the `get_last_batch_number`

I also needed to run 3 queries, all of which return `Option`, even though technically at the moment they should either all be `None` or all be `Some`. It's a bit messy, as the comments show it's not the final article, not even sure it should be in `Connection` eventually if it needs API clients. 

But there's no reason why it should be mutating `state`, I can put `unwrap_or`s at the end.

brunoffranca (approved):
Great job!

RomanBrodetski (dismissed):

brunoffranca (approved):

PR #2341, opened 2024-06-27
feat(eth-sender): handle transactions for different operators separately to increase throughtput

tomg10:
Before this PR, all transactions to be send for the first time landed in one queue.
Whenever a transaction had to be resent, on each block: 
- we chose the first non-blob transaction could a blob transaction be resent.
- If and only if there are no such non-blob txs we resend a blob transaction.

While this works, it's not ideal as those two types of transactions can be safely sent independently to increase throughput.
It's also harder to think about time in mempool now as for instance a blob txs may have to wait for an hour before all pending non-blob txs are sent and confirmed.
My PR changes so that those two types of transaction are completely separate. It makes sense as they are using different sender accounts (operators) and a different set of nonces, so they can be safely sent in parallel.


core/lib/dal/src/eth_sender_dal.rs
```diff
@@ -130,7 +139,8 @@ impl EthSenderDal<'_, '_> {
             FROM
                 eth_txs
             WHERE
-                id > (
+                from_addr IS NOT DISTINCT FROM $2
```
RomanBrodetski, line 142: wow this is a trick to get `null = null` return true right? Nice, wasn't aware. Maybe add a comment..

core/node/eth_sender/src/metrics.rs
```diff
@@ -98,7 +98,7 @@ pub(super) struct EthSenderMetrics {
     /// Last L1 block observed by the Ethereum sender.
     pub last_known_l1_block: Family<BlockNumberVariant, Gauge<usize>>,
     /// Number of in-flight txs produced by the Ethereum sender.
-    pub number_of_inflight_txs: Gauge<usize>,
+    pub number_of_inflight_txs: Family<OperatorType, Gauge<usize>>,
```
RomanBrodetski, line 101: very nice! Alternatively you could use a `Family<ActionTypeLabel, ..>` but probably this one is easier to implement

core/node/eth_sender/src/eth_tx_manager.rs
```diff
@@ -634,9 +610,34 @@ impl EthTxManager {
...
 
-        Ok(l1_block_numbers.latest)
+    #[tracing::instrument(skip(self, storage))]
+    async fn loop_iteration(
+        &mut self,
+        storage: &mut Connection<'_, Core>,
+        l1_block_numbers: L1BlockNumbers,
+    ) {
+        tracing::info!("Loop iteration at block {}", l1_block_numbers.latest);
+        for operator_type in [OperatorType::NonBlob, OperatorType::Blob] {
```
RomanBrodetski, line 625: is it possible that due to this change we'll mark a `prove` or an `execute` as complete before we do the same for `commit` of the same batch (issue similar to the one that brought ENs down some days ago?)

RomanBrodetski (commented):
Looks good - please see comments

core/node/eth_sender/src/eth_tx_manager.rs
tomg10: Fortunately we don't have to worry about it in eth-sender, the aggregator only creates PublishProof transaction if corresponding Commit transaction is confirmed:
https://github.com/matter-labs/zksync-era/blob/main/core/node/eth_sender/src/aggregator.rs#L312
So we only have to make sure that transactions for a single operator are confirmed in-order

tomg10: It's good idea to add comment about that!

perekopskiy (dismissed):

core/node/eth_sender/src/metrics.rs
tomg10: I'm not sure it's possible as PublishProof and Execute transactions share the pool of inflight txs

core/node/eth_sender/src/eth_tx_manager.rs
tomg10: Added a comment

core/lib/dal/src/eth_sender_dal.rs
tomg10: +1, corrected

RomanBrodetski (dismissed):

RomanBrodetski (approved):

PR #2345, opened 2024-06-28
chore(dal): Add migration to drop l2 to l1 logs column in l1 batches table.

AnastasiiaVashchuk:
## What ❔

Follow-up PR for https://github.com/matter-labs/zksync-era/pull/2175

## Checklist


deniallugo-ml:

PR #2350, opened 2024-06-28
fix(db): Fix / extend transaction isolation levels

slowli:
## What ❔

- Makes readonly Postgres transaction have "repeatable read" isolation level by default.
- Allows specifying an isolation level in the transaction builder.

## Why ❔

Readonly transactions usually expect a consistent DB view, hence the "repeatable read" isolation level.

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8128,6 +8128,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sqlx",
+ "test-casing",
  "thiserror",
  "tokio",
  "tracing",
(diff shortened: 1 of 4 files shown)


pompon0 (dismissed):
how about we set the isolation level for every transaction, so that we don't rely on postgres default? I mean, like: if level.is_none() then { level = Some( if read_only then repeatable_read else read_committed) }. If not, let's mention the postgres default in the module documentation.

popzxc (dismissed):
Change LGTM, but no opinion on any further discussions.

popzxc (approved):

PR #2352, opened 2024-06-28
feat(api): Retry `read_value`

perekopskiy:
## What ❔

Retries calling `read_value` when it fails with statement_timeout.

## Why ❔

Eliminate sporadic failures.

## Checklist

### DIFF
--- a/core/lib/state/src/postgres/mod.rs
+++ b/core/lib/state/src/postgres/mod.rs
@@ -1,9 +1,11 @@
 use std::{
     mem,
     sync::{Arc, RwLock},
+    time::Duration,
 };
 
 use anyhow::Context as _;
+use backon::{BlockingRetryable, ConstantBuilder};
 use tokio::{
     runtime::Handle,
     sync::{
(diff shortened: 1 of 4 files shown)


core/lib/state/src/postgres/mod.rs
```diff
@@ -489,11 +491,26 @@ impl ReadStorage for PostgresStorage<'_> {
             values_cache.and_then(|cache| cache.get(self.l2_block_number, hashed_key));
 
         let value = cached_value.unwrap_or_else(|| {
+            const RETRY_INTERVAL: Duration = Duration::from_millis(500);
+            const MAX_TRIES: usize = 20;
+
```
Deniallugo, line 496: Could we mark it as to do for creating more generic solution?

Deniallugo: now, it's applied only for reading value and we can't configure it.

perekopskiy: I'm not sure that we need more generic thing here.. If in future there will be need for it then we will add, but for now I don't think it's needed

slowli (commented):
@montekki This is an example of storage operations in VM having suboptimal interface. If storage accesses were implemented using interrupts, it would be possible to error on execution and/or make it async w/o changing any VM interfaces (i.e., achieve looser coupling between the VM and host code).

core/lib/state/src/postgres/mod.rs
```diff
@@ -489,11 +491,26 @@ impl ReadStorage for PostgresStorage<'_> {
...
+            })
+            .retry(
+                &ConstantBuilder::default()
+                    .with_delay(RETRY_INTERVAL)
+                    .with_max_times(MAX_TRIES),
+            )
+            .when(|e| {
+                e.inner()
+                    .as_database_error()
+                    .is_some_and(|e| e.message() == "canceling statement due to statement timeout")
```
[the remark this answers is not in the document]
perekopskiy: I've used error code initially but then I found out that code 57014 is not only for `canceling statement due to statement timeout` but is generally for cancelled queries (e.g `canceling statement due to user request`). So message should be used to catch only timeouts

core/lib/state/src/postgres/mod.rs
```diff
@@ -489,11 +491,26 @@ impl ReadStorage for PostgresStorage<'_> {
...
+            const MAX_TRIES: usize = 20;
+
             let mut dal = self.connection.storage_web3_dal();
-            let value = self
-                .rt_handle
-                .block_on(dal.get_historical_value_unchecked(hashed_key, self.l2_block_number))
-                .expect("Failed executing `read_value`");
+            let value = (|| {
+                self.rt_handle
+                    .block_on(dal.get_historical_value_unchecked(hashed_key, self.l2_block_number))
```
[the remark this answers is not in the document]
perekopskiy: Introducing retries should put zero additional load on DB (contrary to increasing DB timeout). The goal is to eliminate (or make them much more rare) sporadic panics

perekopskiy:
@slowli I doubt it will have any effect. Holding connections longer than they're needed is irrelevant for timeouts I think

PR #2361, opened 2024-07-01
set_refund_for_current_tx

montekki:
## What ❔

Fixes refund-related bugs in the glue code for the new VM:

- Computes `pubdata_published` as a difference with the previous transaction
- Sets refund for the bootloader state

### DIFF
--- a/core/lib/multivm/src/versions/shadow.rs
+++ b/core/lib/multivm/src/versions/shadow.rs
@@ -3,6 +3,7 @@ use std::{
     fmt,
 };
 
+use pretty_assertions::assert_eq;
 use zksync_state::{ReadStorage, StoragePtr};
 use zksync_types::{StorageKey, StorageLogWithPreviousValue, Transaction, U256};
 use zksync_utils::bytecode::CompressedBytecodeInfo;
@@ -153,7 +154,7 @@ where
             for (slot, value) in memory {
                 if *slot == 111 {
                     // FIXME: this particular memory slot (`OPERATOR_REFUNDS_OFFSET`) differs and is always zero for the new VM
-                    *value = U256::zero();
+                    //*value = U256::zero();
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


core/lib/multivm/src/versions/shadow.rs
```diff
@@ -153,7 +154,7 @@ where
             for (slot, value) in memory {
                 if *slot == 111 {
                     // FIXME: this particular memory slot (`OPERATOR_REFUNDS_OFFSET`) differs and is always zero for the new VM
-                    *value = U256::zero();
+                    //*value = U256::zero();
```
slowli, line 157: You can remove the entire `if let` altogether since it's useless now.

core/lib/multivm/src/versions/vm_fast/vm.rs
```diff
@@ -145,14 +146,15 @@ impl<S: ReadStorage> Vm<S> {
                             .as_u64();
 
                         let pubdata_published = self.inner.world_diff.pubdata.0 as u32;
+                        //pubdata_published -= pubdata_before;
```
slowli, line 149: Please remove if this code isn't needed. BTW, how would pubdata computations work if the entire batch is executed (i.e., `execution_mode == VmExecutionMode::Batch`)? I'd think that `pubdata_before` needs to be updated somewhere in `run()` (in `TxHasEnded` hook handler?).

slowli (commented):
Locally, the changes seem to allow the load test set up to complete, and the load test itself now runs fine 🎉 There are still issues with batch executor unit tests though, like extra storage logs and L2-to-L1 logs being produced by the new VM.

core/lib/multivm/src/versions/vm_fast/vm.rs
montekki: I've changed it to be updated here on the spot, as for now i do not have a better theory about where it should be updated.

slowli: I don't think it's correct (it would probably lead to the previous issue of `pubdata_published` for a transaction including pubdata produced by the bootloader / system contracts after executing the previous transaction), but let's fix this sequentially.

PR #2362, opened 2024-07-02
feat(prover): Add prover_cli stats command

EmilLuta:
Adds stats commands, which provides information (batch number, when the request for proving was created and how long it took) for all L1 Batch proofs. This speeds up proving time reduction (by giving everyone visibility into current process) and can serve as further automation in the future (for instance, emit metrics, or use the tooling for automated reports).


prover/prover_cli/src/commands/stats.rs
```diff
@@ -0,0 +1,63 @@
...
+        .build()
+        .await
+        .context("failed to build a prover_connection_pool")?;
+    let mut conn = prover_connection_pool
+        .connection()
+        .await
+        .context("failed to get connection from pool")?;
+
+    let start_date = match opts.period {
+        StatsPeriod::Day => chrono::offset::Local::now().date_naive(),
```
Artemka374, line 38: Am I getting this right, that we can receive only stats from the beginning of today till now(or for the whole week)? It makes not much sense to me to be honest. As a first iteration of adding this command it should be okay, but I would love to see stats for a specific date

Artemka374: I am not that familiar with clap, but probably it is possible to add one more enum option like CustomDate

Artemka374: Good, let's put this somewhere in the backlog then

Artemka374 (approved):

PR #2363, opened 2024-07-02
feat: Unify and port node storage initialization 

popzxc:
## What ❔

- Introduces the `zksync_node_storage_init` crate: a unified approach to the node storage initialization. Key moments:
  - Storage initializer is a structure that makes sure that, well, the storage is initialized.
  - Initializer understands what does initialized storage means, but defers any real initialization actions to the implementation of the `NodeRole` trait. Currently we have two `NodeRole` implementations: `MainNodeRole` and `ExternalNodeRole`.
  - `MainNodeRole` can only perform genesis. It does not support snapshot recovery or automatic rollbacks.
  - `ExternalNodeRole` can perform either genesis or snapshot recovery; it can also detect reorg and perform a rollback is required.
- Framework integration consists of three parts: `NodeRole` resource, and `NodeStorageInitializer` task and precondition.
- Old genesis code for the main node is fully replaced with the framework.
- The init code is integrated into the EN, but the old code is left for the time being.
- Makes snapshot recovery aware of stop signals. It isn't integrated for the old code, but I assume that since the snapshot recovery happens before we setup sigint handler, it implicitly works out of the box.
- Integrated `reorg_detector` into the EN via framework.

## Why ❔

- Unify codebases and approach to the storage initialization.
- Define the interfaces for future extensions.
- Part of porting the codebase to the framework.

## Checklist


core/node/node_storage_init/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+                        self.app_health.clone(),
+                    )
+                    .await?;
+            }
+        }
+
+        // Now we may check whether we're in the invalid state and should perform a rollback.
+        if let Some(to_batch) = self
+            .node_role
+            .should_rollback_to(stop_receiver, &self.pool)
```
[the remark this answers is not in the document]
popzxc: Maybe initialization is not the best word, but I do think that rollback can be seen as a responsibility of this structure.
We have three kinds of storage states: empty, valid, and invalid. This structure is going to enforce the following invariant: when either `run` or `wait_for_initialized_storage` completes successfully, the storage is in the _valid_ state.

I think having both empty and invalid cases treated/"fixed" by the same entity makes it easer to reason about the node state.

core/node/node_storage_init/src/external_node.rs
```diff
@@ -0,0 +1,141 @@
...
+        let needed = zksync_node_sync::genesis::is_genesis_needed(&mut storage).await?;
+        Ok(!needed)
+    }
+
+    async fn snapshot_recovery(
+        &self,
+        stop_receiver: watch::Receiver<bool>,
+        pool: &ConnectionPool<Core>,
+        recovery_config: SnapshotRecoveryConfig,
+        app_health: Arc<AppHealthCheck>,
```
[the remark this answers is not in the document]
popzxc: I agree generally, but (same with [this concept](https://github.com/matter-labs/zksync-era/pull/2363#discussion_r1662108602)) do we need this kind of generalization right now? 

If we have it as a single trait, we will need a structure per initialization step, and given that the main node/EN is somewhat different, it will _likely_ result in two sets of structures, which will have significant cognitive complexity. 

Wouldn't it be better to wait until we have two more or less aligned implementations, and then rewrite these interfaces accordingly?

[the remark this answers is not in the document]
popzxc: Though I agree about the arguments -- it certainly can be made a part of the role implementation.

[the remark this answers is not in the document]
popzxc: Actually, I thought about the arguments for a while -- I think that passing them explicitly to the abstract interface creates artificial constraints for ourselves, which may be beneficial for the purpose of further unification of EN/main node initialization approaches. IMHO, short-term we can treat it as a feature, not as a bug, and rework the interfaces once we have a common set of components for both EN and main node.

core/node/node_storage_init/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+    Genesis,
+    /// Perform or check snapshot recovery.
+    SnapshotRecovery,
+}
+
+#[derive(Debug)]
+pub struct NodeInitializationStrategy {
+    pub genesis: Box<dyn InitializeStorage>,
+    pub snapshot_recovery: Option<Box<dyn InitializeStorage>>,
+    pub block_reverter: Option<Box<dyn RevertStorage>>,
```
slowli, line 34: Nit: Maybe it makes sense to use `Arc<_>` for field to avoid wrapping strategy itself in an `Arc`?

core/node/node_storage_init/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+        // Wait until the rollback is no longer needed.
+        poll(stop_receiver.clone(), POLLING_INTERVAL, || {
+            self.is_rollback_not_needed(stop_receiver.clone())
+        })
+        .await?;
+
+        Ok(())
+    }
+
+    async fn is_database_initialized(&self, decision: InitDecision) -> anyhow::Result<bool> {
```
slowli, line 166: Not entirely sure, but logically, we should be OK as long as the storage is initialized in any acceptable way. So, roughly speaking, check that either `genesis.is_initialized()` or `strategy.snapshot_recovery.is_initialized()` are true if both are provided. Mostly raising this because `bail!` below looks weird; we don't *perform* snapshot recovery in this method.

core/node/node_storage_init/src/traits.rs
```diff
@@ -0,0 +1,30 @@
...
+    /// Implementors of this method may assume that they have unique access to the storage.
+    async fn initialize_storage(&self, stop_receiver: watch::Receiver<bool>) -> anyhow::Result<()>;
+}
+
+/// An abstract storage revert strategy.
+/// This trait assumes that for any invalid state there exists a batch number to which the storage can be rolled back.
+#[async_trait::async_trait]
+pub trait RevertStorage: fmt::Debug + Send + Sync + 'static {
+    /// Checks if the storage is invalid state and has to be rolled back.
+    async fn is_revert_needed(&self) -> anyhow::Result<Option<L1BatchNumber>>;
```
slowli, line 22: Bikeshedding: Since this method returns richer data than `bool`, IMO, it should be named like `last_correct_l1_batch_for_reorg`, or `detect_reorg` + use a dedicated struct wrapping `L1BatchNumber`:

```rust
pub struct DetectedReorg {
    pub last_correct_l1_batch: L1BatchNumber,
}
```

core/node/node_storage_init/src/external_node/revert.rs
```diff
@@ -0,0 +1,50 @@
...
+        tracing::info!("Reverting to l1 batch number {to_batch}");
+        self.reverter.roll_back(to_batch).await?;
+        tracing::info!("Revert successfully completed");
+        Ok(())
+    }
+
+    async fn is_revert_needed(&self) -> anyhow::Result<Option<L1BatchNumber>> {
+        let mut reorg_detector = ReorgDetector::new(self.client.clone(), self.pool.clone());
+        // We don't need this kind of granularity for detecting stop signal; it's expected to be treated
+        // by the caller.
```
slowli, line 34: Can you explain what you mean by this comment? `run_once()` is cancel-aware to deal with transient errors in the L2 RPC client (e.g., no network). So, making `is_revert_needed()` not cancel-aware can degrade UX for the node operator (it could become unresponsive to SIGINTs).

core/node/node_storage_init/src/traits.rs
popzxc: Changed the name but kept the type; IMHO it's self-explanatory enough, and if not, there exists a doc comment for that.

core/node/node_storage_init/src/external_node/snapshot_recovery.rs
```diff
@@ -0,0 +1,82 @@
...
+    async fn initialize_storage(&self, stop_receiver: watch::Receiver<bool>) -> anyhow::Result<()> {
+        let pool = self.pool.clone();
+        tracing::warn!("Proceeding with snapshot recovery. This is an experimental feature; use at your own risk");
+        let object_store_config =
+            self.recovery_config.object_store_config.clone().context(
+                "Snapshot object store must be presented if snapshot recovery is activated",
+            )?;
+        let object_store = ObjectStoreFactory::new(object_store_config)
+            .create_store()
+            .await?;
```
slowli, line 35: Maybe it makes sense to inject `ObjectStore` directly as a field of `ExternalNodeSnapshotRecovery` instead of using `SnapshotRecoveryConfig.object_store_config` to make it more decoupled from the config?

slowli (dismissed):

core/node/node_storage_init/src/external_node/snapshot_recovery.rs
popzxc: Certainly, but I think the scope of refactoring is much larger (e.g. we need to express different meanings for the same types better, like now we have several kinds of object store, and also several kinds of private key signers), so I think we'll tackle that separately to not over-extend the scope.

slowli (approved):

PR #2369, opened 2024-07-02
feat(vm-runner): make vm runner report time taken

itegulov:
## What ❔

A long overdue addition of `time_taken` to VM runner

## Why ❔

Observability

## Checklist


perekopskiy (commented):
Hmm, for other queue-like tables `time_taken` was the time between "job becomes in-progress" and "job becomes successful". For vm runner we don't have an explicit in-progress status and the task is split into 3 (load storage, run vm, save results). I think we should aim to make `started_at` to represent the earliest point in time when batch is already processing.
I think it's ok to ignore storage loading step here but I would really want that "vm run" time is included in `time_taken`. WDYT?

itegulov:
@perekopskiy I considered that at first but it kind of made the code ugly because there is no clean way to propagate `started_at` from the start of the execution. I have just pushed a commit with one way to do that which requires touching `StateKeeperOutputHandler`. Let me know what you think and if you know of a cleaner way.

perekopskiy (commented):
What do you think about adding DB column `processing_started_at` and filling it in new `VmRunnerIo`'s method?

core/node/vm_runner/src/io.rs
```diff
@@ -31,6 +31,18 @@ pub trait VmRunnerIo: Debug + Send + Sync + 'static {
         conn: &mut Connection<'_, Core>,
     ) -> anyhow::Result<L1BatchNumber>;
 
+    /// Marks the specified batch as the latest completed batch. All earlier batches are considered
+    /// to be completed too. No guarantees about later batches.
+    ///
+    /// # Errors
+    ///
+    /// Propagates DB errors.
+    async fn mark_l1_batch_as_processing(
```
perekopskiy, line 40: Please change comment

perekopskiy (dismissed):

perekopskiy (approved):

PR #2374, opened 2024-07-02
feat: add block timestamp to `eth_getLogs`

ischasny:
## What ❔

Add a new field `blockTimestamp` to the `eth_getLogs` endpoint. 

## Why ❔

That'd allow network indexers to avoid sending a second requests just to get timestamp for each block. More [info](https://ethereum-magicians.org/t/proposal-for-adding-blocktimestamp-to-logs-object-returned-by-eth-getlogs-and-related-requests/11183)


core/lib/basic_types/src/web3/mod.rs
```diff
@@ -327,6 +327,8 @@ pub struct Log {
     pub log_type: Option<String>,
     /// Removed
     pub removed: Option<bool>,
+    /// L2 block timestamp
+    pub block_timestamp: Option<i64>,
```
perekopskiy, line 331: missing serde(rename)

core/lib/dal/src/events_dal.rs
```diff
@@ -222,7 +222,8 @@ impl EventsDal<'_, '_> {
                 tx_hash,
                 tx_index_in_block,
                 event_index_in_block,
-                event_index_in_tx
+                event_index_in_tx,
+                NULL::BIGINT AS "block_timestamp?"
```
perekopskiy, line 226: Fields that are set to null here (block_hash, l1_batch_number) are filled in `get_transaction_receipts`. You should fill block_timestamp as well

perekopskiy (commented):
Please check that `block_timestamp` is indeed returned in integration tests

core/lib/basic_types/src/web3/mod.rs
ischasny: Good catch - added

ischasny:
> Please check that `block_timestamp` is indeed returned in integration tests

The issue is that integrations tests need to have ethers.js library updated as well, otherwise the field won't be propagated. 
* I'll update ethers.js next
* Once that done - I'll bump up the dependency and update integration tests

core/lib/types/src/api/mod.rs
```diff
@@ -443,6 +446,9 @@ pub struct Log {
     pub log_type: Option<String>,
     /// Removed
     pub removed: Option<bool>,
+    /// L2 block timestamp
+    #[serde(rename = "blockTimestamp")]
+    pub block_timestamp: Option<i64>,
```
perekopskiy, line 451: If I'm not mistaken adding `block_timestamp` to receipt is not a part of proposal. You can keep timestamp in `StorageTransactionReceipt`, it should be enough

perekopskiy (commented):
You can use `provider.send('eth_getLogs', ...)` instead of `provider.getLogs(...)`, and add a TODO to switch to proper method after SDK is updated

core/lib/types/src/api/mod.rs
ischasny: good point - skipped serialisation for this field. I'd prefer to leave the field in the struct though because that'd allow to keep `get_transaction_receipts` logic without major changes.

perekopskiy: But the serialization is the main purpose of this struct, I insist on removing `block_timestamp` from it

ischasny: fair enough - removed the field and modified the mapping function

PR #2375, opened 2024-07-02
feat(prover): Add port opt for prover

matias-gonz:
## What ❔

Add port opt for prover


matias-gonz:
Closed in favor of #2385

PR #2382, opened 2024-07-04
feat(prover): Add bellman_cuda_path to FriProofCompressor

matias-gonz:
## What ❔

Add bellman_cuda_path to FriProofCompressor

## Why ❔

To support file based config for gpu feature


matias-gonz:
Closed since the variable is used in compilation time and there is no use to set it while running

PR #2383, opened 2024-07-04
feat: Minimal External API Fetcher

shahar4:
## What ❔

For chains running with a custom base token, use live prices in creating  ETH<->BaseToken conversion ratios.
This PR
- adds a first `PriceAPIClientResource` with a `PriceAPIClient` CoinGecko client.
- Uses this client in the `BaseTokenRatioPersister` to fetch new prices and create new ratios

Not included in this PR and will be follow-up work:
- Redundancy in the number of external price feeds used
- Different "strategies" for how true price converged from multiple price feeds

## Why ❔

For the base token flow, we need to be be able to fetch live prices of BaseToken & ETH from which to create the conversion ratios.

## Checklist


Cargo.toml
```diff
@@ -178,6 +178,8 @@ tracing-subscriber = "0.3"
 tracing-opentelemetry = "0.21.0"
 url = "2"
 web3 = "0.19.0"
+httpmock = "0.7.0"
```
popzxc, line 181: Can be removed now.

core/lib/external_price_api/src/coingecko_api.rs
```diff
@@ -0,0 +1,129 @@
...
+                "Price not found for token: {}",
+                address_str
+            )),
+        }
+    }
+
+    async fn get_token_price_by_id(&self, id: String) -> anyhow::Result<f64> {
+        let price_url = self
+            .base_url
+            .join(format!("/api/v3/simple/price?ids={}&vs_currencies={}", id, USD_ID).as_str())
```
popzxc, line 84: `vs_currencies` seem to [support eth](https://docs.coingecko.com/reference/simple-supported-currencies), so you don't need two requests; one should be enough.
Otherwise your calculation will not be atomic and it will loose precision.

core/lib/types/src/base_token_ratio.rs
```diff
@@ -13,10 +15,39 @@ pub struct BaseTokenRatio {
...
 /// Struct to represent API response containing denominator, numerator, and timestamp.
-#[derive(Debug)]
-pub struct BaseTokenAPIRatio {
-    pub numerator: NonZeroU64,
-    pub denominator: NonZeroU64,
-    /// Either the timestamp of the quote or the timestamp of the request.
+#[derive(Debug, Clone, PartialEq)]
+pub struct BaseTokenAPIPrice {
+    pub base_token_price: BigDecimal,
+    pub eth_price: BigDecimal,
```
popzxc, line 21: ...with a single request you will also be able to use `Fraction` right there, on the API client side, and keep `numerator`/`denominator` structure you had before.

core/node/node_framework/src/implementations/layers/base_token_ratio_persister.rs
```diff
@@ -41,7 +48,18 @@ impl WiringLayer for BaseTokenRatioPersisterLayer {
         let master_pool_resource = context.get_resource::<PoolResource<MasterPool>>()?;
         let master_pool = master_pool_resource.get().await?;
 
-        let persister = BaseTokenRatioPersister::new(master_pool, self.config);
+        let price_api_client = context.get_resource_or_default::<PriceAPIClientResource>();
```
popzxc, line 51: Is it correct to run persister without an API client? I think if you run persister but didn't provide a client resource for it, it should be an error; otherwise it can lead to confusing problems in production (e.g. persister runs, but it persists incorrect rates). I don't think fallback to default is a correct choice here.

core/node/node_framework/src/implementations/layers/external_price_api_client.rs
```diff
@@ -0,0 +1,44 @@
...
+/// Wiring layer for `ExternalPriceApiClient`
+///
+/// Responsible for inserting all the resources to get base token prices from external price feeds to
+/// be used by the `BaseTokenRatioPersister`.
+///
+/// ## Adds resources
+///
+/// - PriceAPIClientResource
+#[derive(Debug)]
+pub struct ExternalPriceApiClientsLayer {
```
popzxc, line 21: An appropriate name would be `CoingeckoClientLayer` (e.g. we're adding a concrete layer here, not abstract).

core/lib/config/src/configs/external_price_api_client.rs
```diff
@@ -0,0 +1,34 @@
+use std::time::Duration;
+
+use serde::Deserialize;
+use url::Url;
+
+pub const DEFAULT_TIMEOUT_MS: u64 = 10_000;
+pub const DEFAULT_COINGECKO_API_URL: &str = "https://pro-api.coingecko.com";
```
popzxc, line 7: I don't think we should have a default here. It's better to have it in the coingecko client impl.

core/lib/config/src/configs/external_price_api_client.rs
```diff
@@ -0,0 +1,34 @@
...
+}
+
+impl ExternalPriceApiClientConfig {
+    fn default_timeout() -> u64 {
+        DEFAULT_TIMEOUT_MS
+    }
+
+    fn default_base_url() -> String {
+        DEFAULT_COINGECKO_API_URL.to_string()
+    }
```
popzxc, line 25: ...e.g. it's rather confusing; if you don't know what API you are going to talk to, the "default" URL doesn't make much sense.

core/node/node_framework/src/implementations/layers/external_price_api_client.rs
shahar4: I was thinking about having
- one resource per external API price provider the chain operator wants to use
- one layer to insert them all
Do you think that's not the way to go? Would it be more appropriate for there to be a layer per external price provider?

core/node/node_framework/src/implementations/layers/base_token_ratio_persister.rs
shahar4: I see what you're saying. My reasoning was to have the default for the i tests, as I didn't want to insert the CG resource there. WDYT? I'd love a suggestion here!

core/node/node_framework/src/implementations/layers/external_price_api_client.rs
popzxc: Layer per external price provider is more flexible. If you have one layer per concrete resource implementation, then you need to choose _just_ layer. If you have layer with several resources changed via "if", then you have to pass the config to layer so that it can choose the resource for you. If we look further, someone may implement their own external API source as a separate crate (one of the problems we're trying to solve with the framework is that we don't want to merge every single feature to our repo, we want to enable people to write their own layers and inject them into our code), you'll get in a weird situation where a few resource implementations are added by layer A, but for others you need B, C, D.

tl;dr One concrete resource per layer is a good option. [API example](https://github.com/matter-labs/zksync-era/tree/main/core/node/node_framework/src/implementations/layers/web3_api/tx_sink), [consensus example](https://github.com/matter-labs/zksync-era/tree/main/core/node/node_framework/src/implementations/layers/consensus).

core/node/node_framework/src/implementations/layers/base_token_ratio_persister.rs
popzxc: You may alter configuration of the node for the integration tests.
I think a problem you may have here is that you miss an option to specify the config provider (e.g. `pub source: String`, and then in the node builder based on this string you can choose between e.g. `coingecko` and `mock`)

core/lib/config/src/configs/external_price_api_client.rs
```diff
@@ -0,0 +1,29 @@
...
+
+use serde::Deserialize;
+
+pub const DEFAULT_TIMEOUT_MS: u64 = 10_000;
+
+#[derive(Debug, Clone, PartialEq, Deserialize)]
+pub struct ExternalPriceApiClientConfig {
+    #[serde(default = "ExternalPriceApiClientConfig::default_source")]
+    pub source: String,
+    pub base_url: Option<String>,
```
popzxc, line 11: Once again, there should be no defaults for these values.
Configs are loaded conditionally, e.g. they are loaded as an `Option` by default, and only dereferenced if actually used. We only use this config if we run base token ratio persister -- a new component -- so we don't have to worry about backward compatibility.

At the same time, declaring these fields as optional can lead to unwanted and unexpected results. For example, we want to run a "real" price fetcher, but we forgot to provide configs. Default configs will be loaded, and the behavior will _likely_ not match what we want.

There is no default that makes sense "in general"; for localhost it's no-op impl, for production it's coingecko, they are both equally important.

So I suggest that you:
a) make `source` required.
b) remove `default_source()` method.
c) add localhost defaults to the dev configs (both env and file-based).

This shouldn't be breaking because, as I said, this is a new component, and the config won't be accessed unless it's needed.

core/node/base_token_adjuster/src/base_token_ratio_persister.rs
```diff
@@ -1,72 +1,82 @@
...
             }
 
-            let new_ratio = self.fetch_new_ratio().await?;
-            self.persist_ratio(&new_ratio, &pool).await?;
-            // TODO(PE-128): Update L1 ratio
+            if let Err(err) = self.loop_iteration().await {
+                tracing::error!(
+                    "Failed to execute a base_token_ratio_persister loop iteration: {err}"
+                );
+            }
```
popzxc, line 48: Do we consider any error to be transient? Normally we treat database errors as irrecoverable and bail, while trying to recover on transient errors only. Otherwise there may be potentially dangerous situations; e.g. imagine that tx for persisting token on L1 reverts -- we will be sending a reverting transaction over and over, draining the operator's balance.

I'm raising it now, because later these lines won't be a part of diff, and it'll be harder to notice the problem.

shahar4: Good point. Made it a persistent error and added a retry mechanism in order to not shut down & alert whenever Coingecko temporarily goes down.

popzxc (dismissed):
I think it's safe to merge, but I would still expect tests for the new functionality.

I think at the very least we need to implement a "testing" price provider which will return a constant price +-20% all the time, and integrate it to our CI matrix when we run tests for custom base token; right now it seems that we still test custom base token with 1:1 ratio, which doesn't seem very helpful.

popzxc (dismissed):

RomanBrodetski (dismissed):

RomanBrodetski (approved):

popzxc (approved):

hatemosphere (approved):

PR #2385, opened 2024-07-04
feat(prover): Add prometheus port to witness generator config

matias-gonz:
## What ❔

Add prometheus port to witness generator config


prover/witness_generator/src/main.rs
```diff
@@ -122,9 +122,12 @@ async fn main() -> anyhow::Result<()> {
     let config = general_config
         .witness_generator
         .context("witness generator config")?;
-    let prometheus_config = general_config
+    let mut prometheus_config = general_config
         .prometheus_config
         .context("prometheus config")?;
+    if let Some(prometheus_listener_port) = config.prometheus_listener_port {
+        prometheus_config.listener_port = prometheus_listener_port;
+    }
```
Deniallugo, line 130: please add a comment, on why you are doing it. 
Without context it looks weird and I'd better have  `config.prometheus_listener_port` as a default and a fallback to the port. 

probably as an additional u32

prover/witness_generator/src/main.rs
```diff
@@ -125,6 +125,14 @@ async fn main() -> anyhow::Result<()> {
     let prometheus_config = general_config
         .prometheus_config
         .context("prometheus config")?;
```
Deniallugo, line 127: you can move it to the line 132 add do not fail if this is not presented

Deniallugo (approved):

Artemka374 (approved):

PR #2386, opened 2024-07-04
feat(tee): add Prometheus metrics to the TEE Prover

pbeza:
## What ❔

This commit adds Prometheus metrics to the TEE Prover. Specifically, the following metrics were added:
- Waiting time for a new batch to be proven
- Proof generation time
- Proof submitting time
- Network error counter
- Last block number processed

## Why ❔

Setting up Prometheus metrics is a prerequisite before rolling them out to staging and testnet environments. Prometheus metrics are useful for monitoring, providing valuable insights into the running system.

## Checklist


pbeza:
@popzxc @haraldh Please review. Hopefully, this should be an easy one. I'm not sure if there's an easy way to test it before rolling it out to staging/testnet.

core/bin/zksync_tee_prover/src/api_client.rs
```diff
@@ -97,11 +99,13 @@ impl TeeApiClient {
             proof: root_hash.as_bytes().into(),
             tee_type,
         }));
+        let started_at = Instant::now();
         self.post::<_, SubmitTeeProofResponse, _>(
             format!("/tee/submit_proofs/{batch_number}").as_str(),
             request,
         )
         .await?;
+        METRICS.proof_submitting_time.observe(started_at.elapsed());
```
popzxc, line 108: I think you don't need to explicitly use `Instant` here; probably it should work as follows:

```rust
let observer = METRICS.proof_submitting_time.start();
// ...
observer.observe();
```

popzxc (approved):

core/bin/zksync_tee_prover/src/api_client.rs
pbeza: Thanks, [fixed](https://github.com/matter-labs/zksync-era/pull/2386/commits/0b5a48ef7e3d56409728b77bb56fdb5f5d9e12e4)! ✅

I was inspired by this code for measuring time for metrics.:
https://github.com/matter-labs/zksync-era/blob/217a4ba7e9b91001429eef952e2b840a5341c591/core/node/tee_verifier_input_producer/src/lib.rs#L219-L227
(I improved it [here](https://github.com/matter-labs/zksync-era/pull/2386/commits/f8fc17c1806dd2cee8bf388151b24f807017f095))

I wish I had an automated tool to let me know if my code could be more idiomatic without explicitly asking ChatGPT for help. Maybe such a tool already exists that could help in this case?

popzxc: I'm not aware of any, especially given that we're talking about our own library.

popzxc (approved):

PR #2388, opened 2024-07-04
feat(base-token): Base token price ratio cache update frequency configurable

cytadela8:
## What ❔

This PR makes base token price ration cache update frequency configurable

## Why ❔

Base token price ratio is updated in DB with configurable frequency, but when used it's taken from cache that is updated in non-configurable manner

## Checklist


etc/env/base/base_token_adjuster.toml
```diff
@@ -4,3 +4,5 @@
 
 # How often to poll external price feeds for the base token price.
 price_polling_interval_ms = "30000"
+
+price_cache_update_interval_ms = "2000"
```
shahar4, line 8: Super nit: new line

shahar4 (dismissed):
Looks good!

core/lib/config/src/configs/base_token_adjuster.rs
```diff
@@ -2,30 +2,46 @@ use std::time::Duration;
...
-            price_polling_interval_ms: Self::default_interval(),
+            price_polling_interval_ms: Self::default_pooling_interval(),
+            price_cache_update_interval_ms: Self::default_cache_update_interval(),
         }
     }
 }
 
 impl BaseTokenAdjusterConfig {
-    fn default_interval() -> u64 {
+    fn default_pooling_interval() -> u64 {
```
perekopskiy, line 32: nit: should be polling not pooling

perekopskiy (approved):

PR #2390, opened 2024-07-05
perf(vm): Fix new VM performance

slowli:
## What ❔

- Fixes `ImmutableStorageView` caching (turns out the new VM doesn't have caching for reads, only for writes).


slowli:
Closing as obsolete.

PR #2391, opened 2024-07-05
perf(vm): Test old revisions of new VM

slowli:
## What ❔

This is a temporary PR that will get removed once the reason of poor loadtest performance in CI is figured out.


slowli:
The problem is figured out and fixed in the VM repo, so this PR can now be safely closed.

PR #2398, opened 2024-07-08
fix(tee): Introduce a 1 second delay in the batch poll

pbeza:
## What ❔

Introduce a 1 second delay in the batch poll.

## Why ❔

We don't want to poll the batches too often if they are unavailable, as it wastes CPU resources.

## Checklist


pbeza:
Opening this PR at @haraldh's request.

While we're at it, I want to check if we should parallelize polling and verifying batches. Right now, we're verifying/proving them [one-by-one](https://github.com/matter-labs/zksync-era/blob/52a4680ed26e755b860e3b97c79618a0c20cb696/core/bin/zksync_tee_prover/src/tee_prover.rs#L102-L123), which means waiting for the `verify()` method to finish before moving on to the next batch, and it's pretty slow (because `verify()` is slow). @haraldh is concerned that it might use too much memory if parallelized since our Azure instances have (very) limited memory. I think `verify()` is so slow that not parallelizing would be a waste from a performance PoV. Any thoughts, @Igor? Should we benchmark, or would that be premature optimization?

core/bin/zksync_tee_prover/src/tee_prover.rs
```diff
@@ -188,15 +188,15 @@ impl Task for TeeProver {
...
-                        .ok();
                     backoff = std::cmp::min(
                         backoff.mul_f32(self.config.retry_backoff_multiplier),
                         self.config.max_backoff,
                     );
                 }
             }
+            tokio::time::timeout(backoff, stop_receiver.0.changed())
+                .await
+                .ok();
```
popzxc, line 199: This will sleep for 1s even if there are consecutive batches (e.g `Some` after `Some`), and I don't think we want that.
Probably you want to handle `Ok(Some(_))`, `Ok(None)` and `Err(_)` differently (e.g. don't wait if we actually processed a batch). Simplest option would be to return `needs_to_sleep` from `match result`, but you will get bonus points if you'll implement a more elegant solution :)

popzxc (commented):
@pbeza let's implement optimizations once we need them and have some initial data to implement them properly.

core/bin/zksync_tee_prover/src/tee_prover.rs
pbeza: Fixed. ✅ Not sure how to implement this better. If we want to parallelize it later, we'll probably have to redesign it anyway.

popzxc (approved):

PR #2399, opened 2024-07-08
feat: Poll DB and sign L1 batches (BFT-474)

aakoshh:
## What ❔

Implements `PersistentBatchStore` methods added in https://github.com/matter-labs/era-consensus/pull/148

- [x] Add `ConsensusDal::unsigned_batch_numbers` to retrieve L1 batch numbers without a QC after a given minimum batch number
- [x] Add `Store::unsigned_batch_numbers` which return the unsigned batches that are max 10 batches older than the last one
- [x] Add `ConsensusDal::get_batch_to_sign` after finding out what to put into `hash`
- [x] Update `ConsensusDal::insert_batch_certificate` to validate the correct `hash` value
- [x] Test `unsigned_batch_numbers`
- [ ] Test `get_batch_to_sign` and `insert_batch_certificate`
- [ ] Point `era-consensus` at the latest version instead of https://github.com/matter-labs/era-consensus/tree/bft-474-pairing-prev-version then delete that branch

The hash check has been implemented against the `StoredBatchInfo::hash()` which isn't available in the DAL crate.

## Why ❔

The consensus `Executor` will poll `unsigned_batch_numbers` and submit attestations over `get_batch_to_sign`. These methods decide 1) which batches to sign and 2) what content to sign.

## Checklist


core/lib/dal/src/consensus_dal.rs
```diff
@@ -480,6 +461,47 @@ impl ConsensusDal<'_, '_> {
...
+                AND NOT EXISTS (
+                    SELECT
+                        1
+                    FROM
+                        l1_batches_consensus c
+                    WHERE
+                        c.l1_batch_number = b.number
+                )
+            ORDER BY
+                b.number
```
RomanBrodetski, line 493: I'd probably do a join and then filter out values for which there is a matching row in consensus table (`WHERE l1_batches_consensus. l1_batch_number IS NULL`) - it's arguably more idiomatic

aakoshh: From my DB days, these two in an ideal test plan both turned into ANTI-JOIN:
* `select * from a where not exists (select 1 from b where b.a_id = a.id)`
* `select a.* from a left outer join b on a.id = b.a_id where b.id is null`

I don't have a preference, if but just to emphasize it has to be a _left outer join_ otherwise it would do the opposite. If you understand it the same way I'm happy to rewrite, no problem.

aakoshh: I'd expect the `not exists` syntax to have some advantages if one really wants to achieve an _anti join_ in that:
* you can't accidentally use the columns of the right side in the select
* under no circumstances will it duplicate data from the left side
* the intent is very clear (to my eyes at least)

I'd assume the `left outer join` syntax might require a foreign key to realise it's actually an _anti join_. 

In Postgres they should be equivalent in terms of plan, if everything goes the way it should. I think I started using this syntax because it was more likely to result in the right plan in Oracle, but not sure why that was. Maybe just some general advice I saw somewhere.

aakoshh:
@RomanBrodetski I think it's correct to sign the hash of `StoredBatchInfo` and not just the `commitment` field because this is a record that gets converted to RLP before hashing, it's meant to be sent to the L1 contract, and therefore the attestation should cover the whole content, not just one field in it. 

It might even be enough to sign just this and not the hash of `attester::Batch` itself, since it already contains the number. Alas, that would mean we couldn't verify a `BatchQC` on its own and someone could change the `number` in it the message. 

So Solidity would have to verify something like `keccak256(abi.encode([batch.number, keccak256(abi.encode(<stored-batch-info>))])`

core/lib/dal/src/consensus_dal.rs
aakoshh: Refactored to use left join in https://github.com/matter-labs/zksync-era/pull/2410

PR #2403, opened 2024-07-08
feat: add zksync_tee_prover and container to nix

haraldh:
```
$ nix build -L .#tee_prover
$ nix build -L .#container-tee_prover-dcap
$ nix build -L .#container-tee_prover-azure
$ export IMAGE_TAG=$(docker load < result | grep -Po 'Loaded image.*: \K.*')

$ docker run -i --env GRAMINE_DIRECT=1 --env TEE_API_URL="http://127.0.0.1:3320" --privileged --init $IMAGE_TAG
```

## Checklist

### DIFF
--- a/core/bin/zksync_tee_prover/Cargo.toml
+++ b/core/bin/zksync_tee_prover/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "zksync_tee_prover"
-version.workspace = true
+version = "0.1.0"
 edition.workspace = true
 authors.workspace = true
 homepage.workspace = true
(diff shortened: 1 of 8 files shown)


pbeza (dismissed):

nix/cargo-vendor.nix
```diff
@@ -0,0 +1,27 @@
+# patched version of cargo to support `cargo vendor` for vendoring dependencies
```
popzxc, line 1: Can we move it to `etc/nix` to not overly pollute the top-level scope, and also add a `README.md` for the directory so that people are aware of why we need it and where is it used?

haraldh: yes, will do

haraldh: done

popzxc (approved):

popzxc (approved):

haraldh:
ok, rewriting, because `cargo vendor` is supposed to work soonish :-)

PR #2408, opened 2024-07-09
feat: add revert tests (external node) to zk_toolbox

manuelmauro:
## What ❔

- Adds revert tests (external node) to zk_toolbox

## Checklist

### DIFF
--- a/core/tests/revert-test/tests/revert-and-restart-en.test.ts
+++ b/core/tests/revert-test/tests/revert-and-restart-en.test.ts
@@ -64,7 +64,7 @@ if (fileConfig.loadFromFile) {
     ethClientWeb3Url = env.ETH_CLIENT_WEB3_URL;
     apiWeb3JsonRpcHttpUrl = env.API_WEB3_JSON_RPC_HTTP_URL;
     baseTokenAddress = env.CONTRACTS_BASE_TOKEN_ADDR;
-    env.EN_ETH_CLIENT_URL`http://127.0.0.1:${env.EN_HTTP_PORT}`;
+    env.EN_ETH_CLIENT_URL = `http://127.0.0.1:${env.EN_HTTP_PORT}`;
 }
 
 interface SuggestedValues {
@@ -153,7 +153,7 @@ async function killServerAndWaitForShutdown(tester: Tester, server: string) {
 }
 
 class MainNode {
-    constructor(public tester: Tester, private proc: child_process.ChildProcess) {}
+    constructor(public tester: Tester, private proc: child_process.ChildProcess) { }
 
     // Terminates all main node processes running.
     public static async terminateAll() {
@@ -209,7 +209,7 @@ class MainNode {
 }
 
 class ExtNode {
-    constructor(public tester: Tester, private proc: child_process.ChildProcess) {}
+    constructor(public tester: Tester, private proc: child_process.ChildProcess) { }
 
     // Terminates all main node processes running.
     public static async terminateAll() {
--- a/core/tests/revert-test/tests/revert-and-restart-en.test.ts
+++ b/core/tests/revert-test/tests/revert-and-restart-en.test.ts
@@ -133,12 +133,11 @@ async function runBlockReverter(args: string[]): Promise<string> {
             `;
     }
 
-    const executedProcess = await utils.exec(
-        `cd ${pathToHome} && RUST_LOG=off cargo run --bin block_reverter --release -- ${args.join(
-            ' '
-        )} ${fileConfigFlags}`
-        // ^ Switch off logs to not pollute the output JSON
-    );
+    const cmd = `cd ${pathToHome} && RUST_LOG=off cargo run --bin block_reverter --release -- ${args.join(
+        ' '
+    )} ${fileConfigFlags}`;
+    console.log(`Running block_reverter: ${cmd}`);
+    const executedProcess = await utils.exec(cmd);
 
     return executedProcess.stdout;
 }
@@ -387,7 +386,7 @@ describe('Block reverting test', function () {
             'print-suggested-values',
             '--json',
             '--operator-address',
-            '0xde03a0B5963f75f1C8485B355fF6D30f3093BDE7'
+            '0xabcf96e1ee478481042a0c4e34cdceceae01b154'
         ]);
         console.log(`values = ${values_json}`);
         const values = parseSuggestedValues(values_json);
--- a/core/tests/revert-test/tests/revert-and-restart-en.test.ts
+++ b/core/tests/revert-test/tests/revert-and-restart-en.test.ts
@@ -11,8 +11,10 @@ import { expect, assert } from 'chai';
 import fs from 'fs';
 import * as child_process from 'child_process';
 import * as dotenv from 'dotenv';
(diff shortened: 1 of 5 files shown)


core/tests/revert-test/tests/revert-and-restart-en.test.ts
```diff
@@ -188,51 +241,37 @@ class ExtNode {
...
+        let tester: Tester = await Tester.init(ethClientWeb3Url, enEthClientUrl, baseTokenAddress);
         while (true) {
             try {
                 await tester.syncWallet.provider.getBlockNumber();
                 break;
             } catch (err) {
-                if (proc.exitCode != null) {
-                    assert.fail(`node failed to start, exitCode = ${proc.exitCode}`);
-                }
+                // NOTE: not an infinite loop, as the test will eventually timeout.
```
Deniallugo, line 264: Seems, that fail fast is better. timeout usually is huge

core/tests/revert-test/tests/revert-and-restart-en.test.ts
```diff
@@ -359,10 +398,11 @@ describe('Block reverting test', function () {
         mainNode = await MainNode.spawn(mainLogs, enableConsensus, true);
 
         console.log('Wait for the external node to detect reorg and terminate');
-        await extNode.waitForExit();
+        await utils.sleep(3);
```
Deniallugo, line 401: could we keep some other checks, not sleeps, most of the time it doesn't work properly

core/tests/revert-test/tests/revert-and-restart-en.test.ts
```diff
@@ -359,10 +398,11 @@ describe('Block reverting test', function () {
...
 
         console.log('Restart external node and wait for it to revert.');
         extNode = await ExtNode.spawn(extLogs, enableConsensus);
+        await utils.sleep(30);
```
Deniallugo, line 405: why do we need to sleep here?

zk_toolbox/crates/zk_supervisor/src/commands/test/revert.rs
```diff
@@ -22,12 +24,27 @@ pub fn run(shell: &Shell, args: RevertArgs) -> anyhow::Result<()> {
...
 ) -> anyhow::Result<()> {
     let _dir_guard = shell.push_dir(&ecosystem_config.link_to_code);
-    let spinner = Spinner::new("Installing and building dependencies...");
+    let spinner = Spinner::new(MSG_REVERT_TEST_INSTALLING_DEPENDENCIES);
     Cmd::new(cmd!(shell, "yarn install")).run()?;
     Cmd::new(cmd!(shell, "yarn utils build")).run()?;
+
+    // Build always zk_server and block_reverter binaries
+    Server::new(None, ecosystem_config.link_to_code.clone()).build(shell)?;
+    Cmd::new(cmd!(shell, "cargo build --release --bin block_reverter")).run()?;
```
Deniallugo, line 37: why do we need it? it suppose to work even without prebuilding

core/tests/revert-test/tests/revert-and-restart-en.test.ts
manuelmauro: Although within `ExtNode.spawn` we wait for the node to be responsive, without this sleep the following L1 txn has a chance of failing. Maybe connected to the time needed to revert?

core/tests/revert-test/tests/revert-and-restart-en.test.ts
manuelmauro: Now that the test uses a dedicated `background` utils, we can restore the previous check. Addressed in https://github.com/matter-labs/zksync-era/pull/2408/commits/459057591a02417a8b0f01f1199861ce041846f0

core/tests/revert-test/tests/revert-and-restart-en.test.ts
manuelmauro: For the reason above, we can restore also this. Addressed in https://github.com/matter-labs/zksync-era/pull/2408/commits/06d0f97a464b697d052be4745162258638086f92

zk_toolbox/crates/zk_supervisor/src/commands/test/revert.rs
manuelmauro: Double-checking this with Agus (it's part of his initial commit).

manuelmauro: Removed in https://github.com/matter-labs/zksync-era/pull/2408/commits/aac7164cacefbd9cef86486761b45c1e37ab7ed5

zk_toolbox/crates/zk_supervisor/src/commands/test/revert.rs
```diff
@@ -22,12 +24,23 @@ pub fn run(shell: &Shell, args: RevertArgs) -> anyhow::Result<()> {
...
     Cmd::new(cmd!(shell, "yarn utils build")).run()?;
+
+    // Build external node if necessary
+    if args.external_node {
+        Cmd::new(cmd!(
+            shell,
+            "cargo build --release --bin zksync_external_node"
+        ))
+        .run()?;
+    }
```
Deniallugo, line 42: why do we need it?

Deniallugo: we only need to run, not build

etc/utils/src/file-configs.ts
```diff
@@ -114,3 +121,12 @@ export function getConfigsFolderPath({
...
 }
+
+export function replaceAggregatedBlockExecuteDeadline(pathToHome: string, fileConfig: any, value: number) {
+    const generalConfigPath = getConfigPath({ pathToHome, chain: fileConfig.chain, config: 'general.yaml' });
+    const generalConfig = fs.readFileSync(generalConfigPath, 'utf8');
+    const regex = /aggregated_block_execute_deadline:\s*\d+/g;
+    const newGeneralConfig = generalConfig.replace(regex, `aggregated_block_execute_deadline: ${value}`);
+
+    fs.writeFileSync(generalConfigPath, newGeneralConfig, 'utf8');
+}
```
Deniallugo, line 132: why you are using regex? 
you could properly parse yaml

manuelmauro: At first I tried to parse the yaml file js' `yaml` package, modify the property and serialize the JS object back to a yaml file, but I encountered some incompatibility:
- for some reason js' yaml package would would convert a `1.0` to an integer when deserializing.
- and would interpret every address `0x...` as an hex value and convert it to an integer

I tried different deserialization oprions ([schema options](https://eemeli.org/yaml/#document-options): core, failsafe, and json), but they all fall short when it's time to deserialize the new yaml file in rust.

zk_toolbox/crates/zk_supervisor/src/commands/test/revert.rs
manuelmauro: Addressed in https://github.com/matter-labs/zksync-era/pull/2408/commits/50abe0c6bec18b21b3f714e3c98dd082150ce3b3

Deniallugo (approved):

PR #2416, opened 2024-07-10
feat(consensus): add L2 contracts VM storage reader (BFT-434)

moshababo:
## What ❔

Adding `eth_call`-based storage reader for consensus L2 contracts.

## Why ❔

Enable consensus executor to read from L2 storage per specific block.

## Notes

* `consensus::storage::VMReader` added to incapsulate read-transactions dispatching against L2 storage.
* `consensus::storage::testonly::VMWriter` added in order to simulate contracts deployments and data initialization in unit tests.
* `consensus::storage::tests::test_vm_reader()` test added.
* Multiple visibility modifiers were changed in order to facilitate the new codebase dependency graph.

## TODOs
- [ ] Remove `contracts` git submodule custom branch reference once https://github.com/matter-labs/era-contracts/pull/555 is merged.

## Checklist


core/node/consensus/src/storage/vm_reader.rs
```diff
@@ -0,0 +1,341 @@
...
+use crate::storage::ConnectionPool;
+
+pub struct VMReader {
+    pool: ConnectionPool,
+    tx_sender: TxSender,
+    consensus_authority_contract: Contract,
+    validator_registry_address: Option<Address>,
+    validator_registry_contract: Contract,
+    attester_registry_address: Option<Address>,
+    attester_registry_contract: Contract,
```
popzxc, line 25: Do we need 3 dedicated contracts for consensus? Our contracts are already very sophisticated, so my fear is that it would complicate them even more.
I do not have concrete proposals, but I strongly suggest discussing the contracts design with @vladbochok and @StanislavBreadless 

Not actionable in this PR I guess.

core/node/consensus/src/storage/vm_reader.rs
```diff
@@ -0,0 +1,341 @@
...
+    pub async fn new(
+        ctx: &Ctx,
+        block_id: BlockId,
+        pool: ConnectionPool,
+        tx_sender: TxSender,
+        consensus_authority_address: Address,
+    ) -> Self {
+        let consensus_authority_contract = load_contract("contracts/l2-contracts/artifacts-zk/contracts/ConsensusAuthority.sol/ConsensusAuthority.json");
+        let validator_registry_contract = load_contract("contracts/l2-contracts/artifacts-zk/contracts/ValidatorRegistry.sol/ValidatorRegistry.json");
+        let attester_registry_contract = load_contract("contracts/l2-contracts/artifacts-zk/contracts/AttesterRegistry.sol/AttesterRegistry.json");
```
popzxc, line 38: `zksync_contracts` crate currently serves (somewhat) as a schema for our FS dependencies, so probably it would be better to load the contracts there.
You can create a module for `consensus_l2_contracts` and load them via `pub fn load_consensus_l2_contracts() -> Option<ConsensusContracts>`.

core/node/consensus/src/storage/vm_reader.rs
```diff
@@ -0,0 +1,341 @@
...
+            pool,
+            tx_sender,
+            consensus_authority_contract,
+            validator_registry_address: None,
+            validator_registry_contract,
+            attester_registry_address: None,
+            attester_registry_contract,
+        };
+
+        reader.validator_registry_address = Some(
```
popzxc, line 50: It seems rather not idiomatic that the structure is created in a half-baked state and then filled.
Maybe you want to split the structure into several ones to avoid that.

core/node/consensus/src/storage/vm_reader.rs
```diff
@@ -0,0 +1,341 @@
...
+        ctx: &Ctx,
+        block_id: BlockId,
+    ) -> Vec<CommitteeValidator> {
+        let mut validators = vec![];
+        let num_committee_validators = self.read_num_committee_validators(ctx, block_id).await;
+        for i in 0..num_committee_validators {
+            let committee_validator = self.read_committee_validator(ctx, block_id, i).await;
+            let validator = self
+                .read_validator(ctx, block_id, committee_validator.0)
+                .await;
```
popzxc, line 94: You seem to instantiate a new DB connection and VM instance for each read request, and it's a somewhat expensive operation. (Actually likely several ones, because `tx_sender` would also create connections internally).

If we want to actually run the VM, I'd expect us to do it efficiently, roughy:
```rust
let vm = self.instantiate_vm().await;
let num_committee_validators = self.read_num_committee_validators(&vm, block_id).await;
// ^ run on an already instantiated VM.
// ... same for the rest of the function.
```

core/node/consensus/src/storage/vm_reader.rs
```diff
@@ -0,0 +1,341 @@
...
+        );
+        let res = self.eth_call(ctx, block_id, tx).await;
+        let tokens = func.decode_output(&res).unwrap();
+        (
+            tokens[0].clone().into_uint().unwrap().as_usize(),
+            tokens[1].clone().into_bytes().unwrap(),
+            tokens[2].clone().into_bytes().unwrap(),
+            tokens[3].clone().into_bool().unwrap(),
+        )
+    }
```
popzxc, line 192: Probably it would be better to implement `zksync_types::web3::contract::Tokenize` on all the required structures instead of mixing parsing with the business logic.

core/node/consensus/src/storage/vm_reader.rs
```diff
@@ -0,0 +1,341 @@
...
+        .unwrap();
+        let call_overrides = CallOverrides {
+            enforced_base_fee: None,
+        };
+
+        let res = self
+            .tx_sender
+            .eth_call(block_args, call_overrides, tx)
+            .await
+            .unwrap();
```
popzxc, line 304: If we want to consider running VM to query contracts (as opposed to pre-calculating storage slots and reading storage directly), I think we need to do quite some preparation work:

- Extract `execution_vm` into a separate, general-purpose crate.
- Change the sandbox so that it can perform multiple requests on a single instance.
- Decouple it from `TxSender`.

Right now, `tx_sender` and dependency on `zksync_node_api_server` feel like a pretty huge hack, and I'm not sure if it's a good idea to depend on that.

popzxc (commented):
A few high-level comments. I understand that the PR is still in draft, but the comments are mostly about design.

core/node/consensus/src/storage/vm_reader.rs
RomanBrodetski: agreed - we now need to remember and rely on the fact that those `Option`s are actually `Some`s once the struct is instantiated. Additionally, it makes it harder to reason about the struct itself: I was confused why it has those addresses as `Option` - are there circumstances when they are not set?

core/node/consensus/src/storage/vm_reader.rs
brunoffranca: That's my fault. I did the design for the contracts and thought that smaller contracts would be preferred. They can be merged into a single larger contract though.

core/node/consensus/src/storage/vm_reader.rs
brunoffranca: @popzxc In your opinion, should we read the storage slots directly instead of using VM read transactions?

popzxc: I can certainly see pros of using VM with `eth_call`, but I believe that implementing it properly would take _a lot_ of engineering efforts. As you can see, the first try is implemented in a hacky way, and making a _proper_ implementation (e.g. detaching sandbox from API) is a complex task even for seasoned platform developers. I would really like to not use hacky ways, as they may backfire, especially given that consensus will be a critical component for the sequencer.

On the other hand, reading storage directly will never look as "good", but it's relatively easy to implement and cover with unit tests (e.g. change value via contract method, then read from storage directly) to make sure that the storage layout doesn't change.

If we can afford to spend time on extracting the sandbox from API, I'm all for it. If not, I would rather prefer direct storage access than hacks.

core/node/consensus/src/storage/vm_reader.rs
moshababo: I agree that it is awkward, but I preferred that over having 2 structs, or not being able to utilize methods (or, transforming them into functions). I'll re-evaluate that 🙂

moshababo: Outdated / irrelevant after https://github.com/matter-labs/zksync-era/pull/2416/commits/5a1a259907b63dc7ce611e9689325f76816e56c4.

core/node/consensus/src/storage/vm_reader.rs
moshababo: Changes here: https://github.com/matter-labs/zksync-era/pull/2416/commits/477ea5ec1acc95cede12a576d8d735b1ba1978dd

core/node/consensus/src/storage/vm_reader.rs
moshababo: Fixed here: https://github.com/matter-labs/zksync-era/pull/2416/commits/5a925c6aced18514645823aa4f3332f1bee354f2

core/node/consensus/src/storage/vm_reader.rs
moshababo: Yes, it was just a PoC. 

It was fixed in https://github.com/matter-labs/zksync-era/pull/2416/commits/a8779b55345bd52f41de32ecebf82e0321b44423, so that only one connection will be acquired per read session. It does not however (yet) re-use that connection within the execution sandbox. The connection is only used to load block args.

core/node/consensus/src/storage/vm_reader.rs
moshababo: I agree that the current solution is a workaround/hack as per the code dependency graph and the lack of execution sandbox instance re-use.

However, falling back to direct storage reads is worse IMHO.

The reads are not targeting U256 variables in which their storage slots’ keys can be easily pre-calculated, be fetched and decoded. It involves dynamic-sized arrays, structs, and dynamic-sized arrays within the structs. This implies packing/padding, which means to re-implement the VM codec. 

Meanwhile, I couldn’t find any relevant documentation, and there’s very limited code examples (mostly in `zksync_types::utils`), for much simpler use-cases. 

I went ahead to experiment with it, and couldn’t even retrieve a simple U256 variable slot, following canonical `solc` storage layout. I might have overlooked something, but couldn’t reverse-engineer the slots back to the pre-hashed keys. In the overall, I don’t think it’s a good idea to go forward with this given the complex data types involved. 

Also, I don’t think such approach is testable, because the layout is supposedly an implementation detail, unless there’s a clear open available specs, backwards compatibility, and lack of potential optimizations. 

Read-transactions are the standard way to read from the VM storage database, while incapsulating the layout and the codec. The consensus component functions as a standard web3 client in this regard, and having direct access to the database does not imply it should use it to bypass the standard interface. 

I suggest to continue with the current approach, and to implement your suggestions for the execution sandbox decoupling, in a follow-up PR. (see BFT-493)

core/node/consensus/src/storage/vm_reader.rs
```diff
@@ -0,0 +1,289 @@
...
+    Nonce, U256,
+};
+use zksync_web3_decl::types::H160;
+
+use crate::storage::ConnectionPool;
+
+/// A struct for reading data from consensus L2 contracts.
+pub struct VMReader {
+    pool: ConnectionPool,
+    tx_sender: TxSender,
```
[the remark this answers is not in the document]
moshababo: Once execution sandbox will be decoupled, we could use it with main DB connection pool, and no changes to `TxSender` would be required.

core/node/consensus/src/storage/vm_reader.rs
```diff
@@ -0,0 +1,289 @@
...
+        let tokens = func.decode_output(&res).context("decode_output()")?;
+        CommitteeAttester::from_tokens(tokens).context("CommitteeAttester::from_tokens()")
+    }
+
+    async fn eth_call(&self, block_args: BlockArgs, tx: L2Tx) -> Vec<u8> {
+        let call_overrides = CallOverrides {
+            enforced_base_fee: None,
+        };
+        self.tx_sender
+            .eth_call(block_args, call_overrides, tx)
```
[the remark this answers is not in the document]
moshababo: Yes, connections are acquired per-call by the execution sandbox. This is indeed suboptimal, and should be part of the needed refactoring. (BFT-493)

core/node/consensus/src/storage/vm_reader.rs
brunoffranca: +1

core/node/consensus/src/storage/testonly.rs
```diff
@@ -189,3 +195,110 @@ impl ConnectionPool {
         Ok(())
     }
 }
+
+/// A struct for writing to consensus L2 contracts.
+pub struct VMWriter {
```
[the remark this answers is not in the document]
moshababo: Fixed here: https://github.com/matter-labs/zksync-era/pull/2416/commits/899f07bdc9c7b2ef65b7b7b103945ca6c4542c5c

core/node/consensus/src/storage/vm_reader.rs
```diff
@@ -0,0 +1,289 @@
...
+    l2::L2Tx,
+    transaction_request::CallOverrides,
+    Nonce, U256,
+};
+use zksync_web3_decl::types::H160;
+
+use crate::storage::ConnectionPool;
+
+/// A struct for reading data from consensus L2 contracts.
+pub struct VMReader {
```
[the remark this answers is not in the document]
moshababo: Added here: https://github.com/matter-labs/zksync-era/pull/2416/commits/d4e6f84fa19bd023154c4438e12f92c3ca92d3e7

It was added to the actual `read_committees` method and not to the struct, which I thought can remain agnostic.

core/node/consensus/src/storage/vm_reader.rs
```diff
@@ -0,0 +1,289 @@
...
+        }
+    }
+
+    /// Reads validator and attester committees from the registry contract.
+    pub async fn read_committees(
+        &self,
+        ctx: &Ctx,
+        block_id: BlockId,
+    ) -> anyhow::Result<(Vec<CommitteeValidator>, Vec<CommitteeAttester>)> {
+        let mut conn = self.pool.connection(ctx).await.unwrap().0;
```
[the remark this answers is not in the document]
moshababo: Fixed here: https://github.com/matter-labs/zksync-era/pull/2416/commits/cbc46a15a5dc4c988603f3fcdb1868e5963dfa57

core/node/consensus/src/storage/vm_reader.rs
popzxc: I don't mind as long as refactoring will really be planned & executed.

core/node/consensus/src/storage/testonly.rs
```diff
@@ -1,18 +1,24 @@
 //! Storage test helpers.
 
 use anyhow::Context as _;
-use zksync_concurrency::{ctx, error::Wrap as _, time};
+use zksync_concurrency::{ctx, ctx::Ctx, error::Wrap as _, time};
```
pompon0, line 4: nit: don't import ctx::Ctx, use the qualified name.

core/node/consensus/src/storage/testonly.rs
```diff
@@ -189,3 +195,111 @@ impl ConnectionPool {
...
+            account,
+        }
+    }
+
+    /// Deploys the consensus registry contract and adds nodes to it.
+    pub async fn deploy_and_add_nodes(
+        &mut self,
+        ctx: &Ctx,
+        owner: Address,
+        nodes: &[&[Token]],
```
pompon0, line 222: can the conversion be done inside the function, passing &[&[Token]] is very loosely typed.

core/node/consensus/src/storage/testonly.rs
```diff
@@ -189,3 +195,111 @@ impl ConnectionPool {
...
+    fn gen_tx_set_attester_committee(
+        &mut self,
+        contract_address: Address,
+        contract: &Contract,
+    ) -> Transaction {
+        let calldata = contract
+            .function("setAttesterCommittee")
+            .unwrap()
+            .short_signature()
+            .to_vec();
```
pompon0, line 290: all these methods wrapping raw contract API deserve a dedicated module.

core/node/consensus/src/storage/testonly.rs
```diff
@@ -189,3 +195,111 @@ impl ConnectionPool {
...
+            vec![],
+            TxType::L2,
+        );
+        txs.push(deploy_tx.tx);
+        for node in nodes {
+            let tx = self.gen_tx_add(&registry_contract.contract, deploy_tx.address, node);
+            txs.push(tx);
+        }
+        txs.push(
+            self.gen_tx_set_validator_committee(deploy_tx.address, &registry_contract.contract),
```
aakoshh, line 239: Is there any rason why the `contract` and the `address` parameters are in opposite order?

core/node/consensus/src/storage/testonly.rs
```diff
@@ -189,3 +195,111 @@ impl ConnectionPool {
...
+        self.gen_tx(contract_address, calldata)
+    }
+
+    fn gen_tx_set_validator_committee(
+        &mut self,
+        contract_address: Address,
+        contract: &Contract,
+    ) -> Transaction {
+        let calldata = contract
+            .function("setValidatorCommittee")
```
aakoshh, line 274: I can't see this method in https://github.com/matter-labs/era-contracts/pull/555 ; is it [commitValidators](https://github.com/matter-labs/era-contracts/blob/c13bd81974c0fbcb284328be6e10c739b77999e5/l2-contracts/contracts/ConsensusRegistry.sol#L329) ?

core/node/consensus/src/storage/tests.rs
```diff
@@ -0,0 +1,92 @@
...
+        for _ in 0..num_nodes {
+            let node_entry = (
+                Address::random(),
+                U256::from(rng.gen::<usize>()),
+                (0..256).map(|_| rng.gen()).collect::<Vec<u8>>(),
+                (0..256).map(|_| rng.gen()).collect::<Vec<u8>>(),
+                U256::from(rng.gen::<usize>()),
+                (0..256).map(|_| rng.gen()).collect::<Vec<u8>>(),
+            )
+                .into_tokens();
```
aakoshh, line 37: Maybe you could use [abigen!](https://docs.rs/ethers/latest/ethers/contract/macro.abigen.html) to generate bindings for all the Solidity structs and then you get the tokenization with static typing.

core/node/consensus/src/storage/vm_reader.rs
```diff
@@ -0,0 +1,294 @@
...
+            .clone();
+
+        let tx = self.gen_l2_call_tx(self.registry_address, func.short_signature().to_vec());
+
+        let res = self.eth_call(block_args, tx).await;
+
+        let tokens = func.decode_output(&res).context("decode_output()")?;
+        U256::from_tokens(tokens)
+            .context("U256::from_tokens()")
+            .map(|t| t.as_usize())
```
aakoshh, line 131: Would it be feasible to do a bit more encapsulation? For example to create a `DeployedContract` or similar type which has an `Address` and a `Contract`, then wrap that into a `RegistryContract` which has methods for these that take whatever is needed to do the `eth_call` but otherwise know which function to call and how to parse the result? Maybe we can use the `abigen!` macro to generate the bindings if the ABI JSON is available at compile time? 

The `VMReader` sounds more general than just having all these methods for this particular contract, even if it's the only one at the moment. 

I did something similar [here](https://github.com/aakoshh/ipc/blob/main/fendermint/vm/interpreter/src/fvm/state/ipc.rs) where a gateway contract implemented multiple facets, each with their own binding, each represented by a [contract caller](https://github.com/aakoshh/ipc/blob/f13c0b6ecdbe16a18e2f3bd7eb6e898cda50552a/fendermint/vm/interpreter/src/fvm/state/fevm.rs#L136) that took the equivalent of the `TxSender` as an input. I suppose that gateway is similar to the `VMReader`, but a lot of the boilerplate and the private methods here could be delegated to contract specific structs, leaving only the highest level stuff in the reader.

core/node/consensus/src/storage/vm_reader.rs
```diff
@@ -0,0 +1,294 @@
...
+    ) -> anyhow::Result<CommitteeValidator> {
+        let func = self
+            .registry_contract
+            .function("validatorCommittee")
+            .unwrap()
+            .clone();
+        let tx = self.gen_l2_call_tx(
+            self.registry_address,
+            func.encode_input(&zksync_types::U256::from(idx).into_tokens())
+                .unwrap(),
```
aakoshh, line 162: These unwraps could be at least `expect` with the method name to give us some context if they fail, in case we can't use bindings.

core/node/consensus/src/storage/vm_reader.rs
```diff
@@ -0,0 +1,294 @@
...
+        U256::from_tokens(tokens)
+            .context("U256::from_tokens()")
+            .map(|t| t.as_usize())
+    }
+
+    async fn read_attester_committee_size(&self, block_args: BlockArgs) -> anyhow::Result<usize> {
+        let func = self
+            .registry_contract
+            .function("attesterCommitteeSize")
+            .unwrap()
```
aakoshh, line 138: A contract wrapper could provide a method to get the function by name and make sure if it's not found the error explains what it was looking for.

core/node/consensus/src/storage/vm_reader.rs
```diff
@@ -0,0 +1,294 @@
...
+#[derive(Debug, Default)]
+#[allow(dead_code)]
+pub struct CommitteeValidator {
+    pub node_owner: Address,
+    pub weight: usize,
+    pub pub_key: Vec<u8>,
+    pub pop: Vec<u8>,
+}
+
+impl Detokenize for CommitteeValidator {
```
aakoshh, line 230: There is an attribute in `ethers` that should work here even without generating the bindings. Here's an [example](https://github.com/aakoshh/ipc/blob/f13c0b6ecdbe16a18e2f3bd7eb6e898cda50552a/fendermint/vm/actor_interface/src/ipc.rs#L343-L350).

core/node/consensus/src/storage/tests.rs
popzxc: We didn't migrate to `ethers` just yet (and in fact it's already deprecated in favor of `alloy`), so using it is not recommended for now. We want the migration to be done consistently, otherwise, we'll end up with a mix of our vendored `web3` remains and some of the newer stuff.

pompon0:
closing this pr, I'll continue this work in https://github.com/matter-labs/zksync-era/pull/2684

PR #2417, opened 2024-07-10
fix: Fix rustls setup for jsonrpsee clients

popzxc:
## What ❔

`jsonrpsee` client expects `rustls` to use the default crypto provider, but `rustls` cannot choose and panics if there are multiple providers available. See [this issue](https://github.com/rustls/rustls/issues/1877) for more detail.
Recently `rustls` changed the default backend, so now some (older) of crates use `ring` backend, while some (newer) use `aws-lc-rs`.
Until the issue is fixed on `jsonrpsee` side, we're making sure that the provider is installed before we create any `jsonrpsee` client.

Added test used to fail before the change and passes now.

## Why ❔

Panics are bad.

## Checklist

### DIFF
--- a/prover/Cargo.lock
+++ b/prover/Cargo.lock
@@ -260,6 +260,33 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0c4b4d0bd25bd0b74681c0ad21497610ce1b7c91b1022cd21c80c6fbdd9476b0"
 
 [[package]]
+name = "aws-lc-rs"
+version = "1.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


perekopskiy (approved):

prover/Cargo.lock
```diff
@@ -8460,6 +8507,7 @@ dependencies = [
  "jsonrpsee",
  "pin-project-lite",
  "rlp",
+ "rustls",
  "serde",
  "serde_json",
  "thiserror",
```
RomanBrodetski, line 8513: are changes to this file done on purpose?

popzxc: Yes; can elaborate on "why" if needed.

PR #2425, opened 2024-07-10
feat(eth-sender): add early return in sending new transactions to not spam logs with errors

tomg10:


core/node/eth_sender/src/eth_tx_manager.rs
```diff
@@ -586,8 +586,16 @@ impl EthTxManager {
                 .await
                 .unwrap();
 
+            tracing::info!("Sending {} new transactions", new_eth_tx.len());
```
perekopskiy, line 589: let's also log operator_type

perekopskiy (approved):

perekopskiy (approved):

PR #2427, opened 2024-07-11
feat(en): Switch EN to use node framework

popzxc:
## What ❔

This PR ports remaining parts of the EN to node framework and makes framework the default way to run it.
In more detail:

- Config for the health check limits are now set from config.
- EN and rust metrics are now exposed; the protocol version update task now runs.
- ⚠️ Connection pool healthcheck was removed. It was controversial initially, its usefulness is not clear, it was supposed to be refactored a year ago but didn't, and it wasn't working well when testing. See [linear issue](https://linear.app/matterlabs/issue/PLA-255/revamp-db-connection-health-check) for more context.
- Tests were reworked to use node framework; some refactoring was also applied to reduce boilerplate.
- Additional tests were added to check for invalid EN configurations.
- ⚠️ Node framework was made the default way to run the EN. There is also a hook to force EN to run the old way, so that we don't have to rollback over small issues.

## Why ❔

- Part of switch to the node framework.

## Checklist


core/lib/health_check/src/lib.rs
```diff
@@ -118,30 +123,59 @@ impl Default for AppHealthCheck {
...
+            components: Vec::default(),
+            slow_time_limit,
+            hard_time_limit,
+        };
+        Self {
+            inner: Mutex::new(inner),
+        }
+    }
+
+    pub fn override_limits(
```
popzxc, line 144: Both this method and `expose_metrics` are somewhat clunky; I think it would be easier to refactor it once we fully switch to the framework and remove the old code.

core/bin/external_node/src/tests/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+        assert_matches!(component_health.status(), HealthStatus::ShutDown);
+    }
+}
+
+#[tokio::test]
+async fn node_reacts_to_stop_signal_during_initial_reorg_detection() {
+    let _guard = zksync_vlog::ObservabilityBuilder::new().build(); // Enable logging to simplify debugging
+    let (env, env_handles) = utils::TestEnvironment::with_genesis_block("core").await;
+
+    let l2_client = MockClient::builder(L2::default())
```
slowli, line 119: Nit: Could make sense to extract this to a function similarly to the L1 client.

core/lib/vm_utils/src/storage.rs
```diff
@@ -83,18 +83,41 @@ pub fn l1_batch_params(
...
 impl L1BatchParamsProvider {
-    pub async fn new(storage: &mut Connection<'_, Core>) -> anyhow::Result<Self> {
-        let snapshot = storage
+    pub fn new() -> Self {
+        Self { snapshot: None }
+    }
+
+    /// Performs the provider initialization. Must only be called with the initialized storage (e.g.
+    /// either after genesis or snapshot recovery).
+    pub async fn initialize(&mut self, storage: &mut Connection<'_, Core>) -> anyhow::Result<()> {
```
slowli, line 98: Bikeshedding: I'd rather retain the `new()` method as-is and create a new `uninitialized()` constructor (or `Default` implementation), so that it's easier to initialize the provider correctly.

core/node/node_storage_init/src/external_node/revert.rs
```diff
@@ -22,8 +22,14 @@ impl RevertStorage for ExternalNodeReverter {
         to_batch: L1BatchNumber,
         _stop_receiver: watch::Receiver<bool>,
     ) -> anyhow::Result<()> {
+        let Some(block_reverter) = self.reverter.as_ref() else {
+            anyhow::bail!(
+                "Revert to block {to_batch} was requested, but the reverter was not provided."
+            );
+        };
```
slowli, line 29: Nit: This can be more concisely expressed by using `self.reverter.as_ref().context("...")?`.

core/bin/external_node/src/tests/mod.rs
slowli, line 1: Nice job adapting tests 👍

core/bin/external_node/src/tests/framework.rs
```diff
@@ -0,0 +1,161 @@
...
+}
+
+#[async_trait::async_trait]
+impl WiringLayer for MockL1ClientLayer {
+    type Input = ();
+    type Output = EthInterfaceResource;
+
+    fn layer_name(&self) -> &'static str {
+        // We don't care about values, we just want to hijack the layer name.
+        QueryEthClientLayer::new(L1ChainId(1), "https://example.com".parse().unwrap()).layer_name()
```
slowli, line 130: Maybe it makes sense to expose the layer name as a constant? Or would it be too single-purpose?

popzxc: Yeah, I did it that way initially, but it looked clunky -- just a few layers that have this constant. Plus we cannot enforce it anyway to rely on it for future similar uses.
It's a shame that static methods in traits break object safety :(

core/lib/vm_utils/src/storage.rs
popzxc: I just thought that it matches the `StateKeeperIO` well, which also has separate methods to create and initialize the object.
Actually, the fact that this structure didn't have an `initialize` method was one of the reasons why I was _extremely_ confused by the `ExternalIO` not picking up the previous batch hash and it took a while to understand what was going on.
Having a method that may either work or not depending on circumstances (which are also pretty rare) sounds a bit dangerous.

RomanBrodetski (approved):

popzxc:
I will fix nitpicks in a follow-up PR.

Deniallugo (approved):

core/bin/external_node/src/main.rs
```diff
@@ -725,10 +722,6 @@ struct Cli {
     external_node_config_path: Option<std::path::PathBuf>,
     /// Path to the yaml with consensus.
     consensus_path: Option<std::path::PathBuf>,
-
-    /// Run the node using the node framework.
-    #[arg(long)]
-    use_node_framework: bool,
```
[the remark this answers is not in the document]
popzxc: Of course. EN wasn't fully ported to the framework until this PR, it wouldn't have worked with it

PR #2431, opened 2024-07-11
fix: Fill `storage_refunds` / `pubdata_costs` data

slowli:
## What ❔

These fields don't seem to be used by ENs, but are expected to be filled during the main state keeper execution cycle.


slowli (commented):
The relevant branch in the `vm2` repo is [this one](https://github.com/matter-labs/vm2/commits/aov/io-batch-stats); probably needs to be merged before this PR.

core/lib/multivm/src/versions/vm_fast/tests/tester/vm_tester.rs
```diff
@@ -69,9 +73,19 @@ impl VmTester {
...
+        let storage = self.storage.clone();
+        {
+            let mut storage = storage.borrow_mut();
+            // Commit pending storage changes (old VM versions commit them on successful execution)
+            for (&(address, slot), &value) in self.vm.inner.world_diff.get_storage_state() {
+                let key = StorageKey::new(AccountTreeId::new(address), u256_to_h256(slot));
+                storage.set_value(key, u256_to_h256(value));
+            }
+        }
+
```
montekki, line 85: Why exactly is this happening in the `reset_state`?

montekki:
I have used this patchset to fix the `require_eip712` test with this diff:

```patch
diff --git a/core/lib/multivm/src/versions/vm_fast/tests/require_eip712.rs b/core/lib/multivm/src/versions/vm_fast/tests/require_eip712.rs
index 97f929927..fce3404f0 100644
--- a/core/lib/multivm/src/versions/vm_fast/tests/require_eip712.rs
+++ b/core/lib/multivm/src/versions/vm_fast/tests/require_eip712.rs
@@ -23,7 +23,13 @@ impl VmTester {
             AccountTreeId::new(L2_BASE_TOKEN_ADDRESS),
             &address,
         );
-        h256_to_u256(self.vm.world.storage.read_value(&key))
+        self.vm
+            .inner
+            .world_diff
+            .get_storage_state()
+            .get(&(L2_BASE_TOKEN_ADDRESS, h256_to_u256(*key.key())))
+            .cloned()
+            .unwrap_or(h256_to_u256(self.vm.world.storage.read_value(&key)))
     }
 }
 
@@ -134,7 +140,8 @@ async fn test_require_eip712() {
         Default::default(),
     );
 
-    let transaction_request: TransactionRequest = tx_712.into();
+    let mut transaction_request: TransactionRequest = tx_712.into();
+    transaction_request.chain_id = Some(chain_id.into());
 
     let domain = Eip712Domain::new(L2ChainId::from(chain_id));
     let signature = private_account

```

I am pasting this here since I have no better place to do it atm I suppose

core/lib/multivm/src/versions/vm_fast/tests/tester/vm_tester.rs
montekki: yeah but this relies on the user of the tester to call the reset at appropriate times and only then checking expected values in storage, this can be easily misused?

slowli: IIUC, it's not really the case, we generally use `WorldDiff`-aware methods to check storage slots, like [this one](https://github.com/matter-labs/zksync-era/blob/d66a4a49d454f0b5fed4cfb80898281b2711bc1d/core/lib/multivm/src/versions/vm_fast/tests/utils.rs#L26). It's just that specifically resetting a VM in existing tests implies that all storage changes are committed, which is difficult to reproduce without committing these changes (since the storage is accessed not by the test code, but by a subsequent VM run in this case).

core/lib/multivm/src/versions/vm_fast/vm.rs
```diff
@@ -503,66 +499,67 @@ impl<S: ReadStorage> VmInterface for Vm<S> {
             // Move the pointer to the next transaction
             self.bootloader_state.move_tx_to_execute_pointer();
             track_refunds = true;
-            // Create a snapshot to roll back the bootloader call frame on halt.
-            self.make_snapshot();
```
montekki, line 507: What is the reason for removing this snapshot and the rollbacks further in this function?

slowli: I've introduced this additional snapshot in [my previous PR](https://github.com/matter-labs/zksync-era/pull/2378/files), and it looks redundant now that I understand VM function better 🙃 The only thing this snapshot did was zeroing the VM outputs before a rollback of the original snapshot (which would occur later), and this is easy to reproduce without taking a snapshot. Another option would be to ignore the VM outputs on the caller side, but it looks more complex to implement / maintain.

slowli:
@joonazan Are VM performance changes OK / expected, or did I do something wrong with the integration?

joonazan:
@slowli Hard to say. This shouldn't improve performance, right?
Probably it is because of using merges rather than rebasing, so the start of the branch is older than the end.

Help me merge this and we will at least know that the same number of instructions are executed: https://github.com/matter-labs/zksync-era/pull/2276

core/lib/multivm/src/versions/vm_fast/tests/code_oracle.rs
```diff
@@ -91,6 +93,16 @@ fn test_code_oracle() {
     assert!(!result.result.is_failed(), "Transaction wasn't successful");
 }
 
+fn find_code_oracle_cost_log(
+    precompiles_contract_address: Address,
+    logs: &[StorageLogWithPreviousValue],
+) -> &StorageLogWithPreviousValue {
+    let log = logs.iter().find(|log| {
+        *log.log.key.address() == precompiles_contract_address && *log.log.key.key() == H256::zero()
```
montekki, line 101:
```suggestion
        *log.log.key.address() == precompiles_contract_address && log.log.key.key().is_zero()
```

[`is_zero()`](https://docs.rs/primitive-types/latest/primitive_types/struct.H256.html#method.is_zero)

core/lib/multivm/src/versions/vm_fast/tests/code_oracle.rs
```diff
@@ -147,3 +159,80 @@ fn test_code_oracle_big_bytecode() {
...
+                    .unwrap(),
+                value: U256::zero(),
+                factory_deps: vec![],
+            },
+            None,
+        );
+
+        vm.vm.push_transaction(tx);
+        let result = vm.vm.execute(VmExecutionMode::OneTx);
+        assert!(!result.result.is_failed(), "Transaction wasn't successful");
```
montekki, line 225:
```suggestion
        assert!(!result.result.is_failed(), "Transaction wasn't successful: {result:#?}");
```

that could save a bit of time

core/lib/multivm/src/versions/vm_fast/vm.rs
joonazan: Is the actual snapshotting done in the sequencer now?

core/lib/multivm/src/versions/vm_fast/vm.rs
```diff
@@ -460,7 +457,6 @@ impl<S: ReadStorage> Vm<S> {
         inner.state.current_frame.sp = 0;
 
         // The bootloader writes results to high addresses in its heap, so it makes sense to preallocate it.
-        inner.state.heaps[vm2::FIRST_HEAP].reserve(get_used_bootloader_memory_bytes(VM_VERSION));
```
montekki, line 463: reservation is removed but the comment is left, maybe should also be removed. besides, what is the reason for this removal?

core/lib/multivm/src/versions/vm_fast/vm.rs
```diff
@@ -503,66 +499,67 @@ impl<S: ReadStorage> VmInterface for Vm<S> {
...
-                    .iter()
-                    .map(|x| x.glue_into())
-                    .collect(),
+                system_l2_to_l1_logs,
                 total_log_queries_count: 0, // This field is unused
-            },
+            }
+        };
+
+        // FIXME: are statistics rolled back on halt as well?
```
montekki, line 558: nitpick: either a tracking issue has to be created and linked here or `FIXME` should be resolved and removed.

core/lib/multivm/src/versions/vm_latest/tests/code_oracle.rs
```diff
@@ -102,6 +107,16 @@ fn test_code_oracle() {
     assert!(!result.result.is_failed(), "Transaction wasn't successful");
 }
 
+fn find_code_oracle_cost_log(
+    precompiles_contract_address: Address,
+    logs: &[StorageLogWithPreviousValue],
+) -> &StorageLogWithPreviousValue {
+    let log = logs.iter().find(|log| {
+        *log.log.key.address() == precompiles_contract_address && *log.log.key.key() == H256::zero()
```
montekki, line 115:
```suggestion
        *log.log.key.address() == precompiles_contract_address && log.log.key.key().is_zero()
```
same as above

core/lib/multivm/src/versions/vm_fast/vm.rs
```diff
@@ -503,66 +499,67 @@ impl<S: ReadStorage> VmInterface for Vm<S> {
...
-                            key: StorageKey::new(AccountTreeId::new(address), u256_to_h256(key)),
-                            value: u256_to_h256(change.after),
-                            kind: if change.is_initial {
-                                StorageLogKind::InitialWrite
-                            } else {
-                                StorageLogKind::RepeatedWrite
-                            },
+        // If the execution is halted, the VM changes are expected to be rolled back by the caller.
+        // Earlier VMs return empty execution logs in this case, so we follow this behavior.
+        let logs = if ignore_world_diff {
```
joonazan, line 513: I can't find anything like this in vm_latest. Are you sure?

core/lib/multivm/src/versions/vm_latest/tests/code_oracle.rs
```diff
@@ -102,6 +107,16 @@ fn test_code_oracle() {
...
+    });
+    log.expect("no code oracle cost log")
```
joonazan, line 117: Nit: don't create the log variable.

core/lib/multivm/src/versions/vm_fast/vm.rs
joonazan: It definitely shouldn't be removed here. It is removed in my PR because its performance impact is minimal and it becomes tricky to do.

joonazan:
I now understand why the benchmarks are faster. It is because the snapshotting is removed.

core/lib/multivm/src/versions/vm_fast/vm.rs
slowli: Good catch! That's essentially a merge artifact at this point.

core/lib/multivm/src/versions/vm_fast/vm.rs
slowli: It was performed in the sequencer previously as well, but it's now performed *entirely* there, yes.

montekki (approved):

PR #2442, opened 2024-07-16
refactor(external_price_api): force erc20 in coingecko's api query

fborello-lambda:
## What ❔

Integrate the `external_price_api` with a dev environment, such as `dev` config or even a different L1 like `sepolia`.

## Why ❔

Provide a workaround and a possible solution.

### Provide a way to force the desired token to calculate the conversion rate.

```toml
[external_price_api_client]
forced_ethereum_erc20_address = "forced_token"
```

And in `core/lib/config/src/configs/external_price_api_client.rs`:

```rust
pub struct ExternalPriceApiClientConfig {
    // [...]
    pub forced_ethereum_erc20_address: Option<String>,
}
```

### Change header if the coingecko api is using the `demo` base_url.

In `core/lib/external_price_api/src/coingecko_api.rs`:

```rust
const PRO_COINGECKO_API_URL: &str = "https://pro-api.coingecko.com";
const DEMO_COINGECKO_API_URL: &str = "https://api.coingecko.com";
const PRO_COINGECKO_AUTH_HEADER: &str = "x-cg-pro-api-key";
const DEMO_AUTH_HEADER: &str = "x-cg-demo-api-key";
//[...]
impl CoinGeckoPriceAPIClient {
    pub fn new(config: ExternalPriceApiClientConfig) -> Self {
        //[...]
        let header = match base_url.as_str() {
            PRO_COINGECKO_API_URL => PRO_COINGECKO_AUTH_HEADER,
            DEMO_COINGECKO_API_URL => DEMO_AUTH_HEADER,
            _ => PRO_COINGECKO_API_URL,
        };
//[...]
```

### Provide according gasprice

In `core/lib/types/src/fee_model.rs`:

```rust
    fn convert_to_base_token(&self, price_in_wei: u64) -> u64 {
        let conversion_ratio = BigDecimal::from(self.conversion_ratio.numerator.get())
            / BigDecimal::from(self.conversion_ratio.denominator.get());
        let converted_price_bd = BigDecimal::from(price_in_wei) / conversion_ratio;
```

Before, the `converted_price_bd` was:

```rust
let converted_price_bd = BigDecimal::from(price_in_wei) * conversion_ratio;
```

Offering a lower gas price compared to an `ETH_BASED` chain, resulting in negligible gas deductions.

## How to reproduce

Using the `etc/env/configs/dev.toml` config as example:

1. Set the `base_token_addr`

The dev `zk init` automatically deploys some ERC20 contracts, the addresses are written in `etc/tokens/localhost.json`.

It uses `create2` so the addresses persist.

If we use `DAI` as example:

```toml
__imports__ = [ "base", "l1-inits/.init.env", "l2-inits/dev.init.env" ]

ETH_SENDER_SENDER_PUBDATA_SENDING_MODE = "Blobs"

# DAI ERC20 address
BASE_TOKEN_ADDR = "0x70a0F165d6f8054d0d0CF8dFd4DD2005f0AF6B55"

[external_price_api_client]
source = "coingecko"
base_url = "https://api.coingecko.com"
api_key = "your_api"
# Since the coingecko api only works for ethereum mainnet and with the dev config we run a reth instance as L1,
# we can force the erc20 address used to get the conversion_rate, the address below is the DAI contract of ethereum mainnet
forced_ethereum_erc20_address = "0x6b175474e89094c44da98b954eedeac495271d0f"
```

2. Setup the environment:

```sh
export ZKSYNC_HOME=$(pwd)
export PATH=$ZKSYNC_HOME/bin:$PATH
```

3. Now we can launch the `zk server`:

```sh
zk && zk clean --all && zk env dev && zk init --base-token-name DAI && zk server --components=api,eth,tree,state_keeper,housekeeper,commitment_generator,proof_data_handler,base_token_ratio_persister
```

4. Try it out
   1. If using the `dev config`, we will have to mint some tokens, and transfer the `ERC20` from L1 to L2. Proposed
      script: [GIST](https://gist.github.com/fborello-lambda/3804938f2b43705b34715cd610e30d2d)
   2. Then we can do a transfer to test that the gasprice is according to the `ERC20` used.

The output of the transaction using the `zksync-ethers` sdk:

```
ZK Network URL: http://127.0.0.1:3050
ZK Network ChainID: 270
(from) balance before tx: 2.000062297325
  (to) balance before tx: 0.00060582619375
Send 1 token
(from): 0x36615Cf349d7F6344891B1e7CA7C72883F5dc049
  (to): 0xde03a0B5963f75f1C8485B355fF6D30f3093BDE7
#####################################################
Tx hash: 0x11df0add38b5ed8276c9652b8405f2ce0b3a507445e3504c955e159c42eae66a
#####################################################
(from) balance after tx: 0.953929033219829376
  (to) balance after tx: 1.00060582619375
#####################################################
```

1 base_token (`DAI`) is being transferred, with a fee of 0.05 `DAI` deducted.

## Note

The ERC20 token must be defined in `etc/tokens/<L1>.json` after executing `zk init`. This limits the default availability of tokens that can be used. Users should be informed about these required steps because failing to define the token could lead to contract deployment failures and `unintended additional costs`.

## Checklist

### DIFF
--- a/core/lib/external_price_api/src/coingecko_api.rs
+++ b/core/lib/external_price_api/src/coingecko_api.rs
@@ -132,17 +132,13 @@ impl CoinGeckoPriceResponse {
             .and_then(|price| price.get(currency))
     }
 }
+
+#[cfg(test)]
 mod test {
-    use std::str::FromStr;
-
-    use zksync_config::configs::ExternalPriceApiClientConfig;
-    use zksync_types::Address;
-
-    use super::CoinGeckoPriceAPIClient;
+    use super::*;
 
     #[tokio::test]
-    // This is not a proper test, as it requires user input and prints information.
-    // It serves as a quick check to verify if the CoinGecko api_key is working correctly.
+    #[ignore = "This is not a proper test, as it requires user input and prints information.\nIt serves as a quick check to verify if the CoinGecko api_key is working correctly."]
     async fn test_coingecko_api() {
         let config = ExternalPriceApiClientConfig {
             source: "coingecko".to_string(),
--- a/core/lib/external_price_api/src/coingecko_api.rs
+++ b/core/lib/external_price_api/src/coingecko_api.rs
@@ -1,4 +1,4 @@
-use std::collections::HashMap;
+use std::{collections::HashMap, str::FromStr};
 
 use async_trait::async_trait;
 use chrono::Utc;
@@ -14,18 +14,31 @@ use crate::{address_to_string, utils::get_fraction, PriceAPIClient};
 pub struct CoinGeckoPriceAPIClient {
     base_url: Url,
     client: reqwest::Client,
+    forced_ethereum_erc20_address: Option<Address>,
 }
 
-const DEFAULT_COINGECKO_API_URL: &str = "https://pro-api.coingecko.com";
-const COINGECKO_AUTH_HEADER: &str = "x-cg-pro-api-key";
+const PRO_COINGECKO_API_URL: &str = "https://pro-api.coingecko.com";
+const DEMO_COINGECKO_API_URL: &str = "https://api.coingecko.com";
+const PRO_COINGECKO_AUTH_HEADER: &str = "x-cg-pro-api-key";
+const DEMO_AUTH_HEADER: &str = "x-cg-demo-api-key";
 const ETH_ID: &str = "eth";
 
 impl CoinGeckoPriceAPIClient {
     pub fn new(config: ExternalPriceApiClientConfig) -> Self {
+        let base_url = config
+            .base_url
+            .clone()
+            .unwrap_or(PRO_COINGECKO_API_URL.to_string());
         let client = if let Some(api_key) = &config.api_key {
+            let header = match base_url.as_str() {
+                PRO_COINGECKO_API_URL => PRO_COINGECKO_AUTH_HEADER,
+                DEMO_COINGECKO_API_URL => DEMO_AUTH_HEADER,
+                _ => PRO_COINGECKO_API_URL,
+            };
+
             let mut headers = reqwest::header::HeaderMap::new();
             headers.insert(
-                reqwest::header::HeaderName::from_static(COINGECKO_AUTH_HEADER),
+                reqwest::header::HeaderName::from_static(header),
                 reqwest::header::HeaderValue::from_str(api_key)
                     .expect("Failed to create header value"),
             );
@@ -39,18 +52,27 @@ impl CoinGeckoPriceAPIClient {
             reqwest::Client::new()
         };
 
-        let base_url = config
-            .base_url
-            .unwrap_or(DEFAULT_COINGECKO_API_URL.to_string());
+        let erc20 = config.forced_ethereum_erc20_address;
+        let forced_ethereum_erc20_address: Option<Address> = match erc20 {
+            Some(erc20) => match Address::from_str(erc20.as_str()) {
+                Ok(address) => Some(address),
+                Err(_) => None,
+            },
+            None => None,
+        };
 
         Self {
             base_url: Url::parse(&base_url).expect("Failed to parse CoinGecko URL"),
             client,
+            forced_ethereum_erc20_address,
         }
     }
 
     async fn get_token_price_by_address(&self, address: Address) -> anyhow::Result<f64> {
-        let address_str = address_to_string(&address);
+        let address_str: String = match self.forced_ethereum_erc20_address {
+            Some(erc20) => address_to_string(&erc20),
+            None => address_to_string(&address),
+        };
         let price_url = self
             .base_url
             .join(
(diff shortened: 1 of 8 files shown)


core/lib/external_price_api/src/coingecko_api.rs
```diff
@@ -110,3 +132,38 @@ impl CoinGeckoPriceResponse {
...
+        };
+        let api = CoinGeckoPriceAPIClient::new(config);
+
+        println!(
+            "{:?}",
+            api.get_token_price_by_address(
+                Address::from_str("0x1f9840a85d5af5bf1d1762f925bdaddc4201f984").unwrap()
+            )
+            .await
+            .unwrap()
```
popzxc, line 166: General observation: such tests are prone to be very flaky because the availability of 3rd party API is not predictable.
If we want to keep it, it's better to mark it as `ignore` so that it only runs if `--ignored` flag is provided.

core/lib/external_price_api/src/coingecko_api.rs
```diff
@@ -110,3 +132,38 @@ impl CoinGeckoPriceResponse {
...
+    use std::str::FromStr;
+
+    use zksync_config::configs::ExternalPriceApiClientConfig;
+    use zksync_types::Address;
+
+    use super::CoinGeckoPriceAPIClient;
+
+    #[tokio::test]
+    // This is not a proper test, as it requires user input and prints information.
+    // It serves as a quick check to verify if the CoinGecko api_key is working correctly.
```
popzxc, line 145: Do I get it right that the expected workflow here is to change `api_key` and `forced_ethereum_erc20_address` to some valid values, then run the test and manually see if it produces some output?
If so, I think it's better to get input via environment variables.

core/lib/external_price_api/src/coingecko_api.rs
```diff
@@ -39,18 +52,27 @@ impl CoinGeckoPriceAPIClient {
...
+            forced_ethereum_erc20_address,
         }
     }
 
     async fn get_token_price_by_address(&self, address: Address) -> anyhow::Result<f64> {
-        let address_str = address_to_string(&address);
+        let address_str: String = match self.forced_ethereum_erc20_address {
+            Some(erc20) => address_to_string(&erc20),
+            None => address_to_string(&address),
+        };
```
popzxc, line 75: Do I get it correctly that idea here is to override token address with some token that is present on mainnet?

core/lib/types/src/fee_model.rs
```diff
@@ -282,7 +282,7 @@ impl FeeParamsV2 {
     fn convert_to_base_token(&self, price_in_wei: u64) -> u64 {
         let conversion_ratio = BigDecimal::from(self.conversion_ratio.numerator.get())
             / BigDecimal::from(self.conversion_ratio.denominator.get());
-        let converted_price_bd = BigDecimal::from(price_in_wei) * conversion_ratio;
+        let converted_price_bd = BigDecimal::from(price_in_wei) / conversion_ratio;
```
popzxc, line 285: The fetched ratio [seems to be](https://github.com/matter-labs/zksync-era/blob/6146fce18673df89485d96c5b38a00d6404aa4b2/core/lib/external_price_api/src/coingecko_api.rs#L80) `token/eth`. So if I understand things correctly, before the unit of calculation was `eth * token / eth = token`, and now it's `eth / (token / eth) = eth * eth / token = eth^2 / token`. Am I wrong?

shahar4: This understanding as well as the existing code are both correct. A simple way to understand this is: the denomination we need stay with is `baseToken` so we need to "cancel out" the ETH denomination.

```
conversion_ratio = numerator(in baseToken)/denominator(in ETH)
gasPrice(in baseToken) = gasPrice(in ETH) * conversion_ratio
```

@ilitteri does it make sense to you?

shahar4 (commented):
I don't think I support the premise of this PR. The benefit of this PR, as I see it, is to be able to use CoinGecko in lower environments. A few thoughts:
- The intention for lower envs was to run with the `ForcedPriceClient`. [See here](https://github.com/matter-labs/zksync-era/blob/main/core/lib/external_price_api/src/forced_price_client.rs#L37)
- I absolutely see a need to test the CoinGecko integration, but don't think it should be more than a test
- CoinGecko should be one of several APIs used by a chain in mainnet
- So to me, this would add relatively a lot of code for not a great reason

Am I missing something? WDYT?

core/lib/types/src/fee_model.rs
fborello-lambda: I've followed this reasoning:
1 usdt &rarr; 1 usd
1 eth &rarr; 3500usd

If i want to have how many `eths` do i need to buy 1 usdt:
$$\text{Price of base token[eth]} = \frac{1[usd] \times 1[eth]}{3500[usd]} = 0.0002857 [eth]$$ 
Which is roughly the value obtained from `coingecko`:
- https://docs.coingecko.com/v3.0.1/reference/simple-token-price

```json
{
  "0xdac17f958d2ee523a2206206994597c13d831ec7": {
    "eth": 0.0002857
  }
}
```

Now, if the gasprice is in `eth`:
 $$\text{gasprice[usdt]} = \frac{\text{gasprice[eth]}}{\text{Price of base token[eth}]}$$

core/lib/external_price_api/src/coingecko_api.rs
fborello-lambda: Yes, to accurately reflect the complete workflow in a testing environment.

core/lib/external_price_api/src/coingecko_api.rs
fborello-lambda: Sounds good, I'm trying this:
```rust
  let config_env = ExternalPriceApiClientConfig::from_env().unwrap();
  let api = CoinGeckoPriceAPIClient::new(config_env);
 ```
 
 And i'm compiling the `dev.toml` config:
 
 ```toml
 [external_price_api_client]
source = "coingecko"
base_url = "https://api.coingecko.com"
api_key = "api"
# Since the coingecko api only works for ethereum mainnet and with the dev config we run a reth instance as L1,
# we can force the erc20 address used to get the conversion_rate, the address below is the DAI contract of ethereum mainnet
forced_ethereum_erc20_address = "0x6b175474e89094c44da98b954eedeac495271d0f"
 ```
 
 But it fails with:
 `Result::unwrap()` on an `Err` value: Cannot load config <external_price_api_client>
 
 Seems that i'm not doing it correctly, i didn't found tests reading from the compiled `env` configs. They are set inside [the test](https://github.com/matter-labs/zksync-era/blob/761bda19844fb3935f8a57c47df39010f88ef9dc/core/lib/env_config/src/external_price_api_client.rs#L34). I would appreciate your insights.

core/lib/external_price_api/src/coingecko_api.rs
fborello-lambda: Fixed!

popzxc:
re using `CoinGecko` instead of forced provider -- I think @fborello-lambda has a point. Also, it would allow us to test the implementation on the testnet/stage, instead of blindly believing that we won't break the mainnet with the new release.

popzxc:
I will close it as stale; @shahar4 feel free to reopen and take ownership.

PR #2446, opened 2024-07-18
chore: add params to 'zk contract register-hyperchain'

dimazhornyk:
## Checklist


Deniallugo:
Do you really need this change, let's use the new tooling and do not upgrade the old one

dimazhornyk:
> Do you really need this change, let's use the new tooling and do not upgrade the old one

Makes sense, I'll close this one and add it to the new tool

PR #2449, opened 2024-07-22
chore: Make eth_tx_manager logs less verbose

popzxc:
## What ❔

Makes eth_tx_manager emit less logs

## Why ❔

Even if there is no activity (e.g. locally), `eth_tx_manager` keeps emitting a lot of logs, which makes it harder to understand what's going on with the server.

## Checklist


itegulov (approved):
Thanks! These logs made me think I have some data locally when I didn't many a time

PR #2453, opened 2024-07-22
refactor: Change prover workspace hierarchy

popzxc:
## What ❔

Prover workspace has got its own "mature" hierarchy, with multiple crates, some data folders, some important files lying in the root, etc.
This makes prover workspace less understandable (especially figuring out the `vk_setup_data_generator_fri/data` folder).

This is the first PR of N that aims to improve prover workspace hierarchy.
It moves all the crates to two subfolders: `crates/bin` and `crates/lib`.
Right now we have most of logic in binaries, but later we'll move some of it to `lib` too. And hopefully, in a foreseeable future we will also have `crates/test`.
Later on we will also have a top-level directory(ies) for _data_ (e.g. keys).

## Why ❔

Make the workspace easier to orient in.

## Checklist


.dockerignore
```diff
@@ -39,7 +39,7 @@ contracts/.git
 !etc/multivm_bootloaders
 !cargo
 !bellman-cuda
-!prover/vk_setup_data_generator_server_fri/data/
+!prover/crates/bin/vk_setup_data_generator_server_fri/data/
```
popzxc, line 42: This was and is suboptimal, but I would prefer to tackle it separately.
This folder should be somewhere near the root and outside of any crate.

PR #2455, opened 2024-07-22
feat: Fallback to consensus secrets file if servers secrets empty

aakoshh:
## What ❔

Related to https://github.com/matter-labs/gitops-kubernetes/pull/5852

Changes the main node entry point to use the `.consensus_secrets.yaml` file (path specified as env var) if the `server-v2-secrets.yaml` file (path specified as CLI arg) doesn't have a `consensus` section. 

## Why ❔

This is to allow us to leave `consensus` empty instead of having to specify secrets in both places, which is what is currently happening in the gitops-kubernetes repo. The external node only reads the `.consensus_secrets.yaml` file, and since this exists for both the main and the external node, it's a bit unintuitive that it's actually ignored by the main node. 

I thought leaving `consensus` empty in the file-based CLI route would be a simple way to make it less confusing without having to change the [Secrets](https://github.com/matter-labs/zksync-era/blob/ce62ddea65e77cd43b9b55f97df6423d2a63e0ca/core/lib/protobuf_config/src/proto/config/secrets.proto#L25) data structure. It allows operators to specify everything in the YAML file they point at with `--server-secrets`, but makes it easier to configure our own nodes. 

I know there is a push to move from envvar based config to file based config. It's not obvious whether the consensus secrets YAML file is considered env var based, since the path is defined as an env var, or file based where the actual secrets are. If there is a better way to do this then please ignore this PR.

## Checklist

### DIFF
--- a/core/bin/zksync_server/src/main.rs
+++ b/core/bin/zksync_server/src/main.rs
@@ -140,18 +140,32 @@ fn main() -> anyhow::Result<()> {
         }
     };
 
-    let secrets: Secrets = match opt.secrets_path {
-        Some(path) => {
-            let yaml =
-                std::fs::read_to_string(&path).with_context(|| path.display().to_string())?;
-            decode_yaml_repr::<zksync_protobuf_config::proto::secrets::Secrets>(&yaml)
-                .context("failed decoding secrets YAML config")?
+    let secrets: Secrets = {
+        let consensus_secrets =
+            config::read_consensus_secrets().context("config::read_consensus_secrets()")?;
+
+        match opt.secrets_path {
+            Some(path) => {
+                let yaml =
+                    std::fs::read_to_string(&path).with_context(|| path.display().to_string())?;
+
+                let mut secrets =
+                    decode_yaml_repr::<zksync_protobuf_config::proto::secrets::Secrets>(&yaml)
+                        .context("failed decoding secrets YAML config")?;
+
+                // Since secrets can be defined in two files, allow the convention of only specifying consensus secrets in one place.
+                if secrets.consensus.is_none() {
+                    secrets.consensus = consensus_secrets;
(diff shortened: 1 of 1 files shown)


core/bin/zksync_server/src/main.rs
```diff
@@ -140,18 +140,32 @@ fn main() -> anyhow::Result<()> {
...
+            Some(path) => {
+                let yaml =
+                    std::fs::read_to_string(&path).with_context(|| path.display().to_string())?;
+
+                let mut secrets =
+                    decode_yaml_repr::<zksync_protobuf_config::proto::secrets::Secrets>(&yaml)
+                        .context("failed decoding secrets YAML config")?;
+
+                // Since secrets can be defined in two files, allow the convention of only specifying consensus secrets in one place.
+                if secrets.consensus.is_none() {
```
pompon0, line 157: IMO it would be better to fail if both ops.secrets_path and the env var with consensus secrets are specified (i.e. disallow coexistence of both config formats altogether). At the very least we should fail if both secrets.consensus.is_some() and env var with consensus secrets are specified.

aakoshh: > it would better to fail if both ops.secrets_path and the env var with consensus secrets are specified

As I understand they want to move to files from env vars, which suggests that the `ops.secrets_path` is not going away, which would mean this would almost surely fail.

> At the very least we should fail if both

Okay, sounds good. Maybe with a check that they are actually different secrets, so it can be backwards compatible with the current gitops.

aakoshh:
It was decided that the `.consensus_secrets.yaml` is considered part of the envvar based vestigial story that should eventually be removed, and that we'll remove this from the configuration files instead as file-based config is rolled out across the environments.

PR #2462, opened 2024-07-22
feat: added consensus_config to general config

pompon0:
Now that we have a support for file-based configuration, we can embed the consensus configuration in general configs file.


core/bin/zksync_server/src/main.rs
```diff
@@ -91,12 +91,25 @@ fn main() -> anyhow::Result<()> {
...
+        }
         Some(path) => {
             let yaml =
                 std::fs::read_to_string(&path).with_context(|| path.display().to_string())?;
-            decode_yaml_repr::<zksync_protobuf_config::proto::general::GeneralConfig>(&yaml)
-                .context("failed decoding general YAML config")?
+            let mut configs =
+                decode_yaml_repr::<zksync_protobuf_config::proto::general::GeneralConfig>(&yaml)
+                    .context("failed decoding general YAML config")?;
+            // Tallback to the consensus_config.yaml file.
```
Deniallugo, line 106:
```suggestion
            // Fallback to the consensus_config.yaml file.
```

core/lib/zksync_core_leftovers/src/temp_config_store/mod.rs
```diff
@@ -111,6 +111,7 @@ impl TempConfigStore {
             snapshot_recovery: self.snapshot_recovery.clone(),
             pruning: self.pruning.clone(),
             external_price_api_client_config: self.external_price_api_client_config.clone(),
+            consensus_config: None,
```
Deniallugo, line 114: could we use `config::read_consensus_config()` here?

pompon0: that's a different crate

Deniallugo (approved):

Deniallugo (approved):

PR #2472, opened 2024-07-24
feat(vlog): New vlog interface + opentelemtry improvements

popzxc:
OK, that's a big PR for its purpose, sorry for that.

Essence of changes:

## Nice observability installation from the config

We can now write the following:
```
    let observability_config = configs
        .observability
        .clone()
        .context("observability config")?;
    let _observability_guard = observability_config.install()?;
```

instead of having to manually map data from `observability_config` to `ObservabilityBuilder` methods.
This is done in a non-intrusive way: the functionality is under a feature flag in `zksync_config` which is only enabled for binaries.

## Nicer `ObservabilityBuilder` interface

Previously we've had "everything in one place", now we have separate structures for logs, opentelemetry, and sentry. It would be nice to integrate prometheus there too, but it's a separate task.

## More service information is exposed via opentelemetry

Previously only service name was exposed. Now we also expose pod name and namespace.
It's done in a non-intrusive way as well, where you can provide _some_ of information if you have it, but otherwise it'll be populated from env variables (which is fine, because it's not a permanent configuration, but rather a machine descriptor).


core/bin/zksync_server/src/main.rs
```diff
@@ -117,31 +117,7 @@ fn main() -> anyhow::Result<()> {
...
-    }
-    let _guard = builder.build();
-
-    // Report whether sentry is running after the logging subsystem was initialized.
-    if let Some(sentry_url) = observability_config.sentry_url {
-        tracing::info!("Sentry configured with URL: {sentry_url}");
-    } else {
-        tracing::info!("No sentry URL was provided");
-    }
+    let _observability_guard = observability_config.install()?;
```
popzxc, line 120: There is a problem here -- server is not outside of tokio context, so we cannot really start the otlp exporter. I will fix that tomorrow.

core/bin/block_reverter/src/main.rs
```diff
@@ -106,21 +106,19 @@ async fn main() -> anyhow::Result<()> {
...
+        .disable_default_logs(); // It's a CLI application, so we only need to show logs that were actually requested.;
+    let sentry: Option<zksync_vlog::Sentry> =
+        TryFrom::try_from(observability_config.clone()).context("sentry")?;
+    let opentelemetry: Option<zksync_vlog::OpenTelemetry> =
+        TryFrom::try_from(observability_config.clone()).context("opentelemetry")?;
+    let _guard = zksync_vlog::ObservabilityBuilder::new()
+        .with_logs(Some(logs))
+        .with_sentry(sentry)
+        .with_opentelemetry(opentelemetry)
+        .build();
```
itegulov, line 121: This is only tangentially related, but what is the purpose of us having a guard? I don't see it being used in a meaningful way anywhere and most of the stuff from `ObservabilityBuilder` is being set up as global anyway. As a user I would expect this to also handle `tracing`'s `DefaultGuard` and maybe some other related things.

Maybe it would make sense to rename `build` to `global` or something and not return a guard at all?

itegulov (commented):
Great refactor, just one nit from my side

core/bin/block_reverter/src/main.rs
popzxc: I see your point, but I feel that the fact that we can't have guards (or other methods of control) for other things here is a bug, not a feature. I don't particularly like the fact that, for example, otlp exporter implicitly installs global handler. It doesn't play well with the node framework either.

A different question is whether we need sentry at all, given that we only use it for panics.

itegulov: I think we could add tracing-related guards here if we really wanted to (open a tracing appender with `WorkerGuard`, then a subscriber with `DefaultGuard`). Not sure about otlp though.

I agree that this is a bug, so my suggestion was more of a "let's not confuse anyone that this behaves a certain way when it really doesn't and we are not investing time in fixing this". In any case not a strong opinion from my side so fine with whatever.

itegulov (dismissed):

PR #2474, opened 2024-07-24
feat(tee): introduce get_tee_proofs RPC method for TEE proofs

pbeza:
## What ❔

Add `get_tee_proofs` RPC method to expose TEE attestations and proofs for a given batch number and TEE type.

Currently, there can be only one proof instance per batch number per TEE type. In the future, we may allow multiple instances since attestations can be revoked if the machine that produced them is compromised.

## Why ❔

We want to enable anyone to download TEE attestations and proofs for any transaction, allowing them to verify it on their own machines using the `verify-attestation` standalone binary (available at https://github.com/matter-labs/teepot/tree/main/bin/verify-attestation).

This is just an intermediate step; the ultimate goal is to have TEE proofs and attestations verification implemented on-chain.

## Checklist


pbeza:
FYI: CI is failing due to a [recent DevOps-related issue](https://matter-labs-workspace.slack.com/archives/C059JA27RE3/p1721898406779919), which is not related to this PR.

core/lib/dal/src/tee_proof_generation_dal.rs
```diff
@@ -16,6 +17,19 @@ pub struct TeeProofGenerationDal<'a, 'c> {
...
+    // signature generated within the TEE enclave, using the privkey corresponding to the pubkey
+    pub signature: Option<Vec<u8>>,
+    // pubkey used for signature verification; each key pair is attested by the TEE attestation
+    // stored in the db
+    pub pubkey: Option<Vec<u8>>,
+    // data that was signed
+    pub proof: Option<Vec<u8>>,
+    // type of TEE used for attestation
+    pub tee_type: Option<TeeType>,
+}
```
pbeza, line 31: This is almost identical to this `struct`:
https://github.com/matter-labs/zksync-era/blob/f42b26027cf864a08b775fa2ed04adcddf5311a9/core/lib/prover_interface/src/outputs.rs#L16-L28
(with a minor change as I needed to introduce `Option`s here)

I don't have a clear idea on how to tidy it up at this point. I had an idea to move `struct TeeProof` definition to
https://github.com/matter-labs/zksync-era/blob/main/core/lib/basic_types/src/tee_types.rs
but then I would need to move this `impl` as well:
https://github.com/matter-labs/zksync-era/blob/f42b26027cf864a08b775fa2ed04adcddf5311a9/core/lib/prover_interface/src/outputs.rs#L60-L69
This would force me to add [`zksync_object_store`](https://github.com/matter-labs/zksync-era/blob/5c932f49fe5a5899df0e7b6a9fe55756d4559800/core/lib/object_store/Cargo.toml#L2C9-L2C28) dependency to the [`zksync_basic_types`](https://github.com/matter-labs/zksync-era/blob/5c932f49fe5a5899df0e7b6a9fe55756d4559800/core/lib/basic_types/Cargo.toml#L2C9-L2C27) crate, which is unexpected and could potentially introduce a circular dependency.

Any suggestions are welcome, @popzxc.

popzxc: You need to define `StorageTeeProof` structure in the `models` module. Take a look at other examples.

core/node/proof_data_handler/src/lib.rs
```diff
@@ -121,6 +126,15 @@ fn create_proof_processing_router(
                         .await
                 },
             ),
+        )
+        .route("/tee/get_proof/:l1_batch_number",
```
popzxc, line 130: I don't think it belongs here. We don't want to expose TEE prover API.
Instead, you need to make an RPC method in the our "main" API (some pointers -- `core/lib/web3_decl` and `core/node/api_server`).

pbeza: @popzxc Is adding my stuff to [`core/node/api_server/src/web3/backend_jsonrpsee`](https://github.com/matter-labs/zksync-era/tree/main/core/node/api_server/src/web3/backend_jsonrpsee) a good idea, or should it belong somewhere else?  Sorry for the basic question – I'm trying to understand which [layer](https://github.com/matter-labs/zksync-era/blob/2fa2249dca15b1968fceec11e485850395f03c9d/core/bin/zksync_server/src/node_builder.rs#L627-L707) I should be looking into ([`Web3ServerLayer`](https://github.com/matter-labs/zksync-era/blob/2fa2249dca15b1968fceec11e485850395f03c9d/core/node/node_framework/src/implementations/layers/web3_api/server.rs#L72-L94)?) and to grasp the overall structure of the RPC server.

core/lib/dal/src/proof_generation_dal.rs
```diff
@@ -1,4 +1,3 @@
-#![doc = include_str!("../doc/ProofGenerationDal.md")]
```
popzxc, line 1: Why did you remove it?

core/lib/dal/src/tee_proof_generation_dal.rs
```diff
@@ -1,122 +1,110 @@
...
-}
-
 impl TeeProofGenerationDal<'_, '_> {
-    pub async fn get_next_block_to_be_proven(
+    pub async fn get_next_batch_to_be_proven(
         &mut self,
+        tee_type: TeeType,
         processing_timeout: Duration,
-    ) -> DalResult<Option<L1BatchNumber>> {
+    ) -> DalResult<Option<(i64, L1BatchNumber)>> {
```
popzxc, line 24: Why do you need to expose IDs? Batch number and TEE type should form a unique identifier of the proof; why do we need something else?

pbeza: Good catch!

In this PR, I initially only wanted to set up an RPC method to get SGX proofs and attestations. However, @haraldh was concerned that my endpoint only returns a single proof for each batch number, so I ended up redesigning the SQL schema for the SGX proofs. This made the PR much bigger in scope than I originally expected.

Harald suggested it might be better if we could submit _multiple_ proofs per `(batch_number, tee_type)` (_not_ just for different TEE types). The main reason is that SGX attestations may not be valid forever; they can be revoked, and so on, so we might need multiple instances. This could apply to other types of TEEs as well.

If you have any suggestions for a better implementation, feel free to share! I can also revert to a single proof per `(batch_number, tee_type)` if you think it's better to introduce complexity to the system more gradually.

Also, thanks for checking things out while it's still a work in progress! It could save a lot of time if I'm heading in the wrong direction with this PR.

core/lib/dal/src/proof_generation_dal.rs
pbeza: SQL schema redesign where we don't need db to be driven by `status` enum anymore. Instead, we can rely on nullable `proved_at`. See related: https://github.com/matter-labs/zksync-era/pull/2474#discussion_r1696601145 I'm open to any suggestions in how to redesign the SQL schema.

core/lib/dal/src/tee_proof_generation_dal.rs
pbeza: Yes, probably — I don't see how it contradicts what I wrote, as I still need to be able to store multiple instances of proofs for a single `(batch_number, tee_type)` pair. We need a way to differentiate between them so that the prover knows which row in the `tee_proof` SQL table to update with the proof.

It's very nuanced when it comes to implementation, so let me briefly summarize how it works to help us be more precise when discussing implementation details.

## Old implementation

The simplified implementation of prover-node communication from the `main` branch:
1. The prover calls the `/tee/proof_inputs` endpoint, which triggers `get_next_batch_to_be_proven` to return the `L1BatchNumber` of the batch to be proven, [along](https://github.com/matter-labs/zksync-era/blob/2fa6bf0ffa5d7a5ff62d595a0efeff9dcd9e5a1a/core/node/proof_data_handler/src/tee_request_processor.rs#L60-L68) with the `TeeVerifierInput` for that batch number.
   https://github.com/matter-labs/zksync-era/blob/c9f76e571e5570d2c4194feee03bb260b24c378f/core/lib/dal/src/tee_proof_generation_dal.rs#L32-L35
   The batch `status` is set to `picked_by_prover` in the database to ensure that the next call to `/tee/proof_inputs` does not return the same batch:
   https://github.com/matter-labs/zksync-era/blob/c9f76e571e5570d2c4194feee03bb260b24c378f/core/lib/dal/src/tee_proof_generation_dal.rs#L41
1. The prover computes the proof and saves it back in the database by calling `/tee/submit_proofs/{l1_batch_number}`, which internally triggers `save_proof_artifacts_metadata`:
   https://github.com/matter-labs/zksync-era/blob/c9f76e571e5570d2c4194feee03bb260b24c378f/core/lib/dal/src/tee_proof_generation_dal.rs#L80-L106
   Using `l1_batch_number` as the unique ID for submitting a proof isn't appropriate here because we want to allow multiple proofs per `l1_batch_number`, even for the same TEE type. That's why I introduced a new implementation in this PR.

## New implementation

The simplified implementation introduced in this PR:
1. The prover calls the `/tee/proof_inputs` endpoint with a `tee_type` payload to retrieve the smallest batch number that has not yet been proved for that `tee_type`, [along](https://github.com/matter-labs/zksync-era/blob/2fa6bf0ffa5d7a5ff62d595a0efeff9dcd9e5a1a/core/node/proof_data_handler/src/tee_request_processor.rs#L60-L68) with the `TeeVerifierInput` for that batch number. Under the hood, this triggers `get_next_batch_to_be_proven`, which returns the row's `ID` (i64) and the `L1BatchNumber` to be proven.
   https://github.com/matter-labs/zksync-era/blob/cea2cf98860a65a7ffcb34094833e344874be8e9/core/lib/dal/src/tee_proof_generation_dal.rs#L20-L24
   Under the hood [we set](https://github.com/matter-labs/zksync-era/blob/c9f76e571e5570d2c4194feee03bb260b24c378f/core/lib/dal/src/tee_proof_generation_dal.rs#L43) `prover_taken_at` to indicate that this `(id, l1_batch_number)` is currently being processed by the prover.
1. The prover computes the proof and saves it in the database by calling `/tee/submit_proofs/{l1_batch_number}` with the `proof_id` in the payload, indicating which row to update with the proof. See the details here:
    https://github.com/matter-labs/zksync-era/blob/cea2cf98860a65a7ffcb34094833e344874be8e9/core/lib/dal/src/tee_proof_generation_dal.rs#L71-L93
   This is the main reason, in response to your original comment, why I return `DalResult<Option<(i64, L1BatchNumber)>>` from `get_next_batch_to_be_proven` instead of `DalResult<Option<L1BatchNumber>>`.

pbeza: For the record: after an offline discussion with @popzxc, we decided to have only a single proof per TEE type per batch. A more extensive redesign will be needed before transitioning to a more flexible system that allows multiple proofs per TEE type per batch. cc @haraldh @thomasknauth

core/lib/dal/src/proof_generation_dal.rs
pbeza: I added it back. 🫡

pbeza:
@popzxc @haraldh, ready to re-review. ✅

The only thing left to do is test the RPC method I added (`unstable_getTeeProofs`). When calling this method, I'm getting `Method not found`.
```
$ curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc": "2.0", "id": 1, "method": "unstable_getTeeProofs", "params": [1, "Sgx"] }' 'http://localhost:3050/'
{"jsonrpc":"2.0","error":{"code":-32601,"message":"Method not found"},"id":1}
```
Same issue with another RPC method from the `unstable` namespace (this method was **not** added by me):
```
$ curl -X POST -H "Content-Type: application/json"   --data '{"jsonrpc": "2.0", "id": 1, "method": "unstable_getTransactionExecutionInfo", "params": [1] }'   'http://localhost:3050/'
{"jsonrpc":"2.0","error":{"code":-32601,"message":"Method not found"},"id":1}
```
On the other hand, I tried another random RPC call (`snapshots_getAllSnapshots`), and it worked:
```
$ curl -X POST -H "Content-Type: application/json" \
  --data '{"jsonrpc": "2.0", "id": 1, "method": "snapshots_getAllSnapshots", "params": [] }' \
  'http://localhost:3050/'
{"jsonrpc":"2.0","result":{"snapshotsL1BatchNumbers":[]},"id":1}
```
Not sure what's the reason.

pbeza:
Re [this](https://github.com/matter-labs/zksync-era/pull/2474#issuecomment-2263558022): after fixing [this bug](https://github.com/matter-labs/zksync-era/pull/2578), I retested this PR and it works. 🎉 PR ready to review. 🫡

```
$ curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc": "2.0", "id": 1, "method": "unstable_getTeeProofs", "params": [1, "Sgx"] }' 'http://localhost:3050'
{"jsonrpc":"2.0","result":[{"l1BatchNumber":1,"teeType":"Sgx","pubkey":[5,6,7,8,9],"signature":[0,1,2,3,4],"proof":[10,11,12,13,14],"provedAt":"2024-08-01T16:18:47.594870Z","attestation":[4,3,2,1,0]}],"id":1}
```

@popzxc, take a look whenever you get a chance. Thanks!

core/lib/dal/doc/ProofGenerationDal.md
```diff
@@ -11,12 +11,9 @@ proof_generation_details
...
 ---
 stateDiagram-v2
-[*] --> ready_to_be_proven : insert_proof_generation_details
-ready_to_be_proven --> picked_by_prover : get_next_block_to_be_proven
-picked_by_prover --> generated : save_proof_artifacts_metadata
-generated --> [*]
-
-[*] --> skipped : mark_proof_generation_job_as_skipped
-skipped --> [*]
+[*] --> ReadyToBeProven : insert_tee_proof_generation_job
```
popzxc, line 14: This is a document for the `proof_generation_dal`, not for the `tee_proof_generation_dal`, so I don't think it should be changed.

core/lib/dal/migrations/20240523085604_add_tee_proof_generation_details_table.down.sql
```diff
@@ -1,4 +1,6 @@
+DROP INDEX IF EXISTS idx_tee_proofs_status_prover_taken_at;
```
popzxc, line 1: You should not edit the migrations, because some of them are already applied on the existing environments.
Instead, you need to create a new migration (see `cargo sqlx` CLI docs for that).

core/lib/dal/src/proof_generation_dal.rs
popzxc, line 1: The line is still removed.

core/lib/dal/src/tee_proof_generation_dal.rs
```diff
@@ -65,58 +63,68 @@ impl TeeProofGenerationDal<'_, '_> {
                         SKIP LOCKED
                 )
             RETURNING
-                tee_proof_generation_details.l1_batch_number
+                tee_proofs.l1_batch_number
             "#,
+            TeeProofGenerationJobStatus::PickedByProver as TeeProofGenerationJobStatus,
```
popzxc, line 68: Are these conversions needed? This looks like no-op.

core/lib/dal/migrations/20240523085604_add_tee_proof_generation_details_table.up.sql
```diff
@@ -1,22 +1,25 @@
+CREATE TYPE tee_proofs_job_status AS ENUM ('ReadyToBeProven', 'PickedByProver', 'Generated', 'Skipped');
```
popzxc, line 1: When thinking about migrations, consider that there are non-empty tables in the existing envs with values in them. Would it work there?

core/lib/dal/src/tee_proof_generation_dal.rs
```diff
@@ -1,60 +1,58 @@
...
+};
 
 #[derive(Debug)]
 pub struct TeeProofGenerationDal<'a, 'c> {
     pub(crate) storage: &'a mut Connection<'c, Core>,
 }
 
-#[derive(Debug, EnumString, Display)]
+#[derive(Debug, sqlx::Type)]
+#[sqlx(type_name = "tee_proofs_job_status")]
```
popzxc, line 20: While I appreciate the strive for strong typing, this looks like a somewhat dangerous change (e.g. things can go wrong with it). With dangerous changes, it's usually better to introduce them separately, so that if something goes wrong, you only have to rollback the _exact_ change and not everything else that was introduced with it.

core/lib/dal/doc/ProofGenerationDal.md
pbeza: Ouch, I mixed up the files, sorry!

core/lib/dal/src/proof_generation_dal.rs
pbeza: [Fixed](https://github.com/matter-labs/zksync-era/pull/2474/commits/aa9d3639050bf2a323c854a9c831750b26146eeb). ✅

core/lib/dal/migrations/20240523085604_add_tee_proof_generation_details_table.down.sql
pbeza: Oh, I naively assumed `sqlx` could automatically handle all the migration steps based on the modifications I made to the SQL table definitions. I'm not sure if I'm doing something wrong or if it's just a real pain to create the right SQL queries for making or undoing changes in the existing SQL tables in the `.up.sql` and `.down.sql` files, but here they are: https://github.com/matter-labs/zksync-era/pull/2474/commits/6e5d39f67318784b8998996dbd7d7e6a9d5fe773. @popzxc please let me know if that looks good enough.

core/lib/dal/migrations/20240523085604_add_tee_proof_generation_details_table.up.sql
pbeza: Same response as here: https://github.com/matter-labs/zksync-era/pull/2474#discussion_r1705158912.

core/lib/dal/src/tee_proof_generation_dal.rs
pbeza: The TEE stuff is at a very early stage, so I wouldn't worry too much about the stability of its already deployed environments. However, you are absolutely right in general. I can split it into two PRs if you insist.

core/lib/dal/src/tee_proof_generation_dal.rs
pbeza: I thought the same until I tried removing `as TeeProofGenerationJobStatus`:
```
   Compiling zksync_dal v0.1.0 (zksync-era/core/lib/dal)
error: unsupported type tee_proofs_job_status for param #1
  --> core/lib/dal/src/tee_proof_generation_dal.rs:35:21
   |
35 |           let query = sqlx::query!(
   |  _____________________^
36 | |             r#"
37 | |             UPDATE tee_proofs
38 | |             SET
...  |
74 | |             &processing_timeout,
75 | |         );
   | |_________^
   |
   = note: this error originates in the macro `$crate::sqlx_macros::expand_query` which comes from the expansion of the macro `sqlx::query` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `zksync_dal` (lib test) due to 1 previous error
error: `cargo check` failed with status: exit status: 101
   Compiling zksync_dal v0.1.0 (zksync-era/core/lib/dal)
error: unsupported type tee_proofs_job_status for param #1
  --> core/lib/dal/src/tee_proof_generation_dal.rs:35:21
   |
35 |           let query = sqlx::query!(
   |  _____________________^
36 | |             r#"
37 | |             UPDATE tee_proofs
38 | |             SET
...  |
74 | |             &processing_timeout,
75 | |         );
   | |_________^
   |
   = note: this error originates in the macro `$crate::sqlx_macros::expand_query` which comes from the expansion of the macro `sqlx::query` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `zksync_dal` (lib test) due to 1 previous error
error: `cargo check` failed with status: exit status: 101
Error: Child process exited with code 1
```
Apparently, it's needed for some reason. 🤷🏼‍♂️

Also, please note that we already have similar code merged into the `main` branch.
https://github.com/matter-labs/zksync-era/blob/b45aa9168dd66d07ca61c8bb4c01f73dda822040/core/lib/dal/src/tee_verifier_input_producer_dal.rs#L112-L114

core/lib/dal/src/tee_proof_generation_dal.rs
popzxc: Even if the component is not in active use, crashloop in any env would cause the on-call team to react and investigate. Depending on the change, a change in component A can cause other components to crash too. So it's always better to develop assuming that your feature can be rolled back, regardless of the readiness state.

pbeza:
As mentioned [here](https://matter-labs-workspace.slack.com/archives/C05ANUCGCKV/p1722949940996389), @RomanBrodetski and @perekopskiy, could you please review? [Unless](https://github.com/matter-labs/zksync-era/pull/2474#discussion_r1705170751) you strongly prefer it, I would rather focus on addressing your code review comments than splitting this PR into two separate PRs and then addressing comments for both.

core/lib/dal/migrations/20240805144000_tee_proofs_reorg.up.sql
```diff
@@ -0,0 +1,24 @@
+DROP INDEX IF EXISTS idx_tee_proof_generation_details_status_prover_taken_at;
+CREATE TYPE tee_proofs_job_status AS ENUM ('ReadyToBeProven', 'PickedByProver', 'Generated', 'Skipped');
```
perekopskiy, line 2: Is there a strong reason to use ENUM here? I'm afraid of 2 things: inconsistency with zk proofs tables that use `TEXT`; and I'm not sure that using ENUM is flexible enough, e.g. if we want to add a new variant, what will it cost us?
IMHO it's not worth it, strong types don't add much value here.

perekopskiy, line 2: Migration is breaking in a sense that we should coordinate applying it and deploying services. If we decide to go with it (which I would like to avoid), then please make sure you disable service, apply migration, deploy updated service.

pbeza: I've radically simplified this file, so please re-review. Thanks!

perekopskiy (approved):

core/lib/dal/migrations/20240805144000_tee_proofs_reorg.up.sql
perekopskiy: I don't see specific reason why not to use ENUM, just raising it. If others don't have as well then probably we should go with it. But I'd strongly prefer to refactor (not necessarily in one go) all the codebase then. cc @EmilLuta

core/lib/dal/migrations/20240805143000_drop_picked_by_column.down.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE tee_verifier_input_producer_jobs ADD COLUMN picked_by TEXT;
```
popzxc, line 1: JFYI: migrations are a tool to change/rollback the database, so generally a single migration per PR is a better approach. Otherwise, if we have to revert PR, we also have to revert 2 migrations, which is unexpected.

core/lib/dal/src/tee_proof_generation_dal.rs
```diff
@@ -67,56 +69,62 @@ impl TeeProofGenerationDal<'_, '_> {
             RETURNING
                 tee_proof_generation_details.l1_batch_number
             "#,
+            &tee_type.to_string(),
+            TeeVerifierInputProducerJobStatus::Successful as TeeVerifierInputProducerJobStatus,
```
popzxc, line 73: Do you need this `as`? It looks like you're casting the type to the same type.

core/lib/dal/src/tee_proof_generation_dal.rs
```diff
@@ -126,53 +134,40 @@ impl TeeProofGenerationDal<'_, '_> {
...
+            .execute(self.storage)
+            .await?;
+        if result.rows_affected() == 0 {
+            let err = instrumentation.constraint_error(anyhow::anyhow!(
+                "Unable to insert TEE proof for batch number {}, TEE type {}",
+                batch_number,
+                tee_type
+            ));
+            return Err(err);
+        }
```
popzxc, line 166: It looks like the query was idempotent before, but right now it disallows inserting the same job twice.
Moreover, without `ON CONFLICT` can it even be that `rows_affected` is 0? I'd assume that on conflict the query will fail, and otherwise we will always insert a single row.
Is there a reason for that?

core/lib/dal/src/tee_proof_generation_dal.rs
```diff
@@ -195,11 +190,76 @@ impl TeeProofGenerationDal<'_, '_> {
             .await?;
         if result.rows_affected() == 0 {
             let err = instrumentation.constraint_error(anyhow::anyhow!(
-                "Unable to insert TEE attestation: given pubkey already has an attestation assigned"
+                "Unable to insert TEE attestation: pubkey {:?} already has an attestation assigned",
+                pubkey
             ));
             return Err(err);
```
popzxc, line 196: I understand that it's not a part of this PR, but same here -- is it a hard error if we will have double attestation? For example, imagine that we receive an attestation request, we save it, but our response is dropped. The TEE machine will retry attestation, and it would cause an error because we've already saved it.

core/lib/dal/src/tee_proof_generation_dal.rs
pbeza: [Fixed](https://github.com/matter-labs/zksync-era/pull/2474/commits/3aa30b230794a2a02721abfce63ac5fc69579273). ✅

core/lib/dal/src/tee_proof_generation_dal.rs
pbeza: [Fixed](https://github.com/matter-labs/zksync-era/pull/2474/commits/3aa30b230794a2a02721abfce63ac5fc69579273). ✅

core/lib/dal/src/tee_proof_generation_dal.rs
pbeza: I already explained it in one of the recent code review rounds here: https://github.com/matter-labs/zksync-era/pull/2474#discussion_r1705192657

This is the error message you receive if you remove `as TeeVerifierInputProducerJobStatus`:
```
error: unsupported type tee_verifier_input_producer_job_status for param #2
  --> core/lib/dal/src/tee_proof_generation_dal.rs:27:21
   |
27 |           let query = sqlx::query!(
   |  _____________________^
28 | |             r#"
29 | |             UPDATE tee_proof_generation_details
30 | |             SET
...  |
63 | |             &processing_timeout,
64 | |         );
   | |_________^
   |
   = note: this error originates in the macro `$crate::sqlx_macros::expand_query` which comes from the expansion of the macro `sqlx::query` (in Nightly builds, run with -Z macro-backtrace for more info)
```

core/lib/dal/migrations/20240805143000_drop_picked_by_column.down.sql
pbeza: [Fixed](https://github.com/matter-labs/zksync-era/pull/2474/commits/caea7a5ccd7a86adcffd4a044b8dd7a1e27ec3d0). ✅

popzxc (approved):

PR #2475, opened 2024-07-24
feat: Update to consensus 0.1.0-rc.4 (BFT-486)

aakoshh:
## What ❔

Updates the `era-consensus` dependency to 0.1.0-rc.4, which brings the following changes:
* `BatchStoreState::last` no longer contains the entire `SyncBatch`, just the number of the latest batch, which should reduce the unintended gossip data volume
* Makes new batch gossip related metrics available to Prometheus (BFT-486)

## Why ❔

Adding attesters to the genesis is a breaking change, so we would like to cut a new release to external node operators. These changes make the gossip less resource intensive as well as make it more observable.

## Checklist


core/node/consensus/src/storage/store.rs
```diff
@@ -446,42 +446,44 @@ impl storage::PersistentBatchStore for Store {
...
-        let Some(last_batch_number) = self.last_batch(ctx).await? else {
-            return Ok(None);
-        };
-        Ok(Some(attester::BatchNumber(
-            last_batch_number.0.saturating_sub(10),
-        )))
-    }
+        // The latest decision is not to store batches with gaps between in the database *of the main node*.
+        // Once we have an API to serve to external nodes the earliest number the main node wants them to sign,
+        // we can get rid of this method: on the main node we can sign from what `last_batc_qc` returns, and
```
brunoffranca, line 469:
```suggestion
        // we can get rid of this method: on the main node we can sign from what `last_batch_qc` returns, and
```

brunoffranca (dismissed):

brunoffranca (approved):

PR #2476, opened 2024-07-24
feat: updated the docker examples to generate consensus keys and use gossipnet

pompon0:
Tested on a hetzner instance.


brunoffranca:
If possible, we could enable consensus by default.

RomanBrodetski (approved):
Approved - but please also wait for @tomg10 's approval

tomg10 (approved):

PR #2477, opened 2024-07-24
perf: writing tx to memory is no longer quadratic

joonazan:
It used to grow proportionally to the number of transactions squared.

### DIFF
--- a/core/lib/multivm/src/versions/vm_fast/bootloader_state/utils.rs
+++ b/core/lib/multivm/src/versions/vm_fast/bootloader_state/utils.rs
@@ -63,13 +63,7 @@ pub(super) fn apply_tx_to_memory(
         (tx_description_offset..tx_description_offset + bootloader_tx.encoded_len())
             .zip(bootloader_tx.encoded.clone()),
     );
-
-    let bootloader_l2_block = if start_new_l2_block {
-        bootloader_l2_block.clone()
-    } else {
-        bootloader_l2_block.interim_version()
-    };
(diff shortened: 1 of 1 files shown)


core/lib/multivm/src/versions/vm_fast/bootloader_state/utils.rs
```diff
@@ -63,13 +63,7 @@ pub(super) fn apply_tx_to_memory(
         (tx_description_offset..tx_description_offset + bootloader_tx.encoded_len())
             .zip(bootloader_tx.encoded.clone()),
     );
-
-    let bootloader_l2_block = if start_new_l2_block {
-        bootloader_l2_block.clone()
-    } else {
-        bootloader_l2_block.interim_version()
```
slowli, line 70: Nit: Can `interim_version()` method be removed after this change? It looks it's not used elsewhere.

core/lib/multivm/src/versions/vm_fast/bootloader_state/utils.rs
```diff
@@ -107,7 +110,12 @@ pub(crate) fn apply_l2_block(
         ),
         (
             block_position + 3,
-            bootloader_l2_block.max_virtual_blocks_to_create.into(),
+            if start_new_l2_block {
+                bootloader_l2_block.max_virtual_blocks_to_create
+            } else {
+                0
+            }
+            .into(),
```
slowli, line 118: Bikeshedding: IMO, moving `.into()` to the `if` clause and using `U256::zero()` would be more readable.

slowli (approved):

PR #2479, opened 2024-07-24
perf: writing tx to bootloader memory is no longer quadratic

joonazan:
It used to grow proportionally to the number of transactions in the L2 block squared.

### DIFF
--- a/core/lib/multivm/src/versions/vm_latest/bootloader_state/l2_block.rs
+++ b/core/lib/multivm/src/versions/vm_latest/bootloader_state/l2_block.rs
@@ -56,12 +56,6 @@ impl BootloaderL2Block {
         self.txs_rolling_hash = concat_and_hash(self.txs_rolling_hash, tx_hash)
     }
 
-    pub(crate) fn interim_version(&self) -> BootloaderL2Block {
-        let mut interim = self.clone();
-        interim.max_virtual_blocks_to_create = 0;
-        interim
-    }
-
     pub(crate) fn make_snapshot(&self) -> L2BlockSnapshot {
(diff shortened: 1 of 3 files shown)


core/lib/multivm/src/versions/vm_latest/bootloader_state/utils.rs
```diff
@@ -110,7 +114,12 @@ pub(crate) fn apply_l2_block(
         ),
         (
             block_position + 3,
-            bootloader_l2_block.max_virtual_blocks_to_create.into(),
+            if start_new_l2_block {
```
slowli, line 117: Bikeshedding: Same nit regarding readability as in the fast VM PR.

core/lib/multivm/src/versions/vm_latest/bootloader_state/l2_block.rs
```diff
@@ -56,12 +56,6 @@ impl BootloaderL2Block {
         self.txs_rolling_hash = concat_and_hash(self.txs_rolling_hash, tx_hash)
     }
 
-    pub(crate) fn interim_version(&self) -> BootloaderL2Block {
-        let mut interim = self.clone();
```
slowli, line 60: Nit: It doesn't look like `Clone` impl for `BootloaderL2Block` is used any more (other than in `Clone` derive in `BootloaderState`, which is also unused). Ditto for the fast VM. IMO, could make sense to remove these derives to not create possibility for performance degradation in the future.

slowli (dismissed):

slowli (approved):

PR #2480, opened 2024-07-24
feat: Added a JSON RPC to simulating L1 for consensus attestation

pompon0:
It just checks what is the last L1 batch certificate stored locally and returns its number (actually, the next number to avoid problems with 0).
ENs will query the main node for this information to determine what is the number of the next batch that they should vote for.


core/lib/types/src/api/en.rs
```diff
@@ -46,3 +46,10 @@ pub struct SyncBlock {
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct ConsensusGenesis(pub serde_json::Value);
+
+/// Status of L1 simulated by the main node.
+/// Used for testing L1 batch signing by consensus attesters.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct SimulatedL1Status {
+    pub next_batch_to_commit: L1BatchNumber,
```
RomanBrodetski, line 54: We do indeed try to simulate L1, but I would still use a more straightforward/descriptive namings for this API. We simply return the first batch that doesn't have quorum yet. I would model the API correspondingly - why load it with this additional context that in the future this will be L1? IMO this feature makes sense even without L1 - we want to collect signatures without any gaps, so we return the current state there.

@pompon0 what do you think?

RomanBrodetski: I also want to note that I think it will take months to get this on L1 (due to audits and protcol upgrade functionality), so, while temporary, it's not like it's only there for some days/weeks. We may even advertise this publicly before this is on L1.

core/node/api_server/src/web3/namespaces/en.rs
```diff
@@ -36,6 +81,23 @@ impl EnNamespace {
...
+        let mut conn = self.state.acquire_connection().await?;
+        let next_batch_to_commit = match conn
+            .consensus_dal()
+            .get_last_batch_certificate_number()
+            .await
+            .context("get_last_batch_certificate_number()")?
+        {
+            Some(n) => to_l1_batch_number(n + 1)?,
+            None => self.first_batch_to_commit,
+        };
```
RomanBrodetski, line 95: why do we have this state at all? Why not apply the logic of `EnNamespace::new` on every API call? I'm probably missing something since now it feels like we have some extra in-memory state (`first_batch_to_commit`) that is not really needed.

core/lib/dal/src/consensus_dal.rs
```diff
@@ -410,40 +406,31 @@ impl ConsensusDal<'_, '_> {
...
             zksync_protobuf::serde::serialize(cert, serde_json::value::Serializer).unwrap(),
         )
         .instrument("insert_batch_certificate")
         .report_latency()
         .execute(self.storage)
         .await?;
-
-        if res.rows_affected().is_zero() {
-            tracing::debug!(l1_batch_number, "duplicate batch certificate");
-        }
```
aakoshh, line 437: I'm not entirely comfortable removing this handling of duplicate certificate insertions. The rest of the code can still produce duplicates due to to the way re-signing works. If we merge this, that can result in an error until something else is changed. 

I might be wrong but I can also think of a convoluted scenario where a duplicate might naturally occur:
1. Say an external node collects a QC for batch N and saves it to its database. 
2. The external node is restarted and asks the main node which is the next batch number we are still voting on. 
3. Say the main node has not collected the QC for N yet (because gossip) and returns N. 
4. All other gossip partners who still have the votes for N propagate it to our external node. The QC is collected again and inserted into the database again, which returns an error now.

IMO it should either ignore it or update it, but not result in an error.

core/lib/dal/src/consensus_dal.rs
```diff
@@ -457,9 +444,10 @@ impl ConsensusDal<'_, '_> {
         .fetch_one(self.storage)
         .await?;
 
-        Ok(row
-            .number
-            .map(|number| attester::BatchNumber(number as u64)))
+        let Some(n) = row.number else { return Ok(None) };
+        Ok(Some(attester::BatchNumber(
+            n.try_into().context("overflow")?,
+        )))
```
aakoshh, line 450: This looks difficult to read. NB the previous pattern is [used all over blocks_dal](https://github.com/matter-labs/zksync-era/blob/46cfdd8f7f2d8003a91355ee4c76ab6836d3c0d7/core/lib/dal/src/blocks_dal.rs#L115), with `u32` at that. I'd say it's impossible that we'll overflow a `u64` with minute or even second level batches, since it can handle milliseconds since unix epoch.

core/node/api_server/src/web3/namespaces/en.rs
pompon0: we need to return consistent responses when the batch certificates table is empty

core/lib/types/src/api/en.rs
aakoshh: I would agree that `SimulatedL1Status` is muddying the waters here - I can't even imagine what else people might think they could stuff in there about L1, chain height maybe? `L1BatchStatus`, perhaps?

core/lib/web3_decl/src/namespaces/en.rs
```diff
@@ -35,6 +35,13 @@ pub trait EnNamespace {
     #[method(name = "genesisConfig")]
     async fn genesis_config(&self) -> RpcResult<GenesisConfig>;
 
+    /// INTERNAL RPC:
+    /// Gets the status of L1 simulated by the main node.
+    /// This is a temporary RPC used for testing L1 batch signing
+    /// by consensus attesters.
+    #[method(name = "simulated_l1_status")]
```
aakoshh, line 42: Every other method uses `camelCasing`.

core/lib/types/src/api/en.rs
pompon0: I'm not returning just a number because then the call won't be extensible. I want to be prepared to return more info here if needed. The simulated L1 status is the least common denominator of what I want to put here. I can rename it to sth like GlobalAttestationStatus and next_batch_to_attest if you want. This API is supposed to provide an abstraction - i.e. not leak how stuff is represented internally. I also want to make it really obvious that this is just a placeholder and that by itself it doesn't provide value.

core/node/api_server/src/web3/namespaces/en.rs
```diff
@@ -14,16 +15,60 @@ use crate::web3::{backend_jsonrpsee::MethodTracer, state::RpcState};
...
+            let info = conn
+                .pruning_dal()
+                .get_pruning_info()
+                .await
+                .context("get_pruning_info()")?;
+            Ok(info
+                .last_soft_pruned_l1_batch
+                .map(|n| n + 1)
+                .unwrap_or(L1BatchNumber(0)))
+        }
```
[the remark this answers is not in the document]
pompon0: Note that the consensus starting point will not be the same as the first stored batch otherwise we would need to backfill information which doesn't make sense. This by itself requires inspecting l1_batches and l1_batches_certificates. pruning_info encapsulates everything else, which IMO is good enough abstraction.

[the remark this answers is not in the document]
aakoshh: It would definitely be nice to move this to the DAL because I don't think I would get it right. I got as far as looking at the [snapshot_recovery_dal](https://github.com/matter-labs/zksync-era/blob/f1cbb74b863b6e0bcfa74ad780beef29844bac6e/core/node/consensus/src/storage/connection.rs#L394) but didn't know about the `pruning_dal`. A bit too much for the `web3` IMO.

core/node/api_server/src/web3/namespaces/en.rs
aakoshh: Can the table get pruned while the API is running, and then fall back to a very old `first_batch_to_commit` that has been surpassed since, regressing the value? 

In other words, can this happen?
1. Start the node, create an `EnNamespace` instance, assign `first_batch_to_commit = 100`
2. Batches are created, QCs are saved, saw we go up to batch number 500
3. DB gets pruned, including the batch QCs
4. Next request sees the 100 again?

aakoshh: What do you mean exactly by inconsistent results?

Looking at the tests, my guess is that because `first_batch_to_commit` relies on `get_sealed_l1_batch_number` before the first batch QC is created, if we called it again without a QC it would move on as soon as there is a new L1 batch.

The problem I see is that if the main node is restarted, it will move on anyway, because it forgets the previous `first_batch_to_commit` value. Currently that seems fine; if we haven't started creating batches yet, we might as well keep returning the latest value until a QC is formed, then go consistently from there. 

I'm not sure how it works with pruning and snapshotting though. Ideally if there _were_ batch QCs before, then it could tell what the latest was before the snapshot, and continue without gaps. However, it can only ask for a QC of something for which it can verify the hash as well, so it might just have to be whatever the next L1 batch record is going to be.

core/node/consensus/src/storage/connection.rs
```diff
@@ -166,6 +158,25 @@ impl<'a> Connection<'a> {
...
+    pub async fn batch_hash(
+        &mut self,
+        ctx: &ctx::Ctx,
+        number: attester::BatchNumber,
+    ) -> ctx::Result<Option<attester::BatchHash>> {
+        let n = L1BatchNumber(number.0.try_into().context("overflow")?);
+        let Some(meta) = ctx
+            .wait(self.0.blocks_dal().get_l1_batch_metadata(n))
+            .await?
+            .context("sqlx")?
```
aakoshh, line 171:
```suggestion
            .context("get_l1_batch_metadata")?
```

core/node/api_server/src/web3/namespaces/en.rs
```diff
@@ -14,16 +15,60 @@ use crate::web3::{backend_jsonrpsee::MethodTracer, state::RpcState};
...
+                return to_l1_batch_number(last + 1);
+            }
+            // Otherwise start with the next sealed L1 batch.
+            if let Some(sealed) = conn
+                .blocks_dal()
+                .get_sealed_l1_batch_number()
+                .await
+                .context("get_sealed_l1_batch_number()")?
+            {
+                return Ok(sealed + 1);
```
aakoshh, line 49: Why is it the last sealed + 1, why not the last sealed itself? How do we sign the fist batch if it already exists when `EnNamespace::new` is called? Say we have batch 0 in the table but no certificate for it, why should we start signing from 1?

core/node/consensus/src/tests.rs
```diff
@@ -661,6 +662,62 @@ async fn test_centralized_fetcher(from_snapshot: bool, version: ProtocolVersionI
...
+        // the first one to sign should be `last_sealed_batch + 1`.
+        let status = api.fetch_simulated_l1_status().await?;
+        assert_eq!(
+            status.next_batch_to_commit,
+            validator.last_sealed_batch() + 1
+        );
+
+        // Insert a cert, then check again.
+        validator.push_random_block(rng).await;
+        validator.seal_batch().await;
```
aakoshh, line 692: It would be good to check here again after the L1 batch is inserted, but before the batch QC is attached to the previous one, that the API returns what you expect, which at the moment IIUC is the original batch, not affected by the new one. 

I am not sure that's correct though, as I suspect if the main node was restarted it would move on to the last batch, so it might as well return the last until it has a certificate, or the one from the snapshot, if it wants to fill gaps.

core/node/api_server/src/web3/namespaces/en.rs
aakoshh: On second thought, I can see why we want to stick to some value while the node is running: I expect that I will also initialise `BatchVotes::min_batch_number` to something, and ideally it should agree with what is returned here. If it doesn't agree, then attesters might be voting on something that the main node ignores, although hopefully just temporarily. A disagreement can stem from a race condition: between the initialisation of `EnNamespace` and `BatchVotes` an extra L1 batch might be created. 

The need to agree would be another reason to delegate this whole query fallback logic to the DAL, so that it's available to both components. Maybe it can be cached there?

aakoshh: A bad race condition would be this:
1. `EnNamespace` is created, there are no batch QC yet, and `first_batch_to_commit` is 100
2. A batch is created on the main node
3. `BatchVotes` is created on the main node and the `min_batch_number` is set to 101
4. External Nodes query this API and get 100, they cast vote on that, but the main node ignores it because it wants at least 101

If the API always returned the latest, then `BatchVotes` would certainly accept votes on 102, 103, etc, even if its minimum is set to 101. 

Otherwise I'd say this number could be in an input for this component, and then the compositor would have to make sure it's correctly initialised to the same value as the other.

core/lib/types/src/api/en.rs
RomanBrodetski: > I want to be prepared to return more info here if needed.

What additional info do you have in mind?

>  I can rename it to sth like GlobalAttestationStatus and next_batch_to_attest if you want

I would definitely prefer it - `LastAttestestedBatch` would be even simpler.

> I also want to make it really obvious that this is just a placeholder and that by itself it doesn't provide value.

Thing is, the feature that we are building currently is being able to collect signatures on the leader node. The next step will be L1, but the feature has value without L1. This feature needs an API - so let's build this API - in this sense it does provide value.

Overall the naming itself is not very important - but I wanted to raise it to underline that for now we don't do anything with L1, so let's not introduce such concepts in code

core/lib/dal/src/consensus_dal.rs
pompon0: ack, if the implementation is not compatible, then I'll leave it as it is. The silent noop behavior needs to be documented though.

core/node/api_server/src/web3/namespaces/en.rs
RomanBrodetski: > we need to return consistent responses when the batch certificates table is empty

core/lib/dal/src/consensus_dal.rs
pompon0: The "as" pattern is a silent wrap, not many people realize that and that's dangerous. I don't see what you find unreadable about try_into() construct - it does exactly what it says: try to cast one integer type to the other and raise an error if it doesn't fit. LMK if you know a more readable way to express that.

core/lib/dal/src/consensus_dal.rs
aakoshh: Thanks. You're right the docs should mention it. The [tests](https://github.com/matter-labs/zksync-era/blob/1c443e5ecfd000279830262a4a35cbc83a9aacec/core/lib/dal/src/consensus_dal.rs#L566-L572) only verify that there is no error raise, and Roman asked for it to be logged, but it was his or Bruno's suggestion to ignore duplicates, so I forgot to mention it as it was known to everyone, my bad.

core/lib/types/src/api/en.rs
pompon0: `chain height` and `L1BatchStatus` are names neither related to attestation nor the fact that this rpc does NOT provide information related to the real chain state.

core/lib/dal/src/consensus_dal.rs
aakoshh: Not the `.try_into` itself, but the whole thing with the `let Some(n)` having a `return` on the same line and then another `Ok(Some(` is a bit busy.

I understand it's dangerous in general, however this data is coming from the database and we know that the rest of the program is casting it to be a `u32`, while we have `u64` here. Maybe wrap it up into some conversion function and move it out of here?

core/lib/web3_decl/src/namespaces/en.rs
pompon0: done

core/node/consensus/src/storage/connection.rs
pompon0: done

core/node/api_server/src/web3/namespaces/en.rs
RomanBrodetski: > we need to return consistent responses when the batch certificates table is empty

Makes sense, but we lose it if the node is restarted anyway right? So we cannot rely on the fact that it's consistent. Do we want it because of

> On second thought, I can see why we want to stick to some value while the node is running: I expect that I will also initialise BatchVotes::min_batch_number to something, and ideally it should agree with what is returned here. If it doesn't agree, then attesters might be voting on something that the main node ignores, although hopefully just temporarily. A disagreement can stem from a race condition: between the initialisation of EnNamespace and BatchVotes an extra L1 batch might be created.

If so, we probably need a single source of truth for the both.

core/node/api_server/src/web3/namespaces/en.rs
pompon0: it doesn't matter whether it is +0,+1,+5,... this is just a starting point and it can be arbitrarily delayed.

aakoshh: Okay, since we're not aiming to fill in the history, it doesn't matter. Might want to leave a comment to explain that it's an arbitrary choice.

core/node/api_server/src/web3/namespaces/en.rs
pompon0: There is no pruning on the main node, my implementation is misleading indeed. It also doesn't matter what this rpc serves on the EN (in fact, it would be better to disable this rpc on the EN, but I guess we don't care).

pompon0: By consistent responses, I mean that on main node startup (with empty l1_batches_consensus), the batch to sign shouldn't be a moving target, i.e. we shouldn't just return the last_sealed_block whatever it is.

core/node/api_server/src/web3/namespaces/en.rs
pompon0: the +1 implementation makes the behavior with empty and nonempty set of sealed batches consistent.

core/lib/dal/src/consensus_dal.rs
pompon0: added more lines

core/lib/types/src/api/en.rs
pompon0: yeah, that's intentional. We may add for example the info that you considered, i.e. how many batches in parallel we want to collect signatures for.

core/lib/dal/src/consensus_dal.rs
aakoshh: It is true that at the moment this is the only place it's used in `consenus_dal`, but this kind of conversion exists in numerous places in `blocks_dal`. I thought in the interest of consistency you could add safe conversions so next time it's the obvious choice.

core/node/consensus/src/tests.rs
aakoshh: Yeah, exactly, so we agree, right? To clarify, what I wanted to say is that adding the test would enshrine the behaviour whether the last one is returned, or the first one, because currently AFAICS it is immediately followed up with a cert, which makes the fact that `first_` exists unobservable in the test.

core/lib/types/src/api/en.rs
pompon0: > LastAttestestedBatch would be even simpler.

This is not extensional name. And note that the EN rpc namespace already is a collection of random stuff. If we end up in a situation that we need more than 1 rpc, I don't want to have a bunch of them mixed among other unrelated rpcs in the same namespace. If you want this rpc to be so very specific (with a name lastAttestestedBatch or similar) then I vote to put it in a separate RPC namespace, which will be just for attestation-related rpcs.

core/node/api_server/src/web3/namespaces/en.rs
aakoshh: I think it would be okay to return whatever the `last_sealed_block` is. Like you pointed out, whether it's +1 or +5 doesn't matter, and after a restart it will move anyway. In fact if it moves, and the `BatchVotes::set_min_batch_number` is updated with it (like the ENs will have to do via polling the API), then we won't get stuck and eventually a QC will be formed. Once a QC is formed, it will stop moving on its own.

pompon0: I've removed querying the pruning_info, ptal

core/lib/types/src/api/en.rs
aakoshh: The examples of extension you mention are still related to batch signing, which is why I proposed `L1BatchStatus`. 

All I'm trying to point out that `SimulatedL1Status` might compel somebody else who has some L1 related thing to add it to this endpoint, even when it's unrelated to what we are doing now, leading to shared ownership of what we say is a temporary API. For example the last finalised block height on L1. If that is intentional, then I rest my case.

pompon0: renamed to AttestationStatus. Is that ok?

core/node/api_server/src/web3/namespaces/en.rs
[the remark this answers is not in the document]
pompon0: I've removed the pruning_dal lookup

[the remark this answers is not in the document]
aakoshh: And snapshots don't have to be considered because the main node doesn't have them?

core/node/consensus/src/tests.rs
pompon0: I've added the assert

core/node/api_server/src/web3/namespaces/en.rs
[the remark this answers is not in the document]
pompon0: yes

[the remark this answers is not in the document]
pompon0: The pruning_info lookup accounted for the snapshots on the main node (not pruning of the main node), but since there are no snapshots on the main node, I've removed the corresponding test case as well.

core/lib/dal/src/consensus_dal.rs
pompon0: I'll update the blocks_dal occurrences in a separate PR

core/node/api_server/src/web3/namespaces/en.rs
pompon0: @aakoshh I don't want a race condition between new batches sealed and bootstraping the attestation.

RomanBrodetski: > By consistent responses, I mean that on main node startup (with empty l1_batches_consensus), the batch to sign shouldn't be a moving target, i.e. we shouldn't just return the last_sealed_block whatever it is.

Why? Won't it be the case anyway if the node restarts?

aakoshh: > I don't want a race condition between new batches sealed and bootstraping the attestation.

Sorry, I didn't want to give the impression that I would like a race condition to exist. I can imagine that the node builder makes this impossible even, with components being initialised first, and during this time I assume it won't create new L1 batches. I just wanted to think through what could go wrong if there was one, without assuming anything about the rest of the system.

My follow up for this PR would be to add a loop that polls what the next batch to vote on should be, and pass it to `BatchVotes::set_min_number`; the source of the data would be the DAL on the main node, and the API on the external nodes. Which is why I thought the DAL could be a good home for this logic, potentially with a cache for consistency, but even easier if it returns the last sealed batch (+1 if you will) until there is a batch QC.

That way I don't think there can be any race conditions, just a bit of eventual consistency between the MN and the ENs: we can initialise the `BatchVotes` with whatever the DAL/API returns to prevent votes from the irrelevant past, then keep increasing the floor and hope that at some point the MN will keep it stable long enough for API queries to refresh votes to arrive within an L1 window; from that point it will be tied to the last batch QC and stay stable.

pompon0: > Why? Won't it be the case anyway if the node restarts?

We don't care about restarts, because the collecting votes happens during single main node execution.

pompon0: > My follow up for this PR would be to add a loop that polls what the next batch to vote on should be, and pass it to BatchVotes::set_min_number

sg, except that (if possible) the main node should rather its own api (for symmetry)

aakoshh: Just to clarify what querying the main node API from the main node itself would mean in terms of deployment:

As I understand the API is running in a different container from the main node, so there will be some new startup dependency between them if I want to initialise it the vote registry. Not sure how it worked with local testing, I guess the component list must have included both consensus and web3. 

I would have thought it would be better to move the logic into the DAL layer and make both web3 and consensus depend on them, keep web3 thin. I see now that even if there was caching of the first in the DAL layer it wouldn’t help keep things consistent because they would be in different containers. 

I still think it could be made stateless to avoid all this head scratching. 

What if the web3 layer is load balanced and everyone sees different values based on when instances were started?

aakoshh: I am also not 100% clear now API based init would work with the node builder. Presumably during wiring the API isn’t running, so the consensus layer would either have to init to 0 or do a lazy init.

aakoshh: Here’s what I think (sorry if I sound like a broken record):
* For now we want the main node and its API to be the single source of truth for what to vote on. 
* After the first batch QC is inserted we don’t want gaps. 
* Before the first batch QC is inserted we want to facilitate attesters agreeing on what to vote on, so a QC is not impossible. 


If we add load balanced scalable web APIs to the picture, potentially with sticky sessions, it has the potential to prevent the first QC from being formed: instead of a single source of truth we have as many as web container instances. If attesters get stuck to querying the same instance, they might each cast their votes on something else forever. 

A stateless DB query might present a value that changes in those initial times, but it does so infrequently at the moment, and gives a minute for the first QC to form each time. 

If we move to a higher batch cadence, we would probably also introduce voting on multiple heights concurrently, in which case we could also adjust the “next” to be less aggressive in how it follows the last, and some QC will form.

pompon0: you are right, storing it in memory won't guarantee consistency because API servers are a distributed service. I'll adjust it as we discussed in the meeting.

aakoshh (dismissed):
Awesome, the new comments explaining the fallbacks are very helpful 👍

RomanBrodetski (dismissed):
I think one of the unit tests is failing though

RomanBrodetski (approved):

PR #2488, opened 2024-07-25
chore: bumped sqlx version

pompon0:


popzxc (commented):
Shouldn't the instructions for `sqlx-cli` be updated too? Also we depend on `sqlx` in `zk_toolbox`.

popzxc (approved):
Thanks! Once merged, please announce it in slack, so that everyone reinstalls the CLI locally.

PR #2505, opened 2024-07-26
docs(prover): Improve default configuration for prover binaries

popzxc:
## What ❔

Changes the default prometheus ports for prover binaries so that there is no need to override them when running together.
Updates the docs accordingly. Also mentions that you can run WVG with multiple threads.

## Why ❔

Convenience.

## Checklist


etc/env/base/fri_witness_vector_generator.toml
```diff
@@ -1,7 +1,7 @@
 [fri_witness_vector_generator]
 prover_instance_wait_timeout_in_secs=200
 prover_instance_poll_time_in_milli_secs=250
-prometheus_listener_port=3314
```
Deniallugo, line 4: could you allign it with file-based configs. just for making our DevOps team life easier

PR #2508, opened 2024-07-26
feat(vm): Allow switching between VMs for latest protocol version

slowli:
## What ❔

- Allows using old (latest) VM by default, new VM only or old + new VM in a shadow mode in the `MainBatchExecutor`.
- Allows configuring this mode for a new `VmRunner`-powered component, VM playground.

## Why ❔

A separate component seems a relatively safe place to start integration from.

## Checklist


core/lib/config/src/configs/experimental.rs
```diff
@@ -60,3 +61,9 @@ impl ExperimentalDBConfig {
         100
     }
 }
+
+#[derive(Debug, Clone, Default, PartialEq, Deserialize)]
+pub struct ExperimentalVmConfig {
+    /// Operation mode for the new fast VM. `None` to disable the fast VM.
+    pub fast_vm_mode: Option<FastVmMode>,
```
slowli, line 68: Not sure whether `Option<_>` makes more sense compared to adding an additional variant to `FastVmMode`. WDYT?

slowli (commented):
A drawback of adding the switch *only* to BWIP is that there's no way to run integration test with a shadowed VM or to run a loadtest with the new VM. Is this OK, or should I add these options? (Could be dealt with later, of course, but then we could miss divergencies triggered by integration tests, or performance degradation on the load test.)

core/lib/config/src/configs/experimental.rs
popzxc: I think a dedicated variant would be more expressive, though either way is fine with me.

core/lib/config/src/configs/vm_runner.rs
```diff
@@ -27,6 +30,10 @@ pub struct BasicWitnessInputProducerConfig {
     pub window_size: u32,
     /// All batches before this one (inclusive) are always considered to be processed.
     pub first_processed_batch: L1BatchNumber,
+
+    #[serde(skip)]
+    // manually added when parsing from env because `envy` doesn't support `serde(flatten)`
+    pub experimental_vm: ExperimentalVmConfig,
```
popzxc, line 36: I guess instead of embedding `ExperimentalVmConfig` into component configs, we may require 2 configs as input for components. Since we're running infra in the distributed mode, we can still override the experimental VM config on a per-component basis.

popzxc (commented):
AFAIU BWIP performance is crucial for the proof generation speed, so probably it's important to not accidentally worsen the performance. Given that VM is single-threaded and isn't _too_ IO-heavy, probably it's fine. But still cc @EmilLuta just in case.

slowli:
@popzxc @EmilLuta This is a good point regarding BWIP availability. Is it possible to run multiple BWIP instances, or is it a singleton by design? Ideally, we'd want to run an additional instance starting from the first L2 block with the supported protocol version, so that it's not a bottleneck and we don't care much about potential panics caused by VM divergencies. If this is impossible / requires significant changes in the BWIP architecture, it could make sense to go with the initial plan of integrating the new VM into the EN state keeper, because in this case it *obviously* wouldn't be a bottleneck, and VM panics wouldn't need to be handled ASAP. (And it would also have the benefit of being able to test the new VM on integration tests; see my comment above.) OTOH, it's relatively easy to convert VM panics into logged errors, but in this case we could get meaningless cascading errors in a single VM run.

core/bin/external_node/src/main.rs
```diff
@@ -94,8 +94,8 @@ async fn build_state_keeper(
         stop_receiver_clone.changed().await?;
         result
     }));
-    let batch_executor_base: Box<dyn BatchExecutor> =
-        Box::new(MainBatchExecutor::new(save_call_traces, true));
```
joonazan, line 98: This actually changes nothing. Committed by accident?

joonazan: Ah, the variable name also changed.

core/lib/config/src/configs/experimental.rs
joonazan: I prefer old, new, shadow over Option(new, shadow)

core/node/node_framework/src/implementations/layers/vm_runner/bwip.rs
```diff
@@ -28,7 +29,7 @@ impl BasicWitnessInputProducerLayer {
         zksync_network_id: L2ChainId,
     ) -> Self {
         Self {
-            basic_witness_input_producer_config,
+            config: basic_witness_input_producer_config,
```
joonazan, line 32: Maybe also rename the argument.

core/lib/config/src/configs/vm_runner.rs
slowli: What would be the benefits? IMO, having VM config embedded into component configs leads to better config structure, especially for file-based configs.

core/bin/external_node/src/main.rs
slowli: Yes, exactly, it's a cosmetic change shortening the var name.

core/lib/config/src/configs/vm_runner.rs
popzxc: Less special handling (here we have to manually add it), less duplication in configuration, easier to remove later (I consider this to be temporary config).

slowli: Yes, I also consider this config temporary, that's why it's marked as experimental. Suppose we integrate the new VM into the API server and EN state keeper as well. I think it's reasonable to expect that we would have the VM switch configurable in 3 places for some time. Do you think it would be better to have a single config with "mode in BWIP" / "mode in state keeper" / "mode in API server sandbox" fields?

popzxc: No, I think it's better to have components accept a tuple: `(component_config, experimental_vm_mode_config)`.
This way, we can gradually add this config to the components, and as long as the configuration is provided for some _env_ (stage/testnet/mainnet), enabling it will be a no-op. If a divergence is found and we need to rollback -- changing the config in one place will change it back everywhere (which IMHO makes sense, because if there is a divergence, likely it will be present in all the components that will process the same batch).

If we want to force a different value for a particular component, we can add override on the component level.

slowli: With the VM playground component, there are no more embedded configs (for now).

core/lib/config/src/configs/experimental.rs
```diff
@@ -60,3 +61,37 @@ impl ExperimentalDBConfig {
...
     }
 }
+
+#[derive(Debug, Deserialize, Clone, PartialEq)]
+pub struct ExperimentalVmPlaygroundConfig {
+    /// Mode in which to run the fast VM implementation. Note that for it to actually be used, L1 batches should have a recent version.
+    #[serde(default)]
+    pub fast_vm_mode: FastVmMode,
+    /// Path to the RocksDB cache directory.
+    #[serde(default = "ExperimentalVmPlaygroundConfig::default_db_path")]
```
popzxc, line 71: Note: `serde(default)` doesn't work with file-based config which we already use on stage.

popzxc (approved):

core/lib/config/src/configs/experimental.rs
slowli: I've [accounted for defaults](https://github.com/matter-labs/zksync-era/blob/83156f08c2c01e282a5eb5d5be6cfcf8f62cc6aa/core/lib/protobuf_config/src/experimental.rs#L75) in `protobuf_config` crate, which AFAIU is sufficient to make file-based config work.

itegulov (approved):
Haven't looked through the whole PR, but the VM runner impl looks good to me

PR #2509, opened 2024-07-26
feat: base token integration tests

cytadela8:
## What ❔

- Introduce 2 new tests for base token:
    - gas calculation
    - proper application of dynamic changes to base token ratio
- Improve forced price client fluctuations implementation to limit difference between consecutive values.

## Why ❔

- We want better integration test coverage of custom base token

## Checklist


core/node/base_token_adjuster/src/base_token_ratio_persister.rs
```diff
@@ -36,15 +36,15 @@ impl BaseTokenRatioPersister {
         let mut timer = tokio::time::interval(self.config.price_polling_interval());
 
         while !*stop_receiver.borrow_and_update() {
-            tokio::select! {
```
cytadela8, line 39: This shouldn't be needed `The first tick completes immediately.`

core/node/external_proof_integration_api/src/lib.rs
```diff
@@ -4,8 +4,6 @@ mod middleware;
 mod processor;
 mod types;
 
-pub use crate::processor::Processor;
-
```
cytadela8, line 8: This was done by `zk fmt` :shrug:

core/tests/ts-integration/tests/fees.test.ts
```diff
@@ -54,12 +56,52 @@ testFees('Test fees', () => {
...
+                undefined,
+                undefined,
+                new Reporter()
+            )
+        );
+        console.log(
+            `After deposit: Alice balance ${ethers.formatEther(
+                await alice.getBalance()
+            )}. Main wallet balance: ${ethers.formatEther(await mainWallet.getBalance())}`
+        );
```
cytadela8, line 104: This may not be needed now.

core/lib/external_price_api/src/forced_price_client.rs
```diff
@@ -1,16 +1,23 @@
...
 use rand::Rng;
+use tokio::sync::RwLock;
 use zksync_config::configs::ExternalPriceApiClientConfig;
 use zksync_types::{base_token_ratio::BaseTokenAPIRatio, Address};
 
 use crate::PriceAPIClient;
 
-// Struct for a a forced price "client" (conversion ratio is always a configured "forced" ratio).
-#[derive(Debug, Clone)]
+const NEXT_VALUE_VARIATION_RANGE: f64 = 0.03; //3%
```
ischasny, line 14: Lets keep it as a configurable parameter please. Different testnets might want their variations to be different.

core/lib/external_price_api/src/forced_price_client.rs
```diff
@@ -36,35 +43,47 @@ impl ForcedPriceClient {
...
-                    let min = (value_f64 * (1.0 - x as f64 / 100.0)).round() as u64;
-                    let max = (value_f64 * (1.0 + x as f64 / 100.0)).round() as u64;
-                    rng.gen_range(min..=max)
-                };
-                let new_numerator = adjust_range(self.ratio.numerator);
-                let new_denominator = adjust_range(self.ratio.denominator);
+        if let Some(fluctation) = self.fluctuation {
+            let mut previous_numerator = self.previous_numerator.write().await;
+            let mut rng = rand::thread_rng();
+            let numerator_range = (
```
ischasny, line 60: Adjusting only numerator might not work here. Imagine USDC as a CBT. If one sets up its ratio to 1:2500 - then it'd never be adjusted. Would be great to either adjust both numerator / denominator as it was before - or at least to check which one is adjustable before choosing which one to adjust.

core/tests/ts-integration/tests/fees.test.ts
```diff
@@ -143,6 +185,116 @@ testFees('Test fees', () => {
...
+        const isETHBasedChain = baseTokenAddress == zksync.utils.ETH_ADDRESS_IN_CONTRACTS;
+
+        if (isETHBasedChain) return;
+
+        await setInternalL1GasPrice(alice._providerL2(), {
+            newL1GasPrice: l1GasPrice.toString(),
+            newPubdataPrice: l1GasPrice.toString(),
+            customBaseToken: true,
+            forcedBaseTokenRatioNumerator: 300,
+            forcedBaseTokenRatioDenumerator: 100,
```
ischasny, line 252: typo "Denominator"

cytadela8: fixed

core/lib/external_price_api/src/forced_price_client.rs
cytadela8: I would rather display a warning or silently multiply both denominator and numerator by something like `100`.  Adjusting the denominator by X% has a weird impact on the ratio, because it changes the ratio by factor of `1/(1+X)`. Enforcing next value variation limit becomes harder.

core/tests/ts-integration/tests/fees.test.ts
```diff
@@ -143,6 +185,116 @@ testFees('Test fees', () => {
...
+            baseTokenAdjusterL1UpdateDeviationPercentage: 0
+        });
+
+        const beginFeeParams = await alice._providerL2().getFeeParams();
+        const mainContract = await alice.getMainContract();
+        const beginL1Nominator = await mainContract.baseTokenGasPriceMultiplierNominator();
+        let changedL2 = false;
+        let changedL1 = false;
+        for (let i = 0; i < 20; i++) {
+            await utils.sleep(0.5);
```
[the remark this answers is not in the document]
cytadela8: Lowered sleep below to 10*0.5

core/tests/ts-integration/tests/fees.test.ts
cytadela8: Turns out it is.

core/lib/external_price_api/src/forced_price_client.rs
cytadela8: done

core/lib/external_price_api/src/forced_price_client.rs
ischasny: Agreed, good point. Can you maybe convert it to float, calculate the next value and then decompose it to numerator / denominator as it's done [here](https://github.com/matter-labs/zksync-era/blob/main/core/lib/external_price_api/src/utils.rs#L7) for example

core/lib/config/src/configs/external_price_api_client.rs
```diff
@@ -4,16 +4,22 @@ use serde::Deserialize;
...
     pub numerator: Option<u64>,
     pub denominator: Option<u64>,
-    /// Forced fluctuation. It defines how much percent numerator /
-    /// denominator should fluctuate from their forced values. If it's None or 0, then ForcedPriceClient
+    /// Forced fluctuation. It defines how much percent the ratio
+    /// should fluctuate from their forced values. If it's None or 0, then ForcedPriceClient
     /// will return the same quote every time it's called. Otherwise, ForcedPriceClient will return
     /// forced_quote +/- forced_fluctuation % from its values.
     pub fluctuation: Option<u32>,
+    /// In order to smoothe out fluctuation consecutive values returned by forced client will not
```
ischasny, line 19: Typo "smooth"

core/tests/ts-integration/tests/fees.test.ts
```diff
@@ -54,12 +56,41 @@ testFees('Test fees', () => {
...
+
+        const mainWallet = new zksync.Wallet(
+            testMaster.environment().mainWalletPK,
+            alice._providerL2(),
+            alice._providerL1()
+        );
+
+        const bridgehub = await mainWallet.getBridgehubContract();
+        const chainId = testMaster.environment().l2ChainId;
+        const baseTokenAddress = await bridgehub.baseToken(chainId);
```
ischasny, line 74: Does it need to be behind `if (!isEthChain)`? As we want deposit to be made only if the chain is on custom base token

core/tests/ts-integration/tests/fees.test.ts
```diff
@@ -143,6 +174,116 @@ testFees('Test fees', () => {
...
+
+        if (isETHBasedChain) return;
+
+        await setInternalL1GasPrice(alice._providerL2(), {
+            newL1GasPrice: l1GasPrice.toString(),
+            newPubdataPrice: l1GasPrice.toString(),
+            customBaseToken: true,
+            forcedBaseTokenRatioNumerator: 300,
+            forcedBaseTokenRatioDenominator: 100,
+            forcedBaseTokenRatioFluctuationsAmplitude: 20,
```
ischasny, line 242: Why not to use the same naming as we have in the forced api client already? I.e. 
 `externalPriceApiClientForcedFluctuation` for example.

core/tests/ts-integration/tests/fees.test.ts
```diff
@@ -143,6 +174,116 @@ testFees('Test fees', () => {
...
+
+            if (
+                ((newFeeParams.V2 as any)['conversion_ratio'].numerator as number) !=
+                ((beginFeeParams.V2 as any)['conversion_ratio'].numerator as number)
+            ) {
+                // @ts-ignore
+                const diff =
+                    (newFeeParams.V2 as any)['conversion_ratio'].numerator -
+                    (beginFeeParams.V2 as any)['conversion_ratio'].numerator;
+                expect(diff).toBeLessThan(75); //75 = 25%*300
```
ischasny, line 265: Can you please explain why `75` here?

ischasny (commented):
Looks really good, please consider a few small comments / suggestions

core/lib/external_price_api/src/forced_price_client.rs
cytadela8: I believe this is going to change denominator in unexpected ways. When you convert a fraction to floating point you loose information and precision.

core/tests/ts-integration/tests/fees.test.ts
cytadela8: updated comment

core/lib/external_price_api/src/forced_price_client.rs
ischasny: But we would have to do such conversion anyway for any "real" quote providers such as coingecko, coinmarketcap and etc (i.e. they won't be supplying numerator / denominator). Also how much precision is critical for us in the forced client (given that we are allowing it to fluctuate on a fairly random number)

ischasny: This is not something thatI'd block the PR on - but please consider converting the quote to float or multiplying onto some number as you suggested before merging.

cytadela8: @ischasny I did add multiplying by 100 if numerator is under 100. I would rebut your argument by that setting ratio to 1:2500 is also something unrealistic as such a setting could not result from converting a "real" quote. Maybe we should go towards changing the forced_price_client config to accept a float for the price itself?

cytadela8: Though I understand your point. I think it would be very unexpected for somebody to see the forced price client to set the price to eg. 4/10000 when they set it 1/2500. They could even see something like 7378697629483821/18446744073709551616 depending on the implementation of `Fraction::from(ratio_f64)` that is not actually documented. (The last numbers where taken from a very real python standard implementation)

core/tests/ts-integration/tests/fees.test.ts
```diff
@@ -143,6 +179,113 @@ testFees('Test fees', () => {
...
+                to: receiver,
+                value: BigInt(1)
+            })
+        ).wait();
+
+        const expectedETHGasPrice =
+            testMaster.environment().l1BatchCommitDataGeneratorMode === DataAvailabityMode.Rollup
+                ? 100_000_000 // 0.1 gwei (the minimum)
+                : 110_000_000; // 0.11 gwei, in validium we need to add compute overhead
+        const expectedCustomGasPrice = expectedETHGasPrice * 0.314;
```
perekopskiy, line 203: You shouldn't hardcode config values here, instead read them with `getFeeParams`

core/tests/ts-integration/tests/fees.test.ts
```diff
@@ -295,9 +440,17 @@ async function killServerAndWaitForShutdown(provider: zksync.Provider) {
 
 async function setInternalL1GasPrice(
```
perekopskiy, line 441: This is no longer only about internal l1 gas price, let's rename to `setFeeParams`

core/tests/ts-integration/tests/fees.test.ts
```diff
@@ -143,6 +179,113 @@ testFees('Test fees', () => {
...
+        if (isETHBasedChain) return;
+
+        await setInternalL1GasPrice(alice._providerL2(), {
+            newL1GasPrice: l1GasPrice.toString(),
+            newPubdataPrice: l1GasPrice.toString(),
+            customBaseToken: true,
+            externalPriceApiClientForcedNumerator: 300,
+            externalPriceApiClientForcedDenominator: 100,
+            externalPriceApiClientForcedFluctuation: 20,
+            baseTokenPricePoolingIntervalMs: 1000,
```
perekopskiy, line 244: typo: p**oll**ing

core/tests/ts-integration/tests/fees.test.ts
```diff
@@ -143,6 +179,113 @@ testFees('Test fees', () => {
...
+            const receipt2 = await (
+                await alice.sendTransaction({
+                    to: receiver,
+                    value: BigInt(1)
+                })
+            ).wait();
+
+            const expectedCustomGasPrice = expectedETHGasPrice * 2.71;
+            expect(receipt2.gasPrice).toBe(BigInt(expectedCustomGasPrice));
+        }
```
perekopskiy, line 229: Probably I miss something, but what's the point of this check, i.e. what does it check that is not covered by other checks?

core/lib/external_price_api/src/forced_price_client.rs
```diff
@@ -29,42 +35,71 @@ impl ForcedPriceClient {
...
+            next_value_fluctuation,
         }
     }
 }
 
 #[async_trait]
 impl PriceAPIClient for ForcedPriceClient {
-    // Returns a ratio which is 10% higher or lower than the configured forced ratio.
+    /// Returns a ratio which is 10% higher or lower than the configured forced ratio,
+    /// but not different more than 3% than the last value
```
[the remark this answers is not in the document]
cytadela8: fixed

core/tests/ts-integration/src/retry-provider.ts
```diff
@@ -57,9 +57,10 @@ export class RetryProvider extends zksync.Provider {
 
     override _wrapTransactionReceipt(receipt: any): zksync.types.TransactionReceipt {
         const wrapped = super._wrapTransactionReceipt(receipt);
-        this.reporter.debug(
-            `Obtained receipt for transaction ${receipt.transactionHash}: blockNumber=${receipt.blockNumber}, status=${receipt.status}`
-        );
+        if (process.env.ZKSYNC_DEBUG_TX_RECEIPT)
+            this.reporter.debug(
+                `Obtained receipt for transaction ${receipt.transactionHash}: blockNumber=${receipt.blockNumber}, status=${receipt.status}`
+            );
```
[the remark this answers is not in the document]
cytadela8: This specific log line is contributing >90% of lines to ZKSYNC_DEBUG_LOGS. I have placed it behind a separate env, so the logs are more useful.

core/tests/ts-integration/tests/fees.test.ts
perekopskiy: `getFeeParams` will return you `FeeModelConfigV2` and `BaseTokenConversionRatio`, so you can replace hardcoded `100_000_000`, `110_000_000`, `0.314` with calculated values

core/tests/ts-integration/tests/fees.test.ts
cytadela8: We have added this check to confirm the gas price changes when the ratio is changed.

perekopskiy: Well, the previous assert checks that gas price depends on the ratio. I wouldn't be against this check unless it required server restart and thus adds 30s of testing time. IMO it isn't worth it. @popzxc WDYT?

popzxc: Agreed.

core/tests/ts-integration/tests/fees.test.ts
cytadela8: done

core/tests/ts-integration/tests/fees.test.ts
cytadela8: done

perekopskiy:
it seems `base ERC20 contract checks › Base token ratio is updated on L1` is flaky (check last run logs)

cytadela8:
Due to removal of `zk` based tests, this PR is blocked by https://github.com/matter-labs/zksync-era/pull/2898

perekopskiy:
@Deniallugo ^

cytadela8:
> it seems `base ERC20 contract checks › Base token ratio is updated on L1` is flaky (check last run logs)

I believe I found where the price was not set properly. Seems this was not "flaky", but permanently broken. pushing a fix...

core/tests/ts-integration/src/utils.ts
```diff
@@ -175,6 +271,26 @@ export class NodeSpawner {
...
+    }
+
+    private writeFileConfig(config: any) {
+        if (this.generalConfigPath == void 0)
+            throw new Error('Trying to set property in config while not in file mode');
+
+        const newGeneralConfig = YAML.stringify(config);
+        fsSync.writeFileSync(this.generalConfigPath, newGeneralConfig, 'utf8');
+    }
 }
```
Deniallugo, line 294: that is obvious, but a little bit tricky part, because js sometimes treat hex values as numbers and store later as number. and it could ruin the parsing logic

core/tests/ts-integration/tests/fees.test.ts
```diff
@@ -213,6 +247,107 @@ testFees('Test fees', function () {
...
+        });
+
+        const doTxToReceiver = () =>
+            alice.sendTransaction({
+                to: receiver,
+                value: BigInt(1)
+            });
+
+        //wait 1 miniblock so the multiplier gets updated
+        await doTxToReceiver().then((tx) => tx.wait(1));
```
perekopskiy, line 265: Multiplier is not guaranteed to be updated after each miniblock, it's updated async. Do I miss something? Does it make sense to poll L1 and wait for multiplier to be updated?

ischasny (dismissed):

core/tests/ts-integration/tests/fees.test.ts
cytadela8: This is a bad comment, it should be `wait 1 miniblock so the l1 gas price gets updated`

cytadela8: Though probably polling is also better.

core/tests/ts-integration/src/utils.ts
cytadela8: I have checked that there are no unexpected changes after parsing and stringifying again:

```
  console.log
    Config differences:
     --- /tmp/original_config.yml       2024-10-23 18:06:48.290023034 +0200
    +++ /tmp/new_config.yml     2024-10-23 18:06:48.290023034 +0200
    @@ -60,7 +60,7 @@
     operations_manager:
       delay_interval: 100
     state_keeper:
    -  transaction_slots: 8192
    +  transaction_slots: 1
       block_commit_deadline_ms: 2500
       miniblock_commit_deadline_ms: 1000
       miniblock_seal_queue_capacity: 10
    @@ -157,6 +157,8 @@
         poll_period: 5
         num_samples_for_blob_base_fee_estimate: 10
         internal_pubdata_pricing_multiplier: 1
    +    internal_enforced_l1_gas_price: 2000000000
    +    internal_enforced_pubdata_price: 2000000000
       watcher:
         confirmations_for_eth_event: 0
         eth_node_poll_interval: 300
    @@ -324,19 +326,19 @@
       max_rows_to_dispatch: 100
       max_retries: 5
     base_token_adjuster:
    -  price_polling_interval_ms: 30000
    -  price_cache_update_interval_ms: 2000
    +  price_polling_interval_ms: 1000
    +  price_cache_update_interval_ms: 500
       max_tx_gas: 80000
       default_priority_fee_per_gas: 1000000
       max_acceptable_priority_fee_in_gwei: 100000000000
    -  l1_receipt_checking_sleep_ms: 30000
    +  l1_receipt_checking_sleep_ms: 1000
       l1_receipt_checking_max_attempts: 3
       l1_tx_sending_max_attempts: 3
    -  l1_tx_sending_sleep_ms: 30000
    +  l1_tx_sending_sleep_ms: 1000
       halt_on_error: false
       price_fetching_max_attempts: 3
       price_fetching_sleep_ms: 5000
    -  l1_update_deviation_percentage: 10
    +  l1_update_deviation_percentage: 0
     basic_witness_input_producer:
       db_path: /home/ert/repos/zksync-era-apu-base-token-integration/./chains/valcus/db/main/basic_witness_input_producer
       window_size: 3
    @@ -344,9 +346,10 @@
     external_price_api_client:
       source: forced
       client_timeout_ms: 10000
    -  forced_numerator: 314
    -  forced_denominator: 1000
    +  forced_numerator: 300
    +  forced_denominator: 100
       forced_next_value_fluctuation: 3
    +  forced_fluctuation: 20
     consensus:
       server_addr: 127.0.0.1:3254
       public_addr: 127.0.0.1:3254

      at NodeSpawner.writeFileConfig (src/utils.ts:312:29)
```

core/tests/ts-integration/tests/fees.test.ts
cytadela8: Ok, so after drafting a wait like this:

```
 let gasPriceGood = false
        for (let i = 0; i < 40; i++) {
            await sleep(0.25);
            const l1GasPrice = (await alice._providerL2().getFeeParams()).V2.l1_gas_price
            if (l1GasPrice === l1GasPrice) {
                gasPriceGood = true;
                break;
            }
            console.log(l1GasPrice)
        }
        if (!gasPriceGood)
            throw new Error("Gas price change doesn't work");
```

I'm having the test fail due to unexpected gasPrice. 

```
    expect(received).toBe(expected) // Object.is equality

    Expected: 33912000n
    Received: 43960000n

      304 |         console.log('gas price', await alice._providerL2().getGasPrice());
      305 |         console.log(await alice._providerL2().getFeeParams());
    > 306 |         expect(receipt.gasPrice).toBe(BigInt(expectedConvertedGasPrice));
          |                                  ^
      307 |     });
      308 |
      309 |     test('Test base token ratio fluctuations', async () => {

      at Object.<anonymous> (tests/fees.test.ts:306:34)
```

This means fee params checked just after the transaction do not match what was used for gasPrice for the transaction. I'm not sure what exactly is causing it / what should we wait for, but I have seen consistently that doing one transaction and waiting for it to be confirmed before doing the tests transaction results in consistent results.

core/lib/config/src/configs/external_price_api_client.rs
```diff
@@ -4,16 +4,21 @@ use serde::Deserialize;
...
-    /// denominator should fluctuate from their forced values. If it's None or 0, then ForcedPriceClient
-    /// will return the same quote every time it's called. Otherwise, ForcedPriceClient will return
-    /// forced_quote +/- forced_fluctuation % from its values.
+    /// Forced fluctuation. It defines how much percent the ratio should fluctuate from its forced
+    /// value. If it's None or 0, then the ForcedPriceClient will return the same quote every time
+    /// it's called. Otherwise, ForcedPriceClient will return quote with numerator +/- fluctuation %.
     pub fluctuation: Option<u32>,
+    /// In order to smooth out fluctuation, consecutive values returned by forced client will not
+    /// differ more than next_value_fluctuation percent. If it's None, a default of 3% will be applied.
+    #[serde(default = "ExternalPriceApiClientConfig::default_forced_next_value_fluctuation")]
```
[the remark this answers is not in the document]
cytadela8: I have done in like this because we still have some compatibility with serde based config parsing (from env vars?). I have separately implemented the defaults in protobuf config. What would you want to me to do if I want this to be optional and have a default? _I think we should overhaul how configs are handled, so there is one code for parsing only... Maybe we should remove serde?_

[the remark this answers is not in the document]
cytadela8: It's done like I did it in other configs eg. prover_job_monitor

core/tests/ts-integration/tests/fees.test.ts
cytadela8: So I want to just update the comment. Unless you have some idea what I can do better here

```suggestion
        //wait 1 miniblock so the gas price gets updated
        await doTxToReceiver().then((tx) => tx.wait(1));
```

perekopskiy: fee params are updated each batch, not miniblock. I'm ok with waiting for batch (please reuse existing function to do that) and changing comment

cytadela8: Got it! Fixed.

core/lib/config/src/configs/external_price_api_client.rs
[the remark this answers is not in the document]
cytadela8: As you are vactioning, I'm closing this.

artmakh (approved):

PR #2511, opened 2024-07-26
chore: disabled get_batch consensus RPC

pompon0:
get_batch is not fully implemented yet (neither in era-consensus nor zksync-era) but it already is consuming resources (db queries and bandwidth). We should finish implementation and loadtest it before enabling it again.


pompon0:
yes, you can see that get_batch RPC takes up to 3s https://grafana.matterlabs.dev/d/eba66cd9-5335-4521-9777-ad34a379f35c/era-consensus?orgId=1&var-datasource=mIPIusa4k . I haven't analyzed the exact impact on the db resources, but I'd like to understand it before we enable it on testnet or mainnet.

PR #2512, opened 2024-07-26
feat: optimize LWG and NWG

0xVolosnikov:
## What ❔

Separating recursive circuits from the AggregationWrapper struct allows them to be used separately and significantly reduces data duplication in the storage.

Processing L/N circuits in parallel speeds up processing, also providing an adjustable limits on peak memory.


0xVolosnikov:
Requires https://github.com/matter-labs/era-zkevm_test_harness/pull/168

0xVolosnikov:
The most important thing here is the fact that   `recursive_circuits` are taken out of `AggregationWrapper` struct. So we will not save them twice for no reason.

Regarding the memory load: it is reduced by more than 2 times, rather by factor of 3-5, since serializing and sending the structure had an additional overhead, proportional to the size of the structure.

prover/crates/bin/witness_generator/src/leaf_aggregation.rs
```diff
@@ -255,29 +245,109 @@ pub async fn prepare_leaf_aggregation_job(
...
         base_vk,
         leaf_params,
     })
 }
 
+// The number of leaf circuits to be processed in parallel.
+// Allows us to easily adjust the balance between processing speed and memory usage.
+// Each circuit requires downloading RECURSION_ARITY (32) proofs, each of which can be roughly estimated at 1 MB.
+// So processing of MAX_IN_FLIGHT_CIRCUITS == 100  should use ~3200 MB of RAM + some overhead during serialization.
+const MAX_IN_FLIGHT_CIRCUITS: usize = 100;
```
popzxc, line 258: Nit: We now have a config variable for that, would it make sense to reuse it here?

prover/crates/bin/witness_generator/src/leaf_aggregation.rs
```diff
@@ -155,7 +152,10 @@ impl JobProcessor for LeafAggregationWitnessGenerator {
         job: LeafAggregationWitnessGeneratorJob,
         started_at: Instant,
     ) -> tokio::task::JoinHandle<anyhow::Result<LeafAggregationArtifacts>> {
-        tokio::task::spawn_blocking(move || Ok(Self::process_job_sync(job, started_at)))
+        let object_store = self.object_store.clone();
+        tokio::task::spawn(async move {
+            Ok(Self::process_job_impl(job, started_at, object_store).await)
+        })
```
popzxc, line 158: This is dangerous. Performing blocking work in the tokio runtime threads can cause the reactor to hand and/or behave unexpectedly otherwise.
If you need to mix blocking and non-blocking work, the preferred way is to obtain a runtime handle via `tokio::runtime::Handle::current()` and pass it into the `spawn_blocking` context, so that you can `handle.block_on(async_stuff)` inside.

prover/crates/bin/witness_generator/src/leaf_aggregation.rs
```diff
@@ -255,29 +245,109 @@ pub async fn prepare_leaf_aggregation_job(
...
+            new_circuit_ids_and_urls
+        });
+
+        handles.push(handle);
+    }
+
+    let mut circuit_ids_and_urls = vec![];
+    for handle in handles {
+        circuit_ids_and_urls.extend(handle.await.unwrap());
+    }
```
popzxc, line 349: `futures::future::join_all` ?

prover/crates/bin/witness_generator/src/node_aggregation.rs
```diff
@@ -186,7 +278,10 @@ impl JobProcessor for NodeAggregationWitnessGenerator {
         job: NodeAggregationWitnessGeneratorJob,
         started_at: Instant,
     ) -> tokio::task::JoinHandle<anyhow::Result<NodeAggregationArtifacts>> {
-        tokio::task::spawn_blocking(move || Ok(Self::process_job_sync(job, started_at)))
+        let object_store = self.object_store.clone();
+        tokio::task::spawn(async move {
+            Ok(Self::process_job_impl(job, started_at, object_store).await)
+        })
```
popzxc, line 284: Same concern

prover/crates/bin/witness_generator/src/utils.rs
```diff
@@ -210,12 +197,12 @@ pub async fn save_node_aggregations_artifacts(
 
 #[tracing::instrument(skip_all)]
 pub async fn load_proofs_for_job_ids(
-    job_ids: &[u32],
+    job_ids: &Vec<u32>,
```
popzxc, line 200: This is not idiomatic, better to leave it as a slice.

prover/crates/bin/witness_generator/src/leaf_aggregation.rs
popzxc: We already have them separate in our infra, this would be defined on the per-deployment level (same as `group_id`).
And in the future, the permits will be memory-based, so a heterogenous load would be handled automatically.

prover/crates/bin/witness_generator/src/leaf_aggregation.rs
popzxc: Indeed, I missed it somehow.

0xVolosnikov: And what about this place?

https://github.com/matter-labs/zksync-era/blob/8b6add2d654235e60ad1b2b5e9cddbb695dcf4d3/prover/crates/bin/witness_generator/src/basic_circuits.rs#L508-L527

prover/crates/bin/witness_generator/src/leaf_aggregation.rs
0xVolosnikov: Made using config. 500 is suitable for LWG/NWG too, but I would prefer 100, since I don't expect a big difference in speed

prover/crates/bin/witness_generator/src/leaf_aggregation.rs
popzxc: This code isn't blocking, yes

Cargo.lock
```diff
@@ -15,9 +15,9 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.21.0"
+version = "0.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
+checksum = "6e4503c46a5c0c7844e948c9a4d6acd9f50cccb4de1c48eb9e291ea17470c678"
```
popzxc, line 20: Did you run `cargo update` or regenerated `Cargo.lock` from scratch?
Looks like much more dependencies were updated than strictly necessary.

While with `crates.io` dependencies it should be generally safe, I would still prefer to not mix `cargo update` with other potentially breaking changes.

Would it be possible to restore the original `Cargo.lock` both here and for prover, and then only apply the required changes (I think `cargo` will do it itself on build)?

prover/crates/bin/witness_generator/src/node_aggregation.rs
```diff
@@ -97,9 +91,11 @@ impl NodeAggregationWitnessGenerator {
         skip_all,
         fields(l1_batch = %job.block_number, circuit_id = %job.circuit_id)
     )]
-    pub fn process_job_sync(
+    pub async fn process_job_impl(
```
popzxc, line 94: It would be nice to leave a comment stating something like "While this function is asynchronous, it contains blocking code, so should only be called from within `spawn_blocking` context" (same in other places).
Can be done in a follow-up, though.

0xVolosnikov: Added

prover/crates/bin/witness_generator/src/leaf_aggregation.rs
```diff
@@ -265,19 +265,93 @@ pub async fn prepare_leaf_aggregation_job(
...
+                    FriProofWrapper::Recursive(_) => {
+                        panic!(
+                            "Expected only base proofs for leaf agg {} {}",
+                            job.circuit_id, job.block_number
+                        );
+                    }
+                })
+                .collect();
+
+            let (_, circuit) = create_leaf_witness(
```
popzxc, line 326: Welp, sorry for raising it again, I should've communicated it better -- but it looks like this function is blocking, and it's in the `tokio::task::spawn` context; e.g. it'll be executed within runtime. I'm not sure if `split_recursion_queue` above is CPU heavy, but if it isn't, then the following changes can be done:

1. It's fine to `tokio::spawn` the `process_leaf_aggregation_job`. Same for `basic_circuits` -- I've checked and in fact we call `zkevm_test_harness::external_calls::run` from `spawn_blocking` context already.
2. Here, `spawn` should be replaced with `spawn_blocking`, and the handle should be passed to block on `save_recursive_layer_prover_input_artifacts`.

0xVolosnikov: Actually it is not CPU heavy at all

0xVolosnikov: I mean `create_leaf_witness` is very light too

0xVolosnikov: `save_recursive_layer_prover_input_artifacts` is just I/O too, should we block on it?

0xVolosnikov: Ok, both `split_recursion_queue` and `create_leaf_witness` are very lightweight. I don't think it will cause any problems.

core/lib/config/src/configs/fri_witness_generator.rs
```diff
@@ -26,12 +26,20 @@ pub struct FriWitnessGeneratorConfig {
 
     pub prometheus_listener_port: Option<u16>,
 
-    /// This value corresponds to the maximum number of circuits kept in memory at any given time for a BWG.
+    /// This value corresponds to the maximum number of circuits kept in memory at any given time for a BWG/LWG/NWG.
```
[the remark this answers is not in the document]
popzxc: We can override the value per-component, so it should be fine.

PR #2513, opened 2024-07-26
perf(vm): Improve snapshot management in batch executor

slowli:
## What ❔

- Does not embed snapshots into each other, and instead use a linear workflow.
- Discards an old snapshot, if any, when processing a new transaction (currently, old snapshots are seemingly never discarded unless the transaction is rolled back).

## Why ❔

- Linear snapshow workflow is easier to grasp and it could allow optimizations in the future.
- Accumulating snapshots introduces an overhead, esp. in the new VM.

## Checklist


core/node/state_keeper/src/batch_executor/traits.rs
```diff
@@ -0,0 +1,238 @@
...
+    tx_result: Box<VmExecutionResultAndLogs>,
+}
+
+/// Object-safe version of `VmInterface` used for batch execution.
+///
+/// # Invariants
+///
+/// - Transaction inspection methods must create a VM snapshot before the execution, and must not roll back these snapshots.
+/// - All rollbacks happen via `rollback_last_transaction`.
+pub(super) trait BatchVm {
```
slowli, line 81: Not sure whether `BatchVm` and `BatchVmFactory` abstractions are warranted given that they are only used only for tests (for now?). IMO, `BatchVm` makes it easier to reason about batch executor – VM interactions and better structures code.

core/node/state_keeper/src/batch_executor/traits.rs
```diff
@@ -0,0 +1,238 @@
...
+
+/// VM factory used by the main batch executor. Encapsulates the storage type used by the executor
+/// so that it's opaque from the implementor's perspective.
+pub(super) trait BatchVmFactory<S>: fmt::Debug + Send + Sync {
+    fn create_vm<'a>(
+        &self,
+        l1_batch_params: L1BatchEnv,
+        system_env: SystemEnv,
+        storage: StoragePtr<StorageView<S>>,
+    ) -> Box<dyn BatchVm + 'a>
```
slowli, line 216: If performance of a dynamically dispatched `BatchVm` is a concern, this could be replaced with a non-object safe trait using GATs:

```rust
trait BatchVmFactory {
    type Vm<S: ReadStorage>: BatchVm;

    fn create_vm<S: ReadStorage>(&self, ..) -> Self::Vm<S>;
}
```

core/node/state_keeper/src/batch_executor/traits.rs
```diff
@@ -0,0 +1,238 @@
...
+    where
+        S: 'a;
+}
+
+/// Default VM factory creating `VmInstance`s.
+impl<S: ReadStorage> BatchVmFactory<S> for () {
```
slowli, line 222: With https://github.com/matter-labs/zksync-era/pull/2508, the default factory would encapsulate the fast VM mode.

core/node/state_keeper/src/batch_executor/tests/batch_vm_factory.rs
```diff
@@ -0,0 +1,94 @@
...
+use zksync_types::Transaction;
+
+use crate::batch_executor::traits::{BatchVm, BatchVmFactory, TraceCalls, VmTransactionOutput};
+
+pub(super) struct VmTrackingSnapshots<S: ReadStorage> {
+    inner: VmInstance<S, HistoryEnabled>,
+}
+
+impl<S: ReadStorage> VmTrackingSnapshots<S> {
+    fn assert_not_too_many_snapshots(&self) {
```
popzxc, line 16: Bikeshedding: `assert_at_most_one_shapshot`?

core/node/state_keeper/src/batch_executor/traits.rs
popzxc: I don't think it's a concern, at least for now.

core/node/state_keeper/src/batch_executor/traits.rs
popzxc: Maybe it'd be better to have a dedicated type instead of using unit type? I think usage of unit type here, while common in libraries, may be hard to understand for the team.

core/node/state_keeper/src/batch_executor/traits.rs
popzxc: The original intention was to make `BatchExecutor` trait to be responsible for that kind of stuff (back then the hypothesis was that the new VM will not be lifetime-bound, so we won't have to spawn a dedicated thread for each batch). I'm not against having `BatchVm`/`BatchVmFactory` per se, but then I think we should remove `BatchExecutor` trait (looks like we only have a single alternative implementation for tests, so hopefully it shouldn't be too hard).

core/node/state_keeper/src/batch_executor/traits.rs
joonazan: You could add a new method to BatchVm instead of the factory if you don't want the BatchVm to be object safe.

core/lib/multivm/src/versions/shadow.rs
```diff
@@ -23,6 +23,17 @@ pub struct ShadowVm<S, T> {
     shadow: vm_fast::Vm<ImmutableStorageView<S>>,
 }
 
+impl<S, T> ShadowVm<S, T>
+where
+    S: ReadStorage,
+    T: VmFactory<StorageView<S>>,
+{
+    #[doc(hidden)] // used in batch executor tests
+    pub fn snapshot_count(&self) -> usize {
```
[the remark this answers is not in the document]
slowli: Yeah, thought about this as well. IIRC, it messes up some `multivm` tests; will look and try to fix them.

core/lib/state/src/rocksdb/mod.rs
```diff
@@ -257,6 +257,12 @@ impl RocksdbStorageBuilder {
     ) -> anyhow::Result<()> {
         self.0.revert(storage, last_l1_batch_to_keep).await
     }
+
+    /// Returns the underlying storage without synchronizing it or checking that it's up-to-date.
+    /// Should not be used other than for tests.
+    pub fn build_unchecked(self) -> RocksdbStorage {
```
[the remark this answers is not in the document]
slowli: Won't work; `#[cfg(test)]` is scoped by the crate, i.e., it would be visible from the `state` crate only, while we want to use it from another crate (`state_keeper`).

core/lib/multivm/src/interface/traits/vm.rs
```diff
@@ -147,13 +147,25 @@ pub trait VmFactory<S>: VmInterface {
 }
 
 /// Methods of VM requiring history manipulations.
+///
+/// # Snapshot workflow
+///
+/// External callers must follow the following snapshot workflow:
+///
+/// - Each new snapshot created using `make_snapshot()` must be either popped or rolled back before creating the following snapshot.
+///   OTOH, it's not required to call either of these methods by the end of VM execution.
```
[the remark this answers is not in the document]
slowli: What's the contradiction? We don't create a snapshot at the end of the VM execution, so it may have either 0 or 1 snapshot at the end.

core/node/state_keeper/src/batch_executor/traits.rs
slowli: It's not that I don't want it to be object-safe (I do want it because it would allow to define custom impls in the node framework). But a boxed VM instance could theoretically lead to performance degradation due to dynamic dispatch, less opportunities for inlining / contextual optimizations etc.

slowli:
I've decided to remove batch VM traits from the PR. They can be added separately, after the new VM is merged into `main`.

slowli:
There are 2 somewhat suspicious failing CI jobs on this PR, but I don't see how they could be related to this PR:

- [First](https://github.com/matter-labs/zksync-era/actions/runs/10145435296/job/28051157963?pr=2513): Missing revert data for `estimateGas` during the upgrade test
- [Second](https://github.com/matter-labs/zksync-era/actions/runs/10167461168/job/28119983480?pr=2513): Missing revert data for `estimateGas` during server integration tests (specifically "ERC20 contract checks › Can perform a deposit with precalculated max value") + reuse of a nonce in another server integration test ("ERC20 contract checks › Should claim failed deposit").

Since the new VM isn't integrated into the API server, I don't see how the `estimateGas` error could be related to this PR or the new VM integration in general, but I may be missing something.

core/lib/multivm/src/interface/traits/vm.rs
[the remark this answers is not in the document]
joonazan: > it's not required to call either of these methods by the end of VM execution.

This confused me (and still does) but I can understand now that it means that you can have multiple TX after your snapshot and then rollback/forget.

core/lib/multivm/src/versions/vm_refunds_enhancement/vm.rs
```diff
@@ -162,10 +160,7 @@ impl<S: WriteStorage> VmInterfaceHistoryEnabled for Vm<S, HistoryEnabled> {
         self.rollback_to_snapshot(snapshot);
     }
 
-    /// Pop the latest snapshot from the memory and destroy it
     fn pop_snapshot_no_rollback(&mut self) {
-        self.snapshots
-            .pop()
-            .expect("Snapshot should be created before rolling it back");
+        self.snapshots.pop();
```
joonazan, line 164: When is it necessary to pop a nonexistent snapshot?

joonazan (commented):
I'm in favor of merging if you are sure that the disabled tests can be fixed.

core/lib/multivm/src/versions/vm_refunds_enhancement/vm.rs
slowli: It happens [here](https://github.com/matter-labs/zksync-era/blob/b51e26d1a4da11bb244e4c31da4609d13fad0839/core/node/state_keeper/src/batch_executor/main_executor.rs#L172) on the first transaction, or after the previous transaction was rolled back.

joonazan: I think you could just always pop the snapshot if it isn't rolled back in the same place where you may roll it back.

PR #2518, opened 2024-07-29
[WIP] feat(consensus): committees data extractor (BFT-443) 

moshababo:
Offshoot of https://github.com/matter-labs/zksync-era/pull/2416

## What ❔

Adding state extractor for committees data, reading from the consensus registry contract to be inserted into `l1_batches_consensus` DB table. 

## Why ❔

Reading directly from VM storage (with `zksync_node_consensus::storage::registry_contract::VMReader`) is inefficient, and is using a (slightly) different data format. A background-running denormalization process would ease ongoing per-batch reads. 

## Changes

* Added `zksync_node_consensus::storage::registry_contract::CommitteeExtractor`
* Added DB migrations for adding 1 new column to `l1_batches_consensus` table
* Added `batch_committee()` & `insert_batch_committee()`, `get_last_batch_committee_number` in `consensus_dal`
* Updated `sqlx::query` of `insert_batch_certificate()`, `get_last_batch_certificate_number` in `consensus_dal`
* Added DB (de)serialization support
* Refactored `zksync_node_consensus::storage::testonly::VMWriter` to support writes across multiple batches
* Added `test_committee_extractor`

TODOs
- [ ] `zksync_dal/migrations`: Update the CHECK constraint on `certificate` column to support nullable column (the constraint name is automatically generated)
- [ ] Add documentation 

## Checklist


core/lib/dal/migrations/20240725194916_l1_batches_consensus_committees.up.sql
```diff
@@ -0,0 +1,12 @@
+ALTER TABLE l1_batches_consensus
+ADD validator_committee JSONB NULL,
```
pompon0, line 2: How about we make it one big json, so that it is extensible to adding more information to the contract? Btw, is the contract upgradeable?

core/lib/dal/src/consensus_dal.rs
```diff
@@ -317,7 +323,39 @@ impl ConsensusDal<'_, '_> {
         else {
             return Ok(None);
         };
-        Ok(Some(zksync_protobuf::serde::deserialize(row.certificate)?))
+        Ok(Some(zksync_protobuf::serde::deserialize(
+            row.certificate.unwrap(),
```
pompon0, line 327: with non null removed, you shouldn't just unwrap

core/lib/dal/src/consensus_dal.rs
```diff
@@ -317,7 +323,39 @@ impl ConsensusDal<'_, '_> {
...
+        )?))
+    }
+
+    pub async fn batch_committees(
+        &mut self,
+        batch_number: attester::BatchNumber,
+    ) -> anyhow::Result<Option<(validator::ValidatorCommittee, attester::AttesterCommittee)>> {
```
pompon0, line 334: this deserves a struct which contains both committees

core/lib/dal/src/consensus_dal.rs
```diff
@@ -402,6 +440,48 @@ impl ConsensusDal<'_, '_> {
         Ok(())
     }
 
+    pub async fn insert_batch_committees(
+        &mut self,
+        number: BatchNumber,
+        validator_committee: ValidatorCommittee,
+        attester_committee: AttesterCommittee,
+    ) -> anyhow::Result<()> {
```
pompon0, line 448: if we are going to query the last batch with committees/certificate, then keeping them in the same table makes little sense, because we will need extra indices for efficient lookups. Do we really want to go this way?

core/node/consensus/src/storage/denormalizer.rs
```diff
@@ -0,0 +1,127 @@
...
+            }
+        })
+        .await;
+        match res {
+            Ok(()) | Err(ctx::Error::Canceled(_)) => Ok(()),
+            Err(ctx::Error::Internal(err)) => Err(err),
+        }
+    }
+
+    async fn denormalize_batch_committees(
```
pompon0, line 35: by "denormalizer" you mean database denormalization? I think it is a way too big a mental shortcut. How about sth more like "CommitteeStateExtractor", or sth else which would actually indicate that this is the thing extracting the committee from the L2 state at the end of the batch.

core/node/consensus/src/storage/mod.rs
```diff
@@ -69,3 +77,101 @@ impl PayloadQueue {
         Ok(())
     }
 }
+
+#[derive(Debug, Default)]
+#[allow(dead_code)]
+pub struct CommitteeValidator {
```
pompon0, line 83: Please move the types which represent the contract API to a separate module.

core/lib/dal/migrations/20240725194916_l1_batches_consensus_committees.up.sql
moshababo: The committees have distinct functionality, so putting them on separate columns makes sense. However they are mostly part of "data derived from the registry contract at a given batch", so it’s possible to consolidate.

core/node/consensus/src/storage/denormalizer.rs
moshababo: I used [database] “denormalizer” to describe an ongoing read/rewrite procedure. I didn’t add (yet) a noun to describe the domain, just to keep extendable. 

As for the verb, I think “extractor” is even more abstract. 
It can make sense if it’s about finding/setting the committees to be used, however I see it as a mere database query optimization / copying of data from one place to another. 

If you don’t want to go for database lingo, I can change that.

core/node/consensus/src/storage/connection.rs
```diff
@@ -339,6 +361,18 @@ impl<'a> Connection<'a> {
             .context("batch_certificate()")?)
     }
 
+    /// Wrapper for `consensus_dal().batch_committees()`.
+    pub async fn batch_committees(
```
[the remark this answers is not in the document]
moshababo: I followed the simple notion of them being "data derived from the registry contract at a given batch".

core/lib/dal/src/consensus_dal.rs
moshababo: Querying the last batch with committees is a one-time read on server start, so I didn’t thought it’s a big deal. I don’t see such an index for the query of the last batch with certificate (unless it’s applied internally via the toolchain).

In the overall, I’m not against separating the tables, and originally suggested that. IIRC @RomanBrodetski thought it’s an overkill, and that having all info in respect to a given batch in one place is useful.

core/lib/dal/migrations/20240725194916_l1_batches_consensus_committees.up.sql
moshababo: The contract is not upgradable.

core/lib/dal/src/consensus_dal.rs
pompon0: there is no index, because the the table serves as index for itself when it has just 1 non null column. With 2 columns, it won't be the case.

core/node/consensus/src/storage/denormalizer.rs
pompon0: I mean, it is the schema of the db that can be denormalized, not the data itself (which is just duplicated in storage then).

core/lib/dal/migrations/20240725194916_l1_batches_consensus_committees.up.sql
pompon0: > I see it as the opposite of https://github.com/matter-labs/zksync-era/pull/2518#discussion_r1695305287. If it's not merely about copying data, but rather about extracting/setting the committees, then the columns shouldn't be unified

Agreed, I'm advocating that it should be about merely copying the data then.

RomanBrodetski: > that it should be about merely copying the data then.

but why? If we just wanted to copy - there would probably less reason to do this feature at all. We want to interpret the data and assign the committee for batch - so that the information, or method how exactly we do this assignment is not needed in other places. Just get it from the DB!

core/lib/dal/src/consensus_dal.rs
RomanBrodetski: >  because we will need extra indices for efficient lookups.

why exactly? `select ... order by ID desc where certificate is not null` will only scan the rows for which there is committee but not certificate. Given that it happens on startup, it shouldn't be a problem - am I missing smt?

core/lib/dal/migrations/20240725194916_l1_batches_consensus_committees.up.sql
moshababo: Latest update is to have 1 column, for the attester committee.

core/lib/dal/src/consensus_dal.rs
moshababo: > there is no index, because the the table serves as index for itself when it has just 1 non null column. With 2 columns, it won't be the case.

Right, I overlooked that in that case there's no need for `WHERE` clause.

> why exactly? `select ... order by ID desc where certificate is not null` will only scan the rows for which there is committee but not certificate. Given that it happens on startup, it shouldn't be a problem - am I missing smt?

If column X is NOT NULL, there's no need for `WHERE X IS NOT NULL`, and `SELECT MAX(PK)` alone is sufficient. 

Since the NULL column is the certificate, and not the committee, this won't happen just on startup. However, I'm not sure about the frequency of this query and whether it justifies an index.

core/node/consensus/src/storage/connection.rs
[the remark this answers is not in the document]
moshababo: Changed to (singular) `batch_committee`, since we're only dealing with the attester committee.

core/node/consensus/src/storage/denormalizer.rs
moshababo: Changed to `CommitteeExtractor`.

core/node/consensus/src/storage/mod.rs
moshababo: This ended up moving everything to `consensus/storage/registry_contract`.

core/lib/dal/src/consensus_dal.rs
moshababo: Fixed here: https://github.com/matter-labs/zksync-era/pull/2518/commits/65d3fa3d928d1667e3e4b0bd466d7d4c392dfaa3

core/lib/dal/src/consensus_dal.rs
aakoshh: FWIW I would keep the committee in a separate table. As I understand we are doing this without concern for performance, but there are many ways to store the committee, and some don't involve repeating the entire committee in every batch, only when it changes. To keep our options open for future changes, I would keep them separate from the beginning, rather than start messing around with existing constraints to make room for this in the same table, having to revisit the (luckily few) existing queries where we now have to add filtering.

core/lib/dal/src/consensus_dal.rs
```diff
@@ -317,7 +320,36 @@ impl ConsensusDal<'_, '_> {
...
             return Ok(None);
         };
-        Ok(Some(zksync_protobuf::serde::deserialize(row.certificate)?))
+        let Some(certificate) = row.certificate else {
+            return Ok(None);
+        };
+        Ok(Some(zksync_protobuf::serde::deserialize(certificate)?))
+    }
+
+    pub async fn batch_committee(
```
aakoshh, line 329: It would be great to state it in docstrings that this returns the committee which should attest to the `batch_number`, and not the one which is the result of executing the batch, ie. the ones attesting for `batch_number+1`.

core/lib/dal/src/consensus_dal.rs
```diff
@@ -402,6 +434,32 @@ impl ConsensusDal<'_, '_> {
...
+            "#,
+            i64::try_from(number.0).context("overflow")?.into(),
+            zksync_protobuf::serde::serialize(&committee, serde_json::value::Serializer).unwrap(),
+        )
+        .instrument("insert_batch_committee")
+        .report_latency()
+        .execute(self.storage)
+        .await?;
+        if res.rows_affected().is_zero() {
+            tracing::debug!(l1_batch_number = ?number, "duplicate batch certificate");
```
aakoshh, line 458:
```suggestion
            tracing::debug!(l1_batch_number = ?number, "duplicate batch committee");
```

core/lib/dal/src/consensus_dal.rs
```diff
@@ -491,6 +552,45 @@ impl ConsensusDal<'_, '_> {
         // and doesn't have prunning enabled.
         Ok(attester::BatchNumber(0))
     }
+
+    pub async fn get_last_batch_committee_number(
```
aakoshh, line 556: Can you add some docstrings to explain what this method is supposed to do? It sounds like "the committee number of the last batch" but there is no such thing. Is it the batch number of the last _executed_ batch, for which a committee is known? Isn't that available somewhere else?

core/node/consensus/src/storage/registry_contract/committee_extractor.rs
```diff
@@ -0,0 +1,136 @@
...
+        Self { pool, reader }
+    }
+
+    pub async fn run(mut self, ctx: &Ctx) -> anyhow::Result<()> {
+        tracing::info!("Running zksync_node_consensus::storage::CommitteeExtractor");
+        let res = scope::run!(ctx, |ctx, s| async {
+            let mut next_batch = self.next_batch_to_extract_committee(ctx).await?;
+            loop {
+                self.wait_batch(ctx, next_batch).await?;
+                self.extract_batch_committee(ctx, next_batch).await?;
```
aakoshh, line 29: So `wait_batch` calls [get_sealed_l1_batch_number](https://github.com/matter-labs/zksync-era/blob/3f2cac6f2a8b81738098eb26d270be8494d2faac/core/lib/dal/src/blocks_dal.rs#L101) which is the last batch in the table, but it's not necessarily execute yet, or is it? What would happen if the contract was called with a number that it doesn't have data for, would it fail, or return an empty result?

core/node/consensus/src/storage/registry_contract/committee_extractor.rs
```diff
@@ -0,0 +1,136 @@
...
+impl CommitteeExtractor {
+    pub fn new(pool: ConnectionPool, reader: VMReader) -> CommitteeExtractor {
+        Self { pool, reader }
+    }
+
+    pub async fn run(mut self, ctx: &Ctx) -> anyhow::Result<()> {
+        tracing::info!("Running zksync_node_consensus::storage::CommitteeExtractor");
+        let res = scope::run!(ctx, |ctx, s| async {
+            let mut next_batch = self.next_batch_to_extract_committee(ctx).await?;
+            loop {
```
aakoshh, line 27: Would it be possible at all to hook into the batch execution itself, so instead of polling and eventually loading the data from here to there, the derived data would be inserted within the same transaction?

core/lib/dal/src/consensus_dal.rs
```diff
@@ -412,11 +470,12 @@ impl ConsensusDal<'_, '_> {
...
-                l1_batches_consensus (l1_batch_number, certificate, created_at, updated_at)
-            VALUES
-                ($1, $2, NOW(), NOW())
-            ON CONFLICT (l1_batch_number) DO NOTHING
+            UPDATE l1_batches_consensus
+            SET
+                certificate = $2,
+                updated_at = NOW()
+            WHERE
+                l1_batch_number = $1
```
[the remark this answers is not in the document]
aakoshh: On second thought this shouldn't happen: before the certificate for batch N can be verified the node needs the committee from the execution of batch N-1, so I assume there is plenty of time between when the committee is extracted from N-1 and inserted for N, especially because we need the commitment for batch N too. 

Still it feels like having two tables would avoid any confusion.

core/node/consensus/src/storage/registry_contract/committee_extractor.rs
```diff
@@ -0,0 +1,136 @@
...
+            members: contract_attester_committee
+                .into_iter()
+                .map(CommitteeAttester::try_into)
+                .collect::<Result<Vec<consensus_dal::Attester>, _>>()?,
+        };
+
+        self.pool
+            .connection(ctx)
+            .await?
+            .insert_batch_committee(ctx, batch_number, db_attester_committee)
```
aakoshh, line 85: Just checking my understanding of which batch number we mean here.

I thought the `committee` in `l1_batches_consensus` for batch number N means which is the committee that is expected to sign batch N. This is the result of executing batch N-1. 

So we call `extract_batch_committee` with the `next_batch` that we don't haven't extracted yet, and then figure out what is the last L2 block included in this batch. We execute the `eth_call` there to get the committee after the execution of the last L2 block. Then we insert the committee _with the same batch number_? Shouldn't it be `batch_number.inc()`?

pompon0:
closing this pr, I'll continue this work in https://github.com/matter-labs/zksync-era/pull/2684

PR #2520, opened 2024-07-29
fix(prover): Parallelize circuit metadata uploading for BWG

EmilLuta:
Basic Witness Generation (BWG) creates circuits for base layer to be proven. BWG runs out-of-circuit VM and provides circuits for the instructions in the execution (note that those are ordered, messing with the order would break proving).

Additionally, harness will do heavy computation -> provide circuits -> heavy computation -> provide circuits. As such, providing circuits is rather fast. Once a circuit is produced, it's sent to BWG, which in turn sends it to GCS and saves it to DB. The queue between BWG runner and harness is of size 1. It was picked in this way to ensure that harness won't overwhelm bwg runner with circuits (if the queue was big, the amount of RAM used would be size of queue * size of circuit). The problem is that the harness will be throttled on circuit submission. It needs to blockingly wait for bwg runner to upload data (GCS + PG).

There are 2 alternatives here:
- make the buffer bigger -- this would work, but it would start consuming more memory. In theory, if you can match the size of the buffer to fit in between CPU cycles, harness wouldn't need to wait for upload and bwg runner would catch up by the time harness has a new set of circuits. The problem with this approach is that it relies on deterministic upload time (nothing can be deterministic in distributed systems) and fixed circuit size (this is not true, because batches can have varying number of circuits, of varying types).
- upload everything async -- this work wonders, but there will be multiple circuits in flight being uploaded at any time. This will also increase RAM. In my testing, I've seen increases smaller than 100MB (acceptable tradeoff). The edge cases are when GCS is down or not working as intended (for instance, refusing connection). In such scenarios the async tasks will grow, adding to a lot more RAM usage (up to 50GB). Whilst this is a real problem, I consider it more of an edge case. Furthermore, there are a few ways to make GCS more reliable.

The second option was picked here. MPSC channel (the buffer) is still of size 1 to have a single point of failure when RAM goes up (we don't want to figure out -- do we have too many things in the buffer, or do we have problems with GCS? -- in the current implementation, it will always be GCS). Furthermore, MPSC(1) doesn't cause much overhead.

The implementation has been tested locally, using mainnet and testnet batches. Whilst the number are super encouraging (00:03:58 time, with ~40GB RAM), the CPUs (and their speed) is slightly lower in GCP.

### DIFF
--- a/etc/env/file_based/general.yaml
+++ b/etc/env/file_based/general.yaml
@@ -182,6 +182,7 @@ witness_generator:
   max_attempts: 10
   shall_save_to_public_bucket: true
   prometheus_listener_port: 3116
+  max_circuits_in_flight: 500
 witness_vector_generator:
   prover_instance_wait_timeout_in_secs: 200
   prover_instance_poll_time_in_milli_secs: 250


prover/crates/bin/witness_generator/src/basic_circuits.rs
```diff
@@ -432,40 +438,83 @@ async fn generate_witness(
...
+    // Sequence is used to determine circuit ordering (the sequencing of instructions) .
+    // If the order is tampered with, proving will fail (as the proof would be computed for a different sequence of instruction).
+    let circuit_sequence = AtomicUsize::new(0);
+
+    // Future which receives circuits and saves them async.
+    let circuit_receiver_handle = async {
+        while let Some(circuit) = circuit_receiver.recv().await {
+            let sequence = circuit_sequence.fetch_add(1, Ordering::SeqCst);
+            let object_store = object_store.clone();
+            save_circuit_handles.push(tokio::task::spawn(save_circuit(
```
popzxc, line 466: Please note that `tokio::task::spawn` can be considered mostly a no-op. So this implementation _in extremum_ has even more memory consumption than two unbounded channels:

Unbound channels: any generated circuit/queue is added to the channel, but we send them to GCS _fully_ sequentially (e.g. we can't even send a circuit and a queue in parallel). So, we may have, say, 10 items in the channel, but only 1 will be in the process of sending.
Spawned tasks: any generated circuit/queue is added to the channel and immediately consumed. A new task is spawned for it. Tasks includes additional steps like serialization, which also consume memory. So, our channel will generally be empty, but we may have 10 spawned jobs at different stages (serialization/sending).

Moreover, since we will be sending data in parallel, it _may_ result in lower bandwidth for each particular task (e.g. 1gbps for 10 tasks => ~100mbps/task), so that each job is processed for longer, and the amount of jobs we have to keep in memory also grows.

It's just an observation, since the old apporach was focused on saving us from OOM, and PR description also touches that -- I think that the new approach _may_ be significantly more prone to OOM in real environments.

0xVolosnikov (approved):
lgtm

prover/crates/bin/witness_generator/src/basic_circuits.rs
```diff
@@ -446,43 +461,103 @@ async fn generate_witness(
...
+    // `circuits_present` stores which circuits exist and is used to filter queues in `recursion_urls` later.
+    let mut circuits_present = HashSet::<u8>::new();
+
+    for task in save_circuit_handles {
+        let (circuit_id, circuit_url) = task.await.expect("failed to save circuit");
+        circuit_urls.push((circuit_id, circuit_url));
+        circuits_present.insert(circuit_id);
+    }
+
+    for task in save_queue_handles {
```
popzxc, line 554: Here you can also do:
```rust
let recursion_urls: Vec<_> = futures::future::join_all(save_queue_handles).filter(|(circuit_id, _, _)| {
  circuits_present.contains(&circuit_id)
}).collect();
```

prover/crates/bin/witness_generator/src/basic_circuits.rs
```diff
@@ -446,43 +461,103 @@ async fn generate_witness(
...
+            .await
+        {
+            let sequence = circuit_sequence.fetch_add(1, Ordering::SeqCst);
+            let object_store = object_store.clone();
+            let semaphore = semaphore.clone();
+            save_circuit_handles.push(tokio::task::spawn(async move {
+                let _permit = semaphore
+                    .acquire()
+                    .await
+                    .expect("failed to get permit for running save circuit task");
```
popzxc, line 504: I don't think this works. You're limiting the number of spawned jobs, but not the number of created futures.
Imagine that we have only 1 permit.
We receive a new circuit, we spawn a task. When it's launched, we get a permit. 1 job in progress, 1 object in memory.
We receive the next circuit, we spawn a task. When it's launched, we can't get a permit. 1 job in progress, 2 objects in memory.
We receiver the third circuit, we spawn a task. When it's launched, we can't get a permit. 1 job in progress, 3 objects in memory. Etc.

For it to work, you need to obtain an [owned permit](https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html#method.acquire_owned) before you spawn the task, move it into the task, and drop it after `save_circuits`. This way lack of permits would stop us from letting witness generation to proceed.

prover/crates/bin/witness_generator/src/basic_circuits.rs
```diff
@@ -446,43 +461,103 @@ async fn generate_witness(
...
-                Some((circuit_id, queue, inputs)) = queue_receiver.recv().instrument(tracing::info_span!("wait_for_queue")) => {
-                    let urls = save_recursion_queue(block_number, circuit_id, queue, &inputs, object_store).await;
-                    recursion_urls.push(urls);
-                }
-                else => break,
-            };
+
+    // Future which receives circuits and saves them async.
+    let circuit_receiver_handle = async {
+        let semaphore = Arc::new(Semaphore::new(MAX_IN_FLIGHT_CIRCUITS));
```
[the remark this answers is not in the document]
popzxc: Yup, that's exactly what I meant -- "multiple witgens" == "all rounds, can pick & process any job".

prover/crates/bin/witness_generator/src/basic_circuits.rs
```diff
@@ -446,43 +461,103 @@ async fn generate_witness(
...
+    let save_queues_span = tracing::info_span!("save_queues");
+
+    // Future which receives recursion queues and saves them async.
+    let queue_receiver_handle = async {
+        while let Some((circuit_id, queue, inputs)) = queue_receiver
+            .recv()
+            .instrument(tracing::info_span!("wait_for_queue"))
+            .await
+        {
+            let object_store = object_store.clone();
```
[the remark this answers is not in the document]
popzxc: It was a genuine question, because from the code it looks like it was missed by accident. Probably worth a comment stating that we don't need semaphore here, because these jobs are small and there is not a lot of them.

core/lib/config/src/configs/fri_witness_generator.rs
```diff
@@ -87,4 +96,13 @@ impl FriWitnessGeneratorConfig {
     pub fn last_l1_batch_to_process(&self) -> u32 {
         self.last_l1_batch_to_process.unwrap_or(u32::MAX)
     }
+
+    // 500 was picked as a mid-ground between allowing enough circuits in flight to speed up circuit generation,
```
popzxc, line 100: Micro-nit: should be a doc-comment (`///`).

popzxc (dismissed):

popzxc (dismissed):

popzxc (approved):

popzxc (approved):

PR #2522, opened 2024-07-29
refactor: replaced wrapping casts with checked casts in dal

pompon0:
more safety


core/lib/dal/src/blocks_dal.rs
```diff
@@ -441,7 +441,7 @@ impl BlocksDal<'_, '_> {
     ) -> DalResult<()> {
         match aggregation_type {
             AggregatedActionType::Commit => {
-                let instrumentation = Instrumented::new("set_eth_tx_id#commit")
+                let i = Instrumented::new("set_eth_tx_id#commit")
```
slowli, line 444: Bikeshedding: I don't really like one-letter var naming here; `i` is usually used for index vars. How about `instr` if `instrumentation` is too long?

core/lib/db_connection/src/instrument.rs
```diff
@@ -527,3 +527,30 @@ mod tests {
             .unwrap();
     }
 }
+
+pub trait DalContext<T> {
```
slowli, line 531: Please move this code to before `mod tests`; it looks misleading otherwise. Also, I'd add this trait / impl to the `error` module, by analogy with `SqlxContext`.

core/lib/db_connection/src/instrument.rs
```diff
@@ -527,3 +527,30 @@ mod tests {
...
+    /// Wraps a provided argument validation error. It is assumed that the returned error
+    /// will be returned as an error cause (e.g., it is logged as an error and observed using metrics).
+    fn arg<Q>(self, i: &Instrumented<Q>, arg_name: &'static str) -> DalResult<T>;
+}
+
+impl<T, E> DalContext<T> for Result<T, E>
+where
+    E: Into<anyhow::Error>,
+{
+    fn arg<Q>(self, i: &Instrumented<Q>, arg_name: &'static str) -> DalResult<T> {
```
slowli, line 541: - Isn't using `Instrumented<'_, ()>` sufficient here? AFAIU, the query is always bound later.
- Bikeshedding: `arg` is an overly generic name; looking at it in isolation, I would think it would be an argument getter or something similar. If we go by analogy with `SqlxContext`, it should be called something like `encode_arg`.

core/lib/dal/src/consensus_dal.rs
```diff
@@ -325,17 +325,11 @@ impl ConsensusDal<'_, '_> {
...
-                end: L2BlockNumber(numbers.end.0.try_into().context("end")?),
-            })
-        })()
-        .map_err(|err| {
-            Instrumented::new("block_payloads")
-                .with_arg("numbers", &numbers)
-                .arg_error("numbers", err)
-        })?;
+        let i = Instrumented::new("block_payloads").with_arg("numbers", &numbers);
+        let numbers = std::ops::Range {
```
slowli, line 329: Off-topic: Explicit `Range` construction looks weird; why not use the ordinary `start..end` syntax?

core/lib/dal/src/blocks_dal.rs
```diff
@@ -570,11 +566,11 @@ impl BlocksDal<'_, '_> {
...
         let initial_bootloader_contents = serde_json::to_value(initial_bootloader_contents)
-            .map_err(|err| instrumentation.arg_error("initial_bootloader_contents", err))?;
+            .arg(&i, "initial_bootloader_contents")?;
         let used_contract_hashes = serde_json::to_value(&header.used_contract_hashes)
-            .map_err(|err| instrumentation.arg_error("header.used_contract_hashes", err))?;
-        let storage_refunds: Vec<_> = storage_refunds.iter().copied().map(i64::from).collect();
-        let pubdata_costs: Vec<_> = pubdata_costs.iter().copied().map(i64::from).collect();
+            .arg(&i, "header.used_contract_hashes")?;
+        let storage_refunds: Vec<i64> = storage_refunds.iter().map(|x| (*x).into()).collect();
+        let pubdata_costs: Vec<i64> = pubdata_costs.iter().map(|x| (*x).into()).collect();
```
slowli, line 573: Dumb question: Is there a particular reason behind these changes? It looks like the old code would still work fine.

slowli (commented):
Looks fine conceptually, but I can't help but bikeshed a couple of things 🙃

core/lib/db_connection/src/instrument.rs
pompon0: done

core/lib/dal/src/blocks_dal.rs
pompon0: There was nothing wrong. I just find it more readable this way, so I've reflexively rewritten it.

core/lib/dal/src/consensus_dal.rs
pompon0: changed

PR #2529, opened 2024-07-29
feat: added dropping all attester certificates when doing hard fork

pompon0:
Clean state at every fork is closer to what will happen during hard fork on L1.

### DIFF
--- a/core/lib/dal/.sqlx/query-3f9543c622a3877dd8612f3fc5f31867b3c0bff34bbacbc31ff4cd786c97ca90.json
+++ b/core/lib/dal/.sqlx/query-3f9543c622a3877dd8612f3fc5f31867b3c0bff34bbacbc31ff4cd786c97ca90.json
@@ -1,0 +1,12 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "\n            DELETE FROM l1_batches_consensus\n            ",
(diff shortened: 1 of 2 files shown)


aakoshh (approved):
When I first saw that `miniblocks_consensus` is deleted, I thought that's okay because L1 doesn't care about L2 blocks, but I thought the QCs of the existing L1 batches can be kept around. 

I see that a specific number of `miniblocks` are kept by the `last_l2_block_to_keep` number of `miniblocks` are retained, so that the API can keep serving transaction data to users. The corresponding L1 batches have already been submitted, and they are the ones that aren't rolled back, so the QC over the constituent L2 blocks are not relevant any more and the validator set could have changed as well during regenesis. 

On the same account we can remove the L1 batch QCs for those past L1 batches since they are not going to be re-submitted, and because attesters might change in regenesis at the moment. In the future when attesters will be based on a smart contract, the historical quorums will stay amenable to validation by looking up the corresponding attesters, at which point we won't have the same argument for getting rid of these records as we do now.

pompon0:
@aakoshh my guess would be that even once the signature verification happens on L1, the signature itself won't be persisted in the contract state, so the qc will be available only in the calldata of the corresponding transactions.

PR #2532, opened 2024-07-29
feat: Revisit base config values

perekopskiy:
## What ❔

- changes some values in base config
- removes `request_timeout`, `account_pks` as those are dead
- adds "override" config, it's not used by code but is helpful when setting new chain

## Why ❔

- unify config values
- remove dead code

## Checklist


etc/env/file_based/overrides/raas_mainnet/general.yaml
Deniallugo, line 1: it's only indirectly related to rass. Let's keep just, testnet/mainnet

Deniallugo (approved):

Deniallugo (approved):

Deniallugo (approved):

PR #2535, opened 2024-07-29
fix(zk_toolbox): Use fast mode

Deniallugo:
## Checklist


Deniallugo:
Close, because it will make only  margin improvement

PR #2536, opened 2024-07-30
feat(vlog): Implement otlp guard with force flush on drop

popzxc:
## What ❔

Improves the `ObservabilityGuard` so that it flushes both sentry & otlp events on drop.

## Why ❔

Without it, some events (at least for otlp) may be missed if the application exits right after events were produced.

## Checklist

### DIFF
--- a/core/lib/vlog/src/lib.rs
+++ b/core/lib/vlog/src/lib.rs
@@ -1,6 +1,8 @@
 //! This crate contains the observability subsystem.
 //! It is responsible for providing a centralized interface for consistent observability configuration.
 
+use std::time::Duration;
+
 use ::sentry::ClientInitGuard;
 use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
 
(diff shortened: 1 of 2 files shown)


core/lib/vlog/src/lib.rs
```diff
@@ -23,7 +25,36 @@ pub struct ObservabilityBuilder {
...
+    pub fn force_flush(&self) {
+        // We don't want to wait for too long.
+        const FLUSH_TIMEOUT: Duration = Duration::from_secs(1);
+
+        if let Some(sentry_guard) = &self.sentry_guard {
+            sentry_guard.flush(Some(FLUSH_TIMEOUT));
+        }
+
+        if let Some(provider) = &self.otlp_provider {
+            for result in provider.force_flush() {
```
itegulov, line 45: Hmm, looks like otlp examples [suggest](https://github.com/open-telemetry/opentelemetry-rust/blob/main@%7B2024-05-20T12:41:02Z%7D/opentelemetry-otlp/examples/basic-otlp/src/main.rs#L137-L139) using `shutdown`. Is there a reason to call `force_flush` instead?

itegulov: Ah, looks like they are doing different things... Maybe let's call both then?

itegulov (approved):

PR #2538, opened 2024-07-30
feat: External prover API

Artemka374:
## What ❔

Adds prover API component, which allows to get proof generation data and verify final proofs against generated by our prover subsystem.

## Why ❔

First step of prover decentralization.

## Checklist


Cargo.toml
```diff
@@ -36,6 +36,7 @@ members = [
     "core/node/api_server",
     "core/node/tee_verifier_input_producer",
     "core/node/base_token_adjuster",
+    "core/node/prover_api",
```
popzxc, line 39: `public_prover_api`? Same for type names and component name.

core/lib/dal/src/proof_generation_dal.rs
```diff
@@ -88,6 +88,48 @@ impl ProofGenerationDal<'_, '_> {
...
 
+    pub async fn get_available_batch(&mut self) -> DalResult<Option<L1BatchNumber>> {
+        let result: Option<L1BatchNumber> = sqlx::query!(
+            r#"
+            UPDATE proof_generation_details
+            SET
+                status = 'picked_by_prover',
+                updated_at = NOW(),
+                prover_taken_at = NOW()
+            WHERE
```
popzxc, line 99: This query locks the batch so that it cannot be picked by anyone else. I think we need a query that doesn't mutate state.

core/node/prover_api/src/processor.rs
```diff
@@ -0,0 +1,196 @@
...
+            last_available_batch_number,
+            last_available_batch_data: None,
+        }
+    }
+
+    #[tracing::instrument(skip_all)]
+    pub(crate) async fn get_proof_generation_data(
+        &mut self,
+        request: Json<ProofGenerationDataRequest>,
+    ) -> Result<Json<ProofGenerationDataResponse>, ProcessorError> {
```
popzxc, line 51: My understanding was that we can give any batch that already has data required for proving (with some caveats, e.g. we probably should only expose batches for the last protocol version, for the sake of simplicity).

I would imagine the request to accept `Option<L1BatchNumber>`, where we either return data for a particular batch, or latest available.

core/node/prover_api/src/processor.rs
```diff
@@ -0,0 +1,196 @@
...
+                Ok(Json(ProofGenerationDataResponse::Success(Some(Box::new(
+                    data,
+                )))))
+            }
+            Err(err) => Err(err),
+        }
+    }
+
+    #[tracing::instrument(skip(self))]
+    async fn proof_generation_data_for_existing_batch(
```
popzxc, line 89: This is copy-pasted from the proof data handler, right? Maybe makes sense to extract this functionality into a function there, add it as a dependency, and reuse it here, just so that we avoid code duplication.

core/node/prover_api/src/processor.rs
```diff
@@ -0,0 +1,196 @@
...
+        Path(l1_batch_number): Path<u32>,
+        Json(payload): Json<VerifyProofRequest>,
+    ) -> Result<(), ProcessorError> {
+        tracing::info!(
+            "Received request to verify proof for batch: {}",
+            l1_batch_number
+        );
+
+        let serialized_proof = bincode::serialize(&payload.0)?;
+        let expected_proof = bincode::serialize(
```
popzxc, line 180: When we fetch proof generation input, right now we pick the job that _doesn't_ have a proof, while here we expect the proof to exist. First of all, we should probably alter the logic for fetching the job (see also other comments on this topic), but secondly -- would it make sense to check if the batch is already proven in the DB before going to GCS?

etc/env/file_based/general.yaml
```diff
@@ -354,3 +354,7 @@ da_dispatcher:
   polling_interval_ms: 5000
   max_rows_to_dispatch: 100
   max_retries: 5
+
+prover_api:
+  http_port: 3073
+  last_available_batch: 1
```
popzxc, line 360: Nit: missing endline.

core/node/prover_api/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+    blob_store: Arc<dyn ObjectStore>,
+    connection_pool: ConnectionPool<Core>,
+    commitment_mode: L1BatchCommitmentMode,
+    config: ProverApiConfig,
+) -> Router {
+    let mut processor = Processor::new(
+        blob_store.clone(),
+        connection_pool.clone(),
+        commitment_mode,
+        L1BatchNumber(config.last_available_batch),
```
popzxc, line 54: This shouldn't be a part of the config. Instead, we should do a database query here and choose the value dynamically.

core/node/prover_api/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+    let verify_proof_processor = processor.clone();
+    Router::new()
+        .route(
+            "/proof_generation_data",
+            post(
+                // we use post method because the returned data is not idempotent,
+                // i.e we return different result on each call.
+                move |payload: Json<ProofGenerationDataRequest>| async move {
+                    processor.get_proof_generation_data(payload).await
+                },
```
popzxc, line 65: I think that with the proposed changes `get` method would be suitable.

core/node/prover_api/src/error.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+impl IntoResponse for ProcessorError {
+    fn into_response(self) -> Response {
+        let (status_code, message) = match self {
+            ProcessorError::ObjectStore(err) => {
+                tracing::error!("GCS error: {:?}", err);
+                (
+                    StatusCode::BAD_GATEWAY,
+                    "Failed fetching/saving from GCS".to_owned(),
```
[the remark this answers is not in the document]
Artemka374: On the other hand - it is more clear that we don't have the proof yet. But definitely we can have another error for this case

core/node/prover_api/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+            ),
+        )
+        .route(
+            "/verify_proof/:l1_batch_number",
+            post(
+                move |l1_batch_number: Path<u32>, payload: Json<VerifyProofRequest>| async move {
```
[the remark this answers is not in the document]
Artemka374: cause it is coming from path of request. Like `host/verify/{l1_batch_number}`

core/lib/dal/src/proof_generation_dal.rs
```diff
@@ -88,6 +88,48 @@ impl ProofGenerationDal<'_, '_> {
         Ok(result)
     }
 
+    pub async fn get_available_batch(&mut self) -> DalResult<Option<L1BatchNumber>> {
+        let result: Option<L1BatchNumber> = sqlx::query!(
+            r#"
+            UPDATE proof_generation_details
```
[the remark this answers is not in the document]
Artemka374: that actually was my fault while copy pasting. Forgot to update query, will fix

core/node/prover_api/src/processor.rs
Artemka374: Huh, I thought we kinda agreed to return the latest one... At least this is what was described in the design doc. But not hard to implement

core/node/prover_api/src/lib.rs
```diff
@@ -0,0 +1,78 @@
...
+    let mut processor = Processor::new(
+        blob_store.clone(),
+        connection_pool.clone(),
+        commitment_mode,
+        L1BatchNumber(config.last_available_batch),
+    );
+    let verify_proof_processor = processor.clone();
+    Router::new()
+        .route(
+            "/proof_generation_data",
```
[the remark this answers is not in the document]
Artemka374: honestly, `get_proof` sounds strange and doesn't really represent what the endpoint is for, same as `get_proof_data` e.g. Someone can think that he is getting the already generated proof

core/bin/zksync_server/src/node_builder.rs
```diff
@@ -562,6 +563,16 @@ impl MainNodeBuilder {
         Ok(self)
     }
 
+    fn add_external_proof_integration_api_layer(mut self) -> anyhow::Result<Self> {
+        let config = try_load_config!(self.configs.external_proof_integration_api_config);
+        self.node.add_layer(ExternalProofIntegrationApiLayer::new(
+            config,
+            self.genesis_config.l1_batch_commit_data_generator_mode,
```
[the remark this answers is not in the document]
Artemka374: it's batch commitment mode that is used for getting the proof data

PR #2539, opened 2024-07-30
refactor: made attestation start from consensus genesis instead of the last sealed batch

pompon0:
Stable starting point prevents race conditions.


core/node/api_server/src/web3/namespaces/en.rs
```diff
@@ -44,18 +37,30 @@ impl EnNamespace {
...
+            .build()
+            .await
+            .context("TransactionBuilder::build()")?
+            .consensus_dal()
+            .attestation_status()
+            .await?;
+
+        Ok(status.map(|s| {
+            en::AttestationStatus(
+                zksync_protobuf::serde::serialize(&s, serde_json::value::Serializer).unwrap(),
```
aakoshh, line 61: What is the rationale behind returning a wrapped JSON value here? 

I see this is what the `consensus_genesis_impl` does as well, but it returns `zksync_consensus_roles::validator::messages::consensus::Genesis`, which seems like a public and well-known type, it's obvious what to deserialise it into, as every consensus client should have access to it. Whereas here we are returning a type defined inside the DAL consensus module. Are those supposed to be part of the web interface? Normally I would not expect this to be the case because I don't expect web DTOs to bring in `sqlx` dependencies, which is why I suppose it's returned as JSON. I am guessing that's because the consensus types aren't available where the web types are declared. What are we supposed to deserialise this into? Can't we have a common type declared somewhere accessible to the DAL and the web _client_ at least?

core/lib/dal/src/consensus/proto/mod.proto
```diff
@@ -114,3 +116,8 @@ message PaymasterParams {
   optional bytes paymaster_address = 1; // required; H160
   optional bytes paymaster_input = 2; // required
 }
+
+message AttestationStatus {
+  optional roles.validator.GenesisHash genesis = 1; // required
+  optional uint64 next_batch_to_attest = 2; // required
+}
```
aakoshh, line 123: Is the purpose of these protobuf definitions solely to serialise an `AttestationStatus` record to JSON? Curious, what would be the downside of relying on `#[derive(serde::Serialize)]` instead?

aakoshh (commented):
Just some conceptual questions.

core/lib/dal/src/consensus/proto/mod.proto
pompon0: a) lack of backward compatibility control over the format
b) GenesisHash doesn't implement Serialize and it shouldn't because then the internal structure becomes public API.
If you don't like it, I can do a wrapper for just GenesisHash instead.

core/node/api_server/src/web3/namespaces/en.rs
pompon0: There are some people which didn't want me to add consensus crates as a dependency of zksync_types crate. Then someone else came and added zksync_protobuf as dependency of zksync_types anyway. One may argue that zksync_consensus_roles is different (not my opinion) and that's why it shouldn't be a dependency of zksync_types. I didn't want to waste time arguing about that, so just followed a pattern that got accepted.

> Can't we have a common type declared somewhere accessible to the DAL and the web client at least?

I'm open to suggestions where to move it.

core/lib/dal/src/consensus_dal.rs
```diff
@@ -466,38 +466,67 @@ impl ConsensusDal<'_, '_> {
...
+        )
+        .instrument("batch_of_block")
+        .report_latency()
+        .fetch_optional(self.storage)
+        .await?
+        else {
+            return Ok(None);
+        };
+        let Some(batch) = row.l1_batch_number else {
+            return Ok(None);
```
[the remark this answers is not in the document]
pompon0: > Although I wonder why it seems to go from 1, when I thought batches start from 0

It is possible because batch 0 is hardcoded. I agree that's an unnecessary assumption for this query.

> I see it's used to query the batch number of the genesis block, and since the status returns the next to vote for regardless of whether it even exists, it would make sense to return the next available batch number without it having to be sealed first.

ok

core/lib/dal/src/consensus/proto/mod.proto
aakoshh: Could we not manually implement `Serialize` for `GenesisHash`? Not that it's needed, just asking. 
I thought it might be backward compatibility. It's certainly an interesting world, because Protobuf would be backward compatible due to the numeric tags, and `serde` supports aliases for fields, but our CI rejects any renamings for example. It's nice to have Protobuf when we use binary _and_ JSON, I was just a little surprised at the level of ceremony needed to return a simple untyped JSON structure. No need to do anything, just wanted to understand why we do it.

core/node/api_server/src/web3/namespaces/en.rs
aakoshh: Thanks for the explanation, I thought it might be something like this. 

I see that `zksync_types` is a dependency for `zksync_dal`, so we could keep the fully fledged `AttestationStatus` where it is now, next to the `SyncBlock`, and return it from the DAL as-is. I'm not sure what that means for the protobuf format which is also defined in the DAL; I don't think it's used for `SyncBlock`, is it? I see `SyncDal::sync_block` converts from `zksync_dal::models::storage_sync::SyncBlock` to `zksync_types::api::en::SyncBlock`, and the former is mapped from a `StorageSyncBlock`, so there is a whole series of DTOs but the point is the DAL returns what the web API serves in this case. Since we query individual fields we could populate it directly. 

I understand why `ConsensusGenesis` return JSON, who would want to repeat all the fields if they don't have to, but I'm not sure we should apply the same pattern just because it's another consensus related thing, while the DTO and the protobuf format that controls its JSON format is inaccessible.

core/lib/dal/src/consensus/proto/mod.proto
aakoshh: I guess you could but you don't know exactly which format to apply in JSON. In the past I used the [is_human_readable](https://docs.rs/serde/latest/serde/trait.Serializer.html#method.is_human_readable) method to see if I should apply the binary format or the string based one. We could use `TextFmt` when it's supposed to be human readable and `ByteFmt` when it's not, to derive `Serialize` and `Deserialize` when both of these traits are implemented for a type.

aakoshh: Argh, of course it doesn't matter if we can't use it because the dependency is inaccessible 🤦

core/node/api_server/src/web3/namespaces/en.rs
aakoshh: I suppose another option is to move the DTO and proto definition over to the consensus crate and use the JSON route in the API. But don't ask me under what package these web-api related types should live.

core/lib/dal/src/consensus/proto/mod.proto
pompon0: we maintain both WIRE and WIRE_JSON compatibility: https://buf.build/docs/breaking/rules . Technically we could limit it to WIRE_JSON for some packages, but that would require totally isolating the stuff which is JSON encodable and binary encodable, which is not worth it as it disallows interoperability. In era-consensus protocol we use binary because it is already there, here we use JSON, well, because it is a json rpc server.

core/node/api_server/src/web3/namespaces/en.rs
pompon0: I mean, I treat this endpoint as a transport layer (we could use binary encoding afaic) and I don't really want/need it to be interpreted outside the scope of the consensus.

AFAICT what you are trying to achieve is to use a stronger type in `zksync_types::api::en`. I see a value in that, but the apparent discrimination against adding dependencies to zksync_types (and any non-serde encoding for that matter) is too discouraging for me to bother with that.

aakoshh: You are right, I focussed on what would be a nice type for the API to expose, and immediately thought that surely the DAL types are private and won't help to deserialise the data. I forgot that we actually only want to use this API from the node itself, where the DAL _is_ accessible, so I won't have any trouble parsing the JSON regardless of whether this type and its proto resides here, or in the era-consensus repo. Of course keeping it here is more convenient, so I'm happy with it.

aakoshh (approved):

PR #2544, opened 2024-07-30
feat: Poll the main node for the next batch to sign (BFT-496)

aakoshh:
## What ❔

Injects an `AttestationStatusWatch` into the `Executor` which on the main node is backed by an `AttestationStatusRunner` polling the `BatchStore::next_batch_to_attest` and external nodes call the `MainNodeApi::fetch_attestation_status`. 

TODO: 
- [x] Rebase after https://github.com/matter-labs/zksync-era/pull/2539 is merged 
- [x] Update the `era-consensus` dependency once https://github.com/matter-labs/era-consensus/pull/161  is merged and `0.1.0-rc.5` is published

### Test failure - pruning the main node

The following test never finished:
```shell
zk test rust -p zksync_node_consensus tests::test_with_pruning::case_0 --no-capture
```
A little extra logging revealed that the `AttesterStatusRunner` never gets initialised, because the blocks get pruned earlier than it could read the result, ie. `ConsensusDal::batch_of_block(genesis.first_block)` probably returns `None` because it's not found, and never will because it's pruned. 

We discussed in https://github.com/matter-labs/zksync-era/pull/2480 that the main node is not supposed to be pruned, which is why the SQL that looks for what to attest on doesn't look for pruning records any more. Yet this test prunes the main node, and if I just try to prune the external node instead it panics because it doesn't have that block yet. 

Either the SQL should take pruning into account after all, or we have to figure out a way to wait in the test until the main node executor is running, or change the test to prune the external node; I did the latter.
 
## Why ❔

This is coordinating attesters to cast their votes on the batch which the main node tells them to do to produce a quorum certificate for all L1 batches without gaps.

## Checklist


core/node/consensus/src/tests.rs
```diff
@@ -616,8 +616,16 @@ async fn test_with_pruning(version: ProtocolVersionId) {
             .wait_for_batch(ctx, validator.last_sealed_batch())
             .await?;
 
+        // The main node is not supposed to be pruned. In particular `ConsensusDal::attestation_status`
+        // does not look for where the last prune happened at, and thus if we prune the block genesis
+        // points at, we might never be able to start the Executor.
```
RomanBrodetski, line 621: the main node is not pruned currently, but there is nothing preventing us from doing that. FWIW, we actually plan to prune it at some point. How hard is it to not rely on this? No need to do now, but let's at least create a ticket to resolve this

pompon0: Not committed batches are never pruned. We don't have such nice guarantee here, because we only simulate the L1 interaction, but once committing a batch requires attestation certificate, pruning won't be an issue. There is nothing to fix there.

core/node/consensus/src/en.rs
```diff
@@ -1,7 +1,14 @@
-use anyhow::Context as _;
+use anyhow::{anyhow, Context as _};
```
pompon0, line 1: nit: format_err (it is an alias, but the name is more descriptive)

core/node/consensus/src/en.rs
```diff
@@ -1,7 +1,14 @@
...
+use async_trait::async_trait;
 use zksync_concurrency::{ctx, error::Wrap as _, scope, time};
-use zksync_consensus_executor as executor;
-use zksync_consensus_roles::validator;
+use zksync_consensus_executor::{
+    self as executor,
+    attestation::{AttestationStatusClient, AttestationStatusRunner},
+};
+use zksync_consensus_roles::{
+    attester::{self, GenesisHash},
```
pompon0, line 9: nit: please use qualified names

core/node/consensus/src/storage/store.rs
```diff
@@ -444,44 +444,18 @@ impl storage::PersistentBatchStore for Store {
...
-            .wrap("get_last_batch_number")?
-        else {
-            return Ok(None);
-        };
-
-        Ok(Some(attester::BatchNumber(
-            last_batch_number.0.saturating_sub(NUM_BATCHES_TO_SIGN),
-        )))
+            .wrap("next_batch_to_attest")?
+            .map(|s| s.next_batch_to_attest))
```
pompon0, line 458: once you rebase, remember to verify that genesis hash is consistent.

aakoshh: This is retrieving the `AttestationStatus` from the database. Do you mean to also retrieve the genesis from the database and compare? How could it be different?

Cargo.toml
```diff
@@ -181,7 +181,7 @@ tower-http = "0.5.2"
 tracing = "0.1"
 tracing-subscriber = "0.3"
 tracing-opentelemetry = "0.25.0"
-time = "0.3.36" # Has to be same as used by `tracing-subscriber`
+time = "0.3.36"                                                               # Has to be same as used by `tracing-subscriber`
```
pompon0, line 184: that's a lot of spaces XD

aakoshh: Yeah I don't know how that happened 😳

core/node/consensus/src/storage/store.rs
pompon0: scenario: the main node restarts, reverts some batches, does regenesis, we don't notice that the next_batch_to_attest moved back, the era-consensus code throws some error, the whole node shuts down with a different error than it should (should be "genesis changed", while we get "next_batch_to_attest decreased" or some even more enigmatic error, depending on what exactly fails).

It is not a huge difference, just preventing some garbage getting into the system.

Cargo.toml
pompon0: dunno, you have quite an aggressive editor ;)

pompon0 (approved):

core/node/consensus/src/storage/store.rs
aakoshh: Okay there are certainly some misunderstandings on my part here: I thought that to do a revert the main node is stopped, some revert CLI is executed, a new genesis is configured, the main node is started, it does [regenesis, and then it starts](https://github.com/matter-labs/zksync-era/blob/c4d8c5266610deda3d43d626191be8c766c9835f/core/node/consensus/src/mn.rs#L39) the stores and the attestation status. If the genesis could change then the `BlockStore` would be in trouble as it caches it in the beginning. 

Another problem is that the `AttesterStatusRunner` simply retries after an error, the only thing it doesn't do is not update the status. Maybe it should have an option to trait errors as fatal, or let the client return a different result to indicate that the whole thing is borked, or do the check itself. This looked fine for the EN as it will eventually stop when it sees the genesis changed.

I'll open another PR that changes the `AttestationStatusClient` to return the whole status so this can be inspected and differentiated from RPC errors.

Cargo.toml
aakoshh: Called Even Better TOML, so... had to move it over the line.

core/node/consensus/src/storage/store.rs
pompon0: > the main node is started, it does regenesis, and then it starts

that's true. Then EN observe the genesis of the main node and restart if it changes. But it is not atomic.

> Maybe it should have an option to trait errors as fatal

I see, I think that's ok for it not to be fatal. Then this is issue is even less important, as the wrong error will be only logged.

aakoshh: https://github.com/matter-labs/era-consensus/pull/167

core/node/consensus/src/tests.rs
aakoshh: For the record the very first attempt that looked at pruning was [here](https://github.com/matter-labs/zksync-era/pull/2480/commits/3953ffd838dc644371aebbf7f860809a4f33ecf1#diff-9958fba2c9e6d7938fdd3bea6e06596e3cff47eb06dda4e440f9117f342495d0R51).

brunoffranca (approved):

PR #2545, opened 2024-07-31
feat(configs): Do not panic if config is only partially filled

Deniallugo:
## What ❔

If config is partially serialized, do not throw an error just print it and the app, for which part the config is required will panic by itself 

## Checklist

### DIFF
--- a/core/lib/protobuf_config/src/lib.rs
+++ b/core/lib/protobuf_config/src/lib.rs
@@ -54,8 +54,18 @@ fn parse_h160(bytes: &str) -> anyhow::Result<H160> {
     Ok(H160::from_str(bytes)?)
 }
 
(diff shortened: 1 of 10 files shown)


core/lib/protobuf_config/src/lib.rs
```diff
@@ -54,8 +54,20 @@ fn parse_h160(bytes: &str) -> anyhow::Result<H160> {
...
+        .as_ref()
+        .map(|x| x.read())
+        .transpose()
+        // This error will printed, only if the config partially filled, allows to debug config issues easier
+        .map_err(print_error)
+        .ok()
+        .flatten()
+}
+
+pub fn print_error(error: anyhow::Error) -> anyhow::Error {
```
popzxc, line 68: Doesn't have to be public (and generally can even be a lambda).

popzxc (dismissed):

PR #2546, opened 2024-07-31
feat(consensus): add tracing instrumentation to consensus store

itegulov:
## What ❔

This PR adds instrumentation to consensus store layer which should serve as a good starting point. Most other layers are heavily intertwined with the `era-consensus` codebase so we will need to introduce instrumentation there first.

## Why ❔

Better visibility of what we spend our time on, including waiting time

## Checklist

### DIFF
--- a/core/node/consensus/src/en.rs
+++ b/core/node/consensus/src/en.rs
@@ -1,4 +1,5 @@
 use anyhow::Context as _;
+use tracing::Instrument;
 use zksync_concurrency::{ctx, error::Wrap as _, scope, time};
 use zksync_consensus_executor as executor;
 use zksync_consensus_roles::validator;
@@ -70,7 +71,11 @@ impl EN {
             s.spawn_bg::<()>(async {
                 let old = genesis;
                 loop {
-                    if let Ok(new) = self.fetch_genesis(ctx).await {
+                    if let Ok(new) = self
+                        .fetch_genesis(ctx)
+                        .instrument(tracing::info_span!("genesis_monitor_fetch"))
+                        .await
+                    {
                         if new != old {
                             return Err(anyhow::format_err!(
                                 "genesis changed: old {old:?}, new {new:?}"
--- a/core/node/consensus/src/storage/connection.rs
+++ b/core/node/consensus/src/storage/connection.rs
@@ -106,6 +106,7 @@ impl<'a> Connection<'a> {
     }
 
     /// Wrapper for `consensus_dal().insert_block_certificate()`.
+    #[tracing::instrument(skip_all, fields(block_number = %cert.message.proposal.number))]
     pub async fn insert_block_certificate(
         &mut self,
         ctx: &ctx::Ctx,
@@ -118,6 +119,7 @@ impl<'a> Connection<'a> {
 
     /// Wrapper for `consensus_dal().insert_batch_certificate()`,
     /// which additionally verifies that the batch hash matches the stored batch.
+    #[tracing::instrument(skip_all, fields(batch_number = %cert.message.number))]
     pub async fn insert_batch_certificate(
         &mut self,
         ctx: &ctx::Ctx,
@@ -223,11 +225,13 @@ impl<'a> Connection<'a> {
     }
 
     /// Wrapper for `consensus_dal().next_block()`.
+    #[tracing::instrument(skip_all)]
     async fn next_block(&mut self, ctx: &ctx::Ctx) -> ctx::Result<validator::BlockNumber> {
         Ok(ctx.wait(self.0.consensus_dal().next_block()).await??)
     }
 
     /// Wrapper for `consensus_dal().block_certificates_range()`.
+    #[tracing::instrument(skip_all)]
     pub(crate) async fn block_certificates_range(
         &mut self,
         ctx: &ctx::Ctx,
@@ -305,6 +309,7 @@ impl<'a> Connection<'a> {
     }
 
     /// Wrapper for `blocks_dal().get_sealed_l1_batch_number()`.
+    #[tracing::instrument(skip_all)]
     pub async fn get_last_batch_number(
         &mut self,
         ctx: &ctx::Ctx,
@@ -390,6 +395,7 @@ impl<'a> Connection<'a> {
     }
 
     /// Construct the [storage::BatchStoreState] which contains the earliest batch and the last available [attester::SyncBatch].
+    #[tracing::instrument(skip_all)]
     pub async fn batches_range(&mut self, ctx: &ctx::Ctx) -> ctx::Result<storage::BatchStoreState> {
         let first = self
             .0
--- a/core/node/consensus/src/storage/store.rs
+++ b/core/node/consensus/src/storage/store.rs
@@ -1,6 +1,7 @@
 use std::sync::Arc;
 
 use anyhow::Context as _;
+use tracing::Instrument;
 use zksync_concurrency::{ctx, error::Wrap as _, scope, sync, time};
 use zksync_consensus_bft::PayloadManager;
 use zksync_consensus_roles::{attester, validator};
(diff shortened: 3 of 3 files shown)


core/node/consensus/src/en.rs
```diff
@@ -70,7 +71,11 @@ impl EN {
             s.spawn_bg::<()>(async {
                 let old = genesis;
                 loop {
-                    if let Ok(new) = self.fetch_genesis(ctx).await {
+                    if let Ok(new) = self
+                        .fetch_genesis(ctx)
+                        .instrument(tracing::info_span!("genesis_monitor_fetch"))
```
pompon0, line 76: Is there any specific rule for this naming (i.e. why not "fetch_genesis")?
Why here though? Shouldn't you rather add `#[tracing...]` to fetch_genesis method instead?
If you want to instrument the whole subtask, you can put it on the outer future, like `s.spawn_bg(async { ... }.instrument(info_span));`

core/node/consensus/src/storage/store.rs
```diff
@@ -175,14 +178,23 @@ impl StoreRunner {
...
                 loop {
-                    let range = pool
-                        .connection(ctx)
-                        .await?
-                        .block_certificates_range(ctx)
-                        .await
-                        .wrap("block_certificates_range()")?;
-                    blocks_persisted.update(range);
-                    ctx.sleep(POLL_INTERVAL).await?;
+                    let span = tracing::info_span!("blocks_persisted_loop");
```
pompon0, line 181: ditto, you can put it on the outer `async { ... }` instead

core/node/consensus/src/storage/store.rs
```diff
@@ -203,93 +215,130 @@ impl StoreRunner {
...
-                    let cert = batch_certificates.recv(ctx).await?;
+                    let span = tracing::info_span!("insert_batch_certificates_loop");
+
+                    async {
+                        let cert = batch_certificates
+                            .recv(ctx)
+                            .instrument(tracing::info_span!("wait_for_certificate"))
+                            .await?;
+
+                        loop {
```
pompon0, line 258: any reason you switched the order of the tasks?

core/node/consensus/src/en.rs
aakoshh: I also think if we want to add more tracing we should add them as an attribute to the method, it keeps the call site easier to read.

core/node/consensus/src/storage/store.rs
```diff
@@ -175,14 +178,23 @@ impl StoreRunner {
...
+                    async {
+                        let range = pool
+                            .connection(ctx)
+                            .await?
+                            .block_certificates_range(ctx)
+                            .await
+                            .wrap("block_certificates_range()")?;
+                        blocks_persisted.update(range);
+                        ctx.sleep(POLL_INTERVAL)
+                            .instrument(tracing::info_span!("sleep"))
```
aakoshh, line 191: Is this going to emit an INFO level log every time it enters/exits sleep? Do we really need this level of detail?

Unlike the potential sleeps involved in saving certificates, this just polls the DB to keep an in-memory data structure in sync with it. It is interesting to see those others as they indicate this node is behind others in some respect, but seeing logs from this every second would be a bit tedious.

core/node/consensus/src/storage/store.rs
```diff
@@ -203,93 +215,130 @@ impl StoreRunner {
...
                 // Loop inserting batch certificates into storage
                 const POLL_INTERVAL: time::Duration = time::Duration::milliseconds(50);
                 loop {
-                    let cert = batch_certificates.recv(ctx).await?;
+                    let span = tracing::info_span!("insert_batch_certificates_loop");
+
+                    async {
+                        let cert = batch_certificates
+                            .recv(ctx)
+                            .instrument(tracing::info_span!("wait_for_certificate"))
```
aakoshh, line 255:
```suggestion
                            .instrument(tracing::info_span!("wait_for_batch_certificate"))
```

core/node/consensus/src/storage/store.rs
```diff
@@ -203,93 +215,130 @@ impl StoreRunner {
...
 
-                    loop {
+            // Loop inserting block certs to storage.
+            const POLL_INTERVAL: time::Duration = time::Duration::milliseconds(50);
+            loop {
+                let span = tracing::info_span!("insert_block_certificates_loop");
+                async {
+                    let cert = block_certificates
+                        .recv(ctx)
+                        .instrument(tracing::info_span!("wait_for_certificate"))
```
aakoshh, line 302:
```suggestion
                        .instrument(tracing::info_span!("wait_for_block_certificate"))
```

aakoshh (commented):
Nice, more visibility is surely going to help you understand what's happening. Keeping the code readable at the same time is a balancing act.

core/node/consensus/src/en.rs
itegulov: Good point, this was a leftover from my attempt at covering more stuff here with spans. Changed to this being a method attribute

core/node/consensus/src/storage/store.rs
itegulov: Correct me if I am wrong, but this way the span would never be exited. The loop spans are ugly but I am unsure on how to achieve something similar with less invasive approach.

core/node/consensus/src/storage/store.rs
itegulov: I didn't change the order of the tasks. Github just generated a weird diff because of the indentation changes.

core/node/consensus/src/storage/store.rs
itegulov: Fair point, removed

itegulov: Also, just to be clear: it won't produce a log but rather a span within a trace so shouldn't pollute the logs at all. That being said I don't think there is much merit to having this tracked.

core/node/consensus/src/storage/connection.rs
```diff
@@ -106,6 +106,7 @@ impl<'a> Connection<'a> {
     }
 
     /// Wrapper for `consensus_dal().insert_block_certificate()`.
+    #[tracing::instrument(skip_all, fields(block_number = %cert.message.proposal.number))]
```
popzxc, line 109: JFYI: In the recently added traces I've tried to use field names `l1_batch` and `l2_block`, so that it's easy to build consistent filters when you're looking for traces. There is no strict notion right now, but it may be beneficial to follow the same convention everywhere. I'll prepare a doc on tracing a bit later.

core/node/consensus/src/storage/store.rs
```diff
@@ -203,93 +213,130 @@ impl StoreRunner {
...
+                                let range =
+                                    conn.batches_range(ctx).await.wrap("batches_range()")?;
+                                next_batch_number = last_batch_number.next();
+                                tracing::info_span!("batches_persisted_send").in_scope(|| {
+                                    batches_persisted.send_replace(range);
+                                });
+                            }
                         }
+                        ctx.sleep(POLL_INTERVAL)
+                            .instrument(tracing::info_span!("sleep"))
```
popzxc, line 234: Again, not strictly followed right now, but I've already noticed that similar spans make the analysis in fact harder. Probably makes sense to make it `debug_span`

core/node/consensus/src/storage/store.rs
popzxc: Yup, the span must be within the loop.

I'm also concerned by the somewhat intrusive spans in the code, but I think if something is interesting for us _as a span_ then it probably deserves its own function.

popzxc (approved):
Mostly LGTM

core/node/consensus/src/storage/store.rs
itegulov: Looking at this one more time I think this particular sleep shouldn't be covered at all. Other sleep are rare (only if a node is behind) so I think it's fine to keep it as `INFO` for now.

core/node/consensus/src/storage/store.rs
itegulov: That's a fair point and I would be glad to refactor loop body into a function if @pompon0 agrees

pompon0 (approved):

pompon0 (approved):

PR #2547, opened 2024-07-31
Adding SLChainID

StanislavBreadless:
## What ❔

Adding settlement layer chain id. Note, that in the future zkSync may settle onto other instances of Era. This PR does not introduce any legacy-inducing changes (e.g. amending configs) in case we'll need to change something.

I also did not rename components / structs that mention "Eth", e.g. `QueryEthClientLayer`. For now, we can say that everything that says Eth means "Eth-compatible".

With this PR:
- L1ChainId should not be used only when *exactly* L1 is meant and not settlement layer.
- SLChainId should be used otherwise.

## Checklist

### DIFF
--- a/core/bin/external_node/src/config/mod.rs
+++ b/core/bin/external_node/src/config/mod.rs
@@ -31,7 +31,7 @@ use zksync_protobuf_config::proto;
 use zksync_snapshots_applier::SnapshotsApplierConfig;
 use zksync_types::{
     api::BridgeAddresses, commitment::L1BatchCommitmentMode, url::SensitiveUrl, Address,
-    L1BatchNumber, L2ChainId, SLChainId, ETHEREUM_ADDRESS,
+    L1BatchNumber, L1ChainId, L2ChainId, SLChainId, ETHEREUM_ADDRESS,
 };
 use zksync_web3_decl::{
     client::{DynClient, L2},
@@ -908,9 +908,11 @@ impl OptionalENConfig {
 /// This part of the external node config is required for its operation.
 #[derive(Debug, Deserialize)]
 pub(crate) struct RequiredENConfig {
-    /// The chain ID of the settlement layer (e.g., 9 for Ethereum mainnet). This ID will be checked against the `eth_client_url` RPC provider on initialization
-    /// to ensure that there's no mismatch between the expected and actual L1 network.
-    pub l1_chain_id: SLChainId,
+    /// The chain ID of the L1 network (e.g., 1 for Ethereum mainnet). In the future, it may be different from the settlement layer.
+    pub l1_chain_id: L1ChainId,
+    /// The chain ID of the settlement layer (e.g., 1 for Ethereum mainnet). This ID will be checked against the `eth_client_url` RPC provider on initialization
+    /// to ensure that there's no mismatch between the expected and actual settlement layer network.
+    pub sl_chain_id: Option<SLChainId>,
     /// L2 chain ID (e.g., 270 for ZKsync Era mainnet). This ID will be checked against the `main_node_url` RPC provider on initialization
     /// to ensure that there's no mismatch between the expected and actual L2 network.
     pub l2_chain_id: L2ChainId,
@@ -932,6 +934,10 @@ pub(crate) struct RequiredENConfig {
 }
 
 impl RequiredENConfig {
+    pub fn get_settlement_layer_id(&self) -> SLChainId {
+        self.sl_chain_id.unwrap_or(self.l1_chain_id.into())
+    }
+
     fn from_env() -> anyhow::Result<Self> {
         envy::prefixed("EN_")
             .from_env()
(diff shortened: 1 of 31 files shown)


kelemeno (approved):

core/bin/external_node/src/config/mod.rs
```diff
@@ -932,6 +934,10 @@ pub(crate) struct RequiredENConfig {
 }
 
 impl RequiredENConfig {
+    pub fn get_settlement_layer_id(&self) -> SLChainId {
```
slowli, line 937: Nit: getters in Rust [usually don't have `get_` prefix](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter), so `settlement_layer_id` would look better.

core/bin/external_node/src/metrics/mod.rs
```diff
@@ -31,13 +31,13 @@ pub(crate) struct ExternalNodeMetrics {
...
         &self,
-        l1_chain_id: L1ChainId,
+        sl_chain_id: SLChainId,
         l2_chain_id: L2ChainId,
         postgres_pool_size: u32,
     ) {
         let info = ExternalNodeInfo {
             server_version: SERVER_VERSION,
-            l1_chain_id: l1_chain_id.0,
+            l1_chain_id: sl_chain_id.0,
```
slowli, line 40: Will L1 chain ID stay as an EN configuration parameter? If so, IMO it'd make more sense to *add* a new field to `ExternalNodeInfo` rather than redefining semantics of an existing field.

core/lib/protobuf_config/src/genesis.rs
```diff
@@ -2,11 +2,11 @@ use std::str::FromStr;
 
 use anyhow::Context as _;
 use zksync_basic_types::{
-    commitment::L1BatchCommitmentMode, protocol_version::ProtocolSemanticVersion, L1ChainId,
-    L2ChainId,
+    commitment::L1BatchCommitmentMode, protocol_version::ProtocolSemanticVersion, L2ChainId,
 };
 use zksync_config::configs;
 use zksync_protobuf::{repr::ProtoRepr, required};
+use zksync_types::L1ChainId;
```
slowli, line 9: Nit: Probably makes sense to import from `zksync_basic_types` for consistency.

core/lib/web3_decl/src/client/network.rs
```diff
@@ -12,9 +12,10 @@ pub trait Network: 'static + Copy + Default + Sync + Send + fmt::Debug {
     fn metric_label(&self) -> String;
 }
 
-/// L1 (i.e., Ethereum) network.
+/// L1-compatible (e.g., Ethereum) network.
+/// Note, that this network does not have to be an L1. It can be any network that is compatible with Ethereum JSON-RPC.
 #[derive(Debug, Clone, Copy, Default)]
-pub struct L1(Option<L1ChainId>);
+pub struct L1(Option<SLChainId>);
```
slowli, line 18: Maybe rename this struct something like `SL` and document that it represents a network for settlement? At least, this is how this marker is used in the `eth_client` crate.

slowli (commented):
IMO, this change makes complete sense.

core/lib/web3_decl/src/client/network.rs
StanislavBreadless: @slowli I keep this struct as `L1` because there are structs like `DynClient<L1>` which support Eth interface as it is on Ethereum and not for e.g. on L2. I am not sure what is better here, I rename it to `L1CompatibleSL`, does it look good?

core/lib/env_config/src/genesis.rs
```diff
@@ -68,7 +68,9 @@ impl FromEnv for GenesisConfig {
             genesis_commitment: contracts_config.genesis_batch_commitment,
             bootloader_hash: state_keeper.bootloader_hash,
             default_aa_hash: state_keeper.default_aa_hash,
-            l1_chain_id: network_config.network.chain_id(),
+            // TODO(X): for now, the settlement layer is always the same as the L1 network
```
popzxc, line 71: X?

zk_toolbox/crates/config/src/genesis.rs
```diff
@@ -13,6 +13,7 @@ use crate::{
 
 pub fn update_from_chain_config(genesis: &mut GenesisConfig, config: &ChainConfig) {
     genesis.l2_chain_id = config.chain_id;
+    // TODO(X): for now, the settlement layer is always the same as the L1 network
```
popzxc, line 16: X?

popzxc (approved):
Please change the PR type from `chore` to `feat` so that it appears in the changelog and more people see it. Also, after merging, announce it additionally in the chat for more awareness.

core/lib/web3_decl/src/client/network.rs
slowli: It's fine to keep it as-is, I suppose, especially given that types containing "eth" in their name are not renamed.

popzxc (approved):

PR #2549, opened 2024-07-31
chore(eth-sender): Do not wait for new blocks to rerun the eth sender loop iteration

StanislavBreadless:
## What ❔

When settling on top of an L2, new blocks may not be produced without transactions from the operator, so we need to re-run the iteration even if no blocks are produced.

## Checklist


core/node/eth_sender/src/eth_tx_manager.rs
```diff
@@ -549,7 +549,12 @@ impl EthTxManager {
             let l1_block_numbers = self.l1_interface.get_l1_block_numbers().await?;
             METRICS.track_block_numbers(&l1_block_numbers);
 
-            if last_known_l1_block < l1_block_numbers.latest {
+            // We need to re-run the iteration even if the latest seen block number is the same as the previous one, since
+            // in case of a low-activity settlement layer network, new blocks may only be produced by the transactions from the operator itself.
+            //
+            // We do not skip the check entirely, to ensure the monotonicity of the `last_known_l1_block`
+            // in case of bugs of the JSON-RPC API provider.
+            if last_known_l1_block <= l1_block_numbers.latest {
```
Deniallugo, line 557: I want to delete this check completely. 
https://github.com/matter-labs/zksync-era/pull/2535.

In this case, it doesn't make sense actually, the blocks most likely won't go down and if they will we have to reiterate on them.

Deniallugo: In this case it won't throttle anything

tomg10: You are right, the code needs to be changed so that if the block number didn't  change, then don't resent txs and only send new ones

StanislavBreadless: @tomg10 But don't we already have the following logic?

```
                if first_sent_at_block == Some(l1_block_numbers.latest.0) {
                    continue;
                }
```

tomg10: it works a bit different, if only prevents tx just sent at block N from being re-sent at block N. It won't prevent tx that was resent at block N not to be resent again in next iteration at block N

tomg10: we can probably modify it though

tomg10: so that it checks against last_sent_at_block, not first

Deniallugo: @tomg10 i've slightly changed the logic, could you verify it?

tomg10: I've just pushed a commit with a test for exactly that

core/lib/dal/src/eth_sender_dal.rs
```diff
@@ -528,6 +528,19 @@ impl EthSenderDal<'_, '_> {
         Ok(sent_at_block.flatten().map(|block| block as u32))
     }
 
+    pub async fn get_block_number_on_last_sent_attempt(
+        &mut self,
+        eth_tx_id: u32,
+    ) -> sqlx::Result<Option<u32>> {
+        let sent_at_block = sqlx::query_scalar!(
+            "SELECT sent_at_block FROM eth_txs_history WHERE eth_tx_id = $1 AND sent_at_block IS NOT NULL ORDER BY created_at DESC LIMIT 1",
```
Deniallugo, line 536: probably we could just use `max` and the query will be pretty simple. 
`SELECT max(sent_at_block) FROM eth_txs_history WHERE eth_tx_id = $1`

core/node/eth_sender/src/eth_tx_manager.rs
```diff
@@ -334,16 +334,22 @@ impl EthTxManager {
             // that `tx` is not mined and we should resend it.
             // We only resend the first un-mined transaction.
             if operator_nonce.latest <= tx.nonce {
+                let last_sent_at_block = storage
+                    .eth_sender_dal()
+                    .get_block_number_on_last_sent_attempt(tx.id)
```
Deniallugo, line 339: I'm a bit afraid of this approach, because we started to do more queries to L1 rpc provider.
The previous solution didn't have this problem

core/lib/dal/src/eth_sender_dal.rs
tomg10: nice, corrected

core/node/eth_sender/src/eth_tx_manager.rs
tomg10: I personally don't worry about it, it's indeed two queries per iteration more worst case, but only if we decreased tx_poll_period by a lot and it's currently set to 20 seconds on every env

tomg10: and this change will simplify code adding Gateway support as with it there isn't a single block numbers, but Gateway has different block number than L1

core/node/eth_sender/src/eth_tx_manager.rs
```diff
@@ -334,16 +334,22 @@ impl EthTxManager {
...
+                    .await
+                    .unwrap();
```
Deniallugo, line 341: is it safe to unwrap?

I'd really stick to the previous approach this one looks more error prone

tomg10: It's safe to unwrap as if a transaction is in-flight, then it has at least one send attempt. And maybe I'm missing something, but there is an unwrap on almost the same method call, but calculating first, not last attempt just 10 lines below.

Deniallugo (approved):

PR #2551, opened 2024-07-31
feat: added a flag to main node to just verify configs (BFT-501)

pompon0:
It will be used on CI to verify configs in gitops repo.
I've considered implementing it as a subcommand instead of a flag, but afaiu adding flags is the current design pattern. I can change that, lmk wdyt.


pompon0:
Probably I should exclude verification of secrets, at least initially.

Deniallugo:
after this pr https://github.com/matter-labs/zksync-era/pull/2545 it doesn't make much sense

Deniallugo:
I mean ci check is good, but this particular implementation, will show nothing

PR #2554, opened 2024-07-31
chore(eth-sender): remove storing gas estimation for eth-sender transactions

tomg10:
removing it as we don't actively use it


perekopskiy:

perekopskiy (commented):

PR #2559, opened 2024-08-01
refactor(state-keeper): Use owned VM storage for batch executor

slowli:
## What ❔

- Refactors `ReadStorageFactory` and `BatchExecutor`.
- Tests the VM runner on multiple batch workloads and fixes bugs discovered by the tests (e.g., a data race in the storage loader).

## Why ❔

- `ReadStorageFactory` and `BatchExecutor` become more composable and work in VM runner with less crutches.
- More tests allow fixing bugs in VM runner earlier.

## Checklist

fix(vm-runner): Fix data race in storage loader
test(vm-runner): Test VM runner with multiple L1 batches

### DIFF
--- a/core/node/state_keeper/src/batch_executor/main_executor.rs
+++ b/core/node/state_keeper/src/batch_executor/main_executor.rs
@@ -1,7 +1,6 @@
 use std::sync::Arc;
 
 use anyhow::Context as _;
-use async_trait::async_trait;
 use once_cell::sync::OnceCell;
 use tokio::{runtime::Handle, sync::mpsc};
 use zksync_multivm::{
@@ -47,9 +46,8 @@ impl MainBatchExecutor {
     }
 }
 
-#[async_trait]
 impl BatchExecutor for MainBatchExecutor {
-    async fn init_batch(
+    fn init_batch(
         &mut self,
         storage: OwnedStorage,
         l1_batch_params: L1BatchEnv,
--- a/core/node/state_keeper/src/batch_executor/main_executor.rs
+++ b/core/node/state_keeper/src/batch_executor/main_executor.rs
@@ -3,10 +3,7 @@ use std::sync::Arc;
 use anyhow::Context as _;
 use async_trait::async_trait;
 use once_cell::sync::OnceCell;
-use tokio::{
-    runtime::Handle,
-    sync::{mpsc, watch},
-};
+use tokio::{runtime::Handle, sync::mpsc};
 use zksync_multivm::{
     interface::{
         ExecutionResult, FinishedL1Batch, Halt, L1BatchEnv, L2BlockEnv, SystemEnv,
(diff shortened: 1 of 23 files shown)


core/node/state_keeper/src/batch_executor/main_executor.rs
```diff
@@ -68,21 +62,17 @@ impl BatchExecutor for MainBatchExecutor {
...
-                )
-                .context("failed accessing state keeper storage")?
-            {
-                executor.run(storage, l1_batch_params, system_env);
-            } else {
-                tracing::info!("Interrupted while trying to access state keeper storage");
-            }
+            let storage = match storage {
+                OwnedStorage::Static(storage) => storage,
+                OwnedStorage::Lending(ref storage) => Handle::current()
```
slowli, line 68: JFYI: This is one of the few scenarios in which `ref` is seemingly required: the branch above moves `storage`, and in this branch, we must borrow from it.

core/node/state_keeper/src/testonly/test_batch_executor.rs
```diff
@@ -805,55 +803,31 @@ impl StateKeeperIO for TestIO {
...
-    ) -> anyhow::Result<Option<PgOrRocksdbStorage<'_>>> {
-        // Presume that the storage is never accessed in mocked environment
-        unimplemented!()
+    ) -> anyhow::Result<Option<OwnedStorage>> {
+        let storage = RocksdbStorage::builder(self.0.path())
+            .await
+            .expect("Cannot create mock RocksDB storage")
+            .build_unchecked();
+        Ok(Some(PgOrRocksdbStorage::Rocksdb(storage).into()))
     }
```
slowli, line 832: This is quite ugly, but I haven't found a way to mock the storage factory otherwise; with the changes, `access_storage` *is* called by the state keeper regardless of the `BatchExecutor` implementation. I guess theoretically it could be possible to add a mock storage variant to `PgOrRocksdbStorage`, but that would introduce unnecessary noise and may marginally hurt performance.

core/node/vm_runner/src/storage.rs
```diff
@@ -34,13 +34,7 @@ pub trait StorageLoader: ReadStorageFactory {
...
         &self,
         l1_batch_number: L1BatchNumber,
-    ) -> anyhow::Result<Option<BatchExecuteData>>;
-
-    /// A workaround for Rust's limitations on upcasting coercion. See
-    /// https://github.com/rust-lang/rust/issues/65991.
-    ///
-    /// Should always be implementable as [`StorageLoader`] requires [`ReadStorageFactory`].
-    fn upcast(self: Arc<Self>) -> Arc<dyn ReadStorageFactory>;
+    ) -> anyhow::Result<Option<(BatchExecuteData, OwnedStorage)>>;
```
slowli, line 37: With this change, there's no more data race possible between getting `BatchExecuteData` and `BatchExecutor` logic. IMO, it's cleaner as well.

popzxc (approved):
👍

core/node/state_keeper/src/testonly/test_batch_executor.rs
itegulov: It would be cool if we could have an associated type in `ReadStorageFactory` that would regulate which storage type is returned from a particular implementation. Don't think it's feasible though due to our heavy usage of `dyn`

itegulov (approved):
Great improvement! ❤️

core/node/state_keeper/src/testonly/test_batch_executor.rs
slowli: Yes, if we could have a varying output type, there wouldn't be a problem, but as you've said, it doesn't look compatible with trait objects. Will think about it a bit more.

JanKortes212:
ok

PR #2560, opened 2024-08-01
test(vm): Fix fast VM unit tests

slowli:
## What ❔

Fixes unit tests that previously failed because of incorrect `Heap` comparisons.

## Why ❔

Better test coverage.

## Checklist

### DIFF
--- a/core/lib/multivm/src/versions/vm_fast/vm.rs
+++ b/core/lib/multivm/src/versions/vm_fast/vm.rs
@@ -337,23 +337,6 @@ impl<S: ReadStorage> Vm<S> {
         }
     }
 
-    /// Returns the current state of the VM in a format that can be compared for equality.
-    #[cfg(test)]
-    #[allow(clippy::type_complexity)] // OK for tests
-    pub(crate) fn dump_state(&self) -> (vm2::State, Vec<((H160, U256), U256)>, Box<[vm2::Event]>) {
-        // TODO this doesn't include all the state of ModifiedWorld
-        (
-            self.inner.state.clone(),
-            self.inner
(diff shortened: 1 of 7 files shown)


core/lib/multivm/src/versions/vm_fast/vm.rs
```diff
@@ -685,6 +675,8 @@ impl<S: ReadStorage> VmInterfaceHistoryEnabled for Vm<S> {
             self.snapshot.is_none(),
             "cannot create a VM snapshot until a previous snapshot is rolled back to or popped"
         );
+
+        self.delete_history_if_appropriate();
```
slowli, line 679: @joonazan Is this additional call acceptable? It removes unnecessary bootloader heap history recorded outside of a snapshot, so that VM state assertions work as expected. I suppose another way to approach this issue would be to change state comparison logic, but I think deleting history makes sense semantically.

montekki (approved):

core/lib/multivm/src/versions/vm_fast/vm.rs
joonazan: I see, RollbackableMap and heap history accumulate even when there is no snapshot. That is somewhat undesirable, so I think this is a positive change.

joonazan: If the VM stored the external snapshot, it would be able to not build up history when there is no snapshot. This was a significant improvement for API speed in the old VM.

On the other hand, if we want to go for the ability to rewind the VM to any historical point in time, we'd want to never clear history.

PR #2563, opened 2024-08-01
feat: Allow tracking l2 fees for L2-based chains

StanislavBreadless:
## What ❔

In this PR we add changes to the GasAdjuster needed to track the fees on L2-based chains. The PR aims to preserve type-safety for places that should work when L1 interface is available only, so `DynClient<L1>` is left in a lot of places and generally `L1` is the default network to be used. However, some methods were extended to work with any network.

Overall what was added:

- A new resource for accessing L2-specific endpoints.
- A new field is returned from `fee_history`. Clients should not rely on this field. However, we added it so we could test things.
- Some refactoring
- While some new new config fields are added into the Rust code, they are not added to the config so that in case something has to be changed, we do not introduce legacy (esp. in protobuf)
- Into protobuf we only added `RelayedL2Calldata` pubdata type, which would symbolize that the calldata is being relayed to L1 through another L2. I am pretty confident that this type will be needed anyway.

## Checklist


core/lib/eth_client/src/lib.rs
```diff
@@ -154,6 +154,23 @@ pub trait EthInterface: Sync + Send {
...
 
+/// Information about the base fees provided by the L1 client.
+#[derive(Debug, Clone, PartialEq, Eq, Default)]
+pub struct L2Fees {
+    pub base_fee_per_gas: u64,
+    pub pubdata_price: U256,
+}
+
+#[async_trait]
+pub trait ZkSyncInterface: Sync + Send {
```
slowli, line 165: Maybe a dumb suggestion: What about providing a single unified interface both for L1 and L2 settlement layers, which would return `base_fee_per_gas` / `base_fee_per_blob_gas` (zero for L2?) / `pubdata_price` (zero for L1)? In this way, some branching code could be simplified.

core/lib/eth_client/src/lib.rs
```diff
@@ -168,13 +185,18 @@ pub trait EthInterface: Sync + Send {
 /// 2. Consider adding the "unbound" version to the `EthInterface` trait and create a default method
 ///   implementation that invokes `contract` / `contract_addr` / `sender_account` methods.
 #[async_trait]
-pub trait BoundEthInterface: AsRef<DynClient<L1>> + 'static + Sync + Send + fmt::Debug {
+pub trait BoundEthInterface<Net: Network>:
```
slowli, line 188: - Just intuitively, `Net` should be an associated type rather than a type param, like in `DynClient`. I.e., an implementation has the "network" property / cannot implement `BoundEthInterface` for multiple networks.
- Dumb question: Is this trait ever used for `Net != L1`? AFAICT, the only instance is the `MockSettlementLayer` impl, but the workspace still compiles after being narrowed to `impl BoundEthInterface<L1> for MockSettlementLayer<L1>`. So it looks like this parameter is redundant. Would `BoundEthInterface` be used with different networks in the future?

core/lib/types/src/api/mod.rs
```diff
@@ -834,6 +834,36 @@ pub struct TransactionExecutionInfo {
     pub execution_info: Value,
 }
 
+/// The fee history type returned from `eth_feeHistory` call.
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct FeeHistory {
+    /// Lowest number block of the returned range.
+    pub oldest_block: BlockNumber,
```
slowli, line 842: This wasn't done for existing API types because of legacy reasons, but for new types, would it be possible to embed the `web3::FeeHistory` type instead of copy-pasting it?

```rust
#[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
pub struct FeeHistory {
    #[serde(flatten)]
    pub inner: web3::FeeHistory,
    /// An array of effective pubdata price. Note, that this field is L2-specific and only provided by L2 nodes.
    #[serde(default)]
    pub pubdata_price: Vec<U256>,
}
```

Would make it more difficult to screw up / easier to maintain, and clearly show the inheritance between the types.

core/lib/web3_decl/src/client/network.rs
```diff
@@ -55,7 +61,7 @@ impl From<L2ChainId> for L2 {
 
 /// Associates a type with a particular type of RPC networks, such as Ethereum or ZKsync Era. RPC traits created using `jsonrpsee::rpc`
 /// can use `ForNetwork` as a client boundary to restrict which implementations can call their methods.
-pub trait ForNetwork {
+pub trait ForEthereumLikeNetwork {
```
slowli, line 64: Dumb question: What's the reason for this renaming? AFAIU, we still don't have non-"Ethereum-like" networks, so why complicate the trait name?

core/node/fee_model/src/l1_gas_price/gas_adjuster/mod.rs
```diff
@@ -18,6 +18,97 @@ mod metrics;
...
+}
+
+struct GasAdjusterFee {
+    pub base_fee_per_gas: u64,
+    pub base_fee_per_blob_gas: U256,
+    pub l2_pubdata_price: U256,
+}
+
+impl GasAdjusterClient {
+    pub fn from_l1(inner: Box<DynClient<L1>>) -> Self {
```
slowli, line 34: Nit: It would be more idiomatic to have `From<Box<DynClient<L1>>>` impl; ditto for `from_l2`.

core/node/fee_model/src/l1_gas_price/gas_adjuster/mod.rs
```diff
@@ -27,31 +118,49 @@ pub struct GasAdjuster {
...
-        eth_client: Box<DynClient<L1>>,
+        client: GasAdjusterClient,
         config: GasAdjusterConfig,
         pubdata_sending_mode: PubdataSendingMode,
         commitment_mode: L1BatchCommitmentMode,
     ) -> anyhow::Result<Self> {
-        let eth_client = eth_client.for_component("gas_adjuster");
+        // A runtime check to ensure consistent config.
+        if config.l2_mode.unwrap_or_default() {
+            assert!(client.is_l2(), "Must be L2 client in L2 mode");
```
slowli, line 142: Please use `anyhow::ensure!` instead of assertions.

core/node/fee_model/src/l1_gas_price/gas_adjuster/metrics.rs
```diff
@@ -7,9 +7,11 @@ use vise::{Gauge, Metrics};
 pub(super) struct GasAdjusterMetrics {
     pub current_base_fee_per_gas: Gauge<u64>,
     pub current_blob_base_fee: Gauge<u64>,
+    pub current_l2_pubdata_price: Gauge<u64>,
     pub median_base_fee_per_gas: Gauge<u64>,
     pub median_blob_base_fee_per_gas: Gauge<u64>,
     pub median_blob_base_fee: Gauge<u64>,
+    pub median_l2_pubdata_price: Gauge<u64>,
```
slowli, line 14: This metric doesn't seem to be used.

core/node/fee_model/src/l1_gas_price/gas_adjuster/mod.rs
```diff
@@ -124,7 +240,23 @@ impl GasAdjuster {
                 }
             }
             self.blob_base_fee_statistics
-                .add_samples(base_fees.iter().map(|fee| fee.base_fee_per_blob_gas));
+                .add_samples(fee_data.iter().map(|fee| fee.base_fee_per_blob_gas));
+
+            if let Some(current_l2_pubdata_price) = fee_data.last().map(|fee| fee.l2_pubdata_price)
+            {
+                // L2 pubdata price overflows `u64` only in very extreme cases.
+                // It doesn't worth to observe exact value with metric because anyway values that can be used
```
slowli, line 248: Grammar: "doesn't worth" -> "isn't worth".

core/node/fee_model/src/l1_gas_price/gas_adjuster/mod.rs
```diff
@@ -197,36 +329,33 @@ impl GasAdjuster {
...
                     * self.config.internal_pubdata_pricing_multiplier;
 
-                self.bound_blob_base_fee(calculated_price)
-            }
-            PubdataSendingMode::Calldata => {
-                self.estimate_effective_gas_price() * self.pubdata_byte_gas()
+                self.bound_pubdata_fee(calculated_price)
             }
+            PubdataSendingMode::Calldata => self.bound_pubdata_fee(
+                (self.estimate_effective_gas_price() * L1_GAS_PER_PUBDATA_BYTE as u64) as f64,
```
slowli, line 335: Dumb question: Wouldn't it make more sense to convert `estimate_effective_gas_price()` output to `f64` so that there's less chance of integer overflow?

core/node/fee_model/src/l1_gas_price/gas_adjuster/mod.rs
```diff
@@ -197,36 +329,33 @@ impl GasAdjuster {
...
 
-    fn pubdata_byte_gas(&self) -> u64 {
-        match self.commitment_mode {
-            L1BatchCommitmentMode::Validium => 0,
-            L1BatchCommitmentMode::Rollup => L1_GAS_PER_PUBDATA_BYTE.into(),
-        }
-    }
-
-    fn bound_blob_base_fee(&self, blob_base_fee: f64) -> u64 {
+    fn bound_pubdata_fee(&self, pubdata_fee: f64) -> u64 {
```
slowli, line 348: Grammar (?): I understand that it's used in multiple other places, but it seem that the fee is "bounded" (by certain numerical limits) rather than "bound" (past participle of "bind"; i.e. tied / connected to something).

core/node/node_framework/src/implementations/layers/query_eth_client.rs
```diff
@@ -1,39 +1,74 @@
...
+    }
+}
+
+#[derive(Debug)]
+pub struct Output {
+    query_client_l1: Client<L1>,
+    query_client_l2: Option<Client<L2>>,
+}
+
+impl IntoContext for Output {
```
slowli, line 36: Why not declare output as containing resources directly?

```rust
#[derive(Debug, IntoContext)]
#[context(crate = crate)]
pub struct Output {
    query_client_l1: EthInterfaceResource,
    query_client_l2: Option<L2InterfaceResource>,
}
```

This is the approach used by other wiring layers.

core/node/node_framework/src/implementations/resources/eth_interface.rs
```diff
@@ -13,9 +13,23 @@ impl Resource for EthInterfaceResource {
...
+/// It is expected to have the same URL as the `EthInterfaceResource`, but have different capabilities.
+///
+/// This resource is provided separately from `EthInterfaceResource`, to provide type safety in places, where the
+/// component must work with L1-interface only and should use `EthInterfaceResource` instead.
+#[derive(Debug, Clone)]
+pub struct L2InterfaceResource(pub Box<DynClient<L2>>);
+
+impl Resource for L2InterfaceResource {
+    fn name() -> String {
+        "common/l2_inteface".into()
```
slowli, line 26: Typo: "inteface" -> "interface".

core/node/node_framework/src/implementations/resources/eth_interface.rs
```diff
@@ -13,9 +13,23 @@ impl Resource for EthInterfaceResource {
     }
 }
 
+/// A resource that provides L2 interface object to the service.
+/// It is expected to have the same URL as the `EthInterfaceResource`, but have different capabilities.
+///
+/// This resource is provided separately from `EthInterfaceResource`, to provide type safety in places, where the
+/// component must work with L1-interface only and should use `EthInterfaceResource` instead.
+#[derive(Debug, Clone)]
+pub struct L2InterfaceResource(pub Box<DynClient<L2>>);
```
slowli, line 22: Bikeshedding: Maybe, call this `L2SettlementInterfaceResource`, so that the intended use case of this resource is clear? (Or simply `L2SettlementResource`; I find the "interface" part in `EthInterface` / `BoundEthInterface` somewhat non-descriptive.)

core/lib/dal/src/blocks_web3_dal.rs
```diff
@@ -564,17 +565,21 @@ impl BlocksWeb3Dal<'_, '_> {
...
 
-    /// Returns `base_fee_per_gas` for L2 block range [min(newest_block - block_count + 1, 0), newest_block]
+    /// Returns `base_fee_per_gas` and `fair_pubdata_price` for L2 block range [min(newest_block - block_count + 1, 0), newest_block]
     /// in descending order of L2 block numbers.
     pub async fn get_fee_history(
         &mut self,
         newest_block: L2BlockNumber,
         block_count: u64,
-    ) -> DalResult<Vec<U256>> {
+    ) -> DalResult<(Vec<U256>, Vec<U256>)> {
```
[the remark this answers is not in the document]
tomg10: For me it looks like It's easier to use in it's current form

core/lib/web3_decl/src/client/network.rs
tomg10: Imo it's easier to understand what this does now without having to see definition

core/node/node_framework/src/implementations/resources/eth_interface.rs
tomg10: I like better what is here now, so naming based on "what it is" rather "what it's used for", I also really like the "Interface" in the names, it tells me instantly that it's some sort of API resource

tomg10 (commented):
I've left a few small comments

core/lib/eth_client/src/lib.rs
tomg10: It will be also used for Net = L2 with the next PRs

core/node/node_framework/src/implementations/layers/query_eth_client.rs
StanislavBreadless: Makes sense, I think I did not notice that the `Option<X>` is possible there

core/node/fee_model/src/l1_gas_price/gas_adjuster/mod.rs
StanislavBreadless: The values will be converted to integers in the end anyway, so if it goes above u64::max, there will be issues anyway

core/lib/eth_client/src/lib.rs
StanislavBreadless: I thought that we wanted to have additional type safety, but if it is acceptable to merge those two, then okay

core/lib/eth_client/src/lib.rs
StanislavBreadless: > but the workspace still compiles after being narrowed to impl BoundEthInterface<L1> for MockSettlementLayer<L1>

The tests don't. And yes as @tomg10 said we will use it for L2

StanislavBreadless: @slowli the problem right now is we want both `DynClient<L1>`-based and `DynClient<L2>`-based structs to be able to implement this trait. I understand the logic that a struct can not implement at the same time both `BoundEthInterface<L1>` and `BoundEthInterface<L2>`, but in this case `DynClient<L1>` will be a requirement for the struct that wanted to implement `BoundEthInterface<L2>` which is hard to achieve. 

So it is easier to say please provide some `DynClient<Net>` and so we have to make `Net` a template param

core/lib/web3_decl/src/client/network.rs
StanislavBreadless: @slowli Just visually it seemed to me that `ForNetwork` should implement only things that could be used for any network. But right now there appeared to be types that are implemented for all networks there, like 

```
/// Subset of the L1 `eth` namespace used by the L1 client.
#[rpc(client, namespace = "eth", client_bounds(Self: ForEthereumLikeNetwork))]
```

So that's why I decided that rename would be necessary to ensure that people understand this network is expected to be Ethereum-like

core/node/fee_model/src/l1_gas_price/gas_adjuster/metrics.rs
StanislavBreadless: Not sure what would be the best place to use it, I'll delete it for now

core/node/fee_model/src/l1_gas_price/gas_adjuster/mod.rs
StanislavBreadless: I will rename to `cap_pubdata_fee`

core/node/fee_model/src/l1_gas_price/gas_adjuster/mod.rs
StanislavBreadless: It unfortunately makes it harder in some places to use the method:

I have to do the following:

```
(Box::new(gateway.clone().into_client()) as Box::<DynClient::<L1>>).into()
```

which is more convoluted than just doing `from_l1`. I will keep both then

core/lib/eth_client/src/lib.rs
StanislavBreadless: It is not very easy unfortunately, an L2 client also has the same interface as the L1 one, so it is not possible to say that "all clients that only support L1 have this trait implementation, but those that also support L2 have another trait implementation". If keep different traits, but the same return data it starts looking rather weird that we have two similar-looking methods that return the same struct, but apparently with different field values

slowli:
IMO, the PR is still unnecessarily complex. The root problem is in the `L1` marker and `EthInterface` semantics. AFAIU, after refactoring, `EthInterface` and `BoundEthInterface` are intended to actually encapsulate methods in *any* settlement layer (at least this looks logical). In this case, it should be reworked as such: renamed to something like `SettlementClient` and its `base_fee_history` method reworked to return an enum (or encapsulate L1 / L2 distinction in another way, but defining an enum is probably the safest option). Similarly, the marker (previously) used in `EthInterface` wildcard implementation and still used `BoundEthInterface` should really have "settlement layer (either L1 or L2)" semantics; if this is the case, there's no need in introducing a type param for it. This approach would also eliminate `ZkSyncInterface` trait, `GasAdjusterClient` enum and so on.

Maybe, there are issues with this approach? The only one that I'm not seeing rn is the wildcard implementation if `EthInterface` for the `jsonrpsee`-derived client. One way to solve it is having `fee_history` return a more complete `zksync_types::api` type, and handling empty `pubdata_price` array inside the `EthInterface` implementation.

tomg10:
imo GasAdjusterClient is a very good abstraction and I'm strongly against removing it, I really like that it has very simple API

core/node/fee_model/src/l1_gas_price/gas_adjuster/mod.rs
```diff
@@ -18,6 +18,75 @@ mod metrics;
 #[cfg(test)]
 mod tests;
 
+#[derive(Debug)]
+pub enum GasAdjusterClient {
```
slowli, line 22: I need to clarify: when I've said to remove the enum, I've meant simplifying it to

```rust
#[derive(Debug)]
pub struct GasAdjusterClient {
    inner: Box<dyn EthFeeInterface>,
}
```

...which is AFAIU is possible. It has downsides, e.g., it requires double-boxing `Box<DynClient<_>>`, which AFAIU is impossible to avoid because Rust doesn't convert among trait objects, but IMO is overall cleaner. If L1 / L2 tags need to be retained (not 100% sure whether it's possible to refactor assertions to avoid them), they could be stored in an additional field. Alternatively, int may be made a part of `EthFeeInterface` itself (i.e., add `fn network_type(&self) -> L1 | L2`); not sure whether this makes sense semantically. Regarding component tags, I think it's reasonable to have this tagging moved to the caller side.

core/node/node_framework/src/implementations/resources/eth_interface.rs
slowli: Naming is subjective; IMO, `_Client` better signals being an API resource 🤷 Also, why is this convention not followed for `EthInterfaceResource` then? If `GasAdjusterClient` is refactored as per another comment, it could be reworked to wrap `EthFeeInterface`, and an additional L2 client resource wouldn't be necessary IIUC.

core/lib/basic_types/src/settlement.rs
```diff
@@ -0,0 +1,16 @@
...
+/// Gateway is an Ethereum-compatible L2 and so it requires different treatment with regards to DA handling.
+#[derive(Default, Debug, Copy, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub enum SettlementMode {
+    #[default]
+    SettlesToL1,
+    Gateway,
+}
+
+impl SettlementMode {
+    pub fn is_gateway(&self) -> bool {
```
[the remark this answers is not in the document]
StanislavBreadless: @slowli will it offer a performance boost? It just sounds a bit non standard to pass `self` by value where it is not needed. Using reference everywhere seems more consistent and future proof

core/lib/eth_client/src/clients/mock.rs
```diff
@@ -221,8 +223,12 @@ impl MockExecutedTxHandle<'_> {
 type CallHandler =
     dyn Fn(&web3::CallRequest, BlockId) -> Result<ethabi::Token, ClientError> + Send + Sync;
 
+pub trait SupportedMockEthNetwork: Network {
```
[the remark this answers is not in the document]
StanislavBreadless: it is used in this function to avoid code repetition, though it is indeed debatable whether it reduces or increases complexity

```
    pub fn build(self) -> MockSettlementLayer<Net> {
        MockSettlementLayer {
            max_fee_per_gas: self.max_fee_per_gas,
            max_priority_fee_per_gas: self.max_priority_fee_per_gas,
            non_ordering_confirmations: self.non_ordering_confirmations,
            inner: self.inner.clone(),
            client: Net::build_client(self),
        }
    }
```

core/node/fee_model/src/l1_gas_price/gas_adjuster/mod.rs
StanislavBreadless: > Regarding component tags, I think it's reasonable to have this tagging moved to the caller side.

It is just not consistent with the code we have right now and more error-prone (easier to remember to tag in one place instead of N places)

core/lib/basic_types/src/settlement.rs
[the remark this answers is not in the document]
tomg10: I agree with @StanislavBreadless

core/node/fee_model/src/l1_gas_price/gas_adjuster/mod.rs
tomg10: I agree with @StanislavBreadless

slowli: Fine, it can still work with `struct GasAdjusterClient` if we don't do `Box<dyn EthFeeInterface>` dependency injection; in this case, the tag will be set when converting an L1 / L2 client to `Box<dyn EthFeeInterface>`.

core/lib/basic_types/src/settlement.rs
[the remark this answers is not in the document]
slowli: Yes, this is a marginal performance boost, and there's [a pedantic clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#/trivially_copy_pass_by_ref) for it that we don't have enabled.

core/lib/types/src/api/mod.rs
```diff
@@ -834,6 +834,17 @@ pub struct TransactionExecutionInfo {
     pub execution_info: Value,
 }
 
+/// The fee history type returned from `eth_feeHistory` call.
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct FeeHistory {
+    #[serde(flatten)]
+    pub inner: zksync_basic_types::web3::FeeHistory,
+    /// An array of effective pubdata price. Note, that this field is L2-specific and only provided by L2 nodes.
```
slowli, line 843: Grammar: "array of effective pubdata price**s**"?

core/lib/eth_client/src/clients/mock.rs
slowli, line 1: Nit: It makes sense to rename `MockEthereumInner` as well 🙃

slowli (commented):
Overall, looks pretty decent, although I'd still prefer refactoring `GasAdjusterClient` to use `Box<dyn EthFeeInterface>` for more high-level / consistent DI (can be done separately). Please remove unused `EthInterface` imports as pointed by clippy.

core/node/fee_model/src/l1_gas_price/gas_adjuster/mod.rs
StanislavBreadless: Okay, I put the tag into the conversion methods

slowli (approved):

tomg10 (approved):

PR #2583, opened 2024-08-02
feat: Poll the main node API for attestation status - relaxed (BFT-496)

aakoshh:
## What ❔

Reverts https://github.com/matter-labs/zksync-era/pull/2574 to re-establish the attestation status API integration, but also bumps the era-consensus version to [0.1.0-rc.6](https://github.com/matter-labs/era-consensus/pull/174) which has changes to make it backwards compatible:
* the `Executor` is started without waiting for the initial attestation batch number to become available, so as not to stop the node completely from participating in gossip without the main node API (this behaviour is internal to the `AttesterStatusRunner`)
* invalid batch vote signatures are not rejected, just ignored, so the node doesn't break connection with a peer who signed a different payload while the feature is in flux

TODO: 
- [x] Update once `0.1.0-rc.6` is published to crates.io

## Why ❔

The first reason is to not be so eager to drop a gossip peer for incompatibilities between newer and existing versions of the software while we're actively working on the features involved.

The second is that the rollout strategy to mainnet seems to be to roll out external nodes first, main nodes second. It is expected that the external node should work with an older version of the main node, or at least not fail to start; it is okay if a new feature doesn't work.

## Checklist


pompon0:
@aakoshh please bump to 0.1.0-rc.7

core/node/consensus/src/en.rs
```diff
@@ -99,6 +105,18 @@ impl EN {
...
+                AttestationStatusRunner::init(
+                    ctx,
+                    Box::new(MainNodeAttestationStatus(self.client.clone())),
+                    time::Duration::seconds(5),
+                    genesis_hash,
+                )
+                .await
+                .wrap("AttestationStatusRunner::init()")?
+            };
+            s.spawn_bg(async { Ok(runner.run(ctx).await?) });
```
pompon0, line 118: wrap() ?

aakoshh: IIRC all these runners return anyhow::Result

core/node/consensus/src/en.rs
```diff
@@ -239,3 +259,31 @@ impl EN {
...
+                let status: consensus_dal::AttestationStatus =
+                    zksync_protobuf::serde::deserialize(&status.0)
+                        .context("deserialize(AttestationStatus")?;
+
+                Ok(Some((status.genesis, status.next_batch_to_attest)))
+            }
+            Ok(None) => Ok(None),
+            Err(err) => {
+                tracing::warn!("AttestationStatus call to main node HTTP RPC failed: {err}");
+                Ok(None)
```
pompon0, line 285: you can just return the error and it will be logged outside, right?

aakoshh: Yes, it would do the same thing. This just followed the error log pattern in another similar API client wrapper in one of the files, can’t remember which

pompon0 (dismissed):

core/node/consensus/src/en.rs
pompon0: then context

core/node/consensus/src/en.rs
pompon0: I suppose it would be clearer if the Client API didn't return Result<Option<Status>> but rather Result<Status>. Anyway, it doesn't matter much.

core/node/consensus/src/en.rs
aakoshh: I know, wanted to use it but in a lot of cases there was nothing, I thought it was deliberate

pompon0: if it is missing, then someone (me) was too lazy to add it, or there was enough context already (for example, in simple wrappers/helper functions)

pompon0 (approved):

core/node/consensus/src/en.rs
aakoshh: Maybe, it looks pretty consistent ([1](https://github.com/matter-labs/zksync-era/blob/dbd481e425b6c87d1e03c732e3db43a8541d1ac5/core/node/consensus/src/en.rs#L51), [2](https://github.com/matter-labs/zksync-era/blob/dbd481e425b6c87d1e03c732e3db43a8541d1ac5/core/node/consensus/src/en.rs#L96), [3](https://github.com/matter-labs/zksync-era/blob/dbd481e425b6c87d1e03c732e3db43a8541d1ac5/core/node/consensus/src/en.rs#L101), [4](https://github.com/matter-labs/zksync-era/blob/dbd481e425b6c87d1e03c732e3db43a8541d1ac5/core/node/consensus/src/en.rs#L106), [5](https://github.com/matter-labs/zksync-era/blob/dbd481e425b6c87d1e03c732e3db43a8541d1ac5/core/node/consensus/src/mn.rs#L46), [6](https://github.com/matter-labs/zksync-era/blob/dbd481e425b6c87d1e03c732e3db43a8541d1ac5/core/node/consensus/src/mn.rs#L51), [7](https://github.com/matter-labs/zksync-era/blob/dbd481e425b6c87d1e03c732e3db43a8541d1ac5/core/node/consensus/src/mn.rs#L62), [8](https://github.com/matter-labs/zksync-era/blob/dbd481e425b6c87d1e03c732e3db43a8541d1ac5/core/node/consensus/src/mn.rs#L74)). 

The only reason this runner should fail is if the genesis returned from the API changed, which has a good description.

pompon0: yes, consistently lazy.

PR #2588, opened 2024-08-04
fix(core): Handle GCS Response retriable errors

EmilLuta:
Response errors that GCS considers retriable are treated as fatal in our codebase. This causes the code to crash, when it could safely retry and continue it's execution.

This PR treats GCS's response errors that are retriable as retriable. Alongside, `is_transient` has been renamed to `is_retriable` to better reflect it's usage.

The change will reduce the amount of job crashes and peak execution times seen across prover subsystems (especially in BWGs).

### DIFF
--- a/core/bin/zksync_tee_prover/src/error.rs
+++ b/core/bin/zksync_tee_prover/src/error.rs
@@ -11,19 +11,19 @@ pub(crate) enum TeeProverError {
 }
 
 impl TeeProverError {
-    pub fn is_transient(&self) -> bool {
+    pub fn is_retriable(&self) -> bool {
         match self {
-            Self::Request(err) => is_transient_http_error(err),
+            Self::Request(err) => is_retriable_http_error(err),
             _ => false,
         }
     }
 }
 
-fn is_transient_http_error(err: &reqwest::Error) -> bool {
+fn is_retriable_http_error(err: &reqwest::Error) -> bool {
     err.is_timeout()
             || err.is_connect()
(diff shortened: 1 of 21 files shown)


core/lib/object_store/src/gcs.rs
```diff
@@ -85,22 +85,22 @@ impl GoogleCloudStore {
...
         Self::Initialization {
             source: err.into(),
-            is_transient,
+            is_retriable,
         }
     }
 }
 
-fn is_transient_http_error(err: &reqwest::Error) -> bool {
+fn is_retriable_http_error(err: &reqwest::Error) -> bool {
```
popzxc, line 99: `is_transient` is an established terminology used in multiple components across the core. `is_retriable` isn't currently used anywhere.

Could you please revert renaming changes, as they decrease the consistency?

popzxc (commented):
Change LGTM, but I would avoid changing the established terminology.
If we want to go this way, I'd expect it to be done consistently across the codebase (though not sure if it's worth it).

core/lib/object_store/src/gcs.rs
popzxc: Not that big of an issue, so no opinion except for consistency concerns. If you want to rename it, please rename all the entries.

popzxc (approved):
After merge please announce in the chat that you changed `transient` to `retriable` across the codebase to be consistent with GCP terminology and ask everyone to use the new term from now on.

PR #2589, opened 2024-08-05
feat: update base token rate on L1

ischasny:
Currently base token to ETH ratio is always locked to 1:1 on the L1. This PR changes that and adds functionality to update L1 numerator and denominator as soon as they get updated on the L2. Updates are done by calling `IChainAdmin.setTokenMultiplier` function. The function can only be called by a new role - `tokenMultiplierSetter`.

This PR introduces the following changes:
* new wallet for `tokenMultiplierSetter`, provisioned with ETH on network initialisation;
* updates to old and new tooling to fully support this new role;
* updates to `base_token_ratio_persister` to do the L1 updates;

What is not included:
* Conditional updates - to save gas cost if numerator / denominator hasn't changed since the last time;
* Support for fluctuating numerator / denominator in EN - that will be added in a separate PR.

[PR](https://github.com/matter-labs/era-contracts/pull/683/files) to update tolling on the `era-contracts` side.


core/lib/config/src/configs/base_token_adjuster.rs
```diff
@@ -41,7 +78,40 @@ impl BaseTokenAdjusterConfig {
...
+        )
+    }
+
+    fn default_max_tx_gas() -> u64 {
+        DEFAULT_MAX_TX_GAS
+    }
+
+    #[deprecated]
+    pub fn private_key(&self) -> anyhow::Result<Option<K256PrivateKey>> {
+        std::env::var("TOKEN_MULTIPLIER_SETTER_PRIVATE_KEY")
```
Deniallugo, line 107: the new function and deprecated? it's a bit of a weird

core/lib/config/src/configs/external_price_api_client.rs
```diff
@@ -14,6 +14,11 @@ pub struct ExternalPriceApiClientConfig {
     /// Forced conversion ratio. Only used with the ForcedPriceClient.
     pub forced_numerator: Option<u64>,
     pub forced_denominator: Option<u64>,
+    /// Forced fluctuation is only used in  ForcedPriceClient. It defines how much percent numerator /
+    /// denominator should fluctuate from their forced values. If it's None or 0, then ForcedPriceClient
+    /// will return the same quote every time it's called. Otherwise, ForcedPriceClient will return
+    /// forced_quote +/- forced_fluctuation % from its values.
+    pub forced_fluctuation: Option<u32>,
```
Deniallugo, line 21: should we have a special config for forcedPriceClient? it's a bit of confusing now

core/lib/env_config/src/external_price_api_client.rs
```diff
@@ -27,6 +27,7 @@ mod tests {
             client_timeout_ms: DEFAULT_TIMEOUT_MS,
             forced_numerator: Some(100),
             forced_denominator: Some(1),
+            forced_fluctuation: None,
```
Deniallugo, line 30: why it's not part of the test?

core/node/base_token_adjuster/src/base_token_ratio_persister.rs
```diff
@@ -112,4 +137,95 @@ impl BaseTokenRatioPersister {
...
+                options,
+            )
+            .await
+            .context("cannot sign a `setTokenMultiplier` transaction")?;
+
+        let hash = (*self.eth_client)
+            .as_ref()
+            .send_raw_tx(signed_tx.raw_tx)
+            .await
+            .context("failed sending `setTokenMultiplier` transaction")?;
```
Deniallugo, line 202: are you going to resend tx, if it stuck?

core/lib/config/src/configs/base_token_adjuster.rs
ischasny: Removed - good spot. Copy / paste from somewhere.

core/lib/config/src/configs/external_price_api_client.rs
ischasny: Moved to a separate struct

core/lib/env_config/src/external_price_api_client.rs
ischasny: Good catch - added to the test

core/node/base_token_adjuster/src/base_token_ratio_persister.rs
ischasny: Good point - added configurable retries

core/node/base_token_adjuster/src/base_token_ratio_persister.rs
```diff
@@ -83,7 +83,23 @@ impl BaseTokenRatioPersister {
...
+                    "Failed to update base token multiplier on L1, attempt {}: {}",
+                    attempt + 1,
+                    err
+                );
+                tokio::time::sleep(sleep_duration).await;
+            } else {
+                return result;
+            }
+        }
+        result
```
Deniallugo, line 102: We have to figure out the consequences if this will happen at the wrong time or not happen at all. For eth txs, we have a eth sender, that makes sure, we send everything properly. In this implementation, you won't resend anything, because you are not updating the gas prices

ischasny: Added mechanism similar to the one that we use in EthSender, where `time_in_mempool` is equal to the passed number of L1 blocks. Also increased default wait time between attempts to 30s.

ischasny: CC @tomg10

core/node/base_token_adjuster/src/base_token_ratio_persister.rs
```diff
@@ -112,4 +170,99 @@ impl BaseTokenRatioPersister {
...
+
+        let gas_price = (*self.eth_client)
+            .as_ref()
+            .get_gas_price()
+            .await
+            .with_context(|| "failed getting gas price")?
+            .as_u64()
+            * 2;
+
+        let base_fee_per_gas = self.gas_adjuster.as_ref().get_base_fee(time_in_mempool);
```
tomg10, line 213: after seeing the whole code, you may actually not want to use get_base_fee, depending on settings it may take 10+ minutes for suggested gas price to get high enough (this method is optimized so that as little gas as possible is spent, but not for transaction inclusion speed).
Tx param provider does have a method that is optimized for transaction fast inclusion:
get_blob_tx_base_fee
Please use it instead. The name is unfortunate, you can safely make "a copy" of this method with better name and add it to TxParamsProvider trait, like get_tx_base_fee_for_fast_inclusion.
Second problem is that ethereum doesn't like resending transactions without bumping prices, see how we have to bump prices on resend in eth-sender https://github.com/matter-labs/zksync-era/blob/main/core/node/eth_sender/src/eth_fees_oracle.rs#L81

ischasny: Gotcha - all makes sense, thank you. Pushed in the changes.

core/node/base_token_adjuster/src/base_token_ratio_persister.rs
```diff
@@ -54,11 +82,86 @@ impl BaseTokenRatioPersister {
...
+            );
+        }
+
+        let gas_price = (*self.eth_client)
+            .as_ref()
+            .get_gas_price()
+            .await
+            .with_context(|| "failed getting gas price")?
+            .as_u64()
+            * 2;
```
perekopskiy, line 162: Why do you need gas price? It seems you use eip1559 txs so it's useless

core/node/base_token_adjuster/src/base_token_ratio_persister.rs
```diff
@@ -112,4 +215,90 @@ impl BaseTokenRatioPersister {
...
+            .send_raw_tx(signed_tx.raw_tx)
+            .await
+            .context("failed sending `setTokenMultiplier` transaction")?;
+
+        let max_attempts = self.config.l1_receipt_checking_max_attempts();
+        let sleep_duration = self.config.l1_receipt_checking_sleep_duration();
+        for _i in 0..max_attempts {
+            let maybe_receipt = (*self.eth_client)
+                .as_ref()
+                .tx_receipt(hash)
```
perekopskiy, line 282: You only wait for receipt and 0 confirmations on L1. I'm wondering how critical reorg is for setting ratio?

perekopskiy (commented):
What's the plan if max_attempts is reached for sending l1 tx?

core/node/base_token_adjuster/src/base_token_ratio_persister.rs
ischasny: Good point - removed

core/node/base_token_adjuster/src/base_token_ratio_persister.rs
ischasny: Reorg won't be critical. Quotes get updated on a regular basis so it's going to be set on the next cycle.

core/lib/config/src/configs/base_token_adjuster.rs
```diff
@@ -1,13 +1,37 @@
...
 pub const DEFAULT_INTERVAL_MS: u64 = 30_000;
 
 /// By default, refetch ratio from db every 0.5 second
 pub const DEFAULT_CACHE_UPDATE_INTERVAL: u64 = 500;
 
+/// Default max amount of gas that a L1 base token update can consume per transaction
+pub const DEFAULT_MAX_TX_GAS: u64 = 80_000;
+
+/// Default max amount of gas that a L1 base token update can consume per transaction
+pub const DEFAULT_PRIORITY_FEE_PER_GAS: u64 = 1_000_000_000;
```
perekopskiy, line 18: nit: incorrect comment

ischasny:
> What's the plan if max_attempts is reached for sending l1 tx?

Good point - currently the server is going to halt, which shouldn't be the case. Updated the logic so that it's going to be just retried on the next iteration.

core/lib/protobuf_config/src/base_token_adjuster.rs
```diff
@@ -11,17 +11,34 @@ impl ProtoRepr for proto::BaseTokenAdjuster {
...
             price_cache_update_interval_ms: self
                 .price_cache_update_interval_ms
                 .expect("price_cache_update_interval_ms"),
+            max_tx_gas: self.max_tx_gas.expect("max_tx_gas"),
+            default_priority_fee_per_gas: self
+                .default_priority_fee_per_gas
+                .expect("default_priority_fee_per_gas"),
+            max_acceptable_priority_fee_in_gwei: self
+                .max_acceptable_priority_fee_in_gwei
+                .expect("max_acceptable_priority_fee_in_gwei"),
```
perekopskiy, line 23: you should use defaults instead of unwrapping

core/tests/ts-integration/tests/base-token.test.ts
```diff
@@ -26,6 +26,24 @@ describe('base ERC20 contract checks', () => {
...
     });
 
+    test('Base token ratio is updated on L1', async () => {
+        if (isETHBasedChain) {
+            return;
+        }
+
+        let govMnemonic = ethers.Mnemonic.fromPhrase(require('../../../../etc/test_config/constant/eth.json').mnemonic);
+        let govWalletHD = ethers.HDNodeWallet.fromMnemonic(govMnemonic, "m/44'/60'/0'/0/1");
+        const adminGovWallet = new ethers.Wallet(govWalletHD.privateKey, alice._providerL1());
```
perekopskiy, line 36: it seems you don't need `adminGovWallet` in this test, `alice` is enough

ischasny: Good point - changed

core/lib/config/src/configs/base_token_adjuster.rs
ischasny: fixed

core/tests/ts-integration/tests/base-token.test.ts
ischasny: Actually alice doesn't seem to have enough permissions to call these functions, reverted back

perekopskiy: why? AFAICT It's view getter so it doesn't matter which wallet to connect, you can even connect none

ischasny: Yes, you are right - I've been passing in a wrong wallet. Changed to use `alice`

core/node/base_token_adjuster/src/base_token_ratio_persister.rs
```diff
@@ -54,11 +88,75 @@ impl BaseTokenRatioPersister {
...
+        result
+    }
+
+    fn get_eth_fees(
+        &self,
+        prev_base_fee_per_gas: Option<u64>,
+        prev_priority_fee_per_gas: Option<u64>,
+    ) -> (u64, u64) {
+        // Use get_blob_tx_base_fee here instead of get_base_fee to optimise for fast inclusion.
+        // get_base_fee will cause the transaction to be stuck in the mempool for 10+ minutes.
```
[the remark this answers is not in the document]
ischasny: fixed

etc/env/base/external_price_api.toml
```diff
@@ -3,6 +3,8 @@
 [external_price_api_client]
 
 # What source to use for the external price API. Currently only options are "forced", "no-op", and "coingecko".
-source = "no-op"
+source = "forced"
 
-client_timeout_ms = 10000
+[external_price_api_client.forced]
+numerator = 3
+denominator = 2
```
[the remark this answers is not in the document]
ischasny: Done and done

perekopskiy (approved):

PR #2591, opened 2024-08-05
test(vm): Improve VM benchmarks

slowli:
## What ❔

- Extends the multi-transaction benchmark to cover simple deployments, transfers and various load test transactions. Optionally includes the snapshot workflow into the benchmark.
- Fixes the multi-transaction benchmark setup so that transactions in it don't fail early in bootloader.

## Why ❔

- Transactions failing early leads to non-representative benchmark results.

## Checklist


core/tests/vm-benchmark/benches/criterion.rs
```diff
@@ -1,20 +1,99 @@
...
+        group.bench_function(bench_name, |bencher| {
+            if FULL {
+                // Include VM initialization / drop into the measured time
+                bencher.iter(|| BenchmarkingVm::<VM>::default().run_transaction(black_box(&tx)));
+            } else {
+                bencher.iter_batched(
+                    BenchmarkingVm::<VM>::default,
+                    |mut vm| {
+                        let result = vm.run_transaction(black_box(&tx));
+                        // FIXME: transactions mostly run out of gas here; is this OK?
```
slowli, line 41: Subj. All transactions other than `deploy_simple_contract` fail with an out-of-gas error. Is this expected / OK?

joonazan: Yes. This way they can be tested with different gas limits to figure out if the time increases roughly linearly as it should.
Also, some of the most degenerate edge cases cannot be stopped any other way.

core/tests/vm-benchmark/harness/src/lib.rs
```diff
@@ -96,33 +181,62 @@ impl BenchmarkingVm {
...
+}
 
+impl<VM: BenchmarkingVmFactory> BenchmarkingVm<VM> {
     pub fn run_transaction(&mut self, tx: &Transaction) -> VmExecutionResultAndLogs {
         self.0.push_transaction(tx.clone());
         self.0.execute(VmExecutionMode::OneTx)
     }
 
+    pub fn run_transaction_full(&mut self, tx: &Transaction) -> VmExecutionResultAndLogs {
+        self.0.pop_snapshot_no_rollback();
```
joonazan, line 196: I'm not a fan of pop, then snapshot like I already stated in the sequencer. I think it is more sensible to snapshot, then pop at the end so there is no unnecessary pop on the first execution.

core/tests/vm-benchmark/harness/src/lib.rs
```diff
@@ -96,33 +181,62 @@ impl BenchmarkingVm {
...
+            self.0.rollback_to_the_latest_snapshot();
+        }
+        tx_result
+    }
+
     pub fn instruction_count(&mut self, tx: &Transaction) -> usize {
         self.0.push_transaction(tx.clone());
-
         let count = Rc::new(RefCell::new(0));
+        self.0.inspect(Default::default(), VmExecutionMode::OneTx); // FIXME
```
joonazan, line 214: Using yesterday's tracer improvements, this can be written as
```
struct InstructionCounter(usize);
impl<T> Tracer<T> for InstructionCounter {
    fn before_instruction(&mut self, ...) { self.0 += 1}
}
```

slowli: This is "JFYI" at this point, right? Sure, it'll be possible to fix instruction counting once tracers are merged.

joonazan: Yes.

joonazan (dismissed):

joonazan (approved):

artmakh (approved):

PR #2595, opened 2024-08-06
feat: Remove old EN code

popzxc:
## What ❔

- Removes old EN code.
- Some configs weren't used in the framework code by mistake. This is fixed now.
- Diamond proxy address was made private in config so that it's obtained in a checked manner.

Some changes are not done in this PR to reduce the scope (e.g. removing unneeded dependencies).

## Why ❔

Switch to framework.

## Checklist


core/node/node_framework/src/implementations/layers/node_storage_init/external_node_strategy.rs
```diff
@@ -73,14 +74,21 @@ impl WiringLayer for ExternalNodeInitStrategyLayer {
...
+                let recovery_pool = input
+                    .master_pool
+                    .get_custom(self.max_postgres_concurrency.get() as u32)
+                    .await?;
+                Some(Arc::new(ExternalNodeSnapshotRecovery {
+                    client: client.clone(),
+                    pool: recovery_pool,
+                    recovery_config,
+                    app_health,
+                }) as Arc<dyn InitializeStorage>)
```
slowli, line 88: Bikeshedding: I'd extract `Arc::new(_)` to a local var since this doesn't look very readable.

core/node/node_framework/src/service/mod.rs
```diff
@@ -59,10 +59,19 @@ impl ZkStackServiceBuilder {
...
-        Ok(Self {
+        Ok(Self::on_runtime(runtime))
+    }
+
+    /// Creates a new builder with the provided Tokio runtime.
+    /// This method can be used if asynchronous tasks must be performed before the service is built.
+    ///
+    /// However, it is not recommended to use this method to spawn any tasks that will not be managed
+    /// by the service itself, so whenever it can be avoided, using [`ZkStackServiceBuilder::new`] is preferred.
+    pub fn on_runtime(runtime: Runtime) -> Self {
```
slowli, line 70: Dumb question: Will this method be used in the future? AFAIU, it's unused now, other than the call above. Should it be used inside `ExternalNodeBuilder::new()`?

core/bin/external_node/src/config/mod.rs
```diff
@@ -1347,6 +1348,29 @@ impl ExternalNodeConfig {
...
+    /// Otherwise, the remote value will be used. However, using remote value has trust implications for the main
+    /// node so relying on it solely is not recommended.
+    pub fn diamond_proxy_address(&self) -> anyhow::Result<Address> {
+        let remote_diamond_proxy_addr = self.remote.diamond_proxy_addr;
+        let diamond_proxy_addr = if let Some(addr) = self.optional.contracts_diamond_proxy_addr {
+            anyhow::ensure!(
+                addr == remote_diamond_proxy_addr,
+                "Diamond proxy address {addr:?} specified in config doesn't match one returned \
+                by main node ({remote_diamond_proxy_addr:?})"
+            );
```
slowli, line 1363: IMO, it could make sense to run this check in `ExternalNodeConfig::fetch_remote()` since this getter may be called multiple times.

slowli (commented):
Can't say I have looked that deep into whether framework-based initialization works identical to the legacy approach, but overall it looks good.

core/node/node_framework/src/service/mod.rs
popzxc: Oh, that's a great catch. Yes, it should be called there.

perekopskiy (dismissed):

core/bin/external_node/src/config/mod.rs
```diff
@@ -1341,6 +1350,24 @@ impl ExternalNodeConfig {
             tree_component: TreeComponentConfig { api_port: None },
         }
     }
+
+    /// Returns a verified diamond proxy address.
+    /// If local configuration contains the address, it will be checked against the one returned by the main node.
+    /// Otherwise, the remote value will be used. However, using remote value has trust implications for the main
+    /// node so relying on it solely is not recommended.
+    pub fn diamond_proxy_address(&self) -> anyhow::Result<Address> {
```
slowli, line 1358: There's no need to return a `Result` now; the method always returns `Ok(_)`.

core/bin/external_node/src/config/mod.rs
```diff
@@ -1341,6 +1350,24 @@ impl ExternalNodeConfig {
...
+        let remote_diamond_proxy_addr = self.remote.diamond_proxy_addr;
+        let diamond_proxy_addr = if let Some(addr) = self.optional.contracts_diamond_proxy_addr {
+            addr
+        } else {
+            tracing::info!(
```
slowli, line 1363: Nit: I would move this log to the `anyhow::ensure!` check above so that it's too not spammed on multiple calls. Then, the method can be simplified to

```rust
self.optional
    .contracts_diamond_proxy_addr
    .unwrap_or(self.remote.diamond_proxy_addr)
```

slowli (dismissed):

slowli (dismissed):

slowli (approved):

PR #2599, opened 2024-08-06
refactor(state-keeper): Make batch executor and storage factory parametric

slowli:
## What ❔

Makes `BatchExecutor` and `ReadStorageFactory` parametric by the storage type. Encapsulates this storage type using a private helper trait (essentially an async closure) used by the state keeper.

## Why ❔

Allows to avoid crutches with the mock storage factory. Potentially extends customizability of batch execution.

## Checklist


core/lib/state/src/storage_factory.rs
```diff
@@ -11,19 +11,20 @@ use crate::{
...
 };
 
-/// Factory that can produce [`OwnedStorage`] instances on demand.
+/// Factory that can produce storage instances on demand. The storage type is encapsulated as a type param
+/// (mostly for testing purposes); the default is [`OwnedStorage`].
 #[async_trait]
-pub trait ReadStorageFactory: Debug + Send + Sync + 'static {
-    /// Creates an [`OwnedStorage`] entity over either a Postgres connection or RocksDB
-    /// instance. The specific criteria on which one are left up to the implementation.
+pub trait ReadStorageFactory<S = OwnedStorage>: Debug + Send + Sync + 'static {
```
slowli, line 17: I'm not sure whether this should be a type param or an associated type. Approach with the type param allows to define a default (i.e., less code changes), so I've chosen it.

popzxc: Looks fine to me.

core/node/state_keeper/src/keeper.rs
```diff
@@ -48,6 +50,44 @@ impl Error {
     }
 }
 
+/// Functionality [`BatchExecutor`] + [`ReadStorageFactory`] with an erased storage type.
```
popzxc, line 53: Do we need this? From its looks, state keeper can just store the batch executor and the storage factory, and `init_batch` can be a method there. One less abstraction to worry about.

core/node/state_keeper/src/keeper.rs
```diff
@@ -48,6 +50,44 @@ impl Error {
...
+        l1_batch_env: L1BatchEnv,
+        system_env: SystemEnv,
+        stop_receiver: &watch::Receiver<bool>,
+    ) -> Result<BatchExecutorHandle, Error>;
+}
+
+/// The only [`ErasedBatchExecutor`] implementation.
+#[derive(Debug)]
+struct ErasedBatchExecutorImpl<S> {
+    batch_executor_base: Box<dyn BatchExecutor<S>>,
```
popzxc, line 67: Bikeshedding: Since `BatchExecutor` itself is no longer a factory, I guess we no longer need the `_base` postfix.

core/node/state_keeper/src/keeper.rs
popzxc: Sorry if this is a stupid question, but can you elaborate? Unless I'm missing something, this is a private trait that only has a single implementation, it's only instantiated within `ZkSyncStateKeeper::new` where we already provide both `dyn BatchExecutor<S>` and `dyn ReadStorageFactory<S>`. From its looks, I was under the impression that you can simply store `Box<dyn BatchExecutor<S>>` and `Arc<dyn ReadStorageFactory<S>>` as fields in `ZkSyncStateKeeper` and implement `init_batch` as a method there, without changing any public interfaces.

Or do you imply that there is some follow-up work planned with this trait in mind?

slowli: > you can simply store `Box<dyn BatchExecutor<S>>` and `Arc<dyn ReadStorageFactory<S>>` as fields in `ZkSyncStateKeeper `

You can, but then `ZkSyncStateKeeper` should be parametric by `S` (at least I don't see how it could avoid it). It can have the `S = OwnedStorage` default, but `S == ()` would need to be used in tests, which can complicate test code somewhat and would be a problem if `ZkSyncStateKeeper` is used in DI (rn, it's not injected directly). These are not major concerns, so if you insist, I'll rework it.

popzxc: I see, that makes sense. I don't really like a new abstraction that is used purely for the generic erasure (StateKeeper already has plenty, so the cognitive cost of any new one is pretty high), but it's not a hill I'm going to die on.

Just in case, won't it work if `access_storage` would return `Box<dyn ReadStorage>`, so that the type is erased there?

If not, up to you whether to keep it or not. But if you decide to keep it, maybe it would be beneficial to mention that we need this type/trait to prevent `StateKeeper` from being generic (at least for me, it wasn't initially clear).

popzxc (dismissed):

core/node/state_keeper/src/keeper.rs
slowli: > Just in case, won't it work if `access_storage` would return `Box<dyn ReadStorage>`, so that the type is erased there?

We would need a new trait for the returned value because `OwnedStorage` doesn't implement `ReadStorage` and AFAIU cannot easily be refactored to do that. The issue is that Postgres-based storage is non-`'static`, avoiding which seemingly requires introducing a self-referential struct (connection + pool where the connection is borrowed from). If you recall, I've implemented something like that for tests using Postgres; but I'm not 100% sure that my implementation was sound 🙃 Such a struct could be useful for other applications as well; I will probably sketch it in a separate PR.

Besides this, there's also performance concern: with `Box<dyn ReadStorage>`, we'd have dynamic dispatch for all storage operations, so IMO it's worth benchmarking this sufficiently (which means tackling the problem separately).

popzxc (approved):

popzxc (approved):

PR #2602, opened 2024-08-06
feat(tee_verifier_input_producer): more flexible config

pbeza:
## What ❔

Improve the flexibility of the `tee_verifier_input_producer` configuration at runtime.

## Why ❔

The `tee_verifier_input_producer` will become more complex over time, so we need to make some of its parameters flexible at runtime.

## Checklist


pbeza:
Looks like we don't need this PR for now. I will reopen this once we have more runtime parameters.

PR #2609, opened 2024-08-07
fix: shrink enclave size of `zksync-tee-prover`

haraldh:
## What ❔

Shrink the enclave size of `zksync-tee-prover`.

## Why ❔

32G was a bit of safe guard while testing. 8G seems to be enough and will fit better the memory size of the Azure nodes.

## Checklist


popzxc (approved):

thomasknauth (commented):
LGTM.

PR #2616, opened 2024-08-07
feat(vm): Allow running new VM in state keeper

slowli:
## What ❔

- Allows running the new VM in the state keeper.
- Adds a CI load test with the new VM.

## Why ❔

Allows to monitor new VM performance regressions.

## Checklist


slowli (commented):
It is possible to not run the VM playground in integration tests now and instead run the state keeper with the shadow VM mode. I think the VM playground has its benefits; it could to complete integration tests and then fail if the playground has crashed (but this would optimally require to run it in a separate process which I haven't implemented). Would this approach make sense, or is it overly complicated?

RomanBrodetski (approved):

PR #2627, opened 2024-08-08
feat: remove TEE from tree

thomasknauth:
TEE input producer jobs are now created by monitoring the db instead of hooking directly into the tree logic.

## Checklist

### DIFF
--- a/core/lib/dal/src/tee_verifier_input_producer_dal.rs
+++ b/core/lib/dal/src/tee_verifier_input_producer_dal.rs
@@ -89,8 +89,12 @@ impl TeeVerifierInputProducerDal<'_, '_> {
         .report_latency()
         .fetch_one(self.storage)
         .await?
-        .number
-        .unwrap();
+        .number;
+
+        let row_high = match row_high {
+            Some(number) => number,
+            None => return Ok(None),
+        };
 
         let row_low = sqlx::query!(
             r#"
--- a/core/lib/dal/src/tee_verifier_input_producer_dal.rs
+++ b/core/lib/dal/src/tee_verifier_input_producer_dal.rs
@@ -72,17 +72,16 @@ impl TeeVerifierInputProducerDal<'_, '_> {
         Ok(())
     }
 
-    // Returns inclusive range, i.e. [low,high]
-    pub async fn get_new_l1_batches(
-        &mut self,
-    ) -> DalResult<Option<(L1BatchNumber, L1BatchNumber)>> {
+    pub async fn get_new_l1_batch(&mut self) -> DalResult<Option<L1BatchNumber>> {
         // Since we depend on Merkle paths, we use the proof_generation_details table to inform us of newly available to-be-proven batches.
-        let row_high = sqlx::query!(
+        let row = sqlx::query!(
             r#"
             SELECT
                 MAX(l1_batch_number) AS "number"
             FROM
                 proof_generation_details
+            WHERE
+                proof_gen_data_blob_url IS NOT NULL
             "#
         )
         .instrument("get_sealed_l1_batch_number")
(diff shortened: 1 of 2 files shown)


core/node/metadata_calculator/src/updater.rs
```diff
@@ -152,10 +152,6 @@ impl TreeUpdater {
             // right away without having to implement dedicated code.
 
             if let Some(object_key) = &object_key {
-                storage
-                    .tee_verifier_input_producer_dal()
-                    .create_tee_verifier_input_producer_job(l1_batch_number)
-                    .await?;
                 // Save the proof generation details to Postgres
                 storage
```
pbeza, line 156: I know it's not your code, but do you happen to know why we have two separate calls here? Why not delete the `.save_merkle_paths_artifacts_metadata()` call and save the Merkle paths within the `.insert_proof_generation_details(l1_batch_number, object_key)` call instead? Just curious if there's a good reason for it. 🤔

core/lib/dal/src/tee_verifier_input_producer_dal.rs
```diff
@@ -72,9 +72,57 @@ impl TeeVerifierInputProducerDal<'_, '_> {
...
+        .fetch_one(self.storage)
+        .await?;
+
+        // Since we depend on Merkle paths, we use the proof_generation_details table to inform us of newly available to-be-proven batches.
+        let row_high = sqlx::query!(
+            r#"
+            SELECT
+                MAX(l1_batch_number) AS "number"
+            FROM
+                proof_generation_details
```
pbeza, line 99: I'm not sure why `proof_gen_data_blob_url` is allowed to be `NULL` in the database, but in this case, shouldn't we ensure that Merkle paths are indeed available?
```suggestion
            SELECT
                MAX(l1_batch_number) AS "number"
            FROM
                proof_generation_details
            WHERE
                proof_gen_data_blob_url IS NOT NULL
```

core/node/metadata_calculator/src/updater.rs
pbeza: Ah, it looks like @Artemka374 fixed a race condition [here](https://github.com/matter-labs/zksync-era/pull/2405). Honestly, I don't fully understand it just by reading the code, but it seems necessary.

core/lib/dal/src/tee_verifier_input_producer_dal.rs
pbeza: Just found that `proof_gen_data_blob_url` used to be non-`NULL`. The constraint was dropped recently here: https://github.com/matter-labs/zksync-era/pull/2405/files#diff-bf5de8ff2822841d4fd0dc5850c7a17039d594382e790b1c877b0949337fde37

core/lib/dal/src/tee_verifier_input_producer_dal.rs
```diff
@@ -72,9 +72,58 @@ impl TeeVerifierInputProducerDal<'_, '_> {
...
+
     pub async fn get_next_tee_verifier_input_producer_job(
         &mut self,
     ) -> DalResult<Option<L1BatchNumber>> {
+        if let Some((low, high)) = self.get_new_l1_batches().await? {
+            for i in low.0..=high.0 {
+                self.create_tee_verifier_input_producer_job(L1BatchNumber(i))
+                    .await?
+            }
+        }
```
RomanBrodetski, line 125: why do we return a range here? We invoke `get_next_tee_verifier_input_producer_job` in a loop - and then inside it we invoke `create_tee_verifier_input_producer_job` in a loop. Why not just have a single loop? Make `get_new_l1_batches` return just one batch and create one job at a time

core/lib/dal/src/tee_verifier_input_producer_dal.rs
thomasknauth: Yeah, makes sense.

core/lib/dal/src/tee_verifier_input_producer_dal.rs
```diff
@@ -72,9 +72,63 @@ impl TeeVerifierInputProducerDal<'_, '_> {
         Ok(())
     }
 
+    pub async fn get_new_l1_batch(&mut self) -> DalResult<Option<L1BatchNumber>> {
+        // Since we depend on Merkle paths, we use the proof_generation_details table to inform us of newly available to-be-proven batches.
+        let row = sqlx::query!(
+            r#"
+            SELECT
```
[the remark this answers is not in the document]
thomasknauth: That should not happen. You only return `max(l1_batch_number)` from the `proof_generation_details` table if the TEE jobs table is empty (which just handles the special case when the thing comes up for the first time, L114). Normal case is to return `max(l1_batch_number)`+1 from the TEE jobs table.

core/lib/dal/src/tee_verifier_input_producer_dal.rs
```diff
@@ -72,9 +72,63 @@ impl TeeVerifierInputProducerDal<'_, '_> {
...
+                MAX(l1_batch_number) AS "number"
+            FROM
+                proof_generation_details
```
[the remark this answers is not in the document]
thomasknauth: What code do you intend to duplicate as a potential solution? I'm open to suggestions in terms of an alternate "signal" to monitor for new batches.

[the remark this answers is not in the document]
popzxc: I think that the right way would be to move `save_merkle_paths_artifacts_metadata` from `proof_generation_dal` to the `blocks_dal` (e.g. existence of the metainformation is the property of the block, regardless of the consumer).

This way we (later, not now!) will be able to get rid of `insert_proof_generation_details` in the `updater.rs` as well -- so that the tree is not aware of any verification logic (as it should be).

PR #2630, opened 2024-08-09
feat: External prover API metrics, refactoring

Artemka374:
## What ❔

Added metrics for external proof integration API, refactored code a little bit

## Checklist


core/node/external_proof_integration_api/src/processor.rs
```diff
@@ -39,20 +42,55 @@ impl Processor {
...
+                .blob_store
+                .get::<L1BatchProofForL1>((l1_batch_number, payload.0.protocol_version))
+                .await?,
+        )?;
+
+        if serialized_proof != expected_proof {
+            return Err(ProcessorError::InvalidProof);
+        }
+
+        guard.mark_successful();
```
popzxc, line 70: I think this approach is somewhat error-prone. Maybe a more generic way to do it is to implement a tower middleware via layer_fn (see [example](https://docs.rs/tower/latest/tower/layer/fn.layer_fn.html)). This would probably also let you add the "unknown" method label.

core/node/external_proof_integration_api/src/metrics.rs
```diff
@@ -0,0 +1,60 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, EncodeLabelSet, EncodeLabelValue)]
+#[metrics(label = "outcome", rename_all = "snake_case")]
+pub(crate) enum CallOutcome {
+    Success,
+    Failure,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, EncodeLabelSet, EncodeLabelValue)]
+#[metrics(label = "type", rename_all = "snake_case")]
+pub(crate) enum MethodType {
```
RomanBrodetski, line 15: nit: it's not really `MethodType`, it's just a `Method`, or `Endpoint`, or `MethodName`.

core/node/external_proof_integration_api/src/metrics.rs
```diff
@@ -0,0 +1,60 @@
...
+}
+
+#[derive(Debug, Metrics)]
+#[metrics(prefix = "external_proof_integration_api")]
+pub(crate) struct ProofIntegrationApiMetrics {
+    pub call_count: Family<MethodType, Counter<u64>>,
+    #[metrics(labels = ["type", "outcome"])]
+    pub call_outcome: LabeledFamily<(MethodType, CallOutcome), Counter<u64>, 2>,
+    #[metrics(buckets = vise::Buckets::LATENCIES)]
+    pub call_latency: Family<MethodType, Histogram<Duration>>,
```
RomanBrodetski, line 28: with this you don't really need the two metrics above. Histogram always creates a `counter` as well.

RomanBrodetski (changes requested):

core/node/external_proof_integration_api/src/processor.rs
Artemka374: UPD: will be done as follow-up due to tight timelines.

RomanBrodetski (approved):

PR #2631, opened 2024-08-09
feat: Evm equivalence (updated with latest changes from `main`)

jrchatruc:
## What ❔

This PR includes only the necessary changes from #1612 for EVM equivalance, starting from the current `main` branch.

The PR includes the following changes:

- Contract address is now considered optional for most of transaction related structs. The contract address with value `None` is used to deploy EVM contracts.
- A new Postgres migration is added with an `evm_simulator_code_hash` column added to the following tables:
  - `protocol_versions`
  - `l1_batches`
  - `miniblocks`
The corresponding `dal` modules were updated to account for these new columns.
- The `contracts` submodule points now contains the `EvmGasManager` and `EvmInterpreter` system contracts.
- The `EvmInterpreter` was added as a base system contract.
- The `decommiter` present in the `vm_latest` now has separate hash calculations for each version, `ZKEvm` and `EVM`. In particular the length of the pre-image differs (EVM is in bytes, zkEVM in words, i.e. 32 bytes)
- An `EvmDeployTracer`was added as another default tracer. Its purpose is to add EVM contracts to the decommiter once an evm deployment ends succesfully.
- For the `vm_latest`, the `TransactionData` now checks if the address is absent and in the afirmative case it writes the corresponding flag on the reserved bit to signal an EVM contract deployment.
- The decommitment processor for the `vm_latest` now has the hash and code of the EVM interpreter.
- The `EvmSimulator` hash is now part of the genesis and it has been added to the corresponding protobuf message.
- The `EthTxAggregator` now adds the call to the getters of the hash and code of the evm interpreter to the multicall.
- The `main_executor` calculates a `new_known_factory_deps` with all the new known EVM bytecodes with the help of the decommiter. These are then added to the regular factory deps.
- The `multivm_bootloader` cached compiled contracts under `etc/multivm_bootloaders` for the `vm_1_5_0_increased_memory` version were updated, accounting for the bootloader changes.
- The precalculated hashes of the chain and system contracts have been updated:
    - `bootloader_hash`
    - `default_aa_hash`
    - `evm_simulator_hash`
    - `GENESIS_ROOT`
    - `GENESIS_BATCH_COMMITMENT`
    - `GENESIS_ROLLUP_LEAF_INDEX`
- New integration tests that run uniswap EVM transactions

## How to test

1. Check out to the correct `era-contracts` branch in the lambda fork of the repo:
   ```
    git submodule update --init
    cd contracts
    git remote add lambda git@github.com:lambdaclass/era-contracts.git
    git fetch lambda
    ```
    The updated version of Solidity contracts is needed. These changes were made in this [PR](https://github.com/matter-labs/era-contracts/pull/688). To be able to test it, you can merge this changes using:
    ```
    git merge --no-commit lambda/update_solidity_version
    ```
    > This can lead to a few conflicts for some reason. Just stick with the incoming changes.
2. Run the following commands back on the root of `zksync-era`:
    ```
    export ZKSYNC_HOME=$(pwd) 
    export PATH=$ZKSYNC_HOME/bin:$PATH
    zk
    zk init --skip-submodules-checkout
    zk server
    ```
3. On another terminal, try to deploy an EVM contract using Forge by running:
    ```
    forge create --rpc-url localhost:3050 --private-key 0x27593fea79697e947890ecbecce7901b0008345e5d7259710d0dd5e500d040be Storage.sol:SimpleStorage --legacy
    ```
    You can use this contract as an example:
    ```solidity
    // Storage.sol
    pragma solidity ^0.8.0;
    
    contract SimpleStorage {
        uint256 private storedData;
    
        function set(uint256 value) public {
            storedData = value;
        }
    
        function get() public view returns (uint256) {
            return storedData;
        }
    }
    ```

## Checklist


core/lib/multivm/src/versions/vm_latest/types/internals/vm_state.rs
```diff
@@ -131,7 +143,7 @@ pub(crate) fn new_vm_state<S: WriteStorage, H: HistoryMode>(
             // For now, the default account hash is used as the code hash for the EVM simulator.
```
0xVolosnikov, line 143: Outdated

0xVolosnikov:
Could you please point out where we set reserved[1] to non-zero value?

core/lib/multivm/src/versions/vm_latest/types/internals/vm_state.rs
IAvecilla: Updated!

core/lib/multivm/src/versions/vm_latest/types/internals/transaction_data.rs
```diff
@@ -85,7 +91,7 @@ impl From<Transaction> for TransactionData {
                     value: execute_tx.execute.value,
                     reserved: [
                         should_check_chain_id,
-                        U256::zero(),
+                        should_deploy_contract,
                         U256::zero(),
                         U256::zero(),
                     ],
```
IAvecilla, line 97: @0xVolosnikov Here is the change on the reserved bytes of the transaction where we set the 0 or 1 depending if we want to deploy an EVM contract or not.

0xVolosnikov:
I think https://github.com/matter-labs/zksync-era/pull/2737 may be required to get green CI here

jrchatruc:
Closed in favor of #2852

PR #2632, opened 2024-08-11
feat(zk_toolbox): Add holesky testnet as layer1 network

fyInALT:
## What ❔

Add `holesky` for `--l1-network` config in ecosystem cmd, that will make the testnet use holesky testnet (chain id: 17000) as l1 network

## Why ❔

It can make us deploy testnet into holesky testnet more easy, in zk_inception, it will write the layer1 's chain id into the config, when use `ecosystem init` cmd, it will start chain init also, if use wrong chain id for layer1 endpoint, it will make cmd run failed, so we cannot just use layer1 endpoint to deploy a testnet in holesky.

Now we add `holesky` for  `--l1-network` config to let it write 17000 as chain id, that will make us to deploy testnet more easy.

## Checklist


zk_toolbox/crates/zk_inception/src/commands/ecosystem/init.rs
```diff
@@ -281,7 +281,7 @@ async fn deploy_ecosystem(
             L1Network::Localhost => {
                 ContractsConfig::get_path_with_base_path(&ecosystem_config.config)
             }
-            L1Network::Sepolia | L1Network::Mainnet => ecosystem_config
```
Deniallugo, line 284: We don't have holesky environment yet, if the Holesky will be chosen this code will fail

Deniallugo (commented):
Thanks for your contribution! 
Code Looks good, please add a fallback for holesky

zk_toolbox/crates/zk_inception/src/commands/ecosystem/init.rs
fyInALT: i think zksync team can deploy a generic ecosystem contracts into holesky for devnets, In currently i think i can just return a error and let user to deploy ecosystem contracts by themself

fyInALT: will show:

```
⚙  Command completed: yarn sc build
│  
◇  Provide the path to the ecosystem contracts or keep it empty and you will use ZkSync ecosystem config. For using this config, you need to have governance wallet
│  
│
■  Not found preexisting ecosystem Contracts with chains Holesky
```

Deniallugo (approved):

PR #2633, opened 2024-08-12
feat: add logs bloom

perekopskiy:
## What ❔

Add logs bloom to block header and tx receipt. Adds a migration task that backfills it for old blocks.

## Why ❔

API compatibility

## Checklist


core/lib/dal/src/blocks_dal.rs
```diff
@@ -2334,6 +2355,84 @@ impl BlocksDal<'_, '_> {
...
+        .fetch_one(self.storage)
+        .await?;
+
+        Ok(row.max.map(|n| L2BlockNumber(n as u32)))
+    }
+
+    pub async fn range_update_logs_bloom(
+        &mut self,
+        from_l2_block: L2BlockNumber,
+        to_l2_block: L2BlockNumber,
```
slowli, line 2402: Bikeshedding: IMO, it would be more idiomatic to use an inclusive range; ditto for other new queries. In this particular case, it's possible to skip `to_l2_block` since it can be trivially restored from other supplied data.

core/lib/dal/src/blocks_dal.rs
```diff
@@ -2334,6 +2355,84 @@ impl BlocksDal<'_, '_> {
...
+        blooms: Vec<Bloom>,
```
slowli, line 2403: Nit: `blooms` are not consumed by the method, so you can use `&[Bloom]` instead.

core/lib/dal/src/blocks_dal.rs
```diff
@@ -2334,6 +2355,84 @@ impl BlocksDal<'_, '_> {
...
+    ) -> DalResult<()> {
+        let numbers: Vec<_> = (i64::from(from_l2_block.0)..=i64::from(to_l2_block.0)).collect();
+        assert_eq!(
+            numbers.len(),
+            blooms.len(),
+            "range_update_logs_bloom: inconsistent array lengths"
+        );
+
+        let blooms: Vec<_> = blooms.into_iter().map(|b| b.0.to_vec()).collect();
```
slowli, line 2412: IIUC, it's possible to avoid allocations here by converting each `Bloom` to `&[u8]` and specifying `&blooms as &[&[u8]]` in the query below.

core/lib/dal/src/blocks_web3_dal.rs
```diff
@@ -87,7 +88,10 @@ impl BlocksWeb3Dal<'_, '_> {
                         .unwrap_or(i64::from(LEGACY_BLOCK_GAS_LIMIT))
                         as u64)
                         .into(),
-                    // TODO: include logs
+                    logs_bloom: row
```
slowli, line 91: Just to understand better: Is returning a dummy value here better than not returning a value at all, because it won't break Web3 clients?

core/lib/types/src/block.rs
```diff
@@ -345,4 +357,66 @@ mod tests {
...
+        ];
+        let iter = logs.iter().flat_map(|log| {
+            log.1
+                .iter()
+                .map(|topic| BloomInput::Raw(topic.as_bytes()))
+                .chain(iter::once(BloomInput::Raw(log.0.as_bytes())))
+        });
+
+        let bloom = build_bloom(iter);
+        let expected = Bloom::from_str("0x00000000040000000000000001000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000200000000000002000040000018000000000000000000000000000000000000000000000000000000000000000000000000000000000800000000002010000000002000000000000000400000000000080000008000000000000000000000000000000000000000000000000000000400000000000100000000000080400000000000000002000000000000000000000004000000000000000000000008002000000000000000000010000000000000000000000000000000000000000000000000").unwrap();
```
slowli, line 418: Nit: It's possible to escape newlines / following whitespace in a string literal, so that it's more readable:

```rust
Bloom::from_str(
    "0x0000000004000000000000000100000000000000000000000000\
     000000000000000000000000000000004000000000000000000000\
     000000000000000000020000000000000200004..."
)
```

core/lib/dal/src/events_dal.rs
```diff
@@ -409,6 +409,56 @@ impl EventsDal<'_, '_> {
...
+        .instrument("get_bloom_items_for_l2_block")
+        .fetch_all(self.storage)
+        .await?;
+
+        let mut items = HashMap::new();
+        for row in rows {
+            let block = L2BlockNumber(row.miniblock_number as u32);
+            let vec: &mut Vec<_> = items.entry(block).or_default();
+            vec.push(row.address);
+            if !row.topic1.is_empty() {
```
slowli, line 446: Nit: IIUC, this can be simplified to something like

```rust
let row_topics = [row.topic1, row.topic2, tow.topic3, row.topic4]
    .into_iter()
    .flatten();
vec.extend(row_topics);
```

core/node/logs_bloom_backfill/src/lib.rs
```diff
@@ -0,0 +1,113 @@
...
+            let blooms: Vec<_> = (left_bound..=right_bound)
+                .map(|block| {
+                    let items = bloom_items
+                        .remove(&L2BlockNumber(block))
+                        .unwrap_or_default();
+                    let iter = items.iter().map(|v| BloomInput::Raw(v.as_slice()));
+                    build_bloom(iter)
+                })
+                .collect();
+            connection
```
slowli, line 93: Theoretically, if pruning is enabled, it may be performed after Bloom items are loaded for a range, but before this call is made, removing the range partially or completely. I *think* this logic will work fine in this case; the updated for removed blocks will be a no-op, and the following iterations (if any) will be no-op as well.

core/node/logs_bloom_backfill/src/lib.rs
```diff
@@ -0,0 +1,113 @@
...
+        };
+        let first_l2_block = connection
+            .blocks_dal()
+            .get_earliest_l2_block_number()
+            .await?
+            .context(
+                "logs_bloom_backfill: missing l2 block in DB after waiting for at least one",
+            )?;
+
+        tracing::info!("starting blooms backfill");
```
slowli, line 69: Nit: IMO, it would make sense to log the range of filled blocks (maybe even have tracing spans for the backfill procedure and each `WINDOW`).

core/node/logs_bloom_backfill/src/lib.rs
```diff
@@ -0,0 +1,113 @@
...
+
+impl LogsBloomBackfill {
+    pub fn new(connection_pool: ConnectionPool<Core>) -> Self {
+        Self { connection_pool }
+    }
+
+    async fn wait_for_l2_block_with_bloom(
+        connection: &mut Connection<'_, Core>,
+        stop_receiver: &mut watch::Receiver<bool>,
+    ) -> anyhow::Result<bool> {
```
slowli, line 21: Bikeshedding: IMO, it would be more idiomatic to introduce an enum for successful results here:

```rust
enum BloomWaitOutcome {
    Ok,
    Canceled,
}
```

core/lib/dal/src/blocks_dal.rs
```diff
@@ -2334,6 +2355,84 @@ impl BlocksDal<'_, '_> {
 
         Ok(results.into_iter().map(L::from).collect())
     }
+
+    pub async fn get_last_l2_block_bloom(&mut self) -> DalResult<Option<Bloom>> {
```
slowli, line 2359: Will return values be used somewhere in the future? Right now, the only caller just checks whether the return value is `Some(_)`; so this method could be refactored to `has_l2_blocks_with_bloom() -> DalResult<bool>`.

slowli (commented):
Dumb question: Are Bloom logs influenced by transfer events we were going to remove in https://github.com/matter-labs/zksync-era/pull/701 (the PR is still open BTW 🙃 )? Or is it the case that because of how Bloom filters work, their consumers won't case about the fact that they contain some extra events?

core/node/logs_bloom_backfill/src/lib.rs
perekopskiy: I've added more logs, IMO spans are an overkill in this case

core/node/logs_bloom_backfill/src/lib.rs
perekopskiy: Yes, given that migration shouldn't take a very long time, I think it's ok to do a no-op for removed blocks

core/lib/dal/src/blocks_web3_dal.rs
perekopskiy: exactly

perekopskiy:
> Dumb question: Are Bloom logs influenced by transfer events we were going to remove in https://github.com/matter-labs/zksync-era/pull/701 (the PR is still open BTW 🙃 )? Or is it the case that because of how Bloom filters work, their consumers won't case about the fact that they contain some extra events?

Bloom filter is influenced by these events. IIUC correctly that PR is not going to be merged and is abandoned in favor of other long-term EVM-compatibility solutions. @RomanBrodetski @Artemka374 should we close that PR?

core/tests/ts-integration/tests/api/web3.test.ts
```diff
@@ -44,21 +44,36 @@ describe('web3 API compatibility tests', () => {
         const blockHash = (await alice.provider.getBlock(blockNumber)).hash!;
         const blockWithTxsByNumber = await alice.provider.getBlock(blockNumber, true);
         expect(blockWithTxsByNumber.gasLimit).toBeGreaterThan(0n);
-        let sumTxGasUsed = 0n;
 
+        // `ethers.Block` doesn't include `logsBloom` for some reason.
+        const blockWithTxsByNumberFull = await alice.provider.send('eth_getBlockByNumber', [blockNumberHex, true]);
```
slowli, line 49: Nit: Is there any particular reason to fetch transactions (by setting the 2nd arg of the call to `true`) instead of hashes? Looks like only hashes are used.

core/node/state_keeper/src/io/seal_logic/mod.rs
```diff
@@ -358,6 +361,17 @@ impl L2BlockSealCommand {
         // Run sub-tasks in parallel.
         L2BlockSealProcess::run_subtasks(self, strategy).await?;
 
+        let progress = L2_BLOCK_METRICS.start(L2BlockSealStage::CalculateLogsBloom, is_fictive);
+        let iter = self.l2_block.events.iter().flat_map(|event| {
+            event
+                .indexed_topics
+                .iter()
+                .map(|topic| BloomInput::Raw(topic.as_bytes()))
+                .chain(std::iter::once(BloomInput::Raw(event.address.as_bytes())))
```
slowli, line 370: Nit: IIUC, you can use `[BloomInput::Raw(_)]` (i.e., a single-item array) instead of `iter::once()` for brevity.

core/node/node_framework/src/implementations/layers/logs_bloom_backfill.rs
```diff
@@ -0,0 +1,61 @@
...
+impl WiringLayer for LogsBloomBackfillLayer {
+    type Input = Input;
+    type Output = Output;
+
+    fn layer_name(&self) -> &'static str {
+        "logs_bloom_backfill_layer"
+    }
+
+    async fn wire(self, input: Self::Input) -> Result<Self::Output, WiringError> {
+        let main_pool = input.master_pool.get_singleton().await.unwrap();
```
slowli, line 40: Makes sense to propagate an error here instead of `unwrap`ping.

core/lib/dal/src/blocks_dal.rs
```diff
@@ -2334,6 +2355,80 @@ impl BlocksDal<'_, '_> {
...
+
+        Ok(row.max.map(|n| L2BlockNumber(n as u32)))
+    }
+
+    pub async fn range_update_logs_bloom(
+        &mut self,
+        from_l2_block: L2BlockNumber,
+        blooms: &[Bloom],
+    ) -> DalResult<()> {
+        let to_l2_block = from_l2_block + (blooms.len() - 1) as u32;
```
slowli, line 2402: Nit: Not sure whether it can be triggered, but adding parentheses means an integer underflow if `blooms` are empty. To be fair, it would be corrected during addition; the only worrying case is if `blooms` are empty and `from_l2_block == L2BlockNumber(0)`. Could return early if `blooms` are empty to avoid these cases.

core/node/logs_bloom_backfill/src/lib.rs
slowli, line 1: Nit: I'd add a test covering the workflow (seems relatively straightforward to set up storage for it and check results), but I understand it's not the top priority.

core/lib/dal/src/blocks_web3_dal.rs
```diff
@@ -219,7 +223,7 @@ impl BlocksWeb3Dal<'_, '_> {
                         .into(),
                     base_fee_per_gas: Some(bigdecimal_to_u256(row.base_fee_per_gas.clone())),
                     extra_data: Bytes::default(),
-                    logs_bloom: H2048::default(),
+                    logs_bloom: Bloom::default(),
```
slowli, line 226: Dumb question: Why is the default value OK here? AFAIU, this method is used in subscriptions, and the `logsBloom` field is visible there.

core/node/node_framework/src/implementations/layers/logs_bloom_backfill.rs
perekopskiy: copied from another place, now propagated in both

core/lib/dal/src/blocks_web3_dal.rs
perekopskiy: This place was missed, nice catch!

slowli (approved):

PR #2638, opened 2024-08-12
feat(vm): Extract VM interface to separate crate

slowli:
## What ❔

- Refactors `zksync_multivm` crate by extracting its interface part into a separate crate.
- Revises `zksync_state` / `zksync_multivm` uses in the workspace.

## Why ❔

- This eliminates `multivm` dependency on `zksync_state`, which is quite heavyweight (depends on `sqlx`, `rocksdb` etc.).
- Makes VM interface better defined and easier to reason about.

## Checklist


slowli (commented):
- The changeset stats look monstrous, but most files just have imports changed.
- I plan to try moving some VM-related types from `zksync_types` and `zksync_utils` into `zksync_vm_interface` in a follow-up PR; this one is large as-is.

Deniallugo (approved):
LGTM, Glad you managed to extract it

core/lib/vm_utils/Cargo.toml
```diff
@@ -11,12 +11,11 @@ keywords.workspace = true
 categories.workspace = true
 
 [dependencies]
-zksync_multivm.workspace = true
-zksync_types.workspace = true
+zksync_contracts.workspace = true
 zksync_dal.workspace = true
-zksync_state.workspace = true
+zksync_types.workspace = true
+zksync_vm_interface.workspace = true
```
popzxc, line 17: With this change, `zksync_vm_utils` now feels even more bogus. Do you think it'll be possible to merge it with the newly added crate?
The main complexity seems to be the fact that it's glued to `zksync_dal`, but if you look at it, it seems to be mostly used to fetch VM inputs. Given that `zksync_state` already provides an implementation for the Postgres-backed storage, maybe it can be solved by introducing a new abstraction (e.g. `VmBootstrap`) that will have a single Postgres-backed implementation in `zksync_state` crate.

Not proposing doing it in this PR though.

popzxc (approved):
I cannot even express how happy I am to see that change, thank you! ❤️ 

I love the fact that traits from `zksync_state` were moved to the new crate. If I don't miss anything, probably it means that `era-test-node` no longer will have to depend on `zksync_dal` (which was a shame, given that it doesn't use Postgres in any way), so the benefits here expand beyond this repository :)

If we think about further steps here, I have the following thing in mind:
We probably should prepare abstractions to extract `execution_sandbox` from the API. I have a feeling that the abstraction for `VmBootstrap` I proposed in the comment may be a significant chunk of work here. If it will encompass stuff like loading system contracts, then using historical/playground system contracts in the API should work out of the box. This will be a path towards having a single interface for the VM instead of 3.5 (state keeper/vm runner, API, era_test_node) and help with (currently somewhat ugly) implementation of `eth_call` in the consensus module.

core/lib/vm_utils/Cargo.toml
slowli: I certainly want to remove `zksync_vm_utils` in some way; as you've said, not in this PR.

PR #2641, opened 2024-08-12
chore: bump consensus to 0.1.0-rc.10

pompon0:


aakoshh (approved):
For the record the only change in rc.10 is that era-consensus is built with the _portable_ version of the BLST library, which will enable at least one EN operator running their node on some machine that doesn't come with BLST support to generate keys and then also to run the node.

PR #2644, opened 2024-08-13
feat(eth-sender): add option to pause aggregator for gateway migration

tomg10:


core/node/eth_sender/src/eth_tx_aggregator.rs
```diff
@@ -355,6 +355,27 @@ impl EthTxAggregator {
...
+                    "Skipping sending operation of type {} for batches {}-{} \
+                as tx_aggregation_paused=true",
+                    agg_op.get_action_type(),
+                    agg_op.l1_batch_range().start(),
+                    agg_op.l1_batch_range().end()
+                );
+                return Ok(());
+            }
+            if self.config.tx_aggregation_only_prove_and_execute
+                && agg_op.get_action_type() == AggregatedActionType::Commit
```
StanislavBreadless, line 369: In the future we might split the batch committing into two steps:
- Publishing of DA-related data to L1
- Committing the batch itself

If we do that we will forget about this line and it wont work for the "publish da" type of the aggregated action. I would suggest creating a function call `is_prove_or_execute` that returns explicitly if the aggregated type is either `Prove` or `Execute`

tomg10: good comment, corrected. I plan to remove those flags as soon as migration is complete, but there are chances I forget about that.

StanislavBreadless (approved):

PR #2645, opened 2024-08-13
refactor(vm): Move more types to VM crates

slowli:
## What ❔

Moves some types from `zksync_types` to `zksync_vm_interface` or `zksync_multivm` crates.

## Why ❔

So that types are separated by domain rather than all collected in `zksync_types`.

## Checklist


core/lib/multivm/src/utils/bytecode.rs
```diff
@@ -0,0 +1,154 @@
+use std::collections::HashMap;
+
+use zksync_types::ethabi::{self, Token};
+use zksync_utils::bytecode::{hash_bytecode, validate_bytecode, InvalidBytecodeError};
```
slowli, line 4: Ideally, I'd like to deal with `zksync_utils` so that it doesn't gather functionality across all domains, but this is beyond the scope of this PR.

core/lib/vm_interface/src/lib.rs
```diff
@@ -1,4 +1,21 @@
 //! ZKsync Era VM interfaces.
+//!
+//! # Developer guidelines
```
slowli, line 3: Are these guidelines visible enough? I've thought about placing them in the crate readme, but it would look slightly out of place IMO.

core/lib/vm_interface/src/types/outputs/execution_result.rs
```diff
@@ -103,3 +105,50 @@ impl VmExecutionResultAndLogs {
...
+#[derive(Debug, Clone, PartialEq)]
+pub struct TransactionExecutionResult {
+    pub transaction: Transaction,
+    pub hash: H256,
+    pub execution_info: VmExecutionMetrics,
+    pub execution_status: TxExecutionStatus,
+    pub refunded_gas: u64,
+    pub operator_suggested_refund: u64,
+    pub compressed_bytecodes: Vec<CompressedBytecodeInfo>,
+    pub call_traces: Vec<Call>,
```
slowli, line 134: Now that I've poked around it a bit, I think `Call` can be moved to `vm_interface` as well. Will try to do separately to keep this PR reasonably sized.

core/lib/multivm/src/utils/bytecode.rs
popzxc: Absolutely.

popzxc (approved):

core/lib/vm_interface/src/lib.rs
perekopskiy: I think it's good enough to have them here

perekopskiy (approved):

PR #2652, opened 2024-08-14
feat(prover): parallelized memory queues simulation in BWG

0xVolosnikov:
## Checklist


0xVolosnikov:
Requires https://github.com/matter-labs/era-zkevm_test_harness/pull/166

0xVolosnikov:
Also requires an update in shivini

0xVolosnikov:
1. We need to simulate sorted and unsorted queues to the end to be able to calculate closed form input for the first ram permutation circuit
2. We need to have subarrays of queues states for every ram permutation circuit
3. These subarrays are huge, they are more than anything else in terms of RAM

So I do not see now any other way to deal with it other then split monolithic ram permutaion circuit artifacts

0xVolosnikov:
Maybe we can come up with a more elegant way to store/recombine data?

PR #2654, opened 2024-08-14
feat(db): Allow creating owned Postgres connections

slowli:
## What ❔

- Changes `Connection` so that it has `'static` lifetime if created from a pool (i.e., when it is non-transactional).
- Simplifies `ReadStorageFactory` and `MainBatchExecutor` accordingly.

## Why ❔

Reduces complexity. `'static` connections can be sent to a Tokio task etc., meaning improved DevEx.

## Checklist


popzxc (approved):
Nice!

PR #2655, opened 2024-08-14
refactor(vm): Move event types to VM interface and multivm crates

slowli:
## What ❔

Continuing from https://github.com/matter-labs/zksync-era/pull/2645, this PR moves VM event types to VM interface / implementation crates.

## Why ❔

So that types are separated by domain rather than all collected in `zksync_types`.

## Checklist


slowli:
@joonazan I'm not entirely sure what performance degradation of legacy VMs is caused by. I haven't been able to reproduce it locally; there are minor changes in estimated cycles on some benches, but they are order of 0.1%. AFAIU, `iai` / cachegrind output still depends on architecture-specific factors like binary layout etc., so it could be a reason (evidence that arm64 architecture is better than x86_64 🙃). I suspect that moving types around disabled some previously used optimizations because they would be cross-crate now (which AFAIU, are not always performed unless LTO is enabled).

joonazan:
A difference this small is probably just a random change in compilation. Iai does not correspond 1:1 to real-world performance so it is unclear if it is better or worse.

core/node/state_keeper/src/updates/l2_block_updates.rs
```diff
@@ -1,23 +1,46 @@
...
     vm_trace::Call,
-    L2BlockNumber, ProtocolVersionId, StorageLogWithPreviousValue, Transaction, VmEvent, H256,
+    L2BlockNumber, ProtocolVersionId, StorageLogWithPreviousValue, Transaction, H256,
 };
 use zksync_utils::bytecode::hash_bytecode;
 
 use crate::metrics::KEEPER_METRICS;
 
+/// Extracts all bytecodes marked as known on the system contracts.
+fn extract_bytecodes_marked_as_known(all_generated_events: &[VmEvent]) -> Vec<H256> {
```
perekopskiy, line 24: Why isn't it a part of `VmEvent` impl?

slowli: In general, I've moved all `VmEvent` uses as close to their uses as possible; thus, all single-case uses were moved to the higher-level crates. I'm not sure this is the best approach, so if you say that this functionality can actually be reused in the future, I'll move it to `VmEvent`.

perekopskiy: I see, no, I don't think it will be reused in future. Makes sense to leave it here

perekopskiy (approved):

montekki (approved):

PR #2666, opened 2024-08-15
feat(prover): Add ProverJobMonitor

EmilLuta:
ProverJobMonitor will be house keeper's counter part in prover subsystem. TL;DR; it's a singleton component, monitoring prover subsystem jobs.

The TL;DR; is that prover and core won't share any databases. This enables:
- core deployments without affecting prover
- removing prover infrastructure (DB) in proverless envs

The release plan is as follows:
- release a component (PJM) that runs in parallel with HK
- migrate all jobs/metrics/dashboards to PJM
- delete their counterparts in HK
- remove redundant infrastructure

This PR contains:
- a new component (PJM)
- fixes for bugs/issues with old metrics (backported to HK)
- refactoring of metrics (PJM metrics cover same metrics as HK, but they are different, as we can cover more with less)
- various other small nits

P.S. Name is up for discussion, feel free to suggest better name.

### DIFF
--- a/core/lib/config/src/configs/prover_job_monitor.rs
+++ b/core/lib/config/src/configs/prover_job_monitor.rs
@@ -9,6 +9,9 @@ pub struct ProverJobMonitorConfig {
     /// Port for prometheus metrics connection.
     pub prometheus_port: u16,
     /// Maximum number of database connections per pool.
+    /// In a balanced system it should match the number of Tasks ran by ProverJobMonitor.
+    /// If lower, components will wait on one another for a connection.
+    /// If more, database will use more resources for idle connections (which drains DB resources needed for other components in Prover Subsystems).
     pub max_db_connections: u32,
     /// Amount of time ProverJobMonitor will wait all it's tasks to finish.
     #[serde(default = "ProverJobMonitorConfig::default_graceful_shutdown_timeout_ms")]


prover/crates/bin/prover_job_monitor/src/job_requeuer/prover_job_requeuer.rs
```diff
@@ -0,0 +1,63 @@
...
+#[async_trait]
+impl PeriodicJob for ProverJobRequeuer {
+    const SERVICE_NAME: &'static str = "ProverJobRequeuer";
+
+    async fn run_routine_task(&mut self) -> anyhow::Result<()> {
+        let stuck_jobs = self
+            .pool
+            .connection()
+            .await
+            .unwrap()
```
popzxc, line 46: There is quite some boilerplate here. I believe that it can be significantly reduced, if we rework the `PeriodicJob` as follows.

This way you will have a builder, so you have to specify less details in each job impl, you don't have to store pool insdide of each job, and spawning will be much easier.

Since it's a thing specific to this crate, you don't really need to extract it into a separate crate (so that you don't have to refactor existing house keeper code). You can move `PeriodicJob` back to the house keeper so that it dies there, and instead define the structure and crate here. Later, if it will be useful, we will generalize the type over DB type (so that it can be used with `Core` as well) and extract.

```rust
#[async_trait::async_trait]
pub trait DalJob {
    async fn invoke(&self, connection: &mut Connection<Prover>) -> anyhow::Result<()>;
}

// Internal repr of a periodic job
struct PeriodicDalJob {
    job: Box<dyn DalJob>,
    pool: ConnectionPool<Prover>,
    name: String,
    interval: Duration,
}

impl PeriodicDalJob {
    async fn run(&self, stop_receiver: sync::Watch<bool>) -> anyhow::Result<()> {
        while !*stop_receiver.borrow() {
            let mut connection = self.pool.get().await?;
            self.job.invoke(&mut connection).await?;
            tokio::time::sleep(self.interval).await;
            // Do whatever reporing/logging you need
        }
    }
}

pub struct DalJobRunner {
    pool: ConnectionPool<Prover>,
    tasks: Vec<PeriodicDalJob>,
}

impl DalJobRunner {
    pub fn new(pool: ConnectionPool<Prover>) -> Self {
        Self { pool, tasks: Vec::new() }
    }

    pub fn push_job<T: DalJob>(&mut self, name: &str, interval: Duration, job: T) {
        self.tasks.push(PeriodicDalJob { name: name.into(), interval, job: Box::new(job), pool: self.pool.clone() });
    }

    pub fn spawn(self, stop_receiver: tokio::sync::Watch<bool>) -> Vec<tokio::task::JoinHandle<anyhow::Result<()>>> {
        self.tasks.into_iter().map(|task| {
            let pool = self.pool.clone();
            let receiver = stop_receiver.clone();
            tokio::spawn(async move {
                task.run(receiver).await
            })
        }).collect()
    }
}
```

prover/crates/bin/prover_job_monitor/src/queue_reporter/prover_queue_reporter.rs
```diff
@@ -0,0 +1,98 @@
...
+#[async_trait]
+impl PeriodicJob for ProverQueueReporter {
+    const SERVICE_NAME: &'static str = "ProverQueueReporter";
+
+    async fn run_routine_task(&mut self) -> anyhow::Result<()> {
+        let mut conn = self.connection_pool.connection().await.unwrap();
+        let stats = conn.fri_prover_jobs_dal().get_prover_jobs_stats().await;
+
+        for (protocol_semantic_version, circuit_prover_stats) in stats {
+            for (
```
popzxc, line 42: Same as comment above about destructing the tuple.

prover/crates/lib/prover_dal/src/fri_prover_dal.rs
```diff
@@ -429,27 +431,47 @@ impl FriProverDal<'_, '_> {
...
+                // not the circuit which is running.
+                // There is a single node level aggregation circuit, which is circuit 2.
+                // This can aggregate multiple leaf nodes (which may belong to different circuits).
+                // This "conversion" is a forced hacky  way to use `circuit_id` 2 for nodes.
+                // A proper fix will be later provided to solve this once new auto-scaler is in place.
+                let circuit_id = if row.aggregation_round == 2 {
+                    2
+                } else {
+                    row.circuit_id as u8
+                };
```
popzxc, line 465: A bit safer way to do it would be to define the conversion of `row` (or its data) to the `CircuitProverStatsEntry`. This conversion would be a pure function and can be covered with a couple of tests for this specific behavior.

prover/crates/bin/prover_job_monitor/src/main.rs
```diff
@@ -0,0 +1,189 @@
...
+        .complete(Duration::from_millis(graceful_shutdown_timeout_ms))
+        .await;
+
+    Ok(())
+}
+
+fn get_tasks(
+    connection_pool: ConnectionPool<Prover>,
+    prover_job_monitor_config: ProverJobMonitorConfig,
+    proof_compressor_config: FriProofCompressorConfig,
```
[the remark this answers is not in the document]
popzxc: I think that having dedicated configs is nice. It makes it easier to deprecate and introduce new jobs, if we will need them.

etc/env/file_based/general.yaml
```diff
@@ -272,6 +272,23 @@ prover_group:
...
     - circuit_id: 18
       aggregation_round: 1
+prover_job_monitor:
+  prometheus_port: 3317
+  max_db_connections: 9
+  graceful_shutdown_timeout_ms: 5000
+  gpu_prover_archiver_run_interval_ms: 86400000
+  gpu_prover_archiver_archive_prover_after_secs: 172800
+  prover_jobs_archiver_run_interval_ms: 1800000
+  prover_jobs_archiver_archive_jobs_after_secs: 172800
```
montekki, line 282:
```suggestion
  prover_jobs_archiver:
    run_interval_ms: 1800000
    jobs_after_secs: 172800
```

another nit: Another layering of config could potentially improve readability here

prover/crates/bin/prover_job_monitor/src/job_requeuer/proof_compressor_job_requeuer.rs
```diff
@@ -0,0 +1,63 @@
...
+#[async_trait]
+impl PeriodicJob for ProofCompressorJobRequeuer {
+    const SERVICE_NAME: &'static str = "ProofCompressorJobRequeuer";
+
+    async fn run_routine_task(&mut self) -> anyhow::Result<()> {
+        let stuck_jobs = self
+            .pool
+            .connection()
+            .await
+            .unwrap()
```
montekki, line 46: Another point where `?` would probably work?

core/lib/config/src/configs/general.rs
```diff
@@ -33,7 +34,7 @@ pub struct GeneralConfig {
     pub prover_gateway: Option<FriProverGatewayConfig>,
     pub witness_vector_generator: Option<FriWitnessVectorGeneratorConfig>,
     pub prover_group_config: Option<FriProverGroupConfig>,
-    pub witness_generator: Option<FriWitnessGeneratorConfig>,
+    pub witness_generator_config: Option<FriWitnessGeneratorConfig>,
```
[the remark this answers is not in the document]
Deniallugo: btw, you don't need to type `zk f`

other than that, it's not breaking, before you change protobuf configs, these configs and representation  are completely separate

deny.toml
```diff
@@ -6,7 +6,9 @@ vulnerability = "deny"
 unmaintained = "warn"
 yanked = "warn"
 notice = "warn"
-ignore = []
+ignore = [
```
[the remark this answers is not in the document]
Deniallugo: I have no objections to update it

core/node/house_keeper/src/prover/queue_reporter/fri_prover_queue_reporter.rs
```diff
@@ -39,45 +39,42 @@ impl PeriodicJob for FriProverQueueReporter {
         let mut conn = self.prover_connection_pool.connection().await.unwrap();
```
Artemka374, line 39: Wait, isn't having prover connection contradicts with PR description? I'm kinda sure that this can be moved to prover job monitor as well.

prover/crates/bin/prover_job_monitor/src/main.rs
```diff
@@ -0,0 +1,276 @@
...
+
+// fn get_tasks(
+//     connection_pool: ConnectionPool<Prover>,
+//     prover_job_monitor_config: ProverJobMonitorConfig,
+//     proof_compressor_config: FriProofCompressorConfig,
+//     prover_config: FriProverConfig,
+//     witness_generator_config: FriWitnessGeneratorConfig,
+//     prover_group_config: FriProverGroupConfig,
+//     stop_receiver: watch::Receiver<bool>,
+// ) -> anyhow::Result<Vec<JoinHandle<anyhow::Result<()>>>> {
```
montekki, line 211: commented code should probably not get into the repo.

core/node/house_keeper/src/prover/queue_reporter/fri_prover_queue_reporter.rs
Artemka374: If the question is in metrics for autoscaler, it shouldn't be hard at all to change them in autoscaler config.

prover/crates/bin/prover_job_monitor/src/queue_reporter/witness_generator_queue_reporter.rs
```diff
@@ -0,0 +1,71 @@
...
+
+use crate::{
+    metrics::{JobStatus, PROVER_JOB_MONITOR_METRICS},
+    task_wiring::Task,
+};
+
+/// `WitnessGeneratorQueueReporter` is a task that reports witness generator jobs status.
+/// Note: these values will be used for auto-scaling witness generators (Basic, Leaf, Node, Recursion Tip and Scheduler).
+#[derive(Debug)]
+pub struct WitnessGeneratorQueueReporter {}
```
montekki, line 16:
```suggestion
pub struct WitnessGeneratorQueueReporter;
```

not needed for empty struct.

prover/crates/bin/prover_job_monitor/src/main.rs
```diff
@@ -0,0 +1,276 @@
...
+    task_runner.add(
+        "WitnessJobQueuer",
+        prover_job_monitor_config.witness_job_queuer_run_interval(),
+        witness_job_queuer,
+    );
+
+    Ok(task_runner.spawn(stop_receiver))
+}
+
+// fn get_tasks(
```
Artemka374, line 203: Eh, what is this?

prover/crates/bin/prover_job_monitor/src/queue_reporter/prover_queue_reporter.rs
```diff
@@ -0,0 +1,83 @@
...
+
+#[async_trait]
+impl Task for ProverQueueReporter {
+    async fn invoke(&self, connection: &mut Connection<Prover>) -> anyhow::Result<()> {
+        let stats = connection
+            .fri_prover_jobs_dal()
+            .get_prover_jobs_stats()
+            .await;
+
+        for (protocol_semantic_version, circuit_prover_stats) in stats {
```
Artemka374, line 32: Ah, I see what are you trying to do... But still, I don't see much value in still having this code in house keeper

prover/crates/bin/prover_job_monitor/src/task_wiring.rs
```diff
@@ -0,0 +1,86 @@
...
+                .context("failed to get database connection")?;
+            self.job
+                .invoke(&mut connection)
+                .instrument(tracing::info_span!("run", service_name = %self.name))
+                .await
+                .context("failed to invoke task")?;
+            // Error here corresponds to a timeout w/o `stop_receiver` changed; we're OK with this.
+            tokio::time::timeout(self.interval, stop_receiver.changed())
+                .await
+                .ok();
```
montekki, line 44: a nit: you probably would want a better expression of a periodic task with a ready to use primitive [tokio::time::interval](https://docs.rs/tokio/latest/tokio/time/fn.interval.html) + couple it with select over it and stop condition. That would be a dash more idiomatic and tell the reader faster whats going on

prover/crates/bin/prover_job_monitor/src/witness_job_queuer.rs
```diff
@@ -0,0 +1,121 @@
...
+
+use crate::{
+    metrics::{WitnessType, PROVER_JOB_MONITOR_METRICS},
+    task_wiring::Task,
+};
+
+/// `WitnessJobQueuer` is a task that moves witness generator jobs from 'waiting_for_proofs' to 'queued'.
+/// Note: this task is the backbone of scheduling/getting ready witness jobs to execute.
+#[derive(Debug)]
+pub struct WitnessJobQueuer {}
```
montekki, line 12:
```suggestion
pub struct WitnessJobQueuer;
```

core/lib/basic_types/src/basic_fri_types.rs
```diff
@@ -136,6 +144,16 @@ impl AggregationRound {
...
+
+    pub fn all_rounds() -> Vec<AggregationRound> {
+        vec![
+            AggregationRound::BasicCircuits,
+            AggregationRound::LeafAggregation,
+            AggregationRound::NodeAggregation,
+            AggregationRound::RecursionTip,
+            AggregationRound::Scheduler,
+        ]
+    }
```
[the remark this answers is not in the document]
montekki: No, but I would suggest to at least improve the style by making it a `const fn` or even making the whole array a values a `const` value. As for exhaustiveness: probably not worth the time though

core/lib/basic_types/src/basic_fri_types.rs
```diff
@@ -187,6 +203,156 @@ impl TryFrom<i32> for AggregationRound {
...
+            2
+        } else {
+            circuit_id as u8
+        };
+        let circuit_id_round_tuple = CircuitIdRoundTuple::new(circuit_id, aggregation_round as u8);
+        CircuitProverStatsEntry {
+            circuit_id_round_tuple,
+            protocol_semantic_version,
+            job_count_statistics,
+        }
```
RomanBrodetski, line 352: hairsplitting, feel free to ignore, but does this really belong here? we seem to make basic_types aware of things that are so close to the DB. IMO this function would better fit to the DAL

core/lib/config/src/configs/prover_job_monitor.rs
```diff
@@ -0,0 +1,182 @@
...
+use serde::{Deserialize, Serialize};
+
+/// Config used for running ProverJobMonitor.
+/// It handles configuration for setup of the binary (like database connections, prometheus) and configuration for jobs that are being ran.
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+pub struct ProverJobMonitorConfig {
+    /// Port for prometheus metrics connection.
+    pub prometheus_port: u16,
+    /// Maximum number of database connections per pool.
+    pub max_db_connections: u32,
```
RomanBrodetski, line 12: hmmm would be good to better understand the required number of DB connections for this component. Or Is it the more the merrier? 

From the prospective of system operator, it's be good to know the valid/recommended ranges for this param. Would the component be functional with just one connection?

RomanBrodetski (approved):

RomanBrodetski (approved):

core/lib/basic_types/src/basic_fri_types.rs
```diff
@@ -187,6 +203,156 @@ impl TryFrom<i32> for AggregationRound {
     }
 }
 
+/// Wrapper for mapping from protocol version to prover circuits job stats
+#[derive(Debug)]
+pub struct ProtocolVersionedCircuitProverStats {
```
RomanBrodetski, line 208: Do you actually, really need this? You first collect a list (iterable) of stats. Then you collect it into this struct. But then you only use it to iterate over and report metrics. Could we just use the iterator instead and spare some dozens of LoC?

I may be wrong here, only had a quick look

RomanBrodetski: you addressed exactly my point. I see now why it's needed. Let's keep as is

Artemka374 (approved):

prover/crates/bin/vk_setup_data_generator_server_fri/proptest-regressions/tests.txt
```diff
@@ -0,0 +1,9 @@
+# Seeds for failure cases proptest has generated in the past. It is
+# automatically read and these particular cases re-run before any
+# novel cases are generated.
+#
+# It is recommended to check this file in to source control so that
+# everyone who runs the test benefits from these saved cases.
+cc ca181a7669a6e07b68bce71c8c723efcb8fd2a4e895fc962ca1d33ce5f8188f7 # shrinks to circuit_id = 1
+cc ce71957c410fa7af30e04b3e85423555a8e1bbd26b4682b748fa67162bc5687f # shrinks to circuit_id = 1
+cc 6d3b0c60d8a5e7d7dc3bb4a2a21cce97461827583ae01b2414345175a02a1221 # shrinks to key = ProverServiceDataKey { circuit_id: 1, round: BasicCircuits }
```
popzxc, line 9: Was this an intended change?

PR #2668, opened 2024-08-16
test(vm): Refactor VM benchmarks

slowli:
## What ❔

- Integrates Prometheus metrics into criterion benches; removes the DIY benchmark correspondingly.
- Merges the main benchmark crate with the harness one.
- Includes benched bytecodes into the crate itself rather than reading them in runtime.

## Why ❔

Makes VM benchmarks more maintainable.

## Checklist


.github/workflows/vm-perf-to-prometheus.yml
```diff
@@ -21,7 +26,7 @@ jobs:
 
       - name: setup-env
         run: |
-          echo PUSH_VM_BENCHMARKS_TO_PROMETHEUS=1 >> .env
+          echo BENCHMARK_PROMETHEUS_PUSHGATEWAY_URL="http://vmagent.stage.matterlabs.corp/api/v1/import/prometheus/metrics/job/vm-benchmark" >> .env
```
slowli, line 24: JFYI, this string was present in the source code previously. May makes sense to move it to repo env / secrets.

core/tests/vm-benchmark/README.md
```diff
@@ -9,35 +9,22 @@ benchmarks, however.
...
-them to "benches/iai.rs".
+You can add new bytecodes to be benchmarked into the [`bytecodes`](src/bytecodes) directory and then add them to the
+`BYTECODES` constant exported by the crate.
 
 ## Profiling (Linux only)
 
 You can also use `sh perf.sh bytecode_file` to produce data that can be fed into the
 [firefox profiler](https://profiler.firefox.com/) for a specific bytecode.
-
-## Fuzzing
```
slowli, line 34: AFAIU, this section is irrelevant for newer VM versions.

slowli (commented):
Couple of general comments:

Iai improvements are probably caused by reducing the setup complexity. Could also to do with removing some deps that have static initialization logic like RocksDB (although it should be filtered out by iai calibration logic).

If removing the DIY benchmark / extending Criterion is controversial, I can split it off from this PR. IMO, having uniform benchmark framework for running locally and Prometheus reporting is good since it allows not to define same benches multiple times. The API looks relatively straightforward. It has its drawbacks though; e.g., collected stats include warm-up iterations (so, e.g. reported means slightly differ between Criterion and Prometheus).

core/tests/vm-benchmark/README.md
joonazan: Yes, fuzzing the old VM is pretty useless because you need to manually edit rust std to get high stability. Fuzzing the new VM to find exceptionally poorly performing inputs may be worthwhile.

joonazan:
Regarding Criterion vs. DIY: DIY gives more useful data. The average time reported by criterion is a very bad metric, especially in noisy CI. Because our programs don't use any randomized algorithms (except hashmap in old VM), the minimum time is a good measurement.

The newer [Divan](https://docs.rs/divan/latest/divan/) framework is as pleasant as Criterion and reports minima. I use it in the vm2 repo.

slowli:
> The average time reported by criterion is a very bad metric, especially in noisy CI.

My understanding is that in CI, we would primarily rely on metrics exported to Prometheus, which do feature the minimum iteration time (along with maximum, mean and median, and the number of iterations). They could be printed to stdout as well (added that in the latest commit). So if it's only a question of whether these values are *displayed*, it's not an issue.

.github/workflows/vm-perf-to-prometheus.yml
popzxc: Yes; please ping DevOps for that.

popzxc (dismissed):

popzxc (approved):

yorik (approved):

PR #2671, opened 2024-08-16
feat(vm): Extract oneshot VM executor interface

slowli:
## What ❔

Extracts oneshot VM executor from the API server crate.

## Why ❔

Simplifies reasoning about oneshot VM execution and its maintenance. Allows for alternative implementations.

## Checklist

### DIFF
--- a/core/node/api_server/src/execution_sandbox/apply.rs
+++ b/core/node/api_server/src/execution_sandbox/apply.rs
@@ -16,7 +16,7 @@ use zksync_multivm::{
     interface::{
         storage::{ReadStorage, StoragePtr, StorageView, WriteStorage},
         BytecodeCompressionError, L1BatchEnv, L2BlockEnv, OneshotEnv, PendingL2BlockEnv, SystemEnv,
-        VmExecutionMode, VmExecutionResultAndLogs, VmInterface,
+        TxExecutionMode, VmExecutionMode, VmExecutionResultAndLogs, VmInterface,
     },
     tracers::StorageInvocations,
     utils::adjust_pubdata_price_for_tx,
@@ -341,8 +341,21 @@ impl<S: ReadStorage> VmSandbox<S> {
     }
 }
 
+/// Main [`OneshotExecutor`] implementation used by the API server.
 #[derive(Debug, Default)]
-pub struct MainOneshotExecutor;
+pub struct MainOneshotExecutor {
+    missed_storage_invocation_limit: usize,
+}
+
+impl MainOneshotExecutor {
+    /// Creates a new executor with the specified limit of cache misses for storage read operations (an anti-DoS measure).
+    /// The limit is applied for calls and gas estimations, but not during transaction validation.
+    pub fn new(missed_storage_invocation_limit: usize) -> Self {
+        Self {
+            missed_storage_invocation_limit,
+        }
+    }
+}
 
 #[async_trait]
 impl<S> OneshotExecutor<S> for MainOneshotExecutor
@@ -358,9 +371,16 @@ where
         args: TxExecutionArgs,
         tracers: Self::Tracers,
     ) -> anyhow::Result<VmExecutionResultAndLogs> {
+        let missed_storage_invocation_limit = match args.execution_mode {
+            // storage accesses are not limited for tx validation
+            TxExecutionMode::VerifyExecute => usize::MAX,
+            TxExecutionMode::EthCall | TxExecutionMode::EstimateFee => {
+                self.missed_storage_invocation_limit
+            }
+        };
+
         tokio::task::spawn_blocking(move || {
-            let tracers =
-                VmSandbox::wrap_tracers(tracers, &env, args.missed_storage_invocation_limit);
+            let tracers = VmSandbox::wrap_tracers(tracers, &env, missed_storage_invocation_limit);
             let executor = VmSandbox::new(storage, env, args);
             executor.apply(|vm, transaction| {
                 vm.push_transaction(transaction);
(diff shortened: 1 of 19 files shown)


core/node/api_server/src/execution_sandbox/mod.rs
```diff
@@ -417,3 +417,28 @@ impl BlockArgs {
         )
     }
 }
+
+/// VM executor capable of executing isolated transactions / calls (as opposed to batch execution).
+#[async_trait]
+trait OneshotExecutor<S: ReadStorage> {
+    type Tracers: Default;
```
slowli, line 424: I don't like this associated type very much; probably would need to get rid of it in the future refactoring iterations (see the overall comment) in favor of something like

```rust
pub struct Tracers<'a> {
    pub calls: Option<&'a mut Vec<Call>>,
    pub validation: Option<(
        ValidationTracerParams,
        &'a mut Result<(), ViolatedValidationRule>,
    )>,
}
```

core/node/api_server/src/execution_sandbox/mod.rs
```diff
@@ -417,3 +417,28 @@ impl BlockArgs {
...
+
+    async fn inspect_transaction(
+        &self,
+        storage: S,
+        env: OneshotEnv,
+        args: TxExecutionArgs,
+        tracers: Self::Tracers,
+    ) -> anyhow::Result<VmExecutionResultAndLogs>;
+
+    async fn inspect_transaction_with_bytecode_compression(
```
slowli, line 434: I've recreated the logic previously used in the sandbox:

- Validation uses `inspect_transaction()` (which translates to `VmInterface::push_transaction` + `inspect()`)
- Execution uses `inspect_transaction_with_bytecode_compression` (which translates to `VmInterface::inspect_transaction_with_bytecode_compression`).

Not sure whether this can / should be simplified to a single method.

core/node/api_server/src/execution_sandbox/apply.rs
slowli, line 1: This module naming / structuring makes less sense now; I haven't touched it that much to keep the change set smaller.

core/node/api_server/src/execution_sandbox/apply.rs
```diff
@@ -37,179 +40,250 @@ use zksync_types::{
...
+    let next_block = if is_pending_block {
+        L2BlockEnv {
+            number: next_block.number + 1,
+            timestamp: resolved_block_info.l1_batch_timestamp,
+            prev_block_hash: resolved_block_info.state_l2_block_hash,
+            // For simplicity, we assume each L2 block create one virtual block.
+            // This may be wrong only during transition period.
+            max_virtual_blocks_to_create: 1,
+        }
+    } else if next_block.number == 0 {
```
slowli, line 118: IIUC, this check has a false positive if the use has actually requested the genesis block. This wasn't covered by tests previously, but now with mock oneshot executor being more accurate, it was uncovered by VM tests. I'm not sure how to fix this, so I've hacked the tests instead by creating an additional block 🙃

core/node/api_server/src/execution_sandbox/apply.rs
```diff
@@ -37,179 +40,250 @@ use zksync_types::{
...
-                prev_block_hash: prev_l2_block_info.l2_block_hash,
-                max_virtual_blocks_to_create: 1,
-            }
+        let mut prev_block_hash = connection
+            .blocks_web3_dal()
+            .get_l2_block_hash(prev_block_number)
+            .await
+            .map_err(DalError::generalize)?;
+        if prev_block_hash.is_none() {
+            // We might need to load the previous block hash from the snapshot recovery metadata
```
slowli, line 143: This was incorrectly not done previously, which led to the RPC method erroring if asked the first retained L2 block. Again, this wasn't covered by tests because the mock oneshot executor returned before this code has a chance to run.

core/node/api_server/src/execution_sandbox/execute.rs
```diff
@@ -1,80 +1,80 @@
...
-use crate::execution_sandbox::api::state_override::StateOverride;
 
+/// Executor-independent arguments necessary to for oneshot transaction execution.
+///
+/// # Developer guidelines
+///
+/// Please don't add fields that duplicate `SystemEnv` or `L1BatchEnv` information, since both of these
+/// are also provided to an executor.
 #[derive(Debug)]
 pub(crate) struct TxExecutionArgs {
```
slowli, line 28: Not 100% sure this grouping makes sense, but it sort of does IMO; these params are related specifically to oneshot execution.

slowli (commented):
Some general notes:

- I don't think it would be reasonable to attempt uniting `OneshotExecutor` and `BatchExecutor`; IMO, they have sufficiently distinct semantics to be kept as separate traits.
- I have left `OneshotExecutor` in the API server crate for now in order to keep PR scope reasonable. It may make sense to put `OneshotExecutor` / `BatchExecutor` implementations in a separate crate (which would also contain `vm_utils` code), maybe `vm_executor`. The `OneshotExecutor` / `BatchExecutor` traits could be put there or in `vm_interface`.

core/node/api_server/src/execution_sandbox/apply.rs
popzxc: I think it's fine for now.

core/node/api_server/src/execution_sandbox/mod.rs
popzxc: I think it's fine to postpone this decision.

core/node/api_server/src/execution_sandbox/validate.rs
```diff
@@ -31,88 +31,46 @@ pub(crate) enum ValidationError {
 }
 
 impl TransactionExecutor {
+    #[tracing::instrument(level = "debug", skip_all)]
     pub(crate) async fn validate_tx_in_sandbox(
         &self,
-        connection_pool: ConnectionPool<Core>,
+        mut connection: Connection<'static, Core>,
```
popzxc, line 37: 🔥 🔥 🔥

core/node/api_server/src/tx_sender/mod.rs
```diff
@@ -368,23 +372,20 @@ impl TxSender {
         stage_latency.observe();
 
         let stage_latency = SANDBOX_METRICS.start_tx_submit_stage(tx_hash, SubmitTxStage::DryRun);
-        let shared_args = self.shared_args().await?;
+        let setup_args = self.call_args(&tx, None).await?;
         let vm_permit = self.0.vm_concurrency_limiter.acquire().await;
         let vm_permit = vm_permit.ok_or(SubmitTxError::ServerShuttingDown)?;
         let mut connection = self.acquire_replica_connection().await?;
         let block_args = BlockArgs::pending(&mut connection).await?;
-        drop(connection);
```
popzxc, line 376: This might improve the latency during high load.

popzxc (approved):

PR #2673, opened 2024-08-16
fix(api_server): fix blob_gas length

tomg10:
Discovered during testing Gateway integration (our own asserts from query.rs started being triggered by our eth api implementation :D )


core/node/api_server/src/web3/namespaces/eth.rs
```diff
@@ -689,6 +689,8 @@ impl EthNamespace {
         effective_pubdata_price_history.reverse();
 
         let oldest_block = newest_l2_block.0 + 1 - base_fee_per_gas.len() as u32;
+        // `base_fee_per_gas` for next L2 block cannot be calculated, appending last fee as a placeholder.
+        base_fee_per_gas.push(*base_fee_per_gas.last().unwrap());
```
perekopskiy, line 693: Nice catch! However, length of `gas_used_ratio `, `reward` and `blob_gas_used_ratio` should left untouched as it depends on `base_fee_per_gas.len()`.
Separately, what about length of `l2_pubdata_price`? I think we shouldn't change it, i.e. don't include "next" value because it's zksync extension and we don't need to follow any eth spec here. But I'm curious what do you think about it

perekopskiy: also, please add an assert for each array len in the test

core/node/api_server/src/web3/namespaces/eth.rs
```diff
@@ -689,6 +689,8 @@ impl EthNamespace {
         effective_pubdata_price_history.reverse();
 
         let oldest_block = newest_l2_block.0 + 1 - base_fee_per_gas.len() as u32;
+        // `base_fee_per_gas` for next L2 block cannot be calculated, appending last fee as a placeholder.
```
slowli, line 692: Drive-by off-topic (?) question: Why can't it be calculated? Isn't this only the case if `newest_l2_block` is the latest one? Otherwise, the next block info can be requested from the DB.

slowli: Aha, so because such calculation is impossible, we prefer to return a dummy value for the next block instead of loading it from the DB? IIUC, Ethereum Web3 providers return the next block value when queried non-latest blocks, but they *are* calculated per EIP1559.

perekopskiy: I don't have a strong opinion on this. Anyway users should not rely on `next_base_fee`

perekopskiy (approved):

PR #2674, opened 2024-08-16
fix(proof_data_handler): TEE blob fetching error handling

pbeza:
We ran into a problem in the staging environment where TEE blob fetching failed because of a 30-day retention policy on blobs in Google Cloud Storage. The TEE prover was failing for all old batches (`l1_batch_number < 58300`). This commit fixes the issue by adding better error handling when the blob for a given batch number isn't available.

## What ❔

Graceful error handling for the TEE proof data handler when there is no blob in Google Cloud Storage for the specified batch number.

## Why ❔

We need more robust error handling.

## Checklist


pbeza:
Tested. It seems to be working. ✅

Not sure who to ask for a review while @popzxc is on vacation. I used to ask him for all my `zksync-era` PRs. Any suggestions? @RomanBrodetski @EmilLuta @slowli

core/node/proof_data_handler/src/tee_request_processor.rs
```diff
@@ -47,26 +47,62 @@ impl TeeRequestProcessor {
...
+            match self.get_blob(l1_batch_number).await {
+                Ok(input) => {
+                    return Ok(Json(TeeProofGenerationDataResponse(Some(Box::new(input)))));
+                }
+                Err(ObjectStoreError::KeyNotFound(_)) => {
+                    tracing::warn!(
+                        "Blob for batch number {} has not been found in the object store. Marking the job as skipped.",
+                        l1_batch_number
+                    );
+                    connection
```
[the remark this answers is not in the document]
pbeza: > In general, skipped has different semantics in our codebase (skipped on purpose, not because input was missing).

I'm fine with renaming it however you like. We just need to make sure we can gracefully handle cases where the corresponding blob for the current batch is missing. We've already had an issue on stage where batches had no corresponding GCP blob, causing the TEE prover to panic.

> I'm not in the loop for this project, but why would TEE try to prove such old proofs? Where does it start? Does it make sense to waste compute on jobs that we know are not valuable at all?

We currently compute TEE proofs for all the batches in the database, starting from the very beginning. Maybe we should consider starting from a preconfigured batch instead of the oldest one.

Here’s how it works right now:
1. We have a [`tee-prover`](https://github.com/matter-labs/zksync-era/blob/main/core/bin/zksync_tee_prover/src/tee_prover.rs) standalone binary that runs an [infinite loop](https://github.com/matter-labs/zksync-era/blob/56d8ee8c0546cc26d412b95cb72bbb1b9a3a6580/core/bin/zksync_tee_prover/src/tee_prover.rs#L178-L218), pulling inputs for proving the next batches from the [`data_proof_handler`](https://github.com/matter-labs/zksync-era/blob/main/core/node/proof_data_handler/src/lib.rs).
1. The `data_proof_handler` [gives](https://github.com/matter-labs/zksync-era/blob/56d8ee8c0546cc26d412b95cb72bbb1b9a3a6580/core/lib/dal/src/tee_proof_generation_dal.rs#L21-L59) the inputs for the next batch to be proven.
1. The `tee-prover` [computes](https://github.com/matter-labs/zksync-era/blob/56d8ee8c0546cc26d412b95cb72bbb1b9a3a6580/core/bin/zksync_tee_prover/src/tee_prover.rs#L117) a TEE proof for the fetched inputs and [sends it back](https://github.com/matter-labs/zksync-era/blob/56d8ee8c0546cc26d412b95cb72bbb1b9a3a6580/core/bin/zksync_tee_prover/src/tee_prover.rs#L118-L125) to the proof handler to be saved in the database.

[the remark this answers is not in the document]
pbeza: > This sounds very sensible. Simply mark everything up to some point (manually, DB update) as skipped and then let the process flow by itself.

I’d really prefer to automate as much as possible. This PR adds automatic skipping. The other option would be manually running an `UPDATE` SQL query after each new DB re-deployment:
```
UPDATE tee_proof_generation_details SET status = 'skipped' WHERE l1_batch_number < 58300 ;
```
(this is the exact SQL query we ran manually when we realized that all batches `< 58300` don’t have matching blobs in GCP)

Whether we decide to manually skip batches below a certain number or not, we still need an easy way to handle missing blobs in GCP for any batch, old or new. This PR offers a way to do that.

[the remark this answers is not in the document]
pbeza: > due to some software bug or changes on retention policy -- we intend to drive that closer to 2 days, if your TEE is always 2 days behind, you'll skip all proofs

Good point too. 🤔

I don't have a strong opinion on how we should handle missing GCP blobs. This PR implements just one of many possible strategies to address this scenario. I'll wait for more review comments to see which approach is most preferable.

This PR isn't super urgent since we're currently working around the issue by manually running [the above SQL query](https://github.com/matter-labs/zksync-era/pull/2674#discussion_r1721896574). CC: @RomanBrodetski, @thomasknauth (and @popzxc, @haraldh when they return from vacation)

core/node/proof_data_handler/src/tee_request_processor.rs
```diff
@@ -47,26 +47,62 @@ impl TeeRequestProcessor {
...
 
-        let response = TeeProofGenerationDataResponse(Some(Box::new(tee_verifier_input)));
+    async fn get_blob(
+        &self,
+        l1_batch_number: L1BatchNumber,
+    ) -> Result<TeeVerifierInput, ObjectStoreError> {
+        let max_blob_store_retries = 3;
+        let mut last_err: Option<ObjectStoreError> = None;
+
+        for _ in 0..max_blob_store_retries {
```
slowli, line 92: There's retries logic [embedded](https://github.com/matter-labs/zksync-era/blob/edfcc7dbb7fb60f0f42fff4f3d350974128127b4/core/lib/object_store/src/retries.rs#L27) into `ObjectStore` implementations; is it insufficient? The only way when a retriable error is returned from `get` if these internal retries are exhausted; so wouldn't it be easier to increase the number of retries 3x in the object store config?

core/node/proof_data_handler/src/tee_request_processor.rs
```diff
@@ -47,26 +47,62 @@ impl TeeRequestProcessor {
...
+                    return Ok(Json(TeeProofGenerationDataResponse(Some(Box::new(input)))));
+                }
+                Err(ObjectStoreError::KeyNotFound(_)) => {
+                    tracing::warn!(
+                        "Blob for batch number {} has not been found in the object store. Marking the job as skipped.",
+                        l1_batch_number
+                    );
+                    connection
+                        .tee_proof_generation_dal()
+                        .mark_proof_generation_job_as_skipped(l1_batch_number, request.tee_type)
```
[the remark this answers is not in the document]
pbeza: > I don't think mutating DB in a logically immutable / idempotent request handler is a good idea (...)

We can rename the request handler to show that it’s not an immutable handler, just like how @popzxc recently changed `get_next_block_to_be_proven` to `lock_batch_for_proving` [here](https://github.com/matter-labs/zksync-era/commit/7c336b1e180b9d5ba1ba74169c61ce27a251e2fc#diff-020dc10234b386f7a7c963d733c86f52e229a1dc0dbccc6300c40d8295dcbfacL33-L44). :P

On a more serious note, you’re probably right – I can add an optional minimum batch number as a parameter for `get_next_batch_to_be_proven`. The tricky part is figuring out how to handle it when we call `get_next_batch_to_be_proven` a few times, and the GCS blob isn’t available, leaving those batches marked as `picked_by_prover`. I could reset their status to ready_to_be_proven in that case, but that doesn't seem like an elegant solution either.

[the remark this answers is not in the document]
pbeza: > (...) I could reset their status to ready_to_be_proven in that case, but that doesn't seem like an elegant solution either.

Oh, maybe that's not that bad to reset the status `unpicked` if @popzxc implemented similar approach [here](https://github.com/matter-labs/zksync-era/commit/7c336b1e180b9d5ba1ba74169c61ce27a251e2fc#diff-020dc10234b386f7a7c963d733c86f52e229a1dc0dbccc6300c40d8295dcbfacR91-R113). I think I should cherry-pick his PR and apply to the largely copy-pasted [`tee_proof_generation_dal.rs`](https://github.com/matter-labs/zksync-era/blob/main/core/lib/dal/src/tee_proof_generation_dal.rs).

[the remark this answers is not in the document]
pbeza: @slowli I just addressed your code review comments and made sure it lines up with the recent refactor Igor [did](https://github.com/matter-labs/zksync-era/pull/2486) in `proof_generation_dal.rs` and `request_processor.rs`. Give it a look when you get a chance!

core/lib/dal/src/tee_proof_generation_dal.rs
```diff
@@ -17,14 +19,23 @@ pub struct TeeProofGenerationDal<'a, 'c> {
...
-    pub async fn get_next_batch_to_be_proven(
+    pub async fn lock_batch_for_proving(
         &mut self,
         tee_type: TeeType,
         processing_timeout: Duration,
+        min_batch_number: Option<L1BatchNumber>,
     ) -> DalResult<Option<L1BatchNumber>> {
         let processing_timeout = pg_interval_from_duration(processing_timeout);
-        let query = sqlx::query!(
+        let min_batch_condition = min_batch_number.map_or(String::new(), |min_batch| {
```
slowli, line 35: AFAIU, it's possible to use [`match_query_as`](https://github.com/matter-labs/zksync-era/blob/87768755e8653e4be5f29945b56fd05a5246d5a8/core/lib/db_connection/src/macro_utils.rs#L84) macro here; you can see some examples of usage in the DAL crate (e.g., [here](https://github.com/matter-labs/zksync-era/blob/87768755e8653e4be5f29945b56fd05a5246d5a8/core/lib/dal/src/blocks_dal.rs#L1969)). This would allow to check the query in compilation time. In this particular case, there's a much simpler solution though: just use 0 as the default value if the batch number isn't supplied.

core/node/proof_data_handler/src/tee_request_processor.rs
```diff
@@ -41,32 +38,71 @@ impl TeeRequestProcessor {
...
+                    return Ok(Json(TeeProofGenerationDataResponse(Some(Box::new(input)))));
+                }
+                Err(ObjectStoreError::KeyNotFound(_)) => {
+                    tracing::warn!(
+                        "Blob for batch number {} has not been found in the object store.",
+                        l1_batch_number
+                    );
+                    self.unlock_batch(l1_batch_number, request.tee_type).await?;
+                    min_batch_number =
+                        min_batch_number.map_or(Some(l1_batch_number + 1), |num| Some(num + 1));
```
slowli, line 63: Nit: You can simplify the RHS to `Some(min_batch_number.unwrap_or(l1_batch_number) + 1)`.

core/node/proof_data_handler/src/tee_request_processor.rs
```diff
@@ -41,32 +38,71 @@ impl TeeRequestProcessor {
...
+                min_batch_number,
+            )
             .await
-            .map_err(RequestProcessorError::Dal)?;
-
-        let l1_batch_number = match l1_batch_number_result {
-            Some(number) => number,
-            None => return Ok(Json(TeeProofGenerationDataResponse(None))),
-        };
+            .map_err(RequestProcessorError::Dal)
```
slowli, line 90: Nit / off-topic (?): You can implement `From<DalError> for RequestProcessorError`, which would allow to skip such implicit conversions; `?` converts the error type automatically if possible.

core/node/proof_data_handler/src/tee_request_processor.rs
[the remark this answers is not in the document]
slowli: I think it's fine; IMO, a background task would be better, but it'd require more time to implement / test. Also, I've made a mistake calling this API method "immutable / idempotent"; it's not (it uses the POST verb).

core/lib/dal/src/tee_proof_generation_dal.rs
pbeza: [Fixed](https://github.com/matter-labs/zksync-era/pull/2674/commits/63e7ada3b727ad1f61cbb70155a3e6972dbf7901). ✅

core/node/proof_data_handler/src/tee_request_processor.rs
pbeza: Good point, [fixed](https://github.com/matter-labs/zksync-era/pull/2674/commits/ef3395481a17618bd995e452b547f7faa7d9b6ab). ✅

slowli (approved):
Parts that I understand look good, but since I'm not a domain expert, I'd suggest to wait for another approval 🙃

core/node/proof_data_handler/src/tee_request_processor.rs
```diff
@@ -41,32 +38,77 @@ impl TeeRequestProcessor {
...
-            .tee_proof_generation_dal()
-            .get_next_batch_to_be_proven(request.tee_type, self.config.proof_generation_timeout())
-            .await
-            .map_err(RequestProcessorError::Dal)?;
-
-        let l1_batch_number = match l1_batch_number_result {
-            Some(number) => number,
-            None => return Ok(Json(TeeProofGenerationDataResponse(None))),
+        let mut min_batch_number: Option<L1BatchNumber> = None;
+        let mut missing_range: Option<(L1BatchNumber, L1BatchNumber)> = None;
```
thomasknauth, line 42: Do you think the additional logic to handle the range scenario is worth it? I'd just log a message if the blob for some batch number is missing w/o trying to keep track of ranges. The only reason I can think of why there are gaps is when we have a situation like now, where we deploy part of the code, accumulate lots of "to_be_proven" batches and only later activate the logic to actually produce a proof. In normal operation, there shouldn't really be any gap and/or missing blob. Not sure.

thomasknauth (approved):

core/node/proof_data_handler/src/tee_request_processor.rs
pbeza: > (...) The only reason I can think of why there are gaps is when we have a situation like now, where we deploy part of the code, accumulate lots of "to_be_proven" batches and only later activate the logic to actually produce a proof. (...)

That reason alone is enough for me to want this logic implemented. I can also imagine a scenario where `proof_data_handler` module temporarily loses access to GCP for whatever reason (like authorization issues, GCP being down, or something else). In that case, we’d get flooded with logs.

Anyway, I’m not strongly attached to this. It’s just an implementation detail that I can easily roll back with [`git revert 35cb71c86781ac8aa3d4d5c7a2ea096b30befaa9`](https://github.com/matter-labs/zksync-era/pull/2674/commits/35cb71c86781ac8aa3d4d5c7a2ea096b30befaa9).

Any strong opinions on this, @slowli? We need a tie-breaker here. I’ll go with whatever suggestion you have, just to get it merged today. :P

pbeza:
@thomasknauth / @haraldh / @slowli, could you merge this? I don't have the permissions. I think 2 approvals ✅✅ should be good enough. Thanks!

core/node/proof_data_handler/src/tee_request_processor.rs
slowli: IMO, it's reasonable to avoid log spamming if it doesn't introduce much complexity.

PR #2675, opened 2024-08-16
feat(zk_toolbox): Run formatters and linterrs

Deniallugo:
## What ❔

Adding an ability to run linters and formatters for zk supervisor

## Checklist


zk_toolbox/crates/zk_supervisor/src/main.rs
```diff
@@ -42,6 +42,8 @@ enum SupervisorSubcommands {
     Snapshot(SnapshotCommands),
     #[command(about = MSG_SUBCOMMAND_LINT_ABOUT, alias = "l")]
     Lint(LintArgs),
+    #[command(about = MSG_SUBCOMMAND_LINT_ABOUT)]
```
matias-gonz, line 45: This is using the same message as lint subcommand

zk_toolbox/crates/zk_supervisor/src/commands/fmt.rs
```diff
@@ -0,0 +1,117 @@
+use std::path::PathBuf;
```
sanekmelnikov, line 1: I store ecosystem folder inside zksync-era (e.g. `zksync-era/my-ecosystem`) and when I run `zk fmt` - everything works, but `zk_supervisor fmt` fails due to access to docker volumes
probably my use case isn't that common, but just letting you know

zk_toolbox/crates/zk_supervisor/src/messages.rs
```diff
@@ -152,5 +156,18 @@ pub(super) fn msg_running_linter_for_extension_spinner(extension: &Extension) ->
     format!("Running linter for files with extension: .{}", extension)
 }
 
+pub(super) fn msg_running_fmt_for_extension_spinner(extension: Extension) -> String {
```
sanekmelnikov, line 159: actually, i see this message twice (sometimes), is it ok?
Screenshot 2024-08-19 at 6 54 59 PM

zk_toolbox/crates/zk_supervisor/src/commands/fmt.rs
Deniallugo: ouch, that is the reason of old obscure code...

Deniallugo: @sanekmelnikov could you try one more time please? 
seems i've fixed but it wasn't reproduced

matias-gonz (approved):

zk_toolbox/crates/zk_supervisor/src/commands/fmt.rs
sanekmelnikov: discussed offline, still not working - need to get rid of glob, mentioned [here](https://github.com/matter-labs/zksync-era/blob/main/etc/utils/src/index.ts#L128-L134)

sanekmelnikov (approved):

PR #2685, opened 2024-08-19
feat(eth-sender): added chain_id column to eth_txs + support for gateway in tx_aggregator

tomg10:


core/node/node_framework/src/implementations/layers/eth_sender/aggregator.rs
```diff
@@ -93,7 +97,13 @@ impl WiringLayer for EthTxAggregatorLayer {
         let master_pool = input.master_pool.get().await.unwrap();
         let replica_pool = input.replica_pool.get().await.unwrap();
 
-        let eth_client = input.eth_client.0;
+        let (eth_client, settlement_mode) = match (input.eth_client, input.eth_client_l2) {
+            (Some(l1_client), None) => (l1_client.0, SettlementMode::SettlesToL1),
+            (None, Some(l2_client)) => (l2_client.0, SettlementMode::Gateway),
+            (_, _) => Err(WiringError::Configuration(
+                "Aggregator requires either l1 or l2 client to operatate".to_string(),
```
StanislavBreadless, line 104: what happens if both of those are provided?

StanislavBreadless, line 104:
```suggestion
                "Aggregator requires either l1 or l2 client to operate".to_string(),
```

core/node/eth_sender/src/eth_tx_aggregator.rs
```diff
@@ -592,9 +607,10 @@ impl EthTxAggregator {
         storage: &mut Connection<'_, Core>,
         from_addr: Option<Address>,
     ) -> Result<u64, EthSenderError> {
+        let is_gateway = self.settlement_mode == SettlementMode::Gateway;
```
StanislavBreadless, line 610: A getter function is already present in the enum itself

core/node/node_framework/src/implementations/layers/eth_sender/aggregator.rs
tomg10: I've changed for not so that it relies on gas_adjuster settlement_mode for which client to chose.

StanislavBreadless (approved):

PR #2688, opened 2024-08-20
feat: add `update-token-multiplier-setter` command

ischasny:
Add a top level command `zk_inception chain update-token-multiplier-setter`. It allows to invoke "IChainAdmin.setTokenMultiplierSetter` which will be useful for the chains with a custom base token that hasn't had that role before.

### DIFF
--- a/zk_toolbox/crates/zk_inception/src/commands/chain/mod.rs
+++ b/zk_toolbox/crates/zk_inception/src/commands/chain/mod.rs
@@ -15,6 +15,7 @@ pub mod deploy_l2_contracts;
 pub mod deploy_paymaster;
 pub mod genesis;
 pub(crate) mod init;
+mod set_token_multiplier_setter;
 
 #[derive(Subcommand, Debug)]
 pub enum ChainCommands {
@@ -35,6 +36,9 @@ pub enum ChainCommands {
     /// Deploy paymaster smart contract
     #[command(alias = "paymaster")]
     DeployPaymaster(ForgeScriptArgs),
+    /// Update Token Multiplier Setter address on L1
+    #[command(alias = "updateTokenMultiplierSetter")]
+    UpdateTokenMultiplierSetter(ForgeScriptArgs),
 }
 
 pub(crate) async fn run(shell: &Shell, args: ChainCommands) -> anyhow::Result<()> {
@@ -52,5 +56,8 @@ pub(crate) async fn run(shell: &Shell, args: ChainCommands) -> anyhow::Result<()
             deploy_l2_contracts::run(args, shell, Deploy2ContractsOption::IntiailizeBridges).await
         }
         ChainCommands::DeployPaymaster(args) => deploy_paymaster::run(args, shell).await,
+        ChainCommands::UpdateTokenMultiplierSetter(args) => {
+            set_token_multiplier_setter::run(args, shell).await
+        }
     }
 }
--- a/zk_toolbox/crates/zk_inception/src/commands/chain/mod.rs
+++ b/zk_toolbox/crates/zk_inception/src/commands/chain/mod.rs
@@ -37,7 +37,6 @@ pub enum ChainCommands {
     #[command(alias = "paymaster")]
     DeployPaymaster(ForgeScriptArgs),
     /// Update Token Multiplier Setter address on L1
-    #[command(alias = "updateTokenMultiplierSetter")]
     UpdateTokenMultiplierSetter(ForgeScriptArgs),
 }
 
--- a/zk_toolbox/crates/zk_inception/src/commands/chain/set_token_multiplier_setter.rs
+++ b/zk_toolbox/crates/zk_inception/src/commands/chain/set_token_multiplier_setter.rs
@@ -1,0 +1,115 @@
+use anyhow::Context;
+use common::{
+    config::global_config,
+    forge::{Forge, ForgeScript, ForgeScriptArgs},
+    logger,
+    spinner::Spinner,
+};
+use config::{forge_interface::script_params::ACCEPT_GOVERNANCE_SCRIPT_PARAMS, EcosystemConfig};
+use ethers::{abi::parse_abi, contract::BaseContract, utils::hex};
+use lazy_static::lazy_static;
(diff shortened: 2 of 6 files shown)


zk_toolbox/crates/zk_inception/src/commands/chain/set_token_multiplier_setter.rs
```diff
@@ -0,0 +1,115 @@
...
+    let l1_url = chain_config
+        .get_secrets_config()?
+        .l1
+        .context(MSG_L1_SECRETS_MUST_BE_PRESENTED)?
+        .l1_rpc_url
+        .expose_str()
+        .to_string();
+    let token_multiplier_setter_address = ecosystem_config
+        .get_wallets()
+        .unwrap()
```
montekki, line 47: this can be a meaningful error, the function returns `anyhow::Result<()>` anyway

zk_toolbox/crates/zk_inception/src/commands/chain/set_token_multiplier_setter.rs
```diff
@@ -0,0 +1,115 @@
...
+    // code assumes that doing only once is enough, but actually we need to accept admin multiple times
+    let mut forge_args = forge_args.clone();
+    forge_args.resume = false;
+
+    let calldata = SET_TOKEN_MULTIPLIER_SETTER
+        .encode(
+            "chainSetTokenMultiplierSetter",
+            (chain_admin_address, target_address),
+        )
+        .unwrap();
```
montekki, line 92: or if you are sure that this has to cause panic, `.expect` with a message would be  bit more useful to the user

zk_toolbox/crates/zk_inception/src/commands/chain/set_token_multiplier_setter.rs
```diff
@@ -0,0 +1,115 @@
...
+    governor: Option<H256>,
+    chain_admin_address: Address,
+    target_address: Address,
+    forge_args: &ForgeScriptArgs,
+    l1_rpc_url: String,
+) -> anyhow::Result<()> {
+    // Resume for accept admin doesn't work properly. Foundry assumes that if signature of the function is the same,
+    // then it's the same call, but because we are calling this function multiple times during the init process,
+    // code assumes that doing only once is enough, but actually we need to accept admin multiple times
+    let mut forge_args = forge_args.clone();
```
montekki, line 84: If you are cloning them anyway that would probably make sense to take a parameter by value, not by reference.

zk_toolbox/crates/zk_inception/src/commands/chain/set_token_multiplier_setter.rs
```diff
@@ -0,0 +1,115 @@
...
+    forge_args.resume = false;
+
+    let calldata = SET_TOKEN_MULTIPLIER_SETTER
+        .encode(
+            "chainSetTokenMultiplierSetter",
+            (chain_admin_address, target_address),
+        )
```
montekki, line 91: I think you can actually do this in a typesafe manner with `abigen`?

Deniallugo: I've not found a way how to encode the call, not make the call in a typesafe manner

zk_toolbox/crates/zk_inception/src/commands/chain/set_token_multiplier_setter.rs
ischasny: good point - fixed.

zk_toolbox/crates/zk_inception/src/commands/chain/mod.rs
```diff
@@ -35,6 +36,9 @@ pub enum ChainCommands {
     /// Deploy paymaster smart contract
     #[command(alias = "paymaster")]
     DeployPaymaster(ForgeScriptArgs),
+    /// Update Token Multiplier Setter address on L1
+    #[command(alias = "updateTokenMultiplierSetter")]
```
Deniallugo, line 40: you don't need to specify an alias i n this way

zk_toolbox/crates/zk_inception/src/commands/chain/set_token_multiplier_setter.rs
ischasny: I.e. mutable to avoid cloning a few times?

zk_toolbox/crates/zk_inception/src/commands/chain/mod.rs
ischasny: What's the best way to do it then?

zk_toolbox/crates/zk_inception/src/commands/chain/set_token_multiplier_setter.rs
ischasny: If it fails here - that means that the tool is broken internally (i.e. contract abi doesn't have the function that it should have). I'd argue that the message won't be too useful to the customer. zk_toolbox also has a cool feature - if an internal error occurs, it prints a pretty message and gives a link to the logs where the user can see what has failed if they want to.

zk_toolbox/crates/zk_inception/src/commands/chain/set_token_multiplier_setter.rs
montekki: So you usually want to be passing things by reference in Rust to kind of signal that the function does only need this object by reference and does not need to do any cloning of it. If you actually need a copy of it, better to make that function argument by value so from the function signature it would be immediately obvious what's going on if that makes sense.

Deniallugo: it will be a clone from a reference anyway. 
or you have to change a lot of signatures.

zk_toolbox/crates/zk_inception/src/commands/chain/mod.rs
ischasny: makes sense - removed

PR #2690, opened 2024-08-20
fix: use lower fair l2 gas price for cbt

ischasny:
this PR reduces fair l2 gas price for chains that use custom base token. This is needed because CBT quote might fluctuate a lot for volatile tokens. By the time a transaction reaches the server, it could have already changed. Such transactions might fail validation without even getting into the mempool.

### DIFF
--- a/core/node/api_server/src/tx_sender/mod.rs
+++ b/core/node/api_server/src/tx_sender/mod.rs
@@ -515,17 +515,6 @@ impl TxSender {
             .get_batch_fee_input()
             .await?;
 
-        // If the chain uses a custom base token (CBT), we need to reduce fair l2 gas price.
-        // This is because CBT quote fluctuates independently of the l1 gas price.
-        // Let's say when the transaction was created the CBT/ETH ratio was 1/10. By the time the transaction
-        // has reached server it got to 2/10. As the gas price was estimated using the former quote and fair l2 gas price
-        // using the latter, the transaction would fail validation without even getting to the mempool.
-        // We want to avoid mass validation failures for volatile tokens and let them into mempool under more relaxed conditions.
-        let mut fair_l2_gas_price = fee_input.fair_l2_gas_price();
-        if !self.0.sender_config.is_eth_based_chain {
-            fair_l2_gas_price /= 2;
-        }
-
         // TODO (SMA-1715): do not subsidize the overhead for the transaction
 
         if tx.common_data.fee.gas_limit > self.0.sender_config.max_allowed_l2_tx_gas_limit.into() {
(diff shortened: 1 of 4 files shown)


core/node/api_server/src/tx_sender/mod.rs
```diff
@@ -507,6 +515,17 @@ impl TxSender {
...
+        // If the chain uses a custom base token (CBT), we need to reduce fair l2 gas price.
+        // This is because CBT quote fluctuates independently of the l1 gas price.
+        // Let's say when the transaction was created the CBT/ETH ratio was 1/10. By the time the transaction
+        // has reached server it got to 2/10. As the gas price was estimated using the former quote and fair l2 gas price
+        // using the latter, the transaction would fail validation without even getting to the mempool.
+        // We want to avoid mass validation failures for volatile tokens and let them into mempool under more relaxed conditions.
+        let mut fair_l2_gas_price = fee_input.fair_l2_gas_price();
+        if !self.0.sender_config.is_eth_based_chain {
+            fair_l2_gas_price /= 2;
+        }
```
perekopskiy, line 527: sorry for miscommunication, I meant that there is no need for `is_eth_based_chain` flag, IMO it's better to keep code simpler and divide by 2 unconditionally

core/node/api_server/src/tx_sender/mod.rs
```diff
@@ -507,6 +515,17 @@ impl TxSender {
             .get_batch_fee_input()
             .await?;
 
+        // If the chain uses a custom base token (CBT), we need to reduce fair l2 gas price.
+        // This is because CBT quote fluctuates independently of the l1 gas price.
+        // Let's say when the transaction was created the CBT/ETH ratio was 1/10. By the time the transaction
+        // has reached server it got to 2/10. As the gas price was estimated using the former quote and fair l2 gas price
+        // using the latter, the transaction would fail validation without even getting to the mempool.
+        // We want to avoid mass validation failures for volatile tokens and let them into mempool under more relaxed conditions.
+        let mut fair_l2_gas_price = fee_input.fair_l2_gas_price();
```
perekopskiy, line 524: nit: you can divide in if statement, no need for additional mutable var.

core/node/api_server/src/tx_sender/mod.rs
ischasny: Would it not let more unexecutable transactions into the mempool for ETH-based chains? If there we do have a way of cleaning those up - then ok, otherwise why would we want to do that?

perekopskiy: Yes, we clean those periodically. It's basically a trade-off between how many txs we allow into mempool and complexity of filter rule. I would drop the `is_eth_based_chain` and allow _some_ more txs into ETH-based chains' mempool

ischasny: Ok amazing - fixing it then.

perekopskiy (approved):

PR #2696, opened 2024-08-20
fix: make set token multiplier optional

ischasny:
For backwards compatibility make setting token multiplier on L1 optional. If private key for token multiplier setter is not provided, L1 quote update will be skipped.

### DIFF
--- a/core/node/base_token_adjuster/src/base_token_ratio_persister.rs
+++ b/core/node/base_token_adjuster/src/base_token_ratio_persister.rs
@@ -3,7 +3,6 @@ use std::{cmp::max, fmt::Debug, sync::Arc, time::Duration};
 use anyhow::Context as _;
 use tokio::{sync::watch, time::sleep};
 use zksync_config::configs::base_token_adjuster::BaseTokenAdjusterConfig;
-use zksync_contracts::chain_admin_contract;
 use zksync_dal::{ConnectionPool, Core, CoreDal};
 use zksync_eth_client::{BoundEthInterface, Options};
 use zksync_external_price_api::PriceAPIClient;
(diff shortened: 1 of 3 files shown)


core/node/base_token_adjuster/src/base_token_ratio_persister.rs
```diff
@@ -90,53 +83,63 @@ impl BaseTokenRatioPersister {
         let new_ratio = self.retry_fetch_ratio().await?;
         self.persist_ratio(new_ratio).await?;
 
-        let max_attempts = self.config.l1_tx_sending_max_attempts;
-        let sleep_duration = self.config.l1_tx_sending_sleep_duration();
-        let mut result: anyhow::Result<()> = Ok(());
-        let mut prev_base_fee_per_gas: Option<u64> = None;
-        let mut prev_priority_fee_per_gas: Option<u64> = None;
+        if let Some(l1_params) = &self.l1_params {
```
perekopskiy, line 86: nit:
`let Some(l1_params) = &self.l1_params else { return Ok(()) }`
will look nicely

core/node/node_framework/src/implementations/layers/base_token/base_token_ratio_persister.rs
```diff
@@ -81,37 +82,37 @@ impl WiringLayer for BaseTokenRatioPersisterLayer {
...
-        let token_multiplier_setter_wallet = self
-            .wallets_config
-            .token_multiplier_setter
-            .expect("base token adjuster wallet is not set")
-            .wallet;
 
-        let tms_private_key = token_multiplier_setter_wallet.private_key();
-        let tms_address = token_multiplier_setter_wallet.address();
-        let EthInterfaceResource(query_client) = input.eth_client;
+        let mut l1_params: Option<BaseTokenRatioPersisterL1Params> = None;
```
perekopskiy, line 86: you can avoid mutability with `let l1_params  = self.wallets_config.token_multiplier_setter.map(...)`

core/node/base_token_adjuster/src/base_token_ratio_persister.rs
ischasny: It's much nicer indeed - fixed

core/node/node_framework/src/implementations/layers/base_token/base_token_ratio_persister.rs
ischasny: thank you - fixed

perekopskiy (approved):

PR #2701, opened 2024-08-20
fix: script fixes

kelemeno:
## Checklist


infrastructure/zk/src/contract.ts
```diff
@@ -405,7 +405,7 @@ export async function registerHyperchain({
         privateKey ? `--private-key ${privateKey}` : '',
         baseTokenName ? `--base-token-name ${baseTokenName}` : '',
         deploymentMode == DeploymentMode.Validium ? '--validium-mode' : '',
-        '--use-governance'
+        '--use-governance true'
```
StanislavBreadless, line 408: I don't think that this change is needed, for boolean vars `--x` is enough to say that it is true

PR #2702, opened 2024-08-21
feat(vm): Extract batch executor to separate crate

slowli:
## What ❔

- Extracts batch executor from `zksync_state_keeper` into a separate crate, `zksync_vm_executor`. Places batch executor interfaces in `zksync_vm_interface`.
- Moves `zksync_vm_utils` remains to this crate, too.
- Removes dependency on the `zksync_state_keeper` in the VM runner.

## Why ❔

Improves encapsulation and versatility (e.g., when using custom VMs).

## Checklist


core/node/state_keeper/src/executor/tests/mod.rs
```diff
@@ -1,35 +1,40 @@
+// FIXME: move storage-agnostic tests to VM executor crate
```
slowli, line 1: Should be relatively straightforward; I plan to do this in a follow-up PR.

core/node/vm_runner/Cargo.toml
```diff
@@ -11,17 +11,16 @@ keywords.workspace = true
...
 zksync_dal.workspace = true
 zksync_contracts.workspace = true
 zksync_state.workspace = true
 zksync_storage.workspace = true
-zksync_state_keeper.workspace = true
 zksync_utils.workspace = true
 zksync_prover_interface.workspace = true
 zksync_object_store.workspace = true
-zksync_vm_utils.workspace = true
+zksync_vm_executor.workspace = true
```
slowli, line 23: This dependency could be removed similarly to dependency on state keeper provided that batch executor is injected into VM runner components via DI rather than assembled inside components.

core/lib/vm_interface/src/executor.rs
```diff
@@ -0,0 +1,120 @@
...
+pub struct StandardOutputs<S>(PhantomData<S>);
+
+impl<S: Send + 'static> BatchExecutorOutputs for StandardOutputs<S> {
+    type Tx = BatchTransactionExecutionResult;
+    type Batch = (FinishedL1Batch, StorageView<S>);
+}
+
+pub trait BatchExecutorFactory<S: Send + 'static>: 'static + Send + fmt::Debug {
+    type Outputs: BatchExecutorOutputs;
+    type Executor: BatchExecutor<Self::Outputs> + ?Sized;
```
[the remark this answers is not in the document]
popzxc: Boxed trait object sounds good to me.

core/node/vm_runner/Cargo.toml
popzxc: While it's a good idea, I don't think it's a high priority for now.

popzxc (dismissed):
Can't say that I like the fact that the number of traits, trait wrappers, and other abstract machinery keeps growing. On the other hand, maybe it's not really avoiable without losing some flexibility.

Nonetheless, that's a good change!

popzxc (approved):

PR #2705, opened 2024-08-21
feat(vm): Do not panic on VM divergence

slowli:
## What ❔

- Allows to continue batch execution on divergence via new `ShadowLenient` VM mode.
- Dumps VM state to logs and optionally a file on divergence.

## Why ❔

Allows to detect divergencies in multiple batches w/o blockers. The dumped VM state will hopefully allow investigating divergencies locally, although this logic isn't implemented yet.

## Checklist


slowli:
May be an overkill, but I think the dump part would be useful even if we debug VM divergencies using other means. It can allow regression testing by snapshotting failing executions (hopefully, with reduction to the block or even transaction) and persisting it in the codebase, e.g. as JSON. From other functionality I've implemented, shadow VM sanity check tests also look nice. Could be extended with positive cases in the future by implementing a VM mutating a VM output.

core/lib/multivm/src/dump.rs
```diff
@@ -0,0 +1,160 @@
...
+use serde::{Deserialize, Serialize};
+use zksync_types::{web3, L1BatchNumber, Transaction, H256, U256};
+use zksync_utils::u256_to_h256;
+use zksync_vm_interface::{
+    storage::{InMemoryStorage, ReadStorage, StoragePtr, StorageView},
+    L1BatchEnv, L2BlockEnv, SystemEnv, VmFactory,
+};
+
+/// Handler for [`VmDump`].
+pub type VmDumpHandler = Arc<dyn Fn(VmDump) + Send + Sync>;
```
popzxc, line 12: You may introduce a wrapper type that would implement `Debug` to not force any type that contains it to have a custom `Debug` implementation.

core/lib/multivm/src/versions/shadow/mod.rs
```diff
@@ -0,0 +1,559 @@
...
+        self.finalize_dump(used_contract_hashes);
+        (self.dump_handler)(self.partial_dump);
+
+        if self.panic_on_divergence {
+            panic!("{err:?}");
+        } else {
+            tracing::error!("{err:#}");
+            tracing::warn!(
+                "New VM is dropped; following VM actions will be executed only on the main VM"
+            );
```
popzxc, line 65: If you want, you can use `zksync_vlog::sentry::capture_message` to emit a sentry event to make it easier to notice.

core/lib/multivm/src/dump.rs
```diff
@@ -0,0 +1,379 @@
...
+        BootloaderMemory, BytecodeCompressionError, CompressedBytecodeInfo, CurrentExecutionState,
+        FinishedL1Batch, L1BatchEnv, L2BlockEnv, SystemEnv, VmExecutionMode,
+        VmExecutionResultAndLogs, VmFactory, VmInterface, VmInterfaceHistoryEnabled,
+        VmMemoryMetrics,
+    },
+    vm_fast, vm_latest, HistoryMode,
+};
+
+/// VM that tracks decommitment of bytecodes during execution. This is required to create a [`VmDump`].
+pub trait VmTrackingContracts: VmInterface {
```
slowli, line 22: It could make sense to make this a part of `VmInterface` trait (maybe, even replace `get_current_execution_state`, which has not quite obvious caveats, like being callable only at the end of VM execution in old VM versions). If so, can do this in a follow-up PR.

core/lib/multivm/src/dump.rs
```diff
@@ -0,0 +1,379 @@
...
+    }
+}
+
+impl<S: ReadStorage> VmTrackingContracts for vm_fast::Vm<S> {
+    fn used_contract_hashes(&self) -> Vec<U256> {
+        self.decommitted_hashes().collect()
+    }
+}
+
+// FIXME: use storage snapshots (https://github.com/matter-labs/zksync-era/pull/2724)
```
slowli, line 39: IDK whether it makes sense to wait for https://github.com/matter-labs/zksync-era/pull/2724 to get merged. Transition to using storage snapshots from there looks relatively straightforward; perhaps, the most non-trivial part is serialization logic.

core/lib/basic_types/src/vm.rs
```diff
@@ -35,5 +35,9 @@ pub enum FastVmMode {
     /// Run only the new Vm.
     New,
     /// Run both the new and old VM and compare their outputs for each transaction execution.
+    /// The VM will panic on divergence.
     Shadow,
+    /// Run both the new and old VM and compare their outputs for each transaction execution.
+    /// The new VM will get dropped on divergence.
+    ShadowLenient,
```
slowli, line 42: An alternative approach would be the dump handler to handle both dumps and errors:

```rust
pub struct VmDumpHandler(Arc<dyn Fn(VmDump, anyhow::Error) + Send + Sync>);
```

This allows to not introduce a new VM mode, and it would allow to implement `zksync_vlog::sentry` suggestion in a less intrusive way (IMO, making `multivm` depend on such a high-level crate is pushing it). OTOH, gluing handling of multiple things together looks slightly hacky. WDYT?

core/node/vm_runner/src/impls/playground.rs
```diff
@@ -120,6 +138,23 @@ impl VmPlayground {
         ))
     }
 
+    async fn dump_vm_state(object_store: &dyn ObjectStore, dump: &VmDump) -> anyhow::Result<()> {
+        let timestamp = SystemTime::now()
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .expect("bogus clock");
+        let timestamp = timestamp.as_millis();
+        let batch_number = dump.l1_batch_number();
+        let dump_filename = format!("shadow_vm_dump_batch{batch_number:08}_{timestamp}.json");
```
slowli, line 147: Not sure whether a timestamp suffix is required. Of course, its purpose is to retain dumps for unrelated errors in the same batch; but as a downside, if the same error occurs, it'll still be dumped to a multitude of files. I guess if a single dump / error handler is implemented (see the other line comment), it could be possible to use a tag obtained by hashing the error message (?).

core/lib/basic_types/src/vm.rs
slowli: Decided to do just that.

slowli:
Ready for review with the caveat that I'm not entirely sure how to deal with storage snapshot serialization. At the end, it's a no-brainer to reuse snapshots from https://github.com/matter-labs/zksync-era/pull/2724, but until that PR is landed, I'm not sure what would be the best approach.

core/lib/multivm/src/dump.rs
```diff
@@ -0,0 +1 @@
+
```
perekopskiy, line 1: should this be removed?

perekopskiy (approved):

core/lib/multivm/src/dump.rs
slowli: Yep, it's a refactoring artifact.

perekopskiy (approved):

PR #2709, opened 2024-08-22
fix(zk_toolbox): Set proper artifacts

Deniallugo:
## Checklist


core/tests/recovery-test/src/index.ts
```diff
@@ -147,6 +147,43 @@ export class NodeProcess {
...
+        }
+
+        let signalNumber;
+        if (signal == 'KILL') {
+            signalNumber = 9;
+        } else {
+            signalNumber = 15;
+        }
+        try {
+            console.log('Main PID', this.childProcess.pid);
```
manuelmauro, line 162: nit: Is this log still useful?

docker-compose.yml
```diff
@@ -3,6 +3,8 @@ services:
   reth:
     restart: always
     image: "ghcr.io/paradigmxyz/reth:v0.2.0-beta.2"
+    ports:
+      - 127.0.0.1:8545:8545
```
manuelmauro, line 7: This is why we use `localhost` now in CI? Why was this needed?

manuelmauro (approved):
LGTM! 🥳

core/tests/recovery-test/src/index.ts
Deniallugo: I think no

docker-compose.yml
Deniallugo: Not this, but `network_mode: host` in `zk` container

PR #2713, opened 2024-08-22
feat(zk_toolbox): Deploy ConsensusRegistry (BFT-504) 

aakoshh:
## What ❔

Adds a `zk_inception chain deploy-consensus-registry` command.

TODO: 
- [x] Change `contracts` submodule back to `main` once https://github.com/matter-labs/era-contracts/pull/735 is merged

### Contract Owner

The agreement was that on testnet the `ConsensusRegistry` contract should be owned by the governor account, which is 0xD64e136566a9E04eb05B30184fF577F52682D182, while on mainnet it should be owned by the [developer multisig account](https://app.safe.global/transactions/queue?safe=eth:0x9e543149DdfEEE18e95A4655D07096398Dd2Bf52). 

The owner is set in [DeployL2ContractsInput::consensus_registry_owner](https://github.com/matter-labs/zksync-era/blob/f4b7c12431d4bb063c735947f74e30c749119b5f/zk_toolbox/crates/config/src/forge_interface/deploy_l2_contracts/input.rs#L19) which has access to contract and wallet configuration and these are written to a config file just before deployment.

~~I added an optional `developer_multisig` wallet to `WalletConfig`, so the address can be added at the same place as the `governor` address is; if `developer_multisig` is missing then `governor` is used. I suppose it could be made part of the `ContractsConfig` instead, but since this is a wallet with funds that developers can access, I thought it wouldn't be out of place in `wallets.yaml` even if one doesn't have any of the corresponding private keys. Let me know if I should be using something else.~~

### Testing

Since the `zk_toolbox` is replacing the `zk` commands, and `zk init` doesn't deploy the consensus registry, we have to use the following commands to see that the contract is built, deployed and its address is written to the config file:

```shell
./bin/zkt
zk_inception ecosystem create
zk_inception containers
zk_inception ecosystem init --dev
```

After this we can check if we see the address in the generated config file:

```console
❯ cat ./chains/era/configs/contracts.yaml | yq .l2.consensus_registry
0x72ada8c211f45e768c9a7781793da84daf1d0d1b
```

Finally clean up:

```shell
zk_supervisor clean all
```

## Why ❔

So that we can deploy the L2 consensus registry contract using the `zk_toolbox`.

## Checklist


contracts
Deniallugo, line 1: please check, that you are using corrrect submodule commit. looks like it's a lot of changes

zk_toolbox/crates/config/src/forge_interface/deploy_l2_contracts/input.rs
```diff
@@ -27,6 +30,10 @@ impl DeployL2ContractsInput {
             bridgehub: contracts.ecosystem_contracts.bridgehub_proxy_addr,
             governance: wallets.governor.address,
             erc20_bridge: contracts.bridges.erc20.l1_address,
+            consensus_registry_owner: wallets
+                .developer_multisig
```
Deniallugo, line 34: I'm not sure it's the correct name. 
Do we need an additional wallet, other than governor?

zk_toolbox/crates/config/src/wallets.rs
```diff
@@ -16,6 +16,10 @@ pub struct WalletsConfig {
     pub fee_account: Wallet,
     pub governor: Wallet,
     pub token_multiplier_setter: Wallet,
+    /// A developer multisig account exists on mainnet.
+    /// The intention here is to use its address to own certain contracts,
+    /// while the private key is either empty, or different for each member.
+    pub developer_multisig: Option<Wallet>,
```
Deniallugo, line 22: I don't think we have to implement it here. We could setup it better manually

contracts
aakoshh: I checked out my branch corresponding to the same feature. Potentially it has more changes because it was branched off `main` relative to what this repo had before. I'll update it after it's been merged. 

If that's not the correct commit to use, can you let me know which one is?

zk_toolbox/crates/config/src/forge_interface/deploy_l2_contracts/input.rs
aakoshh: I don't know, @vladbochok suggested using the "dev multisig" account/wallet. It didn't sound like any of the other wallets, so I made a guess that we need to make room for a new one to be defined, but I could be completely off the mark.

zk_toolbox/crates/config/src/wallets.rs
aakoshh: Okay, but where do you propose to set it manually? 

First I thought that it could be set as an argument to the `deploy-consensus-registry` command, but then it wouldn't have a value if we just go with `deploy-l2-contracts` which does it all and gets all the parameters from the config, and the `governor` is even the preferred value on testnet. Then it automatically writes the config for the forge scripts and executes, so there is no more time to manually set it there.

zk_toolbox/crates/config/src/forge_interface/deploy_l2_contracts/input.rs
Deniallugo: Vlad refered to dev multisig, as a potential governor as the entity that already exists in ML structure. it doesn't mean it has to be in our tooling, because it's really specific to our self. 
I'd rather prefer to have governor as a default entity with an ability to change the governor to whatever we prefer later

aakoshh: Okay great, removed.

zk_toolbox/crates/config/src/wallets.rs
Deniallugo: Let's add a command for changing it. 
set it by default to governor and add an additional step to change it

aakoshh: Can you clarify what exactly the command would change? Do you mean to add a method to `DeployContract.sol` to change the owner of the contract after deployment, and a command to the `zk_inception` tool to invoke it with a new address?

aakoshh: I just had a look at how ownership transfer is supposed to work. 

As I understand I can't call the [initialize](https://github.com/matter-labs/era-contracts/blob/446d391d34bdb48255d5f8fef8a8248925fc98b9/l2-contracts/contracts/ConsensusRegistry.sol#L39) function again, which would immediately transfer the ownership from the governor to the dev multisig, because it's already been initialised when the proxy got deployed. 

Instead I would have to call the public [transferOwnership](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/22489db15621b9a42ebddb1facade6962034e9b9/contracts/access/Ownable2StepUpgradeable.sol#L60) function which would start a 2-step transfer to the dev multisig, and then follow up by the [acceptOwnership](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/22489db15621b9a42ebddb1facade6962034e9b9/contracts/access/Ownable2StepUpgradeable.sol#L79) signed by the multisig to finish the procedure. 

Do you propose adding support to the `zk_toolbox` to with this multiple step process? 
1. If the `governor` wallet is the current owner of the registry, sign a transaction to initiate the transfer
2. I don't know how the Ganache multisig works, but another command would be executed by developers who have access to the private keys in the multisig - it seems a bit overly specific to the use case to have a command for this purpose

Deniallugo: I think we could introduce the function with only transfer ownership

Deniallugo (approved):

contracts
aakoshh: Now that the contracts is merged, I checked `main` out in `contracts` and committed that change in `zksync-era`. It seems to show the right changes (a bit more than related to this ticket, but everything since the current commit), yet it says there are conflicts.

aakoshh: Managed to resolve the conflict, it needed a rebase. It still updates to `main` which brings an extra commit: 446d391d

If that's not okay I made a branch which doesn't have that, only the two registry commits: 5853c3a7 and d3687694 cherry picked, but reckon merging with those would just complicate things later.

brunoffranca (approved):

PR #2716, opened 2024-08-23
feat: Change default_protective_reads_persistence_enabled to false

perekopskiy:
## What ❔

Changes default_protective_reads_persistence_enabled to false both for main and external node

## Why ❔

For EN: it was confirmed that it works well without protective reads
For main node: it's expected that vm_runner_protective_reads is run by default

## Checklist

### DIFF
--- a/core/bin/external_node/src/config/mod.rs
+++ b/core/bin/external_node/src/config/mod.rs
@@ -770,7 +770,7 @@ impl OptionalENConfig {
     }
 
     const fn default_protective_reads_persistence_enabled() -> bool {
-        true
+        false
     }
 
     const fn default_mempool_cache_update_interval_ms() -> u64 {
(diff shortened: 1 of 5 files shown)


RomanBrodetski (approved):

core/bin/external_node/src/config/mod.rs
```diff
@@ -770,7 +770,7 @@ impl OptionalENConfig {
     }
 
     const fn default_protective_reads_persistence_enabled() -> bool {
```
slowli, line 772: Nit: You can remove this function altogether and use `#[serde(default)]`. Also, remove "By default, set to `true`..." as well.

perekopskiy: Updated, also changed default not only for EN but in other places as well

RomanBrodetski (approved):

PR #2720, opened 2024-08-23
feat(Base token): add cbt metrics

ischasny:
* Add cbt-related metrics;
* Move last hardcoded cbt-related properties to the config.

### DIFF
--- a/core/node/base_token_adjuster/src/base_token_ratio_persister.rs
+++ b/core/node/base_token_adjuster/src/base_token_ratio_persister.rs
@@ -182,9 +182,10 @@ impl BaseTokenRatioPersister {
             .observe(Self::duration_from_millis(
                 millis_since_epoch() - start_time,
             ));
-        Err(last_error.unwrap_or(anyhow::format_err!(
-            "Failed to update base token multiplier on L1"
-        )))
+        let error_message = "Failed to update base token multiplier on L1";
+        Err(last_error
+            .map(|x| x.context(error_message))
+            .unwrap_or_else(|| anyhow::anyhow!(error_message)))
     }
 
     async fn retry_fetch_ratio(&self) -> anyhow::Result<BaseTokenAPIRatio> {
--- a/core/node/base_token_adjuster/src/base_token_ratio_persister.rs
+++ b/core/node/base_token_adjuster/src/base_token_ratio_persister.rs
@@ -1,5 +1,6 @@
-use anyhow::Context as _;
 use std::{cmp::max, fmt::Debug, sync::Arc, time::Instant};
+
+use anyhow::Context as _;
 use tokio::{sync::watch, time::sleep};
 use zksync_config::configs::base_token_adjuster::BaseTokenAdjusterConfig;
 use zksync_dal::{ConnectionPool, Core, CoreDal};
--- a/core/node/base_token_adjuster/src/base_token_ratio_persister.rs
+++ b/core/node/base_token_adjuster/src/base_token_ratio_persister.rs
@@ -1,5 +1,5 @@
-use std::{cmp::max, fmt::Debug, sync::Arc, time::Duration};
-
+use std::time::Instant;
+use std::{cmp::max, fmt::Debug, sync::Arc};
 use anyhow::Context as _;
 use tokio::{sync::watch, time::sleep};
 use zksync_config::configs::base_token_adjuster::BaseTokenAdjusterConfig;
@@ -13,7 +13,6 @@ use zksync_types::{
     web3::{contract::Tokenize, BlockNumber},
     Address, U256,
 };
-use zksync_utils::time::millis_since_epoch;
 
 use crate::metrics::{OperationResult, OperationResultLabels, METRICS};
 
@@ -125,7 +124,6 @@ impl BaseTokenRatioPersister {
             return Ok(());
         };
 
-        let start_time = millis_since_epoch();
         let max_attempts = self.config.l1_tx_sending_max_attempts;
         let sleep_duration = self.config.l1_tx_sending_sleep_duration();
         let mut prev_base_fee_per_gas: Option<u64> = None;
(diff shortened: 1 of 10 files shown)


core/node/base_token_adjuster/src/metrics.rs
```diff
@@ -0,0 +1,34 @@
...
+
+/// Roughly exponential buckets for gas (10k – 50M).
+const GAS_BUCKETS: Buckets =
+    Buckets::values(&[1e4, 2e4, 5e4, 1e5, 2e5, 5e5, 1e6, 2e6, 5e6, 1e7, 2e7, 5e7]);
+
+#[derive(Debug, Metrics)]
+#[metrics(prefix = "snapshots_creator")]
+pub(crate) struct BaseTokenAdjusterMetrics {
+    #[metrics(buckets = GAS_BUCKETS)]
+    pub l1_gas_used: Histogram<f64>,
```
RomanBrodetski, line 26: how often is this reported?  Do we update L1 values more often than once per minute or similar? Because if not, I'd just use a gauge and see the exact gas value (as opposed to the buckets)

core/node/base_token_adjuster/src/metrics.rs
```diff
@@ -0,0 +1,34 @@
...
+    #[metrics(buckets = Buckets::LATENCIES)]
+    pub external_price_api_latency: Family<OperationResultLabels, Histogram<Duration>>,
+    #[metrics(buckets = Buckets::LATENCIES)]
+    pub l1_update_latency: Family<OperationResultLabels, Histogram<Duration>>,
```
RomanBrodetski, line 30: Maybe a nit, but I'm not sure it will be nice to show/interpret this data in grafana. We'll have different labels per attempt. This would make it easier to answer questions like "how much time does it take to update L1 **on the second attempt**? **On the third attempt**? I don't think these are the questions we are interested in.

We are interested in "how often does the request result in a error" and maybe what attempt is normally successful. So I'd just remove `attempts` from `OperationResultLabels`, report the counter for errors and maybe a gauge with the successful attempts

RomanBrodetski (dismissed):
looks good but please consider my comments

core/node/base_token_adjuster/src/metrics.rs
ischasny: Good point. Removing attempts. Will report just latency + label of "success" / "failure" for every attempt. Then we can sum up and plot those as we need.

core/node/base_token_adjuster/src/metrics.rs
ischasny: Agreed - switching to gauge

core/node/base_token_adjuster/src/base_token_ratio_persister.rs
```diff
@@ -157,36 +120,120 @@ impl BaseTokenRatioPersister {
...
+
+            match result {
+                Ok(x) => {
+                    tracing::info!(
+                "Updated base token multiplier on L1: numerator {}, denominator {}, base_fee_per_gas {}, priority_fee_per_gas {}",
+                new_ratio.numerator.get(),
+                new_ratio.denominator.get(),
+                base_fee_per_gas,
+                priority_fee_per_gas
+            );
```
perekopskiy, line 150: nit: please align code

core/node/base_token_adjuster/src/base_token_ratio_persister.rs
```diff
@@ -157,36 +120,120 @@ impl BaseTokenRatioPersister {
...
+                        ));
+                    return Ok(());
+                }
+                Err(err) => {
+                    tracing::info!(
+                "Failed to update base token multiplier on L1, attempt {}, base_fee_per_gas {}, priority_fee_per_gas {}: {}",
+                attempt,
+                base_fee_per_gas,
+                priority_fee_per_gas,
+                err
```
perekopskiy, line 169: and here

core/node/base_token_adjuster/src/base_token_ratio_persister.rs
```diff
@@ -157,36 +120,120 @@ impl BaseTokenRatioPersister {
         (base_fee_per_gas, priority_fee_per_gas)
     }
 
+    async fn retry_update_ratio_on_l1(&self, new_ratio: BaseTokenAPIRatio) -> anyhow::Result<()> {
+        let Some(l1_params) = &self.l1_params else {
+            return Ok(());
+        };
+
+        let start_time = millis_since_epoch();
```
perekopskiy, line 128: it's better to use std::time::Instant

core/node/base_token_adjuster/src/base_token_ratio_persister.rs
ischasny: Good point, let me realign it. This is actually what `zk fmt` does by default.

core/node/base_token_adjuster/src/base_token_ratio_persister.rs
ischasny: Fixed

core/node/base_token_adjuster/src/base_token_ratio_persister.rs
ischasny: Fixed

core/node/base_token_adjuster/src/base_token_ratio_persister.rs
```diff
@@ -157,36 +119,110 @@ impl BaseTokenRatioPersister {
...
+                        "Failed to update base token multiplier on L1, attempt {}, base_fee_per_gas {}, priority_fee_per_gas {}: {}",
+                        attempt,
+                        base_fee_per_gas,
+                        priority_fee_per_gas,
+                        err
+                    );
+                    METRICS.l1_update_latency[&OperationResultLabels {
+                        result: OperationResult::Failure,
+                    }]
+                        .observe(Instant::now().duration_since(start_time));
```
perekopskiy, line 171: nit: `start_time.elapsed()`

perekopskiy (approved):

RomanBrodetski (approved):

RomanBrodetski (approved):

PR #2724, opened 2024-08-23
feat(vm-runner): Implement batch data prefetching

slowli:
## What ❔

- Implements prefetching of storage slots / bytecodes accessed by a VM in a batch. Enables it for the VM playground. Optionally shadows prefetched snapshot storage.
- Makes RocksDB cache optional for VM playground.

## Why ❔

- Prefetching will allow to load storage slots / bytecodes for a batch in O(1) DB queries, which is very efficient for local debugging etc. It may be on par or faster than using RocksDB cache. (There's a caveat: prefetching doesn't work w/o protective reads.)
- Disabling RocksDB cache is useful for local testing, since the cache won't catch up during a single batch run anyway.

## Checklist


core/node/node_framework/src/implementations/layers/vm_runner/playground.rs
```diff
@@ -79,10 +82,15 @@ impl WiringLayer for VmPlaygroundLayer {
             window_size: self.config.window_size,
             reset_state: self.config.reset,
         };
+        let storage = if let Some(path) = self.config.db_path {
+            VmPlaygroundStorageOptions::Rocksdb(path)
+        } else {
+            VmPlaygroundStorageOptions::Snapshots { shadow: false }
```
slowli, line 88: I haven't added an option to configure storage shadowing for snapshots. May do so separately.

core/lib/state/src/shadow_storage.rs
```diff
@@ -24,55 +24,73 @@ static METRICS: vise::Global<ShadowStorageMetrics> = vise::Global::new();
 /// to_check_storage -- secondary storage, which will verify it's own return values against source_storage
 /// Note that if to_check_storage value is different than source value, execution continues and metrics/ logs are emitted.
```
itegulov, line 25: > execution continues

nit: no longer the case depending on what `panic_on_divergence` is set to

core/lib/state/src/shadow_storage.rs
```diff
@@ -24,55 +24,73 @@ static METRICS: vise::Global<ShadowStorageMetrics> = vise::Global::new();
...
+                     to be equal to to_check={expected_value:?}",
+                    self.l1_batch_number
+                );
+            } else {
+                tracing::error!(
+                    "read_value({key:?}) -- l1_batch_number={:?} -- expected source={source_value:?} \
+                     to be equal to to_check={expected_value:?}",
+                    self.l1_batch_number
+                );
+            }
```
itegulov, line 71: Maybe it would make the code cleaner if this was a function? Something like `error_or_panic(&str)` so you don't have to repeat formatting twice everywhere

core/lib/state/src/storage_factory.rs
```diff
@@ -75,6 +79,17 @@ pub enum PgOrRocksdbStorage<'a> {
     Rocksdb(RocksdbStorage),
     /// Implementation over a RocksDB cache instance with in-memory DB diffs.
     RocksdbWithMemory(RocksdbWithMemory),
+    /// In-memory storage snapshot.
+    Snapshot(StorageSnapshot),
+    /// Generic implementation. Should be used for testing purposes only.
+    Boxed(Box<dyn ReadStorage + Send + 'a>),
```
itegulov, line 85: Given the amount of variants this enum has at this point, I guess we should consider renaming it from `PgOrRocksdbStorage` to smth else 😅

itegulov (approved):
Looks good overall, just a couple of nits/suggestions

core/lib/state/src/storage_factory.rs
```diff
@@ -153,6 +164,90 @@ impl PgOrRocksdbStorage<'static> {
...
+            .storage_logs_dal()
+            .get_touched_slots_for_l1_batch(l1_batch_number)
+            .await?;
+        tracing::debug!("Loaded {} touched keys", touched_slots.len());
+
+        let all_accessed_keys: Vec<_> = protective_reads
+            .into_iter()
+            .map(|key| key.hashed_key())
+            .chain(touched_slots.into_keys())
+            .collect();
```
perekopskiy, line 214: It probably won't work for every case. E.g. if within an l2 block there were 2 storage write logs: first writes to some slot, second is equal to first but with `rollback: true`; then this slot won't be in `protective_reads` (because of `protective_reads` deduplication rules) nor `touched_slots`(because of l2 block deduplication).
I may be mistaken, please validate

slowli: Nice catch! There are indeed slots that are not covered in this (or similar) way; about 5–10 per batch on the mainnet (~30k–35k storage accesses). I've reworked snapshot logic correspondingly to require a fallback; PTAL.

core/lib/state/src/storage_factory.rs
```diff
@@ -65,19 +69,34 @@ pub struct RocksdbWithMemory {
...
     Postgres(PostgresStorage<'a>),
     /// Implementation over a RocksDB cache instance.
     Rocksdb(RocksdbStorage),
     /// Implementation over a RocksDB cache instance with in-memory DB diffs.
     RocksdbWithMemory(RocksdbWithMemory),
+    /// In-memory storage snapshot with the Postgres storage fallback.
+    Snapshot(StorageWithSnapshot<PostgresStorage<'a>>),
+    /// Generic implementation. Should be used for testing purposes only since it has performance penalty because
+    /// of the dynamic dispatch.
+    Boxed(Box<dyn ReadStorage + Send + 'a>),
```
slowli, line 89: This variant and changes in the `ShadowStorage` are now technically unnecessary, but I'd argue they could still be useful (esp. `ShadowStorage`).

perekopskiy (approved):
Nice!

PR #2725, opened 2024-08-24
feat: Add consensus registry address to the AttestationStatus API (BFT-503)

aakoshh:
## What ❔

Adds a new `contract_registry_address` field to the `AttestationStatus` returned by the main node API to the external nodes, threading it through the main node configuration as `l2_consensus_registry_addr`. 

Depends on https://github.com/matter-labs/zksync-era/pull/2713 which changes the `zk_inception` tool to deploy the `ConsensusRegistry` contract and write the address of its proxy to `./chains/<chain>/configs/contracts.yaml` into the `.l2.consensus_registry` field, which is read by the protobuf formats in this PR.

## Why ❔

We want to move the management of the attestation committee on chain, rather than through Genesis. The idea is to deploy the contract on an existing chain, then configure its address on the main node not in Genesis, but the regular configuration file, then let the external nodes pick it up as they poll the `attestationStatus` endpoint. 

The reason we do it through the API rather than Genesis is to avoid nodes which do not get their clients upgraded to drop out of the gossip network because of a change in the Genesis hash: if the main node added a new field it would calculate a different hash than the ones which do not have that field. While this feature is being developed it's less disruptive to treat is as optional for those who participate in attestations.

## Checklist


core/lib/dal/src/consensus_dal.rs
```diff
@@ -554,6 +554,7 @@ impl ConsensusDal<'_, '_> {
         Ok(Some(AttestationStatus {
             genesis: genesis.hash(),
             next_batch_to_attest,
+            consensus_registry_address: None, // Assigned in the API layer.
```
pompon0, line 557: main node also needs to know this address, while it cannot query api currently. How do we resolve this?

aakoshh: My potentially flawed understanding is by adding it to [the config proto](https://github.com/matter-labs/zksync-era/pull/2725/files#diff-9c82e3284daf363ee808d461fc108a1801ddcdfcfb8e00f23ee491d310ea6a69R24) it will read the value which is written by https://github.com/matter-labs/zksync-era/pull/2713 

```
cat ./chains/era/configs/contracts.yaml | yq .l2.consensus_registry
0x72ada8c211f45e768c9a7781793da84daf1d0d1b
```

This is then read when you use the `zk_toolbox` to [start the server](https://github.com/matter-labs/zksync-era/blob/c162510598b45dc062c2c91085868f8aa966360e/zk_toolbox/crates/zk_inception/src/commands/server.rs#L58)

I also assume that on the kubernetes config this would be something we have to add to some config file manually after the contract has been deployed, for example [here](https://github.com/matter-labs/gitops-kubernetes/blob/f29949bc16ce5072df66c797b6dc2d114d5faa8b/apps/environments/stage/server-v2-config/contracts.yaml#L11).

core/bin/external_node/src/config/mod.rs
```diff
@@ -1394,6 +1394,7 @@ impl From<&ExternalNodeConfig> for InternalApiConfig {
             transparent_proxy_admin_addr: config.remote.transparent_proxy_admin_addr,
             diamond_proxy_addr: config.remote.diamond_proxy_addr,
             l2_testnet_paymaster_addr: config.remote.l2_testnet_paymaster_addr,
+            l2_consensus_registry_addr: None, // Main Node only; on ENs it's etched separately by the consensus module.
```
RomanBrodetski, line 1397:
```suggestion
            l2_consensus_registry_addr: None, // Main Node only; on ENs it's fetched separately by the consensus module.
```

core/lib/dal/src/consensus_dal.rs
pompon0: ok, so you are suggesting to just propagate it to the consensus component from the contracts config. I suppose that's ok.

pompon0 (dismissed):

PR #2729, opened 2024-08-26
docs(prover): Recommend standard provisioning over spot

popzxc:
## What ❔

A few reports have shown that using spot instances is very luck-dependent, so it's not worth trying at the cost of flow disruption.

## Checklist


itegulov (approved):
❤️

PR #2735, opened 2024-08-26
feat: Refactor metrics/make API use binaries

Artemka374:
## What ❔

Add metric-emitting middleware for ExternalProofIntegrationAPI
Make API return binaries/allow uploading binaries for verification

## Why ❔

For better UX

## Checklist


core/node/external_proof_integration_api/src/lib.rs
```diff
@@ -50,25 +54,29 @@ async fn create_router(
     let mut processor =
         Processor::new(blob_store.clone(), connection_pool.clone(), commitment_mode);
     let verify_proof_processor = processor.clone();
+    let specific_proof_processor = processor.clone();
     Router::new()
         .route(
             "/proof_generation_data",
```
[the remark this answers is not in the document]
Artemka374: yeah, that's how axum works. You are not able to have 1 method with optional path, before it used JSON to understand whether you want to get latest/specific, but you cannot path JSON to wget/curl to download the file

core/node/external_proof_integration_api/src/metrics.rs
```diff
@@ -25,30 +27,37 @@ pub(crate) struct ProofIntegrationApiMetrics {
...
+    let start = Instant::now();
+    let path = request.uri().path();
 
-impl MethodCallGuard {
-    pub(crate) fn new(method_type: Method) -> Self {
-        MethodCallGuard {
-            method_type,
-            outcome: CallOutcome::Failure,
-            started_at: Instant::now(),
+    let method = if path.starts_with("/proof_generation_data") {
```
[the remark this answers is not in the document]
popzxc: You can improve the ergonomics by creating a factory for middlewares, e.g.:

```rust
    let middleware_factory = |method: Method| {
        axum::middleware::from_fn(move |req: Request, next: Next| async move {
            metrics_middleware(method);
            next.run(req).await
        })
    };

    let app = axum::Router::new()
        .route(
            "/handler_a",
            axum::routing::get(handler_a).layer(middleware_factory(Method::A)),
        )
        .route(
            "/handler_b/:input",
            axum::routing::get(handler_b).layer(middleware_factory(Method::B)),
        );
```

(note that it _has_ to be a closure, since it works with unnameable types)

core/node/external_proof_integration_api/src/processor.rs
```diff
@@ -45,17 +43,32 @@ impl Processor {
...
+            tracing::error!("Failed to read field: {:?}", err);
+            ProcessorError::InvalidProof
+        })? {
+            while let Some(chunk) = data.chunk().await.map_err(|err| {
+                tracing::error!("Failed to read chunk: {:?}", err);
+                ProcessorError::InvalidProof
+            })? {
+                serialized_proof.extend_from_slice(&chunk);
+            }
+        }
```
popzxc, line 66: Is it correct? I'm not an expert, but shouldn't we find a single exact field with the correct params (`name`/`filename`/`content-type`) and extract it only? Right now it looks like we're going through all the fields and collecting them into a single vector (which would work for forms with a single field, but is logically incorrect).

core/node/external_proof_integration_api/src/processor.rs
```diff
@@ -93,38 +96,87 @@ impl Processor {
...
-                    latest_available_batch
-                );
-                return Err(ProcessorError::BatchNotReady(l1_batch_number));
+        let proof_generation_data = self
+            .proof_generation_data_for_existing_batch_internal(latest_available_batch)
+            .await;
+
+        match proof_generation_data {
+            Ok(data) => {
+                let data = bincode::serialize(&data)?;
```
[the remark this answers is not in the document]
popzxc: I think it would be better to extract into a type that implements `IntoResponse` (and return it instead of opaque `impl IntoResponse`)

zk_toolbox/crates/zk_supervisor/src/commands/test/prover.rs
```diff
@@ -6,7 +6,7 @@ use crate::messages::MSG_PROVER_TEST_SUCCESS;
 
 pub fn run(shell: &Shell) -> anyhow::Result<()> {
     let ecosystem = EcosystemConfig::from_file(shell)?;
-    let _dir_guard = shell.push_dir(&ecosystem.link_to_code.join("prover"));
+    let _dir_guard = shell.push_dir(ecosystem.link_to_code.join("prover"));
```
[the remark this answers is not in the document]
Artemka374: meh, just lint was screaming about it and that's it

core/node/external_proof_integration_api/src/processor.rs
Artemka374: Well, I don't think it's good idea to enforce some exact filename, but other things sound good

prover/docs/05_proving_batch.md
```diff
@@ -72,13 +72,13 @@ input file, called `witness_inputs_<batch>.bin` generated by different core comp
   batch, that was already proven. Example:
 
   ```shell
-  curl -H "Content-Type: application/json" -X POST {address}/proof_generation_data -d 'null'
+  wget --content-disposition {address}/proof_generation_data
```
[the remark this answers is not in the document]
Artemka374: Huh, that's tricky, cause curl doesn't want to save the data with content-disposition(file name that described in response), but wget doesn't allow to post files with multiform, so I'll leave it as is for now

core/node/external_proof_integration_api/src/processor.rs
```diff
@@ -93,38 +124,57 @@ impl Processor {
...
-                return Err(ProcessorError::BatchNotReady(l1_batch_number));
-            }
+        let proof_generation_data = self
+            .proof_generation_data_for_existing_batch_internal(latest_available_batch)
+            .await;
+
+        match proof_generation_data {
+            Ok(data) => Ok(ProofGenerationDataResponse(data)),
+            Err(err) => Err(err),
+        }
```
popzxc, line 134:
```rust
self
    .proof_generation_data_for_existing_batch_internal(latest_available_batch)
    .await
    .map(ProofGenerationDataResponse)
```

core/node/external_proof_integration_api/src/processor.rs
```diff
@@ -93,38 +124,57 @@ impl Processor {
...
+    }
+
+    #[tracing::instrument(skip(self))]
```
popzxc, line 137: Nit: Why do we need tracing here if we have metrics? Given that we don't sample traces, we don't want to record traces of user-triggered calls.

I understand that it wasn't introduced here, just noticed.

core/node/external_proof_integration_api/src/processor.rs
```diff
@@ -93,38 +124,57 @@ impl Processor {
...
         match proof_generation_data {
-            Ok(data) => {
-                guard.mark_successful();
-
-                Ok(Json(ProofGenerationDataResponse::Success(Some(Box::new(
-                    data,
-                )))))
-            }
+            Ok(data) => Ok(ProofGenerationDataResponse(data)),
             Err(err) => Err(err),
```
popzxc, line 172: Same as above.

core/node/external_proof_integration_api/src/processor.rs
```diff
@@ -45,17 +69,32 @@ impl Processor {
...
+
+        while let Some(field) = multipart
+            .next_field()
+            .await
+            .map_err(|_| ProcessorError::InvalidProof)?
+        {
+            if field.name() == Some("proof")
+                && field.content_type() == Some("application/octet-stream")
+            {
+                serialized_proof.extend_from_slice(&field.bytes().await.unwrap());
```
popzxc, line 90: You can `break` here.

core/node/external_proof_integration_api/src/processor.rs
```diff
@@ -45,17 +69,32 @@ impl Processor {
...
+            }
+        }
+
+        tracing::info!("Received proof is size: {}", serialized_proof.len());
+
+        let payload: VerifyProofRequest = bincode::deserialize(&serialized_proof)?;
+
         let expected_proof = bincode::serialize(
```
popzxc, line 98: Pass-by thought: proofs can be wiped from blob store, right? We want to report such cases gracefully, so if the error is "not found" we probably want to check if such a batch exists. If it does -- return `ProofWasPurged` error, otherwise, `ProofNotFound`. Can be tackled separately.

core/node/external_proof_integration_api/src/middleware.rs
```diff
@@ -0,0 +1,21 @@
+use tokio::time::Instant;
+
+use crate::metrics::{CallOutcome, Method, METRICS};
+
+pub(crate) struct MetricsMiddleware {
```
popzxc, line 5: Please derive `Debug`.

popzxc (commented):
Mostly LGTM

core/node/external_proof_integration_api/src/processor.rs
Artemka374: It is done in a bit different way, I have added one more tackling in verify_proof, but generally - it will return object store error with message that batch is too old. Otherwise, if the batch number is bigger than latest proven - it will return BatchNotReady error.

popzxc (approved):

popzxc (dismissed):

popzxc (approved):

PR #2490, opened 2024-07-25
feat(zk_toolbox): use foundry zksync

Deniallugo:
## Checklist

- [ ] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [ ] Tests for the changes have been added / updated.
- [ ] Documentation comments have been added / updated.
- [ ] Code has been formatted via `zk fmt` and `zk lint`.

PR #2737, opened 2024-08-26
feat(contracts)!: integrate protocol defense changes

koloz193:
## What ❔

The work done in the protocol defense project introduced a number of changes, namely custom errors in our solidity contracts. We need to update the server code to handle these new errors.

## Checklist


koloz193:
Need to wait for https://github.com/matter-labs/zksync-era/pull/2490 to get merged for the toolbox tests to pass

core/lib/multivm/src/versions/vm_latest/tests/nonce_holder.rs
```diff
@@ -101,7 +102,7 @@ fn test_nonce_holder() {
     run_nonce_test(
         1u32,
         NonceHolderTestMode::SetValueUnderNonce,
-        Some("Previous nonce has not been used".to_string()),
+        Some("Error function_selector = 0x13595475, data = 0x13595475".to_string()),
```
StanislavBreadless, line 105: it would be really nice to have some programmatic way to generate those. these errors wont be readable by someone who does not know that to search for preimage of the selector you need to do full search over the repo.

also, it is not convenient to generate those test cases

core/tests/ts-integration/tests/contracts.test.ts
```diff
@@ -74,13 +74,13 @@ describe('Smart contract behavior checks', () => {
         await expect(contract.getFooName()).resolves.toBe('Foo');
     });
 
-    test('Should perform "expensive" contract calls', async () => {
+    test.skip('Should perform "expensive" contract calls', async () => {
```
StanislavBreadless, line 77: why? tests should not be skipped in the `main` branch

etc/env/base/contracts.toml
```diff
@@ -26,13 +26,13 @@ RECURSION_NODE_LEVEL_VK_HASH = "0x1186ec268d49f1905f8d9c1e9d39fc33e98c74f91d91a2
...
-GENESIS_BATCH_COMMITMENT = "0x2d00e5f8d77afcebf58a6b82ae56ba967566fe7dfbcb6760319fb0d215d18ffd"
+GENESIS_ROOT = "0x28a7e67393021f957572495f8fdadc2c477ae3f4f413ae18c16cff6ee65680e2"
+GENESIS_BATCH_COMMITMENT = "0xc57085380434970021d87774b377ce1bb12f5b6064af11595e70011965747def"
 PRIORITY_TX_MAX_GAS_LIMIT = 72000000
 DEPLOY_L2_BRIDGE_COUNTERPART_GAS_LIMIT = 10000000
 GENESIS_ROLLUP_LEAF_INDEX = "54"
-GENESIS_PROTOCOL_VERSION = "24"
-GENESIS_PROTOCOL_SEMANTIC_VERSION = "0.24.2"
+GENESIS_PROTOCOL_VERSION = "25"
+GENESIS_PROTOCOL_SEMANTIC_VERSION = "0.25.2"
```
StanislavBreadless, line 35: probably 0.25.0

zk_toolbox/crates/zk_inception/src/accept_ownership.rs
```diff
@@ -94,8 +122,17 @@ async fn accept_ownership(
 ) -> anyhow::Result<()> {
     forge = fill_forge_private_key(forge, governor)?;
     check_the_balance(&forge).await?;
-    let spinner = Spinner::new(MSG_ACCEPTING_GOVERNANCE_SPINNER);
```
StanislavBreadless, line 97: why no spinner anymore?

koloz193: maybe @Deniallugo knows

core/lib/multivm/src/versions/vm_latest/tests/nonce_holder.rs
koloz193: thats a good point, ill see what we can do

perekopskiy (commented):
1. Do you plan to increase `MAXIMUM_L2_GAS_PRICE`, `MAXIMUM_PUBDATA_PRICE` in server? It won't be easy if go over i32::MAX because db columns are of type bigint, so non-trivial migration will be required. Also, no matter what the increase will be (if any) it should be covered with tests
2. Did you check why VM performance degraded?

Deniallugo (approved):

koloz193:
@perekopskiy For the server side increases of those values, given it will take a non-trivial amount of work I think we should do it in a follow up PR, the important this here is that it will be enabled from a protocol perspective. For performance, I ran the performance test locally on main and on this branch. 

Here are the results:
Screenshot 2024-10-11 at 11 33 51 AM

currently im trying to investigate the degraded performance for `access_memory`

perekopskiy (dismissed):

hatemosphere (approved):
CI changes LGTM

Deniallugo (approved):

cytadela8:
> * Do you plan to increase `MAXIMUM_L2_GAS_PRICE`, `MAXIMUM_PUBDATA_PRICE` in server? It won't be easy if go over i32::MAX because db columns are of type bigint, so non-trivial migration will be required. Also, no matter what the increase will be (if any) it should be covered with tests

@perekopskiy [bigint in postgresql is 8bytes](https://www.postgresql.org/docs/current/datatype-numeric.html)

perekopskiy:
@cytadela8 yep, but signed, replace i32 with i64 in my message

PR #2741, opened 2024-08-27
fix(vm): Fix used bytecodes divergence

slowli:
## What ❔

Fixes divergence in used bytecodes info between the old and new VMs.

## Why ❔

The new VM behaved differently to the old VM on far call if decommitting the called contract leads to out-of-gas revert. The old VM records the called contract bytecode as decommitted in this case; the new one didn't.

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8931,6 +8931,7 @@ name = "zksync_multivm"
 version = "0.1.0"
 dependencies = [
  "anyhow",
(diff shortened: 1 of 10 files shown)


slowli:
JFYI, reported changes in the runtime are caused by switching to segmented heap (not directly related to this fix, but it was merged in `vm2` codebase recently and wasn't applied in this repo).

PR #2743, opened 2024-08-27
feat: add da clients

dimazhornyk:
## What ❔

This PR changes the approach to managing 3rd party DA clients. It was assumed before that they will be stored in a separate repository (hyperchain-da), but to simplify the processes and solve a recursive dependency problem, we decided to manage those within `zksync-era`.

The config now defines which DA client will be used, for proto-based configuration it requires adding these lines to general.yaml:
```
da_client:
  avail:
    api_node_url: wss://turing-rpc.avail.so/ws
    bridge_api_url: undefined
    seed: SEED_PHRASE
    app_id: 82
    timeout: 3
    max_retries: 5
``` 
for env-based:
```
    DA_CLIENT="Avail"
    DA_API_NODE_URL="localhost:12345"
    DA_BRIDGE_API_URL="localhost:54321"
    DA_SEED="SEED_PHRASE"
    DA_APP_ID=1
    DA_TIMEOUT=2
    DA_MAX_RETRIES=3
```
If no config is provided - the default behavior is to use NoDA client (same as now).

The `da_client` config might be merged with `da_dispatcher` at some point as the second depends on the first one, so their separation does not make much sense (apart from simplification of the configs). But I'd prefer to do it as a separate PR in case we decide to merge them.

The client was reimplemented using only lightweight libraries from crates.io, so it doesn't have any visible impact on build time.

## Why ❔

To enable seamless integration with 3rd party DA clients in `zksync-era`.

## Checklist


dimazhornyk:
I have a concern regarding the size of `Cargo.lock` diff, @popzxc suggested that it might be due to the `cargo update`, but I attempted regenerating it after reverting to the one used in `main` branch and `cargo clean`, still the same.
Whoever reviews this, please try generating a lock file on your machine to ensure that the diff there is valid

popzxc:
> but I attempted regenerating it after reverting to the one used in main branch and cargo clean, still the same.

I think it's because of avail and git deps, so chances are that we can't do much. I will review the changes though.

core/node/da_clients/src/avail.rs
```diff
@@ -0,0 +1,189 @@
+use std::{
+    fmt::{Debug, Formatter},
+    sync::Arc,
+};
+
+use alloy::{
+    primitives::{B256, U256},
+    sol,
+};
```
popzxc, line 9: `alloy` is only used for rather simple purposes here. We are hesitant to introduce it until we properly migrate.
May I ask you to reimplement this functionality with `zksync_basic_types::web3::contract::{Tokenizable, Tokenize}` (e.g. manually do what the proc macro does for you now?

I understand that it isn't strictly convenient, but if we will enter a phase where half of the code uses old `web3` code and half is on `alloy`, it'll be extremely messy.

core/node/da_clients/src/avail.rs
```diff
@@ -0,0 +1,189 @@
...
+    DataAvailabilityClient,
+};
+
+/// An implementation of the `DataAvailabilityClient` trait that interacts with the Avail network.
+#[derive(Clone)]
+pub struct AvailClient {
+    config: AvailConfig,
+    client: Arc<AvailSubxtClient>,
+    #[allow(unused)]
+    api_client: Arc<reqwest::Client>,
```
popzxc, line 34: Is it needed?

popzxc (commented):
The CI seems to be failing

core/node/da_clients/src/avail.rs
dimazhornyk: makes sense, it was part of the implementation done by Avail team, but it's related to Stage 2 (inclusion verification), which is out of the scope of this PR, so I removed it completely

core/node/da_clients/src/avail.rs
dimazhornyk: removed

dimazhornyk:
I reimplemented the client using lightweight libraries, so it doesn't have impact on build times
The PR can be reviewed now, the failing build is related to the flaky CI, so don't mind it
cc @popzxc

core/lib/config/src/configs/da_client/mod.rs
```diff
@@ -0,0 +1,19 @@
+use serde::Deserialize;
+
+use crate::{AvailConfig, ObjectStoreConfig};
+
+pub mod avail;
+
+#[derive(Debug, Clone, PartialEq, Deserialize)]
+pub struct DAClientConfig {
+    #[serde(flatten)]
+    pub client: DAClient,
```
popzxc, line 10: [This](https://github.com/softprops/envy/issues/26#issuecomment-1072980219) seems to be a better approach to implement this functionality compared to custom deserializer.

We only rely on `serde` as an implementation detail; in the code outside of the config crate we rely on either `FromEnv` trait or the protobuf codegen.

So what you can do is to remove `Deserialize` from this structure, and then use `envy` to deserialize `DAClient` inside of the `FromEnv` implementation. Please leave a comment explaining why this structure doesn't implement `Deserialize` though.

core/lib/config/src/configs/da_client/mod.rs
```diff
@@ -0,0 +1,19 @@
...
+}
+
+#[derive(Debug, Clone, PartialEq, Deserialize)]
+#[serde(tag = "client")]
+pub enum DAClient {
+    NoDA,
```
popzxc, line 16: There seem to be a confusion in logic with this variant. You can serialize `NoDA` using env config, but not with protobuf config. Simultaneously, both lack of config and `NoDA` config result in using `NoDAClientWiringLayer`.

Do we need it as an explicit option? What would be the use cases here?
If we can detect a misconfiguration based on this (e.g. no config was passed in one place, but a related config was placed elsewhere), then deserialization should probably be explicit, and `Some(NoDA)` should be treated differently from `None`. If not, probably we don't need `NoDA` variant at all.

Also, if you keep it, please leave a comment explaining how this option can be used (e.g. the application should err if there is no `DAClient` config provided, but `DADispatcher` config is present).

core/node/da_clients/Cargo.toml
```diff
@@ -9,17 +9,31 @@ repository.workspace = true
 license.workspace = true
 keywords.workspace = true
 categories.workspace = true
+publish = false
```
popzxc, line 12: No longer relevant I guess?

core/node/da_clients/src/avail.rs
```diff
@@ -0,0 +1,404 @@
...
+const PROTOCOL_VERSION: u8 = 4;
+
+/// An implementation of the `DataAvailabilityClient` trait that interacts with the Avail network.
+#[derive(Debug)]
+pub struct AvailClient {
+    config: AvailConfig,
+    keypair: Keypair,
+}
+
+// Utility type needed for encoding the call data
```
popzxc, line 30:
```suggestion
/// Utility type needed for encoding the call data
```

core/node/da_clients/src/avail.rs
```diff
@@ -0,0 +1,404 @@
...
+#[derive(parity_scale_codec::Encode, scale_encode::EncodeAsType)]
+#[encode_as_type(crate_path = "scale_encode")]
+pub struct SubmitData {
+    pub data: BoundedVec<u8>,
+}
+
+// Utility type needed for encoding the call data
```
popzxc, line 37:
```suggestion
/// Utility type needed for encoding the call data
```

core/node/da_clients/src/avail.rs
```diff
@@ -0,0 +1,404 @@
...
+#[derive(Debug)]
+pub struct AvailClient {
+    config: AvailConfig,
+    keypair: Keypair,
+}
+
+// Utility type needed for encoding the call data
+#[derive(parity_scale_codec::Encode, scale_encode::EncodeAsType)]
+#[encode_as_type(crate_path = "scale_encode")]
+pub struct SubmitData {
```
popzxc, line 33: Do these types have to be public? The least possible visibility should be used by default, so if you can make it private or at least `pub(crate)`, it would be much preferred.

core/node/da_clients/src/avail.rs
```diff
@@ -0,0 +1,404 @@
...
+            keypair: self.keypair.clone(),
+        })
+    }
+
+    fn blob_size_limit(&self) -> Option<usize> {
+        Some(AvailClient::MAX_BLOB_SIZE)
+    }
+}
+
+impl AvailClient {
```
popzxc, line 113: Given that (hopefully) we will use a custom implementation for a while, maybe it'd be better to split the concepts for maintainability:

Create a `pub(super)` module `sdk` in the `avail` module. Add a module doc-comment:
```rust
//! Minimal reimplementation of the Avail SDK client required for the DA client implementation.
//! This is considered to be a temporary solution until a mature SDK is available on crates.io
```

Then, create a structure `RawAvailClient` there and move low-level code and types there.
Finally, store an instance of `RawAvailClient` inside of `AvailClient` and use it as you would use the 3rd party SDK.

You may want to define some helper structures to make interactions between `RawAvailClient` and AvailClient` more intuitive, so that you can write `dispatch_blob` roughly as follows:

```rust
let blob = self.raw_client.encode_blob(..); // Returns a "blob wrapper" type from `raw_client`, not bound to the DA client logic.
let response = self.raw_client.dispatch_blob(..).await; // Returns "raw client specific" response type.
Ok(DispatchResponse::from(response)); // `DispatchResponse` is DA client specific and implements `From` trait
```

This way, the DA client implementation would be less specific and it will be easier to migrate to an alternative implementation in the future.

core/node/da_clients/src/lib.rs
```diff
@@ -0,0 +1,5 @@
+extern crate core;
```
popzxc, line 1: Can be safely removed.

core/node/da_clients/src/avail.rs
popzxc: You can look at `ObjectStoreDA` as a hint: imagine that `RawAvailClient` is equivalent to `Arc<dyn ObjectStore>`.

core/node/da_clients/src/object_store.rs
```diff
@@ -84,3 +88,36 @@ impl DataAvailabilityClient for ObjectStoreDAClient {
         None
     }
 }
+
+/// Used as a wrapper for the pubdata to be stored in the GCS.
+#[derive(Debug)]
+pub struct StorablePubdata {
```
popzxc, line 94: Does it have to be `pub`?

core/node/da_clients/src/object_store.rs
```diff
@@ -84,3 +88,36 @@ impl DataAvailabilityClient for ObjectStoreDAClient {
...
+        format!("l1_batch_{key}_pubdata.gzip")
+    }
+
+    fn serialize(&self) -> Result<Vec<u8>, BoxedError> {
+        let mut encoder = GzEncoder::new(Vec::new(), Compression::default());
+        encoder.write_all(&self.data[..])?;
+        encoder.finish().map_err(From::from)
+    }
+
+    fn deserialize(bytes: Vec<u8>) -> Result<Self, BoxedError> {
```
popzxc, line 112: A snapshot test for this functionality would be appreciated, to make sure that we don't accidentally break the (de)serialization logic. You can look at example in `protobuf_interface/tests/job_serialization.rs`.

popzxc (commented):
Tremendous work!

core/node/da_clients/src/object_store.rs
dimazhornyk: not sure that you meant it to use the mock storage, but I wrote one that does

core/node/da_clients/src/avail.rs
dimazhornyk: done

core/lib/config/src/configs/da_client/mod.rs
popzxc: Then you can implement `flatten` manually, e.g. (pseudocode)

```rust
  let client_tag = std::env::var("DA_CLIENT")?;
  let res = match client_tag {
    Self::AVAIL_CLIENT_CONFIG_NAME => Self::Avail(envy_load("da_avail_config", "DA_")?,
    Self::OBJECT_STORE_CLIENT_CONFIG_NAME => Self::ObjectStore(envy_load("da_object_stroe", DA_")?,
  };
  Ok(res)
}
```

core/node/da_clients/src/object_store.rs
popzxc: No, I meant that you encode some data (doesn't necessarily have to be the actual blob), and store it to the file.
Then in the test, you load this file, deserialize it via `StoredObject::deserialize`, and 
a) make sure deserialization succeeds
b) it contains expected data.

Without such test, if serialization format changes (e.g. imagine that we changed `GzEncoder` to something else), we may not notice that it's a breaking change, but once deployed, the code won't be able to access any of the stored blobs.

core/lib/config/src/configs/da_client/mod.rs
dimazhornyk: done

core/node/da_clients/src/object_store.rs
dimazhornyk: added

popzxc (approved):
Thank you! It was a huge effort.

popzxc:
@dimazhornyk FYI CI is not feeling well (as you might have noticed 😅 ). Monitor Slack for updates.

dimazhornyk:
> @dimazhornyk FYI CI is not feeling well (as you might have noticed 😅 ). Monitor Slack for updates.

yeah, trying my luck with it, but it's quite dead 💀

popzxc:
I've removed this PR from the queue so that we [merge the fix PR ](https://github.com/matter-labs/zksync-era/pull/2855) ASAP

PR #2745, opened 2024-08-27
feat(testing): L1 via Gateway to L2 transaction testing

Raid5594:
## Checklist


core/lib/constants/src/contracts.rs
```diff
@@ -173,3 +177,9 @@ pub const SHARED_BRIDGE_ETHER_TOKEN_ADDRESS: Address = H160([
 /// Default `ERA_CHAIN_ID`. All hyperchains start with this chain id and later on during their registration
 /// an "initial upgrade" transaction overrides it with the correct value.
 pub const DEFAULT_ERA_CHAIN_ID: u32 = 270;
+
+/// Default max number of hyperchains used for testing
+pub const MAX_NUMBER_OF_HYPERCHAINS: u32 = 100;
+
+/// Test ZK chain id
+pub const TEST_ZK_CHAIN_ID: u32 = 272;
```
StanislavBreadless, line 185: does it need to be in the common folder for all constants? I think it is used in only one test

core/lib/multivm/Cargo.toml
```diff
@@ -39,6 +40,7 @@ pretty_assertions.workspace = true
 thiserror.workspace = true
 tracing.workspace = true
 vise.workspace = true
+tiny-keccak.workspace = true
```
StanislavBreadless, line 43: why do we need it? look like a rarely used dependency

core/tests/test_account/src/lib.rs
```diff
@@ -191,6 +195,52 @@ impl Account {
         .unwrap()
     }
 
+    pub fn get_l1_prank_tx(
```
StanislavBreadless, line 198: I understand that here we use the same vocabulary as in foundry, but I am not sure that it will be understandable at a first glance for everyone in the platform team, please rename to a more understandable name

core/tests/test_account/src/lib.rs
```diff
@@ -191,6 +195,52 @@ impl Account {
...
+        &self,
+        execute: Execute,
+        serial_id: u64,
+        prank_address: Address,
+    ) -> Transaction {
+        let max_fee_per_gas = U256::from(0u32);
+        let gas_limit = U256::from(20_000_000);
```
StanislavBreadless, line 205: why 20 million? Why not something like 72kk (the maximal gas value for the transaction), just seems like a random constant

core/tests/test_account/src/lib.rs
```diff
@@ -135,14 +135,18 @@ impl Account {
             value: U256::zero(),
         };
 
+        let mut nonce = 0;
```
StanislavBreadless, line 138: please add a comment here that both using txs nonce or serial id is not exactly a correct way to calculate the deployed address on era, but it is used here for easier testing (and also tell which way is the correct one)

core/lib/multivm/src/versions/vm_latest/tests/l1_l2_complete_tx_execution.rs
```diff
@@ -0,0 +1,602 @@
...
+        (deploy_tx.bytecode_hash, account_code_key),
+    ];
+    assert!(!res.result.is_failed());
+
+    verify_required_storage(&vm.vm.state, expected_slots);
+
+    contract_address
+}
+
+fn send_l2_tx_and_verify(
```
StanislavBreadless, line 110: these seem like a commonly used functions, could you please move those to the utils file in the test folder?

core/lib/multivm/src/versions/vm_latest/tests/l1_l2_complete_tx_execution.rs
```diff
@@ -0,0 +1,602 @@
...
+    let (diamond_init_contract_code, _diamond_init_contract) = read_diamond_init();
+    let (diamond_proxy_contract_code, _diamond_proxy_contract) = read_diamond_proxy();
+    let diamond_init_address =
+        deploy_and_verify_contract(vm, &diamond_init_contract_code, None, &mut deploy_nonce);
+
+    // Collect Data to Initialize STM
+
+    let verifier_params = VerifierParams {
+        recursion_node_level_vk_hash: H256::from([0u8; 32]),
+        recursion_leaf_level_vk_hash: H256::from([0u8; 32]),
```
StanislavBreadless, line 187: H256::zero()

core/lib/multivm/src/versions/vm_latest/tests/l1_l2_complete_tx_execution.rs
```diff
@@ -0,0 +1,602 @@
...
+        recursion_circuits_set_vks_hash: H256::from([0u8; 32]),
+    };
+
+    let l2_bootloader_bytecode_hash = H256::from([
+        0x10, 0x00, 0x10, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+        0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+        0x00, 0x00,
+    ]);
+    let l2_default_account_bytecode_hash = H256::from([
```
StanislavBreadless, line 196: i think you can use the same constant for both

core/lib/multivm/src/versions/vm_latest/tests/l1_l2_complete_tx_execution.rs
```diff
@@ -0,0 +1,602 @@
...
+        0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+        0x00, 0x00,
+    ]);
+
+    let priority_tx_max_gas_limit = U256::from(72000000);
+
+    let protocol_version = 21;
+
+    let force_deployments_data = vec![];
+    let hex_one_32_bytes = [
```
StanislavBreadless, line 207: IMHO the following is easier to read:

u256_to_h256(1.into()).0

core/lib/multivm/src/versions/vm_latest/tests/l1_l2_complete_tx_execution.rs
```diff
@@ -0,0 +1,602 @@
...
+        Some(constructor_data),
+        &mut deploy_nonce,
+    );
+
+    // Call BH to mint new chain
+
+    let (_bridgehub_contract_bytecode, bridgehub_contract) = read_bridgehub();
+
+    // Initialize BH
+    let initialize_calldata = bridgehub_contract
```
StanislavBreadless, line 276: you dont really need to call it for the l2 bridgehub as the owner is already set in the constructor

core/lib/multivm/src/versions/vm_latest/tests/l1_l2_complete_tx_execution.rs
```diff
@@ -0,0 +1,602 @@
...
+    ]);
+
+    // Define the L1_CHAIN_ID, sender, and _additionalData
+    let l1_chain_id = Token::Uint(0.into()); // Chain ID hasn't been set
+    const OFFSET: H160 = H160([
+        0x11, 0x11, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+        0x00, 0x00, 0x00, 0x11, 0x11,
+    ]);
+
+    fn sub_h160(lhs: H160, rhs: H160) -> H160 {
```
StanislavBreadless, line 391: convert both numbers to u256.

add 2^160 to lhs
lhs -= rhs
take modulo 2^160
convert back to h160

this would be more readable/less error prone

core/lib/multivm/src/versions/vm_latest/tests/l1_l2_complete_tx_execution.rs
```diff
@@ -0,0 +1,602 @@
...
+            let (sub, overflow) = a.overflowing_sub(*b);
+            let (sub_with_borrow, borrow_overflow) = sub.overflowing_sub(borrow);
+            result[i] = sub_with_borrow;
+            borrow = (overflow || borrow_overflow) as u8;
+        }
+
+        H160(result)
+    }
+
+    fn undo_l1_to_l2_alias(l2_address: H160) -> H160 {
```
StanislavBreadless, line 405: seems like a good utility function, could you please move it to the utils folder

core/lib/multivm/src/versions/vm_latest/tests/l1_l2_complete_tx_execution.rs
```diff
@@ -0,0 +1,602 @@
...
+
+    let bridgehub_contract =
+        prepare_environment_and_deploy_contracts(&mut vm, default_account.address);
+
+    // Collect Data for Test
+
+    let request_params = BridgeHubRequestL2TransactionOnGateway::default();
+
+    // Generate a large number of vectors
+    let small_vector_size: usize = 45_824; // Size of each vector
```
StanislavBreadless, line 523: This is the maximal possible size, right?

core/lib/constants/src/contracts.rs
Raid5594: I thought it could be reused in the future, by I can remove both

core/lib/multivm/src/versions/vm_latest/tests/l1_l2_complete_tx_execution.rs
Raid5594: yes

core/lib/multivm/src/versions/vm_latest/tests/l1_l2_complete_tx_execution.rs
Raid5594: I don't deploy the BH, it' been deployed as system contract, so without initialize the test would fail

core/tests/test_account/src/lib.rs
Raid5594: What is the correct one to calculate deployed address of era? so I can write a precise comment

StanislavBreadless: by using deployment nonce, which is stored separately from the usual nonce and is incrementally only if there are contract deployments

core/lib/multivm/src/versions/vm_latest/tests/l1_l2_complete_tx_execution.rs
```diff
@@ -0,0 +1,500 @@
...
+fn prepare_environment_and_deploy_contracts(
+    vm: &mut VmTester<HistoryEnabled>,
+    deploy_account_address: Address,
+) -> (Address, Contract) {
+    let mut deploy_nonce = 0;
+
+    // Deploy Bridgehub
+
+    let (bridgehub_contract_bytecode, _bridgehub_contract) = read_bridgehub();
+    let l1_chain_id = Token::Uint(0.into()); // Chain ID hasn't been set
```
StanislavBreadless, line 39: probably better to say that the chain id does not matter for the purpose of the test

core/lib/contracts/src/lib.rs
```diff
@@ -59,6 +59,8 @@ const LOADNEXT_CONTRACT_FILE: &str =
     "etc/contracts-test-data/artifacts-zk/contracts/loadnext/loadnext_contract.sol/LoadnextContract.json";
 const LOADNEXT_SIMPLE_CONTRACT_FILE: &str =
     "etc/contracts-test-data/artifacts-zk/contracts/loadnext/loadnext_contract.sol/Foo.json";
+const MAILBOX_CONTRACT_FILE: &str =
```
StanislavBreadless, line 62: please name it "test mailbox contract", right now it gives off impression that it is a usual mailbox

etc/contracts-test-data/contracts/mailbox/mailbox.sol
```diff
@@ -0,0 +1,103 @@
...
+        L2CanonicalTransaction calldata _transaction,
+        bytes[] calldata _factoryDeps,
+        bytes32 _canonicalTxHash,
+        uint64 _expirationTimestamp
+    ) external {
+        _writePriorityOp(_transaction, _factoryDeps, _canonicalTxHash, _expirationTimestamp);
+    }
+
+    /// @notice Stores a transaction record in storage & send event about that
+    function _writePriorityOp(
```
StanislavBreadless, line 93: it does not write into the priority tree.

this operation would involving hashing the entire data + some storage writes

this makes the test less representative of the reality

StanislavBreadless: Would it be possible to reuse the existing mailbox?

PR #2746, opened 2024-08-27
chore: Update to latest cargo deny

EmilLuta:
Cargo deny is a tool used for license checks (initial motivation), vulnerability checks and other checks (such as unmaintained crates).

This tool has been used across core monorepo for a long time, but given one problematic upgrade it's been pinned to a version. There have been breaking changes since then and the tool got better, but we got stuck with the old version. This upgrades to the new version, but is still pinned. A future development is adding renovate bot, which will keep version up to date. Currently in backlog of @matter-labs/devops.

This PR updates the tool & addresses issues where the fix is straightforward. Other issues will need to be prioritized & treated separately.
They can be found in deny.toml under `advisories.ignore`.

There is space for further improvements on our current defaults, again, not tackled in this PR.

### DIFF
--- a/prover/Cargo.lock
+++ b/prover/Cargo.lock
@@ -8315,7 +8315,6 @@ dependencies = [
  "serde_json",
  "sha3 0.10.8",
  "shivini",
- "structopt",
  "toml_edit 0.14.4",
  "tracing",
  "tracing-subscriber",


Deniallugo (approved):

.github/workflows/cargo-license.yaml
```diff
@@ -4,5 +4,5 @@ jobs:
   cargo-deny:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4
-    - uses: EmbarkStudios/cargo-deny-action@68cd9c5e3e16328a430a37c743167572e3243e7e
+    - uses: actions/checkout@v3
```
yorik, line 7: Please keep hashes and put required version in the comment. It's important for security.

Also why to downgrade checkout?

deny.toml
```diff
@@ -1,15 +1,24 @@
+[graph]
+targets = []
 all-features = false
 no-default-features = false
 
+[output]
+feature-depth = 1
+
 [advisories]
-vulnerability = "deny"
```
yorik, line 5: Any reasons to remove those settings?

deny.toml
```diff
@@ -1,15 +1,24 @@
...
+    "RUSTSEC-2024-0320", # yaml_rust dependency being unmaintained, dependency in core, we should consider moving to yaml_rust2 fork
+    "RUSTSEC-2020-0168", # mach dependency being unmaintained, dependency in consensus, we should consider moving to mach2 fork
+    # all below caused by StructOpt which we still use and we should move to clap v3 instead
+    "RUSTSEC-2021-0145",
+    "RUSTSEC-2021-0139",
+
+]
 
 [licenses]
-unlicensed = "deny"
```
yorik, line 12: This is required by legals.

deny.toml
yorik: Removing that we hope that if default will be changed, we will notice.
I don't have that trust.

.github/workflows/cargo-license.yaml
popzxc: We have plenty of non-pinned actions used in the codebase, even [introduced by the DevOps team itself](https://github.com/matter-labs/zksync-era/commit/599142e8f74228d7a82cc337f965350c1660e8a1#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR32) 🙃 

I would only call this a recommendation, not a rule, until we have automation in place to enforce it.

popzxc (approved):

deny.toml
yorik: I see, I missed that. I guess this is fine then.

.github/workflows/cargo-license.yaml
andreyg-matterlabs, line 1: Removing the hash is not a good idea. We need a pin to a specific version.

yorik (dismissed):

yorik (approved):

PR #2748, opened 2024-08-27
feat: conditional cbt l1 updates

ischasny:
For operational efficiency, this PR enables conditional L1 updates for chains with custom base token. It adds a new configuration `BASE_TOKEN_ADJUSTER_L1_UPDATE_DEVIATION` that defines how much the token price needs to fluctuate in order for the update to be propagated to L1. Equal to 10% by default.


core/node/base_token_adjuster/src/base_token_ratio_persister.rs
```diff
@@ -33,6 +40,7 @@ pub struct BaseTokenRatioPersister {
     base_token_address: Address,
     price_api_client: Arc<dyn PriceAPIClient>,
     l1_params: Option<BaseTokenRatioPersisterL1Params>,
+    last_persisted_l1_ratio: Arc<RwLock<Option<BigDecimal>>>,
```
RomanBrodetski, line 43: hmm perhaps hairsplitting but alternatively you could request the ratio on component initialization - then the behaviour wouldn't depend on component restart (I think now on restart you'd always reset the value on the smart contract).

In that case this value would only be a `None` when `l1_params` above is `None`. 

Actually, I'm not a big fan of this `l1_params: Option<...>` - which is `Some` or `None` depending on the config passed - and then somewhere inside the implementation we dynamically check its value. Alternatively we could do something like `enum BaseTokenL1Behaviour {UpdateOnL1, NoOp} ` and implement all the l1-related logic in its `impl`. 

I won't block the merge by this, but please think about it and do this change if you like it.

RomanBrodetski (commented):
Please consider my comments

core/lib/config/src/configs/base_token_adjuster.rs
```diff
@@ -79,6 +82,11 @@ pub struct BaseTokenAdjusterConfig {
     #[serde(default = "BaseTokenAdjusterConfig::default_l1_tx_sending_sleep_ms")]
     pub l1_tx_sending_sleep_ms: u64,
 
+    /// How many percent a quote needs to change in order for update to be propagated to L1.
+    /// Exists to save on gas.
+    #[serde(default = "BaseTokenAdjusterConfig::default_l1_update_deviation")]
+    pub l1_update_deviation: u32,
```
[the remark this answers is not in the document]
ischasny: Fixed - good point

core/node/base_token_adjuster/src/base_token_ratio_persister.rs
```diff
@@ -124,6 +133,21 @@ impl BaseTokenRatioPersister {
...
 
+        if let Some(ref prev_ratio) = *self.last_persisted_l1_ratio.read().unwrap() {
+            let current_ratio = BigDecimal::from(new_ratio.numerator.get())
+                .div(BigDecimal::from(new_ratio.denominator.get()));
+            let deviation = (prev_ratio - current_ratio.clone())
+                .abs()
+                .div(prev_ratio.clone())
+                .mul(BigDecimal::from(100))
+                .to_u32()
+                .unwrap();
```
perekopskiy, line 144: 1. This will panic easily (e.g. if prev_ratio = 1/2, current_ratio = 1/3: |1/2-1/3|/(1/2)=1/3, 100/3 is not an integer). Instead convert `self.config.l1_update_deviation_percentage` to BigDecimal and compare
2. Let's extract this code in a separate function and cover with unit tests

ischasny: Can you help me to understand why it'd fail? 

is that the scenario you are talking about? 

```
    let prev_ratio = BigDecimal::from(1).div(BigDecimal::from(3));
    let current_ratio = BigDecimal::from(1).div(BigDecimal::from(2));

    let deviation = (prev_ratio.clone() - current_ratio.clone())
        .abs()
        .div(prev_ratio.clone())
        .mul(BigDecimal::from(100))
        .to_u32()
        .unwrap();

    assert_eq!(deviation, 50);
```

ischasny: Moved to a function and covered with tests anyway.

perekopskiy: > is that the scenario you are talking about?

No, you swapped prev_ratio and current_ratio in your example. Now looks good anyway

core/node/base_token_adjuster/src/base_token_ratio_persister.rs
perekopskiy: I wouldn't make any assumptions about number of restarts if possible

ischasny: Ok 2vs1 - deal, will read the prev ratio from the chain on startup.

ischasny: Addressed both points - fetching ratios from L1 on start and moved L1 logic to a separate struct

core/node/base_token_adjuster/src/base_token_l1_behaviour.rs
```diff
@@ -0,0 +1,326 @@
...
+pub struct UpdateOnL1Params {
+    pub eth_client: Box<dyn BoundEthInterface>,
+    pub gas_adjuster: Arc<dyn TxParamsProvider>,
+    pub token_multiplier_setter_account_address: Address,
+    pub chain_admin_contract: Contract,
+    pub getters_facet_contract: Contract,
+    pub diamond_proxy_contract_address: Address,
+    pub chain_admin_contract_address: Option<Address>,
+    pub config: BaseTokenAdjusterConfig,
+    pub last_persisted_l1_ratio: Arc<RwLock<Option<BigDecimal>>>,
```
perekopskiy, line 32: 1. Arc<RwLock> is an overkill in this case, you can just use `Option<BigDecimal>` and mutate it (`&self` should be changed to `&mut self` in `retry_update_l1`/`update_l1`
2. I think it would be nice to move `last_persisted_l1_ratio` out of `UpdateOnL1Params` and make it a field of `UpdateOnL1` enum variant. This way params will be immutable part, and `last_persisted_l1_ratio` will be changed on each iteration

ischasny: good points - addressed both

core/node/base_token_adjuster/src/base_token_ratio_persister.rs
RomanBrodetski: thanks! re: setting value on restart - I agree it's not very important, but for me it's not so much spending gas money, but more it doesn't feel right that the L1 behaviour depends on restarts. imagine someone looking at the list of updates - they always update when the price differs by 10%, but _sometimes_ just in random times.

Thanks for addressing!

core/node/base_token_adjuster/src/base_token_l1_behaviour.rs
```diff
@@ -0,0 +1,331 @@
...
+            // Increase `priority_fee_per_gas` by at least 20% to prevent "replacement transaction under-priced" error.
+            priority_fee_per_gas = max(priority_fee_per_gas, (x * 6) / 5 + 1);
+        }
+
+        if let Some(x) = prev_base_fee_per_gas {
+            // same for base_fee_per_gas but 10%
+            base_fee_per_gas = max(base_fee_per_gas, x + (x / 10) + 1);
+        }
+
+        // Extra check to prevent sending transaction will extremely high priority fee.
```
RomanBrodetski, line 280:
```suggestion
        // Extra check to prevent sending transaction with extremely high priority fee.
```

core/node/base_token_adjuster/src/base_token_l1_behaviour.rs
```diff
@@ -0,0 +1,331 @@
...
+        let prev_ratio = if let Some(prev_ratio) = last_persisted_l1_ratio {
+            prev_ratio.clone()
+        } else {
+            let prev_ratio = self.get_current_ratio_from_l1(l1_params).await?.clone();
+            self.update_last_persisted_l1_ratio(prev_ratio.clone());
+            tracing::info!(
+                "Fetched current base token ratio from the L1: {}",
+                prev_ratio.to_bigint().unwrap()
+            );
+            prev_ratio.clone()
```
perekopskiy, line 62: nit: looks like first and third clones and needless, not sure why lint check doesn't catch it

perekopskiy (dismissed):

core/node/base_token_adjuster/src/base_token_l1_behaviour.rs
```diff
@@ -0,0 +1,331 @@
...
+    pub diamond_proxy_contract_address: Address,
+    pub chain_admin_contract_address: Option<Address>,
+    pub config: BaseTokenAdjusterConfig,
+}
+
+#[derive(Debug, Clone)]
+pub enum BaseTokenL1Behaviour {
+    UpdateOnL1 {
+        params: UpdateOnL1Params,
+        last_persisted_l1_ratio: Option<BigDecimal>,
```
RomanBrodetski, line 38: does it need to be an `Option`? It's only `None` on startup, so we do initialization implicitely inside normal logic - and have a slightly more complicated state space (needing `Option` now). Could we request the persisted l1 ration on startup (in a `new` method) instead?

Well you'd probably still want to have several attempts of requesting it from l1, which may be complicated - but maybe not, especially if retry logic is abstracted away.

Ok, so overall not critical - feel free to leave as is

core/node/base_token_adjuster/src/base_token_l1_behaviour.rs
```diff
@@ -0,0 +1,331 @@
...
+                    .into_tokens(),
+            )
+            .context("failed encoding `setTokenMultiplier` input")?;
+
+        let nonce = (*l1_params.eth_client)
+            .as_ref()
+            .nonce_at_for_account(
+                l1_params.token_multiplier_setter_account_address,
+                BlockNumber::Pending,
+            )
```
RomanBrodetski, line 182: Maybe it's an edge case, but what if we send an underprice transaction, and then the server restarts? My understanding is that this transaction will be stuck forever. Do we want to think about this case?

perekopskiy: good catch, yes, let's use BlockNumber::Latest instead

ischasny: Well in the worst case multiplier is going to be updated twice then (which is not bad)

core/node/base_token_adjuster/src/base_token_l1_behaviour.rs
ischasny: Good point - let me make a ticket for that.

core/node/base_token_adjuster/src/base_token_l1_behaviour.rs
ischasny: Good catch - fixed

core/node/base_token_adjuster/src/base_token_l1_behaviour.rs
perekopskiy: The worst (and I'd say pretty realistic) case is that updating on L1 will be fully stuck

perekopskiy: IMO yes

ischasny: Ok just chatted to Roman - agreed, it should be Latest, changed and pushed. Thank you both for catching it.

perekopskiy (approved):

PR #2750, opened 2024-08-27
chore: simplify storage cache

joonazan:
Attempts to simplify how the cache of read storage slots is used to prepare for optimization that avoids accessing storage twice on storage write.

Removes bwip storage cache wrapper and implements ReadStorage on the wrapped value directly.


joonazan:
I have changed `unwrap_or_default` to `unwrap` because witness generation shouldn't require storage that execution didn't. The old code was dangerous especially in is_initial, as the default value of bool is false but true would be more appropriate.

joonazan:
Or maybe it was like that to not break on free storage slots? I hope we have tests for this.

core/lib/vm_interface/src/storage/storage_witness.rs
```diff
@@ -0,0 +1,21 @@
+use zksync_types::{witness_block_state::WitnessStorageState, StorageKey, StorageValue, H256};
+
+use super::ReadStorage;
+
+impl ReadStorage for WitnessStorageState {
```
slowli, line 5: What benefit does this change produce? From the engineering perspective, I'd say it makes things worse: Peculiar usage patterns of `WitnessStorageState` (wrapper) were confined in the witness generator crate before, but now they are still there, and the storage implementation is here (in another workspace!). This make maintaining this code more difficult (e.g., it's easier to forget about storage use invariants if they are changed in the witgen crate). Logically the prover workspace is downstream of the main one; I understand that there are common points out of necessity, but having proving-related logic infiltrating the main workspace is still bad.

slowli (commented):
Is the first part of the PR description ("simplify how the cache of read storage slots is used") obsolete? IIUC, it's not present in the PR 🙃

core/lib/vm_interface/src/storage/storage_witness.rs
joonazan: The reasoning is that storage does not provide witgen with some map, it provides witgen with a storage, so implementing the methods is the storage crate's job.

This also removes an unnecessary wrapper. Maybe even more layers can be removed; I haven't had time to look at the serialization layer.

PR #2759, opened 2024-08-28
feat(zk_toolbox): Migrate CI unit tests to zk_toolbox

matias-gonz:
## What ❔
Migrate CI unit tests to zk_toolbox


.github/workflows/ci-core-reusable.yml
```diff
@@ -58,13 +58,14 @@ jobs:
...
           ci_run zk contract build
+          ci_run ./bin/zkt
 
       - name: Contracts unit tests
         run: ci_run yarn l1-contracts test
 
       - name: Rust unit tests
         run: |
-          ci_run zk test rust
+          ci_run zk_supervisor test rust --link-to-code=/usr/src/zksync --test-server-url=postgres://postgres:notsecurepassword@localhost:5433/zksync_local_test --test-prover-url=postgres://postgres:notsecurepassword@localhost:5433/prover_local_test
```
Deniallugo, line 68: i don't think it's necessary. it's good enough to use default settings

zk_toolbox/crates/zk_supervisor/src/commands/test/rust.rs
```diff
@@ -1,32 +1,73 @@
...
-        .context(MSG_CHAIN_NOT_FOUND_ERR)?;
-    let general_config = chain.get_general_config()?;
-    let postgres = general_config
-        .postgres_config
-        .context(MSG_POSTGRES_CONFIG_NOT_FOUND_ERR)?;
+    let link_to_code = if let Some(link_to_code) = args.link_to_code {
+        PathBuf::from(link_to_code)
+    } else {
+        EcosystemConfig::from_file(shell)?.link_to_code
+    };
```
Deniallugo, line 27: do we really need it? 
Because ecosystem always exists

matias-gonz: True! We only needed the dbs, i will change this

Deniallugo:
let's remove `zk f` from `ci_run zk f cargo test --release -p vm-benchmark --bench oneshot --bench batch` and all zk installation

.github/workflows/ci-core-reusable.yml
```diff
@@ -52,22 +52,17 @@ jobs:
...
           ci_run run_retried rustup show
-          ci_run zk run yarn
-          ci_run zk db setup
-          ci_run zk compiler all
-          ci_run zk contract build
-
-      - name: Contracts unit tests
-        run: ci_run yarn l1-contracts test
+          ci_run yarn l1-contracts build
+          ci_run yarn l2-contracts build
```
Deniallugo, line 57: You have to install it with zk supervisor, and you also have to install  yarn. 
Moreover for l1 contracts we already use foundry, so please use it instead of hardhat

Deniallugo (approved):

onyxet (approved):

PR #2760, opened 2024-08-28
feat(vm): Simplify VM interface

slowli:
## What ❔

Simplifies low-level VM interface (i.e., `VmInterface` trait).

## Why ❔

To make it easier to use / maintain.

## Checklist


slowli:
This is a simple / less controversial part of refactoring. It still leaves multiple ways to execute transactions on the VM (`push_transaction` + `inspect` or `inspect_transaction`). Also, I'm not sure what `Bootloader` execution mode does; if it's removed, I *think* `inspect` could be removed altogether in favor of `finish_batch` and `inspect_transaction`.

core/node/state_keeper/src/batch_executor/main_executor.rs
```diff
@@ -205,7 +205,7 @@ impl CommandReceiver {
         }
 
         let tx_metrics = ExecutionMetricsForCriteria::new(Some(tx), &tx_result);
-        let gas_remaining = vm.gas_remaining();
+        let gas_remaining = tx_result.statistics.gas_remaining;
```
Deniallugo, line 208: is it equivalent?

slowli: Do you mean in theory, or in practice? AFAICT, semantically, it's the same value, and it's set correctly for all VMs other than M5, M6 and 1.3.2 (e.g., [here's the latest VM](https://github.com/matter-labs/zksync-era/blob/8ed086afecfcad30bfda44fc4d29a00beea71cca/core/lib/multivm/src/versions/vm_latest/implementation/statistics.rs#L34), where `gas_remaining_after` is obtained [here](https://github.com/matter-labs/zksync-era/blob/8ed086afecfcad30bfda44fc4d29a00beea71cca/core/lib/multivm/src/versions/vm_latest/implementation/execution.rs#L76)); but these 3 can be fixed to return a correct value as well. An alternative would be to add a `gas_remaining` field to `VmExecutionResultAndLogs`, but it looks obviously redundant there since it includes statistics as well. Maybe, I'm misinterpreting `VmExecutionStatistics` (like, are its fields approximations / filled at best-effort basis)?

core/lib/vm_interface/src/vm.rs
```diff
@@ -38,67 +32,48 @@ pub trait VmInterface {
...
-            final_bootloader_memory: Some(bootloader_memory),
-            pubdata_input: None,
-            state_diffs: None,
-        }
+    fn finish_batch(&mut self) -> FinishedL1Batch;
+}
+
+/// Extension trait for [`VmInterface`] that provides some additional methods.
+pub trait VmInterfaceExt: VmInterface {
+    /// Executes the next VM step (either next transaction or bootloader or the whole batch).
```
joonazan, line 56: What is the purpose of putting the methods with default implemenations into an extension trait?

popzxc: To improve usage experience without polluting the main trait (and allowing to unexpectedly alter its behavior); it's a common practice, see [FutureExt](https://docs.rs/futures/latest/futures/future/trait.FutureExt.html), [StreamExt](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html) (note that [Stream](https://docs.rs/futures/latest/futures/stream/trait.Stream.html) is defined within the same crate), [Itertools](https://docs.rs/itertools/latest/itertools/trait.Itertools.html).

popzxc (approved):
Nice!

core/lib/vm_interface/src/vm.rs
slowli: So that there's no ambiguousness of whether an implementation may implement these methods (e.g., for performance reasons). This, in turn, simplifies reasoning around implementations; note e.g. how I had to include optional methods in the shadow VM in case they were overridden by either of main and shadow VMs. In the general case, an extension trait could be non-object safe even if the extended trait is, but this is not relevant in this case.

core/node/state_keeper/src/batch_executor/main_executor.rs
slowli: I've added remaining gas tracking (I think correctly) for these 3 old VMs [in this commit](https://github.com/matter-labs/zksync-era/pull/2760/commits/056312f37c5e541098ccefbb8a234f0463cb42bd).

popzxc (approved):

Deniallugo (approved):

PR #2763, opened 2024-08-28
fix(api): Fix duplicate DB connection acquired in `eth_call`

slowli:
## What ❔

Fixes a duplicate DB connection getting acquired in the `eth_call` handler.

## Why ❔

Extra connection leads to performance degradation.

## Checklist


popzxc:
Thank you! This problem bugged me since forever. Glad to see it solved.

PR #2774, opened 2024-08-30
refactor(external-prover-api): Polish the API implementation

popzxc:
## What ❔

- Use `L1BatchProofForL1` instead of `VerifyProofRequest`.
- Rework errors: do not use "branching" variants that are handled separately in `IntoResponse`; instead use one variant per possible error.
- Use `thiserror` to improve ergonomics of errors.
- Do not use `Multipart` directly, instead use a dedicated type that implements `FromRequest`.
- Introduce `Api` structure to implement axum API (instead of procedural approach) -- aligns better with the framework design.
- Better separate `Processor` and `Api` in a way that `Processor` is backend-agnostic (e.g. know nothing about `axum`).
- Remove dependency on `zksync_config`.
- Improve framework integration.
- Other minor things.

## Why ❔

Ergonomics, maintainability, and readability.


Artemka374 (approved):
That looks very nice!

PR #2781, opened 2024-09-02
feat(zk_toolbox): Add zks contracts

matias-gonz:
## What ❔

Add zks contracts. Builds contracts


zk_toolbox/crates/zk_supervisor/src/commands/contracts.rs
```diff
@@ -0,0 +1,41 @@
...
+pub fn run(shell: &Shell) -> anyhow::Result<()> {
+    logger::info(MSG_BUILDING_CONTRACTS);
+
+    let ecosystem = EcosystemConfig::from_file(shell)?;
+    let link_to_code = ecosystem.link_to_code.clone();
+
+    let spinner = Spinner::new(MSG_CONTRACTS_DEPS_SPINNER);
+    let _dir_guard = shell.push_dir(&link_to_code);
+    Cmd::new(cmd!(shell, "yarn install")).run()?;
+    spinner.finish();
```
Deniallugo, line 20: can we add an ability to choose what contracts we want to build

Deniallugo: I think that optional enum, without multiple choices is better.

matias-gonz: 3 arguments instead of vec?

Deniallugo: I mean, it'd be more convinient to call it 
`zks contracts l1` instead of `zks contracts -c l1` and I don't see a lot of value, in having multiple arguments, because it's only 3 options

zk_toolbox/crates/zk_supervisor/src/commands/contracts.rs
```diff
@@ -9,13 +9,47 @@ use xshell::{cmd, Shell};
...
 #[derive(Debug, Parser)]
 pub struct ContractsArgs {
-    #[clap(long, short = 'c')]
-    pub contracts: Vec<ContractType>,
+    #[clap(long, help = MSG_BUILD_L1_CONTRACTS_HELP, default_missing_value = "true", num_args = 0..=1)]
+    pub l1_contracts: Option<bool>,
+    #[clap(long, help = MSG_BUILD_L2_CONTRACTS_HELP, default_missing_value = "true", num_args = 0..=1)]
+    pub l2_contracts: Option<bool>,
+    #[clap(long, help = MSG_BUILD_SYSTEM_CONTRACTS_HELP, default_missing_value = "true", num_args = 0..=1)]
+    pub system_contracts: Option<bool>,
```
Deniallugo, line 23: let's add aliases. 
`l1`,`l2`,`sc`

Deniallugo (approved):

PR #2782, opened 2024-09-02
perf(vm): Fix VM performance regression on CI loadtest

slowli:
## What ❔

Fixes VM performance regression on CI loadtest introduced in https://github.com/matter-labs/zksync-era/pull/2760.

## Why ❔

Changes in the VM interface made the VM eagerly clone compressed bytecodes if compression hasn't failed. Compressed bytecodes aren't used during sandboxed VM execution in the API server (the sandbox only checks that compression is successful).For new VMs, bytecodes can be borrowed from the VM state, which is what this PR does using `Cow`.

## Checklist

### DIFF
--- a/.github/workflows/ci-core-reusable.yml
+++ b/.github/workflows/ci-core-reusable.yml
@@ -91,7 +91,7 @@ jobs:
       - name: Loadtest configuration
         run: |
           echo EXPECTED_TX_COUNT=${{ matrix.vm_mode == 'new' && 22000 || 16000 }} >> .env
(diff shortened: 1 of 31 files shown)


slowli:
Not sure whether it's just wishful thinking, but the changes in the load test config seem to make the load test more stable (decreasing the number of accounts and concurrent API requests; increasing the number of in-flight txs and randomizing polling interval between polling tx status).

Deniallugo (approved):

onyxet (approved):

alexandrst88 (approved):

PR #2786, opened 2024-09-03
feat(genesis): Validate genesis config against L1

perekopskiy:
## What ❔

Validate protocol version and vk hash from genesis config against L1

## Why ❔

Right now nothing prevents from initializing contracts and node with different protocol versions or vk hashs, and it already happened 🥲 

## Checklist

### DIFF
--- a/core/node/genesis/src/lib.rs
+++ b/core/node/genesis/src/lib.rs
@@ -6,9 +6,12 @@ use std::fmt::Formatter;
 
 use anyhow::Context as _;
 use zksync_config::GenesisConfig;
-use zksync_contracts::{BaseSystemContracts, BaseSystemContractsHashes, SET_CHAIN_ID_EVENT};
+use zksync_contracts::{
+    hyperchain_contract, verifier_contract, BaseSystemContracts, BaseSystemContractsHashes,
+    SET_CHAIN_ID_EVENT,
+};
 use zksync_dal::{Connection, Core, CoreDal, DalError};
-use zksync_eth_client::EthInterface;
+use zksync_eth_client::{CallFunctionArgs, EthInterface};
 use zksync_merkle_tree::{domain::ZkSyncTree, TreeInstruction};
 use zksync_multivm::utils::get_max_gas_per_pubdata_byte;
 use zksync_system_constants::PRIORITY_EXPIRATION;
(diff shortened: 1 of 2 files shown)


core/node/node_storage_init/src/main_node/genesis.rs
```diff
@@ -30,6 +30,12 @@ impl InitializeStorage for MainNodeGenesis {
         }
 
         let params = GenesisParams::load_genesis_params(self.genesis.clone())?;
+        zksync_node_genesis::validate_genesis_params(
+            &params,
+            &self.l1_client,
+            self.contracts.diamond_proxy_addr,
+        )
+        .await?;
```
popzxc, line 38: This check would only be performed on a fresh setup, so if someone will change the parameters afterwards -- it won't be checked. Is that expected?

core/node/genesis/src/lib.rs
```diff
@@ -110,11 +113,8 @@ impl GenesisParams {
                 },
             )));
         }
-        // Try to convert value from config to the real protocol version and return error
-        // if the version doesn't exist
-        let _: ProtocolVersionId = config
+        let _ = config
             .protocol_version
-            .map(|p| p.minor)
             .ok_or(GenesisError::MalformedConfig("protocol_version"))?;
```
popzxc, line 118: Nit: `if config.protocol_version.is_none() { return Err(..) }`

core/node/genesis/src/lib.rs
```diff
@@ -264,6 +264,55 @@ pub async fn is_genesis_needed(storage: &mut Connection<'_, Core>) -> Result<boo
...
+) -> anyhow::Result<()> {
+    let hyperchain_abi = hyperchain_contract();
+    let verifier_abi = verifier_contract();
+
+    let packed_protocol_version: U256 = CallFunctionArgs::new("getProtocolVersion", ())
+        .for_contract(diamond_proxy_address, &hyperchain_abi)
+        .call(query_client)
+        .await?;
+
+    let (minor_version, patch_version) = packed_protocol_version.div_mod((1u64 << 32).into());
```
popzxc, line 280: Is there no dedicated function for that? If we will ever have to use `major`, catching all such places would be troublesome.

core/node/node_storage_init/src/main_node/genesis.rs
perekopskiy: Yes, they are inserted into db after fresh setup so it's fine if they are changed later as long as genesis config is not used (it shouldn't be changed nor used)

popzxc (approved):

PR #2798, opened 2024-09-04
fix(tee): lowercase enum TEE types

pbeza:
## What ❔

We encountered an issue where the staging environment was unable to deserialize `sgx` to `TeeType::Sgx`. Relevant code:
- https://github.com/matter-labs/zksync-era/blob/87b02e3ab5c1f61d59dd0f0eefa9ec33a7b55488/core/lib/basic_types/src/tee_types.rs#L7
- https://github.com/matter-labs/teepot/blob/537521f0ee2bd704fb839fe336f43f8aab5887df/bin/tee-key-preexec/src/main.rs#L53

Relevant logs:
- https://grafana.matterlabs.dev/goto/Q5ENugeSR?orgId=1

## Why ❔

To fix a panic in the staging environment.

## Checklist


core/lib/basic_types/src/tee_types.rs
```diff
@@ -2,8 +2,57 @@ use serde::{Deserialize, Serialize};
 use strum::{Display, EnumString};
 
 #[derive(Debug, Clone, Copy, PartialEq, EnumString, Display, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
+#[strum(serialize_all = "snake_case")]
 #[non_exhaustive]
 pub enum TeeType {
-    #[strum(serialize = "sgx")]
     Sgx,
 }
```
popzxc, line 10: Why both `strum` and `serde` tho? I think the real problem here is that the type was _serialized_ with `strum` and deserialized with `serde`.
So wouldn't it be better to just use `serde` to reduce risks of it going out of sync again?

pbeza: [Fixed](https://github.com/matter-labs/zksync-era/pull/2798/commits/e524817b6c77f71631b82a453b80c34a2048ed80). ✅

core/lib/basic_types/src/tee_types.rs
```diff
@@ -1,9 +1,71 @@
...
+        assert_eq!(TeeType::Sgx.to_string(), "sgx");
+    }
+
+    #[test]
+    fn test_envy_teetype() {
+        std::env::set_var("TEE_TYPE", "sgx");
+
+        let config = envy::from_env::<Config>().unwrap();
+        assert_eq!(config.tee_type, TeeType::Sgx);
+    }
```
popzxc, line 70: This test is not needed. It just calls `serde` under the hood.
On a separate note, env variables are global state, so this test may break other tests and can break itself (unless access to env vars is put under a mutes.

pbeza: That test was suggested to be added by @haraldh. I just removed it [here](https://github.com/matter-labs/zksync-era/pull/2798/commits/72f96936edc2a21bf8df579e0d954277fd4dc390). ✅ But to be fair, he also suggested the same code change as you did [here](https://github.com/matter-labs/zksync-era/pull/2798#discussion_r1743537744).

core/lib/basic_types/src/tee_types.rs
```diff
@@ -1,9 +1,58 @@
...
+}
+
+impl FromStr for TeeType {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        serde_json::from_str(&format!("\"{}\"", s.to_lowercase()))
+            .map_err(|_| format!("Invalid value for TeeType: {}", s))
+    }
```
popzxc, line 26: Sorry for many iterations of review, but:
1. Why do we need to allocate a string and then deserialize it just to get the value? Why not check on static strings?
2. It is somewhat unconventional to have case-insensitive logic in `FromStr`. I would prefer the logic to be strict, and if we don't know if the string is lowercased, then the _caller_ of this method can call `TeeType::from_str(s.to_lowercase())`.

core/lib/basic_types/src/tee_types.rs
```diff
@@ -1,9 +1,58 @@
...
+impl fmt::Display for TeeType {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            TeeType::Sgx => write!(f, "sgx"),
+        }
+    }
+}
+
+impl FromStr for TeeType {
+    type Err = String;
```
popzxc, line 21: `String` is not a good error type. `anyhow::Error` would work better here.

core/lib/basic_types/src/tee_types.rs
```diff
@@ -1,9 +1,58 @@
...
     Sgx,
 }
+
+impl fmt::Display for TeeType {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            TeeType::Sgx => write!(f, "sgx"),
+        }
+    }
+}
```
popzxc, line 18: What do we need `Display` implementation for here? Display is [meant for user-facing output](https://doc.rust-lang.org/std/fmt/trait.Display.html), and I don't think we utilize this type for that purpose.
If it's used for _serialization_ (e.g. to store and then get data), then `serde` should be used instead.
For printing in logs, `Debug` is preferred.

If `Display` and `FromStr` are used to serialize and deserialize the type but not for anything else, I'd suggest removing this logic and rely on `serde` everywhere instead.

core/lib/basic_types/src/tee_types.rs
pbeza: I completely removed the implementation of the std::str::FromStr trait because we don't use it anyway.

core/lib/basic_types/src/tee_types.rs
pbeza: I completely removed the implementation of the `std::str::FromStr` trait because we don't use it anyway.

core/lib/basic_types/src/tee_types.rs
pbeza: As discussed offline, we need `Display` implementation in [`core/lib/dal/src/tee_proof_generation_dal.rs`](https://github.com/matter-labs/zksync-era/pull/2798/commits/ae107002f1d526df084b078cd7f66876eb9aefb0#diff-46d00b49fb68d6da13c9ecdc5ab3dab6adf502ec216349721586e96553652444).

popzxc (approved):

PR #2804, opened 2024-09-04
test: Fix "missing revert data" error; fix / debug integration tests

slowli:
## What ❔

- Fixes the "missing revert data" error by updating the used reth Docker image. The error is probably caused by [this issue](https://github.com/paradigmxyz/reth/issues/7381) fixed in the new reth versions.
- Removes "web3 API compatibility tests › Should check API returns error when there are too many logs in eth_getLogs" test as fundamentally flaky and able to poison other tests.
- Adds logging for upgrade test to investigate L1 "nonce too low" errors.

## Why ❔

Flaky CI bad.

## Checklist


slowli:
Got [this new error](https://github.com/matter-labs/zksync-era/actions/runs/10715375714/job/29710731744?pr=2804#step:8:36086):

```text
  ● web3 API compatibility tests › Should check API returns error when there are too many logs in eth_getLogs

    expect(received).rejects.toThrow(expected)

    Expected substring: "Query returned more than 10000 results."
    Received message:   "missing response for request (value=[ { \"error\": { \"code\": -32011, \"data\": \"Exceeded max limit of 10485760\", \"message\": \"The batch response was too large\" }, \"id\": null, \"jsonrpc\": \"2.0\" } ], info={ \"payload\": { \"id\": 7967, \"jsonrpc\": \"2.0\", \"method\": \"eth_getLogs\", \"params\": [ { \"fromBlock\": \"0x61\", \"toBlock\": \"0x73\" } ] } }, code=BAD_DATA, version=6.12.1)"

  ● web3 API compatibility tests › Should check API returns correct block for every tag

    missing response for request (value=[ { "error": { "code": -32011, "data": "Exceeded max limit of 10485760", "message": "The batch response was too large" }, "id": null, "jsonrpc": "2.0" } ], info={ "payload": { "id": 7966, "jsonrpc": "2.0", "method": "eth_getLogs", "params": [ { "fromBlock": "0x61", "toBlock": "0x61" } ] } }, code=BAD_DATA, version=6.12.1)
```

Besides some fixes in the test logic (it doesn't await for some `Promise`s currently), it looks weird; it's as if a single transaction (the third one with many logs) generates >10M of logs, *and* the request limiting doesn't work properly.

core/tests/ts-integration/tests/api/web3.test.ts
```diff
@@ -355,7 +355,6 @@ describe('web3 API compatibility tests', () => {
         // ...though the gap should not be *too* big.
         expect(newBlock).toBeLessThan(currentBlock + 100);
         await tx.wait(); // To not leave a hanging promise.
-        wsProvider.removeAllListeners();
```
slowli, line 358: This call seems useless if not awaited (the unsubscribe RPC requests scheduled internally likely won't be sent since the WebSocket is closed immediately). But if it is awaited, then `destroy()` below will fail because unsubscribe RPC requests get rejected. I haven't found a good way to wait until these requests are sent (something like `await wsProvider.sendPendingRequests()`), so the only option seems to have a hard-coded timeout, which doesn't look good.

core/tests/upgrade-test/tests/upgrade.test.ts
```diff
@@ -280,9 +280,11 @@ describe('Upgrade test', function () {
         );
         executeOperation = chainUpgradeCalldata;
 
+        console.log('Sending scheduleTransparentOperation');
```
slowli, line 283: Motivated by this test [sometimes failing](https://github.com/matter-labs/zksync-era/actions/runs/10715375714/job/29717077740?pr=2804) with "nonce has already been used" error on one of the operations below.

core/bin/zksync_server/src/node_builder.rs
```diff
@@ -364,6 +364,7 @@ impl MainNodeBuilder {
             subscriptions_limit: Some(rpc_config.subscriptions_limit()),
             batch_request_size_limit: Some(rpc_config.max_batch_request_size()),
             response_body_size_limit: Some(rpc_config.max_response_body_size()),
+            with_extended_tracing: rpc_config.extended_api_tracing,
```
slowli, line 367: Maybe this was skipped intentionally, but I don't see a particular reason why the WS server should have extended tracing and the HTTP one shouldn't.

slowli:
Looks like now there's an issue with using expired nonces on L1 ([1](https://github.com/matter-labs/zksync-era/actions/runs/10715375714/job/29717077740?pr=2804), [2](https://github.com/matter-labs/zksync-era/actions/runs/10719479298/job/29723661260?pr=2804), [3](https://github.com/matter-labs/zksync-era/actions/runs/10721704587/job/29730904726?pr=2804)).

perekopskiy:
Was there any "expired nonce" failure outside this PR?

slowli:
@perekopskiy Yes, there were ([1](https://github.com/matter-labs/zksync-era/actions/runs/10617480068/job/29430442131#step:8:20521), [2](https://github.com/matter-labs/zksync-era/actions/runs/10628748919/job/29464231503#step:9:16386), both from last week), but subjectively, they've become more frequent (as mentioned above, I've encountered them 3 times on this PR alone).

core/bin/zksync_server/src/node_builder.rs
popzxc: Either I missed it, or it was configured this way in the old API server configuration code (there were other desynchronized configs in there too). Either way, it was not intentional.

perekopskiy (approved):

PR #2806, opened 2024-09-05
refactor(api): Extract oneshot VM executor to executor crate

slowli:
## What ❔

Extracts oneshot VM executor to the executor crate.

## Why ❔

To make executor logic more reusable and maintainable.

## Checklist


core/lib/vm_interface/src/types/tracer.rs
```diff
@@ -37,3 +41,78 @@ pub enum VmExecutionStopReason {
     VmFinished,
     TracerRequestedStop(TracerExecutionStopReason),
 }
+
+/// Transaction validation parameters.
+#[derive(Debug, Clone)]
+pub struct ValidationParams {
```
slowli, line 47: I'm not sure this belongs to the `tracer` module. Logically, I think validation doesn't necessary correlate to VM tracing (hence, the separate `TransactionValidator` trait).

core/node/api_server/src/execution_sandbox/apply.rs
```diff
@@ -30,18 +23,15 @@ use zksync_system_constants::{
...
+use zksync_utils::{h256_to_u256, time::seconds_since_epoch};
 
 use super::{
-    vm_metrics::{self, SandboxStage, SANDBOX_METRICS},
-    ApiTracer, BlockArgs, OneshotExecutor, TxExecutionArgs, TxSetupArgs,
+    vm_metrics::{SandboxStage, SANDBOX_METRICS},
+    BlockArgs, TxSetupArgs,
 };
 
 pub(super) async fn prepare_env_and_storage(
```
slowli, line 37: With some creativity, `BlockArgs`, `TxSetupArgs` and this logic can be (mostly) moved to the executor crate. Didn't do that to keep the PR size reasonable, and because it's not *completely* trivial; e.g., I'd like to isolate the moved code from knowing that `BlockArgs` are obtained from `api::BlockId` etc., so that oneshot executor isn't 100% tied to the API server domain.

core/node/api_server/src/execution_sandbox/execute.rs
```diff
@@ -114,72 +51,99 @@ impl TransactionExecutor {
...
                     .await
             }
-            Self::Mock(executor) => executor.inspect_transaction(storage, env, args, ()).await,
         }
     }
+}
 
-    async fn inspect_transaction_with_bytecode_compression(
+#[async_trait]
+impl<S> TransactionValidator<S> for TransactionExecutor
```
slowli, line 127: It seems possible to use dynamically dispatched executor instead (`Box<dyn TransactionValidator< StorageWithOverrides>>`), but this can have performance implications, so haven't done that.

perekopskiy (dismissed):

popzxc (dismissed):
🔥 🔥 🔥

slowli:
Recently merged consensus-related PR https://github.com/matter-labs/zksync-era/pull/2684 is an obvious candidate to use the oneshot executor. But since this change is non-trivial and requires moving more types from the API server crate as I've described above, it makes sense to do separately.

popzxc (approved):

PR #2800, opened 2024-09-04
feat(zk_toolbox): Allow running docker images for provers

Artemka374:
## What ❔

Add `--docker` flag for `zk_inception prover run` which will allow prover components to run from docker images.

## Why ❔

To decrease setup time/improve UX

## Checklist

- [ ] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [ ] Tests for the changes have been added / updated.
- [ ] Documentation comments have been added / updated.
- [ ] Code has been formatted via `zk fmt` and `zk lint`.

PR #2811, opened 2024-09-05
feat(zk_toolbox): Add setup keys step to prover init

matias-gonz:
## What ❔
Add setup keys step to prover init


Deniallugo (approved):
LGTM, but i'm waiting for @EmilLuta or @Artemka374

Artemka374:
I don't think that generating/downloading setup keys by default is a good idea. It can take a lot of time(generating ~40minutes/downloading depends on your internet connection, but it is 26GB anyway), and it is needed only for FRI provers. Instead, I'd add an option to choose whether you want to download setup keys right now or not. The same also refers to initializing bellman-cuda - for that I need drivers/CUDA and other things, but it is used only in compressor. I don't need it e.g. [here](https://github.com/matter-labs/zksync-era/pull/2800), where bellman cuda is baked into compressor's image. Same way it can be changed for downloading compressor setup key, but it is small, so it definitely wouldn't hurt to have it by default

matias-gonz:
Hello @Artemka374 right now it is optional to download/generate setup keys and the [default is false](https://github.com/matter-labs/zksync-era/pull/2811/files#diff-18bdfce168eddebd57cf0d13a909d5e2209ff2bb7be93c8d348be5c1a0dd0b12R375-R379)

I will make downloading the compressor key and setting up bellman-cuda default false too.

Artemka374:
@matias-gonz I'd suggest not to make compressor key dowloading optional because we don't have alternative command to do that later... So it's either leave it as is, or add separate command for this

matias-gonz:
@Artemka374 compressor key downloading was already optional. I will add a separate command

Artemka374 (approved):
LGTM!

Deniallugo (approved):

PR #2814, opened 2024-09-06
chore(ci): Fix cargo deny check and make output readable

popzxc:
## What ❔

- Allow new advisory (low impact; hard to fix).
- Omit printing the tree in output; because with it the output is not readable.

## Why ❔

- Unblock CI
- Make cargo deny CI usable.


deny.toml
```diff
@@ -51,7 +51,7 @@ ignore = false
 registries = []
 
 [bans]
-multiple-versions = "warn"
+multiple-versions = "allow"
```
popzxc, line 54: We inherently depend on multiple versions (because of multivm), so it cannot realistically be fixed without a humongous effort.

hatemosphere (approved):

PR #2817, opened 2024-09-06
fix(tee-prover): fix deserialization of `std::time::Duration` in `envy` config

pbeza:
## What ❔

This PR fixes the deserialization issue of `std::time::Duration` in the `envy` configuration.

Relevant logs from the `stage` environment showcasing the issue:
https://grafana.matterlabs.dev/goto/IC-9k4eIR?orgId=1

Error message from the above logs:
```
Error: missing value for field initial_retry_backoff
```

The root cause of the problem supposedly boils down to the mismatch between the expected format of `TEE_PROVER_INITIAL_RETRY_BACKOFF` and the actual format. We export it as follows:
```
export TEE_PROVER_INITIAL_RETRY_BACKOFF=1
```
which is not supported as explained here:
https://github.com/serde-rs/serde/issues/339#issuecomment-539453327

## Why ❔

To fix the bug.

## Checklist

### DIFF
--- a/core/bin/zksync_tee_prover/src/config.rs
+++ b/core/bin/zksync_tee_prover/src/config.rs
@@ -2,11 +2,13 @@ use std::{path::PathBuf, time::Duration};
 
 use secp256k1::SecretKey;
 use serde::Deserialize;
+use serde_with::{serde_as, DurationSecondsWithFrac};
 use url::Url;
 use zksync_env_config::FromEnv;
 use zksync_types::tee_types::TeeType;
 
 /// Configuration for the TEE prover.
+#[serde_with::serde_as]
 #[derive(Debug, Clone, Deserialize)]
 pub(crate) struct TeeProverConfig {
     /// The private key used to sign the proofs.
@@ -22,10 +24,12 @@ pub(crate) struct TeeProverConfig {
     pub max_retries: usize,
     /// Initial back-off interval when retrying recovery on a retriable error. Each subsequent retry interval
     /// will be multiplied by [`Self.retry_backoff_multiplier`].
+    #[serde_as(as = "DurationSecondsWithFrac")]
     pub initial_retry_backoff: Duration,
     /// Multiplier for the back-off interval when retrying recovery on a retriable error.
     pub retry_backoff_multiplier: f32,
     /// Maximum back-off interval when retrying recovery on a retriable error.
+    #[serde_as(as = "DurationSecondsWithFrac")]
     pub max_backoff: Duration,
 }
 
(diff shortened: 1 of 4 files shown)


pbeza:
@popzxc please review when you get a chance. Sorry, I must have overlooked it while testing it locally.

To give you more context – that's odd that we get this error on `stage` (see the [relevant logs](https://grafana.matterlabs.dev/goto/IC-9k4eIR?orgId=1)):
```
Error: missing value for field initial_retry_backoff
```
if we have it defined `TEE_PROVER_INITIAL_RETRY_BACKOFF` [here](https://github.com/matter-labs/gitops-kubernetes/blob/190fe318ab52183abd36c40ba98ff20717931f00/apps/clusters/azure-staging/stage2/zksync-tee-prover/zksync-tee-prover.yaml#L21-L22):
```
      - name: TEE_PROVER_INITIAL_RETRY_BACKOFF
        value: 1
```
I would rather expect this error instead:
```
Error: invalid type: string "1", expected struct Duration
```

I hope `stage` will correctly pick up the `TEE_PROVER_INITIAL_RETRY_BACKOFF` variable once we merge this PR. 🤞🏼

artmakh (dismissed):

core/bin/zksync_tee_prover/src/config.rs
```diff
@@ -22,10 +24,12 @@ pub(crate) struct TeeProverConfig {
     pub max_retries: usize,
     /// Initial back-off interval when retrying recovery on a retriable error. Each subsequent retry interval
     /// will be multiplied by [`Self.retry_backoff_multiplier`].
+    #[serde_as(as = "DurationSecondsWithFrac")]
     pub initial_retry_backoff: Duration,
```
popzxc, line 28: We usually load timeouts like seconds or milliseconds (and add corresponding postfix, e.g. `_sec`), so that the value is predictably stored as integer, and add a convenience getter, e.g. `fn initial_retry_backoff(&self) -> Duration { Duration::from_secs(self.initial_retry_backoff_sec) }`.

popzxc (approved):

haraldh (approved):

core/bin/zksync_tee_prover/src/config.rs
pbeza: I prefer to introduce abstractions (like `Duration`) as early as possible to avoid the need for verbose intermediary getters that act as translators between abstraction layers (`u64` <--> `Duration`), but that's fine – I can stick to the conventions we have here. 🫡

popzxc (approved):

PR #2822, opened 2024-09-06
test: Improve revert integration test

slowli:
## What ❔

- Removes a data race from the revert integration test (a single node one); a batch gets executed after it is verified that not all batches are executed.
- Removes copy-paste between single-node and EN tests.
- Structures revert tests into smaller steps.

## Why ❔

Makes the test less flaky and easier to maintain.

## Checklist


etc/utils/src/logs.ts
```diff
@@ -1,7 +1,7 @@
 import path from 'path';
 import fs from 'node:fs/promises';
 
-const pathToHome = path.join(__dirname, '../../../..');
+const pathToHome = path.join(__dirname, '../../..');
```
slowli, line 4: This may break something in CI or zk_toolbox, but the previous location looks incorrect; it's the *parent* directory for the Era repo, not the repo directory itself.

core/tests/revert-test/tests/utils.ts
```diff
@@ -100,3 +108,389 @@ export function exec(command: string, options: ProcessEnvOptions) {
...
+    }
+
+    public async spawnExtNode(): Promise<Node<NodeType.EXT>> {
+        const env = this.env ?? process.env;
+        const { pathToHome, fileConfig, logs, options } = this;
+
+        let args = []; // FIXME: unused
+        if (options.enableConsensus) {
+            args.push('--enable-consensus');
+        }
```
slowli, line 358: Is this intentional that `args` aren't passed anywhere? [Here's this code in `main`](https://github.com/matter-labs/zksync-era/blob/b8d4424c0be72a22df0c2a828785442822825b21/core/tests/revert-test/tests/revert-and-restart-en.test.ts#L264-L268).

slowli (commented):
Not 100% sure this will work with `zk_toolbox`; all relevant CI jobs so far have failed *before* getting to a revert test (e.g. [hung up on integration tests](https://github.com/matter-labs/zksync-era/actions/runs/10741422424/job/29862337123?pr=2822) or [failed minting the custom token](https://github.com/matter-labs/zksync-era/actions/runs/10741422424/job/29857891766?pr=2822)), which I don't think is related to this PR. Will try to set up `zk_toolbox` locally and run tests in this way.

core/tests/revert-test/tests/utils.ts
Deniallugo: I think it's not we have to propagate it

perekopskiy (approved):

popzxc (approved):

Deniallugo (approved):

core/tests/revert-test/tests/utils.ts
slowli: Will do so separately then to not block merging this PR.

PR #2837, opened 2024-09-10
feat(zk_toolbox): deploy legacy bridge

Deniallugo:
## What ❔

Support legacy bridges in zk toolbox 

## Why ❔

We still have some legacy bridges in production, so we have to have an opportunity to test it 

## Checklist


zk_toolbox/crates/zk_supervisor/src/commands/test/mod.rs
```diff
@@ -43,6 +44,8 @@ pub enum TestCommands {
     Prover,
     #[clap(about = MSG_TEST_WALLETS_INFO)]
     Wallet,
+    #[clap(about = MSG_TEST_WALLETS_INFO)]
```
perekopskiy, line 47: please change message

zk_toolbox/crates/zk_inception/src/commands/args/run_server.rs
```diff
@@ -17,4 +17,6 @@ pub struct RunServerArgs {
     additional_args: Vec<String>,
     #[clap(long, help = MSG_SERVER_BUILD_HELP)]
     pub build: bool,
+    #[clap(help=MSG_SERVER_URING_HELP, long, default_missing_value = "true", num_args = 0..=1)]
```
perekopskiy, line 20: zk tool uses false by default, do you think it should be true?

alexandrst88 (dismissed):

zk_toolbox/crates/zk_inception/src/commands/args/run_server.rs
Deniallugo: here false is by default, just `--uring` makes it true

perekopskiy: ah, got it, thanks

zk_toolbox/crates/zk_supervisor/src/commands/test/mod.rs
Deniallugo: fixed

alexandrst88 (approved):

perekopskiy (approved):

PR #2840, opened 2024-09-10
refactor(vm-runner): Improve VM runner / VM playground

slowli:
## What ❔

Various minor improvements to VM runner / VM playground:

- Get batch storage asynchronously, so that it works efficiently with snapshot storage.
- Add metrics / logs to ensure that snapshot storage works as expected.

## Why ❔

Improves usability.

## Checklist

### DIFF
--- a/core/node/vm_runner/src/process.rs
+++ b/core/node/vm_runner/src/process.rs
@@ -1,19 +1,21 @@
 use std::{sync::Arc, time::Duration};
 
 use anyhow::Context;
-use tokio::{sync::watch, task::JoinHandle};
+use tokio::{
+    sync::{watch, Mutex},
+    task::JoinHandle,
+};
 use zksync_dal::{ConnectionPool, Core};
 use zksync_state::OwnedStorage;
-use zksync_types::{block::L2BlockExecutionData, L1BatchNumber};
-use zksync_vm_interface::{
-    executor::{BatchExecutor, BatchExecutorFactory},
-    L2BlockEnv,
-};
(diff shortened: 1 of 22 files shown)


core/node/vm_runner/src/process.rs
```diff
@@ -148,39 +177,16 @@ impl VmRunner {
...
-
-            self.io
-                .mark_l1_batch_as_processing(&mut self.pool.connection().await?, next_batch)
-                .await?;
-            let handle = tokio::task::spawn(Self::process_batch(
-                batch_executor,
-                batch_data.l2_blocks,
-                output_handler,
-            ));
+            let handle = tokio::spawn(self.clone().process_batch(next_batch));
```
slowli, line 188: I'm not 100% sure this change is correct for existing VM runner components other than the VM playground. With this async pipeline, it means that `mark_l1_batch_as_processing()` can be called out of order (e.g., first for batch 5, then 3, then 4).

itegulov: The order of `mark_l1_batch_as_processing` shouldn't matter as long as `mark_l1_batch_as_completed` is sequential. All existing implementations use processing stage purely for metrics/observability. Maybe we should reflect that the implementation should not rely on order of execution in `mark_l1_batch_as_processing`'s rustdoc

itegulov: Ah, sorry, I skimmed through the code somehow missed that

itegulov (approved):
Big fan of the changes to VM runner!

PR #2844, opened 2024-09-11
feat: Selector generator tool

mm-zk:
## What ❔

* A small tool to generate the selector hashes based on the ABI from json files

## Why ❔

* The output json can be useful for humans to better understand some of the errors (and calldata)
* It can also be read by our tools, to make the debugging easier.

In the future, we could call this tool regularly on each contracts version change, but for now it can stay as manual.


core/bin/selector_generator/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+categories.workspace = true
+publish = false
+
+
+[dependencies]
+serde = { workspace = true, features = ["derive"] }
+serde_json.workspace = true
+sha3.workspace = true
+glob.workspace = true
+clap = { workspace = true,  features = ["derive"] }
```
popzxc, line 19: Nit: Empty line.

core/bin/selector_generator/selectors.json
popzxc, line 1: Could you move it to `etc/selector_generator_data/` and add a README.md there stating that this is a directory used by the `[selector_generator](../core/bin/selector_generator)` crate?

core/bin/selector_generator/src/main.rs
```diff
@@ -0,0 +1,105 @@
+use std::{
+    collections::HashMap,
+    fs::{File, OpenOptions},
+    io::{self},
+};
+
+use clap::Parser;
+use glob::glob;
+use serde::{Deserialize, Serialize};
+use serde_json::{from_reader, to_writer_pretty, Value};
```
popzxc, line 10: Nit: We usually use fully qualified paths for short-names functions/macros from common crates like `serde_json`, e.g. `serde_json::from_str`/`tracing::info!`. Value is up to use, but can use a fully qualified path too.

koloz193:
can this be extended to also generate a file of selector to custom error? that would be much better than the current 4 byte array we propagate on vm failures 😄

mm-zk:
@koloz193 - sure we can add the custom errors support in next PRs.

popzxc:
> this is a good candidate for zk_supervisor.

I also thought so initially, but then decided that probably not.
1. We want to have tools that are actually used there. Otherwise it'll become a pile of dead code after some time.
2. This tool updates a generally static file. Not sure if people outside of ML would want to run it.
3. `zks` already has a lot of functionality, so adding anything new should be done with caution.

Let's wait a bit and see how it's used, and only then add it there.

popzxc:
I've removed this PR from the queue so that we [merge the fix PR ](https://github.com/matter-labs/zksync-era/pull/2855) ASAP

PR #2846, opened 2024-09-11
feat(state-keeper): pre-insert unsealed L1 batches

itegulov:
## What ❔

`MempoolIO` and `ExternalIO` pre-insert unsealed batches into DB as soon as they open them.

I chose to populate non-null fields with default values to minimize the impact of this PR, but I am open for a discussion if anyone thinks we need to make those fields nullable (specifically `l1_tx_count`, `l2_tx_count`, `bloom`, `priority_ops_onchain_data`, `initial_bootloader_heap_content`, and `used_contract_hashes`).

## Why ❔

* Better visibility of what's going on for node operators
* This opens a road to fix the issue with gas fluctuations on chains with low amounts of traffic as `api_server` will have access to the most up-to-date fee input

## Checklist


core/lib/dal/src/sync_dal.rs
```diff
@@ -35,6 +35,8 @@ impl SyncDal<'_, '_> {
                             (MAX(number) + 1)
                         FROM
                             l1_batches
+                        WHERE
+                            is_sealed
```
perekopskiy, line 39: [Out of this PR scope]
As you can see the pattern `COALESCE(miniblocks.l1_batch_number, SELECT (MAX(number) + 1) FROM l1_batches WHERE is_sealed)` is quite common. We can consider saving l1 batch number when inserting miniblock. This should simplify such places. Though we should be careful, AFAIR we use `miniblocks.l1_batch_number IS NULL` to get miniblocks in pending batch (e.g. to re-execute them in state-keeper). We can replace it with `SELECT ... FROM miniblocks JOIN l1_batches ON miniblocks.l1_batch_number = l1_batch.number WHERE l1_batch.is_sealed = FALSE` but I don't know how indexing will work for such query.
Please investigate indexing and consider this option

core/node/genesis/src/lib.rs
```diff
@@ -413,6 +413,14 @@ pub async fn create_genesis_l1_batch(
     transaction
         .blocks_dal()
         .insert_l1_batch(
+            genesis_l1_batch_header.number,
+            genesis_l1_batch_header.timestamp,
+            BatchFeeInput::default(),
```
perekopskiy, line 418: `BatchFeeInput::default()` returns `L1Pegged` variant which is legacy one. I'm not sure if there is a reason why `default` does so. I would explicitly create `PubdataIndependent` variant.

core/lib/dal/src/blocks_dal.rs
```diff
@@ -2481,8 +2556,10 @@ impl BlocksDal<'_, '_> {
         Ok(())
     }
 
-    pub async fn insert_mock_l1_batch(&mut self, header: &L1BatchHeader) -> DalResult<()> {
-        self.insert_l1_batch(
+    pub async fn insert_mock_l1_batch(&mut self, header: &L1BatchHeader) -> anyhow::Result<()> {
+        self.insert_l1_batch(header.number, header.timestamp, BatchFeeInput::default())
```
perekopskiy, line 2560: let's avoid using `BatchFeeInput::default()`, same for other occurrences in this file

core/lib/dal/src/blocks_dal.rs
```diff
@@ -176,7 +183,8 @@ impl BlocksDal<'_, '_> {
             FROM
                 l1_batches
             WHERE
-                hash IS NOT NULL
+                is_sealed
+                AND hash IS NOT NULL
```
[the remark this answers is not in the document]
itegulov: Sorry you had to review the PR in its current form. I added all these extra `is_sealed` clauses in https://github.com/matter-labs/zksync-era/pull/2846/commits/2542b092ae4209fd478929e2794f69de9e8b6c1a to gauge if that's the reason why a couple of tests were not passing (that was indeed the case, one of the queries I thought was safe actually needed `is_sealed`). I will revert those changes.

core/node/genesis/src/lib.rs
itegulov: If you look at the start of the file we have been using `L1Pegged` even before this PR. I changed it to `PubdataIndependent` as suggested, lmk if that's what not what you intended.

core/lib/dal/src/sync_dal.rs
itegulov: That is indeed be the next planned step after PLA-1011. It will simplify some code, also this is somewhat of a weird pattern as some node operators are confused when `miniblocks.l1_batch` is set to null when we actually know which batch it is supposed to belong to.

core/lib/dal/src/blocks_dal.rs
```diff
@@ -557,17 +568,78 @@ impl BlocksDal<'_, '_> {
...
+                    NOW(),
+                    NOW(),
+                    FALSE
+                )
+            "#,
+            i64::from(number.0),
+            timestamp as i64,
+            batch_fee_input.l1_gas_price() as i64,
+            batch_fee_input.fair_l2_gas_price() as i64,
+            batch_fee_input.fair_pubdata_price() as i64,
```
perekopskiy, line 620: i64 is too small

core/node/state_keeper/src/io/mempool.rs
```diff
@@ -203,6 +203,13 @@ impl StateKeeperIO for MempoolIO {
                 continue;
             }
 
+            self.pool
+                .connection()
+                .await?
+                .blocks_dal()
+                .insert_l1_batch(cursor.l1_batch, timestamp, self.filter.fee_input)
```
perekopskiy, line 210: What if batch is inserted and right after that server restarts? I think you should check for batch with `is_sealed = false` at the beginning of this function. Perhaps you should also save `protocol_version`, `operator_address`

core/lib/dal/src/blocks_dal.rs
itegulov: Don't we cap pubdata price at 10^15, so it should always fit in i64? I don't like that this invariant is very implicit but I don't see an easy way to avoid lossy conversions unless we want to invest some time into "custom" capped primitives

perekopskiy: yes, it's at the moment, but in the very next upgrade that is going to happen very soon this limit is increased to `2^64 - 1`. Can we use `numeric` type?

itegulov: Ah, didn't know that. We can try `numeric` type, sure

core/node/state_keeper/src/io/mempool.rs
itegulov: Alright I just pushed dirty code that should do the trick. Still need to add a test for this scenario but I need to temporarily prioritize devx work so will get back to this PR next week.

As a side effect I had to bring `fee_address` back to the `l1_batches` table thus reverting some of your changes in https://github.com/matter-labs/zksync-era/pull/1617. Hope that's okay.

perekopskiy: looks good! waiting for test

itegulov: I have just added a test, PTAL

perekopskiy (dismissed):

perekopskiy (dismissed):

PR #2849, opened 2024-09-11
feat(zk_inception): Add ecosystem ports scanner

sanekmelnikov:
## What ❔
Added `EcosystemPortsScanner` tool to scan ecosystem directory config and find all values for ports.
Returns `EcosystemPorts` object with ports found and custom description of a service.
`EcosystemPorts` also contains methods to: check whether the port is assigned, allocate a new port in range, etc.

## Why ❔
With multiple chains and number of services (like explorer, etc.) growing - it is useful to have a tool (`EcosystemPorts`) that can be used to allocate new ports, thus, lowering the chance of port conflicts within created configuration.

## Checklist


sanekmelnikov:
Marking as draft as ports in docker files are not parsed and tool is not yet used for allocation.

zk_toolbox/crates/zk_inception/src/utils/ports.rs
```diff
@@ -0,0 +1,364 @@
...
+impl EcosystemPorts {
+    /// Creates a new EcosystemPorts instance with default port configurations.
+    /// These ports are not configured in ecosystem yaml file configs, but are
+    /// commonly used or pre-assigned within the ecosystem.
+    pub fn with_default_ports() -> Self {
+        let mut ports = HashMap::new();
+        ports.insert(OBSERVABILITY_PORT, vec!["Observability".to_string()]);
+        ports.insert(POSTGRES_DB_PORT, vec!["PostgreSQL".to_string()]);
+        ports.insert(LOCAL_HTTP_RPC_PORT, vec!["Local HTTP RPC".to_string()]);
+        ports.insert(LOCAL_WS_RPC_PORT, vec!["Local WebSockets RPC".to_string()]);
```
Deniallugo, line 31: do we need it if we scan docker compose?

matias-gonz: No, I don't think we need it anymore

zk_toolbox/crates/zk_inception/src/utils/ports.rs
```diff
@@ -0,0 +1,350 @@
...
+    }
+
+    /// Finds the smallest multiple of the offset that, when added to the base ports,
+    /// results in a set of available ports within the specified range.
+    ///
+    /// The function iteratively checks port sets by adding multiples of the offset to the base ports:
+    /// base, base + offset, base + 2*offset, etc., until it finds a set where all ports are:
+    /// 1) Within the specified range
+    /// 2) Not already assigned
+    pub fn find_offset(
```
matias-gonz, line 75: We can remove this function since it is not used

zk_toolbox/crates/zk_inception/src/utils/ports.rs
sanekmelnikov: We may or may not have era-observability folder, so I'd leave OBSERVABILITY_PORT (3000).

zk_toolbox/crates/zk_inception/src/utils/ports.rs
```diff
@@ -0,0 +1,350 @@
...
+                            for port_entry in ports {
+                                if let Some(port_str) = port_entry.as_str() {
+                                    let parts: Vec<&str> = port_str.split(':').collect();
+
+                                    // Handle 3-part (IP:host:container) or 2-part (host:container) port mappings
+                                    let host_port_str = match parts.len() {
+                                        3 => parts.get(1),
+                                        2 => parts.first(),
+                                        _ => None,
+                                    };
```
sanekmelnikov, line 211: Just going to highlight cases where this wouldn't work and we can discuss whether it's too crazy to support:
- '[::1]:3010:3010' (IPV6)
- '${HOST_PORT:-3010}:3010' (env configurable, although nothing we can do about it)
- 3010-3014:3010-3014 (port ranges)

I suggest just doing `if parts.len() > 1` then take second to the last part.

zk_toolbox/crates/zk_inception/src/utils/ports.rs
```diff
@@ -0,0 +1,350 @@
...
+        match value {
+            Value::Mapping(map) => {
+                for (key, val) in map {
+                    let new_path = if path.is_empty() {
+                        key.as_str().unwrap_or_default().to_string()
+                    } else {
+                        format!("{}:{}", path, key.as_str().unwrap_or_default())
+                    };
+
+                    if key.as_str() == Some("ports") {
```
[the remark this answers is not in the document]
sanekmelnikov: I guess, whatever, better to keep hacky logic in one place.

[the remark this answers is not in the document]
matias-gonz: Sounds good, but we should accept docker-compose files with `yaml` [extension too](https://docs.docker.com/compose/compose-application-model/#the-compose-file) so there would have to be another way of telling them apart.

zk_toolbox/crates/zk_inception/src/commands/chain/init.rs
```diff
@@ -54,10 +57,12 @@ pub async fn init(
     ecosystem_config: &EcosystemConfig,
     chain_config: &ChainConfig,
 ) -> anyhow::Result<()> {
+    let mut ecosystem_ports = EcosystemPortsScanner::scan(shell)?;
     copy_configs(shell, &ecosystem_config.link_to_code, &chain_config.configs)?;
 
     let mut general_config = chain_config.get_general_config()?;
-    apply_port_offset(init_args.port_offset, &mut general_config)?;
+    ecosystem_ports.allocate_ports(&mut general_config)?;
```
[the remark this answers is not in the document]
matias-gonz: I like this idea. We should implement `ConfigWithChainPorts` for `GeneralConfig` and `ExplorerBackendPorts`.

What do you think @Deniallugo ? Do we want to use the internal chain id as offset?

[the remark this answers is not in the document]
Deniallugo: yeah, that sounds reasonable.

zk_toolbox/crates/zk_inception/src/utils/ports.rs
```diff
@@ -0,0 +1,288 @@
...
+                .allocate_port(PORT_RANGE, "Web3 JSON RPC WS".to_string())?,
+            healthcheck_port: self.allocate_port(PORT_RANGE, "Healthcheck".to_string())?,
+            merkle_tree_port: self.allocate_port(PORT_RANGE, "Merkle Tree".to_string())?,
+            prometheus_listener_port: self.allocate_port(PORT_RANGE, "Prometheus".to_string())?,
+            contract_verifier_port: self
+                .allocate_port(PORT_RANGE, "Contract Verifier".to_string())?,
+            consensus_port: self.allocate_port(PORT_RANGE, "Consensus".to_string())?,
+        };
+
+        config::update_ports(general_config, &ports)?;
```
Deniallugo, line 62: We do scan now, why do we need to specify all of them?

zk_toolbox/crates/zk_inception/src/utils/ports.rs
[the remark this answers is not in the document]
Deniallugo: can we detect docker-compose by having keys `services` and `version`?

zk_toolbox/crates/zk_inception/src/utils/ports.rs
```diff
@@ -0,0 +1,288 @@
...
+                                                    file_path.display(),
+                                                    new_path
+                                                );
+                                                ecosystem_ports.add_port_info(port, description);
+                                            }
+                                        }
+                                    }
+                                }
+                            }
+                        }
```
Deniallugo, line 193: i'd really appreciate if you split this logic to smaller functions and add tests for them

zk_toolbox/crates/zk_inception/src/commands/chain/args/init.rs
```diff
@@ -100,5 +67,4 @@ pub struct InitArgsFinal {
     pub genesis_args: GenesisArgsFinal,
     pub deploy_paymaster: bool,
     pub l1_rpc_url: String,
-    pub port_offset: u16,
```
Deniallugo, line 103: can we add --no-ports-reallocation instead?

zk_toolbox/crates/zk_inception/src/utils/ports.rs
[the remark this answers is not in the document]
sanekmelnikov: `version` is optional though, if we go this route - then isn't`DockerComposeConfig::read` with iteration over `services` better?

zk_toolbox/crates/zk_inception/src/commands/chain/args/init.rs
sanekmelnikov: what's the use-case?

Deniallugo: for devops team. 
They prefer to have the same ports for all envs, since they are all launching in independent envs

zk_toolbox/crates/zk_inception/src/utils/ports.rs
```diff
@@ -0,0 +1,495 @@
...
+        }
+    }
+}
+
+pub struct EcosystemPortsScanner {}
+
+impl EcosystemPortsScanner {
+    /// Scans the ecosystem directory for YAML files and extracts port information.
+    /// Specifically, it looks for keys ending with "port" and collects their values.
+    /// Note: Port information from Docker Compose files will not be picked up by this method.
```
sanekmelnikov, line 105: now, they will be picked up :)

zk_toolbox/crates/zk_inception/src/utils/ports.rs
```diff
@@ -0,0 +1,495 @@
...
+        // - Ecosystem configs directory
+        // - Chain configs directories
+        // - ZKsync repo directory (docker-compose files)
+        let mut dirs = vec![ecosystem_config.config.clone()];
+        for chain in ecosystem_config.list_of_chains() {
+            if let Some(chain_config) = ecosystem_config.load_chain(Some(chain)) {
+                dirs.push(chain_config.configs.clone())
+            }
+        }
+        dirs.push(ecosystem_config.link_to_code);
```
sanekmelnikov, line 119: nit: actually, should it be `shell.current_dir()` instead?
if ecosystem is created outside of the zksync-era repo, then the actual docker compose file that is running will be inside ecosystem repo, right?

matias-gonz: nice catch!!

zk_toolbox/crates/zk_inception/src/commands/ecosystem/args/init.rs
```diff
@@ -92,6 +93,8 @@ pub struct EcosystemInitArgs {
     pub dev: bool,
     #[clap(long, short = 'o', help = MSG_OBSERVABILITY_HELP, default_missing_value = "true", num_args = 0..=1)]
     pub observability: Option<bool>,
+    #[clap(long, help = MSG_NO_PORT_REALLOCATION_HELP, default_missing_value = "true", num_args = 0..=1)]
```
sanekmelnikov, line 96: default_missing_value = "false", no? I think we just adding that for devops team.

zk_toolbox/crates/config/src/consts.rs
```diff
@@ -64,6 +64,18 @@ pub const DEFAULT_EXPLORER_API_PORT: u16 = 3002;
...
+/// Default port for the web3 ws rpc service
+pub const DEFAULT_WEB3_WS_RPC_PORT: u16 = 3051;
+/// Default port for the healthcheck service
+pub const DEFAULT_HEALTHCHECK_PORT: u16 = 3071;
+/// Default port for the merkle tree
+pub const DEFAULT_MERKLE_TREE_PORT: u16 = 3072;
+/// Default port for the prometheus service
+pub const DEFAULT_PROMETHEUS_PORT: u16 = 3314;
+/// Default port for the contract verifier service
+pub const DEFAULT_CONTRACT_VERIFIER_PORT: u16 = 3070;
```
Deniallugo, line 78: do we still need them?

zk_toolbox/crates/zk_inception/src/commands/ecosystem/args/init.rs
matias-gonz: `default_missing_value` means that when you provide just the argument the value is true. Making the next two commands the same:
```bash
zki e init --no-port-reallocation
zki e init --no-port-reallocation=true
```

matias-gonz: When the option is not provided:
https://github.com/matter-labs/zksync-era/pull/2849/files#diff-7187f86e54c67642efffe168253f12dcdd2dc0da07e3b4cd08b0814348d700b5R57

matias-gonz: This should make it more clear:
https://github.com/matter-labs/zksync-era/pull/2849/commits/2eb62c52438dbf15df12ff48706d439a0f948127

zk_toolbox/crates/zk_inception/src/utils/ports.rs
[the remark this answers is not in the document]
Deniallugo: probably services is fine

zk_toolbox/crates/zk_inception/src/utils/ports.rs
```diff
@@ -0,0 +1,494 @@
...
+            (
+                "prometheus_listener_port".to_string(),
+                DEFAULT_PROMETHEUS_PORT,
+            ),
+            (
+                "contract_verifier_port".to_string(),
+                DEFAULT_CONTRACT_VERIFIER_PORT,
+            ),
+            ("consensus_port".to_string(), DEFAULT_CONSENSUS_PORT),
+        ])
```
Deniallugo, line 314: why do we need this function? We have already allocated ports in config file

zk_toolbox/crates/zk_inception/src/utils/ports.rs
```diff
@@ -0,0 +1,494 @@
...
+    }
+
+    fn set_ports(&mut self, ports: HashMap<String, u16>) -> anyhow::Result<()> {
```
Deniallugo, line 317: Do I understand it right, that if we have some another port, which is not specified, it'd fail?

Deniallugo: In this case, i'd rather save the whole paths to the value and write a small macros for updating it

zk_toolbox/crates/config/src/consts.rs
matias-gonz: These are used for default port + offset allocation

zk_toolbox/crates/zk_inception/src/utils/ports.rs
matias-gonz: I will see if I can use those instead

zk_toolbox/crates/zk_inception/src/utils/ports.rs
matias-gonz: `set_ports` is used [here](https://github.com/matter-labs/zksync-era/pull/2849/files#diff-7847b0da60e1ab89e4bb36e78a39f1e249133f2c82b78d86aecf1bbfd0b19d1eR60)

This means `set_ports` uses the same ports as the ones in `get_default_ports`
So it wouldn't fail as long as `get_default_ports` is consistent with `set_ports`. Both functions belong to the same trait: `ConfigWithChainPorts` making it easy to keep it consistent.

How do you think we can improve this?

zk_toolbox/crates/zk_inception/src/commands/chain/init.rs
```diff
@@ -54,15 +57,29 @@ pub async fn init(
...
+            &chain_config.path_to_general_config(),
+            chain_config.id,
+        )?;
+    }
     let mut general_config = chain_config.get_general_config()?;
-    apply_port_offset(init_args.port_offset, &mut general_config)?;
-    let ports = ports_config(&general_config).context(MSG_PORTS_CONFIG_ERR)?;
+    let offset = ((chain_config.id - 1) * 100) as u16;
+    let consensus_port_range = DEFAULT_CONSENSUS_PORT + offset..PORT_RANGE_END;
+    let consensus_port =
```
Deniallugo, line 73: can't we reuse the ecosystem ports for it?

matias-gonz: You mean adding a function to `EcosystemPorts` to allocate consensus port?

matias-gonz: Another approach is to add `ConsensusConfig` to file_based/general.yaml
including the following fields:
* server_addr
* public_addr
* max_payload_size
* max_batch_size
* gossip_dynamic_inbound_limit

Then we need a special case for `server_addr` and `public_addr` in the function `traverse_allocate_ports_in_yaml`

Deniallugo (dismissed):

Deniallugo (approved):

artmakh (approved):

PR #2852, opened 2024-09-11
feat: make `to` address optional for transaction data

IAvecilla:
## What ❔

This PR is the first part of [this one](https://github.com/matter-labs/zksync-era/pull/2631). 

We decided to separate it into smaller PRs to make it easier to read and review.

This PR contains only the changes to the `TransactionData` structs for the different VM versions, modifying them for an optional field.

The other parts are currently available in our repo fork, as they are based on this branch. The other PRs are:
- [feat: evm simulator as base system contract](https://github.com/lambdaclass/zksync-era/pull/256): integrates the EVM interpreter contract as a base system contract and includes all necessary updates for zksync-era to run with it.
- [feat: make evm simulator optional for the server](https://github.com/lambdaclass/zksync-era/pull/257): makes the evm simulator optional to run using a config file.

## Checklist


core/lib/dal/src/transactions_dal.rs
```diff
@@ -58,7 +58,13 @@ impl TransactionsDal<'_, '_> {
         tx: &L1Tx,
         l1_block_number: L1BlockNumber,
     ) -> DalResult<()> {
-        let contract_address = tx.execute.contract_address.as_bytes();
+        let contract_address = tx.execute.contract_address;
+        let unwrapped_contract_address = contract_address.unwrap_or_default();
+        let contract_address_b: &[u8] = if contract_address.is_none() {
```
0xVolosnikov, line 63:
```suggestion
        let contract_address_as_bytes: &[u8] = if contract_address.is_none() {
```

core/lib/dal/src/transactions_dal.rs
```diff
@@ -161,7 +167,13 @@ impl TransactionsDal<'_, '_> {
     }
 
     pub async fn insert_system_transaction(&mut self, tx: &ProtocolUpgradeTx) -> DalResult<()> {
-        let contract_address = tx.execute.contract_address.as_bytes().to_vec();
+        let contract_address = tx.execute.contract_address;
+        let unwrapped_contract_address = contract_address.unwrap_or_default();
+        let contract_address_b: &[u8] = if contract_address.is_none() {
```
0xVolosnikov, line 172:
```suggestion
        let contract_address_as_bytes: &[u8] = if contract_address.is_none() {
```

core/lib/dal/src/transactions_dal.rs
```diff
@@ -818,7 +843,7 @@ impl TransactionsDal<'_, '_> {
             &l2_inputs as &[&[u8]],
             &l2_datas,
             &l2_tx_formats,
-            &l2_contract_addresses as &[&[u8]],
```
0xVolosnikov, line 821: Nit: do we need this change?

core/lib/types/src/l2/mod.rs
```diff
@@ -233,7 +233,7 @@ impl L2Tx {
 
     /// Returns recipient account of the transaction.
     pub fn recipient_account(&self) -> Address {
```
0xVolosnikov, line 235: Shouldn't it be also `-> Option<Address>`?

prover/crates/lib/keystore/src/utils.rs
```diff
@@ -115,6 +115,7 @@ pub fn calculate_snark_vk_hash(keystore: &Keystore) -> anyhow::Result<H256> {
 #[cfg(test)]
 mod tests {
     use std::str::FromStr;
+
```
0xVolosnikov, line 118: Nit: unnecessary diff

core/lib/multivm/src/versions/vm_1_3_2/transaction_data.rs
```diff
@@ -170,7 +170,7 @@ impl TransactionData {
         encode(&[Token::Tuple(vec![
             Token::Uint(U256::from_big_endian(&self.tx_type.to_be_bytes())),
             Token::Address(self.from),
-            Token::Address(self.to),
+            Token::Address(self.to.unwrap_or_default()),
```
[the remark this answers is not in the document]
IAvecilla: You mean to apply this change for older VMs? At least for `vm_latest`, we can have a `None` at this point, so we need to take the default `U256` to avoid panicking.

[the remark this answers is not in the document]
0xVolosnikov: > This method is to be used only in tests, when we want to bypass the checks imposed

I don't think the `None` scenario is implied in tests. On the other hand, I don't think it's a problem.

prover/crates/lib/keystore/src/utils.rs
IAvecilla: I think that `zk fmt` is adding it.

core/lib/dal/src/transactions_dal.rs
IAvecilla: I also updated the implementation for obtaining the `contract_address_bytes` to reduce the boilerplate.

core/lib/types/src/l2/mod.rs
IAvecilla: Indeed, updated!

core/lib/multivm/src/versions/vm_1_3_2/transaction_data.rs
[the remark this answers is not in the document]
IAvecilla: Correct me if I'm wrong, but this function `abi_encode_with_custom_factory_deps` is used every time we send a new transaction since it's called by the `BootloaderTx::new()` function, which happens every time we push a new transaction into the bootloader memory.

Changing this line to an `unwrap()` in `vm_latest` causes a panic every time I try to deploy an EVM contract. 
Maybe I'm missing something here, so please let me know.

0xVolosnikov:
Lgtm, waiting for code owners to review

core/lib/types/src/tx/execute.rs
```diff
@@ -72,7 +72,10 @@ impl EIP712TypedStructure for Execute {
     const TYPE_NAME: &'static str = "Transaction";
 
     fn build_structure<BUILDER: StructBuilder>(&self, builder: &mut BUILDER) {
-        builder.add_member("to", &U256::from(self.contract_address.as_bytes()));
+        builder.add_member(
+            "to",
+            &U256::from(self.contract_address.unwrap_or_default().as_bytes()),
+        );
```
perekopskiy, line 78: is it correct way to do it? AFAIU it means that signatures for txs with `tx.contract_address = null` and `tx. contract_address = Address::zero()` will be the same which doesn't look correct

core/lib/types/src/transaction_request.rs
```diff
@@ -818,9 +818,7 @@ impl L2Tx {
         validate_factory_deps(&meta.factory_deps)?;
 
         let mut tx = L2Tx::new(
-            value
-                .to
-                .ok_or(SerializationTransactionError::ToAddressIsNull)?,
+            value.to,
```
perekopskiy, line 821: so this means we start allowing txs with to=null in our system but process them as if `to=0x00..00`. I would continue rejecting such txs until evm equivalence is on, otherwise it will be confusing for everyone

core/lib/dal/src/consensus/mod.rs
```diff
@@ -401,9 +401,11 @@ impl ProtoRepr for proto::Transaction {
                 }
             },
             execute: Execute {
-                contract_address: required(&execute.contract_address)
-                    .and_then(|x| parse_h160(x))
-                    .context("execute.contract_address")?,
+                contract_address: Some(
+                    required(&execute.contract_address)
```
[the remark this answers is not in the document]
IAvecilla: Right, I changed it to get the actual `Option` instead of using `required`, is that the correct way to do it?

core/lib/dal/src/transactions_dal.rs
```diff
@@ -58,7 +58,10 @@ impl TransactionsDal<'_, '_> {
         tx: &L1Tx,
         l1_block_number: L1BlockNumber,
     ) -> DalResult<()> {
-        let contract_address = tx.execute.contract_address.as_bytes();
+        let contract_address = tx.execute.contract_address;
+        let contract_address_as_bytes = contract_address
+            .map(|addr| addr.as_bytes().to_vec())
+            .unwrap_or_default();
```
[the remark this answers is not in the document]
IAvecilla: Done! I had to add some type conversions in a few queries since that seems to be the correct way to handle Null inserts with `sqlx`

core/lib/types/src/tx/execute.rs
IAvecilla: I agree that it might be incorrect for the signature to remain the same when the address is null or `0x00`. I've changed it to only include the `to` field as a member when the contract address is `Some(..)`, and not include it when it's `None`. 

I'm not entirely sure if this is the best approach or if it could introduce additional issues. While reviewing other parts of the code, I noticed that when a member has an `Option` type, it usually defaults when it is `None`, like [here](https://github.com/matter-labs/zksync-era/blob/3f1d1a26076d5bf2fd65c31876d2a32a3cbfff0c/core/lib/types/src/transaction_request.rs#L339), but this might be a case where it's not possible to get `None` as a value. Let me know it there's a better way to do this

core/lib/types/src/transaction_request.rs
IAvecilla: Done! I added the check above and included a comment for its removal once the EVM interpreter is integrated.

core/lib/types/src/l1/mod.rs
```diff
@@ -274,7 +274,7 @@ impl From<L1Tx> for abi::NewPriorityRequest {
             transaction: abi::L2CanonicalTransaction {
                 tx_type: PRIORITY_OPERATION_L2_TX_TYPE.into(),
                 from: address_to_u256(&t.common_data.sender),
-                to: address_to_u256(&t.execute.contract_address),
+                to: address_to_u256(&t.execute.contract_address.unwrap_or_default()),
```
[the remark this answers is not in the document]
jrchatruc: L1 transactions come from `L2TransactionRequest`s, which have to contain an address, so I don't think `contract_address` is nullable.

I imagine making it non-nullable through the type system would involve making a new `Execute` struct specific for L1 transactions where the `contract_address` field is not an `Option` and make L1Tx contain it instead of the regular `Execute`. Is that reasonable or do you think it's better to unwrap/use try_from to return an error instead?

core/lib/dal/src/consensus/mod.rs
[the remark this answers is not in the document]
perekopskiy: Yes! Please also change comment in .proto file (`optional bytes contract_address = 1; // required; H160` -> `optional bytes contract_address = 1; // optional; H160`

core/lib/types/src/l1/mod.rs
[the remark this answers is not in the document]
perekopskiy: > L1 transactions come from L2TransactionRequests, which have to contain an address, so I don't think contract_address is nullable.

Yes, that's the current state, I was asking more about future plans. I didn't expect a definite answer but I wanted to raise this topic

> Is that reasonable or do you think it's better to unwrap/use try_from to return an error instead?

I think it's reasonable long-term but not as a part of this PR. So let's unwrap/return error for now but keep it in mind, cc @0xVolosnikov

perekopskiy (approved):

PR #2863, opened 2024-09-12
feat: vm2 account validation

joonazan:
Implements an account validation gas limit and the validation tracer for vm2, along with better tests for validation.

Instead of a second gas limit like in vm_latest, the normal gas limit is used. Unfortunately this means that the VM is not safe to use in the sequencer until we forbid the use of gasleft. I didn't do it here because it requires something like taint analysis and could break existing contracts that didn't know that gasleft is forbidden.


core/lib/multivm/src/versions/vm_fast/vm.rs
```diff
@@ -86,6 +86,18 @@ impl VmRunResult {
     }
 }
 
+pub trait TracerExt: Tracer {
```
joonazan, line 89: Having to add this is not great. It could instead be added to the tracer interface, which is ok-ish. I don't want to duplicate the hook logic in the tracer. Do you have any other ideas?

core/lib/vm_executor/src/oneshot/mod.rs
```diff
@@ -188,39 +189,64 @@ where
...
+                        vm_fast::validation_tracer::ValidationTracer::new(validation_params);
+                    let result_and_logs = vm.inspect(&mut tracer, InspectExecutionMode::OneTx);
+                    if let Some(violation) = tracer.validation_error() {
+                        return Err(ValidationError::ViolatedRule(violation));
+                    }
+                    match result_and_logs.result {
+                        ExecutionResult::Halt { reason } => Err(ValidationError::FailedTx(reason)),
+                        ExecutionResult::Revert { .. } => {
+                            unreachable!("Revert can only happen at the end of a transaction")
+                        }
```
slowli, line 231: Dumb question: Would it be difficult to process reverts here? Are they *logically* impossible, or is how the bootloader *should* work?

core/lib/vm_executor/src/oneshot/mod.rs
```diff
@@ -188,39 +189,64 @@ where
...
+                        ExecutionResult::Success { .. } => Ok(()),
+                    }
+                }
+
+                _ => {
+                    unimplemented!("Shadow validation is not supported.")
+                }
```
slowli, line 238: Why not? AFAICT, `ValidationError`s can be compared (both `Halt`s and `ViolatedValidationRule`s are comparable), so there should be nothing preventing running both VMs with necessary settings and comparing their results.

core/lib/multivm/src/versions/testonly/account_validation_rules.rs
```diff
@@ -0,0 +1,109 @@
...
+        test_rule::<VM>(2),
+        Some(ViolatedValidationRule::CalledContractWithNoCode(_))
+    );
+
+    // Disallowing gasleft is difficult because it isn't immediately followed
+    // by far call in EraVM bytecode.
+    /*assert_matches!(
+        test_rule::<VM>(100),
+        Some(ViolatedValidationRule::TouchedDisallowedContext)
+    );*/
```
slowli, line 33: I don't understand how this comment and the commented code relate to each other. AFAIU, `TouchedDisallowedContext` is raised by the new VM on `ContextMeta` opcode, which differs from `ErgsLeft`. Is it impossible to trigger `ContextMeta` opcode execution from Solidity? (Even if so, the comment is still misleading, and commented code should be either restored or removed.)

core/lib/multivm/src/versions/testonly/account_validation_rules.rs
```diff
@@ -0,0 +1,109 @@
...
+            ContractToDeploy::account(bytecode, aa_address).funded()
+        ])
+        .with_execution_mode(TxExecutionMode::VerifyExecute)
+        .with_rich_accounts(1)
+        .build::<VM>();
+
+    let mut private_account = vm.rich_accounts[0].clone();
+
+    // Set the type of misbehaviour of the AA contract
+    let function = contract.function("setTypeOfRuleBreak").unwrap();
```
slowli, line 56: Nit: IMO, it would be acceptable to set the contract storage directly in this case to simplify the contract, but since this simplification is marginal, feel free to leave as-is.

core/lib/multivm/src/versions/testonly/account_validation_rules.rs
```diff
@@ -0,0 +1,109 @@
...
+    assert!(!vm
+        .vm
+        .execute(InspectExecutionMode::OneTx)
+        .result
+        .is_failed());
+
+    // Use account abstraction
+    let chain_id: u32 = 270;
+
+    let tx_712 = L2Tx::new(
```
slowli, line 76: Nit: Could make sense to move this functionality to the `test_account` crate, so that it can be reused in the `require_eip712` module etc.

etc/contracts-test-data/contracts/custom-account/validation-rule-breaker.sol
```diff
@@ -0,0 +1,138 @@
...
+        if (typeOfRuleBreak == 1) {
+            // The balance of another account may not be read
+            // I'm writing assertions because otherwise the compiler would optimize away the access
+            require(BOOTLOADER_FORMAL_ADDRESS.balance != 0);
+        } else if (typeOfRuleBreak == 2) {
+            // May not call an EOA
+            address(1234567890).call("");
+        } else if (typeOfRuleBreak == 3) {
+            // TODO make test that calls upgradeable proxy
+        } else if (typeOfRuleBreak == 100) {
```
slowli, line 42: - Do you plan to add this in this PR? If not, makes sense to remove this for now and restore when the test is added.
- What about reading allowed / disallowed storage slots?

slowli (commented):
Sorry for a delayed review.

core/lib/multivm/src/versions/testonly/account_validation_rules.rs
joonazan: The comment should be expanded a bit. The issue is that I want to test gasleft but we currently allow gasleft because there is no easy way to forbid it.
Using gasleft would also trigger TouchedDisallowedContext.

core/lib/multivm/src/versions/testonly/account_validation_rules.rs
joonazan: That's an interesting idea. I guess the number is always at slot 0, so it would be easy.

etc/contracts-test-data/contracts/custom-account/validation-rule-breaker.sol
joonazan: Yes, I want to add this today. The upgradeable proxy is one case of allowed storage slots.

core/lib/multivm/src/versions/vm_fast/vm.rs
joonazan: If the Vm knew where in the composite tracer ValidationTracer is, it could simply set its in_validation boolean.

I'm not a fan of the built-in CircuitsTracer, though. But I think I'll figure something out.

core/lib/multivm/src/versions/vm_fast/validation_tracer.rs
```diff
@@ -0,0 +1,215 @@
...
+                let address = state.current_frame().address();
+                let caller = state.current_frame().caller();
+
+                // Can unwrap because the instruction pointer does not point to a panic instruction
+                let pc = state.current_frame().program_counter().unwrap();
+                let word = pc / 4;
+                let part = pc % 4;
+                let instruction =
+                    state.current_frame().read_contract_code(word).0[3 - part as usize];
+                let slot = state.read_register((instruction >> 16) as u8 & 0b1111).0;
```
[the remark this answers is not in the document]
joonazan: It is extremely low-level. I didn't want to design an interface for getting instruction targets because this is the only place it is needed, so I don't know what a good interface would be like.

core/lib/multivm/src/versions/vm_fast/validation_tracer.rs
```diff
@@ -0,0 +1,215 @@
...
+            && (caller == MSG_VALUE_SIMULATOR_ADDRESS
+                || caller == CONTRACT_DEPLOYER_ADDRESS
+                || caller == BOOTLOADER_ADDRESS)
+        // allow getting chain_id
+        || address == SYSTEM_CONTEXT_ADDRESS && slot == U256::zero()
+        // allow reading code hashes of existing contracts
+        || address == ACCOUNT_CODE_STORAGE_ADDRESS && !value.is_zero()
+    }
+
+    fn set_error(&mut self, error: ViolatedValidationRule) {
```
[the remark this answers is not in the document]
joonazan: It is intentional but also an interesting design question. With the performant tracers, we could even implement bootloader hooks via tracers!

core/lib/vm_executor/src/oneshot/mod.rs
joonazan: Previously reverts were handled like successes. I much prefer this over that.
They are not impossible but getting a revert here would be strange enough to warrant investigation.

core/lib/multivm/src/versions/vm_fast/validation_tracer.rs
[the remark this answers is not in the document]
joonazan: So I thought this through and it should theoretically be zero-cost to return ExecutionStatus from Tracer::after_instruction. It should also work well with tail calls, there will just have to be a branch that is compiled away in the usual case where the tracer doesn't want to stop the execution.

One tricky thing is combining multiple stops. I think I will have stops take priority over suspends.

I'll make another integration PR for it so we can see how it performs in practice.

joonazan:
Can't merge yet because it does not point to master branch of vm2.

Also @slowli what methods should I call to report that there's a mismatch in shadow validation?

core/lib/multivm/src/versions/vm_fast/validation_tracer.rs
```diff
@@ -0,0 +1,291 @@
...
+    Tracer,
+};
+use zksync_vm_interface::tracer::{
+    TimestampAsserterParams, ValidationParams, ValidationTraces, ViolatedValidationRule,
+};
+
+use super::utils::read_fat_pointer;
+use crate::tracers::TIMESTAMP_ASSERTER_FUNCTION_SELECTOR;
+
+pub trait ValidationMode: Tracer + Default {
```
slowli, line 20: Bikeshedding: IMO, `ValidatingTracer` would be a better name (or maybe `ValidationTracer` + rename `ValidationTracer` -> `FullValidationTracer`). As it's named right now, I'd not expect `Tracer` to be a supertrait, but rather have it as an associated type.

core/lib/multivm/src/versions/vm_fast/tests/mod.rs
```diff
@@ -78,7 +81,11 @@ impl PartialEq for VmStateDump {
     }
 }
 
-impl TestedVm for Vm<ImmutableStorageView<InMemoryStorage>> {
+impl<E: Tracer + Default, V: ValidationMode> TestedVm
+    for Vm<ImmutableStorageView<InMemoryStorage>, WithBuiltinTracers<E, V>>
```
slowli, line 85: Nit: You can create an alias for this type, similarly to [`TestedLatestVm`](https://github.com/matter-labs/zksync-era/blob/98803effbca2d285cf600dbaeb78818f56b25951/core/lib/multivm/src/versions/vm_latest/tests/mod.rs#L64).

core/lib/vm_executor/src/oneshot/mod.rs
```diff
@@ -189,44 +194,130 @@ where
...
-            match (exec_result.result, validation_result) {
-                (_, Err(violated_rule)) => Err(ValidationError::ViolatedRule(violated_rule)),
-                (ExecutionResult::Halt { reason }, _) => Err(ValidationError::FailedTx(reason)),
-                _ => Ok(validation_traces.lock().unwrap().clone()),
-            }
+            let version = sandbox.env.system.version.into();
+
+            sandbox.execute_in_vm_with_tracer(|vm, transaction| match vm {
+                Vm::Legacy(vm) => {
+                    let validation_tracer = ValidationTracer::<HistoryDisabled>::new(
```
slowli, line 201: - Would extracting common parts for legacy / fast VMs into functions require too much work (so that they are reused for standalone and shadowed VM versions)?
- Dumb question: Why does this place use `HistoryDisabled`, but logic for the standalone legacy VM uses `HistoryEnabled`?

slowli (commented):
Some preliminary comments; will look through the PR once again tomorrow.

core/lib/multivm/src/versions/vm_fast/vm.rs
```diff
@@ -596,18 +644,9 @@ impl<S: ReadStorage, Tr: Tracer + Default> Vm<S, Tr> {
...
-        );
-        let result = self.run(
-            execution_mode,
-            &mut full_tracer,
-            track_refunds,
-            pubdata_builder,
-        );
-        let ((external_tracer, circuits_tracer), _) = full_tracer;
-        *tracer = external_tracer; // place the tracer back
+        tracer.insert_dynamic_bytecodes_handle(self.world.dynamic_bytecodes.clone());
```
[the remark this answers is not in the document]
joonazan: I didn't really like the `'static` in the interface and the fact that you pass in a reference but the contents of it then get moved. I think I'd prefer a solution where the reference is stored over the move.

You are right that this could cause tracers to be incorrectly reused which is something I did not think about.

IIRC this was also done so the validation result can be taken out of the tracer without extra synchronization primitives.

slowli (approved):

PR #2866, opened 2024-09-12
feat(zk_toolbox): Add zk_supervisor config-writer command

matias-gonz:
## What ❔
Add zk_supervisor override configs command

## Why ❔
In some cases we want to override some default configs. Right now it is done manually and we need proper automation for regular use/ci.


zk_toolbox/crates/common/src/yaml.rs
Deniallugo, line 1: it's a confusing name, better to rename it to `config writer` or something simillar

zk_toolbox/crates/common/src/yaml.rs
```diff
@@ -0,0 +1,472 @@
...
+            key3: null
+            "#,
+        )
+        .unwrap();
+
+        let expected: serde_yaml::Value = serde_yaml::from_str(
+            r#"
+            key1: value1
+            key2: value2
+            key3: null
```
Deniallugo, line 457: can we remove this flag instead?

Deniallugo:
As an integration test you can remove this sed 
https://github.com/matter-labs/zksync-era/blob/3609ea6f1b6d767c0f1d64a9303f0331db920931/.github/workflows/ci-core-reusable.yml#L129-L131

etc/env/file_based/overrides/legacy-new.yaml
```diff
@@ -0,0 +1,7 @@
+db:
+  merkle_tree:
+    mode: LIGHTWEIGHT
+experimental_vm:
+  state_keeper_fast_vm_mode: NEW
+mempool:
+  delay_interval: 50
```
perekopskiy, line 7: nit: config itself has nothing to do with legacy, I would change name to loadtest-new.yaml and loadtest-old.yaml

etc/env/file_based/overrides/no_proofs.yaml
```diff
@@ -0,0 +1,3 @@
+eth:
+  sender:
+    proof_sending_mode: ONLY_REAL_PROOFS
```
perekopskiy, line 3: ONLY_REAL_PROOFS and no_proofs are completely different

etc/env/file_based/overrides/legacy-new.yaml
Deniallugo, line 1: let's at least put it under tests/folder

etc/env/file_based/overrides/validium.yaml
```diff
@@ -0,0 +1,5 @@
+eth:
+  sender:
+    pubdata_sending_mode: CUSTOM
+state_keeper_config:
+  pubdata_overhead_part: 0.0
```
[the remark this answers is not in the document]
matias-gonz: Looks like we missed that, it is added now

Deniallugo (approved):

perekopskiy (approved):

Deniallugo (approved):

artmakh (approved):

PR #2867, opened 2024-09-12
feat(eth-watch): redesign to support multiple chains

tomg10:
This PR contains code from sync-layer-stable + I removed old upgrades processor and updated unit-tests to use the new one


core/lib/dal/src/eth_watcher_dal.rs
```diff
@@ -0,0 +1,152 @@
...
+
+#[derive(Debug, Copy, Clone, sqlx::Type)]
+#[sqlx(type_name = "event_type")]
+pub enum EventType {
+    ProtocolUpgrades,
+    PriorityTransactions,
+}
+
+impl ProcessedEventsDal<'_, '_> {
+    pub async fn get_or_set_next_block_to_process(
```
perekopskiy, line 18: nit: comment explaining what method does would be appreciated

core/lib/dal/src/eth_watcher_dal.rs
```diff
@@ -0,0 +1,152 @@
+use zksync_db_connection::{connection::Connection, error::DalResult, instrument::InstrumentExt};
+use zksync_types::SLChainId;
+
+use crate::Core;
+
+pub struct ProcessedEventsDal<'a, 'c> {
```
perekopskiy, line 6: nit: let's make file name and struct name same. IMHO EthWatcherDal is better than ProcessedEventsDal

core/node/eth_watch/src/event_processors/mod.rs
```diff
@@ -49,10 +54,14 @@ pub(super) trait EventProcessor: 'static + fmt::Debug + Send + Sync {
     async fn process_events(
```
perekopskiy, line 54: Please add what method should return in the comment

core/node/eth_watch/src/client.rs
```diff
@@ -21,10 +21,14 @@ pub trait EthClient: 'static + fmt::Debug + Send + Sync {
...
         from: BlockNumber,
         to: BlockNumber,
+        topic1: H256,
+        topic2: Option<H256>,
         retries_left: usize,
     ) -> EnrichedClientResult<Vec<Log>>;
     /// Returns finalized L1 block number.
     async fn finalized_block_number(&self) -> EnrichedClientResult<u64>;
+
+    async fn get_total_priority_txs(&self) -> Result<u64, ContractCallError>;
```
perekopskiy, line 31: Please add comment. Also IMHO the name is confusing (I understand that it comes from contracts but still), maybe smth like `get_last_ready_priority_tx_id` will be better?

core/node/eth_watch/src/event_processors/priority_ops.rs
```diff
@@ -70,33 +71,48 @@ impl EventProcessor for PriorityOpsEventProcessor {
...
 
         let stage_latency = METRICS.poll_eth_node[&PollStage::PersistL1Txs].start();
         APP_METRICS.processed_txs[&TxStage::added_to_mempool()].inc();
         APP_METRICS.processed_l1_txs[&TxStage::added_to_mempool()].inc();
+        let processed_priority_transactions = client.get_total_priority_txs().await?;
+        let mut processed_events_count = 0;
         for new_op in new_ops {
+            if processed_priority_transactions <= new_op.serial_id().0 {
+                break;
+            }
```
perekopskiy, line 92: can we avoid using `mut`, `break` and e.g. do `take_while` instead? I think it will look nicer

core/node/eth_watch/src/lib.rs
```diff
@@ -34,70 +31,54 @@ mod tests;
...
+    l1_client: Box<dyn EthClient>,
+    sl_client: Box<dyn EthClient>,
     poll_interval: Duration,
     event_processors: Vec<Box<dyn EventProcessor>>,
-    last_processed_ethereum_block: u64,
     pool: ConnectionPool<Core>,
 }
 
 impl EthWatch {
+    #[allow(clippy::too_many_arguments)]
```
perekopskiy, line 47: nit: it's not needed

core/node/eth_watch/src/lib.rs
```diff
@@ -172,34 +132,60 @@ impl EthWatch {
...
-                .cloned()
-                .collect();
-            processor
-                .process_events(storage, &*self.client, processor_events)
+            let client = match processor.event_source() {
+                EventsSource::L1 => self.l1_client.as_ref(),
+                EventsSource::SL => self.sl_client.as_ref(),
+            };
+            let chain_id = client.chain_id().await?;
+            let finalized_block = client.finalized_block_number().await?;
```
perekopskiy, line 141: nit: you can query `finalized_block` out of event_processors' loop

core/node/eth_watch/src/event_processors/priority_ops.rs
```diff
@@ -36,10 +36,11 @@ impl EventProcessor for PriorityOpsEventProcessor {
     async fn process_events(
         &mut self,
         storage: &mut Connection<'_, Core>,
-        _client: &dyn EthClient,
+        client: &dyn EthClient,
```
perekopskiy, line 39: let's rename to sl_client so it's clear what client it is

core/node/eth_watch/src/event_processors/priority_ops.rs
perekopskiy: also `processed_priority_transactions` seems to be off here as it doesn't account for `skip_while` executed before.

core/node/eth_watch/src/lib.rs
```diff
@@ -172,34 +132,60 @@ impl EthWatch {
...
+                    None,
+                    RETRY_LIMIT,
+                )
                 .await?;
+            let processed_events_count = processor
+                .process_events(storage, &*self.sl_client, processor_events.clone())
+                .await?;
+
+            let next_block_to_process = if processed_events_count == processor_events.len() {
+                finalized_block
```
perekopskiy, line 167: super nit: we can even do `finalized_block + 1` here

core/node/eth_watch/src/tests.rs
```diff
@@ -116,15 +145,19 @@ impl EthClient for MockEthClient {
...
+            if let Some(ops) = self.inner.read().await.upgrade_timestamp.get(&number) {
                 logs.extend_from_slice(ops);
             }
         }
-        Ok(logs)
+        Ok(logs
+            .into_iter()
+            .filter(|log| {
+                log.topics.contains(&topic1)
+                    && (topic2.is_none() || log.topics.contains(&topic2.unwrap()))
```
perekopskiy, line 156: nit: contains is incorrect here as e.g. forth topic is `log.topics` can be equal to `topic1`

core/node/eth_watch/src/lib.rs
tomg10: In theory yes, but as the client can be different for different processors with current number of processors, it wouldn't have resulted in less queries (it's queried once for L1 and once for SL) and it would complicate code so I vote not to touch it

core/node/eth_watch/src/client.rs
tomg10: I've changed the name to `get_number_of_processed_priority_txs`. Your name would be great, but the problem is that it's actually equal to last_ready_priority_tx - 1 as they are indexed from 0.

core/node/eth_watch/src/event_processors/priority_ops.rs
tomg10: nice catch, I've also refactored this part of the code into more functional as you suggested, it looks much nicer now

perekopskiy (approved):

PR #2868, opened 2024-09-13
feat(api_server): report gas price based on open batch

itegulov:
## What ❔

`api_server` should report gas price based on the currently open L1 batch as opposed to the latest sealed L2 block.

Note: adding `BatchFeeInput` to L1 batch models did not end up being necessary but I think it still makes sense.

## Why ❔

To avoid gas fluctuation issues on chains with low amount of traffic

## Checklist


core/node/fee_model/src/lib.rs
```diff
@@ -130,22 +130,35 @@ impl BatchFeeModelInputProvider for ApiFeeInputProvider {
...
+            let last_l2_block_params = self
+                .connection_pool
+                .connection_tagged("api_fee_input_provider")
+                .await?
+                .blocks_dal()
+                .get_last_sealed_l2_block_header()
+                .await?;
+            Ok(last_l2_block_params
+                .map(|header| inner_input.stricter(header.batch_fee_input))
+                .unwrap_or(inner_input))
```
itegulov, line 160: Not 100% sure what logic we need to follow when we don't have an open batch. Went with the old logic for now, especially given that the L1 batches before #2846 won't have correct gas information in them.

thomas-nguy: In my humble opinion, I suggest the following change

- Get the latest batch number in DB (could be unsealed or sealed)
We can reuse the function `get_sealed_l1_batch_number`  (to be renamed `get_latest_l1_batch_number`)

- Change `get_unsealed_l1_batch_fee_input` implementation to `get_l1_batch_fee_input` which take a batch number input and return the fee params

- Use `get_l1_batch_fee_input` on the latest batch number in this function


With these changes, you don't have to care if batch is sealed or unsealed. In fact what matter is that you always return the fee_params of the `latest L2 block` which is equal to the latest fee_params of the `latest batch stored in DB` after this change
https://github.com/matter-labs/zksync-era/pull/2846/files

itegulov: @thomas-nguy thanks for your insight! Originally I operated under the assumption that this and #2846 would be a part of the same release in which case the approach suggested by you wouldn't work as there is a corner case on the very first start when the latest sealed batch's fee input would not be populated. Now, since #2846 has been rolled out as a part of 24.29.0 release, I think what you laid out above would work and makes a lot of sense.

@perekopskiy could you verify my thinking here? If you agree with the direction I'll complete the PR ASAP

perekopskiy: @itegulov yes, makes sense

core/lib/types/src/fee_model.rs
```diff
@@ -46,6 +46,30 @@ impl BatchFeeInput {
...
+        l1_gas_price: u64,
+        fair_l2_gas_price: u64,
+        fair_pubdata_price: Option<u64>,
+    ) -> Self {
+        protocol_version
+            .filter(|version: &ProtocolVersionId| version.is_post_1_4_1())
+            .map(|_| {
+                Self::PubdataIndependent(PubdataIndependentBatchFeeModelInput {
+                    fair_pubdata_price: fair_pubdata_price
+                        .expect("No fair pubdata price for 1.4.1 miniblock"),
```
perekopskiy, line 61: nit: not alwaays "a miniblock", I'd just leave it as "No fair pubdata price for 1.4.1"

perekopskiy (dismissed):

perekopskiy (dismissed):

perekopskiy (approved):

PR #2872, opened 2024-09-13
feat: deployment of multicall3 on L2 by zk_toolbox

pompon0:
For complex atomic operations on consensus registry.
Should be merged after https://github.com/matter-labs/era-contracts/pull/805
Tested manually by running `cargo -p zk_inception -- e init --dev`, which populated the multicall3 address in contract.yaml to non-zero value.


zk_toolbox/crates/config/src/contracts.rs
```diff
@@ -152,4 +158,5 @@ pub struct L2Contracts {
     pub testnet_paymaster_addr: Address,
     pub default_l2_upgrader: Address,
     pub consensus_registry: Option<Address>,
+    pub multicall3: Address,
```
Deniallugo, line 161: non option field is breaking change

pompon0: done

PR #2878, opened 2024-09-14
fix(zk_toolbox): increase confirmations in testing

manuelmauro:
## What ❔

Increase confirmations in testing.

## Why ❔

Should make CI workflows less flaky.

## Checklist

### DIFF
--- a/infrastructure/zk/src/docker.ts
+++ b/infrastructure/zk/src/docker.ts
@@ -1,4 +1,4 @@
-import {Command} from 'commander';
+import { Command } from 'commander';
 import * as utils from 'utils';
 
 const IMAGES = [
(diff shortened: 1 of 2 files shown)


Deniallugo (commented):
While we are here, lets increase nonce manually and send all mint transactions simultaneously, now we wait 8 seconds instead of 2 for nothing

manuelmauro:
> While we are here, lets increase nonce manually and send all mint transactions simultaneously, now we wait 8 seconds instead of 2 for nothing

Addressed in https://github.com/matter-labs/zksync-era/pull/2878/commits/fee848963f839cdaf955bc34ae20fac7b86a0221

Deniallugo (approved):

PR #2879, opened 2024-09-15
docs: fix aa docs

wetkeyboard:
Hi, when I was trying to learn about the paymaster process, I noticed that the link was broken. I hope the update I made is correct.


wetkeyboard:
@Deniallugo  @perekopskiy  GM, would you mind reviewing this?

PR #2883, opened 2024-09-16
chore(vm): Bump vm2 repo revision

slowli:
## What ❔

Bumps the `vm2` crate revision and updates new VM usage correspondingly.

## Why ❔

To keep the `vm2` dependency up to date and track possible regressions.

## Checklist


slowli:
The performance regression is mostly encapsulated in the `jump` instruction handler and is caused by the change in program counter computation (i.e., removing `ptr::offset_from` – it's slightly more performant than computing the offset manually). IMO, it's safer to accept this regression in this case.

joonazan (approved):

PR #2887, opened 2024-09-16
feat(tee): use hex serialization for RPC responses

pbeza:
## What ❔

Following Anton's suggestion, we have switched to hex serialization for API/RPC requests and responses. Previously, we used default JSON serialization for Vec<u8>, which resulted in a lengthy comma-separated list of integers.

This change standardizes serialization, making it more efficient and reducing the size of the responses. The previous format, with a series of comma-separated integers for pubkey-like fields, looked odd.

Then:
```
curl -X POST\
     --data '{"jsonrpc": "2.0", "id": 1, "method": "unstable_getTeeProofs", "params": [491882, "sgx"] }' \
        https://mainnet.era.zksync.io
{"jsonrpc":"2.0","result":[{"attestation":[3,0,2,0,0,0,0,0,10,<dozens of comma-separated integers here>
```

Now:
```
$ curl -X POST \
       --data '{"jsonrpc": "2.0", "id": 1, "method": "unstable_getTeeProofs", "params": [1, "sgx"] }' \
          http://localhost:3050
{"jsonrpc":"2.0","result":[{"l1BatchNumber":1,"teeType":"sgx","pubkey":"0506070809","signature":"0001020304","proof":"0a0b0c0d0e","provedAt":"2024-09-16T11:53:38.253033Z","attestation":"0403020100"}],"id":1}
```

This change needs to be deployed in lockstep with:
https://github.com/matter-labs/teepot/pull/209.

## Why ❔

To improve user experience (UX) and optimize the RPC response size.

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10898,6 +10898,7 @@ dependencies = [
  "secp256k1",
  "serde",
  "serde_json",
+ "serde_with",
  "strum",
  "thiserror",
  "tokio",
--- a/core/lib/prover_interface/Cargo.toml
+++ b/core/lib/prover_interface/Cargo.toml
@@ -20,7 +20,7 @@ circuit_sequencer_api_1_5_0.workspace = true
 
 serde.workspace = true
 strum = { workspace = true, features = ["derive"] }
-serde_with = { workspace = true, features = ["base64"] }
+serde_with = { workspace = true, features = ["base64", "hex"] }
 chrono = { workspace = true, features = ["serde"] }
 
 [dev-dependencies]
--- a/core/lib/prover_interface/src/api.rs
+++ b/core/lib/prover_interface/src/api.rs
@@ -2,6 +2,7 @@
 //! This module defines the types used in the API.
 
 use serde::{Deserialize, Serialize};
+use serde_with::{hex::Hex, serde_as};
 use zksync_types::{
     protocol_version::{L1VerifierConfig, ProtocolSemanticVersion},
     tee_types::TeeType,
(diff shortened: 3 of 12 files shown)


haraldh:
is that backwards compatible?

pbeza:
> is that backwards compatible?

Definitely not, which means we need to update both the prover and verifier in sync. I see three options here:

1. Since, realistically, no one is relying on the TEE Prover yet, we can use this time to iterate quickly and ignore breaking compatibility changes, deferring API versioning until later.
2. We can introduce versioning for the API/RPC interface.
3. We can quietly support both formats in the TEE proof verifier using a series of `if` statements, without explicit versioning.

Any other idea? Any preferences? It would be interesting to understand how API/RPC interface versioning is handled in other parts of the project. (cc @popzxc)

haraldh:
I don't understand this change... nobody is looking at the raw data anyway. And potential http compression makes the byte saving negligible

RomanBrodetski:
> This change will make serialization more efficient and reduce the size of the responses.

I don't think we care about serialization efficiency in this case, but I agree it doesn't look right to have those integer arrays. Indeed raw data is not often observed, but I imagine us having to document it - would look less awkward with hex serialization. Not saying it's critical, but since we have it i implemented here already let's merge

pbeza:
tl;dr Feel free to re-run the tests later and merge. ✅

---

@haraldh @RomanBrodetski This PR is ready to be merged, but the flaky state of the tests is preventing them from passing (as discussed [here](https://matter-labs-workspace.slack.com/archives/C05ANUCGCKV/p1726654356732959?thread_ts=1726475710.170479&cid=C05ANUCGCKV)).

Regarding backward compatibility: If we deploy PR https://github.com/matter-labs/teepot/pull/209 in lockstep with this one, we should be fine. The only tool processing the fields that are now being serialized differently is the one from https://github.com/matter-labs/teepot/pull/209.

pbeza:
JFYI: PR https://github.com/matter-labs/zksync-era/pull/2912 should hopefully fix [the test flakiness](https://github.com/matter-labs/zksync-era/actions/runs/10922135532/job/30316250020?pr=2887) that is blocking this PR.

pbeza:
@haraldh/@RomanBrodetski It seems all the tests have finally passed. Please feel free to review/merge.

RomanBrodetski (approved):

PR #2894, opened 2024-09-16
docs: Update system requirements

brunoffranca:
## What ❔
- 30 GB of free disk space are no longer sufficient to run a testnet EN. Upped the system requirements to 50 GB.
- State growth is normally used to refer to the growth of the state tree, not of the historical state. What we have is historical state pruning, not state pruning.

### DIFF
--- a/docs/guides/external-node/00_quick_start.md
+++ b/docs/guides/external-node/00_quick_start.md
@@ -65,12 +65,12 @@ The HTTP JSON-RPC API can be accessed on port `3060` and WebSocket API can be ac
 
 > [!NOTE]
 >
-> To stop state growth, you can enable state pruning by uncommenting `EN_PRUNING_ENABLED: true` in docker compose file,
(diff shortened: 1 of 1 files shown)


docs/guides/external-node/00_quick_start.md
```diff
@@ -65,12 +65,12 @@ The HTTP JSON-RPC API can be accessed on port `3060` and WebSocket API can be ac
 
 > [!NOTE]
 >
-> To stop state growth, you can enable state pruning by uncommenting `EN_PRUNING_ENABLED: true` in docker compose file,
+> To stop historical DB growth, you can enable DB pruning by uncommenting `EN_PRUNING_ENABLED: true` in docker compose file,
```
popzxc, line 68: @slowli we have merkle tree pruning too, don't we? Or wasn't this project completed?

pompon0 (approved):

docs/guides/external-node/00_quick_start.md
brunoffranca: We do. But I think we are only pruning historical tree states.

PR #2899, opened 2024-09-17
feat(prover): WitnessGenerator refactoring #2

Artemka374:
## What ❔

Introduce WitnessGenerator trait
Rename some methods

## Checklist


Artemka374:
Actually I beliebe we really can do it the way that Leaf/node/rt/scheduler have the minimum amount of different code and I think much more changes can be done on this field, but I want to do these changes step by step(to not break anything). But yes, I think we can do something like that in the next PRs

PR #2908, opened 2024-09-18
feat(circuit_prover): Add circuit prover

EmilLuta:
### Summary

This PR introduces a new component `circuit_prover`, which is multiple WVGs & GPU prover running together, groupless. The changes are non-destructive, old setup and new setup must work together in-tandem.

### What?

Circuit prover is a component that runs X WVGs alongside a GPU prover. Making full use of CPUs on the GPU machine, WVGs as a component can be removed altogether. Prover groups are not needed anymore. Based on empiric testing we can (almost -- there will be follow-up fixes to make it efficient) fully run everything on a single machine.

The current implementation can sunset the old setup. Current metrics show that circuit prover is > 60% efficient than old one (but quirks are needed for node proofs to unlock it -- will be treated as a follow-up).

The purpose is to have the `circuit_prover` deprecate old `prover_fri` & `witness_vector_generator`.

### Why?

The changes will allow us to reduce our infrastructure footprint by ~2x and fix plenty of issues we had in the past. Namely:
- fully decoupled of GCP
- better resource utilization & reduce costs
- reduce overall infrastructure needs (which solves the GPU unavailability we've been facing)
- reduce complexity & other inefficiencies (no more prover groups!)
- and more

### Ask

We want to unblock folks running on AWS. This PR is done as is to speed up release process on DevOps side, as it's the longest pole.

NOTE: This is the first PR out of a longer set of PRs.

Comments are more than welcome, but the following concerns will be addressed in follow-up PRs and are out of scope for this PR:
- tracing implementation is subpar; in fact, I'm confident that most metrics could be done via traces
- there's a lot of code duplication (both old/new prover, but runner interface between new WVG & CP)
- tests
- concern separation between job scheduling & job execution
- job priority based on resource consumption
- other nits (such as no README, constants being hard-coded instead of configuration, etc.)

### Reviewer suggestion

This is basically a merge between `prover_fri`, `witness_vector_generation` and `JobProcessor`. Checking code alongside should give you a better view of what's going on. Sorry for making this hard. :/

### DIFF
--- a/docs/guides/setup-dev.md
+++ b/docs/guides/setup-dev.md
@@ -243,17 +243,20 @@ commands related to deployment, you can pass flags for Foundry integration.
 
 ## Non-GPU setup
 
-Circuit Prover requires a GPU (& CUDA bindings) to run. If you still want to be able to build everything locally on non-GPU setup, you'll need to change your rustflags.
+Circuit Prover requires a GPU (& CUDA bindings) to run. If you still want to be able to build everything locally on
+non-GPU setup, you'll need to change your rustflags.
 
 For a single run, it's enough to export it on the shell:
+
 ```
 export RUSTFLAGS='--cfg=no_cuda'
 ```
 
-For persistent runs, you can either echo it in your ~/.<shell>rc file (discouraged), or configure it for your taste in `config.toml`.
+For persistent runs, you can either echo it in your ~/.<shell>rc file (discouraged), or configure it for your taste in
+`config.toml`.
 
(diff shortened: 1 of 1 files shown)


Artemka374 (commented):
Generally LGTM, happy to see those changes!

popzxc (commented):
The review is super rough, just sending quick thoughts. It's being discussed with Emil in DMs too.

popzxc (approved):
Approving with expectation that the comments will be addressed in the follow up PRs.

yorik (approved):

PR #2910, opened 2024-09-18
feat(toolbox): Port toolbox to work with gateway

StanislavBreadless:
## Checklist


zk_toolbox/crates/zk_inception/src/commands/chain/migrate_to_gateway.rs
```diff
@@ -0,0 +1,465 @@
...
+    if receipt.status == Some(U64::from(1)) {
+        println!("Transaction completed successfully!");
+    } else {
+        panic!("Transaction failed!");
+    }
+
+    Ok(())
+}
+
+async fn call_script(
```
kelemeno, line 406: This function is duplicated, probably ok for now

.github/workflows/ci-core-reusable.yml
```diff
@@ -318,45 +367,6 @@ jobs:
       #     ci_run sleep 10
       #     ci_run zk test i upgrade
 
-      - name: Show server.log logs
```
kelemeno, line 321: Do we not want to show logs? Probably something else is displaying them

core/lib/dal/src/eth_sender_dal.rs
```diff
@@ -746,6 +750,43 @@ impl EthSenderDal<'_, '_> {
 
         Ok(())
     }
+
+    pub async fn get_batch_finalization_info(
```
kelemeno, line 754: I don't understand what happens here?

.github/workflows/ci-core-reusable.yml
StanislavBreadless: yeah, the Upload logs action publishes all the logs to github and can be accessible via a link (very helpful when we'll run tests for multiple chains at the same time)

zk_toolbox/crates/zk_inception/src/commands/chain/migrate_to_gateway.rs
StanislavBreadless: let's keep it this way for now, I have a feeling that Danil would want to refactor lots of stuff here anywhere

kelemeno (approved):

zk_toolbox/crates/common/src/forge.rs
```diff
@@ -69,6 +69,17 @@ impl ForgeScript {
                 return Ok(res?);
             }
         }
+
+        // TODO: This line is very helpful for debugging purposes,
+        // maybe it makes sense to make it conditionally displayed.
+        let command = format!(
```
kelemeno, line 75: not a big issue, but small code duplication here as well

PR #2917, opened 2024-09-18
fix(api): Return correct flat call tracer

Deniallugo:
## What ❔

It was an inconsistency between call tracer results

## Checklist


core/lib/types/src/api/mod.rs
```diff
@@ -728,6 +733,22 @@ pub enum BlockStatus {
     Verified,
 }
 
+/// Result tracers need to have a nested result field for compatibility. So we have two different
+/// structs 1 for blocks tracing and one for txs and call tracing
```
joonazan, line 737: This seems overly complicated. Why can't we always have `Vec<DebugCall>`?

Deniallugo: For compatibility reasons... i don't want to break our clients

core/lib/types/src/api/mod.rs
```diff
@@ -728,6 +733,22 @@ pub enum BlockStatus {
...
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(untagged)]
+pub enum CallTracerResultWithNestedResult {
```
slowli, line 740: - Bikeshedding: This name looks overly long.
- Since this type is only used as `Vec<CallTracerResultWithNestedResult>` (in the output of `debug_traceBlock`), maybe it makes sense to put `Vec<_>` *inside* the enum? AFAIU, it would represent domain better (it's impossible that an output mixes variants from both flat and ordinary tracers). I believe it would make conversions easier as well. Also, it would solve the naming problem; you could call the enum something like `CallTracerBlockResult`.

core/lib/types/src/debug_flat_call.rs
```diff
@@ -34,222 +32,3 @@ pub struct CallResult {
     pub output: Bytes,
     pub gas_used: U256,
 }
-
-pub fn flatten_debug_calls(calls: Vec<ResultDebugCall>) -> Vec<DebugCallFlat> {
```
slowli, line 38: Dumb question: Would this method work here if it was supplied transaction hashes in addition to calls (i.e., `Vec<(ResultDebugCall, H256)>`), assuming that the provided calls correspond to all transactions in in a block sequentially? If so, I'd consider it cleaner than the current approach. It would also allow not loading transaction hashes in a block if it's not required to process the request (although IDK whether this would have any performance impact). Rather, they could be loaded in a separate DB request (I don't think we have it right now, but maybe I've missed it).

core/node/api_server/src/web3/namespaces/debug.rs
```diff
@@ -71,6 +104,66 @@ impl DebugNamespace {
...
+            transaction_position,
+            transaction_hash,
+            r#type: DebugCallType::Call,
+        });
+
+        if !only_top_call {
+            call.calls
+                .into_iter()
+                .enumerate()
+                .for_each(|(number, call)| {
```
slowli, line 152: Nit: `for` cycle is arguably more idiomatic, and would be easier to read. Yes, I understand it's copy-pasted code, but still.

core/lib/dal/src/blocks_web3_dal.rs
```diff
@@ -570,7 +572,11 @@ impl BlocksWeb3Dal<'_, '_> {
         .fetch_all(self.storage)
         .await?
         .into_iter()
-        .map(|call_trace| call_trace.into_call(protocol_version))
+        .map(|call_trace| {
+            let hash = H256::from_slice(&call_trace.hash);
+            let index = call_trace.index_in_block.unwrap_or_default() as usize;
```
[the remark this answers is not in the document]
Deniallugo: Theoretically it's true, but it's strong assumption, i'd rather keep it in this way

core/lib/types/src/debug_flat_call.rs
Deniallugo: The problem with the previous approach, is that it didn't server the needs to have the only method for returning traces. and it also did 2 conversions instead of 1. 
So we first converted to one type of debug calls and then to the next one

core/lib/types/src/api/mod.rs
Deniallugo: I've refactored this part, but I am not sure it makes it better.

core/node/api_server/src/web3/namespaces/debug.rs
```diff
@@ -71,6 +104,63 @@ impl DebugNamespace {
         }
     }
 
+    fn map_flatten_call(
```
slowli, line 107: Bikeshedding: I'd call this method just `flatten_call`.

core/node/api_server/src/web3/namespaces/debug.rs
```diff
@@ -71,6 +104,63 @@ impl DebugNamespace {
...
+                call_type: debug_type,
+                from: call.from,
+                to: call.to,
+                gas: U256::from(call.gas),
+                value: call.value,
+                input: web3::Bytes::from(call.input),
+            },
+            result,
+            subtraces,
+            traceaddress: trace_address.clone(), // Clone the current trace address
```
slowli, line 142: BTW, is this naming correct? [The PR](https://github.com/ethereum/go-ethereum/pull/26377/files) in go-ethereum seems to use `traceAddress` naming. I.e., the field should be named `trace_address` with `#[serde(rename_all)]` enabled.

core/node/api_server/src/web3/namespaces/debug.rs
```diff
@@ -99,53 +186,48 @@ impl DebugNamespace {
...
-    pub async fn debug_trace_block_flat_impl(
-        &self,
-        block_id: BlockId,
-        options: Option<TracerConfig>,
-    ) -> Result<Vec<DebugCallFlat>, Web3Error> {
-        let call_trace = self.debug_trace_block_impl(block_id, options).await?;
-        let call_trace_flat = flatten_debug_calls(call_trace);
-        Ok(call_trace_flat)
+        let mut calls = vec![];
+        let mut flat_calls = vec![];
```
slowli, line 191: This looks awkward. IMO, it'd make sense to "inline" `map_call` code taking the tracer type into account:

```rust
match options.tracer {
    SupportedTracers::FlatCallTracer => {
        let mut flat_calls = vec![];
        for (call, tx_hash, tx_index) in call_traces {
            let mut traces = vec![index];
            Self::map_flatten_call(
                call,
                &mut flat_calls,
                &mut vec![tx_index],
                only_top_call,
                tx_index,
                tx_hash,
            );
        }
        Ok(CallTracerBlockResult::FlatCallTrace(flat_calls))
    }
    // ...
}
```

core/node/api_server/src/web3/namespaces/debug.rs
```diff
@@ -99,53 +186,48 @@ impl DebugNamespace {
...
-            .unwrap_or(false);
+    ) -> Result<Option<CallTracerResult>, Web3Error> {
         let mut connection = self.state.acquire_connection().await?;
         let call_trace = connection
             .transactions_dal()
             .get_call_trace(tx_hash)
             .await
             .map_err(DalError::generalize)?;
-        Ok(call_trace.map(|call_trace| Self::map_call(call_trace, only_top_call)))
+        Ok(call_trace.map(|call_trace| Self::map_call(call_trace, 0, tx_hash, options)))
```
slowli, line 217: Is it OK to return a potentially incorrect `transactionPosition` here? It may make sense to separate the passed trace address index from a transaction index in general; the other case when these two values differ is in `debug_traceBlock*` methods if traces for *some* transactions in a block are absent for whatever reason (although in all fairness, this missing data will likely make the caller upset in any case).

core/node/api_server/src/web3/tests/vm.rs
```diff
@@ -557,12 +567,19 @@ impl HttpTest for TraceCallTestAfterSnapshotRecovery {
     ) -> anyhow::Result<()> {
         self.fee_input.expect_default(TraceCallTest::FEE_SCALE);
         let call_request = CallTest::call_request(b"pending");
-        let call_result = client.trace_call(call_request.clone(), None, None).await?;
+        let CallTracerResult::CallTrace(call_result) =
```
slowli, line 570: Nit: You can introduce conversion methods for `CallTracerResult` like `unwrap_hierarchical` (`unwrap_default`?) / `unwrap_flat` to make them shorter.

core/lib/types/src/api/mod.rs
slowli: Obviously subjective, but I do like it better 👍

core/lib/types/src/api/mod.rs
```diff
@@ -701,33 +705,100 @@ impl ProtocolVersion {
...
 
+/// Result tracers need to have a nested result field for compatibility. So we have two different
+/// structs 1 for blocks tracing and one for txs and call tracing
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(untagged)]
+pub enum CallTracerBlockResult {
+    CallTrace(Vec<ResultDebugCall>),
+    FlatCallTrace(Vec<DebugCallFlat>),
+}
+impl CallTracerBlockResult {
```
slowli, line 755: Nit: Missing empty line between items.

core/lib/types/src/api/mod.rs
```diff
@@ -701,33 +705,100 @@ impl ProtocolVersion {
...
+    pub fn unwrap_flatten(self) -> Vec<DebugCallFlat> {
+        match self {
+            Self::CallTrace(_) => {
+                unreachable!()
```
slowli, line 759: Nit: Semantically, it's a `panic!`; there's no local reasoning why this variant is unreachable. Ditto for `unwrap_default`.

core/lib/types/src/api/mod.rs
```diff
@@ -701,33 +705,100 @@ impl ProtocolVersion {
...
+/// Result tracers need to have a nested result field for compatibility. So we have two different
+/// structs 1 for blocks tracing and one for txs and call tracing
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(untagged)]
+pub enum CallTracerBlockResult {
+    CallTrace(Vec<ResultDebugCall>),
+    FlatCallTrace(Vec<DebugCallFlat>),
+}
+impl CallTracerBlockResult {
+    pub fn unwrap_flatten(self) -> Vec<DebugCallFlat> {
```
slowli, line 756: Bikeshedding: IMO, `unwrap_flat` would be a better name; after all, the returned struct is called `_Flat`, not `_Flatten`, and there's no flattening happening inside the method.

core/node/api_server/src/web3/namespaces/debug.rs
```diff
@@ -222,6 +318,6 @@ impl DebugNamespace {
             revert_reason,
             result.call_traces,
         );
-        Ok(Self::map_call(call, false))
+        Ok(Self::map_call(call, 0, hash, options))
```
slowli, line 321: Just to clarify: Do we know what transaction hash / index geth returns for `debug_traceCall`? Is it zeroed, or skipped from the response?

slowli (dismissed):

core/node/api_server/src/web3/namespaces/debug.rs
Deniallugo: Its unpredictable and doesn't make much sense

slowli (approved):

PR #2927, opened 2024-09-19
fix(eth-sender): print better error message in case of missing blob prices

tomg10:


core/node/eth_sender/src/eth_fees_oracle.rs
```diff
@@ -40,6 +40,12 @@ impl GasAdjusterFeesOracle {
...
     ) -> Result<EthFees, EthSenderError> {
         let base_fee_per_gas = self.gas_adjuster.get_blob_tx_base_fee();
+        if base_fee_per_gas == 0 {
+            panic!(
+                "L1 RPC incorrectly reported blob prices, either it doesn't return them at all \
+            or returns 0's, eth-sender cannot continue without proper blob prices!"
+            );
+        }
         let priority_fee_per_gas = self.gas_adjuster.get_blob_tx_priority_fee();
         let blob_base_fee_per_gas = Some(self.gas_adjuster.get_blob_tx_blob_base_fee());
```
perekopskiy, line 50: You are comparing `base_fee_per_gas` and panic message is about blob prices. Let's do it like
```
if base_fee_per_gas == 0 || blob_base_fee_per_gas == 0 { panic!(..) }
```

core/node/eth_sender/src/eth_fees_oracle.rs
```diff
@@ -40,6 +40,12 @@ impl GasAdjusterFeesOracle {
         previous_sent_tx: &Option<TxHistory>,
     ) -> Result<EthFees, EthSenderError> {
         let base_fee_per_gas = self.gas_adjuster.get_blob_tx_base_fee();
```
perekopskiy, line 42: it also makes sense to check if `base_fee_per_gas == 0` in `calculate_fees_no_blob_sidecar `

perekopskiy (dismissed):

perekopskiy (approved):

PR #2931, opened 2024-09-20
refactor: Improve EN Grafana dashboards

brunoffranca:
## What ❔
Adding more information to the EN Grafana dashboards, which barely had any.

## Why ❔
Better observability.

Fixes BFT-507


brunoffranca:
Checked that it works on my EN. Some documentation fixes are still to be done (on the external node guide).

Deniallugo:
why it's not a part of era observability?

brunoffranca:
It's simpler for the user, there's no need to start a separate service. Besides we already start a Prometheus and Grafana containers together with the EN, it doesn't make sense to start yet another set of containers with era-observability.

pompon0 (approved):
Approved, but it is not like I manually reviewed the generated json files :P

brunoffranca:
Well, you can always import the dashboard.

PR #2934, opened 2024-09-23
feat(zk_toolbox): Add options for selective installation of zk_inception and zk_supervisor

matias-gonz:
## What ❔

- Introduced `--inception` and `--supervisor` options to allow users to specify which binaries to install.
- Removed the `ZKUP_SKIP_ZK_SUPERVISOR` option.

## Why ❔

Make it clear what to install, focusing on what to include rather what to exclude


sanekmelnikov (approved):
lgtm

PR #2937, opened 2024-09-23
perf(api): More efficient gas estimation

slowli:
## What ❔

Makes gas estimation more efficient by using heuristics similar to ones used by Ethereum clients:

- Defines the lower gas limit bound as the gas consumed with an effectively infinite gas limit.
- Defines an initial binary search pivot as a slightly larger value so that in the best / average case, most of the search space is immediately discarded.

## Why ❔

Improves gas estimation latency in the best and average case.

## Checklist


core/node/api_server/src/execution_sandbox/storage.rs
```diff
@@ -56,6 +56,10 @@ impl<S: ReadStorage> StorageWithOverrides<S> {
                 let code_key = get_code_key(account);
                 let code_hash = code.hash();
                 self.overridden_slots.insert(code_key, code_hash);
+                // FIXME: is this addition correct?
+                let known_code_key = get_known_code_key(&code_hash);
+                self.overridden_slots
+                    .insert(known_code_key, H256::from_low_u64_be(1));
```
slowli, line 62: Without adding a bytecode to the known bytecodes storage, decommitting it via the code oracle will fail (which seems to differ from how real contracts work). OTOH, I'm not 100% certain that adding this slot override won't have negative side effects.

core/node/api_server/src/tx_sender/gas_estimation.rs
```diff
@@ -0,0 +1,511 @@
...
+use super::{result::ApiCallResult, SubmitTxError, TxSender};
+use crate::execution_sandbox::{BlockArgs, SandboxAction, VmPermit, SANDBOX_METRICS};
+
+#[derive(Debug, Clone, Copy)]
+pub(crate) enum BinarySearchKind {
+    /// Full binary search.
+    Full,
+    /// Binary search with an optimized initial pivot.
+    #[allow(dead_code)] // FIXME
+    Optimized,
```
slowli, line 28: Do we want to use optimized binary search from the get go, or make it opt-in? Right now, it's only used in tests, hence `allow(dead_code)`. Even without enabling it, we can still collect metrics for the lower bound / initial pivot (see below).

core/node/api_server/src/tx_sender/gas_estimation.rs
```diff
@@ -0,0 +1,511 @@
...
+                });
+                (lower_bound..=upper_bound, initial_pivot)
+            }
+        };
+
+        let unscaled_gas_limit =
+            Self::binary_search(&estimator, bounds, initial_pivot, acceptable_overestimation)
+                .await?;
+        // Metrics are intentionally reported regardless of the binary search mode, so that the collected stats can be used to adjust
+        // optimized binary search params (e.g., the initial pivot multiplier).
```
slowli, line 80: Does it make sense to log the lower bound / initial gas limit, or would it produce too much spam?

core/node/api_server/src/tx_sender/gas_estimation.rs
```diff
@@ -0,0 +1,511 @@
...
+        }
+        total_gas_bound.checked_sub(self.operator_overhead)
+    }
+
+    pub fn optimistic_gas_limit_without_overhead(&self) -> Option<u64> {
+        let gas_charged_without_overhead = self
+            .total_gas_charged?
+            .checked_sub(self.operator_overhead)?;
+        // 21/20 is an empirical multiplier. It is higher than what empirically suffices for some common transactions;
+        // one can argue that using 64/63 multiplier would be more accurate due to the 63/64 rule for far calls
```
slowli, line 248: More precisely, judging by unit tests, the only exception I've found so far is a transaction decommitting a very large bytecode (~320kB).

core/node/api_server/src/tx_sender/gas_estimation.rs
```diff
@@ -0,0 +1,511 @@
...
+            // and getting how much pubdata it used.
+
+            let (result, _) = self
+                .unadjusted_step(self.max_gas_limit)
+                .await
+                .context("estimate_gas step failed")?;
+            // If the transaction has failed with such a large gas limit, we return an API error here right away,
+            // since the inferred gas bounds would be unreliable in this case.
+            // FIXME: double-check that returning early is OK
+            result.check_api_call_result()?;
```
slowli, line 369: Subj. I *think* it's OK, at least if the transaction success / failure is monotonic w.r.t. the gas limit. In the old implementation, such a transaction would end binary search with the gas limit equal to the upper bound, and then would fail on `into_api_call_result()`. It's difficult to analyze implications in the general case though.

slowli:
I'd imagine that switching between full and optimized binary search would be a node config parameter (probably opt-in). I'd prefer to add it in a separate PR, since this one is large as-is (although a significant part of it are unit tests). OTOH, not enabling it for integration tests decreases coverage. (I've run integration tests locally with the optimized binary search; seems to work fine.)

core/node/api_server/src/execution_sandbox/storage.rs
Deniallugo: Yep, that's correct

slowli:
Collected some stats w.r.t. the number of binary search iterations in the old and new approach, using the added gas estimation unit tests (i.e., transfers, various parameterizations of load test transactions etc.). With `acceptable_overestimation == 1_000` (the default value), the number of iterations in the old approach is almost always 17; in the new approach, it varies from 5 to 9, with the average being ~7; so it's ~2.5x improvement. I think the number of iterations could be improved further with skewed binary search as discussed privately, but IMO, it's a good improvement as is (and especially if the new VM will be used on top of it).

core/node/api_server/src/tx_sender/gas_estimation.rs
perekopskiy: I also think it's OK

core/node/api_server/src/tx_sender/gas_estimation.rs
```diff
@@ -0,0 +1,533 @@
...
+
+        let optimized_lower_bound = initial_estimate.lower_gas_bound_without_overhead();
+        // Perform an initial search iteration with the pivot slightly greater than `gas_used` to account for 63/64 rule for far calls etc.
+        // If the transaction succeeds, it will discard most of the search space at once.
+        let optimistic_gas_limit = initial_estimate.optimistic_gas_limit_without_overhead();
+
+        let (bounds, initial_pivot) = match kind {
+            BinarySearchKind::Full => {
+                let lower_bound = initial_estimate.gas_charged_for_pubdata;
+                let upper_bound = MAX_L2_TX_GAS_LIMIT;
```
perekopskiy, line 62: `upper_bound` should be `initial_estimate.gas_charged_for_pubdata + MAX_L2_TX_GAS_LIMIT`, that's how it works in old code, no? `MAX_L2_TX_GAS_LIMIT`'s name is somewhat wrong, in fact it means `MAX_L2_TX_COMPUTATIONAL_GAS_LIMIT`

core/node/api_server/src/tx_sender/gas_estimation.rs
```diff
@@ -0,0 +1,533 @@
...
+                (lower_bound..=upper_bound, None)
+            }
+            BinarySearchKind::Optimized => {
+                let lower_bound = optimized_lower_bound.unwrap_or(0);
+                let upper_bound = MAX_L2_TX_GAS_LIMIT;
```
perekopskiy, line 67: same here

perekopskiy (commented):
Looks nice!

core/node/api_server/src/execution_sandbox/storage.rs
slowli: Sorry, I phrased it really weirdly. I meant that if this additional log is not added, decommitments don't work; deploying real contract adds such a log, so decommitment works for them.

core/node/api_server/src/tx_sender/gas_estimation.rs
RomanBrodetski: I would log it TBH - it's less than ~30 per second currently, will not make an impact on the total API logs

perekopskiy (approved):

RomanBrodetski (dismissed):
Looks good!
Let's follow the deployment on all envs - we should see the change in response times (and potentially in estimation v alues) on stage already once this is merged (https://grafana.matterlabs.dev/d/cb075009-049f-44ed-81c5-2b73dae42c16/era-watchdog?orgId=1&var-datasource=mIPIusa4k)

core/node/api_server/src/tx_sender/gas_estimation.rs
RomanBrodetski: @slowli is it worth it to rename the var? Feel free to keep as is too

perekopskiy (approved):

PR #2943, opened 2024-09-23
fix(vm): small fixes for sync-layer-stable

perekopskiy:
## What ❔

fixes state diff filter (leftover from debug)
uses correct protocol version id to get log tree size

## Checklist


.github/workflows/ci-core-reusable.yml
```diff
@@ -60,10 +60,6 @@ jobs:
           ci_run ./bin/zkt
           ci_run zk_supervisor contracts
 
-      # FIXME: enable contract tests once tehy are stable
-      #- name: Contracts unit tests
-      #  run: ci_run yarn l1-contracts test
```
StanislavBreadless, line 65: we do run those in the main branch:  https://github.com/matter-labs/zksync-era/blob/93bc66f21f9f67a440f06f1c4402e0d687698741/.github/workflows/ci-core-reusable.yml#L63

but I agree that we do not really need them since we already have tests in the era-contracts repo

core/lib/multivm/src/versions/vm_latest/tracers/pubdata_tracer.rs
```diff
@@ -51,13 +53,15 @@ impl<S: WriteStorage> PubdataTracer<S> {
         l1_batch_env: L1BatchEnv,
         execution_mode: VmExecutionMode,
         subversion: MultiVMSubversion,
+        protocol_version: ProtocolVersionId,
```
StanislavBreadless, line 56: why do we need protocol version here? it is for `vm_latest`

.github/workflows/ci-core-reusable.yml
perekopskiy: sorry, missed it, I will add it back and let's remove separately

StanislavBreadless (approved):

PR #2949, opened 2024-09-24
feat(en): periodically fetch bridge addresses

perekopskiy:
## What ❔

Periodically fetch bridge addresses

## Why ❔

Addresses will be changed during gateway upgrade

## Checklist


core/node/api_server/src/web3/state.rs
```diff
@@ -241,6 +247,51 @@ impl SealedL2BlockNumber {
...
+        initial_value: BridgeAddresses,
+        update_interval: Duration,
+        stop_receiver: watch::Receiver<bool>,
+    ) -> (Self, impl Future<Output = anyhow::Result<()>>) {
+        let this = Self(Arc::new(RwLock::new(initial_value)));
+        let updater = this.clone();
+
+        let update_task = async move {
+            loop {
+                if *stop_receiver.borrow() {
```
slowli, line 270: Nit: You can use `while !*stop_receiver.borrow()` as the loop statement instead, and move the log after the loop.

core/node/api_server/src/web3/state.rs
```diff
@@ -241,6 +247,51 @@ impl SealedL2BlockNumber {
...
+
+                match main_node_client.get_bridge_contracts().await {
+                    Ok(bridge_addresses) => {
+                        *updater.0.write().await = bridge_addresses;
+                    }
+                    Err(err) => {
+                        tracing::error!("Failed to query `get_bridge_contracts`, error: {err:?}");
+                    }
+                }
+                tokio::time::sleep(update_interval).await;
```
slowli, line 283: It makes sense to use [the `timeout` construction](https://github.com/matter-labs/zksync-era/blob/79b6fcf8b5d10a0ccdceb846370dd6870b6a32b5/core/node/api_server/src/web3/pubsub.rs#L74-L76), so that the node shutdown is more responsive. TBH, this is not done in the neighboring `SealedL2BlockNumber` task, but it's expected to have much lower sleep interval, at least judging by the default values. (Although ideally, it can be refactored as well.)

core/node/api_server/src/web3/mod.rs
```diff
@@ -285,6 +297,11 @@ impl ApiBuilder {
         self
     }
 
+    pub fn with_main_node_client(mut self, main_node_client: Box<DynClient<L2>>) -> Self {
```
slowli, line 300: Bikeshedding: IMO, it's not clear that setting the client determines whether the bridge addresses task will be enabled. IMO, it'd make sense to accept `BridgeAddressesHandle` directly; this would be more in line with more modular DI we use almost (?) everywhere except for the API server because of legacy reasons.

slowli: Yep, LGTM. Maybe makes sense to ask @popzxc's opinion as well.

core/node/api_server/src/web3/state.rs
```diff
@@ -243,6 +213,29 @@ impl SealedL2BlockNumber {
     }
 }
 
+impl Default for SealedL2BlockNumber {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
slowli, line 220: Nit: Can't this implementation be derived?

core/node/node_framework/src/implementations/layers/web3_api/server/sealed_l2_block.rs
```diff
@@ -0,0 +1,50 @@
+use std::time::Duration;
+
+use zksync_dal::{Core, CoreDal};
+use zksync_db_connection::connection_pool::ConnectionPool;
+use zksync_node_api_server::web3::state::SealedL2BlockNumber;
+
+use crate::{StopReceiver, Task, TaskId};
+
+#[derive(Debug)]
+pub struct SealedL2BlockUpdaterTask {
```
slowli, line 10: Nit: AFAIU, the idea with the `node_framework` crate is to eventually invert its dependency on node components (so that components depend on the framework). So, it makes slightly more sense to define tasks in the component crates so they don't need to be moved there in the future. This is a very minor thing however.

core/node/node_framework/src/implementations/layers/web3_api/server/bridge_addresses.rs
```diff
@@ -0,0 +1,47 @@
...
+}
+
+#[async_trait::async_trait]
+impl Task for BridgeAddressesUpdaterTask {
+    fn id(&self) -> TaskId {
+        "bridge_addresses_updater_task".into()
+    }
+
+    async fn run(self: Box<Self>, mut stop_receiver: StopReceiver) -> anyhow::Result<()> {
+        let update_interval = self.update_interval.unwrap_or(Duration::from_secs(30));
```
slowli, line 25: Nit: Could make sense to extract `Duration::from_secs(30)` to a constant.

core/node/node_framework/src/implementations/layers/web3_api/server/bridge_addresses.rs
```diff
@@ -0,0 +1,47 @@
...
+
+        while !*stop_receiver.0.borrow_and_update() {
+            match self.main_node_client.get_bridge_contracts().await {
```
slowli, line 28: Nit: Since calling the main node could take a large time, there *may* make sense to use `tokio::time::interval` here. OTOH, it may put additional load to an already struggling server, so maybe it's better to retain the current approach.

slowli (approved):

core/node/node_framework/src/implementations/layers/web3_api/server/bridge_addresses.rs
perekopskiy: IMO it doesn't matter, in practice bridge contracts will only be changed once and it's not a big deal if they will reach ENs not in 30 seconds but 35

core/node/node_framework/src/implementations/layers/web3_api/server/sealed_l2_block.rs
perekopskiy: Do you mean to define tasks in component crate but implement in `node_framework`? If so then I find it confusing (for now) and would prefer to do it later when we start inverting globally

core/lib/config/src/configs/en_config.rs
```diff
@@ -19,4 +19,5 @@ pub struct ENConfig {
     pub main_node_rate_limit_rps: Option<NonZeroUsize>,
 
     pub gateway_url: Option<SensitiveUrl>,
+    pub bridge_addresses_refresh_interval_sec: Option<NonZeroU64>,
```
StanislavBreadless, line 22: if a partner forgets to provide the value, I guess the consequences will be uncomfortable, right? maybe it makes sense to replace it with some sensible default constant in case the value is not provided

StanislavBreadless (approved):

core/lib/config/src/configs/en_config.rs
perekopskiy: @StanislavBreadless I'm merging it for now, please reply and I will adjust later if required

PR #2950, opened 2024-09-24
feat(zk_toolbox): Add SQL format for zk supervisor

matias-gonz:
## What ❔
Add SQL format for zk supervisor


zk_toolbox/crates/zk_supervisor/Cargo.toml
```diff
@@ -29,3 +29,4 @@ futures.workspace = true
 types.workspace = true
 serde_yaml.workspace = true
 zksync_basic_types.workspace = true
+sqruff-lib = { git = "https://github.com/quarylabs/sqruff", version = "0.18.2" }
```
perekopskiy, line 32: any reason to not use crate from crates.io?

perekopskiy: also, did you consider https://crates.io/crates/sqlformat? It has much more downloads

matias-gonz: Yes! They still didn't publish 1.18.2 (I already asked on their slack channel) so I had to import it like that.
I did try sqlformat but we weren't happy with the formatted queries.

perekopskiy (approved):

zk_toolbox/crates/zk_supervisor/src/commands/sql_fmt.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    let base_indent = line
+        .find(|c: char| !c.is_whitespace())
+        .context(MSG_FAILED_TO_DETERMINE_BASE_INDENT)?
+        + 4;
+
+    let prefix = &line[..query_start];
+
+    let query_end = if is_raw_string {
+        line.find(r#""#)
```
slowli, line 90: Does it work like that? IIUC, `r#""#` is an empty string; you'd probably want `"\"#"`.

core/lib/dal/src/blocks_dal.rs
```diff
@@ -986,14 +986,21 @@ impl BlocksDal<'_, '_> {
         sqlx::query!(
             r#"
             INSERT INTO
```
slowli, line 988: Another example of lenient formatting: the formatter would accept here

```rust
        sqlx::query!(
            r#"
            INSERT INTO commitments (
                l1_batch_number,
                events_queue_commitment,
                bootloader_initial_content_commitment
            )
            VALUES ($1, $2, $3)
            ON CONFLICT (l1_batch_number) DO NOTHING
            "#,
            _
        )
```

or even a monstrosity like

```rust
        sqlx::query!(
            r#"
            INSERT
            INTO commitments (
                l1_batch_number, events_queue_commitment,
                bootloader_initial_content_commitment
            )
            VALUES (
                $1, $2, $3
            )
            ON CONFLICT
            (l1_batch_number)
            DO NOTHING
            "#,
            _
        )
```

slowli (commented):
I don't quite like the new formatter because it looks overly lenient. For example, take this mangled query from `events_dal`:

```rust
        let logs: Vec<_> = sqlx::query_as!(
            StorageWeb3Log,
            r#"
            SELECT  address,
                topic1,
                topic2,
              topic3,
                topic4, value,
                NULL::bytea AS "block_hash",
                NULL::bigint AS "l1_batch_number?",
                miniblock_number ,
                tx_hash,
                   tx_index_in_block,
                event_index_in_block,
                event_index_in_tx, NULL::BIGint AS "block_timestamp?"
            FROM events
            WHERE
                tx_hash  = ANY     ( $1 )
            ORDER  BY
                miniblock_number ASC, event_index_in_block ASC
            "#,
            &hashes[..] as &[&[u8]],
        )
```

Its formatted variant is

```rust
        let logs: Vec<_> = sqlx::query_as!(
            StorageWeb3Log,
            r#"
            SELECT
                address,
                topic1,
                topic2,
                topic3,
                topic4, value,
                NULL::bytea AS "block_hash",
                NULL::bigint AS "l1_batch_number?",
                miniblock_number,
                tx_hash,
                tx_index_in_block,
                event_index_in_block,
                event_index_in_tx, NULL::bigint AS "block_timestamp?"
            FROM events
            WHERE
                tx_hash = ANY($1)
            ORDER BY
                miniblock_number ASC, event_index_in_block ASC
            "#,
            &hashes[..] as &[&[u8]],
        )
```

That is, the formatter doesn't have an opinion on how many list items are put on a single line. Obviously, having *any* sort-of reasonable formatter is better than not having one, but maybe it may be tweaked to be more opinionated?

core/lib/dal/src/blocks_dal.rs
Deniallugo: But, i'd say the original option was even worse, no?

Deniallugo:
@slowli Actually it support pretty big configuration https://docs.sqlfluff.com/en/1.3.0/configuration.html

we are free to set it more precise

matias-gonz:
@slowli @Deniallugo 
Actually these are the docs:
https://github.com/quarylabs/sqruff/blob/main/docs/rules.md

Deniallugo (approved):

Deniallugo (approved):

PR #2954, opened 2024-09-24
feat(da-clients): add secrets

dimazhornyk:
## What ❔

Add an opportunity to store the sensitive information of DA clients in the secrets.

## Why ❔

Some sensitive info is currently stored as plaintext, which is not secure, we should use secrets for those values

## Checklist


core/node/da_clients/src/avail/sdk.rs
```diff
@@ -40,7 +40,7 @@ struct BoundedVec<_0>(pub Vec<_0>);
 impl RawAvailClient {
     pub(crate) const MAX_BLOB_SIZE: usize = 512 * 1024; // 512kb
 
-    pub(crate) async fn new(app_id: u32, seed: String) -> anyhow::Result<Self> {
+    pub(crate) async fn new(app_id: u32, seed: &String) -> anyhow::Result<Self> {
```
Deniallugo, line 43:
```suggestion
    pub(crate) async fn new(app_id: u32, seed: &str) -> anyhow::Result<Self> {
```

Deniallugo (approved):

PR #2956, opened 2024-09-25
build: Set consensus and pruning as default for EN

brunoffranca:
## What ❔
ENs, started from docker compose, now have consensus and pruning enabled as default. Also, the EN docs were updated.


RomanBrodetski (dismissed):

docs/guides/external-node/00_quick_start.md
```diff
@@ -10,28 +10,28 @@ To start a mainnet instance, run:
 
 ```sh
 cd docker-compose-examples
-docker compose --file mainnet-external-node-docker-compose.yml up
+sudo docker compose --file mainnet-external-node-docker-compose.yml up
```
pompon0, line 13: pls don't add sudo to commands. It really depends on the setup on whether sudo is needed or not

brunoffranca: The default setup requires you to use sudo. If you change your setup to not require calling sudo, then you know that.

pompon0 (approved):

PR #2957, opened 2024-09-25
chore(prover-gateway): Speed up polling

EmilLuta:
16 minutes is just ridiculous. This should've been done a long time ago, but has been avoided due to other priorities and always put on the back-burner/forgotten. Today we change it.

### DIFF
--- a/etc/env/base/fri_prover_gateway.toml
+++ b/etc/env/base/fri_prover_gateway.toml
@@ -1,6 +1,6 @@
 [fri_prover_gateway]
 api_url="http://127.0.0.1:3320"
-api_poll_duration_secs=1000
+api_poll_duration_secs=15
 prometheus_listener_port=3314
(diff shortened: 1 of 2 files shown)


etc/env/base/fri_prover_gateway.toml
```diff
@@ -1,6 +1,6 @@
 [fri_prover_gateway]
 api_url="http://127.0.0.1:3320"
-api_poll_duration_secs=1000
+api_poll_duration_secs=15
```
popzxc, line 3: It's a localhost config, why not 3 or 5?

popzxc (approved):
I guess it was `ms` at some point, what could go wrong 😅

etc/env/base/fri_prover_gateway.toml
popzxc: Spam logs must be lowered in priority, probably.
No strong opinion though.

PR #2959, opened 2024-09-25
refactor: Deleted unnecessary files

brunoffranca:
## What ❔

- ~`.cargo/config.toml` is a remnant from when we used private repos. It should no longer be necessary.~
- ~`configs/` doesn't seem to be used at all.~
- `building-from-scratch` just contains an old docker file that was used for Coinbase. No longer used.


.cargo/config.toml
```diff
@@ -1,2 +0,0 @@
-[net]
-git-fetch-with-cli = true
```
Deniallugo, line 2: If we one day want to use private dependencies. 
e.g. some critical fixes, we will need to install this value again

configs/.gitkeep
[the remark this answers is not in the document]
brunoffranca: Got it. Can't zk toolbox just create an empty folder when initializing an ecosystem if there's none?

.cargo/config.toml
brunoffranca: To fix a critical issue we most likely just create a private copy of the repo. We can set the value again if the need arises anyway.

Deniallugo: I just don't want to fight with it in a rush.

PR #2960, opened 2024-09-25
fix: fix state_diff_hash usage

perekopskiy:
## What ❔

- squashes some migration (there is no need for them to be separate)
- makes `state_diff_hash`, `local_root`, `aggregated_root` optional in L1BatchMetadata
- `metadata.state_diff_hash` is used for new batches, system log is used for old batches
- removes pre boojum code from proof_data_handler 

## Checklist


core/lib/constants/src/system_logs.rs
```diff
@@ -1,8 +1,8 @@
 /// The key of the system log with value of the L2->L1 logs tree root hash
 pub const L2_TO_L1_LOGS_TREE_ROOT_KEY: u32 = 0;
 
-/// The key of the system log with value of the state diff hash
-pub const STATE_DIFF_HASH_KEY: u32 = 2;
+/// The key of the system log with value of the state diff hash for pre-gateway protocol versions
```
mm-zk, line 4: what is in that slot AFTER gateway? 
(did we move slot numbers or do we simply put empty data there)

core/node/api_server/src/web3/namespaces/zks.rs
```diff
@@ -352,14 +352,14 @@ impl ZksNamespace {
         let (root, mut proof) = MiniMerkleTree::new(merkle_tree_leaves, Some(tree_size))
             .merkle_root_and_path(l1_log_index);
 
+        // FIXME Definitely refactor all of it
```
mm-zk, line 355: seems we have a bunch of 'helper' printf here...

Can you create a linear ticket and add it here (so that we don't forget).

core/lib/types/src/commitment/mod.rs
```diff
@@ -95,8 +92,9 @@ pub struct L1BatchMetadata {
     pub bootloader_initial_content_commitment: Option<H256>,
     pub state_diffs_compressed: Vec<u8>,
 
-    pub aggregation_root: H256,
-    pub local_root: H256,
+    pub state_diff_hash: Option<H256>,
```
mm-zk, line 95: can you add a comment in which circuimstances these are None?

mm-zk (approved):

core/lib/constants/src/system_logs.rs
perekopskiy: we've moved slots around, IIUC some were removed and some new were added. After gateway `PREV_BATCH_HASH_KEY` has key = 2

core/node/api_server/src/web3/namespaces/zks.rs
perekopskiy: I don't want to create a ticket because I'm not yet sure what to put there. It's on my radar, I will get to it soon

perekopskiy:
Merging cause I only added comments after approve. Feel free to leave additional comment, I can address them in next PR

PR #2962, opened 2024-09-25
fix: ignore unknown fields in rpc json response

pompon0:
In some previous pr, I've added a field to json RPC response (consensusGlobalConfig) and it turned out to be incompatible due to the default parsing behavior: an error was returned when unknown fields were present. In general, for programmatic communications the recommended behaviour is to silently ignore unknown fields, so that adding new fields is compatible.

I've also weakened the restriction on accepting new consensus genesis by external nodes - when the unknown fields are silently dropped, it is possible that the node will perceive genesis with the same fork number differently after a binary update (because it receives more fields).


Cargo.lock
```diff
@@ -2319,18 +2319,6 @@ dependencies = [
  "subtle",
 ]
 
-[[package]]
-name = "ff_ce"
-version = "0.14.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5b538e4231443a5b9c507caee3356f016d832cf7393d2d90f03ea3180d4e3fbc"
-dependencies = [
```
pompon0, line 2327: I'll set the correct dependency once https://github.com/matter-labs/era-consensus/pull/200 needs is merged and published

RomanBrodetski (approved):

PR #2966, opened 2024-09-25
feat: Evm Equivalence

IAvecilla:
## What ❔

This PR contains the rest of the changes required for EVM Equivalence. Changes have been made to make sure it can be merged to `main` currently without having to point to an `era-contracts` branch containing the EVM interpreter and the EVM Gas Manager. The changes required for this are all [here](https://github.com/lambdaclass/zksync-era/pull/276). Also, the evm simulator hash and bytecode are currently set to be the same as the default AA contract as otherwise things will not work until the evm interpreter is present in the `era-contracts` branch.

The EVM equivalence changes are:

- A `use_evm_simulator` (defaulted to `false`) config value to set whether the zk stack is initialized with the evm equivalence feature or not. Most changes below are conditional on whether this flag is set or not.
- A new Postgres migration is added with an `evm_simulator_code_hash` column added to the following tables:
  - `protocol_versions`
  - `l1_batches`
  - `miniblocks`
The corresponding `dal` modules were updated to account for these new columns.
- The `EvmInterpreter` is added as a base system contract.
- The `decommiter` present in the `vm_latest` now has separate hash calculations for each version, `ZKEvm` and `EVM`. In particular the length of the pre-image differs (EVM is in bytes, zkEVM in words, i.e. 32 bytes)
- An `EvmDeployTracer` is added as another default tracer. Its purpose is to add EVM contracts to the decommiter once an evm deployment ends succesfully.
- For the `vm_latest`, the `TransactionData` now checks if the address is absent and in the afirmative case it writes the corresponding flag on the reserved bit to signal an EVM contract deployment.
- The decommitment processor for the `vm_latest` now has the hash and code of the EVM interpreter.
- The `EthTxAggregator` now adds the call to the getters of the hash and code of the evm interpreter to the multicall.
- The `main_executor` calculates a `new_known_factory_deps` with all the new known EVM bytecodes with the help of the decommiter. These are then added to the regular factory deps.

## Checklist


core/bin/system-constants-generator/Cargo.toml
```diff
@@ -15,6 +15,8 @@ zksync_types.workspace = true
 zksync_utils.workspace = true
 zksync_contracts.workspace = true
 zksync_multivm.workspace = true
+zksync_config.workspace = true
+zksync_env_config.workspace = true
```
slowli, line 19: This change looks unnecessary; see the overall comment.

core/lib/multivm/src/versions/vm_latest/tests/nonce_holder.rs
```diff
@@ -101,7 +102,7 @@ fn test_nonce_holder() {
     run_nonce_test(
         1u32,
         NonceHolderTestMode::SetValueUnderNonce,
-        Some("Previous nonce has not been used".to_string()),
+        Some("Error function_selector = 0x13595475, data = 0x13595475".to_string()),
```
slowli, line 105: These tests seem to fail locally. Have you forgot to check in the `contracts` submodule? I see it's present [in the follow-up PR](https://github.com/matter-labs/zksync-era/pull/2967), but not here.

core/lib/dal/migrations/20240911161714_evm-simulator.down.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE protocol_versions DROP COLUMN IF NOT EXISTS evm_simulator_code_hash;
+ALTER TABLE l1_batches DROP COLUMN IF NOT EXISTS evm_simulator_code_hash;
+ALTER TABLE miniblocks DROP COLUMN IF NOT EXISTS evm_simulator_code_hash;
```
slowli, line 3: Is this still necessary given that the "up" migration doesn't create this column?

core/lib/dal/migrations/20240911161714_evm-simulator.down.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE protocol_versions DROP COLUMN IF NOT EXISTS evm_simulator_code_hash;
```
slowli, line 1: IIUC, it should be `IF EXISTS` (without `NOT`).

core/lib/contracts/src/lib.rs
```diff
@@ -331,13 +340,35 @@ impl BaseSystemContracts {
         let hash = hash_bytecode(&bytecode);
 
         let default_aa = SystemContractCode {
-            code: bytes_to_be_words(bytecode),
+            code: bytes_to_be_words(bytecode.clone()),
             hash,
         };
 
+        // If evm simulator is not enabled, use the default account bytecode and hash.
+        let (evm_simulator_bytecode, evm_simulator_hash) =
```
slowli, line 348: This adds the EVM simulator retroactively for all protocol versions. I'd prefer not to do that. See the overall comment for more discussion.

core/lib/multivm/src/versions/vm_latest/utils/mod.rs
```diff
@@ -1,6 +1,58 @@
 /// Utility functions for the VM.
+use ethabi;
```
slowli, line 2: Nit: This is useless.

core/lib/multivm/src/versions/vm_latest/utils/mod.rs
```diff
@@ -1,6 +1,58 @@
...
+use once_cell::sync::Lazy;
+use zk_evm_1_4_1::sha2;
```
slowli, line 4: Why not import `sha2` from `zk_evm_1_5_0` like the items below?

core/lib/multivm/src/versions/vm_latest/utils/mod.rs
```diff
@@ -1,6 +1,58 @@
...
+use zksync_types::{H256, KNOWN_CODES_STORAGE_ADDRESS};
+use zksync_vm_interface::VmEvent;
+
 pub mod fee;
 pub mod l2_blocks;
 pub(crate) mod logs;
 pub mod overhead;
 pub mod transaction_encoding;
+
+/// TODO: maybe move to a different folder
```
slowli, line 18: IMO, it belongs to `zksync_utils` near `hash_bytecode`.

core/lib/multivm/src/versions/vm_latest/vm.rs
```diff
@@ -79,6 +79,23 @@ impl<S: WriteStorage, H: HistoryMode> Vm<S, H> {
         self.state.local_state.callstack.current.ergs_remaining
     }
 
+    pub(crate) fn ask_decommitter(&self, hashes: Vec<H256>) -> Vec<Vec<u8>> {
+        let mut result = vec![];
+        for hash in hashes {
```
slowli, line 84: Nit: `hashes` aren't really consumed, so you can pass `&[H256]`. Also, since bytecode hashes correspond 1-to-1 to resolved bytecodes, it's possbile to use `hashes.iter().map(..).collect()` workflow.

core/lib/protobuf_config/src/proto/config/genesis.proto
```diff
@@ -28,5 +28,6 @@ message Genesis {
   optional Prover prover = 10;
   optional L1BatchCommitDataGeneratorMode l1_batch_commit_data_generator_mode = 29; // optional, default to rollup
   optional string genesis_protocol_semantic_version = 12; // optional;
+  optional string evm_simulator_hash = 13; // required; h256
```
slowli, line 31: No, it's not required.

core/lib/vm_interface/src/types/outputs/execution_result.rs
```diff
@@ -118,6 +118,7 @@ pub struct VmExecutionResultAndLogs {
     pub logs: VmExecutionLogs,
     pub statistics: VmExecutionStatistics,
     pub refunds: Refunds,
+    pub new_known_factory_deps: Option<Vec<(H256, Vec<u8>)>>,
```
slowli, line 121: I think I understand the semantics of the field now: it includes dynamically deployed bytecodes (i.e., ones not mentioned in factory deps of a transaction). If this is the case, this raises further questions:

- Why is this field optional? If old VMs don't support dynamically deployed bytecodes, then I'd argue it would be caller's responsibility to not provide them with EVM simulator-enabled env.
- I'd argue that a `HashMap<H256, Vec<u8>>` would represent field semantics better. IIUC, bytecodes *can* be deduplicated by hash.

core/lib/config/src/configs/use_evm_simulator.rs
```diff
@@ -0,0 +1,7 @@
+use serde::Deserialize;
+
+/// Configure whether to enable the EVM simulator on the stack.
+#[derive(Debug, Deserialize, Clone, PartialEq)]
+pub struct UseEvmSimulator {
```
slowli, line 5: AFAIU, this config is never read properly in the main / external node initialization logic. Is it here for future use? I understand it's used as a surrogate for dynamically switching stuff in the codebase, but this is an anti-pattern and if it stays, it can be replaced with `std::env::var(_)` to not obscure the fact that it's a temporary hack.

slowli (changes requested):
Here's [a draft](https://github.com/matter-labs/zksync-era/compare/evm-simulator-base-sc...aov-evm-simulator-base-sc) making `evm_simulator` fields optional. It is somewhat incomplete e.g. in regards to transaction conversion in the API server.

- The easier option here would be to unconditionally allow `to == None` (essentially, it means reverting changes related to transaction conversions), but I'm not sure what the impact on UX would be.
- An alternative would require defining how `BaseSystemContracts.evm_simulator` is loaded. (Naturally, `to == None` is allowed iff the loaded `BaseSystemContracts` include the EVM simulator; AFAICT, *all* tx conversions in the API server are indeed tied to VM execution.) AFAIU, in the API the loaded `BaseSystemContracts` [are solely dictated](https://github.com/matter-labs/zksync-era/blob/df6876221936a44fa2fb8c80c01d043d229621fc/core/lib/vm_executor/src/oneshot/contracts.rs) by the protocol version; would this still be the case for the EVM simulator? Or would it be initially controlled by the node configuration? It would still only be enabled for the latest protocol version, I suppose – it would be weird to enable the simulator retroactively.

deniallugo-ml:
Closing this PR as the experimental work it belongs to has either moved to a dedicated repository or been deprioritized in favour of a different approach. Keeping it open here creates confusion about the current state of this feature.

If this work is continuing, it will be tracked in the appropriate repo or branch.

PR #2970, opened 2024-09-26
test: Base token price client, fix: ratio calculation multiplicative inverse error

cytadela8:
## What ❔

- I introduce tests to external_price_api crate. Some work was taken from https://github.com/matter-labs/zksync-era/pull/2315
- fix multiplicative inverse ratio calculation bug
- fix missing checks in get_fraction

## Why ❔

- We want to make sure the flow of data from API response to token ratio is correct

## Checklist


core/lib/external_price_api/src/coingecko_api.rs
```diff
@@ -110,3 +113,174 @@ impl CoinGeckoPriceResponse {
...
+                then.status(200).body("{}");
+            };
+        });
+    }
+
+    fn get_config(base_url: String, api_key: Option<String>) -> ExternalPriceApiClientConfig {
+        ExternalPriceApiClientConfig {
+            base_url: Some(base_url),
+            api_key,
+            source: "FILLER".to_string(),
```
encody, line 179: Should this be `"coingecko"`?

core/lib/external_price_api/src/coingecko_api.rs
```diff
@@ -110,3 +113,174 @@ impl CoinGeckoPriceResponse {
...
+        address: Address,
+        base_token_price: f64,
+    ) -> SetupResult {
+        happy_day_setup(None, server, address, base_token_price)
+    }
+
+    #[tokio::test]
+    async fn test_happy_day_with_no_api_key() {
+        happy_day_test(happy_day_setup_no_key).await
+    }
```
encody, line 234: Could be an inline lambda fn.

core/lib/external_price_api/src/tests.rs
```diff
@@ -0,0 +1,52 @@
...
+pub(crate) struct SetupResult {
+    pub(crate) client: Box<dyn PriceAPIClient>,
+}
+
+pub(crate) type SetupFn =
+    fn(server: &MockServer, address: Address, base_token_price: f64) -> SetupResult;
+
+pub(crate) async fn happy_day_test(setup: SetupFn) {
+    let server = MockServer::start();
+    let address_str = "0x1f9840a85d5af5bf1d1762f925bdaddc4201f984"; //Uniswap (UNI)
```
encody, line 20: It's probably worth declaring this as a constant.

core/lib/external_price_api/src/utils.rs
```diff
@@ -13,3 +13,33 @@ pub fn get_fraction(ratio_f64: f64) -> (NonZeroU64, NonZeroU64) {
...
+        assert_get_fraction_value(1337.0, 1337, 1);
+        assert_get_fraction_value(0.1, 1, 10);
+        assert_get_fraction_value(3.141, 3141, 1000);
+        assert_get_fraction_value(1_000_000.0, 1_000_000, 1);
+        assert_get_fraction_value(3123.47, 312347, 100);
+        // below tests assume some not necessarily required behaviour of get_fraction
+        assert_get_fraction_value(0.2, 1, 5);
+        assert_get_fraction_value(0.5, 1, 2);
+        assert_get_fraction_value(3.1415, 6283, 2000);
+    }
```
encody, line 44: Should include some `#[should_panic]` tests for zero num/den

cytadela8: We should never get panic from get_fraction as it converts from float to fraction so there should never be 0 denum, but I should test the value of zero and negative values also for completeness. Good catch!

core/lib/external_price_api/src/coingecko_api.rs
```diff
@@ -49,6 +49,7 @@ impl CoinGeckoPriceAPIClient {
         }
     }
 
+    /// returns ETH/BaseToken price of a token by address
```
ischasny, line 52: nit: It's base to eth, that's why you've had to flip it to be eth / base down in the file

core/lib/external_price_api/src/coingecko_api.rs
```diff
@@ -110,3 +113,174 @@ impl CoinGeckoPriceResponse {
...
+
+#[cfg(test)]
+mod test {
+    use httpmock::MockServer;
+    use zksync_config::configs::external_price_api_client::DEFAULT_TIMEOUT_MS;
+
+    use super::*;
+    use crate::tests::*;
+
+    fn auth_check(when: httpmock::When, api_key: Option<String>) -> httpmock::When {
```
ischasny, line 125: nit: can be simplified

```
api_key.map(|key| when.header(COINGECKO_AUTH_HEADER, key)).unwrap_or(when)
```

core/lib/external_price_api/src/tests.rs
```diff
@@ -0,0 +1,52 @@
...
+    let address = Address::from_str(address_str).unwrap();
+    let base_token_price = 0.00269; //ETH costs one token
+
+    let SetupResult { client } = setup(&server, address, base_token_price);
+    let api_price = client.fetch_ratio(address).await.unwrap();
+
+    let (num_in_eth, denom_in_eth) = get_fraction(base_token_price);
+    let (ratio_num, ratio_denom) = (denom_in_eth, num_in_eth);
+    assert!(((ratio_num.get() as f64) / (ratio_denom.get() as f64) - 371.74).abs() < 0.1);
```
ischasny, line 29: nit: worth adding a comment here why `371.74` or extract it to a variable near `base_token_price`. `0.1` can also be a var like `precision`

ischasny (approved):
Looks good - just a few small nits. 

General observation - originally I've split up the tests across multiple files so that CMC and CG can reuse the same setup functions while implementing their specifics in the files of their own. Do you think it's still useful to keep that separation, or should we consolidate all tests code in a single place?

core/lib/external_price_api/src/coingecko_api.rs
cytadela8: So in terms on numerator denominator it's (here) numerator=eth, denominator=base, and it's flipped below. Please if you have any idea, suggest how to comment this properly, so there is no confusion in the future.

core/lib/external_price_api/src/coingecko_api.rs
cytadela8: changed, does not matter as this is unsed as most of those fields.

cytadela8:
> General observation - originally I've split up the tests across multiple files so that CMC and CG can reuse the same setup functions while implementing their specifics in the files of their own. Do you think it's still useful to keep that separation, or should we consolidate all tests code in a single place?

@ischasny I would think that this may be still useful as tests.ts provides some framework to properly test each integration by implementing only the relevant mock and client setup in each test. If this doesn't work well with CMC we can get rid of this and consolidate. Hard to tell IMHO without implementing the CMC tests.

core/lib/external_price_api/src/utils.rs
cytadela8: Actually we are panicing sometimes! I believe we might not want to.

core/lib/external_price_api/src/coingecko_api.rs
ischasny: Yes, you are right, resolving

ischasny (dismissed):

encody (approved):

RomanBrodetski (approved):

deny.toml
```diff
@@ -8,6 +8,7 @@ feature-depth = 1
 
 [advisories]
 ignore = [
+    "RUSTSEC-2024-0375", # atty dependency being unmaintained, dependency of clap and criterion, we would need to update to newer major of dependencies
```
perekopskiy, line 11: nit: it's already in the list

perekopskiy (approved):

deny.toml
cytadela8: Ok, I'm not fixing this, as this PR blocks my colleague's work.

PR #2971, opened 2024-09-26
feat: Add CoinMarketCap external API

encody:
## What ❔

Adds support for CoinMarketCap as a price API for base token prices.

## Why ❔

Currently, the only supported API is CoinGecko. It's not ideal to only support a single 3rd party API here, so this provides some more degrees of freedom, and makes base token less reliant on a single third-party provider.

## Checklist

- [x] Tests for the changes have been added / updated. (See #2970)


core/lib/external_price_api/src/cmc_api.rs
```diff
@@ -0,0 +1,199 @@
...
+    data: Vec<CryptocurrencyObject>,
+}
+
+#[derive(Debug, Deserialize)]
+struct CryptocurrencyObject {
+    id: i32,
+    // name: String,
+    // symbol: String,
+    // slug: String,
+    // is_active: u8, // TODO: This field is available, should we at least emit a warning if the listing is not marked as active?
```
encody, line 149: Do we care to do "liveness" checks (i.e. could emit a warning if the API says that the token is inactive)?

ischasny (commented):
Looking good! Thank you for bringing CMC API over and brushing up wiring. While we are here - would it make sense to bring the tests over too? 
Also, is CMC client repeating the [error](https://github.com/matter-labs/zksync-era/pull/2970/commits/3b219d08f594efd1b9eac77786808a94a301d486) discovered by @cytadela8 (with flipped numerator / denominator)?

core/lib/external_price_api/src/cmc_api.rs
ischasny: That's a good point - lets absolutely do it!

encody:
> is CMC client repeating the [error](https://github.com/matter-labs/zksync-era/pull/2970/commits/3b219d08f594efd1b9eac77786808a94a301d486) discovered by @cytadela8 (with flipped numerator / denominator)?

Yes, thank you for the call-out. To avoid this sort of thing in the future, could we consider renaming `BaseTokenAPIRatio` to something like `BaseTokenApiPriceInEth`, or renaming the fields to that end?

encody:
@ischasny With regard to the tests, it might be better to merge this branch into @cytadela8 's to avoid duplicating test infra now and then de-duplicating again when they merge later.

encody:
Happy & failure tests added ✅

core/lib/external_price_api/src/cmc_api.rs
```diff
@@ -0,0 +1,358 @@
...
+    fn get(&self, path: &str) -> reqwest::RequestBuilder {
+        self.client
+            .get(self.base_url.join(path).expect("Failed to join URL path"))
+    }
+
+    async fn get_token_id(&self, address: Address) -> anyhow::Result<i32> {
+        if let Some(x) = self.cache_token_id_by_address.read().await.get(&address) {
+            return Ok(*x);
+        }
+        // drop read lock
```
ischasny, line 62: Is this comment needed?

ischasny (approved):
Left one small comment. 

Please verify that the new wiring logic works with different config settings. Otherwise looks good.

encody:
(merged with main because `zkt` was failing)

Deniallugo (approved):

PR #2973, opened 2024-09-26
fix(zk_toolbox): Do not panic if mint is not successful

Deniallugo:
## What ❔

Use governor for minting process and do not panic if minting is not successful

## Checklist

### DIFF
--- a/zk_toolbox/crates/zk_inception/src/commands/chain/common.rs
+++ b/zk_toolbox/crates/zk_inception/src/commands/chain/common.rs
@@ -111,7 +111,7 @@ pub async fn mint_base_token(
         let amount = AMOUNT_FOR_DISTRIBUTION_TO_WALLETS * base_token.nominator as u128
             / base_token.denominator as u128;
         common::ethereum::mint_token(
-            wallets.operator,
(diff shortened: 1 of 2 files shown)


zk_toolbox/crates/zk_inception/src/commands/chain/common.rs
```diff
@@ -111,7 +111,7 @@ pub async fn mint_base_token(
         let amount = AMOUNT_FOR_DISTRIBUTION_TO_WALLETS * base_token.nominator as u128
             / base_token.denominator as u128;
         common::ethereum::mint_token(
-            wallets.operator,
+            wallets.governor,
```
dutterbutter, line 114: Shouldnt this be `wallets.deployer` as `mint_token` is minting tokens to `wallets.governor` ?

dutterbutter: On second thought, it makes for sense for `governor` to be able to `mint` tokens rather then the `deployer`. Excuse my original comment.

matias-gonz (approved):

PR #2976, opened 2024-09-27
fix: sync layer compatibility

perekopskiy:
## What ❔

Sync layer code backward compatibility. Makes it possible to run server from this branch if chain was initialized with version < 25 

## Checklist


core/node/state_keeper/src/io/seal_logic/l2_block_seal_subtasks.rs
```diff
@@ -333,10 +333,9 @@ impl L2BlockSealSubtask for InsertTokensSubtask {
         connection: &mut Connection<'_, Core>,
     ) -> anyhow::Result<()> {
         let is_fictive = command.is_l2_block_fictive();
-        let mut deployer_address = command.l2_native_token_vault_proxy_addr;
-        if command.l2_legacy_shared_bridge_addr != H160::zero() {
-            deployer_address = command.l2_legacy_shared_bridge_addr;
-        }
+        let deployer_address = command
+            .l2_shared_bridge_addr
```
StanislavBreadless, line 337: This is not exactly correct... Since for chains without legacy bridge there is a shared bridge (the l2 asset router), but no NTV

perekopskiy: I see where the confusion comes from. In my understanding L2 asset router is not a shared bridge, it's L2 asset router. Shared bridge is only `L2SharedBridge`. IMO current state is good enough (operator doesn't need to do any config changes, `deployer_address` and `l2_shared_bridge` in API will return correct values automatically. `deployer_address` for EN is a bit ugly as I had to add another field in API response)

core/node/node_framework/src/implementations/layers/state_keeper/output_handler.rs
```diff
@@ -93,6 +89,38 @@ impl OutputHandlerLayer {
...
+                .blocks_dal()
+                .get_l2_block_header(l2_block)
+                .await
+                .context("failed to load L2 block header")?
+                .context("missing L2 block header")?;
+            let protocol_version = header
+                .protocol_version
+                .unwrap_or_else(ProtocolVersionId::last_potentially_undefined);
+
+            if protocol_version.is_pre_gateway() && self.l2_legacy_shared_bridge_addr.is_none() {
```
StanislavBreadless, line 115: maybe also do the contrary: if it is post gateway and we provided shared bridge, fail also?

core/node/state_keeper/src/io/seal_logic/mod.rs
```diff
@@ -339,7 +335,7 @@ impl L2BlockSealCommand {
     /// the bootloader enters the "tip" phase in which it can still generate events (e.g.,
     /// one for sending fees to the operator).
     ///
-    /// `l2_shared_bridge_addr` is required to extract the information on newly added tokens.
+    /// `l2_legacy_shared_bridge_addr` is required to extract the information on newly added tokens.
```
StanislavBreadless, line 338: I am not sure it is required (it is not present for the new chains for instance)

core/node/node_framework/src/implementations/layers/state_keeper/output_handler.rs
perekopskiy: IIUC it's not always correct. E.g. zk_toolbox currently init system with genesis_protocol_v = v25 and l2_legacy_shared_bridge_addr.is_some(), no?

core/node/state_keeper/src/io/seal_logic/mod.rs
perekopskiy: tbh this comment looks out of place completely (`l2_legacy_shared_bridge_addr` is not passed/used around), I will remove it

core/node/node_framework/src/implementations/layers/state_keeper/output_handler.rs
StanislavBreadless: ah damn, this is not correct for production, only correct for local development

StanislavBreadless (approved):

PR #2979, opened 2024-09-27
feat(vm): EVM emulator support – base

slowli:
## What ❔

Modifies the Era codebase to support the EVM emulator. Intentionally avoids changing the `contracts` submodule yet; as a consequence, there are no EVM emulation tests.

## Why ❔

Stepping stone for EVM equivalence.

## Checklist


core/node/eth_sender/src/eth_tx_aggregator.rs
```diff
@@ -268,12 +287,31 @@ impl EthTxAggregator {
...
+            };
+
             let base_system_contracts_hashes = BaseSystemContractsHashes {
                 bootloader,
                 default_aa,
+                evm_simulator,
             };
 
-            call_results_iterator.next().unwrap();
+            call_results_iterator.next().unwrap(); // FIXME: why is this value requested?
```
slowli, line 314: It looks like this value (`getVerifierParams` output) is always unused.

slowli (commented):
- It could be possible to simplify transaction conversions if `TransactionRequest.to` etc. is allowed to be unconditionally nullable. Not sure whether this will degrade UX.
- Not sure how upgrade logic should be handled. For now, I've [removed it](https://github.com/matter-labs/zksync-era/pull/2979/commits/d2bb42c0ec759e86d46964db364a98c25c5ad80d) because it was incorrect. It parsed the new upgrade format even when the old one was supplied from L1, which led to garbage stored in some `ProtocolUpgrade` fields (like protocol version). What's the adequate way to distinguish whether an upgrade contains an EVM simulator hash?

core/lib/contracts/src/lib.rs
```diff
@@ -339,14 +345,27 @@ impl BaseSystemContracts {
...
     }
-    // BaseSystemContracts with proved bootloader - for handling transactions.
+
+    /// BaseSystemContracts with proved bootloader - for handling transactions.
     pub fn load_from_disk() -> Self {
         let bootloader_bytecode = read_proved_batch_bootloader_bytecode();
         BaseSystemContracts::load_with_bootloader(bootloader_bytecode)
     }
 
+    /// Loads the latest EVM simulator for these base system contracts. Logically, it only makes sense to do for the latest protocol version.
```
0xVolosnikov, line 358: Will this comment be misleading if we have multiple versions with a EVM simulator in the future?

core/lib/dal/src/models/storage_transaction.rs
```diff
@@ -364,12 +364,12 @@ impl From<StorageTransactionReceipt> for TransactionReceipt {
             to: storage_receipt
                 .transfer_to
                 .or(storage_receipt.execute_contract_address)
-                .map(|addr| {
-                    serde_json::from_value::<Address>(addr)
+                .and_then(|addr| {
+                    serde_json::from_value::<Option<Address>>(addr)
                         .expect("invalid address value in the database")
+                    // For better compatibility with various clients, we never return null.
```
0xVolosnikov, line 370: nit: shouldn't it be closer to `.or_else(|| Some(Address::zero()))` line?

core/lib/multivm/src/versions/vm_latest/tests/migration.rs
0xVolosnikov, line 1: Why did we remove this?

core/lib/contracts/src/lib.rs
slowli: This comment will be updated (at least that's the idea). Once we have stable EVM simulators, they would be loaded in the new "historic" `BaseSystemContracts` constructors, and in the "default" constructor.

core/lib/multivm/src/versions/vm_latest/tests/migration.rs
slowli: That's a leftover from the previous PR; probably has to deal with updating system contracts. Will try to restore these tests and see if they work.

core/lib/l1_contract_interface/src/multicall3/mod.rs
```diff
@@ -7,6 +7,7 @@ use zksync_types::{
 };
 
 /// Multicall3 contract aggregate method input vector struct.
+#[derive(Debug)]
```
0xVolosnikov, line 10: It is done for tests?

slowli: Yes. Can the derived `Debug` representation contain confidential data?

core/lib/multivm/src/versions/vm_latest/types/internals/transaction_data.rs
```diff
@@ -62,6 +62,12 @@ impl From<Transaction> for TransactionData {
                     U256::zero()
                 };
 
+                let should_deploy_contract = if execute_tx.execute.contract_address.is_none() {
```
0xVolosnikov, line 65: Looks somewhat dangerous. Especially if we will allow to use `None` as `to` even without EVM simulator

core/lib/l1_contract_interface/src/multicall3/mod.rs
0xVolosnikov: Don't think so

core/lib/multivm/src/versions/vm_latest/types/internals/transaction_data.rs
slowli: Fixed by asserting that the EVM simulator is enabled before setting the field; PTAL.

core/node/api_server/src/web3/namespaces/debug.rs
```diff
@@ -259,7 +259,11 @@ impl DebugNamespace {
         };
 
         let call_overrides = request.get_call_overrides()?;
-        let call = L2Tx::from_request(request.into(), MAX_ENCODED_TX_SIZE)?;
+        let call = L2Tx::from_request(
+            request.into(),
+            MAX_ENCODED_TX_SIZE,
+            false, // Even with EVM simulation enabled, calls must specify `to` field
```
perekopskiy, line 265: hmm, why?

core/node/eth_sender/src/eth_tx_aggregator.rs
perekopskiy: leftover, `getVerifierParams` call can be removed

core/node/api_server/src/web3/namespaces/debug.rs
slowli: According to various Ethereum docs, like [Alchemy](https://docs.alchemy.com/reference/eth-call), and [the copied docs for `CallRequest`](https://github.com/matter-labs/zksync-era/blob/a5ac9c164bc0d16a879bef1cab1ccdaeac205bdc/core/lib/basic_types/src/web3/mod.rs#L783). Logically speaking, `to == None` corresponds to a deployment, and deployments don't work with the call execution mode in case of "pure" EraVM transactions (I've checked recently when [adding more unit tests for API server](https://github.com/matter-labs/zksync-era/pull/3001); e.g., load test transactions with non-zero deploys always revert in this mode). I'm not 100% sure, though; mechanically, it's easy to treat calls in the same way as transactions. @0xVolosnikov What's the correct choice here?

core/lib/multivm/src/versions/vm_latest/tracers/evm_deploy_tracer.rs
```diff
@@ -0,0 +1,105 @@
...
+        // We check if ContractDeployer was called with provided evm bytecode.
+        // It is assumed that by that time the user has already paid for its size.
+        // So even if we do not revert the addition of the this bytecode it is not a ddos vector, since
+        // the payment is the same as if the bytecode publication was reverted.
+        let current_callstack = &state.local_state.callstack.current;
+
+        // Here we assume that the only case when PC is 0 at the start of the execution of the contract.
+        let known_code_storage_call = current_callstack.this_address == KNOWN_CODES_STORAGE_ADDRESS
+            && current_callstack.pc == 0
+            && current_callstack.msg_sender == CONTRACT_DEPLOYER_ADDRESS;
```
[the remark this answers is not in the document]
slowli: Did this, please check whether the approach is correct.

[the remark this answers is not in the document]
perekopskiy: Looks correct. One unrelated thing that I am not sure about is `read_pointer`: it reads region `[start+offset..start+length]`; however in call tracer it doesn't use offset:
```
let packed_abi =
    state.vm_local_state.registers[CALL_IMPLICIT_CALLDATA_FAT_PTR_REGISTER as usize];
assert!(packed_abi.is_pointer);
let far_call_abi = FarCallABI::from_u256(packed_abi.value);
memory.read_unaligned_bytes(
    far_call_abi.memory_quasi_fat_pointer.memory_page as usize,
    far_call_abi.memory_quasi_fat_pointer.start as usize,
    far_call_abi.memory_quasi_fat_pointer.length as usize,
)
```
so it looks like a bug in call_tracer or maybe offset = 0 for some reason. Do you know something about it?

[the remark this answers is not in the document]
slowli: Will need to investigate some more, but it could be the case that `offset == 0` for fat pointers to calldata / returndata. E.g., the new VM [narrows the pointer](https://github.com/matter-labs/vm2/blob/a233d44bbe61dc6a758a754c3b78fe4f83e56699/crates/vm2/src/instruction_handlers/far_call.rs#L204) (so that `offset == 0`) in these situations.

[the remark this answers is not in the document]
slowli: Looks like the old VM implementation maintains the same invariant: all [calldata](https://github.com/matter-labs/zksync-protocol/blob/a3369809a760c448542d6e6877b95bbabac25d14/crates/zk_evm/src/opcodes/execution/far_call.rs#L328-L335) / [returndata](https://github.com/matter-labs/zksync-protocol/blob/a3369809a760c448542d6e6877b95bbabac25d14/crates/zk_evm/src/opcodes/execution/ret.rs#L86-L93) fat pointers are expected to have `offset == 0`.

perekopskiy (approved):

core/node/api_server/src/web3/namespaces/debug.rs
slowli: Decided to keep `to` mandatory.

PR #2980, opened 2024-09-27
feat(proof-data-handler): exclude batches without object file in GCS 

pbeza:
## What ❔

`/tee/proof_inputs` endpoint no longer returns batches that have no corresponding object file in Google Cloud Storage for an extended period.

## Why ❔

TEE's `proof-data-handler` on `mainnet` was flooded with warnings.

Since the recent `mainnet`'s `24.25.0` redeployment, we've been [flooded with warnings][warnings] for the `proof-data-handler` on `mainnet` (the warnings are actually _not_ fatal in this context):
```
Failed request with a fatal error

(...)

Blobs for batch numbers 490520 to 490555 not found in the object store. Marked as unpicked.
```
The issue is caused [by the code][code] behind the `/tee/proof_inputs` [endpoint][endpoint_proof_inputs] (which is equivalent to the `/proof_generation_data` [endpoint][endpoint_proof_generation_data]) – it finds the next batch to send to the [requesting][requesting] `tee-prover` by looking for the first batch that has a corresponding object in the Google object store. As it skips over batches that don’t have the objects, [it logs][logging] `Failed request with a fatal error` for each one (unless the skipped batch was successfully proven, in which case it doesn’t log the error). This happens with every request the `tee-prover` sends, which is why we're getting so much noise in the logs.

One possible solution is to flag the problematic batches as `permanently_ignored`, like [Thomas did before][Thomas] on `mainnet`.

[warnings]: https://grafana.matterlabs.dev/goto/TjlaXQgHg?orgId=1
[code]: https://github.com/matter-labs/zksync-era/blob/3f406c7d0c0e76d798c2d838abde57ca692822c0/core/node/proof_data_handler/src/tee_request_processor.rs#L35-L79
[endpoint_proof_inputs]: https://github.com/matter-labs/zksync-era/blob/3f406c7d0c0e76d798c2d838abde57ca692822c0/core/node/proof_data_handler/src/lib.rs#L96
[endpoint_proof_generation_data]: https://github.com/matter-labs/zksync-era/blob/3f406c7d0c0e76d798c2d838abde57ca692822c0/core/node/proof_data_handler/src/lib.rs#L67
[requesting]: https://github.com/matter-labs/zksync-era/blob/3f406c7d0c0e76d798c2d838abde57ca692822c0/core/bin/zksync_tee_prover/src/tee_prover.rs#L93
[logging]: https://github.com/matter-labs/zksync-era/blob/3f406c7d0c0e76d798c2d838abde57ca692822c0/core/lib/object_store/src/retries.rs#L56
[Thomas]: https://matter-labs-workspace.slack.com/archives/C05ANUCGCKV/p1725284962312929

## Checklist


pbeza:
@popzxc, I remember you mentioned not to ask for code reviews this wave, but you're probably the most familiar with this code (along with @slowli). So, if you could make an exception this time, I’d really appreciate it. If you're busy, no worries – feel free to ignore, and I’ll ask @RomanBrodetski to find someone else. Thanks!

pbeza:
Kindly ping @slowli @RomanBrodetski. I need a reviewer.

core/lib/object_store/src/retries.rs
```diff
@@ -28,6 +28,43 @@ impl Request<'_> {
...
+    {
+        self.retry_internal(max_retries, f).await
+    }
+
+    #[tracing::instrument(
+        name = "object_store::Request::retry_optional",
+        skip(f), // output request and store as a part of structured logs
+        fields(retries) // Will be recorded before returning from the function
+    )]
+    async fn retry_optional<T, Fut, F>(
```
slowli, line 45: AFAICT, this function is unused. It's not detected as such because of the `instrument` attribute.

core/lib/types/src/tee_types.rs
slowli, line 1: Dumb question: Why was this module moved to `zksync_types`? It has a lot of deps which are not usually necessary.

core/lib/types/src/tee_types.rs
```diff
@@ -17,6 +19,17 @@ impl fmt::Display for TeeType {
     }
 }
 
+/// Representation of a locked batch. Used in DAL to fetch details about the locked batch to
```
slowli, line 22: If this belongs to the DAL domain, it makes sense to define this type in DAL. AFAICT, there are some TEE-related types defined there already (like `TeeProofGenerationJobStatus`).

core/node/proof_data_handler/src/tee_request_processor.rs
```diff
@@ -42,34 +48,50 @@ impl TeeRequestProcessor {
...
+                Ok(None) => {
                     missing_range = match missing_range {
-                        Some((start, _)) => Some((start, l1_batch_number)),
-                        None => Some((l1_batch_number, l1_batch_number)),
+                        Some((start, _)) => Some((start, batch_number)),
+                        None => Some((batch_number, batch_number)),
+                    };
+                    let datetime_utc = Utc.from_utc_datetime(&locked_batch.created_at);
+                    let duration = Utc::now().signed_duration_since(datetime_utc);
+                    let status = if duration > ChronoDuration::days(10) {
```
slowli, line 71: Nit: It may make sense to extract this value into a constant so that it's more visible.

core/node/proof_data_handler/src/tee_request_processor.rs
```diff
@@ -42,34 +48,50 @@ impl TeeRequestProcessor {
...
+                        TeeProofGenerationJobStatus::PermanentlyIgnored
```
slowli, line 72: Nit: Maybe it makes sense to log this status assignment, so that it is easier to debug?

core/node/proof_data_handler/src/tee_request_processor.rs
```diff
@@ -42,34 +48,50 @@ impl TeeRequestProcessor {
...
+                    break Ok(Json(TeeProofGenerationDataResponse(Some(Box::new(input)))))
+                }
+                Ok(None) => {
                     missing_range = match missing_range {
-                        Some((start, _)) => Some((start, l1_batch_number)),
-                        None => Some((l1_batch_number, l1_batch_number)),
+                        Some((start, _)) => Some((start, batch_number)),
+                        None => Some((batch_number, batch_number)),
+                    };
+                    let datetime_utc = Utc.from_utc_datetime(&locked_batch.created_at);
```
slowli, line 69: It may make sense to store the timestamp as `DateTime<Utc>` in `LockedBatch`.

RomanBrodetski (commented):
@pbeza to be honest I don't fully follow this solution. I understand what we are trying to do (mark older unresolved jobs as skipped), but I'm not sure I understand the `Why` here. We can discuss over a huddle or async

pbeza:
JFYI: This PR is on hold because the code it is based on was recently radically redesigned/refactored here: https://github.com/matter-labs/zksync-era/pull/3017. This PR may be cherry-picked once https://github.com/matter-labs/zksync-era/pull/3017 is merged into `main`.

core/lib/object_store/src/retries.rs
pbeza: Removed. ✅

core/lib/types/src/tee_types.rs
pbeza: Reverted. ✅

core/lib/types/src/tee_types.rs
pbeza: Moved. ✅

core/node/proof_data_handler/src/tee_request_processor.rs
pbeza: Extracted. ✅

core/node/proof_data_handler/src/tee_request_processor.rs
pbeza: Fixed. ✅

core/node/proof_data_handler/src/tee_request_processor.rs
pbeza: Good point. Added. ✅

core/lib/dal/migrations/20240930110000_tee_add_permanently_ignored_state.down.sql
```diff
@@ -0,0 +1 @@
+UPDATE tee_proof_generation_details SET status = 'permanently_ignore' WHERE status = 'permanently_ignored';
```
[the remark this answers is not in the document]
pbeza: Removed it as suggested. ✅

core/lib/types/src/tee_types.rs
```diff
@@ -17,6 +19,17 @@ impl fmt::Display for TeeType {
...
+/// determine whether it should be flagged as permanently ignored if it has no corresponding file in
+/// the object store for an extended period.
```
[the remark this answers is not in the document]
pbeza: Elaborated [here](https://github.com/matter-labs/zksync-era/pull/2980/files#diff-46d00b49fb68d6da13c9ecdc5ab3dab6adf502ec216349721586e96553652444R36-R42). ✅

pbeza:
I rebased this PR on the latest `origin/main`, added a few more fixes, and did a careful manual retest. Trust me – it works. Please take another look, @slowli!

BTW, sorry for the force-push instead of merging `origin/main` into this branch.

pbeza:
@slowli, kindly `ping` for a code review.

core/node/proof_data_handler/src/tee_request_processor.rs
```diff
@@ -47,49 +51,52 @@ impl TeeRequestProcessor {
     ) -> Result<Option<Json<TeeProofGenerationDataResponse>>, RequestProcessorError> {
         tracing::info!("Received request for proof generation data: {:?}", request);
 
-        let mut min_batch_number = self.config.tee_config.first_tee_processed_batch;
-        let mut missing_range: Option<(L1BatchNumber, L1BatchNumber)> = None;
+        let batch_ignored_timeout = ChronoDuration::days(10);
```
haraldh, line 54: Hardcode or config parameter?

core/node/proof_data_handler/src/tee_request_processor.rs
```diff
@@ -47,49 +51,52 @@ impl TeeRequestProcessor {
...
-        let result = loop {
-            let Some(l1_batch_number) = self
+        loop {
+            let Some(locked_batch) = self
                 .lock_batch_for_proving(request.tee_type, min_batch_number)
                 .await?
             else {
-                // No job available
-                return Ok(None);
+                return Ok(None); // no job available
```
haraldh, line 62: nit: this can now be `break`, too... Either change all `break` to `return` or the other way round

core/lib/object_store/src/retries.rs
```diff
@@ -28,6 +28,18 @@ impl Request<'_> {
...
+    where
+        Fut: Future<Output = Result<T, ObjectStoreError>>,
+        F: FnMut() -> Fut,
+    {
+        self.retry_internal(max_retries, f).await
+    }
+
+    async fn retry_internal<T, Fut, F>(
+        &self,
+        max_retries: u16,
```
pbeza, line 42: JFYI: this is an artifact that I'm gonna revert.

core/node/proof_data_handler/Cargo.toml
```diff
@@ -24,12 +24,12 @@ axum.workspace = true
...
 
 [dev-dependencies]
 hyper.workspace = true
-chrono.workspace = true
 zksync_multivm.workspace = true
 serde_json.workspace = true
 tower.workspace = true
-zksync_basic_types.workspace = true
 zksync_contracts.workspace = true
+zksync_basic_types.workspace = true
```
slowli, line 35: Nit: You don't usually need `zksync_basic_types` as a direct dep if you depend on `zksync_types`; the latter re-exports a substantial part of basic types.

core/node/proof_data_handler/src/tee_request_processor.rs
```diff
@@ -47,49 +51,52 @@ impl TeeRequestProcessor {
...
-                        None => Some((l1_batch_number, l1_batch_number)),
+                    let duration = Utc::now().signed_duration_since(locked_batch.created_at);
+                    let status = if duration > batch_ignored_timeout {
+                        TeeProofGenerationJobStatus::PermanentlyIgnored
+                    } else {
+                        TeeProofGenerationJobStatus::Failed
                     };
-                    self.unlock_batch(l1_batch_number, request.tee_type).await?;
-                    min_batch_number = l1_batch_number + 1;
+                    self.unlock_batch(batch_number, request.tee_type, status)
```
slowli, line 80: I feel we've already had this conversation: This looks like backend driven by frontend anti-pattern; batches are only unlocked in response to client requests. I'd imagine that batches should be unlocked on a timeout (currently hard-coded as 10 days) with `PermanentlyIgnored` status, right? Or is there no harm if a batch is unlocked later?

core/node/proof_data_handler/src/tee_request_processor.rs
```diff
@@ -47,49 +51,52 @@ impl TeeRequestProcessor {
...
+                    tracing::warn!(
+                        "Assigned status {} to batch {} created at {}",
+                        status,
+                        batch_number,
+                        locked_batch.created_at
+                    );
                 }
                 Err(err) => {
-                    self.unlock_batch(l1_batch_number, request.tee_type).await?;
+                    self.unlock_batch(
```
slowli, line 90: Dumb question: Why does this unlocking work like that? Suppose this server loses connection to Postgres for a moment, resulting in a `RequestProcessorError::Dal` error. IIUC, the batch will be marked as unlocked here, but there's seemingly no reason to do so.

core/node/proof_data_handler/src/tee_request_processor.rs
haraldh: There is also a `unlock` part in the SQL query for new jobs ... 
```
OR ( tee.status = `picked_by_prover` AND tee.prover_taken_at < NOW() - processing_timeout::INTERVAL )
```

core/node/proof_data_handler/src/tee_request_processor.rs
pbeza: Notice that the batch that failed due to `RequestProcessorError::Dal` or `RequestProcessorError::GeneralError` is guaranteed to be reconsidered by the SQL query that fetches the next batch to be TEE-proven:

https://github.com/matter-labs/zksync-era/blob/9188f8a3107f9a46d5b74c99efe7a1db5ccda56d/core/lib/dal/src/tee_proof_generation_dal.rs#L83-L84

Would you prefer me to implement explicit retry logic in case of `RequestProcessorError::Dal` errors?

core/node/proof_data_handler/src/tee_request_processor.rs
pbeza: > (...) Or is there no harm if a batch is unlocked later?

There is no harm I can think of. Also, please consider the point @haraldh mentioned above (and that I elaborated on further [here](https://github.com/matter-labs/zksync-era/pull/2980/files/9188f8a3107f9a46d5b74c99efe7a1db5ccda56d#r1847009731)): batch status can transition from `failed` to `picked_by_prover`, and this transition is also driven by frontend (since [`lock_batch_for_proving`](https://github.com/matter-labs/zksync-era/blob/9188f8a3107f9a46d5b74c99efe7a1db5ccda56d/core/lib/dal/src/tee_proof_generation_dal.rs#L53) is ultimately triggered by a user request).

core/lib/object_store/src/retries.rs
pbeza: Reverted [here](https://github.com/matter-labs/zksync-era/pull/2980/commits/a559c401a66b170c4ae2c15e205306a448c28e21). ✅

core/node/proof_data_handler/Cargo.toml
pbeza: Fixed [here](https://github.com/matter-labs/zksync-era/pull/2980/commits/16a0e555a1cf746bcbf2e62dac00122132d4ea0e). ✅

core/node/proof_data_handler/src/tee_request_processor.rs
pbeza: Fixed [here](https://github.com/matter-labs/zksync-era/pull/2980/commits/5de035be5d8577a07a8685ee7dd1a0ad506d9b99). ✅

core/node/proof_data_handler/src/tee_request_processor.rs
pbeza: I can make it a configurable parameter indeed.

I hardcoded it because I felt it might be confusing in a scenario where, for example, the `batch_ignored_timeout` config parameter is set to `10 days` and later changed to `1 year`. In such a case, it might be unclear at first glance why some entries in the database are marked as `permanently_ignored` even though they are only 10 days old and not at least one year.

core/node/proof_data_handler/src/tee_request_processor.rs
slowli: I see your point; I've misunderstood what `Failed` status means (I thought it means proof generation itself fails, but AFAIU it's also used for failures related to preparing inputs for proof generation). Regarding the query: Do I read it correctly that it uses the same timeout strategy that is used for force-unlocking batches? If so, why not use a smaller timeout or no timeout at all? Going with my previous example, IIUC, a transient network connectivity issue may lead to a batch getting ignored for 10 minutes (with the default config); this sounds inefficient.

core/node/proof_data_handler/src/tee_request_processor.rs
slowli: > batch status can transition from failed to picked_by_prover,

This one is expected, because picking batches is *expected* to be driven by the client requesting proof inputs (unless I'm horribly misunderstanding the entire workflow). It is unlocking that looks suboptimal because *conceptually*, rows in `tee_proof_generation_details` should be unlocked w/ `PermanentlyIgnored` state on a certain timeout after their creation (of course only if they don't reach another terminal state before that). If this is indeed the ideal workflow, it can be implemented as a background task rather than being driven by the frontend. Can be implemented separately, of course.

pbeza: > (...) If this is indeed the ideal workflow, it can be implemented as a background task rather than being driven by the frontend. (...)

Do we already have a similar background task implemented somewhere? I would rather not reinvent the wheel if we have some building blocks for creating this kind of background service that I can reuse.

> (...) Can be implemented separately, of course.

Alright, I’ll probably handle it in a separate PR then.

core/node/proof_data_handler/src/tee_request_processor.rs
pbeza: Fixed [here](https://github.com/matter-labs/zksync-era/pull/2980/commits/0dd01210668c1fb25defa0cbe9723fc5ff3b1abd#diff-9cf7f2a4779fdd713eabf6a97c007aa4114de190846ee8c5e13be555d14c49e3R54-R57). ✅

core/node/proof_data_handler/src/tee_request_processor.rs
pbeza: > (...) I've misunderstood what Failed status means (I thought it means proof generation itself fails, but AFAIU it's also used for failures related to preparing inputs for proof generation). (...)

I just refined the docs a bit to clear it up:
https://github.com/matter-labs/zksync-era/blob/0dd01210668c1fb25defa0cbe9723fc5ff3b1abd/core/lib/dal/src/tee_proof_generation_dal.rs#L32-L36

> (...) Regarding the query: Do I read it correctly that it uses the same timeout strategy that is used for force-unlocking batches? (...)

So, we are talking about two different timeouts here – I refined the docs for them as well:
https://github.com/matter-labs/zksync-era/blob/0dd01210668c1fb25defa0cbe9723fc5ff3b1abd/core/lib/config/src/configs/proof_data_handler.rs#L12-L17

> (...) If so, why not use a smaller timeout or no timeout at all? Going with my previous example, IIUC, a transient network connectivity issue may lead to a batch getting ignored for 10 minutes (with the default config); this sounds inefficient. (...)

The [timeout used in the SQL query](https://github.com/matter-labs/zksync-era/blob/9188f8a3107f9a46d5b74c99efe7a1db5ccda56d/core/lib/dal/src/tee_proof_generation_dal.rs#L84) is `tee_proof_generation_timeout_in_secs`, and if we set it to `0`, it means we keep retrying all [`Failed`](https://github.com/matter-labs/zksync-era/blob/0dd01210668c1fb25defa0cbe9723fc5ff3b1abd/core/lib/dal/src/tee_proof_generation_dal.rs#L36) batches immediately,  before proceeding with any other batches. If we have a batch where preparing inputs for proof generation always fails (for any reason), it would effectively block TEE proving progress. This happens because we would keep retrying until either the input preparation for that batch succeeds or the `tee_batch_permanently_ignored_timeout_in_hours` timeout expires.

slowli: What I mean by "force-unlocking" is the transition between `picked_by_prover` to "unpicked" (which would be transformed to `picked_by_prover` again; it's not a real state). So, the question is whether it makes sense to make the timeout for this operation the same as the timeout for `failed` batches. IDK; intuitively, at least for some types of errors, the default timeout value of 10 minutes looks too large, but if a typical error takes this long to resolve, happy to roll with it.

pbeza: Gotcha. Decreased the retry interval for failed batches from 10 minutes to 1 minute [here](https://github.com/matter-labs/zksync-era/pull/2980/commits/ada62febcbf6c01e21ebdb0242823c4feb1303f5). ✅ Hopefully, that works better now.

pbeza:
@slowli I just retested everything manually to ensure that transitioning to the `permanently_ignored` state works correctly when there are no object files (you can't imagine how time-consuming and painful it is to reproduce). Everything seems to be working perfectly. Ready to merge from my perspective. 🫡

slowli (approved):

PR #2984, opened 2024-09-30
fix(eth-client): fix l2_fee_history

perekopskiy:
## What ❔

There was a wrong assertion: `if fee_history.inner.base_fee_per_gas.len() != fee_history.l2_pubdata_price.len()`. Instead `base_fee_per_gas.len()` should equal `l2_pubdata_price.len() + 1`.

Refactors code of l*_fee_history methods: 
- adds exhaustive assertions
- returns errors instead of only logging

## Checklist

### DIFF
--- a/core/lib/eth_client/src/clients/http/query.rs
+++ b/core/lib/eth_client/src/clients/http/query.rs
@@ -15,7 +15,7 @@ use crate::{
     BaseFees, EthFeeInterface, EthInterface, RawTransactionBytes,
 };
 
-const FEE_HISTORY_MAX_REQUEST_CHUNK: usize = 1024;
+const FEE_HISTORY_MAX_REQUEST_CHUNK: usize = 1023;
 
 #[async_trait]
 impl<T> EthInterface for T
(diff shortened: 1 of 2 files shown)


core/lib/eth_client/src/clients/mock.rs
```diff
@@ -415,25 +415,33 @@ fn l2_eth_fee_history(
     let from_block = from_block.as_usize();
     let start_block = from_block.saturating_sub(block_count.as_usize() - 1);
 
+    let base_fee_per_gas = base_fee_history[start_block..=from_block]
+        .iter()
+        .chain([&base_fee_history[from_block]]) // duplicate last value
```
StanislavBreadless, line 420: better to add a comment on why we do it

core/lib/eth_client/src/clients/http/query.rs
```diff
@@ -407,19 +433,45 @@ where
...
+            return Err(EnrichedClientError::custom(message, "l2_fee_history")
+                .with_arg("chunk_size", &chunk_size)
+                .with_arg("chunk_end", &chunk_end));
         }
 
         for (base, l2_pubdata_price) in fee_history
             .inner
             .base_fee_per_gas
             .into_iter()
+            .take(chunk_size)
```
StanislavBreadless, line 474: better to document why do we do it here

core/lib/eth_client/src/clients/http/query.rs
```diff
@@ -324,21 +324,47 @@ where
...
             );
+            return Err(EnrichedClientError::custom(message, "l1_fee_history")
+                .with_arg("chunk_size", &chunk_size)
+                .with_arg("chunk_end", &chunk_end));
         }
 
         for (base, blob) in fee_history
             .base_fee_per_gas
             .into_iter()
+            .take(chunk_size)
```
StanislavBreadless, line 367: better to write a comment on why do we need it here

StanislavBreadless (approved):
The logic is correct, I would suggest to add a bit more comments for future readers without context

PR #2986, opened 2024-09-30
feat(configs): Add port parameter to ConsensusConfig

matias-gonz:
## What ❔
Update consensus config:
* Add `port` parameter

## Why ❔

Make it consistent with other config structs.


zk_toolbox/crates/zk_inception/src/consts.rs
```diff
@@ -12,27 +10,6 @@ pub const DEFAULT_UNSIGNED_TRANSACTIONS_DIR: &str = "transactions";
...
-/// Max payload size for consensus in bytes
-pub const MAX_PAYLOAD_SIZE: usize = 2_500_000;
-/// Max batch size for consensus in bytes
-/// Compute a default batch size, so operators are not caught out by the missing setting
-/// while we're still working on batch syncing. The batch interval is ~1 minute,
-/// so there will be ~60 blocks, and an Ethereum Merkle proof is ~1kB, but under high
-/// traffic there can be thousands of huge transactions that quickly fill up blocks
-/// and there could be more blocks in a batch then expected. We chose a generous
-/// limit so as not to prevent any legitimate batch from being transmitted.
-pub const MAX_BATCH_SIZE: usize = MAX_PAYLOAD_SIZE * 5000 + kB;
```
matias-gonz, line 27: We can remove because of:
https://github.com/matter-labs/zksync-era/blob/eed8198f97fd0606884271c86eebd700785ccbd8/core/lib/protobuf_config/src/consensus.rs#L139

core/lib/config/src/configs/consensus.rs
```diff
@@ -115,12 +115,13 @@ impl RpcConfig {
...
 pub struct ConsensusConfig {
+    pub port: u16,
     /// Local socket address to listen for the incoming connections.
-    pub server_addr: std::net::SocketAddr,
+    pub server_url: std::net::SocketAddr,
     /// Public address of this node (should forward to `server_addr`)
     /// that will be advertised to peers, so that they can connect to this
     /// node.
-    pub public_addr: Host,
+    pub public_url: Host,
```
pompon0, line 124: these are not URLs

core/lib/config/src/configs/consensus.rs
```diff
@@ -115,6 +115,7 @@ impl RpcConfig {
 /// Config (shared between main node and external node).
 #[derive(Clone, Debug, PartialEq)]
 pub struct ConsensusConfig {
+    pub port: u16,
```
pompon0, line 118: what is this port? Is it related to server_addr/public_addr? Why would you want to duplicate this information? Note that there is no requirement for the server_addr port and the public_addr port to be the same.

matias-gonz: The information is duplicated in some other configs:
https://github.com/matter-labs/zksync-era/blob/main/etc/env/file_based/general.yaml#L30-L31
https://github.com/matter-labs/zksync-era/blob/main/etc/env/file_based/general.yaml#L32-L33
https://github.com/matter-labs/zksync-era/blob/main/etc/env/file_based/general.yaml#L84-L85

We want the configs to be consistent so we can automate the task to allocate ports and avoid overlaps when creating new chains/external node
If at some point we add a new port/url/address, this port allocation will continue working and nothing will break

pompon0 (approved):

Deniallugo (dismissed):

Deniallugo (approved):

PR #2992, opened 2024-10-01
feat(zk_toolbox): added support for setting attester committee defined in a separate file

pompon0:
So far only setting committee to the value in genesis was supported which is too constraining for real world use. I've added support for specifying the attester committee in a separate yaml file and passing it to the set-attester-committee command.


brunoffranca (dismissed):
LGTM. You might want to wait for @Deniallugo to have a quick look.

brunoffranca (approved):

onyxet (approved):

PR #2994, opened 2024-10-02
fix(vm): Prepare new VM for use in API server and fix divergences

slowli:
## What ❔

Prepares the new VM for use in the API server (without tracers):

- Fills in more statistics
- Covers more statistic fields in divergence checks and fixes the corresponding divergence in `pubdata_published`
- Fixes a divergence if revert reaches the bootloader call frame (happens in the call execution mode)
- Fixes a panic in `ShadowVm` if the transaction hash is not set and makes `impl Debug for Transaction` non-panicking. (A hash is not set if a transaction is converted from a call.)

## Why ❔

Part of preparations for integrating the new VM into the API server.

## Checklist


slowli:
Not sure where performance improvements are coming from (yet), but they look legitimate in the sense I haven't commented parts of the new VM execution etc.

core/lib/vm_interface/src/vm.rs
```diff
@@ -44,9 +44,6 @@ pub trait VmInterface {
         with_compression: bool,
     ) -> (BytecodeCompressionResult<'_>, VmExecutionResultAndLogs);
 
-    /// Record VM memory metrics.
-    fn record_vm_memory_metrics(&self) -> VmMemoryMetrics;
```
slowli, line 48: AFAIU, metrics are specific to the old tracer-powered VM implementation, and these metrics are not critical for correctness (they are only used for metrics). I haven't found a way to fill them for the new VM.

core/lib/multivm/src/versions/vm_fast/vm.rs
```diff
@@ -744,20 +782,23 @@ impl<S: ReadStorage, T: Tracer> World<S, T> {
...
-    fn cost_of_writing_storage(&mut self, initial_value: Option<U256>, new_value: U256) -> u32 {
-        let is_initial = initial_value.is_none();
-        let initial_value = initial_value.unwrap_or_default();
+    fn read_storage_value(&mut self, contract: H160, key: U256) -> U256 {
+        let key = &StorageKey::new(AccountTreeId::new(contract), u256_to_h256(key));
+        U256::from_big_endian(self.storage.read_value(key).as_bytes())
+    }
 
-        if initial_value == new_value {
+    fn cost_of_writing_storage(&mut self, initial_slot: StorageSlot, new_value: U256) -> u32 {
```
joonazan, line 800: Not sure if bug or poor naming. If initial_slot contains the initial value of the slot it doesn't make sense that it is a StorageSlot because it will always be initial according to the name.

core/lib/multivm/src/versions/vm_fast/vm.rs
```diff
@@ -744,20 +782,23 @@ impl<S: ReadStorage, T: Tracer> World<S, T> {
...
         let key = &StorageKey::new(AccountTreeId::new(contract), u256_to_h256(key));
-        if self.storage.is_write_initial(key) {
-            None
-        } else {
-            Some(self.storage.read_value(key).as_bytes().into())
+        let value = U256::from_big_endian(self.storage.read_value(key).as_bytes());
+        let is_write_initial = self.storage.is_write_initial(key);
+        StorageSlot {
+            value,
+            is_write_initial,
```
joonazan, line 791: Maybe `has_been_written_in_this_block` or whatever descriptive name or comment would help the confusion around this.

joonazan:
Performance improvements can be related to storage changes, though it looks like there are still two separate hash map accesses to get storage slot and initialness, so I'm not sure.

core/lib/multivm/src/versions/vm_fast/vm.rs
slowli: Bad naming; "initial" here means that the provided value is the initial value. I'll shorten the name to `slot`.

core/lib/multivm/src/versions/vm_fast/vm.rs
slowli: "is write initial" / "is initial" is used extensively throughout the codebase, so I don't think changing naming is really an option without changing those occurrences as well. I'll add a comment.

core/lib/multivm/src/versions/vm_fast/tests/rollbacks.rs
```diff
@@ -142,3 +145,33 @@ fn test_vm_loadnext_rollbacks() {
...
+    let (compression_result, vm_result) = vm
+        .vm
+        .execute_transaction_with_bytecode_compression(tx, true);
+    compression_result.unwrap();
+    assert_matches!(
+        vm_result.result,
+        ExecutionResult::Revert { output }
+            if output.to_string().contains("This method always reverts")
+    );
+    // The new VM performs deduplication internally and doesn't record reads, hence this crude comparison.
```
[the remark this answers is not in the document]
slowli: Just removed a comment to avoid confusion. You can see a non-crude version in the `vm_latest` counterpart; it checks that there are no deduplicated write logs.

joonazan (approved):

PR #3006, opened 2024-10-03
feat: gateway preparation

perekopskiy:
## What ❔

- adds new fields to DB tables and rust structs
- adds new config variables
- update commitment generator to work with post-gateway
- adds new vm subversion (vm fast is not changed yet)

## Why ❔

prepare for gateway, reduce sync-layer-stable diff

## Checklist


core/lib/multivm/src/versions/vm_latest/types/internals/pubdata.rs
```diff
@@ -9,17 +15,19 @@ pub(crate) struct PubdataInput {
     pub(crate) l2_to_l1_messages: Vec<Vec<u8>>,
     pub(crate) published_bytecodes: Vec<Vec<u8>>,
     pub(crate) state_diffs: Vec<StateDiffRecord>,
+    pub(crate) l2_to_l1_logs_tree_size: usize,
```
slowli, line 18: AFAICT, for now, this field only depends on the protocol version. Would it not be true in the future? If it does only depend on the protocol version, it could make sense to not make it a part of `PubdataInput` (input seems like something produced by the VM), but rather supply the protocol version to the `PubdataInput` mapper in some other way.

core/lib/multivm/src/versions/vm_latest/types/internals/pubdata.rs
```diff
@@ -1,4 +1,10 @@
-use zksync_types::writes::{compress_state_diffs, StateDiffRecord};
+use zksync_mini_merkle_tree::MiniMerkleTree;
```
slowli, line 1: Not really a fan of the `PubdataInput` dataflow. Optimally, I'd imagine there would be a trait mapping `PubdataInput` to `Vec<u8>` published pubdata (for bootloader memory and external consumption in `FinishedL1Batch`; maybe 2 versions if there is no well-defined transform between both; or maybe it makes sense to change `FinishedL1Batch.pubdata_input` type to `Option<PubdataInput>` to only keep the bootloader memory version). This trait would be used by the VM similarly to storage (probably with dynamic dispatch; it doesn't look performance-critical). This would allow to not thread `PubdataParams` everywhere and to not include additional deps for `multivm`. OTOH, I suspect this would require non-trivial amount of refactoring.

core/lib/types/src/commitment/mod.rs
```diff
@@ -92,6 +93,16 @@ pub struct L1BatchMetadata {
...
     pub bootloader_initial_content_commitment: Option<H256>,
     pub state_diffs_compressed: Vec<u8>,
+    /// Hash of packed state diffs. It's present only for post-gateway batches.
+    pub state_diff_hash: Option<H256>,
+    /// Root hash of the local logs tree. Tree contains logs that were produced on this chain.
+    /// It's present only for post-gateway batches.
+    pub local_root: Option<H256>,
+    /// Root hash of the aggregated logs tree. Tree aggregates `local_root`s of chains that settle on this chain.
+    /// It's present only for post-gateway batches.
+    pub aggregation_root: Option<H256>,
```
slowli, line 103: Maybe a dumb suggestion: If these fields are supposed to be present together, maybe it makes sense to extract them to a new struct?

```rust
pub struct L1BatchGatewayMetadata {
    pub state_diff_hash: H256,
    // ...
}

pub struct L1BatchMetadata {
    // ...
    pub gateway: Option<L1BatchGatewayMetadata>,
}
```

core/lib/types/src/commitment/mod.rs
```diff
@@ -401,11 +430,38 @@ impl L1BatchAuxiliaryOutput {
...
                     blob_linear_hashes,
                     blob_commitments,
+                    local_root,
+                    aggregated_root,
                 }
             }
         }
     }
 
+    pub fn get_local_root(&self) -> H256 {
```
slowli, line 440: Bikeshedding: Although this naming is used in the codebase already, `get_` prefix [is considered](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter) redundant in Rust.

core/lib/types/src/commitment/mod.rs
```diff
@@ -670,6 +729,8 @@ pub enum CommitmentInput {
         state_diffs: Vec<StateDiffRecord>,
         aux_commitments: AuxCommitments,
         blob_commitments: Vec<H256>,
+        blob_linear_hashes: Vec<H256>,
```
slowli, line 732: Could make sense to define `blob_commitments_and_hashes: Vec<(H256, H256)>` to make (some) invalid states non-representable. Or even define a struct with `commitment` and `linear_hash` fields.

core/lib/types/src/commitment/mod.rs
```diff
@@ -634,6 +690,9 @@ impl L1BatchCommitment {
             aux_commitments: self.aux_commitments(),
             compressed_initial_writes,
             compressed_repeated_writes,
+            local_root: self.auxiliary_output.get_local_root(),
+            aggregation_root: self.auxiliary_output.get_aggregated_root(),
```
slowli, line 694: Bikeshedding: Naming mismatch ("aggregation" vs "aggregated").

core/lib/types/src/l2_to_l1_log.rs
```diff
@@ -80,10 +80,14 @@ pub fn l2_to_l1_logs_tree_size(protocol_version: ProtocolVersionId) -> usize {
...
 
 /// Returns the blob hashes parsed out from the system logs
-pub fn parse_system_logs_for_blob_hashes(
+pub fn parse_system_logs_for_blob_hashes_pre_gateway(
     protocol_version: &ProtocolVersionId,
     system_logs: &[SystemL2ToL1Log],
 ) -> Vec<H256> {
+    if !protocol_version.is_pre_gateway() {
+        panic!("Cannot parse blob linear hashes from system logs for post gateway");
+    }
```
slowli, line 89: Nit: Can be replaced with an `assert!`.

core/node/commitment_generator/src/utils.rs
```diff
@@ -234,3 +241,79 @@ pub(crate) fn convert_vm_events_to_log_queries(events: &[VmEvent]) -> Vec<LogQue
...
+pub(crate) fn pubdata_to_blob_linear_hashes(
+    blobs_required: usize,
+    mut pubdata_input: Vec<u8>,
+) -> Vec<H256> {
+    // Now, we need to calculate the linear hashes of the blobs.
+    // Firstly, let's pad the pubdata to the size of the blob.
+    if pubdata_input.len() % ZK_SYNC_BYTES_PER_BLOB != 0 {
+        let padding =
+            vec![0u8; ZK_SYNC_BYTES_PER_BLOB - pubdata_input.len() % ZK_SYNC_BYTES_PER_BLOB];
+        pubdata_input.extend(padding);
```
slowli, line 254: Nit: There's no need to allocate an additional `Vec` here; it's possible to use [`Vec::resize()`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.resize).

core/node/commitment_generator/src/utils.rs
```diff
@@ -234,3 +241,79 @@ pub(crate) fn convert_vm_events_to_log_queries(events: &[VmEvent]) -> Vec<LogQue
...
+
+    let (_, last_l2_block) = connection
+        .blocks_dal()
+        .get_l2_block_range_of_l1_batch(l1_batch_number)
+        .await?
+        .context("No range for batch")?;
+
+    let agg_tree_height_slot = StorageKey::new(
+        AccountTreeId::new(L2_MESSAGE_ROOT_ADDRESS),
+        u256_to_h256(AGG_TREE_HEIGHT_KEY.into()),
```
slowli, line 287: Nit: It's possible to use `H256::from_low_u64_be` here.

core/node/commitment_generator/src/utils.rs
```diff
@@ -234,3 +241,79 @@ pub(crate) fn convert_vm_events_to_log_queries(events: &[VmEvent]) -> Vec<LogQue
...
+
+    result
+}
+
+pub(crate) async fn read_aggregation_root(
+    connection: &mut Connection<'_, Core>,
+    l1_batch_number: L1BatchNumber,
+) -> anyhow::Result<H256> {
+    // Position of `FullTree::_height` in `MessageRoot`'s storage layout.
+    const AGG_TREE_HEIGHT_KEY: usize = 3;
```
slowli, line 274: AFAICT, some other contract layout constants are defined in the `system_constants` crate (e.g., for `SystemContext`). Maybe, it would make sense to define these constants there as well?

core/node/node_framework/src/implementations/layers/state_keeper/output_handler.rs
```diff
@@ -84,6 +89,38 @@ impl OutputHandlerLayer {
         self.protective_reads_persistence_enabled = protective_reads_persistence_enabled;
         self
     }
+
+    async fn validate_l2_legacy_shared_bridge_addr(
```
slowli, line 93: In general, I don't think it's a good idea to either define business logic in `node_framework` (mentioned in https://github.com/matter-labs/zksync-era/pull/2949), or to run it during wiring (as opposed to a task).

core/node/node_framework/src/implementations/layers/state_keeper/output_handler.rs
```diff
@@ -84,6 +89,38 @@ impl OutputHandlerLayer {
...
+                .blocks_dal()
+                .get_l2_block_header(l2_block)
+                .await
+                .context("failed to load L2 block header")?
+                .context("missing L2 block header")?;
+            let protocol_version = header
+                .protocol_version
+                .unwrap_or_else(ProtocolVersionId::last_potentially_undefined);
+
+            if protocol_version.is_pre_gateway() && self.l2_legacy_shared_bridge_addr.is_none() {
```
slowli, line 115: I don't think this check will work with pruning enabled (i.e., it can eventually lead to a node panicking). Maybe, it makes sense to query the protocol version of the genesis block instead, and skip the check if there's no genesis (i.e., a node was recovered from a snapshot and/or has pruning enabled)? Or will such a check be not strict enough?

core/node/proof_data_handler/src/request_processor.rs
```diff
@@ -226,58 +226,64 @@ impl RequestProcessor {
...
+                    .metadata
+                    .bootloader_initial_content_commitment
+                    .expect("No bootloader_initial_content_commitment");
+
+                if events_queue_state != events_queue_state_from_prover
+                    || bootloader_heap_initial_content
+                        != bootloader_heap_initial_content_from_prover
+                {
+                    let server_values = format!("events_queue_state = {events_queue_state}, bootloader_heap_initial_content = {bootloader_heap_initial_content}");
+                    let prover_values = format!("events_queue_state = {events_queue_state_from_prover}, bootloader_heap_initial_content = {bootloader_heap_initial_content_from_prover}");
```
slowli, line 248: Nit: I get that this is copy-pasted, but I don't understand the reasoning behind `format!` if it's only used inside another interpolated string.

core/lib/types/src/block.rs
```diff
@@ -91,6 +91,7 @@ pub struct L2BlockHeader {
     /// amount of gas can be spent on pubdata.
     pub gas_limit: u64,
     pub logs_bloom: Bloom,
+    pub pubdata_params: PubdataParams,
```
slowli, line 94: Dumb question: Can `pubdata_params` be changed in a middle of batch? (Looks like not, judging by the current implementation.) Or are they present in blocks rather than L1 batches because of the issue with storing these params for a pending batch? In any case, it looks misleading to place these params in `SystemEnv` (vs `L1BatchEnv`) because they can change per L1 batch without any sort of coordination (vs protocol version or base system contracts).

core/lib/multivm/src/versions/vm_latest/types/internals/pubdata.rs
perekopskiy: I agree with you and IMO you described the end goal accurately. It doesn't seem like something is blocking it, I will try to implement

core/lib/types/src/commitment/mod.rs
perekopskiy: What's the benefit? Yeah, it seems like with current implementation they will be present together but I don't think we should rely on it. Comment I left is somewhat misleading, they won't be present only for post-gateway, but they are expected to be present for each post-gateway batch.

StanislavBreadless (dismissed):

core/node/node_framework/src/implementations/layers/state_keeper/output_handler.rs
perekopskiy: if info about earliest batches is missing then check will still work correctly (but won't be that strict) because `earliest_batch.protocol_version >= genesis_batch.protocol_version` and `is_pre_gateway` is monotonic. Also, check will be performed for main nodes that always have full history and ENs fetch data from main node so transitively it means data they receive is already strict-checked

core/lib/types/src/block.rs
perekopskiy: Yes, they're present in blocks because of the pending batch issue. I hope that `pending_batch` will be in DB really soon and then we can remove pubdata_params from miniblocks.
Also, extending Stan's answer, it requires coordination but it's not a protocol version parameter. It's not decided how that coordination will look like nor implemented yet, effectively it means that operator is not allowed to change `pubdata_params` for now

core/lib/basic_types/src/protocol_version.rs
```diff
@@ -120,6 +121,7 @@ impl ProtocolVersionId {
             ProtocolVersionId::Version23 => VmVersion::Vm1_5_0SmallBootloaderMemory,
             ProtocolVersionId::Version24 => VmVersion::Vm1_5_0IncreasedBootloaderMemory,
             ProtocolVersionId::Version25 => VmVersion::Vm1_5_0IncreasedBootloaderMemory,
+            ProtocolVersionId::Version26 => VmVersion::VmGateway,
```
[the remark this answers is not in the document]
perekopskiy: I've allocated Version25 for protocol defense. It's not blocking. We will move gateway further in terms of version later if needed

StanislavBreadless (approved):

core/lib/multivm/src/versions/vm_latest/types/internals/pubdata.rs
perekopskiy: PTAL, I have updated PR. Now it uses `Rc<dyn PubdataBuilder>` inside VM (`Rc` is used instead of `Box` only  because I need `Clone` for shadow VM), and converts `PubdataParams` to `PubdataBuilder` in all places where VM is inited

core/lib/vm_interface/src/executor.rs
```diff
@@ -20,6 +20,7 @@ pub trait BatchExecutorFactory<S: Send + 'static>: 'static + Send + fmt::Debug {
         storage: S,
         l1_batch_params: L1BatchEnv,
         system_env: SystemEnv,
+        pubdata_params: PubdataParams,
```
slowli, line 23: Dumb question: Do `PubdataParams` not fit into either `L1BatchEnv` or `SystemEnv`? Or would they fit *now*, but cease to fit in the future when custom `PubdataBuilder`s would be implemented? In general, what's the relationship between `PubdataParams` and a `PubdataBuilder`? Is it possible to create a builder from params, or not necessarily?

core/lib/vm_interface/src/pubdata/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+    pub user_logs: Vec<L1MessengerL2ToL1Log>,
+    pub l2_to_l1_messages: Vec<Vec<u8>>,
+    pub published_bytecodes: Vec<Vec<u8>>,
+    pub state_diffs: Vec<StateDiffRecord>,
+}
+
+/// Trait that encapsulates pubdata building logic. It is implemented for rollup and validium cases.
+/// If chains needs custom pubdata format then another implementation should be added.
+pub trait PubdataBuilder: std::fmt::Debug {
+    fn pubdata_params(&self) -> Option<PubdataParams> {
```
slowli, line 80: Why are the returned params optional? AFAICT, both existing impls return `Some(_)`. If the returned params were mandatory, `l2_da_validator` would be redundant IIUC (the address is a part of params).

core/lib/vm_interface/src/pubdata/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+        protocol_version: ProtocolVersionId,
+    ) -> Vec<u8>;
+
+    fn settlement_layer_pubdata(
+        &self,
+        input: PubdataInput,
+        protocol_version: ProtocolVersionId,
+    ) -> Vec<u8>;
+
+    fn bootloader_memory_input(
```
slowli, line 98: Dumb question: Would customizing this implementation work (i.e., not break invariants)? If it's not supposed to be overridden, it makes sense to move it to the `multivm` crate.

core/lib/vm_interface/src/vm.rs
```diff
@@ -75,7 +77,12 @@ impl<T: VmInterface> VmInterfaceExt for T {}
 /// Encapsulates creating VM instance based on the provided environment.
 pub trait VmFactory<S>: VmInterface {
     /// Creates a new VM instance.
-    fn new(batch_env: L1BatchEnv, system_env: SystemEnv, storage: StoragePtr<S>) -> Self;
+    fn new(
+        batch_env: L1BatchEnv,
+        system_env: SystemEnv,
+        storage: StoragePtr<S>,
+        pubdata_builder: Option<Rc<dyn PubdataBuilder>>,
```
slowli, line 84: - Dumb question: Why is `pubdata_builder` optional? Isn't publishing pubdata a mandatory part of finishing a batch? (Initially, I thought that it could be set to `None` for oneshot executor, but it's set there as well.)
- Related: could it be possible to move the `PubdataBuilder` arg to `VmInterface::finish_batch`, so that its lifecycle is clearer? (And maybe it could be possible to supply it by ref in this case, although not 100% sure it'll work for old VMs.) Or can pubdata be requested at any point?

core/lib/multivm/src/versions/vm_fast/vm.rs
```diff
@@ -537,6 +538,7 @@ where
         batch_env: L1BatchEnv,
         system_env: SystemEnv,
         storage: StoragePtr<StorageView<S>>,
+        _pubdata_builder: Option<Rc<dyn PubdataBuilder>>,
```
slowli, line 548: Does it mean that the new VM doesn't correctly build pubdata now? How difficult would be to support it?

core/lib/vm_executor/src/storage.rs
```diff
@@ -53,7 +54,8 @@ pub fn l1_batch_params(
     protocol_version: ProtocolVersionId,
     virtual_blocks: u32,
     chain_id: L2ChainId,
-) -> (SystemEnv, L1BatchEnv) {
+    pubdata_params: PubdataParams,
+) -> (SystemEnv, L1BatchEnv, PubdataParams) {
```
slowli, line 58: Adding an arg just to get it back looks weird.

core/lib/vm_interface/Cargo.toml
```diff
@@ -14,6 +14,8 @@ categories.workspace = true
 zksync_contracts.workspace = true
 zksync_system_constants.workspace = true
 zksync_types.workspace = true
+zksync_utils.workspace = true
+zksync_mini_merkle_tree.workspace = true
```
slowli, line 18: If possible, I'd advise against adding either of these 2 deps, since it makes the interface crate too heavy. Could it be possible e.g. to add `PubdataBuilder` implementations to the executor crate?

core/lib/vm_interface/src/pubdata/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+pub trait PubdataBuilder: std::fmt::Debug {
+    fn pubdata_params(&self) -> Option<PubdataParams> {
+        None
+    }
+
+    fn l2_da_validator(&self) -> Address;
+
+    fn l1_messenger_operator_input(
+        &self,
+        input: PubdataInput,
```
slowli, line 88: It's relatively straightforward to make this method to take input by ref. Ditto for the methods below.

slowli (commented):
Not a full review, just some quick feedback. Will try to look at the PR more exhaustively tomorrow.

Not going to lie, the current `PubdataBuilder` design looks overly complicated, and in some aspects worse than the ad-hoc design during the last review. I think it could be salvaged though; see line comments.

core/lib/vm_interface/src/executor.rs
perekopskiy: So `PubdataParams` is data that we persist for batches in DB and that we use for building `PubdataBuilder`s. `PubdataParams` is not customizable for now as it requires additional effort and it's not yet clear how to make it so and wire it altogether with node framework. I'm leaving it as an intermediate step until we really need custom `PubdataBuilder`s. Recent changes were to prepare `VmInterface` for custom impls but not the node itself.

core/lib/vm_interface/src/pubdata/mod.rs
perekopskiy: here customizing won't work, makes sense to move it

core/lib/vm_interface/src/vm.rs
perekopskiy: it's optional for older versions, they don't require it. I will investigate possibility of moving it to `finish_batch`, from first glance it can be done

core/lib/multivm/src/versions/vm_fast/vm.rs
perekopskiy: Yes, not supported for now. It is easy to support, basically duplicate changes from old_vm

perekopskiy:
> Not going to lie, the current PubdataBuilder design looks overly complicated, and in some aspects worse than the ad-hoc design during the last review. I think it could be salvaged though; see line comments.

I agree but the previous version was rather naive. I'm trying to find some spot between "keep it as simple as possible and delay dealing with complexity till we need first custom builder" and "implement complex solution that fully allows custom pubdata"

core/lib/vm_interface/src/vm.rs
perekopskiy: I have moved it to `finish_batch`

core/lib/vm_interface/Cargo.toml
perekopskiy: I've moved impls to `multivm`, otherwise it's not possible to run `vm.finish_batch` in multivm tests. It also required moving vm dump playback functions.

core/node/commitment_generator/src/lib.rs
```diff
@@ -263,22 +266,45 @@ impl CommitmentGenerator {
...
             } else {
-                vec![H256::zero(); num_blobs_required(&protocol_version)]
+                vec![Default::default(); num_blobs_required(&protocol_version)]
             };
 
+            let mut connection = self
+                .connection_pool
+                .connection_tagged("commitment_generator")
+                .await?;
+            let aggregation_root = read_aggregation_root(&mut connection, l1_batch_number).await?;
```
slowli, line 299: Dumb question: Won't this root be always zero and unused for `protocol_version < Gateway`? Would not reading it from storage in this case complicate the code too much?

core/node/commitment_generator/src/lib.rs
```diff
@@ -357,14 +383,10 @@ impl CommitmentGenerator {
...
-                L1BatchCommitmentMode::Validium,
-                CommitmentInput::PostBoojum {
-                    blob_commitments, ..
-                },
-            ) => {
-                blob_commitments.fill(H256::zero());
+            (L1BatchCommitmentMode::Validium, CommitmentInput::PostBoojum { blob_hashes, .. }) => {
+                blob_hashes
+                    .iter_mut()
+                    .for_each(|b| b.commitment = H256::zero());
```
slowli, line 389: Bikeshedding: `for` cycle is generally considered preferable to `for_each`.

core/node/consensus/src/storage/store.rs
```diff
@@ -38,6 +38,7 @@ fn to_fetched_block(
         l1_gas_price: payload.l1_gas_price,
         l2_fair_gas_price: payload.l2_fair_gas_price,
         fair_pubdata_price: payload.fair_pubdata_price,
+        pubdata_params: payload.pubdata_params.unwrap_or_default(),
```
slowli, line 41: Dumb question: Is the default value reasonable here (ditto for JSON-RPC-based syncing) because it's unused before the protocol version upgrade, at which point it's expected to be returned by the main node / peers? If the default value is read, it could silently produce unexpected results. So just from this perspective, it'd be safer to use an `Option` here to ensure that the value is indeed never read.

slowli (commented):
Sorry for merging https://github.com/matter-labs/zksync-era/pull/3109 which requires to fix edit conflicts in multivm unit tests (seems quite straightforward – you should move the changes to the test logic in `versions::testonly` submodules.

core/lib/multivm/src/dump.rs
```diff
@@ -0,0 +1,49 @@
...
+pub fn play_back_dump<Vm>(dump: VmDump) -> Vm
+where
+    Vm: VmFactory<StorageView<StorageSnapshot>>,
+{
+    play_back_dump_custom(dump, Vm::new)
+}
+
+/// Plays back dump on a VM created using the provided closure.
+#[doc(hidden)] // too low-level
+pub fn play_back_dump_custom<Vm: VmInterface>(
```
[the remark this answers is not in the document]
perekopskiy: yes, it will work

core/lib/vm_interface/src/utils/dump.rs
```diff
@@ -129,12 +93,14 @@ impl<S: ReadStorage, Vm: VmTrackingContracts> DumpingVm<S, Vm> {
...
 
-    pub fn dump_state(&self) -> VmDump {
+    pub fn dump_state(&self, pubdata_builder: Option<Rc<dyn PubdataBuilder>>) -> VmDump {
         VmDump {
             l1_batch_env: self.l1_batch_env.clone(),
             system_env: self.system_env.clone(),
             l2_blocks: self.l2_blocks.clone(),
             storage: create_storage_snapshot(&self.storage, self.inner.used_contract_hashes()),
+            pubdata_params: pubdata_builder
+                .map(|p| p.pubdata_params().expect("pubdata builder is not dumpable")),
```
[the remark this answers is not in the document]
perekopskiy: I made it possible to leave it as None so we can replay part of execution before finish_batch, it's useful when divergence happens before `finish_batch`. I see that I didn't support it properly as way of running it partially wasn't added. Anyway, I would go with a suggestion to not run `finish_batch` when playing back, so pubdata_builder/pubdata_params will be removed from dump-related code

core/lib/vm_interface/src/vm.rs
```diff
@@ -46,13 +48,13 @@ pub trait VmInterface {
 
     /// Execute batch till the end and return the result, with final execution state
     /// and bootloader memory.
-    fn finish_batch(&mut self) -> FinishedL1Batch;
+    fn finish_batch(&mut self, pubdata_builder: Option<Rc<dyn PubdataBuilder>>) -> FinishedL1Batch;
```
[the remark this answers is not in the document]
perekopskiy: I added `None` so caller doesn't have to create some pubdata builder if it calls VM that doesn't need it. But you are right, it shouldn't cause any issues for them, they can use e.g. `RollupPubdataBuilder::new(Address::zero())`. So I will remove optionality to avoid some boilerplate

core/node/commitment_generator/src/lib.rs
perekopskiy: yes, it will be zero. No, it won't complicate, makes sense to wrap it into if

core/node/commitment_generator/src/utils.rs
```diff
@@ -234,3 +242,75 @@ pub(crate) fn convert_vm_events_to_log_queries(events: &[VmEvent]) -> Vec<LogQue
...
+
+pub(crate) async fn read_aggregation_root(
+    connection: &mut Connection<'_, Core>,
+    l1_batch_number: L1BatchNumber,
+) -> anyhow::Result<H256> {
+    let (_, last_l2_block) = connection
+        .blocks_dal()
+        .get_l2_block_range_of_l1_batch(l1_batch_number)
+        .await?
+        .context("No range for batch")?;
```
slowli, line 280: Nit: Could make sense to include batch number in the message (although such an error is quite unlikely to happen).

core/lib/vm_interface/src/pubdata/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+}
+
+/// Struct based on which the pubdata blob is formed
+#[derive(Debug, Clone, Default)]
+pub struct PubdataInput {
+    pub user_logs: Vec<L1MessengerL2ToL1Log>,
+    pub l2_to_l1_messages: Vec<Vec<u8>>,
+    pub published_bytecodes: Vec<Vec<u8>>,
```
slowli, line 70: Dumb question / probably not important rn: Will this include EVM bytecodes if EVM emulator is enabled? (AFAIU, these bytecodes are at least known to the decommitment processor via [the EVM deployment tracer](https://github.com/matter-labs/zksync-era/blob/main/core/lib/multivm/src/versions/vm_latest/tracers/evm_deploy_tracer.rs).) *Should* it include such bytecodes?

slowli (dismissed):
Can't say I've looked thoroughly at 100% of the code, but it looks good in general.

core/lib/vm_interface/src/types/inputs/execution_mode.rs
```diff
@@ -13,3 +13,22 @@ pub enum VmExecutionMode {
     /// Stop after executing the entire bootloader. But before you exit the bootloader.
     Bootloader,
 }
+
+/// Subset of `VmExecutionMode` variants that do not require any additional input
+/// and can be invoked with `inspect` method.
+#[derive(Debug, Copy, Clone)]
+pub enum InspectExecutionMode {
```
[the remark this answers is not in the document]
perekopskiy: I do not have strong opinion, both variants look okayish

core/lib/vm_interface/src/pubdata/mod.rs
perekopskiy: I've checked `stable/evm-emulator` branch in era-contracts, there `publishEVMBytecode` calls `L1_MESSENGER_CONTRACT.sendToL1` so these bytecodes will be included in `l2_to_l1_messages`. Historically we had a bit of a mess, compressed bytecodes were published to L1 through messages, uncompressed as `published_bytecodes` or vice versa I don't remember. It seems it has changed and now all eraVM bytecode go through `published_bytecodes`, I'm not sure why EVM bytecodes are different, cc @0xVolosnikov.
Overall it's not really important for node to which field bytecodes are included

0xVolosnikov: It would nice to switch to `published_bytecodes`, but i'm not sure what exactly is blocking us from this

@slowli any ideas?

perekopskiy: Changing `sendToL1` to `requestBytecodeL1Publication` [here](https://github.com/matter-labs/era-contracts/blob/stable/evm-emulator/system-contracts/contracts/KnownCodesStorage.sol#L101) should generally work

0xVolosnikov: I haven't had a chance to delve into it yet

slowli (approved):

slowli (approved):

PR #3014, opened 2024-10-04
refactor(eth): Brush up `eth_signer` crate

slowli:
## What ❔

Brushes up the `eth_signer` crate so that its API is easier to use (e.g., it doesn't require an async runtime).

## Why ❔

Improved DevEx.

## Checklist


popzxc (commented):
Meta-comment: we may want to have [AWS Signer](https://docs.rs/alloy-signer-aws/latest/alloy_signer_aws/struct.AwsSigner.html), [GCP Signer](https://docs.rs/alloy-signer-gcp/latest/alloy_signer_gcp/struct.GcpSigner.html) and maybe custom signer for a signing microservice in foreseeable future (at least I hear requests for it from PE more and more often), so I'm not sure if changing the trait to be sync is a good decision.

If we are to refactor it, I'd probably align interfaces somewhat with [alloy_signer](https://docs.rs/alloy-signer/latest/alloy_signer/) to simplify future migration. It provides [SignerSync](https://docs.rs/alloy-signer/latest/alloy_signer/trait.SignerSync.html) abstraction, but I'm not sure how useful is it for us today.

slowli:
@popzxc The signer trait isn't changed at all, I've just added inherent methods to `PrivateKeySigner`.

popzxc:
Ah, sorry, I just read the description.

perekopskiy (approved):

PR #3015, opened 2024-10-04
chore(vm): Remove tests for old VM versions

slowli:
## What ❔

Removes unit test files for old VM versions in the `multivm` crate.

## Why ❔

These tests are not built (not included into the module tree of the crate), so they are just deadweight.

## Checklist


joonazan:
Having to maintain these tests is bad but also the old VM's code changes from time to time. If we could disallow any changes to the old VM's I'd be happy.

slowli:
> Having to maintain these tests is bad

@joonazan These tests are *not* maintained (I think they are quire bit-rotten already) – they are not compiled or run; that's the reason for removing them 🙃 Regarding changes, I don't think it's realistically possible at this point because of technical debt in VM interfaces (`VmInterface` etc.). These aren't changes to VM semantics though; I'm not aware of any (somewhat) recent changes to old VMs.

perekopskiy (approved):

PR #3017, opened 2024-10-04
fix(tee_verifier): correctly initialize storage for re-execution

haraldh:
## What ❔

With this commit, the TEE verifier uses `WitnessStorageState` of `VMRunWitnessInputData` to initialize the storage. This requires waiting for the BasicWitnessInputProducer to complete and therefore the TEE verifier input producer can be removed. The input for the TEE verifier is now collected in the `proof_data_handler`, which enables to remove the whole job queue for the TEE verifier input producer.

## Why ❔

Previously the storage for VM re-execution was initialized just from `WitnessInputMerklePaths`. This although misses the storage values for slots, which are only read/written to by rolled back transactions. This led to failed verification of blocks, which would normally pass.

## Checklist


core/lib/dal/src/tee_proof_generation_dal.rs
```diff
@@ -41,44 +38,72 @@ impl TeeProofGenerationDal<'_, '_> {
...
-                    LIMIT
-                        1
-                    FOR UPDATE
-                    SKIP LOCKED
-                )
+                    )
+                    AND (
+                        proofs.l1_batch_number IS NULL
+                        OR proofs.l1_batch_number >= $5
+                    )
```
pbeza, line 76: This condition feels more natural for me (I hope it's correct):
```
                    AND (
                        (
                            proofs.tee_type IS NULL
                            AND proofs.status IS NULL
                            AND proofs.l1_batch_number IS NULL
                        ) OR (
                            proofs.tee_type = $1
                            AND (
                                proofs.status = $3
                                OR (
                                    proofs.status = $2
                                    AND proofs.prover_taken_at < NOW() - $4::INTERVAL
                                )
                            )
                            AND proofs.l1_batch_number >= $5
                        )
                    )
```
I think your variant evaluates to `true` in more cases, e.g. when `proofs.tee_type IS NULL AND proofs.status = $3 AND proofs.l1_batch_number >= $5`. It's probably not possible because of how DB schema is defined (`NOT NULL` columns), but it's not so obvious. 🤔

haraldh:
```
                    AND (
                        (
                            proofs.tee_type IS NULL
                            AND proofs.status IS NULL
                            AND proofs.l1_batch_number IS NULL
                        ) OR (
                            proofs.l1_batch_number >= $5
                            AND (
                                proofs.tee_type != $1
                                OR
                                (
                                    proofs.tee_type = $1
                                    AND (
                                        proofs.status = $3
                                        OR (
                                            proofs.status = $2
                                            AND proofs.prover_taken_at
                                            < NOW() - $4::INTERVAL
                                        )
                                    )
                                )
                            )
                        )
                    )

```

haraldh: Pushed a changing commit with your version... I added pre filtering of the tee proof table with `tee_type` though, to exclude other TEE techs from the query.

pbeza: I believe [this line](https://github.com/matter-labs/zksync-era/pull/3017/commits/fa06b18517993810ee4ab3513029002ee2d57a21#diff-46d00b49fb68d6da13c9ecdc5ab3dab6adf502ec216349721586e96553652444R76) is now redundant.

BTW, one of the linter tests [is failing](https://github.com/matter-labs/zksync-era/actions/runs/11211617460/job/31160930751?pr=3017).

core/bin/zksync_server/src/main.rs
```diff
@@ -45,7 +45,7 @@ struct Cli {
     /// Comma-separated list of components to launch.
     #[arg(
         long,
-        default_value = "api,tree,eth,state_keeper,housekeeper,tee_verifier_input_producer,commitment_generator,da_dispatcher,vm_runner_protective_reads"
```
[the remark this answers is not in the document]
haraldh: yes... `tee_verifier_input_producer` is gone

core/lib/object_store/src/file.rs
```diff
@@ -42,7 +42,6 @@ impl FileBackedObjectStore {
             Bucket::SchedulerWitnessJobsFri,
             Bucket::ProofsFri,
             Bucket::StorageSnapshot,
-            Bucket::TeeVerifierInput,
```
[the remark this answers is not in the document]
haraldh: Honestly, IDK

core/node/vm_runner/src/impls/bwip.rs
```diff
@@ -257,6 +258,7 @@ async fn get_updates_manager_witness_input_data(
         storage_refunds,
         pubdata_costs,
         witness_block_state,
+        l2_chain_id,
```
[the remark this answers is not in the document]
haraldh: yes please... otherwise it gets super complicated again

core/lib/dal/src/tee_proof_generation_dal.rs
```diff
@@ -41,44 +38,81 @@ impl TeeProofGenerationDal<'_, '_> {
...
-                        tee_verifier_input_producer_jobs AS inputs
-                        ON proofs.l1_batch_number = inputs.l1_batch_number
-                    WHERE
-                        inputs.status = $3
-                        AND (
-                            proofs.status = $4
-                            OR (
-                                proofs.status = $1
-                                AND proofs.prover_taken_at < NOW() - $5::INTERVAL
+            WITH
```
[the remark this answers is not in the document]
haraldh: isn't it the same with `proof_generation_details` table?

core/lib/prover_interface/src/inputs.rs
```diff
@@ -161,7 +163,7 @@ impl StoredObject for VMRunWitnessInputData {
     serialize_using_bincode!();
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
```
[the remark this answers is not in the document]
haraldh: Does it cause issues? I used it for asserts in tests

core/lib/object_store/src/file.rs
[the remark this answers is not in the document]
pbeza: I just opened a series of PRs for each environment to delete this bucket from each one:
1. https://github.com/matter-labs/terraform-configurations/pull/4330 (`stage`)
1. https://github.com/matter-labs/terraform-configurations/pull/4331 (`testnet`)
1. https://github.com/matter-labs/terraform-configurations/pull/4332 (`mainnet`)

core/lib/dal/src/tee_proof_generation_dal.rs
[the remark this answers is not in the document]
haraldh: let me know, if you like the current version of the SQL query more

pbeza (commented):
We should probably remove [this function](https://github.com/matter-labs/zksync-era/blob/3a9716829e7a1d64953c20bdf824ed02c767c629/core/lib/dal/src/tee_proof_generation_dal.rs#L199-L229) altogether:

https://github.com/matter-labs/zksync-era/blob/3a9716829e7a1d64953c20bdf824ed02c767c629/core/lib/dal/src/tee_proof_generation_dal.rs#L199-L229

You are now inserting new entries in the `lock_batch_for_proving` function instead:

https://github.com/matter-labs/zksync-era/blob/3a9716829e7a1d64953c20bdf824ed02c767c629/core/lib/dal/src/tee_proof_generation_dal.rs#L78-L104

core/node/vm_runner/src/impls/bwip.rs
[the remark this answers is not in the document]
haraldh: removed... found an easy way

core/lib/dal/src/tee_proof_generation_dal.rs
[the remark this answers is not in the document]
haraldh: it shouldn't blow up on big tables anymore, because I removed the sort... so it should be O(n)

haraldh:
> I've left my concerns on the current implementation. If there's a doc for me to understand what's going on, that'd be great, could provide better input. Otherwise, maybe writing it down (maybe as part of PR?) would help moving this forward.

@EmilLuta which part is unclear, not enough explained in the PR comment?

core/lib/tee_verifier/src/lib.rs
```diff
@@ -50,30 +51,43 @@ impl Verify for V1TeeVerifierInput {
...
-        }
+        let initial_writes_ops = self
+            .vm_run_data
+            .witness_block_state
+            .is_write_initial
+            .into_iter();
 
-        let block_output_with_proofs =
-            get_bowp_and_set_initial_values(self.witness_input_merkle_paths, &mut raw_storage);
+        let storage = read_storage_ops
```
slowli, line 69: Bikeshedding: I'd add a comment explaining why the storage is built like this (e.g., why enumeration indices don't matter).

core/lib/tee_verifier/src/lib.rs
```diff
@@ -191,36 +212,51 @@ fn map_log_tree(
...
                     value
                 );
+                anyhow::bail!("Failed to map LogQuery to TreeInstruction");
             }
             TreeInstruction::Read(key)
         }
-        (false, TreeLogEntry::ReadMissingKey { .. }) => TreeInstruction::Read(key),
-        _ => {
-            tracing::error!("Failed to map LogQuery to TreeInstruction");
+        (true, StorageLogKind::RepeatedWrite, TreeLogEntry::Read { .. }) => {
```
slowli, line 237: Is there a particular reason this match hand is treated separately from `(true, _, TreeLogEntry::Read { .. })` below? AFAIU, it uses same handling. If it were removed, the added match on `storage_log.kind` would become unnecesssary.

core/node/proof_data_handler/Cargo.toml
```diff
@@ -17,6 +17,8 @@ zksync_dal.workspace = true
 zksync_object_store.workspace = true
 zksync_prover_interface.workspace = true
 zksync_types.workspace = true
+zksync_vm_executor.workspace = true
```
slowli, line 20: Off-topic: This dependency is unfortunate given that `vm_executor` transitively depends on all VM implementations. The assumption behind including `L1BatchParamsProvider` in `vm_executor` was that its users would also execute VM based on returned `(SystemEnv, L1BatchEnv)`, but apparently this is not the case here.

core/node/proof_data_handler/src/errors.rs
```diff
@@ -19,6 +20,13 @@ impl From<DalError> for RequestProcessorError {
 impl IntoResponse for RequestProcessorError {
     fn into_response(self) -> Response {
         let (status_code, message) = match self {
+            RequestProcessorError::GeneralError(err) => {
+                tracing::error!("Error: {:?}", err);
+                (
+                    StatusCode::BAD_GATEWAY,
```
slowli, line 26: Bikeshedding: "Bad gateway" [doesn't look like](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/502) the best status code here:

> indicates that a server was acting as a gateway or [proxy](https://developer.mozilla.org/en-US/docs/Glossary/Proxy_server) and that it received an invalid response from the upstream server.

IMO, [500 Internal Server Error](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500) would work better.

core/node/proof_data_handler/src/lib.rs
```diff
@@ -88,17 +96,23 @@ fn create_proof_processing_router(
...
+            TeeRequestProcessor::new(blob_store, connection_pool, config.clone(), l2_chain_id);
         let submit_tee_proof_processor = get_tee_proof_gen_processor.clone();
         let register_tee_attestation_processor = get_tee_proof_gen_processor.clone();
 
         router = router.route(
             "/tee/proof_inputs",
             post(
                 move |payload: Json<TeeProofGenerationDataRequest>| async move {
-                    get_tee_proof_gen_processor
+                    let result: Result<Json<zksync_prover_interface::api::TeeProofGenerationDataResponse>, errors::RequestProcessorError> = get_tee_proof_gen_processor
```
slowli, line 107: Nit: Type annotation looks unnecessary, and it makes code less readable.

core/node/proof_data_handler/src/lib.rs
```diff
@@ -88,17 +96,23 @@ fn create_proof_processing_router(
...
                         .get_proof_generation_data(payload)
-                        .await
+                        .await;
+
+                    match result {
+                        Ok(Json(TeeProofGenerationDataResponse(None))) => (StatusCode::NO_CONTENT, Json("No new TeeVerifierInputs are available yet")).into_response(),
```
slowli, line 112: "No content" implies empty response body; not sure whether it'll be correctly handled by clients though (but a string response would probably require special handling as well).

core/node/proof_data_handler/src/lib.rs
```diff
@@ -88,17 +96,23 @@ fn create_proof_processing_router(
...
         let submit_tee_proof_processor = get_tee_proof_gen_processor.clone();
         let register_tee_attestation_processor = get_tee_proof_gen_processor.clone();
 
         router = router.route(
             "/tee/proof_inputs",
             post(
                 move |payload: Json<TeeProofGenerationDataRequest>| async move {
-                    get_tee_proof_gen_processor
+                    let result: Result<Json<zksync_prover_interface::api::TeeProofGenerationDataResponse>, errors::RequestProcessorError> = get_tee_proof_gen_processor
                         .get_proof_generation_data(payload)
```
slowli, line 108: Off-topic (?): The design of `get_proof_generation_data()` looks really weird; it mixes data access and response representation layers which should ideally be cleanly separated. Representing responses in the responsibility of the API server, i.e. *this* code.

etc/env/file_based/general.yaml
```diff
@@ -312,7 +312,7 @@ prometheus:
...
...ification_server=info,zksync_node_api_server=info,zksync_node_framework=info,zksync_block_reverter=info,zksync_commitment_generator=debug,zksync_node_db_pruner=info,zksync_eth_sender=info,zksync_node_fee_model=info,zksync_node_genesis=info,zksync_house_keeper=info,zksync_proof_data_handler=info,zksync_shared_metrics=info,zksync_node_test_utils=info,zksync_vm_runner=info,zksync_consensus_bft=info,zksync_consensus_network=info,zksync_consensus_storage=info,zksync_core_leftovers=debug,zksync_server=debug,zksync_contract_verifier=debug,zksync_dal=info,zksync_db_connection=info,zksync_eth_client=info,zksync_eth_watch=debug,zksync_storage=info,zksync_db_manager=info,zksync_merkle_tree=info,zksync_state=debug,zksync_utils=debug,zksync_queued_job_processor=info,zksync_types=info,zksync_mempool=debug,loadnext=info,vm=info,zksync_object_store=info,zksync_external_node=info,zksync_witness_generator=info,zksync_prover_fri=info,zksync_witness_vector_generator=info,zksync_web3_decl=debug,zksync_health_check=debug,zksync_proof_fri_compressor=info,vise_exporter=error,snapshots_creator=debug,zksync_base_token_adjuster=debug,zksync_external_price_api=debug,zksync_external_proof_integration_api=info"
```
slowli, line 315: Off-topic (?): It may be time to use `zksync=info` directive to get rid of most directives with this level here; we already use it in our K8s configs.

Cargo.toml
```diff
@@ -1,85 +1,84 @@
 [workspace]
 members = [
-  # Binaries
```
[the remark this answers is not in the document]
haraldh: thx

core/lib/dal/migrations/20241001110000_remove_tee_verifier_input_producer_job.down.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE tee_proof_generation_details (
```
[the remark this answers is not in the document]
haraldh: @pbeza ?

Cargo.toml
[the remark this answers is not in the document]
haraldh: fixed

core/lib/tee_verifier/src/lib.rs
haraldh: Please give me the comment you want to see here, as you are the expert :)

core/node/proof_data_handler/src/errors.rs
haraldh: good catch!

etc/env/file_based/general.yaml
haraldh: not the topic of this PR :shrug:

core/node/proof_data_handler/Cargo.toml
haraldh: No clue how to fix this..

core/lib/dal/src/tee_proof_generation_dal.rs
```diff
@@ -271,25 +257,54 @@ impl TeeProofGenerationDal<'_, '_> {
...
+            )
+            VALUES
+            ($1, $2, $3, NOW(), NOW())
+            ON CONFLICT (l1_batch_number, tee_type) DO NOTHING
+            "#,
+            batch_number,
+            tee_type.to_string(),
+            TeeProofGenerationJobStatus::Unpicked.to_string(),
+        );
+        let instrumentation = Instrumented::new("insert_tee_proof_generation_job")
```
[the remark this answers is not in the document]
haraldh: fixed

core/node/proof_data_handler/src/errors.rs
haraldh: fixed

core/node/proof_data_handler/src/lib.rs
haraldh: fixed

core/node/proof_data_handler/src/lib.rs
haraldh: @pbeza ?

core/node/proof_data_handler/src/lib.rs
haraldh: @pbeza ?

core/lib/prover_interface/src/inputs.rs
```diff
@@ -254,37 +254,37 @@ impl StoredObject for WitnessInputData {
...
 pub struct L1BatchMetadataHashes {
     pub root_hash: H256,
     pub meta_hash: H256,
     pub aux_hash: H256,
 }
 
-/// Version 1 of the data used as input for the TEE verifier.
+/// Version 2 of the data used as input for the TEE verifier.
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct V1TeeVerifierInput {
```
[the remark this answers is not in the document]
haraldh: Ah, sorry, at first I thought about changing the version, but then again, we haven't released anything, what we call `stable`, so I thought V1 is still fine.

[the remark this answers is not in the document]
haraldh: reverted to "1"

core/node/proof_data_handler/Cargo.toml
slowli: No need to fix it rn; I'll probably create a dedicated task for it. Abstractly, it could make sense to place `L1BatchParamsProvider` elsewhere, but where exactly isn't clear to me yet (maybe, `zksync_state`, but it would look sort of out of place).

core/node/proof_data_handler/src/lib.rs
slowli: Yeah, no need to rework it rn, that's why I've marked discussion as off-topic. IMO, this firmly falls into the category of tech debt (i.e., we'd want to eventually refactor this logic unless it would be removed shortly).

core/lib/tee_verifier/src/lib.rs
slowli: Something like this:

> We need to define storage slots read during batch execution, and their initialness; hence, the use of both `read_storage_ops` and `initial_writes_ops`. `StorageSnapshot` also requires providing enumeration indices, but they only matter at the end of execution when creating pubdata for the batch, which is irrelevant in this case. Thus, enumeration indices are set to dummy values.

haraldh: added

core/lib/dal/migrations/20241001110000_remove_tee_verifier_input_producer_job.down.sql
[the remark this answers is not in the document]
pbeza: Good catch! I’ve fixed it [here](https://github.com/matter-labs/zksync-era/pull/3017/commits/fc30e5bbc666c69dace1630a1012c2682140f934). All credits goes to [Copilot](https://copilot.microsoft.com/) – it’s awesome for handling the tedious stuff like creating SQL queries for existing tables.

By the way, what’s the point of creating rollback scripts (.down.sql files) if we don’t even test them? AFAIU, it’s more of a "best effort" thing here.

core/node/proof_data_handler/src/lib.rs
pbeza: Great catch! I forgot to test the client after making this change.

I've addressed it in this PR: https://github.com/matter-labs/zksync-era/pull/3067 and have re-tested it manually end-to-end. Please merge it if you approve.

pbeza: @slowli Would you mind reviewing #3067, which addresses your comment?

I’ve resolved the issue, but now I'm returning a `StatusCode::NOT_FOUND` (`404`) HTTP status code instead of `StatusCode::NO_CONTENT` (`204`). This is because we already have a mapping from `zksync_dal::SqlError::RowNotFound` to `StatusCode::NOT_FOUND` here:

https://github.com/matter-labs/zksync-era/blob/748ee1b831c5fc81f6df2442dab3c862f5e19462/core/node/proof_data_handler/src/errors.rs#L32-L34

and I'm now using:

https://github.com/matter-labs/zksync-era/blob/ab6e56552377ea74ccf619f940a03735c0e4d5ea/core/lib/dal/src/tee_proof_generation_dal.rs#L108

(that throws `zksync_dal::SqlError::RowNotFound`), instead of:

https://github.com/matter-labs/zksync-era/blob/3d77b90315a67ce361fc86315021ec0c403d0cbb/core/lib/dal/src/tee_proof_generation_dal.rs#L108

@haraldh suggested that we should return `StatusCode::NO_CONTENT` if no new `TeeVerifierInput` is available yet. However, I don't see an easy way to bypass this existing mapping unless we modify it globally to return `StatusCode::NO_CONTENT` instead of `StatusCode::NOT_FOUND`.

pbeza: I've prepared an alternative solution in https://github.com/matter-labs/zksync-era/pull/3082. Feel free to choose whichever you prefer, or suggest another solution that I can implement.

pbeza: @slowli, could you please review? We would like to get this PR wrapped up and merged so we can move on to other tasks that depend on it. It's a big one that affects a lot, so I would really prefer to have it merged before we move forward.

slowli: I like "not found" errors better; they look more idiomatic.

core/lib/tee_verifier/src/lib.rs
```diff
@@ -191,36 +216,43 @@ fn map_log_tree(
     idx: &mut u64,
 ) -> anyhow::Result<TreeInstruction> {
     let key = storage_log.key.hashed_key_u256();
-    Ok(match (storage_log.is_write(), *tree_log_entry) {
-        (true, TreeLogEntry::Updated { leaf_index, .. }) => {
+    let tree_instruction = match (storage_log.is_write(), storage_log.kind, *tree_log_entry) {
```
slowli, line 219: AFAIU, `storage_log.kind` is not matched now, so it can be removed from the matched expression.

core/lib/tee_verifier/src/lib.rs
```diff
@@ -313,14 +356,11 @@ mod tests {
...
-            vec![(H256([1; 32]), vec![0, 1, 2, 3, 4])],
         );
         let tvi = TeeVerifierInput::new(tvi);
-        let serialized = <TeeVerifierInput as StoredObject>::serialize(&tvi)
-            .expect("Failed to serialize TeeVerifierInput.");
+        let serialized = bincode::serialize(&tvi).expect("Failed to serialize TeeVerifierInput.");
         let deserialized: TeeVerifierInput =
-            <TeeVerifierInput as StoredObject>::deserialize(serialized)
-                .expect("Failed to deserialize TeeVerifierInput.");
+            bincode::deserialize(&serialized).expect("Failed to deserialize TeeVerifierInput.");
```
slowli, line 363: Dumb question: What does this test now? Is `bincode` serialization applied to `TeeVerifierInput` elsewhere?

core/node/proof_data_handler/src/tee_request_processor.rs
```diff
@@ -78,6 +92,66 @@ impl TeeRequestProcessor {
...
+
+        let merkle_paths: WitnessInputMerklePaths = self
+            .blob_store
+            .get(l1_batch_number)
+            .await
+            .map_err(RequestProcessorError::ObjectStore)?;
+
+        let mut connection = self
+            .pool
+            .connection()
```
slowli, line 114: Nit: Please use `connection_tagged` for non-test code; it allows more precise debugging / tracing.

slowli (dismissed):
Other than API changes in #3067, looks fine.

core/node/proof_data_handler/src/lib.rs
pbeza: @haraldh, the `not found` error is from PR #3067. Are you fine with merging it?

core/lib/tee_verifier/src/lib.rs
haraldh: We might want to use it for optimization in the future. Bandwidth plays a bigger role here

core/node/proof_data_handler/src/lib.rs
haraldh: merged with changes

core/lib/tee_verifier/src/lib.rs
haraldh: removed

slowli (approved):

pbeza:
I think [`core/lib/dal/doc/TeeProofGenerationDal.md`](https://github.com/matter-labs/zksync-era/blob/tee_prover_new_2/core/lib/dal/doc/TeeProofGenerationDal.md) got outdated.

core/node/proof_data_handler/src/errors.rs
```diff
@@ -19,34 +20,35 @@ impl From<DalError> for RequestProcessorError {
...
-                    _ => (
-                        StatusCode::BAD_GATEWAY,
-                        "Failed fetching/saving from db".to_owned(),
-                    ),
-                }
+                (
+                    StatusCode::BAD_GATEWAY,
+                    "Failed fetching/saving from db".to_owned(),
+                )
+                    .into_response()
```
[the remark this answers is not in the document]
haraldh: I had that first and didn't like it... I don't care about rustfmt formatting...

core/node/proof_data_handler/src/errors.rs
```diff
@@ -19,34 +20,35 @@ impl From<DalError> for RequestProcessorError {
 
 impl IntoResponse for RequestProcessorError {
     fn into_response(self) -> Response {
-        let (status_code, message) = match self {
+        match self {
             RequestProcessorError::GeneralError(err) => {
```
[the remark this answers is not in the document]
haraldh: changed

core/node/proof_data_handler/src/errors.rs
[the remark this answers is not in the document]
haraldh: anyway.. changed it back to your liking

core/node/proof_data_handler/src/errors.rs
```diff
@@ -19,34 +20,35 @@ impl From<DalError> for RequestProcessorError {
...
-                        "Failed fetching/saving from db".to_owned(),
-                    ),
-                }
+                (
+                    StatusCode::BAD_GATEWAY,
+                    "Failed fetching/saving from db".to_owned(),
+                )
+                    .into_response()
+            }
+            RequestProcessorError::NoJob => {
```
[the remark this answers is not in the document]
haraldh: changed to non error, hope you like it

slowli (approved):

PR #3031, opened 2024-10-08
feat: add `block.timestamp` asserter for AA

ischasny:
This PR adds the ability to use `block.timestamp` in custom AA contracts. AAs will still not have direct access to `block.timestamp`, but can utilize it via a proxy that enforces certain constraints. The PR introduces a `TimestampAsserter` contract that is deployed on every chain to the user space, similar to `Multicall3`. This contract has a single function, `assertTimestampInRange(start, end)`, which can be used by AAs at their discretion.

The `TimestampAsserter` contract ensures that `block.timestamp` falls within the specified `(start, end)` range. Additionally, the sequencer verifies that the `block.timestamp` is sufficiently far from the range’s end. This is to prevent DoS attacks where transactions pass validation but get stuck in the mempool during execution. This constraint is configurable and can be adjusted without requiring protocol update.

The PR also introduces two new fields to the `transactions` table: `timestamp_asserter_range_start` and `timestamp_asserter_range_end`. These fields are extracted during transaction execution in the sandbox by the `ValidationTracer`. If multiple assertions are made in a single transaction, the system captures the maximum of the starts and the minimum of the ends, resulting in the narrowest possible time range.

Transactions with time range constraints will undergo additional verification before being included in a block. If the current time falls outside the transaction’s specified time range, the transaction will be rejected with an appropriate message.

Sister PR in `era-contracts`: https://github.com/matter-labs/era-contracts/pull/843


core/node/state_keeper/src/io/mempool.rs
```diff
@@ -298,6 +298,31 @@ impl StateKeeperIO for MempoolIO {
                         .await?;
                     continue;
                 }
+
+                // Reject transactions that violate block.timestamp constraints. Such transactions should be
+                // rejected at the API level, but we need to protect ourselves in case if a transaction
+                // goes outside of the allowed range while being in the mempool
+                let now = SystemTime::now()
```
slowli, line 305: Same question as for the VM tracer: why is time acquired here instead of using one from the latest block?

core/node/state_keeper/src/io/mempool.rs
```diff
@@ -298,6 +298,31 @@ impl StateKeeperIO for MempoolIO {
                         .await?;
                     continue;
                 }
+
+                // Reject transactions that violate block.timestamp constraints. Such transactions should be
+                // rejected at the API level, but we need to protect ourselves in case if a transaction
+                // goes outside of the allowed range while being in the mempool
```
slowli, line 304: Dumb question: The PR description says that changing these params doesn't require a protocol upgrade. How will params work in a decentralized network? How will these values be coordinated among validators? (I'd imagine the timestamp asserter contract would actually contain these values in this case.) Or is this too far too ahead to plan?

core/node/api_server/src/tx_sender/tests/send_tx.rs
```diff
@@ -298,3 +308,75 @@ async fn sending_transaction_out_of_gas() {
     let (_, vm_result) = tx_sender.submit_tx(tx, block_args).await.unwrap();
     assert_matches!(vm_result.result, ExecutionResult::Revert { .. });
 }
+
+#[tokio::test]
+async fn submitting_tx_with_validation_traces() {
```
slowli, line 313: Bikeshedding: IMO, this test is mostly useless; it essentially tests that the mock tx executor works as expected. I'd advise to test the *real* executor (as most tests in this module do) with the changed custom account contract; so essentially, move the new integration tests here. Likewise, you can test gas estimation as well.

core/node/api_server/src/execution_sandbox/validate.rs
```diff
@@ -39,13 +39,16 @@ impl SandboxExecutor {
...
+    ) -> Result<ValidationTraces, ValidationError> {
         let total_latency = SANDBOX_METRICS.sandbox[&SandboxStage::ValidateInSandbox].start();
         let validation_params = get_validation_params(
             &mut connection,
             &tx,
             self.options.eth_call.validation_computational_gas_limit(),
             whitelisted_tokens_for_aa,
+            self.timestamp_asserter_address,
+            self.timestamp_asserter_min_range_sec,
+            self.timestamp_asserter_min_time_till_end_sec,
```
slowli, line 51: IIUC, during node synchronization it can still accept transaction requests (i.e., validate transactions). If this is the case, is the fact that validation can be called with `timestamp_asserter_address == Some(_)` before the asserter contract is deployed a concern?

core/lib/dal/src/transactions_dal.rs
```diff
@@ -347,7 +357,9 @@ impl TransactionsDal<'_, '_> {
                 execution_info,
                 received_at,
                 created_at,
-                updated_at
+                updated_at,
+                timestamp_asserter_range_start,
+                timestamp_asserter_range_end
```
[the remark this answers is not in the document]
ischasny: Good point - fixed

core/lib/dal/src/models/storage_transaction.rs
```diff
@@ -321,6 +324,15 @@ impl From<StorageTransaction> for Transaction {
...
 }
 
+impl From<StorageTransaction> for TransactionTimeRangeConstraint {
+    fn from(tx: StorageTransaction) -> Self {
+        Self {
+            timestamp_asserter_range_start: tx.timestamp_asserter_range_start,
+            timestamp_asserter_range_end: tx.timestamp_asserter_range_end,
+        }
+    }
+}
```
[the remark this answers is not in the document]
ischasny: Fixed

core/lib/types/src/lib.rs
```diff
@@ -416,3 +417,19 @@ impl Transaction {
...
+    Debug,
+    Default,
+    Eq,
+    PartialEq,
+    Hash,
+    Copy
+)]
+pub struct TransactionTimeRangeConstraint {
+    pub timestamp_asserter_range_start: Option<NaiveDateTime>,
+    pub timestamp_asserter_range_end: Option<NaiveDateTime>,
```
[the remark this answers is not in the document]
ischasny: Changed to `i64`

core/lib/multivm/src/tracers/validator/vm_latest/mod.rs
```diff
@@ -81,6 +86,56 @@ impl<H: HistoryMode> ValidationTracer<H> {
                             called_address,
                         ));
                     }
+                    // If this is a call to the timestamp asserter, extract the function arguments and store them in ValidationTraces.
+                    // These arguments are used by the mempool for transaction filtering. The call data length should be 68 bytes:
+                    // a 4-byte function selector followed by two U256 values.
+                    // TODO: if any more functions will be added to the TimestampAsserter contract in the future, verify the function selector too
```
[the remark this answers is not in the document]
ischasny: Good point - added

core/lib/multivm/src/tracers/validator/vm_latest/mod.rs
```diff
@@ -81,6 +86,56 @@ impl<H: HistoryMode> ValidationTracer<H> {
...
+                    if Some(called_address) == self.timestamp_asserter_address
+                        && far_call_abi.memory_quasi_fat_pointer.length == 68
```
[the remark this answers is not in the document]
ischasny: The reason I implemented it that way - was because that's the way it's done right above for [KECCAK](https://github.com/matter-labs/zksync-era/blob/main/core/lib/multivm/src/tracers/validator/vm_latest/mod.rs#L52). 

I tried doing in the way you suggested above - calldata gets red always as zeroes. Does any modificaiton need to be done to the snippet you provided?

core/lib/multivm/src/tracers/validator/vm_latest/mod.rs
```diff
@@ -81,6 +86,56 @@ impl<H: HistoryMode> ValidationTracer<H> {
...
+                        let start = U256::from_big_endian(
+                            &calldata[calldata.len() - 64..calldata.len() - 32],
+                        );
+                        let end = U256::from_big_endian(&calldata[calldata.len() - 32..]);
+
+                        if end.as_u32() - start.as_u32() < self.timestamp_asserter_min_range_sec {
+                            return Err(ViolatedValidationRule::TimestampAssertionShortRange);
+                        }
+
+                        let now = SystemTime::now()
```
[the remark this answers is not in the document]
ischasny: Good point - fixed

core/lib/multivm/src/tracers/validator/vm_latest/mod.rs
```diff
@@ -81,6 +86,56 @@ impl<H: HistoryMode> ValidationTracer<H> {
...
+                            .as_secs();
+
+                        if end.as_u32() < now as u32 + self.timestamp_asserter_min_time_till_end_sec
+                        {
+                            return Err(ViolatedValidationRule::TimestampAssertionCloseToRangeEnd);
+                        }
+
+                        {
+                            let mut traces_mut = self.traces.lock().unwrap();
+                            traces_mut.timestamp_asserter_range_start =
```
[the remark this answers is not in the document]
ischasny: Reworked it to be `Option<std::ops::RangeInclusive<_>>` - like.

core/lib/vm_interface/src/types/tracer.rs
```diff
@@ -57,6 +57,13 @@ pub struct ValidationParams {
     pub trusted_address_slots: HashSet<(Address, U256)>,
     /// Number of computational gas that validation step is allowed to use.
     pub computational_gas_limit: u32,
+    /// Address of the timestamp asserter. This contract is allowed to touch block.timestamp regardless
+    /// of the calling context.
+    pub timestamp_asserter_address: Option<Address>,
+    /// Minimum difference in seconds between the range start and range end
+    pub timestamp_asserter_min_range_sec: u32,
+    /// Minimum time between current block.timestamp and the end of the asserted range
+    pub timestamp_asserter_min_time_till_end_sec: u32,
```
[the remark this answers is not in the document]
ischasny: Good point - changed

core/node/api_server/src/execution_sandbox/execute.rs
```diff
@@ -100,13 +100,19 @@ pub(crate) struct SandboxExecutor {
     engine: SandboxExecutorEngine,
     pub(super) options: SandboxExecutorOptions,
     storage_caches: Option<PostgresStorageCaches>,
+    pub(super) timestamp_asserter_address: Option<Address>,
+    pub(super) timestamp_asserter_min_range_sec: u32,
+    pub(super) timestamp_asserter_min_time_till_end_sec: u32,
```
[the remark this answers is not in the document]
ischasny: Fixed

core/node/api_server/src/execution_sandbox/execute.rs
```diff
@@ -100,13 +100,19 @@ pub(crate) struct SandboxExecutor {
...
 }
 
 impl SandboxExecutor {
     pub fn real(
         options: SandboxExecutorOptions,
         caches: PostgresStorageCaches,
         missed_storage_invocation_limit: usize,
+        timestamp_asserter_address: Option<Address>,
+        timestamp_asserter_min_range_sec: u32,
+        timestamp_asserter_min_time_till_end_sec: u32,
```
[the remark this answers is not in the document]
ischasny: Grouped them up into a single struct - let me know if it looks better now.

core/node/state_keeper/src/io/mempool.rs
ischasny: We could in theory move these checks down to the `TimestampAsserter` contract and have those variables set there. @StanislavBreadless  would love your thoughts here.

StanislavBreadless: I think it is better not to do things on the protocol side that can be moved somewhere else. (In the worst case, consensus can be done offchain). 

The reason why communicating it offchain may make sense is because this restriction is on the API side only. I.e. if someone proposes a miniblock, the validators that actually check the block should not be affected by this restriction, i.e. if the validation of the custom account passes, the block is valid. Otherwise it is not. The restrictions here are just to protect the block builder and not the other validators

core/lib/multivm/src/tracers/validator/vm_latest/mod.rs
```diff
@@ -81,6 +83,50 @@ impl<H: HistoryMode> ValidationTracer<H> {
...
+                            let end = U256::from_big_endian(&calldata[calldata.len() - 32..]);
+
+                            let params = self.timestamp_asserter_params.as_ref().unwrap();
+
+                            if end.as_u32() - start.as_u32() < params.min_range_sec {
+                                return Err(ViolatedValidationRule::TimestampAssertionShortRange);
+                            }
+
+                            if end.as_u32()
+                                < self.l1_batch_env.timestamp as u32 + params.min_time_till_end_sec
```
slowli, line 116: I think it could be worth clarifying in a comment why using `l1_batch_env.timestamp` is OK here. AFAIU, it is because this tracer is always used in a oneshot execution mode (executing a single transaction).

core/lib/vm_interface/src/types/errors/halt.rs
```diff
@@ -42,6 +42,7 @@ pub enum Halt {
     VMPanic,
     TracerCustom(String),
     FailedToPublishCompressedBytecodes,
+    FailedBlockTimestampAssertion,
```
slowli, line 45: Future work (?): AFAICT, right now this error is only raised outside the VM, by `MempoolIO`. Is it possible to convert actual `require` failure in the timestamp asserter to this variant? Or would it require too much work?

core/lib/vm_interface/src/types/tracer.rs
```diff
@@ -104,6 +127,29 @@ pub enum ValidationError {
...
+pub struct ValidationTraces {
+    /// Represents a range from-to. Each field is a number of seconds since the epoch.
+    pub timestamp_asserter_range: Option<(i64, i64)>,
+}
+
+impl ValidationTraces {
+    /// Merges two ranges together by taking the maximum of the starts and the minimum of the ends
+    /// resulting into the narrowest possible time window
+    pub fn apply_range(&mut self, new_range: (i64, i64)) {
+        if let Some(mut range) = self.timestamp_asserter_range {
```
slowli, line 144: This copies the range, which is probably not what you want to do here. To modify the range in place, use `if let Some(range) = &mut self.timestamp_asserter_range`. I'd suggest to add a couple of unit tests to check that this method works as intended.

core/node/api_server/src/tx_sender/mod.rs
```diff
@@ -368,7 +390,7 @@ impl TxSender {
         let submission_res_handle = self
             .0
             .tx_sink
-            .submit_tx(&tx, execution_output.metrics)
+            .submit_tx(&tx, execution_output.metrics, validation_result.unwrap())
```
slowli, line 393: Please replace the `let Err(_) = validation_result` to make it clear that this is safe:

```rust
let validation_traces = validation_result?;
// ...
.submit_tx(&tx, execution_output.metrics, validation_traces)
```

core/node/api_server/src/tx_sender/mod.rs
```diff
@@ -234,6 +247,9 @@ pub struct TxSenderConfig {
     pub validation_computational_gas_limit: u32,
     pub chain_id: L2ChainId,
     pub whitelisted_tokens_for_aa: Vec<Address>,
+    pub timestamp_asserter_address: Option<Address>,
+    pub timestamp_asserter_min_range_sec: u32,
+    pub timestamp_asserter_min_time_till_end_sec: u32,
```
slowli, line 252: Why not use `Option<TimestampAsserterParams>` here? Do `min_range_sec` / `min_time_till_end_sec` params still make sense even if the asserter address isn't set?

core/node/state_keeper/Cargo.toml
```diff
@@ -31,6 +31,8 @@ zksync_node_test_utils.workspace = true
 zksync_vm_executor.workspace = true
 zksync_system_constants.workspace = true
 zksync_base_token_adjuster.workspace = true
+chrono = { workspace = true, features = ["serde"] }
+
```
slowli, line 35: This dependency seems unused as well.

core/node/state_keeper/src/io/mempool.rs
```diff
@@ -298,6 +299,23 @@ impl StateKeeperIO for MempoolIO {
...
+                let matches_range = constraint.timestamp_asserter_range.map_or(true, |x| {
+                    x.0 < l2_block_timestamp as i64 && x.1 > l2_block_timestamp as i64
```
slowli, line 307: Bikeshedding: IMO, a range with both exclusive ends is somewhat esoteric, even not taking Rust into account (there's no easy way to construct such ranges, unlike `x..y` / `x..=y`). If I'd read the `assertTimestampInRange` name, I'd assume that both range ends are inclusive. I understand that this is very subjective though.

core/node/state_keeper/src/io/tests/mod.rs
```diff
@@ -644,3 +666,121 @@ async fn insert_unsealed_batch_on_init(commitment_mode: L1BatchCommitmentMode) {
...
+    #[allow(deprecated)]
+    let expected_tx = tester.insert_tx(
+        &mut guard,
+        want_filter.fee_per_gas,
+        want_filter.gas_per_pubdata,
+        TransactionTimeRangeConstraint {
+            timestamp_asserter_range: Some((system_time - 1000, system_time + 1000)),
+        },
+    );
+    #[allow(deprecated)]
```
slowli, line 726: Dumb question: What's deprecated in this statement? It seems to work fine if the attribute is removed.

core/node/state_keeper/src/io/tests/mod.rs
```diff
@@ -644,3 +666,121 @@ async fn insert_unsealed_batch_on_init(commitment_mode: L1BatchCommitmentMode) {
...
+    );
+    insert_l2_transaction(&mut storage, &rejected_tx_1).await;
+    insert_l2_transaction(&mut storage, &expected_tx).await;
+    insert_l2_transaction(&mut storage, &rejected_tx_2).await;
+
+    let tx = mempool
+        .wait_for_next_tx(Duration::from_secs(2), system_time as u64)
+        .await
+        .expect("No expected transaction in the mempool")
+        .unwrap();
```
slowli, line 743: Nit: judging by `expect` messages, `expect` should be switched around with `unwrap()`; i.e., `unwrap` should be applied to a `Result`, and `expect` to the underlying `Option`. Ditto below.

core/tests/ts-integration/contracts/custom-account/custom-account.sol
```diff
@@ -9,6 +9,10 @@ import './SystemContractsCaller.sol';
 
 import './interfaces/IAccount.sol';
 
+interface ITimestampAsserter {
+	function assertTimestampInRange(uint256 start, uint256 end) external view;
```
slowli, line 13: Bikeshedding: I think (fortunately), tab indentation isn't usually used in Solidity. Would be somewhat to change in this case though.

core/tests/ts-integration/tests/custom-account.test.ts
```diff
@@ -95,6 +113,94 @@ describe('Tests for the custom account behavior', () => {
...
+            rangeStart,
+            rangeEnd
+        );
+
+        const promise = alice.provider.estimateGas({
+            ...(await erc20.transfer.populateTransaction(alice.address, TRANSFER_AMOUNT)),
+            from: await customAccount.getAddress()
+        });
+
+        // requires custom assertion because the transaction would fail in "estimateFee" that is not handled correctly by zksync ethers
```
slowli, line 187: Dumb question: Would it be possible to use another RPC method then? (E.g., submit a transaction w/o estimating gas for it.)

core/tests/ts-integration/tests/custom-account.test.ts
```diff
@@ -95,6 +113,94 @@ describe('Tests for the custom account behavior', () => {
         ).toBeAccepted([erc20BalanceChange, feeCheck]);
     });
 
+    test('Should fail transaction validation due to timestamp assertion in the validation tracer - short time window', async () => {
+        const now = Math.floor(Date.now() / 1000);
+        const rangeStart = now - 60;
+        const rangeEnd = now + 60;
```
slowli, line 119: How do we know that this range is too short? Maybe, it would make sense to read node config and then create a shorter range? Ditto for the next test.

core/tests/ts-integration/tests/custom-account.test.ts
```diff
@@ -95,6 +113,94 @@ describe('Tests for the custom account behavior', () => {
...
+            from: await customAccount.getAddress()
+        });
+
+        // requires custom assertion because the transaction would fail in "estimateFee" that is not handled correctly by zksync ethers
+        promise
+            .then((_) => {
+                fail('The transaction was expected to fail due to timestamp assertion');
+            })
+            .catch((error) => {
+                // If it's an error object with the structure you're mentioning, check its properties
```
slowli, line 193: Nit: Who's "you"? 🙃

core/tests/ts-integration/tests/custom-account.test.ts
```diff
@@ -95,6 +113,94 @@ describe('Tests for the custom account behavior', () => {
...
+                if (error.info && error.info.error && error.info.error.message) {
+                    expect(error.info.error.message).toEqual(
+                        'failed to validate the transaction. reason: Timestamp is out of range'
+                    );
+                } else {
+                    fail('The transaction was expected to fail with a different message');
```
slowli, line 199: Please use `expect(null).fail(..)` as in other integration tests.

core/tests/ts-integration/tests/custom-account.test.ts
```diff
@@ -40,7 +45,18 @@ describe('Tests for the custom account behavior', () => {
 
     test('Should deploy custom account', async () => {
         const violateRules = false;
-        customAccount = await deployContract(alice, contracts.customAccount, [violateRules], 'createAccount');
+        customAccount = await deployContract(
+            alice,
+            contracts.customAccount,
+            // 2555971200 is a number of seconds up to 30/12/2050
```
slowli, line 51: Nit: This comment should be moved to the `DEFAULT_TIMESTAMP_ASSERTER_RANGE_END` declaration.

core/lib/config/src/configs/chain.rs
```diff
@@ -244,3 +244,11 @@ impl MempoolConfig {
         Duration::from_millis(self.delay_interval)
     }
 }
+
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Default)]
+pub struct TimestampAsserterConfig {
```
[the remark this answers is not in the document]
ischasny: I think it's a general pattern if you look into how configs are structured across the board. Agree that it'd make sense to enforce reasonable defaults everywhere if such exist, but if I do this for one specific config it'd rather become a weird outlier.

core/lib/protobuf_config/src/timestamp_asserter.rs
```diff
@@ -0,0 +1,19 @@
+use zksync_config::configs::chain::TimestampAsserterConfig;
+use zksync_protobuf::{required, ProtoRepr};
+
+impl ProtoRepr for crate::proto::config::timestamp_asserter::TimestampAsserter {
+    type Type = TimestampAsserterConfig;
+    fn read(&self) -> anyhow::Result<Self::Type> {
+        Ok(Self::Type {
+            min_range_sec: *required(&self.min_range_sec).unwrap_or(&0),
+            min_time_till_end_sec: *required(&self.min_time_till_end_sec).unwrap_or(&0),
```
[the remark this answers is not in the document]
ischasny: Fixed - errors are propagated back up now.

core/lib/types/src/lib.rs
```diff
@@ -416,3 +416,19 @@ impl Transaction {
...
+    Debug,
+    Default,
+    Eq,
+    PartialEq,
+    Hash,
+    Copy
+)]
+pub struct TransactionTimeRangeConstraint {
+    /// Represents a range from-to. Each field is a number of seconds since the epoch.
+    pub timestamp_asserter_range: Option<(i64, i64)>,
```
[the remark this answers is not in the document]
ischasny: Changed to Range<u64> for both;

core/lib/dal/migrations/20241008073940_add_block_timestamp_asserter.down.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE transactions DROP COLUMN timestamp_asserter_range_start;
+ALTER TABLE transactions DROP COLUMN timestamp_asserter_range_end;
```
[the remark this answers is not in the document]
ischasny: Fixed

core/lib/dal/src/transactions_dal.rs
```diff
@@ -1819,8 +1837,14 @@ impl TransactionsDal<'_, '_> {
         .fetch_all(self.storage)
         .await?;
 
-        let transactions = transactions.into_iter().map(|tx| tx.into()).collect();
-        Ok(transactions)
+        let transactions_with_constraints = transactions
+            .into_iter()
+            .map(|tx| {
+                let constraint: TransactionTimeRangeConstraint = (&tx).into();
```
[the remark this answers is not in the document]
ischasny: Applied

core/lib/mempool/Cargo.toml
```diff
@@ -13,3 +13,5 @@ categories.workspace = true
 [dependencies]
 zksync_types.workspace = true
 tracing.workspace = true
+chrono = { workspace = true, features = ["serde"] }
+
```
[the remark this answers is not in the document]
ischasny: Good spot - removed

core/lib/multivm/src/tracers/validator/vm_latest/mod.rs
```diff
@@ -81,6 +83,50 @@ impl<H: HistoryMode> ValidationTracer<H> {
...
+
+                        if calldata[..4] == TIMESTAMP_ASSERTER_FUNCTION_SELECTOR {
+                            let start = U256::from_big_endian(
+                                &calldata[calldata.len() - 64..calldata.len() - 32],
+                            );
+                            let end = U256::from_big_endian(&calldata[calldata.len() - 32..]);
+
+                            let params = self.timestamp_asserter_params.as_ref().unwrap();
+
+                            if end.as_u32() - start.as_u32() < params.min_range_sec {
```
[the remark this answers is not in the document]
ischasny: * `end` is guaranteed to be greater than `start`, this is enforced in the contract, i.e.
```
        require(start < block.timestamp && end > block.timestamp, "Timestamp is out of range");
```
* added an extra bit of logic to avoid overflows - values are capped at `u64::MAX` / `u32::MAX`

core/lib/multivm/src/tracers/validator/vm_latest/mod.rs
ischasny: indeed that's the case - added clarification

core/lib/multivm/src/tracers/validator/vm_latest/mod.rs
```diff
@@ -81,6 +83,50 @@ impl<H: HistoryMode> ValidationTracer<H> {
                             called_address,
                         ));
                     }
+                    // If this is a call to the timestamp asserter, extract the function arguments and store them in ValidationTraces.
+                    // These arguments are used by the mempool for transaction filtering. The call data length should be 68 bytes:
+                    // a 4-byte function selector followed by two U256 values.
+                    let address = self.timestamp_asserter_params.as_ref().map(|x| x.address);
+                    if Some(called_address) == address
```
joonazan, line 90: I'd write something like`if self.timestamp_asserter_params.map(|x| x.address == called_address).unwrap_or_default()`

core/lib/vm_interface/src/types/tracer.rs
ischasny: Great point - changed and added tests

core/lib/multivm/src/tracers/validator/vm_latest/mod.rs
joonazan: Actually, to avoid the `.unwrap()` later on, you should do `if let Some(address) = ...` here.

core/node/state_keeper/Cargo.toml
ischasny: Removed

core/node/api_server/src/tx_sender/mod.rs
ischasny: Makes sense - changed

core/node/state_keeper/src/io/mempool.rs
ischasny: Refined the logic with `range.contains` instead

core/node/state_keeper/src/io/tests/mod.rs
ischasny: Thanks, good spot - it's a leftover from previous iterations where I used `NaiveDateTime` type

core/node/state_keeper/src/io/tests/mod.rs
ischasny: Fixed for all, thanks

core/tests/ts-integration/contracts/custom-account/custom-account.sol
ischasny: Do you mean changing TAB to 4 spaces?

core/lib/types/src/lib.rs
```diff
@@ -416,3 +416,19 @@ impl Transaction {
         })
     }
 }
+
+#[derive(
+    Clone,
+    Serialize,
+    Deserialize,
```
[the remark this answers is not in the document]
ischasny: Indeed - removed.

core/tests/ts-integration/tests/custom-account.test.ts
ischasny: You'd still have to fill all required fields though no? Even if you call 'sendRawTransaction` without estimating fee first.

core/tests/ts-integration/tests/custom-account.test.ts
ischasny: Fixed

core/lib/vm_interface/src/types/errors/halt.rs
ischasny: Thanks. Created a ticket for future investigation.

joonazan:
I believe that the "time range is wide enough" check does nothing and should be removed. Suppose you want the time range [1, 2]. You can get it by requiring two ranges: [-500, 2] and [1, 500].

The check that the range's end is far enough in the future is sufficient to prevent ultra-short ranges, I think.

ischasny:
> if Some(called_address) == address

Great point thanks. Indeed we don't need to check how wide the range is. Just making sure that:
* `block.timestamp` is within the specified range (done by the contract);
* and that the end is enough in the future 
should be sufficient.

Removed the check and updated the PR description.

core/node/state_keeper/src/io/tests/mod.rs
```diff
@@ -644,3 +666,118 @@ async fn insert_unsealed_batch_on_init(commitment_mode: L1BatchCommitmentMode) {
     assert_eq!(l1_batch_params.fee_input, fee_input);
     assert_eq!(l1_batch_params.first_l2_block.timestamp, 2);
 }
+
+#[test_casing(2, COMMITMENT_MODES)]
+#[tokio::test]
+async fn test_mempool_with_timestamp_assertion(commitment_mode: L1BatchCommitmentMode) {
+    let connection_pool = ConnectionPool::<Core>::constrained_test_pool(2).await;
+    let tester = Tester::new(commitment_mode);
```
slowli, line 674: Nit: I don't think the commitment mode matters here, so it would be sufficient to set it to any single value.

core/node/state_keeper/src/mempool_actor.rs
```diff
@@ -122,16 +123,22 @@ impl MempoolFetcher {
                 )
                 .await
                 .context("failed syncing mempool")?;
+
+            let transactions = transactions_with_constraints
+                .iter()
+                .map(|(t, _c)| t)
+                .collect_vec();
```
slowli, line 130: Bikeshedding: It's easy to avoid dependency on `Itertools` by using `collect` with a type annotation or a turbofish.

core/node/state_keeper/src/io/mod.rs
```diff
@@ -137,8 +137,11 @@ pub trait StateKeeperIO: 'static + Send + Sync + fmt::Debug + IoSealCriteria {
 
     /// Blocks for up to `max_wait` until the next transaction is available for execution.
     /// Returns `None` if no transaction became available until the timeout.
-    async fn wait_for_next_tx(&mut self, max_wait: Duration)
-        -> anyhow::Result<Option<Transaction>>;
+    async fn wait_for_next_tx(
+        &mut self,
+        max_wait: Duration,
+        l2_block_timestamp: u64,
```
slowli, line 143: If it's guaranteed that `wait_for_new_batch_params` / `wait_for_new_l2_block_params` output is used for L2 blocks, the L2 block timestamp can be encapsulated inside `MempoolIO`, and this method can be left unchanged.

core/tests/ts-integration/tests/custom-account.test.ts
```diff
@@ -95,6 +112,68 @@ describe('Tests for the custom account behavior', () => {
...
+
+        await expect(
+            sendCustomAccountTransaction(
+                tx as zksync.types.Transaction,
+                alice.provider,
+                await customAccount.getAddress(),
+                testMaster.environment().l2ChainId
+            )
+        ).toBeRejected(
+            'failed to validate the transaction. reason: Violated validation rules: block.timestamp is too close to the range end'
```
slowli, line 139: Nit: Would match via a regex work here (like `/block\.timestamp/`)? It's usually a bad idea to tie tests to human-readable error messages.

core/tests/ts-integration/tests/custom-account.test.ts
slowli: Yes, but I thought that in this case it would be possible to simplify error assertions below similarly to other tests.

zkstack_cli/crates/zkstack/completion/_zkstack.zsh
slowli, line 1: Dumb question: What are changes in this file caused by? It doesn't look like this PR changes `zkstack` command-line args.

core/lib/dal/src/transactions_dal.rs
```diff
@@ -346,6 +357,8 @@ impl TransactionsDal<'_, '_> {
                 paymaster_input,
                 execution_info,
                 received_at,
+                timestamp_asserter_range_start,
+                timestamp_asserter_range_end,
```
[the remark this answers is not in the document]
ischasny: There was a discussion about that in the internal notion - @RomanBrodetski agreed to keep it as separate fields.

core/node/api_server/src/tx_sender/mod.rs
```diff
@@ -234,6 +244,14 @@ pub struct TxSenderConfig {
     pub validation_computational_gas_limit: u32,
     pub chain_id: L2ChainId,
     pub whitelisted_tokens_for_aa: Vec<Address>,
+    pub timestamp_asserter_params: Option<TimestampAsserterParams>,
+}
+
+#[derive(Debug, Clone)]
+pub struct TimestampAsserterParams {
```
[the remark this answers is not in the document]
ischasny: Because it won't work in the external node as it's not aware of that crate.

zkstack_cli/crates/zkstack/completion/_zkstack.zsh
ischasny: Not sure - the PR started failing lint check in the CI - maybe job config has changed

core/bin/external_node/src/config/mod.rs
```diff
@@ -685,6 +698,16 @@ impl OptionalENConfig {
             contracts_diamond_proxy_addr: None,
             gateway_url: enconfig.gateway_url.clone(),
             bridge_addresses_refresh_interval_sec: enconfig.bridge_addresses_refresh_interval_sec,
+            timestamp_asserter_min_range_sec: general_config
+                .timestamp_asserter_config
+                .as_ref()
+                .map(|x| x.min_range_sec)
+                .unwrap_or(0),
```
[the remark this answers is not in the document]
ischasny: Changed to use more reasonable default

core/lib/multivm/src/tracers/validator/mod.rs
```diff
@@ -60,8 +67,14 @@ impl<H> ValidationTracer<H> {
     pub fn new(
         params: ValidationParams,
         vm_version: VmVersion,
-    ) -> (Self, Arc<OnceCell<ViolatedValidationRule>>) {
+        l1_batch_env: L1BatchEnv,
+    ) -> (
+        Self,
+        Arc<OnceCell<ViolatedValidationRule>>,
+        Arc<Mutex<ValidationTraces>>,
```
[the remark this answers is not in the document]
ischasny: Added

core/lib/vm_interface/src/types/tracer.rs
```diff
@@ -104,6 +127,30 @@ pub enum ValidationError {
...
+    pub timestamp_asserter_range: Option<Range<u64>>,
+}
+
+impl ValidationTraces {
+    /// Merges two ranges by selecting the maximum of the start values and the minimum of the end values,
+    /// producing the narrowest possible time window. Note that overlapping ranges are essential;
+    /// a lack of overlap would have triggered an assertion failure in the `TimestampAsserter` contract,
+    /// as `block.timestamp` cannot satisfy two non-overlapping ranges.
+    pub fn apply_range(&mut self, new_range: Range<u64>) {
+        if let Some(ref mut range) = self.timestamp_asserter_range.as_mut() {
```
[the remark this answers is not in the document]
ischasny: Changed

core/lib/vm_interface/src/types/tracer.rs
```diff
@@ -116,3 +163,45 @@ impl fmt::Display for ValidationError {
...
+    use std::ops::Range;
+
+    use super::*;
+
+    #[test]
+    fn test_apply_range_when_none() {
+        let mut validation_traces = ValidationTraces {
+            timestamp_asserter_range: None,
+        };
+        let new_range = Range { start: 10, end: 20 };
```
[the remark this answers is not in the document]
ischasny: changed

core/node/state_keeper/src/mempool_actor.rs
ischasny: changed

core/tests/ts-integration/tests/custom-account.test.ts
ischasny: Agreed - this is rather a pattern that is used across all integration tests. Ideally i'd match against a defined set of error codes.

core/node/state_keeper/src/io/mod.rs
ischasny: I chose to change the interface because `UpdatesManager` gets initialised (outside)[https://github.com/matter-labs/zksync-era/blob/main/core/node/state_keeper/src/keeper.rs#L151] of the wiring framework. Getting it encapsulated into the `MempoolIO` would require larger refactoring.

core/lib/dal/src/models/storage_transaction.rs
```diff
@@ -321,6 +324,30 @@ impl From<StorageTransaction> for Transaction {
...
 }
 
+impl From<&StorageTransaction> for TransactionTimeRangeConstraint {
+    fn from(tx: &StorageTransaction) -> Self {
+        if tx.timestamp_asserter_range_start.is_some() && tx.timestamp_asserter_range_end.is_some()
+        {
+            Self {
+                timestamp_asserter_range: Some(
+                    tx.timestamp_asserter_range_start
+                        .unwrap()
```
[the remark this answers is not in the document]
ischasny: Simplified it in a bit different way

core/lib/dal/src/transactions_dal.rs
```diff
@@ -2212,6 +2237,29 @@ impl TransactionsDal<'_, '_> {
         .fetch_optional(self.storage)
         .await
     }
+
+    pub async fn get_storage_tx_by_hash(
```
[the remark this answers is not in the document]
ischasny: I agree - I wouldn't be asserting a error message if we've had some standardised failure reason.

core/lib/vm_interface/src/types/tracer.rs
```diff
@@ -57,6 +57,19 @@ pub struct ValidationParams {
...
+
+#[derive(Debug, Clone)]
+pub struct TimestampAsserterParams {
+    /// Address of the timestamp asserter. This contract is allowed to touch block.timestamp regardless
+    /// of the calling context.
+    pub address: Address,
+    /// Minimum difference in seconds between the range start and range end
+    pub min_range_sec: u32,
+    /// Minimum time between current block.timestamp and the end of the asserted range
+    pub min_time_till_end_sec: u32,
```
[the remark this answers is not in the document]
ischasny: Good point - changed

core/bin/zksync_server/src/node_builder.rs
```diff
@@ -322,6 +323,13 @@ impl MainNodeBuilder {
                     .fee_account
                     .address(),
                 self.genesis_config.l2_chain_id,
+                self.contracts_config
+                    .l2_timestamp_asserter_addr
+                    .map(|address| TimestampAsserterParams {
+                        address,
+                        min_range_sec: timestamp_asserter_config.min_range_sec,
+                        min_time_till_end_sec: timestamp_asserter_config.min_time_till_end_sec,
```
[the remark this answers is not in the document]
ischasny: I'd be ok with leaving it as is, but also happy to change it in a follow up PR if you feel strong about it.

core/tests/ts-integration/tests/custom-account.test.ts
ischasny: Very good point - changed the test

zkstack_cli/crates/zkstack/completion/_zkstack.zsh
ischasny: FYI - tagged you in relevant slack thread - probably an unformatted file has made it's way into the main somehow.

core/lib/dal/src/transactions_dal.rs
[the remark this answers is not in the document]
slowli: Can (IMO should) be fixed in a follow-up PR.

core/lib/protobuf_config/src/timestamp_asserter.rs
```diff
@@ -0,0 +1,19 @@
+use anyhow::Context;
+use zksync_config::configs::chain::TimestampAsserterConfig;
+use zksync_protobuf::{required, ProtoRepr};
+
+impl ProtoRepr for crate::proto::config::timestamp_asserter::TimestampAsserter {
+    type Type = TimestampAsserterConfig;
+    fn read(&self) -> anyhow::Result<Self::Type> {
+        Ok(Self::Type {
+            min_time_till_end_sec: *required(&self.min_time_till_end_sec)
```
slowli, line 9: Don't forget to add this config param to stage / other auto-deployed envs; IIUC, they'll get broken otherwise.

slowli (approved):

core/bin/external_node/src/config/mod.rs
```diff
@@ -146,6 +147,10 @@ impl RemoteENConfig {
             .get_main_contract()
             .rpc_context("get_main_contract")
             .await?;
+        let timestamp_asserter_address = client
```
slowli, line 150: Is there a risk that the main node won't have this method implemented (e.g., on the mainnet)? If so, it makes sense to handle `MethodNotFound` errors as it's done below.

core/bin/external_node/src/config/mod.rs
```diff
@@ -1447,6 +1467,14 @@ impl From<&ExternalNodeConfig> for TxSenderConfig {
             chain_id: config.required.l2_chain_id,
             // Does not matter for EN.
             whitelisted_tokens_for_aa: Default::default(),
+            timestamp_asserter_params: config.remote.l2_timestamp_asserter_addr.map(|address| {
+                TimestampAsserterParams {
+                    address,
+                    min_time_till_end: Duration::from_secs(
+                        config.optional.timestamp_asserter_min_time_till_end_sec as u64,
```
slowli, line 1474: Nit: Since `u32` -> `u64` conversion is lossless, you may use `timestamp_asserter_min_time_till_end_sec.into()` to signal it. Or change the field type to `u64`.

core/lib/dal/src/transactions_dal.rs
```diff
@@ -2212,6 +2238,29 @@ impl TransactionsDal<'_, '_> {
...
+                *
+            FROM
+                transactions
+            WHERE
+                hash = $1
+            "#,
+            hash.as_bytes()
+        )
+        .map(Into::into)
+        .instrument("get_tx_error_by_hash")
```
slowli, line 2259: Nit: Mismatch between the method name / semantics and the name here. As we've discussed, I'd prefer if the method actually just loaded the error btw 🙃

core/lib/multivm/src/tracers/validator/vm_latest/mod.rs
```diff
@@ -81,6 +83,52 @@ impl<H: HistoryMode> ValidationTracer<H> {
...
+                                {
+                                    return Err(
+                                        ViolatedValidationRule::TimestampAssertionCloseToRangeEnd,
+                                    );
+                                }
+
+                                {
+                                    let mut traces_mut = self.traces.lock().unwrap();
+                                    traces_mut.apply_range(start..end);
+                                }
```
slowli, line 128: Nit: Since the outer scope ends here anyway, it's not necessary to create an additional scope for the mutex guard. Also, since it's used for a single operation, it's possible to apply it directly w/o declaring a local var.

core/lib/vm_interface/src/types/tracer.rs
```diff
@@ -104,6 +120,30 @@ pub enum ValidationError {
...
+pub struct ValidationTraces {
+    pub timestamp_asserter_range: Option<Range<u64>>,
+}
+
+impl ValidationTraces {
+    /// Merges two ranges by selecting the maximum of the start values and the minimum of the end values,
+    /// producing the narrowest possible time window. Note that overlapping ranges are essential;
+    /// a lack of overlap would have triggered an assertion failure in the `TimestampAsserter` contract,
+    /// as `block.timestamp` cannot satisfy two non-overlapping ranges.
+    pub fn apply_range(&mut self, new_range: Range<u64>) {
```
slowli, line 137: Bikeshedding: If we plan to add new fields to `ValidationTraces` in the future, this method name / docs look too ambiguous. It would make sense to call it like `(apply|narrow)_timestamp_asserter_range`.

core/node/api_server/src/tx_sender/mod.rs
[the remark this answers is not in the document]
slowli: It can be re-exported from the API server (the simplest option, probably). Conceptually, there would be no overhead adding a direct dependency for the EN either because it is included in its dependency graph anyway.

core/tests/ts-integration/tests/custom-account.test.ts
```diff
@@ -192,19 +192,41 @@ describe('Tests for the custom account behavior', () => {
...
+        } catch (e) {
+            const err = e as Error;
+            expect(err.message).toContain(
+                'failed to validate the transaction. reason: Validation revert: Account validation error'
+            );
+            const functionSelectorMatch = err.message.match(/function_selector\s=\s(0x[0-9a-fA-F]{8})/);
+            const calldataMatch = err.message.match(/data\s=\s(0x[0-9a-fA-F]+)/);
+
+            if (!functionSelectorMatch || !calldataMatch) {
+                expect(null).fail('Function selector or calldata not found in the error message');
```
slowli, line 215: Nit: You can use [`toBeTruthy`](https://jestjs.io/docs/expect#tobetruthy) matcher here to be more idiomatic. Also, `!` below would fulfill the same goal (albeit without an error message).

core/tests/ts-integration/tests/custom-account.test.ts
```diff
@@ -192,19 +192,41 @@ describe('Tests for the custom account behavior', () => {
...
-        );
+            );
+            expect(null).fail('The transaction was expected to fail');
+        } catch (e) {
+            const err = e as Error;
+            expect(err.message).toContain(
+                'failed to validate the transaction. reason: Validation revert: Account validation error'
+            );
+            const functionSelectorMatch = err.message.match(/function_selector\s=\s(0x[0-9a-fA-F]{8})/);
+            const calldataMatch = err.message.match(/data\s=\s(0x[0-9a-fA-F]+)/);
```
slowli, line 212: Off-topic, probably: Parsing function selector / calldata from the human-readable message looks weird. Is this what Web3 clients expect? I'd imagine an optimal implementation would define separate fields in the JSON-RPC response.

core/tests/ts-integration/tests/custom-account.test.ts
```diff
@@ -192,19 +192,41 @@ describe('Tests for the custom account behavior', () => {
...
+
+            const functionSelector = functionSelectorMatch![1];
+            const calldata = calldataMatch![1];
+
+            const startHex = calldata.slice(74, 138);
+            const endHex = calldata.slice(138);
+            const start = BigInt(`0x${startHex}`);
+            const end = BigInt(`0x${endHex}`);
+
+            expect(functionSelector).toBe('0x3d5740d9');
```
slowli, line 226: Nit: I'd move this next to the `functionSelector` declaration; if this assertion doesn't hold, parsing `start` / `end` may fail leading to hard-to-attribute errors.

slowli (commented):
Other than a potential error getting remote EN config, looks fine.

core/bin/external_node/src/config/mod.rs
ischasny: Agreed and fixed

core/tests/ts-integration/tests/custom-account.test.ts
ischasny: Agreed - I'd prefer custom errors to be returned as a part of JSON RPC response.

core/bin/external_node/src/config/mod.rs
```diff
@@ -227,10 +227,36 @@ impl RemoteENConfig {
...
+async fn handle_rpc_response_with_fallback<T, F>(
+    rpc_call: F,
+    fallback: Option<T>,
+    context: String,
+) -> anyhow::Result<T>
+where
+    F: Future<Output = Result<T, ClientError>>,
+    T: Clone,
+{
+    match rpc_call.await {
```
slowli, line 244: Nit: You can match the call result directly instead of `await`ing for it.

core/bin/external_node/src/config/mod.rs
```diff
@@ -227,10 +227,36 @@ impl RemoteENConfig {
...
+        Err(ClientError::Call(err))
+            if [
+                ErrorCode::MethodNotFound.code(),
+                // This what `Web3Error::NotImplemented` gets
+                // `casted` into in the `api` server.
+                ErrorCode::InternalError.code(),
+            ]
+            .contains(&(err.code())) =>
+        {
+            fallback.ok_or_else(|| anyhow::anyhow!("RPC method not found or internal error"))
```
slowli, line 254: I don't understand how `fallback == None` works; as it's implemented currently, it's not a fallback at all, and a node executing this code will still break down here unless the main node is updated. Isn't `None` a valid value for the `timestamp_asserter_address`? If so, it looks like `fallback` should have type `T` and it should be returned here unconditionally.

core/bin/zksync_server/src/node_builder.rs
```diff
@@ -303,6 +305,20 @@ impl MainNodeBuilder {
...
         let rpc_config = try_load_config!(self.configs.api_config).web3_json_rpc;
+
+        let timestamp_asserter_params = match self.contracts_config.l2_timestamp_asserter_addr {
+            Some(address) => {
+                let timestamp_asserter_config =
+                    try_load_config!(self.configs.timestamp_asserter_config);
+                Some(TimestampAsserterParams {
+                    address,
+                    min_time_till_end: Duration::from_secs(
+                        timestamp_asserter_config.min_time_till_end_sec as u64,
```
slowli, line 316: Nit: This lossless conversion can be replaced with `_.into()` as well.

core/bin/external_node/src/config/mod.rs
ischasny: Good point - fixed

slowli (approved):

artmakh (dismissed):

.github/workflows/new-build-core-template.yml
```diff
@@ -208,6 +211,14 @@ jobs:
           path: |
             ./contracts
 
+      # TODO Remove it when we migrate to foundry inside contracts repository
+      - name: Create necessary artefacts
+        shell: bash
+        run: |
+          mkdir -p contracts/l1-contracts/artifacts/
```
yorik, line 218: nit: you can create all the directories with single command:
`mkdir -p contracts/l1-contracts/artifacts contracts/l1-contracts/out`

yorik (dismissed):
Approve for workflows.

perekopskiy (approved):

PR #3033, opened 2024-10-08
feat(api): Integrate new VM into API server (no tracers)

slowli:
## What ❔

Integrates the new VM into API server for 3 use cases: gas estimation, calls, and tx execution (without the validation stage).

## Why ❔

These use cases do not require tracers and could benefit from faster VM execution (particularly gas estimation, which runs the VM multiple times).

## Checklist


core/lib/multivm/src/versions/vm_fast/vm.rs
```diff
@@ -205,7 +205,22 @@ impl<S: ReadStorage, Tr: Tracer> Vm<S, Tr> {
...
+                            let tx_has_failed = self.get_tx_result().is_zero();
+                            if tx_has_failed {
+                                let output = VmRevertReason::General {
+                                    msg: "Transaction reverted with empty reason. Possibly out of gas"
+                                        .to_string(),
+                                    data: vec![],
+                                };
+                                ExecutionResult::Revert { output }
+                            } else {
+                                ExecutionResult::Success { output: vec![] }
```
joonazan, line 220: Can there be success without getting a result?

core/lib/vm_executor/src/oneshot/mod.rs
```diff
@@ -49,10 +53,52 @@ mod contracts;
...
+mod tests;
+
+/// Fast VM modes utilized for different kinds of operations.
+#[derive(Debug, Clone, Copy, Default)]
+pub struct OneshotExecutorVmModes {
+    /// Mode used for gas estimation.
+    pub gas_estimation: FastVmMode,
+    /// Mode used for `eth_call`s.
+    pub call: FastVmMode,
+    /// Mode used for transaction execution (doesn't include the validation step).
```
joonazan, line 66: Please also add the validation step so I can work on it.

slowli: Validation will be covered by this param once it's supported by the fast VM; I'll make the doc comment clearer ("doesn't include the validation step for now since it's not supported by the fast VM").

core/lib/multivm/src/versions/vm_fast/vm.rs
slowli: I'm not sure. I've reproduced functionality from the results tracer for the latest VM:

https://github.com/matter-labs/zksync-era/blob/3fd2fb14e7283c6858731e162522e70051a8e162/core/lib/multivm/src/versions/vm_latest/tracers/result_tracer.rs#L298-L312

core/lib/vm_executor/src/oneshot/mod.rs
```diff
@@ -182,25 +254,89 @@ where
...
+            .push(StorageInvocations::new(missed_storage_invocation_limit).into_tracer_pointer());
+        tracers.into()
+    }
+}
+
 #[derive(Debug)]
 struct VmSandbox<S: ReadStorage> {
-    vm: Box<LegacyVmInstance<S, HistoryDisabled>>,
-    storage_view: StoragePtr<StorageView<S>>,
+    vm: Vm<StorageWithOverrides<S>>,
```
joonazan, line 321: I think the VM can built when something is executed, so it doesn't need to live in this struct.

core/lib/multivm/src/versions/vm_fast/vm.rs
joonazan: Looks like the else would be reached if the bootloader runs out of gas after a successful transaction. Apparently that is considered a success.

core/lib/vm_executor/src/oneshot/mod.rs
joonazan: A bit of a chicken and egg problem 😄 I guess I will add that option in my PR implementing validation then.

slowli:
With optimized gas estimation and new VM, the CI load test [yields](https://github.com/matter-labs/zksync-era/actions/runs/11237161477/job/31238987753?pr=3033#step:8:855) around 115 tps (vs 70 tps baseline and [approx. 60 tps for the old VM](https://github.com/matter-labs/zksync-era/actions/runs/11237161477/job/31238987004?pr=3033#step:8:489)).

core/lib/config/src/configs/experimental.rs
```diff
@@ -106,4 +106,8 @@ pub struct ExperimentalVmConfig {
     /// the new VM doesn't produce call traces and can diverge from the old VM!
     #[serde(default)]
     pub state_keeper_fast_vm_mode: FastVmMode,
+
+    /// Fast VM mode to use for gas estimation in the API server.
+    #[serde(default)]
+    pub api_fast_vm_mode_for_gas_estimation: FastVmMode,
```
slowli, line 112: I haven't added corresponding option for calls and transaction execution; they can be added iteratively. Not sure what's the best approach here.

etc/env/file_based/overrides/tests/loadtest-new.yaml
slowli, line 1: @Deniallugo What's the best way to set `api_fast_vm_mode_for_gas_estimation = SHADOW` for integration tests? Create a similar file and call `zks config-writer` like in the CI load test?

core/lib/vm_executor/src/oneshot/mod.rs
```diff
@@ -275,30 +373,90 @@ impl<S: ReadStorage> VmSandbox<S> {
...
+                storage_view.clone(),
+            )),
+            FastVmMode::Shadow => {
+                let mut vm =
+                    ShadowVm::new(self.env.l1_batch, self.env.system, storage_view.clone());
+                if !self.panic_on_divergence {
+                    let transaction = format!("{:?}", transaction);
+                    let handler = DivergenceHandler::new(move |errors, _| {
+                        tracing::error!(transaction, ?mode, "{errors}");
+                    });
```
slowli, line 425: Not sure what the best option for the divergence handler would look like. We probably don't want to panic on divergence, but maybe we want to store a VM dump somewhere?

core/lib/config/src/configs/experimental.rs
perekopskiy: Probably one variable for all 3 cases is enough?

core/lib/vm_executor/src/oneshot/mod.rs
perekopskiy: It does, maybe we can just dump it to local file with some safety measures against spamming files, e.g. limit it to 10 dump files and overwrite/stop dumping when limit is reached?

etc/env/file_based/overrides/tests/loadtest-new.yaml
Deniallugo: yes, config-writer is the best effort. 
You can create this file on flight

core/lib/vm_executor/src/oneshot/mod.rs
slowli: Will we be able to get these files from e.g. mainnet pods (where divergences are empirically most likely to happen)? AFAIU, this would require K8s privileges mere mortals don't have (like executing commands in a pod).

core/lib/config/src/configs/experimental.rs
slowli: Do you think it should be used for all 3 cases from the start? Or should it be used for gas estimation rn, and then extended to other 2 cases? cc @RomanBrodetski

perekopskiy: IMO it's fine to use it for all 3. If we encounter problems with any of them, we can temporarily disable the new VM mode. But if you have any reasons to keep it only for gas estimation for now, I'm more than fine with keeping it

slowli: No particular reason; I'll simplify configuration then.

perekopskiy (dismissed):

perekopskiy (approved):

artmakh (approved):

PR #3035, opened 2024-10-08
test: Brush up test contracts

slowli:
## What ❔

Extracts test contracts to a separate crate with a reasonable build pipeline.

## Why ❔

For now, test contracts are distributed across multiple crates (e.g., loaded using hardcoded paths in the workspace). This is not maintainable and makes the codebase harder to publish and use.

## Checklist


slowli:
The regression in the `iai` benchmark is entirely explained by the increased time to read system contracts (more specifically, most additional instructions are in `serde_json::read::next_or_eof` function, so it looks like parsing Solidity contracts in `zksync_contracts`). Not entirely sure which changes in `zksync_contracts` slowed it down; logically, the only change is removal of the load test contract.

slowli (commented):
This PR isn't 100% brushed up, but I think it should give a good general understanding of the approach. Essentially, building test contracts is now moved to a build script (which still relies on `yarn`; I haven't explored options like [`foundry-compilers`](https://crates.io/crates/foundry-compilers) yet), so that the building process is fully encapsulated inside the corresponding crate. IMO, this approach has significantly better in terms of DevEx than the previously used one, but there may be significant downsides I'm missing.

deny.toml
```diff
@@ -34,6 +32,7 @@ allow = [
     "OpenSSL",
     "Apache-2.0 WITH LLVM-exception",
     "0BSD",
+    "BSL-1.0",
```
slowli, line 35: Is this license addition OK?

core/lib/test_contracts/contracts/simple-transfer/simple-transfer.sol
```diff
@@ -19,13 +19,15 @@ contract SimpleTransfer {
     // Function to withdraw Ether to the owner's address
     function withdraw(uint _amount) public onlyOwner {
         require(address(this).balance >= _amount, "Insufficient balance in contract");
-        payable(owner).transfer(_amount);
+        (bool success, ) = owner.call{value: _amount}("");
+        require(success, "transfer reverted");
```
slowli, line 23: Not a Solidity expert: is this change OK? AFAICT, these methods aren't currently used in `SimpleTransfer` anyway. It's used only in [a single test for the prestate tracer](https://github.com/matter-labs/zksync-era/blob/474bc04e9a2e4ab03b261b55fda6cc37a27961f8/core/lib/multivm/src/versions/vm_latest/tests/prestate_tracer.rs#L62), but I had no luck so far replacing it with another contract. AFAIU, it matters that the contract is stateful, its state is written to in the constructor, and it accepts payments. Generally: is the prestate tracer used anywhere? It doesn't look used in the Era repo.

slowli (commented):
This is mostly ready for review (at least the approach is ready), with the exception of a couple of tests disabled because it's currently impossible to suppress `zksolc` errors in `foundry-compilers` (see FIXMEs).

core/tests/ts-integration/tests/contracts.test.ts
```diff
@@ -423,35 +423,6 @@ describe('Smart contract behavior checks', () => {
         expect(receipt.status).toEqual(1);
     });
 
-    test('Should check transient storage', async () => {
```
perekopskiy, line 426: Are those tests removed because of `etc/contracts-test-data` removal? If so then I would add contracts to `ts-integration/contracts` and keep the tests

perekopskiy (commented):
Overall looks nice! Is zksolc error suppressing in work?

core/tests/ts-integration/tests/contracts.test.ts
slowli: I've moved the relevant 2 tests to the `state_keeper::executor` module, so they work as unit tests now.

Deniallugo (approved):

perekopskiy (approved):

popzxc (approved):

Deniallugo (approved):

onyxet (approved):

PR #3044, opened 2024-10-09
fix: bincode deserialization for VM run data

Artemka374:
## What ❔

Implement custom deserialize for VM run data

## Why ❔

serde(deserialize_if) is not working with bincode

## Checklist

### DIFF
--- a/core/lib/prover_interface/src/inputs.rs
+++ b/core/lib/prover_interface/src/inputs.rs
@@ -198,7 +198,7 @@ impl StoredObject for VMRunWitnessInputData {
 
     fn deserialize(bytes: Vec<u8>) -> Result<Self, BoxedError> {
         zksync_object_store::bincode::deserialize::<VMRunWitnessInputData>(&bytes).or_else(|_| {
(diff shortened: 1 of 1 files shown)


core/lib/prover_interface/src/inputs.rs
```diff
@@ -160,7 +190,19 @@ impl StoredObject for VMRunWitnessInputData {
...
+        bincode::serialize(self).map_err(Into::into)
+    }
+
+    fn deserialize(bytes: Vec<u8>) -> Result<Self, BoxedError> {
+        if let Ok(res) = bincode::deserialize::<VMRunWitnessInputData>(&bytes).map_err(From::from) {
+            return Ok(res);
+        } else {
+            bincode::deserialize::<VMRunWitnessInputDataLegacy>(&bytes)
+                .map_err(From::from)
+                .map(Into::into)
```
perekopskiy, line 203: nit: you can use `or_else` instead of `if let Ok(res) ... return Ok(res);`, also let's use either From::from or Into::into for consistency

also a small comment about why we need Legacy struct is preferred

perekopskiy (approved):

PR #3048, opened 2024-10-09
chore: Bump version of consensus

brunoffranca:
## What ❔

- Bumps consensus dependencies versions to 0.4.
- Adds a note to EN docs about debug page.
- As a side-effect, two cargo files were formatted.


RomanBrodetski:
sorry for nitpicking, but I'd revert formatting. Visually I agree it's more pleasant to have them aligned, but when we add a dependency with new max length name, we'd need to change the whole file again. Also not sure it will be enforced by the linter. 
TBH I'd keep the formatting as is

PR #3058, opened 2024-10-10
feat(zk_toolbox): improved wallet handling

pompon0:
Made the tool fail with an error when wallets.yaml has parsing errors, rather than displaying "not found".
Made the tool fail when the wallets.yaml is unparseable instead of blindly overwriting it (in Localhost mode).
Made the tool check that the address actually matches the private key when parsing the config.
Made the tool check that the private key is actually a valid private key when parsing the config.
Made the code pass around Wallet, rather than weakly typed H256.


Deniallugo (approved):

zk_toolbox/crates/common/src/wallets.rs
```diff
@@ -30,7 +30,7 @@ impl<'de> Deserialize<'de> for Wallet {
                 let k = LocalWallet::from_bytes(k.as_bytes()).map_err(serde::de::Error::custom)?;
                 if k.address() != x.address {
                     return Err(serde::de::Error::custom(format!(
-                        "address does not match private key: got address {}, want {}",
```
Deniallugo: i think we need debug formatting.

pompon0: yes, I've changed it to "{:#x}"

Deniallugo (approved):

PR #3059, opened 2024-10-10
fix(nix): separate `tee_prover` and `zksync` deps

haraldh:
## What ❔

While `zksync` deps might not compile with nix, `tee_prover` deps must.

To enable nix incompatible changes in the workspace, separate the build process completely.

## Why ❔

To enable nix incompatible changes in the workspace, which are unrelated to `tee_prover`.

## Checklist


slowli (approved):
Can't say I'm a Nix expert (it's rather the opposite), but the CI job output looks valid.

PR #3070, opened 2024-10-10
feat: Implement gas relay mode and inclusion data for data attestation

QEDK:
## What ❔

This PR adds gas relay API support for gasless submission to the Avail network, it also provides the attestation implementation necessary for data attestation.

## Why ❔

Gas relay API support is required for Avail partners that choose to pay in a different token. Data attestation ensures that arbitrary tx data cannot be used for rollup finality and that no data withholding attack can occur.

## Checklist

Supersedes #2987


core/lib/config/src/configs/da_client/avail.rs
```diff
@@ -1,16 +1,40 @@
...
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Ok(GasRelayAPIKey(s.parse()?))
+    }
+}
+
 #[derive(Clone, Debug, PartialEq, Deserialize)]
 pub struct AvailConfig {
-    pub api_node_url: String,
+    pub api_node_url: Option<String>,
```
dimazhornyk, line 27: it seems like this conditional presence of some values in the config is quite confusing (especially for someone who would read the code and not know that usual client and gas relay are mutually exclusive), it would be much more understandable if we had 2 separate configs: AvailConfig and AvailGasRelayConfig, and then you can think maybe it makes sense to group them in enum (but I don't have an opinion here, up to you)

core/node/da_clients/src/avail/client.rs
```diff
@@ -1,33 +1,132 @@
...
+    // bridge root to check proof against, or reconstruct the data root
+    bridge_root: H256,
+    // leaf being proven
+    leaf: H256,
+    // index of the leaf in the blob/bridge root tree
+    leaf_index: U256,
+}
+
+impl Tokenize for MerkleProofInput {
+    fn into_tokens(self) -> Vec<Token> {
```
dimazhornyk, line 82: great job here! thanks for rewriting from alloy

core/node/da_clients/src/avail/client.rs
```diff
@@ -39,37 +138,175 @@ impl DataAvailabilityClient for AvailClient {
...
+
+        let url = format!(
+            "{}/eth/proof/{}?index={}",
+            self.config.bridge_api_url, block_hash, tx_idx
+        );
+        let mut response: reqwest::Response;
+        let mut retries = self.config.max_retries;
+        let mut response_text: String;
+        let mut bridge_api_data: BridgeAPIResponse;
+        loop {
```
dimazhornyk, line 270: we don't need this loop with a sleep here as this logic is already present in da_dispatcher
you just need to figure out if the request to the bridge API actually failed, or if the proof is just not present there, if it is not yet present - return `Ok(None)`

dimazhornyk (changes requested):
the logic seems good, but let's work on separating the configs/clients so it is always clear which one is used and what config fields are present

core/node/da_clients/src/avail/client.rs
QEDK: Fixed, closing.

core/node/da_clients/src/avail/client.rs
QEDK: ty!

core/lib/protobuf_config/src/secrets.rs
```diff
@@ -110,6 +110,12 @@ impl ProtoRepr for proto::DataAvailabilitySecrets {
                     )
                     .unwrap(),
                 ),
+                gas_relay_api_key: Some(
+                    GasRelayAPIKey::from_str(
+                        required(&avail_secret.gas_relay_api_key).context("seed_phrase")?,
+                    )
+                    .unwrap(),
```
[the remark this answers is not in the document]
QEDK: fixed, closing!

core/lib/config/src/configs/da_client/avail.rs
```diff
@@ -1,16 +1,40 @@
+use std::str::FromStr;
+
+use secrecy::ExposeSecret as _;
+use secrecy::Secret;
 use serde::Deserialize;
 use zksync_basic_types::seed_phrase::SeedPhrase;
 
+#[derive(Clone, Debug)]
+pub struct GasRelayAPIKey(pub Secret<String>);
```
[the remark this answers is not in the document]
QEDK: Fixed, closing.

core/node/da_clients/src/avail/client.rs
```diff
@@ -1,33 +1,132 @@
...
+};
 
 use crate::avail::sdk::RawAvailClient;
 
 /// An implementation of the `DataAvailabilityClient` trait that interacts with the Avail network.
 #[derive(Debug, Clone)]
 pub struct AvailClient {
     config: AvailConfig,
-    sdk_client: Arc<RawAvailClient>,
+    sdk_client: Option<Arc<RawAvailClient>>,
```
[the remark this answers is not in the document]
QEDK: Fixed, closing.

core/lib/config/src/configs/da_client/avail.rs
QEDK: Fixed, closing.

core/node/da_clients/src/avail/client.rs
```diff
@@ -39,37 +138,175 @@ impl DataAvailabilityClient for AvailClient {
         _: u32, // batch_number
         data: Vec<u8>,
     ) -> anyhow::Result<DispatchResponse, DAError> {
+        if self.config.gas_relay_mode {
```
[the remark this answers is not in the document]
QEDK: This has been addressed with an `enum` and `match` flow, closing.

core/lib/env_config/src/da_client.rs
```diff
@@ -110,8 +120,8 @@ mod tests {
             DA_API_NODE_URL="localhost:12345"
             DA_BRIDGE_API_URL="localhost:54321"
             DA_APP_ID="1"
-            DA_TIMEOUT="2"
-            DA_MAX_RETRIES="3"
+            DA_GAS_RELAY_MODE="false"
+            DA_GAS_RELAY_API_URL="localhost:23456"
```
dimazhornyk, line 124: this seems like it won't work, please try running `cargo test` in this directory to make sure it passes

core/lib/protobuf_config/src/proto/config/da_client.proto
```diff
@@ -8,9 +8,10 @@ message AvailConfig {
   optional string api_node_url = 1;
   optional string bridge_api_url = 2;
   optional uint32 app_id = 4;
-  optional uint64 timeout = 5;
```
dimazhornyk, line 11: I guess timeout is necessary

core/lib/protobuf_config/src/proto/config/da_client.proto
```diff
@@ -8,9 +8,10 @@ message AvailConfig {
...
-  optional uint64 max_retries = 6;
-  reserved 3; reserved "seed";
+  optional bool gas_relay_mode = 5;
```
dimazhornyk, line 11: we can do oneof here so we don't need this flag and don't have any optionality

core/lib/protobuf_config/src/proto/config/da_client.proto
QEDK: Fixed, closing.

core/lib/config/src/configs/da_client/avail.rs
```diff
@@ -1,16 +1,33 @@
...
+pub enum AvailClientConfig {
+    Default(AvailDefaultConfig),
+    GasRelay(AvailGasRelayConfig),
+}
+
 #[derive(Clone, Debug, PartialEq, Deserialize)]
 pub struct AvailConfig {
-    pub api_node_url: String,
     pub bridge_api_url: String,
+    pub gas_relay_mode: bool,
```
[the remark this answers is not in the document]
QEDK: Fixed, closing.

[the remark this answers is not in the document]
QEDK: Fixed, closing.

core/lib/env_config/src/da_client.rs
```diff
@@ -34,7 +34,14 @@ impl FromEnv for DataAvailabilitySecrets {
                     .ok()
                     .map(|s| s.parse())
                     .transpose()?;
-                Self::Avail(AvailSecrets { seed_phrase })
+                let gas_relay_api_key = env::var("DA_GAS_RELAY_API_KEY")
+                    .ok()
+                    .map(|s| s.parse())
+                    .transpose()?;
```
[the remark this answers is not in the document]
QEDK: It will just evaluate to `None` right? I've added another check in case both are `None`.

core/lib/protobuf_config/src/proto/config/da_client.proto
QEDK: Fixed, closing.

core/lib/protobuf_config/src/da_client.rs
```diff
@@ -17,17 +17,36 @@ impl ProtoRepr for proto::DataAvailabilityClient {
...
-                    .clone(),
-                bridge_api_url: required(&conf.bridge_api_url)
-                    .context("bridge_api_url")?
-                    .clone(),
-                app_id: *required(&conf.app_id).context("app_id")?,
-                timeout: *required(&conf.timeout).context("timeout")? as usize,
-                max_retries: *required(&conf.max_retries).context("max_retries")? as usize,
-            }),
+            proto::data_availability_client::Config::Avail(conf) => {
+                Avail(match conf.config.as_ref() {
```
perekopskiy, line 21: nit: you can move parsing shared params (`bridge_api_url`, `timeout`) out of this match

core/lib/protobuf_config/src/secrets.rs
```diff
@@ -103,14 +103,37 @@ impl ProtoRepr for proto::DataAvailabilitySecrets {
...
-            DaSecrets::Avail(avail_secret) => DataAvailabilitySecrets::Avail(AvailSecrets {
-                seed_phrase: Some(
-                    SeedPhrase::from_str(
-                        required(&avail_secret.seed_phrase).context("seed_phrase")?,
+            DaSecrets::Avail(avail_secret) => {
+                let seed_phrase = if avail_secret.seed_phrase.is_some() {
+                    Some(
+                        SeedPhrase::from_str(
+                            required(&avail_secret.seed_phrase).context("seed_phrase")?,
+                        )
```
perekopskiy, line 111: it looks weird to check if it's some and then use `required`, please refactor. Same for `gas_relay_api_key` below

core/node/da_clients/src/avail/client.rs
```diff
@@ -39,37 +137,93 @@ impl DataAvailabilityClient for AvailClient {
...
+                .map_err(to_retriable_da_error)?;
+            if bridge_api_data.error.is_none() {
+                break bridge_api_data;
+            }
+            if current_timestamp.elapsed().as_secs() > self.config.timeout as u64 {
+                return Err(to_non_retriable_da_error(anyhow!(
+                    "Inclusion check timeout exceeded"
+                )));
+            }
+            tokio::time::sleep(tokio::time::Duration::from_secs(60)).await;
```
perekopskiy, line 211: Please use [backon](https://docs.rs/backon/latest/backon/) instead of custom loop + check. 
Also, this sleep looks unreasonably long. At least it should be capped by  `config.timeout`

core/node/da_clients/src/avail/sdk.rs
```diff
@@ -369,3 +372,91 @@ fn ss58hash(data: &[u8]) -> Vec<u8> {
...
+        // send the data to the gas relay
+        let submit_response = self
+            .api_client
+            .post(&submit_url)
+            .body(Bytes::from(data))
+            .header("Content-Type", "text/plain")
+            .header("Authorization", &self.api_key)
+            .send()
+            .await
+            .map_err(to_retriable_da_error)?;
```
perekopskiy, line 424: Seems retryability is managed by caller of `post_data` so it doesn't make sense to wrap errors in `to_retriable_da_error` in this method

core/node/da_clients/src/avail/sdk.rs
```diff
@@ -369,3 +372,91 @@ fn ss58hash(data: &[u8]) -> Vec<u8> {
...
+
+        let status_url = format!(
+            "{}/user/get_submission_info?submission_id={}",
+            self.api_url, submit_response.submission_id
+        );
+
+        let (block_hash, extrinsic_index) = loop {
+            // usually takes around 40s to finalize, but polling every 5s is enough here
+            tokio::time::sleep(tokio::time::Duration::from_secs(5)).await;
```
perekopskiy, line 438: Does it make sense to set max retries/timeout here? If so please use `backon`

core/lib/protobuf_config/src/da_client.rs
QEDK: Fixed, closing.

core/lib/protobuf_config/src/secrets.rs
QEDK: Fixed, closing.

core/node/da_clients/src/avail/client.rs
QEDK: I tried `backon` but imo the implementation is much less cleaner because of having to use async closures. Admittedly, maybe I'm just not used to writing good code with `backon`.

core/node/da_clients/src/avail/sdk.rs
QEDK: Good catch, fixed.

contracts
perekopskiy, line 1: I think you accidentally changes contracts, please revert

core/node/da_clients/src/avail/client.rs
perekopskiy: >Also, this sleep looks unreasonably long. At least it should be capped by config.timeout

what about this?

core/node/da_clients/src/avail/sdk.rs
perekopskiy: Current version is bad, please add max retries regardless. If for some reason API doesn't return response it means loop will be stuck and we will have to put some effort into understanding what's going on. If code returns a meaningful error after some failed attempts, it makes it much easier to debug etc

core/node/da_clients/src/avail/client.rs
QEDK: Let me address it without using `backon` for now, I've asked @dimazhornyk to take a look at implementing using backon.

QEDK: > At least it should be capped by config.timeout
Will do.

contracts
QEDK: Fixed, closing.

core/node/da_clients/src/avail/sdk.rs
QEDK: Fixed, leaving open if @dimazhornyk wants to implement using `backon`.

core/node/da_clients/src/avail/client.rs
dimazhornyk: removed the retry logic from here completely as we have it in `da_dispatcher`, so there's no need in the duplication

perekopskiy (approved):

dimazhornyk (approved):

PR #3073, opened 2024-10-11
fix(api): Adapt `eth_getCode` to EVM emulator

slowli:
## What ❔

Fixes returned bytecodes with EVM bytecodes by removing the length prefix and padding suffix.

## Why ❔

Callers are interested in the original EVM bytecode rather than its transformed version.

## Checklist


slowli:
This PR (currently?) doesn't provide a way to distinguish between EVM and EraVM bytecodes; it just ensures that returned EVM bytecodes are well-formed.

core/node/api_server/src/web3/tests/mod.rs
```diff
@@ -1102,3 +1109,122 @@ impl HttpTest for GenesisConfigTest {
...
+#[derive(Debug)]
+struct GetBytecodeTest;
+
+impl GetBytecodeTest {
+    async fn insert_evm_bytecode(
+        connection: &mut Connection<'_, Core>,
+        at_block: L2BlockNumber,
+        address: Address,
+    ) -> anyhow::Result<()> {
+        let evm_bytecode_hash = hash_evm_bytecode(RAW_EVM_BYTECODE);
```
slowli, line 1122: @0xVolosnikov This doesn't matter in the context for this test, but in general: Does the padded / prefixed bytecode get hashed (seems like that from the EVM deploy tracer implementation), or the original one?

core/node/api_server/src/web3/namespaces/eth.rs
```diff
@@ -385,6 +385,8 @@ impl EthNamespace {
         address: Address,
         block_id: Option<BlockId>,
     ) -> Result<Bytes, Web3Error> {
+        const EVM_BYTECODE_MARKER: u8 = 2;
```
0xVolosnikov, line 388: nit: we have this const in two places. Would it be better to import it from `core/lib/utils/src/bytecode.rs`?

core/node/api_server/src/web3/tests/mod.rs
0xVolosnikov: Yes, padded and prefixed code. So we can enforce check that raw bytecode has odd amount of 32-byte words.

core/lib/utils/src/bytecode.rs
```diff
@@ -40,6 +41,7 @@ pub fn validate_bytecode(code: &[u8]) -> Result<(), InvalidBytecodeError> {
     Ok(())
 }
 
+/// Hashes the provided EraVM bytecode.
 pub fn hash_bytecode(code: &[u8]) -> H256 {
     let chunked_code = bytes_to_chunks(code);
     let hash = zk_evm::zkevm_opcode_defs::utils::bytecode_to_code_hash(&chunked_code)
```
0xVolosnikov, line 47: Maybe we should move `hash_evm` to `zk_evm::zkevm_opcode_defs` as well later.

0xVolosnikov (approved):

PR #3077, opened 2024-10-11
fix(api): Accept integer block count in `eth_feeHistory`

slowli:
## What ❔

...besides hexadecimal count accepted previously.

## Why ❔

To be more compatible with Web3 tooling.

## Checklist


core/node/api_server/src/web3/tests/mod.rs
```diff
@@ -1102,3 +1112,114 @@ impl HttpTest for GenesisConfigTest {
...
+        assert_eq!(history.inner.oldest_block, 0.into());
+        assert_eq!(
+            history.inner.base_fee_per_gas,
+            [0, 100, 200, 200].map(U256::from) // The latest value is duplicated
+        );
+        assert_eq!(history.l2_pubdata_price, all_pubdata_prices);
+        // Values below are not filled.
+        assert_eq!(history.inner.gas_used_ratio, [0.0; 3]);
+        assert_eq!(history.inner.base_fee_per_blob_gas, [U256::zero(); 4]);
+        assert_eq!(history.inner.blob_gas_used_ratio, [0.0; 3]);
```
slowli, line 1159: @0xVolosnikov Is duplicating the last value in `base_fee_per_gas` and not filling `gas_used_ratio`, `base_fee_per_blob_gas` and `blob_gas_used_ratio` OK for tooling we want to support?

0xVolosnikov: Yes, users may incorrectly assume that due to `gas_used_ratio` being < 0.5 the base fee for the next block should decrease. This may create problems with gas estimation, but I think that's a separate issue.

0xVolosnikov (dismissed):

core/node/api_server/src/web3/tests/mod.rs
slowli: Let's try as-is and then fix the returned value if necessary.

0xVolosnikov (approved):

perekopskiy (approved):

PR #3078, opened 2024-10-11
feat(gateway): refine l2 l1 logs proofs

perekopskiy:
## What ❔

Reworks how node generates proofs for l2-l1 logs. Adds new eth-watch processor that listens to `AppendedChainBatchRoot` events, updates batch tree and saves proof to DB. API reads those proofs on requests  

## Why ❔

- optimize for performance -- no need to rebuild batch tree each time, now eth watch holds only one tree
- brush up code 

## Checklist


core/node/eth_watch/src/event_processors/appended_chain_batch_root_signature.rs
```diff
@@ -0,0 +1,219 @@
...
+            .collect();
+
+        let new_events = grouped_events
+            .into_iter()
+            .skip_while(|(_sl_l1_batch_number, events)| {
+                let first_event = events.first().unwrap();
+                let last_event = events.first().unwrap();
+
+                match (
+                    first_event.0 < self.next_expected_batch_number,
```
StanislavBreadless, line 99: is this value `self.next_expected_batch_number` ever modified?

core/tests/ts-integration/tests/base-token.test.ts
```diff
@@ -170,6 +171,8 @@ describe('base ERC20 contract checks', () => {
         const withdrawalTx = await withdrawalPromise;
         const l2Receipt = await withdrawalTx.wait();
         await waitForBlockToBeFinalizedOnL1(alice, l2Receipt!.blockNumber);
+        // Sleep to give some time for eth watch to process events, l2 l1 log proof will be available only after it's done.
+        await sleep(3000);
```
StanislavBreadless, line 175: is it possible to create an API method to track it? `sleep` leads to flackiness of tests

StanislavBreadless (commented):
Generally looks good, we also discussed in private that a bit more long-term approach is that each chain calculates chain root on its own. However, it is not needed in this release, but we need to plan to be able to change it.

Also, using `sleep` in tests is not recommended since I already worked on removing those by creating `waitForBlockToBeFinalizedOnL1`

core/node/eth_watch/src/event_processors/appended_chain_batch_root_signature.rs
perekopskiy: fixed

core/node/eth_watch/src/event_processors/appended_chain_batch_root_signature.rs
```diff
@@ -0,0 +1,221 @@
...
+                (sl_l1_batch_number, group)
+            })
+            .collect();
+
+        let next_expected_batch_number = self.next_expected_batch_number;
+        let new_events = grouped_events
+            .into_iter()
+            .skip_while(|(_sl_l1_batch_number, events)| {
+                let first_event = events.first().unwrap();
+                let last_event = events.first().unwrap();
```
StanislavBreadless, line 97: looks like a typo

core/node/eth_watch/src/lib.rs
```diff
@@ -91,9 +106,26 @@ impl EthWatch {
...
+        let batch_hashes = storage
+            .blocks_dal()
+            .get_executed_batch_roots_on_sl(sl_chain_id)
+            .await?;
+
+        let next_expected_batch_number = batch_hashes
+            .last()
+            .map(|(n, _)| *n + 1)
+            .unwrap_or(L1BatchNumber(0));
+        let tree_leaves = batch_hashes.into_iter().map(|(batch_number, batch_root)| {
```
StanislavBreadless, line 119: just in case, the message root aggregation will only work on gateway (not L1)

core/lib/eth_client/src/clients/http/query.rs
```diff
@@ -266,6 +266,20 @@ where
         Ok(logs)
     }
 
+    async fn logs_extended(
+        &self,
+        filter: &web3::Filter,
+    ) -> EnrichedClientResult<Vec<zksync_types::api::Log>> {
```
StanislavBreadless, line 272: I personally agree that this is the easiest choice, but I recall I had to refactor my types with some pain to ensure that certain methods are available only for types that do support it (e.g. I added L2-like base fee history only for `ForWeb3Network<Net = L2>`). 

But if you decide to keep it this way, it may make sense to add a comment on what will happen if someone tries to use the method with l1 provider

core/lib/dal/src/blocks_dal.rs
```diff
@@ -1849,6 +1849,150 @@ impl BlocksDal<'_, '_> {
...
+        Ok(Some(
+            bincode::deserialize(&batch_chain_merkle_path).unwrap(),
+        ))
+    }
+
+    pub async fn get_executed_batch_roots_on_sl(
+        &mut self,
+        sl_chain_id: SLChainId,
+    ) -> DalResult<Vec<(L1BatchNumber, H256)>> {
+        let result = sqlx::query!(
```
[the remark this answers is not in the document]
perekopskiy: Should be simple: insert corresponding rows to `eth_txs_history` and `eth_tx` tables. However, if it happens and server restarts before we insert these rows it potentially can lead to incorrect `batch_chain_merkle_path` being saved. I guess I will add a sanity check to events processor: batch root calculated on eth watcher side should be the same as on SL

core/node/eth_watch/src/lib.rs
```diff
@@ -91,9 +106,26 @@ impl EthWatch {
             .await?
             .context("expected at least one (genesis) version to be present in DB")?;
 
+        let sl_chain_id = sl_client.chain_id().await?;
+        let batch_hashes = storage
+            .blocks_dal()
+            .get_executed_batch_roots_on_sl(sl_chain_id)
+            .await?;
+
+        let next_expected_batch_number = batch_hashes
```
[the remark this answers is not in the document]
perekopskiy: It will work. Variable name is misleading, it's not next expected but "lower bound for the next batch number". I will try to come up with better name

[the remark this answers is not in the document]
perekopskiy: I ended up with `next_batch_number_lower_bound`

core/lib/eth_client/src/clients/http/query.rs
perekopskiy: I have reworked eth watcher client

PR #2915, opened 2024-09-18
feat(vm): Split old and new VM implementations

slowli:
## What ❔

Splits old and new VM implementations in the `multivm` crate:

- Old VMs are encapsulated in the `LegacyVmInstance` enum, while new ones in the `FastVmInstance` enum (which includes plain and shadowed VM variants).
- Fast VM and `FastVmInstance` now expose a tracer type.
- Usage of the Fast VM in the batch executor are updated correspondingly.

## Why ❔

It seems infeasible to unify the tracer model for old and new VMs, so keeping them all in a single enum makes little sense.

## Checklist

- [x] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [x] Tests for the changes have been added / updated.
- [x] Documentation comments have been added / updated.
- [x] Code has been formatted via `zk fmt` and `zk lint`.

PR #3080, opened 2024-10-11
fix(vm): Check protocol version for fast VM

slowli:
## What ❔

Fixes a regression introduced by https://github.com/matter-labs/zksync-era/pull/2915: The protocol version is now not checked before instantiating a fast VM.

## Why ❔

Without this check, fast VM can be used with old protocol versions, which most likely will cause divergencies.

## Checklist

### DIFF
--- a/core/lib/multivm/src/vm_instance.rs
+++ b/core/lib/multivm/src/vm_instance.rs
@@ -1,6 +1,6 @@
 use std::mem;
 
-use zksync_types::{vm::VmVersion, Transaction};
+use zksync_types::{vm::VmVersion, ProtocolVersionId, Transaction};
 use zksync_vm2::interface::Tracer;
 
 use crate::{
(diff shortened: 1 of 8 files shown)


core/lib/multivm/src/vm_instance.rs
```diff
@@ -328,3 +328,11 @@ impl<S: ReadStorage, Tr: Tracer + Default + 'static> FastVmInstance<S, Tr> {
         Self::Shadowed(ShadowedFastVm::new(l1_batch_env, system_env, storage_view))
     }
 }
+
+/// Checks whether the protocol version is supported by the fast VM.
+pub fn is_supported_by_fast_vm(protocol_version: ProtocolVersionId) -> bool {
```
Deniallugo, line 333: We have a lot of such switches inside protocol version, probably it's better to keep it there and do not spread it across the code base

slowli: I feel that in this case, it is justified. The existing switches tell something inherent about protocol versions, while this one tells something about VM implementations (which are downstream of protocol versions; protocol versions should not be aware of how switching between implementations is implemented).

perekopskiy (approved):

PR #3081, opened 2024-10-11
fix: restore instruction count functionality

joonazan:
Brings back instruction counts that show that bytecode has changed rather than VM performance. They were previously disabled because vm2 didn't support tracers.


core/tests/vm-benchmark/src/vm.rs
```diff
@@ -90,6 +93,29 @@ impl BenchmarkingVmFactory for Fast {
     ) -> Self::Instance {
         vm_fast::Vm::custom(batch_env, system_env, storage)
     }
+
+    fn count_instructions(tx: &Transaction) -> usize {
+        let mut vm = BenchmarkingVm::<Fast<InstructionCount>>::default();
```
joonazan, line 98: I had to create quite a type tangle here to make it work. Probably there's a better way that I can't think of right now.

koloz193 (dismissed):

joonazan:
heap_read_write is still N/A, which seems wrong.

koloz193 (approved):

PR #3099, opened 2024-10-15
fix: remove l1 tx merkle tree from eth watch

perekopskiy:
## What ❔

Previously eth sender relied on eth watch to populate shared l1 tx merkle tree with new hashes. It works as long as eth sender and eth watch are run in the same process but it's not guaranteed.
I've changed the approach, now eth sender loads new txs from DB and populates its tree. We don't need tree in eth watch anymore

## Why ❔

Node should work if eth sender and eth watch are run as different processes.

## Checklist


core/node/eth_sender/src/aggregator.rs
```diff
@@ -40,20 +41,33 @@ pub struct Aggregator {
...
         commitment_mode: L1BatchCommitmentMode,
-        priority_merkle_tree: SyncMerkleTree<L1Tx>,
-    ) -> Self {
+        connection: &mut Connection<'_, Core>,
+    ) -> anyhow::Result<Self> {
         let pubdata_da = config.pubdata_sending_mode.into();
 
-        Self {
+        let priority_op_hashes =
+            if let Some(priority_tree_start_index) = config.priority_tree_start_index {
```
ly0va, line 58: If `priority_tree_start_index` is `None`, the behavior is identical to assuming it is equal to the last `priority_op_id`. A bit counter-intuitive, I would expect it to behave as `.unwrap_or(0)` since for new chains that won't undergo the migration (and hence don't need this config var) that would be the case

ly0va (approved):
:+1:

core/node/eth_sender/src/aggregator.rs
perekopskiy: actually it should work as None=Some(0) in the end because ops starting from id = 0 will be inserted at the first iteration here
```
let new_l1_tx_hashes = storage
                    .transactions_dal()
                    .get_l1_transactions_hashes(self.priority_merkle_tree.length())
                    .await
                    .unwrap();
for hash in new_l1_tx_hashes {
    self.priority_merkle_tree.push_hash(hash);
}
```
And this doesn't work as intended when priority_tree_start_index is Some(x) if x > 0. I will push fixes

ly0va (approved):

PR #3103, opened 2024-10-15
feat(eth-sender): better handling of failed L1 transactions and fixing block-reverter's clear-failed-transactions 

tomg10:
This PR changes two things to make removing L1 transactions a lot safer:
- instead of instantly panicking in case of failed L1 transaction, eth-sender will continue resending transactions (make sure there are no in-flight txs) and panic only as soon as there are none in-flight txs
- block-reverter clear-failed-transactions will not allow to proceed if there are in-flight txs

I'm making this change as waiting for all in-flight to either be confirmed or failed is relatively cheap, but makes the system safer by making it impossible to remove potentially confirmed tx from database


core/node/eth_sender/src/eth_tx_aggregator.rs
```diff
@@ -402,6 +381,24 @@ impl EthTxAggregator {
             )
             .await
         {
+            if storage
+                .eth_sender_dal()
+                .get_number_of_failed_transactions()
+                .await
+                .unwrap()
+                != 0
```
perekopskiy, line 389: Does it make sense to move this check at the start of `loop_iteration`. Then we won't get info about `agg_op` but IMO it's useless in this case

core/node/eth_sender/src/eth_tx_aggregator.rs
```diff
@@ -597,11 +594,14 @@ impl EthTxAggregator {
     ) -> Result<EthTx, EthSenderError> {
         let mut transaction = storage.start_transaction().await.unwrap();
         let op_type = aggregated_op.get_action_type();
-        // We may be using a custom sender for commit transactions, so use this
+        // We may be using a custom sender for commit tester.assert_inflight_txs_count_equals(0).await;transactions, so use this
```
perekopskiy, line 597: nit: accidental change I suppose

core/lib/dal/src/eth_sender_dal.rs
```diff
@@ -686,26 +636,53 @@ impl EthSenderDal<'_, '_> {
...
 
-    pub async fn clear_failed_transactions(&mut self) -> sqlx::Result<()> {
+    pub async fn clear_failed_transactions(&mut self) -> anyhow::Result<()> {
+        // we don't allow in-flight txs as they might be confirmed on Ethereum, but not in db
+        if self.count_all_inflight_txs().await.unwrap() != 0 {
+            return Err(anyhow!(
+                "There are still some in-flight txs, cannot proceed. \
+            Please wait for eth-sender to process all in-flight txs and try again!"
+            ));
+        }
```
perekopskiy, line 646: This logic seems to be too high-level for DAL method. I would rather check `count_all_inflight_txs  == 0` on block reverter side. Also, it makes sense to either change method name (as it no longer only clears failed transactions but also removes unsent ones) or move removing unsent ones to a different method. And maybe `#[doc(hidden)]` + comment "this method should be used only by block-reverter" wouldn't hurt

core/node/eth_sender/src/eth_tx_manager.rs
```diff
@@ -461,7 +461,6 @@ impl EthTxManager {
             tx_status.receipt,
             failure_reason
         );
-        panic!("We can't operate after tx fail");
```
perekopskiy, line 464: Concern: with this approach we will be paged with "failed tx" alert only when number_of_inflight_txs==0, it adds latency. I think it makes sense to report metric about failed tx, ideally without waiting for block finalization (this would require some refactoring) and add alert based on it

core/node/eth_sender/src/eth_tx_aggregator.rs
tomg10: good idea

core/node/eth_sender/src/eth_tx_manager.rs
tomg10: Good idea, I've added a new metric for failed transactions

core/lib/dal/src/eth_sender_dal.rs
tomg10: I've moved the check to block reverter, I've also changed the method name, it's super verbose now, but I still think it's a change for better

perekopskiy (approved):

PR #3105, opened 2024-10-16
test: Improve instruction-counting VM benchmark

slowli:
## What ❔

Replaces `iai` with an alternative; brushes up instruction counting in general.

## Why ❔

- The library currently used for the benchmark (`iai`) is unmaintained.
- It doesn't work with newer valgrind versions.
- It doesn't allow measuring parts of program execution, only the entire program run.

## Checklist


slowli:
Observations so far:

- Completely subjectively, the new approach has better DevEx; e.g., it allows filtering run benches and allows integrating reporters directly into the benchmark logic (see code).
- Instruction / cycle count measured using the new approach seems to correspond to the old approach if ~90M instruction overhead on general and VM initialization is subtracted.
- The new approach seems to better correlate with real-time benchmarks (more w.r.t. instructions than cycles), although there are still outliers. E.g., here are test results on my M2 Macbook:

```text
                               time               cycles    instructions   cycles/s, B    instr/s, B
fast/deploy_simple_contract    1.4662 ms       148594653        13479007         101.4          9.19
legacy/deploy_simple_contract  2.4865 ms       175808750        31190368          70.7          12.5
fast/access_memory             39.774 ms      1080974146       715607457          27.1          18.0
legacy/access_memory           615.71 ms     11989387670      7320044515          19.5          11.9
fast/call_far                  31.002 ms       538142795       419103438          17.4          13.5
legacy/call_far                123.89 ms      2214133575      1252371237          17.9          10.1
fast/decode_shl_sub            22.284 ms       638405039       462780306          28.6          20.8
legacy/decode_shl_sub          513.15 ms     11170751394      6856193817          21.8          13.4
fast/event_spam                38.736 ms       804507408       517321574          20.8          13.5
legacy/event_spam              335.45 ms      6875852503      4141145639          20.5          12.3
```

So, the number of instructions per second is *roughly* the same for all benches and it has the expected order of magnitude 🙃

Not so good observations:

- As expected, due to measuring parts of program execution, the benches are more sensitive to the setup logic. I've observed ~1% instruction / cycle changes caused by trivial changes in the benchmark source (e.g., iterating over benchmarks in the reverse order; running the `init` bench before / after other benches or not running it at all, etc.). To be fair, fluctuating results were partially true for the old approach as well, but probably to a lesser degree. Maybe, the results would be more stable with [cachegrind instrumentation enabled](https://slowli.github.io/yab/yab/#instrumentation), but that'd require installing a new version of valgrind.

core/tests/vm-benchmark/src/vm.rs
```diff
@@ -282,4 +292,22 @@ mod tests {
...
+    #[test]
+    fn instruction_count_matches_on_both_vms_for_transfer() {
+        let tx = get_transfer_tx(0);
+        let legacy_count = Legacy::count_instructions(&tx);
+        let fast_count = Fast::count_instructions(&tx);
+        assert_eq!(legacy_count, fast_count);
+    }
+
+    #[test]
+    fn instruction_count_matches_on_both_vms_for_fuzzed_bytecodes() {
```
joonazan, line 305: confusing naming: I think BYTECODES is the benchmark bytecodes, not the fuzzing corpus.

core/tests/vm-benchmark/src/bin/instruction_counts.rs
```diff
@@ -1,16 +1,96 @@
...
+                    .iter()
+                    .filter_map(|(&name, &new_count)| {
+                        let old_count = *old_counts.get(name)?;
+                        (old_count != new_count).then_some((name, old_count, new_count))
+                    })
+                    .collect();
+
+                if !differing_counts.is_empty() {
+                    println!("## Detected differing instruction counts");
+                    println!("| Benchmark | Old count | New count |");
```
joonazan, line 83: Are differing instruction counts now reported separately?
They will almost always cause a performance difference, too, which is why I have reported them side by side so far.

slowli: Subjectively, these are two substantially differing pieces of information; presence of any differing instruction counts is a cause of concern (maybe it makes sense to add some "⚠️"s in the relevant header), which is lost if it's reported alongside with perf differences. Another thing is that now instruction counts don't necessarily correspond to `BYTECODES` (although they still *mostly* do).

joonazan: The warning sign is a good idea because these should very rarely change. The instruction counts are usually just clutter because they say 0 difference.

core/tests/vm-benchmark/benches/instructions.rs
```diff
@@ -0,0 +1,198 @@
...
+
+        #[vise::register]
+        static VM_CACHEGRIND_METRICS: vise::Global<VmCachegrindMetrics> = vise::Global::new();
+
+        let id = self.0.to_string();
+        VM_CACHEGRIND_METRICS.instructions[&id].set(output.stats.total_instructions());
+        if let Some(&full) = output.stats.as_full() {
+            let summary = AccessSummary::from(full);
+            VM_CACHEGRIND_METRICS.l1_accesses[&id].set(summary.l1_hits);
+            VM_CACHEGRIND_METRICS.l2_accesses[&id].set(summary.l3_hits);
```
joonazan, line 84: Typo or inconsistent naming?

core/tests/vm-benchmark/benches/instructions.rs
```diff
@@ -0,0 +1,198 @@
...
+
+#[derive(Debug, Clone, Copy)]
+struct Comparison {
+    current_cycles: u64,
+    prev_cycles: Option<u64>,
+}
+
+impl Comparison {
+    fn percent_difference(a: u64, b: u64) -> f64 {
+        ((b as f64) - (a as f64)) / (a as f64) * 100.0
```
joonazan, line 99: Better to subtract first, then convert to floating point.

core/tests/vm-benchmark/benches/instructions.rs
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    fn new(output: &BenchmarkOutput) -> Option<Self> {
+        let current_cycles = AccessSummary::from(*output.stats.as_full()?).estimated_cycles();
+        let prev_cycles = if let Some(prev_stats) = &output.prev_stats {
+            Some(AccessSummary::from(*prev_stats.as_full()?).estimated_cycles())
+        } else {
+            None
+        };
```
joonazan, line 108: Looks like map would be cleaner than if let.

core/tests/vm-benchmark/benches/instructions.rs
```diff
@@ -0,0 +1,198 @@
...
+            .map(|prev_cycles| Self::percent_difference(prev_cycles, self.current_cycles))
+    }
+}
+
+/// Reporter that outputs diffs in a Markdown table to stdout after all benchmarks are completed.
+///
+/// Significant diff level can be changed via `BENCHMARK_DIFF_THRESHOLD_PERCENT` env var; it is set to 1% by default.
+#[derive(Debug, Default)]
+struct ComparisonReporter {
+    comparisons: Arc<Mutex<BTreeMap<String, Comparison>>>,
```
joonazan, line 127: I'd use a queue over a mutex here.
https://doc.rust-lang.org/std/sync/mpsc/index.html

core/tests/vm-benchmark/benches/instructions.rs
```diff
@@ -0,0 +1,198 @@
...
+}
+
+impl Drop for ComparisonReporter {
```
joonazan, line 130: This feels like an abuse of Drop. Maybe just make a method that consumes self instead?

core/tests/vm-benchmark/benches/instructions.rs
```diff
@@ -0,0 +1,198 @@
+//! Measures instruction counts for the fuzzed bytecodes.
```
joonazan, line 1:
```suggestion
//! Measures the number of host instructions required to run the benchmark bytecodes.
```

core/tests/vm-benchmark/benches/instructions.rs
slowli: No objections, but just to clarify: are you concerned about contention? IMO, it's not much of an issue in this particular case since the number of test cases is small.

core/tests/vm-benchmark/benches/instructions.rs
slowli: Yes, makes sense. I'll change the `yab` library to allow for that.

core/tests/vm-benchmark/src/vm.rs
```diff
@@ -135,7 +140,12 @@ impl BenchmarkingVmFactory for Legacy {
         let storage = StorageView::new(storage).to_rc_ptr();
         vm_latest::Vm::new(batch_env, system_env, storage)
     }
+}
 
+#[derive(Debug)]
+pub struct BenchmarkingVm<VM: BenchmarkingVmFactory>(VM::Instance);
```
joonazan, line 146: Move this down next to its impl.

core/tests/vm-benchmark/benches/instructions.rs
joonazan: No, I just prefer message passing over low-level constructs.

core/tests/vm-benchmark/benches/instructions.rs
slowli: There's a shortcut `?` after `as_full` though (i.e., we cannot compute stats if there's no cycles output in the original cachegrind data), so using a `map` would be non-trivial.

core/tests/vm-benchmark/benches/instructions.rs
slowli: The difference may be negative. Do you want to convert `a` and `b` to `i64` before subtraction? If so, IMO, it's an unnecessary complication given that this place isn't performance-critical.

joonazan: You are correct that it would be a bit ugly to convert to signed first. The reason I suggested is that subtraction is perfectly accurate for integers but not for floats.

f64 is accurate enough for all simple math, though, so it hardly matters. We're not going to bother to do the whole thing in fixed point, even though that would be more accurate in extreme cases.

core/tests/vm-benchmark/benches/instructions.rs
joonazan: Is Cachegrind's L1 a 4-cycle latency cache or more like L2?

slowli: AFAICT, it's intended to be L1 cache; see [this article](https://pythonspeed.com/articles/consistent-benchmarking-in-ci/) and [valgrind docs](https://valgrind.org/docs/manual/cg-manual.html).

joonazan (approved):

artmakh (approved):

PR #3109, opened 2024-10-16
test(vm): Deduplicate `multivm` tests

slowli:
## What ❔

Deduplicates unit tests in the `multivm` crate by retaining a single generic copy parameterized by a VM (with test-specific functionality encapsulated in a new `TestedVm` trait).

## Why ❔

2 copies of almost every test is difficult to maintain.

## Checklist


slowli:
The change diff is smaller than it looks; it mostly consists of moving stuff around. I've needed to massage a couple of tests so that they work in the same way for both legacy and fast VMs, but I've tried to not reduce the test scope whenever possible.

core/lib/multivm/src/versions/testonly/mod.rs
```diff
@@ -0,0 +1,213 @@
...
+        base_system_smart_contracts: BaseSystemContracts::playground(),
+        bootloader_gas_limit: BATCH_COMPUTATIONAL_GAS_LIMIT,
+        execution_mode: TxExecutionMode::VerifyExecute,
+        default_validation_computational_gas_limit: BATCH_COMPUTATIONAL_GAS_LIMIT,
+        chain_id: L2ChainId::from(270),
+    }
+}
+
+pub(super) fn default_l1_batch(number: L1BatchNumber) -> L1BatchEnv {
+    let timestamp = unix_timestamp_ms();
```
joonazan, line 148: Please remove all nondeterminism. It would makes debugging easier and is required for CallTracer tests.

core/lib/multivm/src/versions/testonly/mod.rs
```diff
@@ -145,7 +144,7 @@ pub(super) fn default_system_env() -> SystemEnv {
 }
 
 pub(super) fn default_l1_batch(number: L1BatchNumber) -> L1BatchEnv {
-    let timestamp = unix_timestamp_ms();
+    let timestamp = number.0.into();
```
joonazan, line 147: I would use "random" values with more entropy. The `repeat_byte` values are probably ok but this zero is not, as somebody could just set a field to zero accidentally and pass a test because of this.

slowli: Would a value like `1_700_000_000 + number.0` work? Also, are you suggesting changing existing zero values in tests? This could be a large piece of work, so I'd suggest to do it separately.

joonazan: That's a good idea! I was suggesting you generate a random number and put it there but that's even better.

joonazan (approved):

joonazan (approved):

PR #3112, opened 2024-10-16
feat(data-proof-handler): add first processed batch option

pbeza:
## What ❔

Add an option to the proof data handler to allow the first verified batch to be set.

## Why ❔

To be able to skip some batches if we need to.

## Checklist


haraldh (approved):
LGTM

core/lib/dal/.sqlx/query-4498e1c1ff179eacd03bd9ec24a7dc982c8cfb0e2277aff8dfaa9654255451ac.json
haraldh, line 1: Err.. this is an old file... it contains `l1_batches`

haraldh (changes requested):

core/lib/dal/.sqlx/query-4498e1c1ff179eacd03bd9ec24a7dc982c8cfb0e2277aff8dfaa9654255451ac.json
pbeza: Fixed.

haraldh (dismissed):

core/lib/config/src/configs/proof_data_handler.rs
```diff
@@ -1,16 +1,60 @@
...
+use std::{fmt, fmt::Display, marker::PhantomData, str::FromStr, time::Duration};
 
-use serde::Deserialize;
+use serde::{de, Deserialize};
+use zksync_basic_types::L1BatchNumber;
+
+#[derive(Debug, Deserialize, Clone, PartialEq)]
+pub struct TeeConfig {
+    /// If true, the TEE support is enabled.
+    #[serde(deserialize_with = "deserialize_stringified_any")]
```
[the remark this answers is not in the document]
pbeza: Fixed [here](https://github.com/matter-labs/zksync-era/pull/3112/commits/9e96fd1b3ea4532e9c1ea32acd4cc42c4d45cfec).

core/lib/protobuf_config/src/proto/config/prover.proto
```diff
@@ -108,4 +108,5 @@ message ProofDataHandler {
   optional uint32 http_port = 1; // required; u16
   optional uint32 proof_generation_timeout_in_secs = 2; // required; s
   optional bool tee_support = 3; // required
+  optional uint64 first_tee_processed_batch = 4; // required
```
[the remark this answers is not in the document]
pbeza: How do you set the default of `first_tee_processed_batch` to `0` in a `*.proto` file? To be honest, I don’t really understand how these proto files work in our case – I treat it as boilerplate that I need to revisit whenever modifying configuration based on environment variables. For example, we have:
```
```
Notice `uint32` vs `u16`? Is this just for documentation to indicate that `http_port` is actually represented as `u16` in the code? If so, it can easily get out of sync (unless it's checked in CI), and I wouldn't rely on it anyway.

core/lib/env_config/src/proof_data_handler.rs
```diff
@@ -4,12 +4,22 @@ use crate::{envy_load, FromEnv};
 
 impl FromEnv for ProofDataHandlerConfig {
     fn from_env() -> anyhow::Result<Self> {
-        envy_load("proof_data_handler", "PROOF_DATA_HANDLER_")
+        Ok(Self {
+            http_port: std::env::var("PROOF_DATA_HANDLER_HTTP_PORT")?.parse()?,
+            proof_generation_timeout_in_secs: std::env::var(
+                "PROOF_DATA_HANDLER_PROOF_GENERATION_TIMEOUT_IN_SECS",
+            )?
+            .parse()?,
```
slowli, line 12: AFAICT, this works:

```rust
Ok(Self {
    tee_config: envy_load("proof_data_handler.tee", "PROOF_DATA_HANDLER_")?,
    ..envy_load("proof_data_handler", "PROOF_DATA_HANDLER_")?
})
```

core/lib/env_config/src/proof_data_handler.rs
```diff
@@ -4,12 +4,22 @@ use crate::{envy_load, FromEnv};
...
+            tee_config: envy_load("proof_data_handler", "PROOF_DATA_HANDLER_")?,
```
slowli, line 13: Please use a distinguishing config name, like `proof_data_handler.tee` from the comment above.

core/lib/protobuf_config/src/proto/config/prover.proto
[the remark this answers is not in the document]
slowli: Yes, it's just a documentation thing. It's sort of checked in CI by [parsing sample file-based configs](https://github.com/matter-labs/zksync-era/blob/b29be7d9a8c664beac5d8384548db54de0ba882f/core/lib/protobuf_config/src/tests.rs#L61). You do not set a default in the proto file, you set it in the glue code, i.e., in a conversion from the Protobuf type to the domain type.

My moderately strong opinion is that the language / tooling constraints are not a reason to write lower-quality code; i.e., in this case lack of defaults / constrained Protobuf type system are not good reasons to degrade UX or desync config parsing logic compared to env-based config. (As a side note, AFAIU, we plan to eventually remove either Protobuf-based or serde-based config, but haven't 100% decided which one yet 🙃)

slowli (commented):
Other than a couple of small improvements look good. I think these improvements are worth implementing though.

core/lib/env_config/src/proof_data_handler.rs
pbeza: [Fixed](https://github.com/matter-labs/zksync-era/pull/3112/commits/a75a0a2c7b8b86fd972e5364619c61f4e482fe1b). ✅

core/lib/env_config/src/proof_data_handler.rs
pbeza: [Fixed](https://github.com/matter-labs/zksync-era/pull/3112/commits/a75a0a2c7b8b86fd972e5364619c61f4e482fe1b). ✅

core/lib/protobuf_config/src/proto/config/prover.proto
```diff
@@ -108,4 +108,5 @@ message ProofDataHandler {
   optional uint32 http_port = 1; // required; u16
   optional uint32 proof_generation_timeout_in_secs = 2; // required; s
   optional bool tee_support = 3; // required
```
haraldh, line 110: should this be `// optional` then also?

pbeza: Thanks for noticing it. Fixed [here](https://github.com/matter-labs/zksync-era/pull/3112/commits/d943314829de7e41bbbea6bc0d8cf30fdd83a6f1). If you have a strong opinion to make it required, please lemme know @slowli / @haraldh.

I guess we want to have it optional, defaulting to `false` to make it consistent with [this](https://github.com/matter-labs/gitops-kubernetes/blob/04bd23eb714622298c04adb750804ed35a66976e/apps/common/zksync-v2-config/zksync-v2-config.yaml#L616).

If we decide to make it required, we [would need](https://github.com/matter-labs/zksync-era/pull/3112#discussion_r1810406067) to add the new param for all envs using file-based configs beforehand.

pbeza:
@slowli, ready for re-review. 🫡

slowli (approved):

PR #3113, opened 2024-10-16
fix(mempool): minor mempool improvements

perekopskiy:
## What ❔

fixes two mempool issues:
1) `gc` does not really ensure `size <= capatity` (found by @jcsec-security), fix: purge some accounts with lowest score from priority queue
2) Mempool actor built l2 tx filter based on current l1 gas prices. This doesn't make sense when there is an open batch, if there is some then we should use its fee params.

## Checklist

### DIFF
--- a/core/lib/mempool/src/mempool_store.rs
+++ b/core/lib/mempool/src/mempool_store.rs
@@ -1,4 +1,4 @@
-use std::collections::{hash_map, BTreeSet, HashMap, HashSet};
+use std::collections::{hash_map, BTreeSet, HashMap};
 
 use zksync_types::{
     l1::L1Tx, l2::L2Tx, Address, ExecuteTransactionCommon, Nonce, PriorityOpId, Transaction,
@@ -221,22 +221,53 @@ impl MempoolStore {
     }
 
     fn gc(&mut self) -> Vec<Address> {
-        if self.size >= self.capacity {
-            let index: HashSet<_> = self
+        if self.size > self.capacity {
+            let mut transactions = std::mem::take(&mut self.l2_transactions_per_account);
+            let mut possibly_kept: Vec<_> = self
                 .l2_priority_queue
                 .iter()
-                .map(|pointer| pointer.account)
+                .rev()
+                .filter_map(|pointer| {
+                    transactions
+                        .remove(&pointer.account)
+                        .map(|txs| (pointer.account, txs))
+                })
                 .collect();
-            let transactions = std::mem::take(&mut self.l2_transactions_per_account);
-            let (kept, drained) = transactions
+            let mut number_of_accounts_kept = possibly_kept
+                .iter()
+                .scan(0, |sum, (_, txs)| {
+                    *sum += txs.len();
+                    (*sum <= self.capacity as usize).then_some(())
+                })
+                .count();
+            if number_of_accounts_kept == 0 {
+                tracing::warn!("mempool capacity is too low to handle txs from single account, consider increasing capacity");
+                // Keep at least one entry, otherwise mempool won't return any new L2 tx to process.
+                number_of_accounts_kept = 1;
+            }
+            let (kept, drained) = {
+                let mut drained: Vec<_> = transactions.into_keys().collect();
+                let also_drained = possibly_kept
+                    .split_off(number_of_accounts_kept)
+                    .into_iter()
+                    .map(|(address, _)| address);
+                drained.extend(also_drained);
+
+                (possibly_kept, drained)
+            };
+
+            let l2_priority_queue = std::mem::take(&mut self.l2_priority_queue);
+            self.l2_priority_queue = l2_priority_queue
                 .into_iter()
-                .partition(|(address, _)| index.contains(address));
-            self.l2_transactions_per_account = kept;
+                .rev()
+                .take(number_of_accounts_kept)
+                .collect();
+            self.l2_transactions_per_account = kept.into_iter().collect();
             self.size = self
                 .l2_transactions_per_account
                 .iter()
-                .fold(0, |agg, (_, tnxs)| agg + tnxs.len() as u64);
-            return drained.into_keys().collect();
+                .fold(0, |agg, (_, txs)| agg + txs.len() as u64);
+            return drained;
         }
         vec![]
     }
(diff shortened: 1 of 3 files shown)


core/lib/mempool/src/mempool_store.rs
```diff
@@ -221,22 +221,53 @@ impl MempoolStore {
...
                 .collect();
-            let transactions = std::mem::take(&mut self.l2_transactions_per_account);
-            let (kept, drained) = transactions
+            let mut number_of_accounts_kept = possibly_kept
+                .iter()
+                .scan(0, |sum, (_, txs)| {
+                    *sum += txs.len();
+                    (*sum <= self.capacity as usize).then_some(())
+                })
+                .count();
```
slowli, line 242: Bikeshedding: IMO, `scan` is not the most obvious choice here; an imperative approach with a `for` loop would be better. Alternatively, maybe [`take_while`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.take_while) would work here?

core/lib/mempool/src/mempool_store.rs
```diff
@@ -221,22 +221,53 @@ impl MempoolStore {
...
+            if number_of_accounts_kept == 0 {
+                tracing::warn!("mempool capacity is too low to handle txs from single account, consider increasing capacity");
+                // Keep at least one entry, otherwise mempool won't return any new L2 tx to process.
+                number_of_accounts_kept = 1;
+            }
+            let (kept, drained) = {
+                let mut drained: Vec<_> = transactions.into_keys().collect();
+                let also_drained = possibly_kept
+                    .split_off(number_of_accounts_kept)
```
slowli, line 251: Dumb question not looking at the unchanged code 🙃: is it possible for `possibly_kept` to be empty? If so, increasing `number_of_accounts_kept` above will lead to a panic here.

core/lib/mempool/src/tests.rs
```diff
@@ -321,32 +321,26 @@ fn stashed_accounts() {
...
-    );
-    // verify that existing good-to-go transactions and new ones got picked
-    mempool.insert(
-        vec![gen_l2_tx_with_timestamp(
-            account1,
-            Nonce(0),
-            unix_timestamp_ms() + 1,
-        )],
-        HashMap::new(),
+        HashSet::<_>::from_iter(vec![account2, account3]),
```
slowli, line 341: Nit: you can use a conversion from an array here: `HashSet::from([account2, account3])`.

core/node/state_keeper/src/mempool_actor.rs
```diff
@@ -89,12 +89,30 @@ impl MempoolFetcher {
...
+                .get_unsealed_l1_batch()
+                .await
+                .context("failed getting unsealed batch")?
+            {
+                let (base_fee, gas_per_pubdata) = derive_base_fee_and_gas_per_pubdata(
+                    unsealed_batch.fee_input,
+                    protocol_version.into(),
+                );
+
+                L2TxFilter {
```
slowli, line 103: Nit: creating `L2TxFilter` looks unnecessary here; IIUC, only `fee_per_gas` and `gas_per_pubdata` fields are used below, so it could be possible to return `derive_base_fee_and_gas_per_pubdata()` output directly.

core/node/state_keeper/src/mempool_actor.rs
```diff
@@ -89,12 +89,30 @@ impl MempoolFetcher {
                 .await
                 .context("failed getting pending protocol version")?;
 
-            let l2_tx_filter = l2_tx_filter(
-                self.batch_fee_input_provider.as_ref(),
-                protocol_version.into(),
-            )
-            .await
-            .context("failed creating L2 transaction filter")?;
+            let l2_tx_filter = if let Some(unsealed_batch) = storage
```
slowli, line 92: Dumb question: Does this change allow for more accurate transaction filtering? What's the benefit from it?

core/lib/mempool/src/mempool_store.rs
perekopskiy: nice catch, yes, it's possible

core/node/state_keeper/src/mempool_actor.rs
perekopskiy: yes, more accurate filtering:
1) if current L1 gas price is less than batch's gas price it means that we can potentially insert some transaction to mempool with low fee params, those transactions will be in insert->stash->insert loop till the next batch
2) if current L1 gas price is higher than batch's gas price it means that some transaction that can be included in the batch will not make it to mempool

core/lib/mempool/src/mempool_store.rs
perekopskiy: pushed imperative approach

core/node/state_keeper/src/mempool_actor.rs
```diff
@@ -89,20 +89,35 @@ impl MempoolFetcher {
...
+                    protocol_version.into(),
+                );
+                (fee_per_gas, gas_per_pubdata as u32)
+            } else {
+                let filter = l2_tx_filter(
+                    self.batch_fee_input_provider.as_ref(),
+                    protocol_version.into(),
+                )
+                .await
+                .context("failed creating L2 transaction filter")?;
```
itegulov, line 109: Why do we even create a filter here? Can't we do something like this instead:
```rust
let fee_input = if let Some(unsealed) = ... {
    unsealed_batch.fee_input
} else {
    batch_fee_input_provider.get_batch_fee_input().await?
}
let (fee_per_gas, gas_per_pubdata) = derive_base_fee_and_gas_per_pubdata(
    unsealed_batch.fee_input,
    protocol_version.into(),
);
```

itegulov: Also, looking at it again, aren't you doing pretty the same as what https://github.com/matter-labs/zksync-era/pull/2868 is doing just on a smaller scale? Or do you intend this to be a temporary fix specifically for mempool?

perekopskiy: > Why do we even create a filter here?

I'm creating filter so mempool_actor and mempool_io are using the same `l2_tx_filter` method, so there are less chances that someone changes it only for one of them.

It's connected to 2868 but separate: your PR solves that API may return incorrect fee params, my PR solves that mempool actor may use incorrect fee params

itegulov: Ah, yes, it's using a different `BatchFeeModelInputProvider` impl underneath, makes sense.

itegulov (approved):

jcsec-security (approved):
The issue looks fixed to me

PR #3121, opened 2024-10-17
refactor: Refactor fee-related types

slowli:
## What ❔

- Moves basic logic for fee types close to the type declaration (i.e., to `zksync_types`).
- Unifies `PubdataSendingMode` and `PubdataDA`.
- Removes dependency of `zksync_types` on `zksync_config`.

## Why ❔

This allows external apps depending on fee types (e.g., the test node) to not have behemoths like `zksync_dal` in their dependency graph.

## Checklist


core/lib/basic_types/src/pubdata_da.rs
```diff
@@ -1,15 +1,17 @@
...
+
+use crate::L1BatchNumber;
 
 /// Enum holding the current values used for DA Layers.
 #[repr(u8)]
-#[derive(Debug, Clone, Copy, Deserialize, PartialEq, Serialize)]
-#[derive(TryFromPrimitive)]
-pub enum PubdataDA {
+#[derive(Debug, Clone, Copy, Default, PartialEq, Deserialize, Serialize, TryFromPrimitive)]
+pub enum PubdataSendingMode {
```
slowli, line 12: This refactoring assumes that `PubdataSendingMode` and `PubdataDA` in fact refer to the same concept. I'm not 100% sure that is the case. The current definitions are obviously identical, but maybe they'll diverge in the future?

dimazhornyk: my guess was that these were duplicated as a workaround to some circular dependency, so they shouldn't diverge in the future, but maybe I'm missing some other reason

dimazhornyk (approved):

perekopskiy (approved):

PR #3124, opened 2024-10-17
feat: update MAXIMUM_L2_GAS_PRICE and MAXIMUM_PUBDATA_PRICE

cytadela8:
## What ❔

- increase limits to `2**63 - 1` 

## Why ❔

- protocol upgrade removes this limits, but we still need to limit due to Postgres schema type used (bigint)
- -1 maybe not needed, but added for safety.

## Checklist

### DIFF
--- a/core/node/fee_model/src/lib.rs
+++ b/core/node/fee_model/src/lib.rs
@@ -247,18 +247,18 @@ fn compute_batch_fee_model_input_v2(
     }
 }
 
-/// Bootloader places limitations on fair_l2_gas_price and fair_pubdata_price.
-/// (MAX_ALLOWED_FAIR_L2_GAS_PRICE and MAX_ALLOWED_FAIR_PUBDATA_PRICE in bootloader code respectively)
+/// Postgres schema places limitations on the gas and pubdata price. Previously this was also
+/// limited by bootloader. This is no longer the case.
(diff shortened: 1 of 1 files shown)


core/node/fee_model/src/lib.rs
```diff
@@ -247,18 +247,18 @@ fn compute_batch_fee_model_input_v2(
...
+/// TODO (PE-153): to be removed when Postgres schema is updated
 fn clip_batch_fee_model_input_v2(
     fee_model: PubdataIndependentBatchFeeModelInput,
 ) -> PubdataIndependentBatchFeeModelInput {
     /// MAX_ALLOWED_FAIR_L2_GAS_PRICE
-    const MAXIMUM_L2_GAS_PRICE: u64 = 10_000_000_000_000;
+    const MAXIMUM_L2_GAS_PRICE: u64 = 2_u64.pow(63) - 1;
     /// MAX_ALLOWED_FAIR_PUBDATA_PRICE
-    const MAXIMUM_PUBDATA_PRICE: u64 = 1_000_000_000_000_000;
+    const MAXIMUM_PUBDATA_PRICE: u64 = 2_u64.pow(63) - 1;
```
perekopskiy, line 261: please add integration test that justifies the node works correctly when limits are reached

perekopskiy (changes requested):

deniallugo-ml:

PR #3125, opened 2024-10-18
fix(external-node): delete empty unsealed batch on EN initialization

itegulov:
## What ❔

This PR reverts #3088 as I have realized it is going to be very hard to make this fix work by going in that direction. Basically initializing with an empty unsealed batch causes a lot of issues and the existing state keeper/external IO flow heavily relies on us having at least one at the start to initialize correctly. Will leave more context in the comments.

Feel free to review individual commits to not see revert changelog.

## Why ❔

This bug causes EN to panic

## Checklist


core/node/node_sync/src/external_io.rs
```diff
@@ -155,6 +155,14 @@ impl StateKeeperIO for ExternalIO {
...
+            tracing::info!(
+                l1_batch_number = %cursor.l1_batch,
+                "Empty unsealed batch found; deleting it as we need at least one L2 block to initialize"
+            );
+            storage
+                .blocks_dal()
+                .delete_unsealed_l1_batch(cursor.l1_batch - 1)
+                .await?;
             return Ok((cursor, None));
         };
```
itegulov, line 167: An alternative here is to fetch the pending L2 block as it is guaranteed to exist on the main node (we can only receive `SyncAction::OpenBatch` command along with `SyncAction::SealL2Block`; essentially this bug only reproduces if EN restarts in between of opening a batch and processing the L2 block).

I didn't go with the approach as it seems to be error-prone and I would rather finally unblock the next release to mainnet, but happy to explore it if anyone prefers it.

itegulov: Another more robust approach is to refactor `L1BatchParamsProvider::load_l1_batch_params` in such a way that it does not rely on the first L2 block anymore as I think we don't actually need anything block-specific there. This requires much more effort and time though.

slowli: The latter option sounds the most reasonable in the long term. Dumb question regarding alternatives: Is inserting an unsealed batch atomically with its first block not an option (in general or for ENs)? I'd imagine it would help preventing issues like this, but it could be the case that it makes an L1 batch appear in the storage too late (esp. for the main node).

core/node/node_sync/src/sync_action.rs
```diff
@@ -33,6 +33,21 @@ impl ActionQueueSender {
...
+    /// Useful to simulate situations where only a part of the sequence was executed on the node.
+    #[cfg(test)]
+    pub async fn push_action_unchecked(&self, action: SyncAction) -> anyhow::Result<()> {
+        self.0
+            .send(action)
+            .await
+            .map_err(|_| anyhow::anyhow!("node action processor stopped"))?;
+        QUEUE_METRICS
+            .action_queue_size
+            .set(self.0.max_capacity() - self.0.capacity());
```
slowli, line 47: Nit: I don't think metrics are relevant for tests. Similarly, it could make sense to call `expect` on the send result.

core/node/node_sync/src/external_io.rs
```diff
@@ -155,6 +155,14 @@ impl StateKeeperIO for ExternalIO {
                 )
             })?;
         let Some(mut pending_l2_block_header) = pending_l2_block_header else {
+            tracing::info!(
+                l1_batch_number = %cursor.l1_batch,
+                "Empty unsealed batch found; deleting it as we need at least one L2 block to initialize"
```
slowli, line 160: Dumb question: Does an unsealed batch necessarily exist at this point? AFAICT, `delete_unsealed_l1_batch` is a no-op if it doesn't (so it would work correctly), but the log message may be misleading.

slowli (commented):
Can a similar situation reproduce on the main node (i.e., an unsealed batch w/o blocks / transactions)? Is it handled OK?

- AFAICT, `MempoolIO::initialize()` can insert an unsealed L1 batch via `ensure_unsealed_l1_batch_exists` (as off-topic / bikeshedding, I think this method is too high-level to be placed in DAL). OTOH, we only reach the `ensure_unsealed_l1_batch_exists` call if `L1BatchParamsProvider::load_l1_batch_env()` above returns `Some(_)`, i.e. there is at least one L2 block persisted for the batch. So it doesn't look like the `ensure_unsealed_l1_batch_exists` may actually persist a batch without blocks.
- In `MempoolIO::wait_for_new_batch_params()`, a new batch is inserted into the storage proactively. If a node crashes immediately afterwards, it looks like the situation *can* be reproduced.

itegulov:
@slowli 
> Can a similar situation reproduce on the main node (i.e., an unsealed batch w/o blocks / transactions)? Is it handled OK?

I believe that is handled OK as main node has a slightly different mechanism: `MempoolIO::wait_for_new_batch_params` checks if there is an unsealed batch before doing anything else and returns it proactively if it does. ENs IO logic is a lot more tied to `fethcher` and relies on a specific order of actions arriving which makes things complicated. Another strange difference on EN is that it expects L1 batches to not have `protocol_version` populated until a certain point which also complicates things (I still need to figure out what's up with that in a separate PR).

But I will write a test just to validate that everything is ok.

> So it doesn't look like the ensure_unsealed_l1_batch_exists may actually persist a batch without blocks.

Ok, so the logic is a bit confusing and I should have left a comment but essentially this is a compatibility mechanism. Imagine that EN starts for the very first time after unsealed batches PR made it in - we have `N` L2 blocks that belong to the pending batch but the pending batch is not inserted as unsealed. Then this bit makes sure that it is as the rest of the flow depends on it.

> as off-topic / bikeshedding, I think this method is too high-level to be placed in DAL

Fair, this was a bit of dirty hack to be honest with you to fix stage which was in crash loop at the time. This entire logic is IMHO something that should've been a Rust migration script but AFAICT we don't have anything like that. Anyway, I'll try to refactor this in a separate PR.

> If a node crashes immediately afterwards, it looks like the situation can be reproduced.

As I said there is actually a fail-safe mechanism at the very top of the function that prevents this from happening. Unless you mean something else?

core/node/node_sync/src/external_io.rs
itegulov: > Is inserting an unsealed batch atomically with its first block not an option (in general or for ENs)?

This was an option I considered at the start but I really wanted to minimize the amount of time when we misreport gas on low-traffic envs (the entire reason we are doing this work). We often spend more than a minute without sealing a miniblock which would mean anyone who makes gas-related requests during that time might get incorrect information.

Now, on ENs it might be less important (although above reasoning still applies to them too) but would involve some amount of refactoring on action queue logic side.

itegulov: In general I find it strange that we depend on the first miniblock so much and hopefully something I might address in the future

core/node/node_sync/src/sync_action.rs
itegulov: True, it snuck in when I copy-pasted the other method

core/node/node_sync/src/external_io.rs
itegulov: Good catch, clarified

perekopskiy (approved):

itegulov:
@slowli I am merging this PR but thanks for the critical review and if you still have questions/concern please feel free to continue the discussion here. I will deliver extra tests/improvements in a follow-up PR.

PR #3126, opened 2024-10-18
feat(vm): Return compressed bytecodes from `push_transaction()`

slowli:
## What ❔

Returns compressed bytecodes from `VmInterface::push_transaction()`.

## Why ❔

This can be used by some external VM users. It is a more idiomatic replacement of removed `VmInterface::get_last_tx_compressed_bytecodes()`, and it's more efficient for newer VMs (doesn't clone bytecodes, instead providing a reference from the bootloader state).

## Checklist


joonazan (approved):
Nice use of Cow!

PR #3128, opened 2024-10-18
feat(proof-data-handler): add tee_proof_generation_timeout_in_secs param

pbeza:
## What ❔

Add `tee_proof_generation_timeout_in_secs` parameter to the `proof-data-handler` configuration to avoid sharing the same `proof_generation_timeout_in_secs` timeout with the ZK prover.

This timeout is for retrying TEE proof generation if it fails. Retries continue indefinitely until successful.

## Why ❔

The TEE prover is much faster than the ZK prover, so some of the ZK timeouts are too long to be shared with the TEE-specific code.

## Checklist


pbeza:
Perhaps I should extract TEE-specific options, like I did in other PR that is not yet merged:
https://github.com/matter-labs/zksync-era/blob/6f0b18ef1068904bf92c5b00a2a410f2fc216eff/core/lib/config/src/configs/proof_data_handler.rs#L20-L21

Lemme know.

haraldh (approved):

core/lib/config/src/configs/proof_data_handler.rs
```diff
@@ -28,6 +33,14 @@ impl TeeConfig {
     pub fn default_first_tee_processed_batch() -> L1BatchNumber {
         L1BatchNumber(0)
     }
+
+    pub fn default_tee_proof_generation_timeout_in_secs() -> u16 {
+        600
+    }
+
+    pub fn tee_proof_generation_timeout(&self) -> Duration {
+        Duration::from_secs(self.tee_proof_generation_timeout_in_secs as u64)
```
slowli, line 42: Nit: You can use `tee_proof_generation_timeout_in_secs.into()` to clearly signal that the conversion is lossless.

slowli (dismissed):

core/lib/config/src/configs/proof_data_handler.rs
pbeza: [Fixed](https://github.com/matter-labs/zksync-era/pull/3128/commits/391ce36c36abfe54dc7eea14ad5b9efaef511b06). ✅

haraldh (approved):

PR #3135, opened 2024-10-21
feat(api): Implement eth_maxPriorityFeePerGas

### LINKED ISSUE #13: eth_maxPriorityFeePerGas not available
While zkSync era using EIP1559,  eth_maxPriorityFeePerGas isn't implemented.
```
curl --url https://zksync2-mainnet.zksync.io \
-X POST \
-d '{ "id": 1, "jsonrpc": "2.0", "method": "eth_maxPriorityFeePerGas" }'

{"jsonrpc":"2.0","error":{"code":-32601,"message":"Method not found"},"id":1}
```

Although eth_maxPriorityFeePerGas is not a standard rpc method of Ethereum, it is very important in practical use, especially for determining gas fee when sending transactions.

Thanks!

popzxc:
## What ❔

Fixes #13 
Implements `eth_maxPriorityFeePerGas` method.

## Why ❔

This method is a de-facto standard now, and SDKs (e.g. viem) can use it assuming that it's supported.
Even given that we're not really using EIP1559, we should still support it and return 0.

## Checklist


core/tests/ts-integration/tests/api/web3.test.ts
```diff
@@ -271,7 +272,8 @@ describe('web3 API compatibility tests', () => {
 
         const eip1559ApiReceipt = await alice.provider.getTransaction(eip1559Tx.hash);
         expect(eip1559ApiReceipt.maxFeePerGas).toEqual(eip1559Tx.maxFeePerGas!);
-        expect(eip1559ApiReceipt.maxPriorityFeePerGas).toEqual(eip1559Tx.maxPriorityFeePerGas!);
+        // `ethers` will use value provided by `eth_maxPriorityFeePerGas`, and we return 0 there.
+        expect(eip1559ApiReceipt.maxPriorityFeePerGas).toEqual(0n);
```
itegulov, line 276: Omg that was it? This is highly tangential but any chance we have a TS/JS wizard available to us? I have noticed that `jest` traces SUCK when they are run against async TS code. It is very hard to judge what actually went wrong sometimes. There is even an [issue](https://github.com/kulshekhar/ts-jest/issues/4177) about that but no one cared to address it.

itegulov (approved):

core/tests/ts-integration/tests/api/web3.test.ts
popzxc: There is a lot to unpack here. From what I understood:
After this change, `maxPriorityFeePerGas` in the transaction became `null` (I guess ethers does that because we return 0 from API), and in receipt we have 0 as BigInt. So while comparing 0n to null, jest uses implicit comparison (`==` instead of `===`), so both arguments are stringified. But BigInt cannot be stringified, so it crashes Jest worker 😓 
Same behavior will happen if you will change `0n` to just `0`.

Javascript.

itegulov: Rust-based integration tests when???

popzxc: Once budgeted.

PR #3140, opened 2024-10-21
feat(zkstack_cli): use docker-managed volumes

itegulov:
## What ❔

Host-bound volumes never get cleaned even if you do `docker compose down -v`. Existing dev flows don't seem to rely on volumes being on the host machine, so this is PoC of how we can move to Docker-managed volumes.

## Why ❔

Avoid pesky bugs that prevent user from deleting `./volumes`, rely on Docker to persist and dispose of data as need be

## Checklist

### DIFF
--- a/docker-compose-gpu-runner-cuda-12-0.yml
+++ b/docker-compose-gpu-runner-cuda-12-0.yml
@@ -6,8 +6,8 @@ services:
     ports:
       - 127.0.0.1:8545:8545
     volumes:
-      - type: bind
-        source: ./volumes/reth/data
+      - type: volume
+        source: reth-data
         target: /rethdata
       - type: bind
         source: ./etc/reth/chaindata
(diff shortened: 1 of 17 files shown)


docker-compose-gpu-runner-cuda-12-0.yml
```diff
@@ -69,3 +69,7 @@ services:
     environment:
       # We bind only to 127.0.0.1, so setting insecure password is acceptable here
       - POSTGRES_PASSWORD=notsecurepassword
+
+volumes:
+  postgres-data:
+  reth-data:
```
itegulov, line 75: Do we even use `docker-compose-cpu-runner.yml` and `docker-compose-gpu-runner.yml` anymore? Pretty sure those can just be deleted, but happy to update if anyone insists.

amakhortov (approved):

artmakh (approved):

Deniallugo (approved):

PR #3163, opened 2024-10-23
feat(vm): Support EVM emulation in fast VM

slowli:
## What ❔

Supports EVM emulation if it is signaled via `SystemEnv`.

## Why ❔

Fast VM should have feature parity with the legacy VM.

## Checklist


core/lib/multivm/src/versions/vm_fast/vm.rs
```diff
@@ -842,15 +874,35 @@ impl<S: ReadStorage, T: Tracer> zksync_vm2::StorageInterface for World<S, T> {
...
+/// This would be incorrect behavior because bytecode deployments must be reverted along with transactions.
+///
+/// In reality, this cannot happen because both `decommit()` and `decommit_code()` calls perform storage-based checks
+/// before a decommit:
+///
+/// - `decommit_code()` is called from the `CodeOracle` system contract, which checks that the decommitted bytecode is known.
+/// - `decommit()` is called during far calls, which obtains address -> bytecode hash mapping beforehand.
+///
+/// Thus, if storage is reverted correctly, additional EVM bytecodes occupy the cache, but are unreachable.
+// FIXME: can it be used for DoS?
```
slowli, line 891: - I don't quite like the reasoning above, but it seems to work for now at least.
- Regarding DoS, the hypothetical scenario is that the user can somehow cheaply add EVM bytecodes into the cache (perhaps eventually running out of gas), so that they persist for the entire VM run as deadweight. This could a problem only for batch processing (i.e., state keeper and VM runner-based components), not for the API server.

core/lib/multivm/src/versions/vm_fast/evm_deploy_tracer.rs
```diff
@@ -0,0 +1,101 @@
...
+            }
+            Err(err) => tracing::error!("Unable to decode `publishEVMBytecode` call: {err}"),
+        }
+    }
+}
+
+impl Tracer for EvmDeployTracer {
+    #[inline(always)]
+    fn after_instruction<OP: OpcodeType, S: GlobalStateInterface>(&mut self, state: &mut S) {
+        if !matches!(OP::VALUE, Opcode::FarCall(CallingMode::Normal)) {
```
joonazan, line 96: I'd remove the negation and put the next line inside.

core/lib/multivm/src/versions/vm_fast/evm_deploy_tracer.rs
```diff
@@ -0,0 +1,101 @@
...
+    fn handle_far_call(&self, state: &mut impl GlobalStateInterface) {
+        let from = state.current_frame().caller();
+        let to = state.current_frame().code_address();
+        if from != CONTRACT_DEPLOYER_ADDRESS || to != KNOWN_CODES_STORAGE_ADDRESS {
+            return;
+        }
+
+        let (calldata_ptr, is_pointer) =
+            state.read_register(CALL_IMPLICIT_CALLDATA_FAT_PTR_REGISTER + 1);
+        assert!(
```
joonazan, line 60: I wouldn't assert things that are not critical for this function's correctness.

core/lib/multivm/src/versions/vm_fast/vm.rs
```diff
@@ -578,8 +585,12 @@ impl<S: ReadStorage, Tr: Tracer + Default + 'static> VmInterface for Vm<S, Tr> {
...
         let gas_before = self.gas_remaining();
 
-        let mut full_tracer = (mem::take(tracer), CircuitsTracer::default());
+        let mut full_tracer = (
+            (mem::take(tracer), CircuitsTracer::default()),
+            EvmDeployTracer::new(self.world.dynamic_bytecodes.clone()),
+        );
         let result = self.run(execution_mode, &mut full_tracer, track_refunds);
+        let (full_tracer, _) = full_tracer;
         *tracer = full_tracer.0; // place the tracer back
```
joonazan, line 594: Doing this in two steps is weird. Either fully pattern match or do `full_tracer.0.0`.

core/lib/multivm/src/versions/vm_fast/vm.rs
joonazan: I'm ok with this. If there is DoS it is not this code's fault. Adding a bytecode but not paying for it shouldn't be possible, even if reverted.

joonazan (commented):
I did not review the tests. The VM changes look good.

core/lib/multivm/src/versions/vm_fast/evm_deploy_tracer.rs
slowli: I'd argue that in this case, both assertions are critical for correctness, in the sense that if they don't hold, the logic below doesn't work:

- If `r1` is not a pointer, this may mean a changed calling convention or a similar failure. The value in `r1` in this case cannot be interpreted.
- If `calldata_ptr.offset` is not zero, this means that the range of data to be read should change.

Would you prefer to allow for arbitrary `calldata_ptr.offset`? (I would be OK with this.) To return early if `r1` is not a pointer instead of panicking? (IMO, panicking is better because would allow to catch not updating calling convention logic etc. earlier.)

joonazan: My account validation PR adds a helper function `read_fat_pointer`, which does take offset into acount. I'd use that.

Switching to a different calling convention isn't possible, as it would break all existing bytecodes.

core/lib/multivm/src/versions/vm_fast/vm.rs
```diff
@@ -637,6 +648,11 @@ impl<S: ReadStorage, Tr: Tracer + Default + 'static> VmInterface for Vm<S, Tr> {
         let gas_remaining = self.gas_remaining();
         let gas_used = gas_before - gas_remaining;
 
+        // We need to filter out bytecodes the deployment of which may have been reverted; the tracer is not aware of reverts.
+        // To do this, we check bytecodes against deployer events.
+        let factory_deps_marked_as_known = extract_bytecodes_marked_as_known(&logs.events);
+        let new_known_factory_deps = self.world.decommit_bytecodes(&factory_deps_marked_as_known);
```
perekopskiy, line 654: Is it possible to include only EVM bytecodes here? IIUC there is no point in storing era-vm code in `new_known_factory_deps `.

perekopskiy (dismissed):

core/lib/multivm/src/versions/vm_fast/vm.rs
slowli: There's a TODO for this; I'd prefer to work on it separately (looks quite easy to do, maybe deceptively so).

https://github.com/matter-labs/zksync-era/blob/8ae06b237647715937fb3656d881c0fd460f2a07/core/lib/vm_interface/src/types/outputs/execution_result.rs#L124-L125

perekopskiy (approved):

PR #3166, opened 2024-10-24
fix: extend allowed storage slots for validation as per EIP-7562

ly0va:
## What ❔

[EIP-7562](https://eips.ethereum.org/EIPS/eip-7562#validation-rules) allows reading `keccak256(address || x) + n` where `x` is `bytes32` and `n` is `0..128`.

This PR adds support for the `+ n` as we didn't have it before.

## Why ❔

To support reading larger-than-1-slot structs from mappings, during validation

## Checklist


etc/env/base/chain.toml
```diff
@@ -90,8 +90,8 @@ fee_model_version = "V2"
 validation_computational_gas_limit = 300000
 save_call_traces = true
 
-bootloader_hash = "0x010008c37ecadea8b003884eb9d81fdfb7161b3b309504e5318f15da19c500d8"
-default_aa_hash = "0x0100055da70d970f98ca4677a4b2fcecef5354f345cc5c6d13a78339e5fd87a9"
+bootloader_hash = "0x010008c3be57ae5800e077b6c2056d9d75ad1a7b4f0ce583407961cc6fe0b678"
```
vladbochok, line 93: Why?

ly0va: server would complain that calculated hashes don't match the config

but not sure why that happened

core/lib/multivm/src/tracers/validator/mod.rs
```diff
@@ -133,7 +139,11 @@ impl<H> ValidationTracer<H> {
         // The user is allowed to touch its own slots or slots semantically related to him.
         let valid_users_slot = address == self.user_address
             || u256_to_account_address(&key) == self.user_address
```
vladbochok, line 141: Interesting, but not an issue. `u256_to_account_address` just takes the last 20 bytes, and doesn't enforce that the leading 12 bytes are zeroes. The original EIP proposed to not allow key with address + dirty leading bytes.

core/lib/multivm/src/tracers/validator/mod.rs
```diff
@@ -133,7 +139,11 @@ impl<H> ValidationTracer<H> {
...
-            || self.auxilary_allowed_slots.contains(&u256_to_h256(key));
+            || self
+                .auxilary_allowed_slots
+                .range(u256_to_h256(key - Self::MAX_ALLOWED_SLOT_OFFSET)..=u256_to_h256(key))
```
vladbochok, line 144: What will happen if this is underflow?

etc/env/base/chain.toml
StanislavBreadless: this definitely should not be required in this pr

core/lib/multivm/src/tracers/validator/mod.rs
ly0va: I thought that this won't matter since probability of getting a keccak < uint256(127) is basically 0, but I see now that it underflow can happen regardless. Will fix

etc/env/base/chain.toml
kelemeno: Also you're still using zk and not zk toolbox?

ly0va: @kelemeno oh ok that was the issue, yes

core/lib/multivm/src/tracers/validator/mod.rs
```diff
@@ -131,9 +137,17 @@ impl<H> ValidationTracer<H> {
...
 
         // The user is allowed to touch its own slots or slots semantically related to him.
+        let from = u256_to_h256(key.saturating_sub(Self::MAX_ALLOWED_SLOT_OFFSET.into()));
+        let to = u256_to_h256(key);
         let valid_users_slot = address == self.user_address
-            || u256_to_account_address(&key) == self.user_address
-            || self.auxilary_allowed_slots.contains(&u256_to_h256(key));
+            // we don't allow dirty bytes in front of the address
+            || key < U256::from(2).pow(160.into())
+                && u256_to_account_address(&key) == self.user_address
```
vladbochok, line 145: Maybe?
```suggestion
            || key == address_to_u256(self.user_address)
```

vladbochok (approved):

vladbochok (approved):
`GENESIS_ROOT` still changed, no idea of implications of this though

kelemeno (approved):

PR #3168, opened 2024-10-24
fix(consensus): payload encoding protected by protocol_version

pompon0:
Changing payload encoding without protocol version change would invalidate consensus signatures.


core/lib/dal/src/consensus/conv.rs
```diff
@@ -169,11 +180,17 @@ impl ProtoFmt for Payload {
...
+        };
+        anyhow::ensure!(this.pubdata_params.is_none() == this.protocol_version.is_pre_gateway());
+        Ok(this)
     }
 
     fn build(&self) -> Self::Proto {
+        assert!(
+            self.protocol_version.is_pre_gateway() == self.pubdata_params.is_none(),
+            "pubdata_params should be None iff protocol_version is pre-gateway"
+        );
```
perekopskiy, line 193: nit: assert_eq! can be used

core/lib/dal/src/models/storage_sync.rs
```diff
@@ -139,7 +139,11 @@ impl SyncBlock {
             operator_address: self.fee_account_address,
             transactions,
             last_in_batch: self.last_in_batch,
-            pubdata_params: Some(self.pubdata_params),
+            pubdata_params: if self.protocol_version.is_pre_gateway() {
+                Some(self.pubdata_params)
+            } else {
+                None
+            },
```
perekopskiy: vice versa 🙃

perekopskiy (approved):

artmakh (approved):

PR #3169, opened 2024-10-24
fix(tee_prover): add prometheus pull listener

haraldh:
## What ❔

Add a prometheus pull listener.

## Why ❔

To get the metrics out of the zk_tee_prover

## Checklist

### DIFF
--- a/core/bin/zksync_tee_prover/src/main.rs
+++ b/core/bin/zksync_tee_prover/src/main.rs
@@ -45,11 +45,12 @@ fn main() -> anyhow::Result<()> {
         .add_layer(SigintHandlerLayer)
         .add_layer(TeeProverLayer::new(tee_prover_config));
 
-    if let Some(gateway) = prometheus_config.gateway_endpoint() {
-        let exporter_config =
-            PrometheusExporterConfig::push(gateway, prometheus_config.push_interval());
-        builder.add_layer(PrometheusExporterLayer(exporter_config));
-    }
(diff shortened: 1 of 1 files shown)


core/bin/zksync_tee_prover/src/main.rs
```diff
@@ -45,11 +45,12 @@ fn main() -> anyhow::Result<()> {
...
-        let exporter_config =
-            PrometheusExporterConfig::push(gateway, prometheus_config.push_interval());
-        builder.add_layer(PrometheusExporterLayer(exporter_config));
-    }
+    let exporter_config = if let Some(gateway) = prometheus_config.gateway_endpoint() {
+        PrometheusExporterConfig::push(gateway, prometheus_config.push_interval())
+    } else {
+        PrometheusExporterConfig::pull(prometheus_config.listener_port)
+    };
+    builder.add_layer(PrometheusExporterLayer(exporter_config));
```
pbeza, line 53: It may be worth adding `tracing::info` in case both `listener_port` and `gateway_endpoint()` were set? Same as they did here:
https://github.com/matter-labs/zksync-era/blob/ffa18e1d84a4bb1ca9b897fbc0a55b9e3ef0964c/core/bin/external_node/src/config/observability.rs#L70-L79

pbeza (approved):

core/bin/zksync_tee_prover/src/main.rs
haraldh: listener_port is no Option. It is set every time.

PR #3172, opened 2024-10-25
feat(en): Support Merkle tree recovery with pruning enabled

slowli:
## What ❔

Generalizes Merkle tree recovery so that it supports recovery after the node has pruned some of its data.

## Why ❔

Improves node UX; allows to recover a tree of a full / non-archive node.

## Checklist


core/node/metadata_calculator/src/recovery/mod.rs
```diff
@@ -279,13 +313,18 @@ impl AsyncTreeRecovery {
...
         let finalize_latency = RECOVERY_METRICS.latency[&RecoveryStage::Finalize].start();
         let actual_root_hash = tree.root_hash().await;
-        anyhow::ensure!(
-            actual_root_hash == snapshot.expected_root_hash,
-            "Root hash of recovered tree {actual_root_hash:?} differs from expected root hash {:?}. \
-             If pruning is enabled and the tree is initialized some time after node recovery, \
-             this is caused by snapshot storage logs getting pruned; this setup is currently not supported",
-            snapshot.expected_root_hash
-        );
+        if let Some(expected_root_hash) = init_params.expected_root_hash {
```
slowli, line 316: It would be beneficial to record the state root hash of the last hard-pruned L1 batch somewhere (in pruning logs?) so that it can be checked here. If this sounds OK, I'd prefer to implement it in a follow-up PR.

core/node/metadata_calculator/src/recovery/mod.rs
```diff
@@ -340,6 +379,21 @@ impl AsyncTreeRecovery {
         Ok(output)
     }
 
+    async fn check_pruning_info(
```
slowli, line 382: - I *think* this check has low probability of false negatives. Optimally though, we'd want to check the tree root hash so that we know for sure that we don't commit to a broken recovery; see another line comment.
- It would be nice to somehow prevent pruning until the recovery is complete, but I'm not sure how to do that. It could be possible to signal this via Postgres, but there could be issues if the tree recovery is aborted for whatever reason. Alternatively, it may be possible to check whether pruning is enabled at the start / continuation of tree recovery, and error if it is. This may lead to false positives if pruning is configured to never happen in practice, but may be good enough.

perekopskiy: WDYM by " if pruning is configured to never happen in practice"?

perekopskiy (approved):

PR #3173, opened 2024-10-25
feat(zkstack_cli): l2 contract verification

pompon0:
Added command for verifying L2 contracts deployed with zkstack_cli.

Tested with local env and on testnet.


pompon0:
Can you elaborate on what is this complication? If anything, this change:
* decreases the amount of code
* removes useless abstraction on script results
* makes script usage more uniform (single entry point -> single output, generalization to arbitrary set of contracts instead of a bunch of chosen ones)

zkstack_cli/Cargo.toml
```diff
@@ -36,6 +36,7 @@ zksync_consensus_crypto = "=0.5.0"
 zksync_consensus_utils = "=0.5.0"
 zksync_protobuf = "=0.5.0"
 zksync_protobuf_build = "=0.5.0"
+zksync_concurrency = "=0.5.0"
```
popzxc, line 39: Is it used anywhere?

popzxc (commented):
Generally LGTM, but I'd suggest to also get a review from @sanekmelnikov

PR #3174, opened 2024-10-25
feat(l1-recover): DO NOT MERGE, work in progress

tomg10:


core/node/l1_recovery/src/l1_fetcher/l1_fetcher.rs
```diff
@@ -0,0 +1,1083 @@
...
+    }
+
+    fn hyperchain_contract() -> Contract {
+        hyperchain_contract()
+    }
+
+    fn commit_functions() -> Result<Vec<Function>> {
+        let contract = Self::hyperchain_contract();
+        Ok(vec![
+            contract.functions_by_name("commitBatches").unwrap()[0].clone(),
```
pompon0, line 113: `function(...)`

core/node/l1_recovery/src/l1_fetcher/l1_fetcher.rs
```diff
@@ -0,0 +1,1083 @@
...
+    }
+
+    fn rollup_event_by_type(event_type: RollupEventType) -> Result<Event> {
+        let event_name = match event_type {
+            RollupEventType::Commit => "BlockCommit",
+            RollupEventType::Prove => "BlocksVerification",
+            RollupEventType::Execute => "BlockExecution",
+            RollupEventType::NewPriorityTx => "NewPriorityRequest",
+        };
+        Ok(Self::hyperchain_contract().events_by_name(event_name)?[0].clone())
```
pompon0, line 128: `event(...)`

core/node/l1_recovery/src/l1_fetcher/l1_fetcher.rs
```diff
@@ -0,0 +1,1083 @@
...
+
+    fn extract_l1_batch_number_from_rollup_event_log(
+        event_type: RollupEventType,
+        log: &Log,
+    ) -> L1BatchNumber {
+        let topic_index = match event_type {
+            RollupEventType::Commit => 1,
+            RollupEventType::Prove => 2,
+            RollupEventType::Execute => 1,
+            _ => panic!("{event_type:?} event doesn't have l1_batch_number"),
```
pompon0, line 139: return error

core/node/l1_recovery/src/l1_fetcher/l1_fetcher.rs
```diff
@@ -0,0 +1,1083 @@
...
+        };
+        L1BatchNumber(log.topics[topic_index].to_low_u64_be() as u32)
+    }
+
+    fn extract_priority_op_id_from_rollup_event_log(
+        event_type: RollupEventType,
+        log: &Log,
+    ) -> PriorityOpId {
+        assert_eq!(event_type, RollupEventType::NewPriorityTx);
```
pompon0, line 148: why do you need the event_type argument in the first place, if you are just asserting the value?

core/node/l1_recovery/src/l1_fetcher/l1_fetcher.rs
```diff
@@ -0,0 +1,1083 @@
...
+        {
+            Ok(Some(tx)) => Ok(tx),
+            Ok(None) => Err(anyhow!("unable to find transaction with hash: {}", hash)),
+            Err(e) => Err(e),
+        }
+    }
+
+    fn query_client(eth_client: &Box<DynClient<L1>>) -> &dyn EthInterface {
+        eth_client
+    }
```
pompon0, line 771: private getter looks weird imo

core/node/l1_recovery/src/l1_fetcher/l1_fetcher.rs
```diff
@@ -0,0 +1,1083 @@
...
+    /// Get the last published L1 block.
+    async fn get_last_l1_block_number(eth_client: &Box<DynClient<L1>>) -> Result<U64> {
```
pompon0, line 773: why it doesn't use &self instead?

core/node/l1_recovery/src/l1_fetcher/l1_fetcher.rs
```diff
@@ -0,0 +1,1083 @@
...
+        let last_block = L1Fetcher::retry_call(
+            || L1Fetcher::query_client(eth_client).block(BlockId::Number(BlockNumber::Finalized)),
+            L1FetchError::GetEndBlockNumber,
+        )
+        .await?;
+
+        last_block
+            .expect("last block must be present")
```
pompon0, line 781: return error

core/node/l1_recovery/src/l1_fetcher/l1_fetcher.rs
```diff
@@ -0,0 +1,1083 @@
+use std::{cmp, cmp::PartialEq, collections::HashMap, future::Future, sync::Arc};
+
+use anyhow::{anyhow, Result};
```
pompon0, line 3: nit: use "format_err" (alias of anyhow) instead. It is a more meaningful name

pompon0, line 3: use qualified `anyhow::Result` instead, so as not to shadow the Result from prelude, especially since you define custom errors in this file as well.

core/node/l1_recovery/src/l1_fetcher/l1_fetcher.rs
```diff
@@ -0,0 +1,1083 @@
...
+
+impl L1Fetcher {
+    pub fn new(config: L1FetcherConfig, eth_client: Box<DynClient<L1>>) -> Result<Self> {
+        tracing::info!("Initialized L1 fetcher with config: {config:?}");
+        Ok(L1Fetcher { eth_client, config })
+    }
+
+    fn hyperchain_contract() -> Contract {
+        hyperchain_contract()
+    }
```
pompon0, line 108: not very useful method, isn't it :P

core/node/l1_recovery/src/l1_fetcher/l1_fetcher.rs
```diff
@@ -0,0 +1,1083 @@
...
+        };
+        L1BatchNumber(log.topics[topic_index].to_low_u64_be() as u32)
+    }
+
+    fn extract_priority_op_id_from_rollup_event_log(
+        event_type: RollupEventType,
+        log: &Log,
+    ) -> PriorityOpId {
+        assert_eq!(event_type, RollupEventType::NewPriorityTx);
+        L1Tx::try_from(log.clone()).unwrap().common_data.serial_id
```
pompon0, line 149: return error

core/node/l1_recovery/src/l1_fetcher/l1_fetcher.rs
```diff
@@ -0,0 +1,1083 @@
...
+    }
+
+    pub async fn find_block_near_genesis(&self) -> U64 {
+        let mut start_block: U64 = U64::zero();
+        let last_l1_block_number = L1Fetcher::get_last_l1_block_number(&self.eth_client)
+            .await
+            .unwrap();
+        let mut end_block = last_l1_block_number;
+        // we use step override as such big step value guarantees that we don't miss
```
pompon0, line 158: I don't understand this comment. Can you elaborate?

core/node/l1_recovery/src/l1_fetcher/l1_fetcher.rs
```diff
@@ -0,0 +1,1083 @@
...
+            if !logs.is_empty() {
+                end_block = mid;
+            } else {
+                start_block = mid;
+            }
+        }
+        end_block = last_l1_block_number;
+        let mut current_block = start_block;
+        loop {
+            let filter_to_block = cmp::min(current_block + step - 1, end_block);
```
pompon0, line 175: fyi, instead of `cmp::min` you can use `min` method (no imports required), for example
`let filter_to_block = end_block.min(...)`.

core/node/l1_recovery/src/l1_fetcher/l1_fetcher.rs
```diff
@@ -0,0 +1,1083 @@
...
+                    current_block,
+                    filter_to_block,
+                    RollupEventType::NewPriorityTx,
+                )
+                .await;
+            if !logs.is_empty() {
+                if L1Fetcher::extract_priority_op_id_from_rollup_event_log(
+                    RollupEventType::NewPriorityTx,
+                    &logs[0],
+                ) == PriorityOpId(0)
```
pompon0, line 187: does that mean that this function is supposed to find the L1 block which contains the first priority transaction? Please document.

core/node/l1_recovery/src/l1_fetcher/types/mod.rs
```diff
@@ -0,0 +1,308 @@
...
+    InvalidCompressedValue(String),
+
+    #[error("invalid pubdata source: {0}")]
+    InvalidPubdataSource(String),
+
+    #[error("blob storage error: {0}")]
+    BlobStorageError(String),
+
+    #[error("blob format error: {0}")]
+    BlobFormatError(String, String),
```
pompon0, line 52: second field is omitted, is that intentional?

core/node/l1_recovery/src/l1_fetcher/types/mod.rs
```diff
@@ -0,0 +1,308 @@
...
+pub enum CommitBlockInfo {
+    V1(V1),
+    V2(V2),
+}
+
+/// Block with all required fields extracted from a [`CommitBlockInfo`].
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct CommitBlock {
+    /// ZKSync batch number.
+    pub l1_batch_number: u64,
```
pompon0, line 88: why not L1BatchNumber?

core/node/l1_recovery/src/l1_fetcher/types/mod.rs
```diff
@@ -0,0 +1,308 @@
...
+}
+
+impl CommitBlock {
+    pub fn try_from_token<'a, F>(value: &'a ethabi::Token) -> Result<Self, ParseError>
+    where
+        F: CommitBlockFormat + TryFrom<&'a ethabi::Token, Error = ParseError>,
+    {
+        let commit_block_info = F::try_from(value)?;
+        Ok(Self::from_commit_block(commit_block_info.to_enum_variant()))
+    }
```
pompon0, line 117: wouldn't https://github.com/matter-labs/zksync-era/tree/main/core/lib/l1_contract_interface be a better place for this logic?

core/node/l1_recovery/src/l1_fetcher/l1_fetcher.rs
```diff
@@ -0,0 +1,1083 @@
...
+        let genesis_block = self.find_block_near_genesis().await;
+        let logs = self
+            .get_logs(
+                genesis_block,
+                genesis_block + self.config.block_step,
+                RollupEventType::Commit,
+            )
+            .await;
+        let tx =
+            L1Fetcher::get_transaction_by_hash(&self.eth_client, logs[0].transaction_hash.unwrap())
```
pompon0, line 226: if you don't want to make these methods of L1Fetcher, then consider wrapping eth_client into yet another type instead and making functions like get_transaction_by_hash methods of that new type instead.

core/node/l1_recovery/src/l1_fetcher/l1_fetcher.rs
```diff
@@ -0,0 +1,1083 @@
...
+    pub fn new(config: L1FetcherConfig, eth_client: Box<DynClient<L1>>) -> Result<Self> {
+        tracing::info!("Initialized L1 fetcher with config: {config:?}");
+        Ok(L1Fetcher { eth_client, config })
+    }
+
+    fn hyperchain_contract() -> Contract {
+        hyperchain_contract()
+    }
+
+    fn commit_functions() -> Result<Vec<Function>> {
```
pompon0, line 110: any specific reason why those functions are members of L1Fetcher class?

core/node/l1_recovery/src/l1_fetcher/l1_fetcher.rs
```diff
@@ -0,0 +1,1083 @@
...
+
+    pub async fn get_all_blocks_to_process(
+        &self,
+        blob_client: &Arc<dyn BlobClient>,
+        commit_blocks_cache: Option<&Arc<dyn ObjectStore>>,
+        stop_receiver: &watch::Receiver<bool>,
+    ) -> Vec<CommitBlock> {
+        let end_block = L1Fetcher::get_last_l1_block_number(&self.eth_client)
+            .await
+            .unwrap();
```
pompon0, line 211: return error

core/node/l1_recovery/src/l1_fetcher/l1_fetcher.rs
```diff
@@ -0,0 +1,1083 @@
...
+
+            let should_retry = |err_code, err_message: String| {
+                // All of these can be emitted by either API provider.
+                err_code == Some(-32603)             // Internal error
+                    || err_message.contains("failed")    // Server error
+                    || err_message.contains("timed out") // Time-out error
+            };
+
+            // check whether the error is related to having too many results
+            if err_message.contains(TOO_MANY_RESULTS_INFURA)
```
pompon0, line 366: just to confirm: there is no error code for these?

core/node/l1_recovery/src/l1_fetcher/l1_fetcher.rs
```diff
@@ -0,0 +1,1083 @@
...
+                        return result;
+                    }
+                };
+
+                // divide range into two halves and recursively fetch them
+                let mid = (from_number + to_number) / 2;
+
+                // safety check to prevent infinite recursion (quite unlikely)
+                if from_number >= mid {
+                    tracing::warn!("Infinite recursion detected while getting events: from_number={from_number:?}, mid={mid:?}");
```
pompon0, line 394: why warn? Isn't this the regular execution path for a binary search?

core/node/l1_recovery/src/l1_fetcher/l1_fetcher.rs
```diff
@@ -0,0 +1,1083 @@
...
+                        RETRY_LIMIT,
+                    )
+                    .await?;
+
+                first_half.append(&mut second_half);
+                result = Ok(first_half);
+            } else if should_retry(err_code, err_message) && retries_left > 0 {
+                tracing::warn!("Retrying. Retries left: {retries_left}");
+                result = self
+                    .get_events_inner(from, to, topics1, topics2, addresses, retries_left - 1)
```
pompon0, line 425: nit: it seems wasteful to use recursion for retries instead of a loop, no?
Also the layout/block nesting of this function could be improved, by inverting success and failure branches, i.e.:
```
let err = match result {
  Ok(res) => return res;
  Err(err) => err
};
// error handling.
```

core/node/l1_recovery/src/l1_fetcher/l1_fetcher.rs
```diff
@@ -0,0 +1,1083 @@
...
+            .ok_or_else(|| anyhow!("found latest block, but it contained no block number"))
+    }
+
+    async fn retry_call<T, E, Fut>(callback: impl Fn() -> Fut, err: L1FetchError) -> Result<T>
+    where
+        Fut: Future<Output = Result<T, E>>,
+        E: std::fmt::Display,
+    {
+        sleep(Duration::from_millis(100)).await;
+        for attempt in 1..=100 {
```
pompon0, line 792: that's a lot of retries. Why 100?

core/node/l1_recovery/src/l1_fetcher/l1_fetcher.rs
```diff
@@ -0,0 +1,1083 @@
...
+            .number
+            .ok_or_else(|| anyhow!("found latest block, but it contained no block number"))
+    }
+
+    async fn retry_call<T, E, Fut>(callback: impl Fn() -> Fut, err: L1FetchError) -> Result<T>
+    where
+        Fut: Future<Output = Result<T, E>>,
+        E: std::fmt::Display,
+    {
+        sleep(Duration::from_millis(100)).await;
```
pompon0, line 791: this is an artificial delay. Why is it needed?

core/node/l1_recovery/src/l1_fetcher/l1_fetcher.rs
```diff
@@ -0,0 +1,1083 @@
...
+    let _new_blocks_data = parsed_input.pop().unwrap();
+    let stored_block_info = parsed_input.pop().unwrap();
+    return Ok(StoredBatchInfo::from_token(stored_block_info).unwrap());
+}
+
+pub async fn parse_calldata(
+    protocol_versioning: &ProtocolVersioning,
+    l1_block_number: u64,
+    commit_candidates: &[Function],
+    calldata: &[u8],
```
pompon0, line 827: this parsing looks very manual. Can't we make it more robust by using ethabi library somehow? Also l1_contracts_interface could be a better place for this logic.

core/node/l1_recovery/src/l1_fetcher/blob_http_client.rs
```diff
@@ -0,0 +1,137 @@
...
+                        }
+                    },
+                    Err(e) => {
+                        tracing::error!("attempt {}: {} failed: {:?}", attempt, full_url, e);
+                        sleep(Duration::from_secs(FAILED_FETCH_RETRY_INTERVAL_S)).await;
+                    }
+                },
+                Err(e) => {
+                    tracing::error!("attempt {}: GET {} failed: {:?}", attempt, full_url, e);
+                    sleep(Duration::from_secs(FAILED_FETCH_RETRY_INTERVAL_S)).await;
```
pompon0, line 121: invert control flow, deduplicate sleeps

core/node/l1_recovery/src/l1_fetcher/l1_fetcher.rs
```diff
@@ -0,0 +1,1083 @@
...
+        .await
+        {
+            Ok(Some(tx)) => Ok(tx),
+            Ok(None) => Err(anyhow!("unable to find transaction with hash: {}", hash)),
+            Err(e) => Err(e),
+        }
+    }
+
+    fn query_client(eth_client: &Box<DynClient<L1>>) -> &dyn EthInterface {
+        eth_client
```
pompon0, line 770: this function does explicit casting to an interface. Does rust compiler have problems without it?

tomg10: yeah, I haven't been able to convince the compiler to not throw errors without that "hack"

core/node/l1_recovery/src/l1_fetcher/l1_fetcher.rs
tomg10: it's a piece of code directly taken from old repo, I agree it's super ugly, but I'd still prefer to not have to refactor it as it feels like it's not worth it, that is also the reason why I don't want to move it to l1_contracts_interface, to keep the code's style consistent there. I will add a comment about that though

core/node/l1_recovery/src/l1_fetcher/l1_fetcher.rs
tomg10: it's there to not spam the provider with requests, I've added a comment

core/node/l1_recovery/src/l1_fetcher/l1_fetcher.rs
tomg10: 100 was high enough number to realistically never run out of. I've refactored it to just be a loop as we realistically always want to retry

PR #3189, opened 2024-10-29
refactor(contract-verifier): Brush up contract verifier

slowli:
## What ❔

Brushes up contract verifier code.

## Why ❔

Would make it easier to support EVM emulation in the contract verifier.

## Checklist


slowli (commented):
- Is the contract verifier library used somewhere externally? (Looks like that judging by commit history.) These uses would probably need to be updated, and I'm not quite sure which verifier APIs should be public.
- I don't see a compelling reason why the contract verifier API should be a node component, especially given that the verifier backend is not a component. I think it could work autonomously provided that it uses some interface to interact with a node (namely, fetch the contract bytecode, contract deployer events and the deploying transaction). All of these seem to be covered by the Web3 RPC. Would such a refactoring be too deep to handle? (Obviously, won't do it in this PR since it's large as-is.)

perekopskiy (dismissed):
> Is the contract verifier library used somewhere externally?

@AntonD3 please help with it.

>I don't see a compelling reason why the contract verifier API should be a node component

It shouldn't. It was a compromise to put it there. Overall I wouldn't proceed with any deep refactoring unless it's clearly needed/beneficial short term. There were discussions about moving contract verifier out of zksync-era somewhere closer to block-explorer but this was deprioritized. So its future is not determined

core/lib/contract_verifier/Cargo.toml
```diff
@@ -27,8 +26,11 @@ ethabi.workspace = true
...
 hex.workspace = true
 serde = { workspace = true, features = ["derive"] }
-lazy_static.workspace = true
 tempfile.workspace = true
 regex.workspace = true
 tracing.workspace = true
 semver.workspace = true
+
+[dev-dependencies]
+zksync_node_test_utils.workspace = true
```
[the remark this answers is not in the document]
slowli: Created a ticket for it; will do separately.

yorik (dismissed):
Approve for workflows

perekopskiy (approved):

yorik (approved):

core/lib/contract_verifier/src/resolver.rs
```diff
@@ -0,0 +1,220 @@
...
+                .context("failed reading zkvyper dir")?,
+        })
+    }
+
+    async fn resolve_solc(
+        &self,
+        versions: &CompilerVersions,
+    ) -> Result<Box<dyn Compiler<ZkSolcInput>>, ContractVerifierError> {
+        let zksolc_version = versions.zk_compiler_version();
+        let zksolc_path = self
```
popzxc, line 136: Nit: This part feels overly repeated. Could we encapsulate it to something like `CompilerPaths::try_detect(&self.home_dir, Compiler::ZkSolc).await?` or `Compiler::ZkSolc::try_detect(&self.home_dir).await?` ?

core/lib/contract_verifier/src/tests/real.rs
```diff
@@ -0,0 +1,141 @@
...
+//! with at least one `solc` and `zksolc` version. If there are no compilers, the tests will be ignored
+//! unless the `RUN_CONTRACT_VERIFICATION_TEST` env var is set to `true`, in which case the tests will fail.
+
+use std::{env, sync::Arc, time::Duration};
+
+use zksync_utils::bytecode::validate_bytecode;
+
+use super::*;
+
+fn assert_no_compilers_expected() {
```
popzxc, line 11: Nit: maybe the code would be a bit easier to read if it was a macro like 
```rust
macro_rules! skip_if_not_enabled {
   () => {
       if std::env::var("RUN_CONTRACT_VERIFICATION_TEST").ok().as_deref() != Some("true") {
          println!("Test is not enabled, skipping");
          return;
       }
   }
}
```

this way the logic to skip the test will not be mixed with the actual test logic.

popzxc (approved):
Sorry for being late to the party. Looks great!
Left a few nits, but feel free to ignore.

PR #3193, opened 2024-10-29
feat(zksync_cli): Health checkpoint improvements

manuelmauro:
## What ❔

Add three new components to node's healthcheck:
- General (i.e., version, last migration)
- State Keeper 
- Eth Sender

## Checklist


core/lib/dal/src/system_dal.rs
```diff
@@ -86,4 +98,29 @@ impl SystemDal<'_, '_> {
...
+
+    pub async fn get_last_migration(&mut self) -> DalResult<DatabaseMigration> {
+        let row = sqlx::query!(
+            r#"
+            SELECT *
+            FROM _sqlx_migrations
+            WHERE _sqlx_migrations.version = (
+                SELECT MAX(_sqlx_migrations.version)
+                FROM _sqlx_migrations
+            );
```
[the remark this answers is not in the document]
manuelmauro: Thank you for the review @slowli ! While the two methods have some subtle differences, I agree that maintaining consistency with the existing `order by` pattern in the codebase might be more valuable than optimizing for the relatively minor performance gains. What do you think?"

1. Using Subquery with MAX:
```sql
SELECT *
FROM table
WHERE number = (SELECT MAX(number) FROM table)
```

2. Using ORDER BY with LIMIT:
```sql
SELECT *
FROM table
ORDER BY number DESC
LIMIT 1
```

**Performance**:
- Subquery approach typically performs two scans but can use indexes efficiently
- ORDER BY approach requires sorting the entire table, which is usually more expensive

**Memory Usage**:
- Subquery uses less memory as it doesn't require sorting
- ORDER BY needs additional memory for sorting operation

**Multiple Maximum Values**:
- Subquery approach will return ALL rows that have the maximum value
- ORDER BY approach will return only ONE row, even if multiple rows have the same maximum value

core/node/house_keeper/src/database.rs
```diff
@@ -0,0 +1,42 @@
...
+    pub connection_pool: ConnectionPool<Core>,
+    pub database_health_updater: HealthUpdater,
+}
+
+#[async_trait]
+impl PeriodicJob for DatabaseHealthTask {
+    const SERVICE_NAME: &'static str = "DatabaseHealth";
+
+    async fn run_routine_task(&mut self) -> anyhow::Result<()> {
+        let mut conn = self.connection_pool.connection().await?;
```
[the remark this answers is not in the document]
manuelmauro: Addressed in https://github.com/matter-labs/zksync-era/pull/3193/commits/b31233b1561a3a78058fab39f0eaee2070a54aae

core/node/house_keeper/src/eth_sender.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+impl From<EthSenderInfo> for Health {
+    fn from(details: EthSenderInfo) -> Self {
+        Self::from(HealthStatus::Ready).with_details(details)
+    }
+}
+
+#[derive(Debug)]
+pub struct EthSenderHealthTask {
```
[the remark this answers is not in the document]
manuelmauro: Added reactive health check to state keeper in https://github.com/matter-labs/zksync-era/pull/3193/commits/5396e5ac703ae8c3cca86d58fd3faa103fa54b2a

core/lib/dal/src/system_dal.rs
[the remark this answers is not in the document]
perekopskiy: There is no gain as `version` is a private key and thus is indexed, so query with `ORDER BY version DESC LIMIT 1` doesn't actually sort table, postgres is smart enough to just take last row from the index.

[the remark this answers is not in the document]
manuelmauro: Thanks for the observation @perekopskiy ! Addressed in https://github.com/matter-labs/zksync-era/pull/3193/commits/bb5acfc3a99676226aa28fc2d48d28b3a2852dda

core/node/house_keeper/src/eth_sender.rs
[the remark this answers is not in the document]
manuelmauro: Added reactive health check to eth sender in https://github.com/matter-labs/zksync-era/pull/3193/commits/fd35543fe875a662bdf88c68ea30490d088b6036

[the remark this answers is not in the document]
manuelmauro: @Deniallugo I attached an healthcheck response to the PR description, it is currently pretty verbose and does not fully cover the requirements of the task. Could you give me a feedback on what information would you like to add/remove?

core/node/node_framework/src/implementations/layers/house_keeper.rs
```diff
@@ -59,8 +70,26 @@ impl WiringLayer for HouseKeeperLayer {
             replica_pool.clone(),
         );
 
+        let app_health = input.app_health.0;
+        app_health
+            .insert_custom_component(Arc::new(NodeVersionInfo::default()))
```
Deniallugo, line 75: I think we have to include it into every component, not only to house keeper

core/node/eth_sender/src/health.rs
```diff
@@ -0,0 +1,84 @@
...
+
+impl From<AggregatedActionType> for ActionType {
+    fn from(action_type: AggregatedActionType) -> Self {
+        match action_type {
+            AggregatedActionType::Commit => Self::Commit,
+            AggregatedActionType::PublishProofOnchain => Self::PublishProofOnchain,
+            AggregatedActionType::Execute => Self::Execute,
+        }
+    }
+}
```
Deniallugo, line 24: why do you need a new enum here?

core/node/node_framework/src/implementations/layers/house_keeper.rs
slowli: I.e., include it in a separate layer, right? (I'd suggest to extend `PostgresMetricsLayer`, renaming it to `PostgresLayer`.)

core/node/house_keeper/src/database.rs
```diff
@@ -0,0 +1,45 @@
...
+    fn from(details: DatabaseInfo) -> Self {
+        Self::from(HealthStatus::Ready).with_details(details)
+    }
+}
+
+#[derive(Debug)]
+pub struct DatabaseHealthTask {
+    pub polling_interval_ms: u64,
+    pub connection_pool: ConnectionPool<Core>,
+    pub database_health_updater: HealthUpdater,
```
slowli, line 23: Bikeshedding: To make encapsulation better, I'd suggest to make the fields private and use a constructor taking `connection_pool` and `polling_interval_ms` (a health check can be obtained via a getter, as with other components).

core/node/state_keeper/src/keeper.rs
```diff
@@ -65,48 +67,58 @@ impl Error {
...
 
-    pub async fn run(mut self) -> anyhow::Result<()> {
-        match self.run_inner().await {
+    pub async fn run(mut self, stop_receiver: watch::Receiver<bool>) -> anyhow::Result<()> {
+        match self.run_inner(stop_receiver).await {
             Ok(_) => unreachable!(),
-            Err(Error::Fatal(err)) => Err(err).context("state_keeper failed"),
+            Err(Error::Fatal(err)) => {
+                self.health_updater
+                    .update(Health::from(HealthStatus::ShuttingDown));
```
slowli, line 101: Nit: It's really redundant to set the `ShuttingDown` status just before the `HealthUpdater` is dropped, since it'll immediately be overwritten as `ShutDown`. `ShuttingDown` makes sense to set before non-trivial shut down logic.

core/node/state_keeper/src/lib.rs
```diff
@@ -12,6 +12,7 @@ pub use self::{
 };
 
 pub mod executor;
+pub mod health;
```
slowli, line 15: Nit: Does this module need to be public?

core/node/state_keeper/src/health.rs
```diff
@@ -0,0 +1,30 @@
...
+use zksync_health_check::{Health, HealthStatus};
+use zksync_types::{L1BatchNumber, L2BlockNumber, H256};
+
+use crate::io::IoCursor;
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct StateKeeperHealthDetails {
+    pub next_l2_block: L2BlockNumber,
+    pub prev_l2_block_hash: H256,
+    pub prev_l2_block_timestamp: u64,
```
slowli, line 11: Bikeshedding: Do these 2 fields provide valuable info? IMO, it'd make more sense to provide info like the number of txs in the current block.

core/node/node_framework/src/implementations/layers/state_keeper/mod.rs
```diff
@@ -123,11 +133,32 @@ impl WiringLayer for StateKeeperLayer {
...
-    io: Box<dyn StateKeeperIO>,
-    executor_factory: Box<dyn BatchExecutorFactory<OwnedStorage>>,
-    output_handler: OutputHandler,
-    sealer: Arc<dyn ConditionalSealer>,
-    storage_factory: Arc<dyn ReadStorageFactory>,
+    state_keeper: ZkSyncStateKeeper,
+}
+
+impl StateKeeperTask {
+    pub fn new(
```
slowli, line 140: Nit: This constructor probably doesn't need to be public (so it can be removed altogether in favor of constructing `state_keeper` at the only callsite.

slowli (commented):
Nice 👍 Still have a couple of comments, but overall the PR looks significantly better than previously.

core/lib/protobuf_config/src/proto/config/house_keeper.proto
```diff
@@ -17,4 +17,5 @@ message HouseKeeper {
     reserved 15; reserved "prover_job_archiver_archive_after_secs";
     reserved 16; reserved "fri_gpu_prover_archiver_archiving_interval_ms";
     reserved 17; reserved "fri_gpu_prover_archiver_archive_after_secs";
+    optional uint64 database_health_polling_interval_ms = 18; // required; ms
```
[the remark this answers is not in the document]
manuelmauro: Addressed in https://github.com/matter-labs/zksync-era/pull/3193/commits/8d4c17617317586cca9af1993ca2670d0461a540

core/node/node_framework/src/implementations/layers/house_keeper.rs
manuelmauro: Moved database health task to PostgresLayer in https://github.com/matter-labs/zksync-era/pull/3193/commits/5e235b13c54baac79e8dbc05b0bdd59c1650b407

core/node/state_keeper/src/keeper.rs
manuelmauro: Addressed in https://github.com/matter-labs/zksync-era/pull/3193/commits/53cca2fb8d510e0cea578a02c73a987bb48909be

core/node/state_keeper/src/lib.rs
manuelmauro: Addressed in https://github.com/matter-labs/zksync-era/pull/3193/commits/9fe98d2fd980f07bc741178d93fb07d78099b7ee

core/node/node_framework/src/implementations/layers/state_keeper/mod.rs
manuelmauro: Addressed in https://github.com/matter-labs/zksync-era/pull/3193/commits/36ad7c5c2c4247510c52767f41f08ee7591d02cc

core/node/house_keeper/src/database.rs
manuelmauro: Addressed in https://github.com/matter-labs/zksync-era/pull/3193/commits/36ad7c5c2c4247510c52767f41f08ee7591d02cc and https://github.com/matter-labs/zksync-era/pull/3193/commits/1585d5dfbbe206e2d67a52c52d1b2910b244f409

core/node/house_keeper/src/version.rs
```diff
@@ -0,0 +1,39 @@
+use async_trait::async_trait;
+use serde::{Deserialize, Serialize};
+use zksync_health_check::{CheckHealth, Health, HealthStatus};
+
+const GIT_VERSION: &str = zksync_git_version_macro::build_git_revision!();
```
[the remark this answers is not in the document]
manuelmauro: RFC on https://github.com/matter-labs/zksync-era/pull/3193/commits/26aa8241480446c6df6f335d293c4bbb0e01dacd and https://github.com/matter-labs/zksync-era/pull/3193/commits/f06a4158fb041794fbe24df7ebfd664bc6549b8c

I added git version information to `shared_metrics` as suggested, renamed the involved structs, and directly integrated this health check component into `healthcheck_server`. I'd appreciate your thoughts, particularly regarding the integration within `healthcheck_server`.

Deniallugo (approved):
LGTM, but wanna approve from @slowli

core/node/state_keeper/src/health.rs
Deniallugo: > the number of txs in the current block

I don't feel this info is somehow interesting. We have normally 2-3 tps  so we have to update it pretty frequent

core/node/node_framework/src/implementations/layers/postgres.rs
```diff
@@ -0,0 +1,186 @@
...
+    service::StopReceiver,
+    task::{Task, TaskId, TaskKind},
+    wiring_layer::{WiringError, WiringLayer},
+    FromContext, IntoContext,
+};
+
+const SCRAPE_INTERVAL: Duration = Duration::from_secs(60);
+
+#[derive(Debug, Deserialize, Clone, PartialEq)]
+pub struct Config {
```
slowli, line 24: This config looks redundant; it's not used and is always initialized to a default value. Also, you can probably reuse `SCRAPE_INTERVAL` for the new task.

core/node/shared_metrics/src/binary.rs
```diff
@@ -0,0 +1,58 @@
...
+use serde::Serialize;
+use vise::{EncodeLabelSet, Info, Metrics};
+
+mod values {
+    use super::BinMetadata;
+    include!(concat!(env!("OUT_DIR"), "/metadata_values.rs"));
+}
+
+pub use values::BIN_METADATA;
+use zksync_health_check::{CheckHealth, Health, HealthStatus};
```
slowli, line 11: - Please don't split `use`s with `mod` declarations; it'll lead to bogus import grouping (like here `zksync_health_check` belongs to external imports).
- Please prepend `self::` for internal imports so that they are easy to distinguish from external ones.

core/node/shared_metrics/build.rs
```diff
@@ -4,43 +4,75 @@ use std::{
...
+
+/// Outputs the current git commit hash as a string literal.
+pub fn git_revision() -> String {
+    run_cmd("git", &["rev-parse", "--short", "HEAD"])
+}
+
+/// Tries to run the command, only returns `Some` if the command
+/// succeeded and the output was valid utf8.
+fn run_cmd(cmd: &str, args: &[&str]) -> String {
+    run_cmd_opt(cmd, args).unwrap_or("unknown".to_string())
```
slowli, line 56: Bikeshedding: Is "unknown" a better value than no value at all (i.e., `None`)? IMO, the latter is more idiomatic, and it fits better at least for metrics.

core/node/shared_metrics/src/binary.rs
```diff
@@ -0,0 +1,58 @@
...
+        self.info.set(BIN_METADATA).ok();
+    }
+}
+
+#[vise::register]
+pub static BIN_METRICS: vise::Global<BinMetrics> = vise::Global::new();
+
+impl From<&BinMetadata> for Health {
+    fn from(details: &BinMetadata) -> Self {
+        Self::from(HealthStatus::Ready).with_details(details)
```
slowli, line 45: Future work (?): It's weird seeing this reported as a component, since this data clearly doesn't have its own lifecycle (not even a sort of surrogate one like Postgres does). It'd probably make sense to add a way to set details for `AppHealth` (note that it contains `Health` already).

core/node/state_keeper/src/keeper.rs
```diff
@@ -104,9 +105,14 @@ impl ZkSyncStateKeeper {
...
     /// Fallible version of `run` routine that allows to easily exit upon cancellation.
-    async fn run_inner(&mut self) -> Result<Infallible, Error> {
+    async fn run_inner(
+        &mut self,
+        mut stop_receiver: watch::Receiver<bool>,
+    ) -> Result<Infallible, Error> {
         let (cursor, pending_batch_params) = self.io.initialize().await?;
         self.output_handler.initialize(&cursor).await?;
+        self.health_updater
+            .update(Health::from(HealthStatus::Ready));
```
slowli, line 115: - Why not update health from `cursor` immediately, like you do below?
- To be really nitpicky, one could argue that the state keeper is affected until it processes the pending batch, since it doesn't respond to environment signals (mempool txs and timeouts) until then. This is subjective though, so feel free to ignore.

slowli (commented):
Mostly looks good. If this functionality isn't urgent, I'd still prefer at least non-nit / bikeshedding comments be addressed in this PR.

core/lib/dal/src/system_dal.rs
```diff
@@ -12,6 +14,16 @@ pub(crate) struct TableSize {
...
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+pub struct DatabaseMigration {
+    pub version: i64,
+    pub description: String,
+    pub installed_on: DateTime<chrono::Utc>,
+    pub success: bool,
+    pub checksum: String,
+    pub execution_time: i64,
```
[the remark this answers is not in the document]
manuelmauro: Addressed in https://github.com/matter-labs/zksync-era/pull/3193/commits/9964ee7d62abf689bae57b45e5e0617c1d1e2042

core/node/node_framework/src/implementations/layers/postgres.rs
manuelmauro: Addressed in https://github.com/matter-labs/zksync-era/pull/3193/commits/19275b6681b6a1ff7fe84f6a11a088d6d4a137f2

core/node/shared_metrics/build.rs
manuelmauro: Addressed in https://github.com/matter-labs/zksync-era/pull/3193/commits/6bc90e2f8364418fb8632548d59e397bf94ca29d

core/node/eth_sender/src/health.rs
```diff
@@ -0,0 +1,67 @@
...
+            created_at_timestamp: tx.created_at_timestamp,
+            predicted_gas_cost: tx.predicted_gas_cost,
+        }
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct EthTxManagerHealthDetails {
+    pub last_mined_tx: EthTxDetails,
+    pub tx_status: TxStatus,
```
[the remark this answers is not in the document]
manuelmauro: Merged `TxStatus` into `EthTxDetails` in https://github.com/matter-labs/zksync-era/pull/3193/commits/25d8b26951f58b0be76ed85afa9dccc9b47e1f5f

core/node/state_keeper/src/keeper.rs
manuelmauro: Addressed in https://github.com/matter-labs/zksync-era/pull/3193/commits/d2c983c9e3685e7958bdf4a40c82ada48090889c

core/node/shared_metrics/src/binary.rs
Deniallugo: I agree, looks like we have to include it general AppHealth report

Deniallugo (commented):
generally lgtm, please fix the last comment from @slowli

core/lib/health_check/Cargo.toml
```diff
@@ -20,6 +20,7 @@ serde_json.workspace = true
 thiserror.workspace = true
 tokio = { workspace = true, features = ["sync", "time"] }
 tracing.workspace = true
+zksync_bin_metadata.workspace = true
```
slowli, line 23: This dependency looks out of place; it makes a more-or-less isolated library depend on a library with a build script etc. An inverse dependency (`bin_metadata` depending on `health_check`) would make more sense.

core/lib/bin_metadata/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+    pub rustc_commit_date: Option<&'static str>,
+    pub rustc_channel: &'static str,
+    pub host: &'static str,
+    pub llvm: Option<&'static str>,
+    pub git_branch: Option<&'static str>,
+    pub git_revision: Option<&'static str>,
+}
+
+#[derive(Debug, Metrics)]
+#[metrics(prefix = "rust")]
```
slowli, line 25: Eh, new `git_*` metrics probably shouldn't be grouped into the `rust_info` INFO metric. Optimally, I'd add a new INFO metric with these two labels and name it like `git_info` or `version_control_info`.

slowli (approved):
Comments are nits; I've left them mostly to be addressed in a follow-up PR. I can produce such a PR once I've got less swamped with tasks 🙃

manuelmauro:
> Comments are nits; I've left them mostly to be addressed in a follow-up PR. I can produce such a PR once I've got less swamped with tasks 🙃

No worries! I can address the here. 👍 Thanks for the review!

core/lib/health_check/Cargo.toml
manuelmauro: I see your point here. The problem with inverting the dependency is that I wouldn't be able to add `BIN_METADATA` at the moment I create the inner `Health` instance:

```rust
let inner = Health::with_details(aggregated_status.into(), BIN_METADATA);
```

WDYT?

core/lib/bin_metadata/src/lib.rs
manuelmauro: Addressed in https://github.com/matter-labs/zksync-era/pull/3193/commits/b6bdda43ed6164bccc3025e95f5f95c469501306

Deniallugo (approved):

PR #3200, opened 2024-10-30
fix(merkle-tree): Repair stale keys for tree in background

slowli:
## What ❔

Implements a background task to remove bogus stale keys for the Merkle tree.

## Why ❔

These keys could have been produced during L1 batch reverts before https://github.com/matter-labs/zksync-era/pull/3178.

## Checklist


core/bin/external_node/src/config/mod.rs
```diff
@@ -382,6 +382,9 @@ pub(crate) struct OptionalENConfig {
     /// Timeout to wait for the Merkle tree database to run compaction on stalled writes.
     #[serde(default = "OptionalENConfig::default_merkle_tree_stalled_writes_timeout_sec")]
     merkle_tree_stalled_writes_timeout_sec: u64,
+    /// Enables the stale keys repair task for the Merkle tree.
+    #[serde(default)]
+    pub merkle_tree_repair_stale_keys: bool,
```
slowli, line 387: I'm not sure whether it's worth starting with ENs or the main node to test the task (bogus stale keys could theoretically be present in either case). EN looks like a less risky option.

core/lib/merkle_tree/src/repair.rs
```diff
@@ -0,0 +1,404 @@
...
+        }
+        stats.repaired_key_count += step_stats.repaired_key_count;
+    }
+
+    /// Runs this task indefinitely.
+    ///
+    /// # Errors
+    ///
+    /// Propagates RocksDB I/O errors.
+    pub fn run(mut self) -> anyhow::Result<()> {
```
slowli, line 257: Currently, the task will run indefinitely checking stale keys for new versions as they are added to the tree. An alternative would be to to persist the latest tree version when the task is started the first time, and stop the task when this version is reached. The reasoning is that stale keys should be correct after this version due to https://github.com/matter-labs/zksync-era/pull/3178.

perekopskiy (approved):

PR #3204, opened 2024-10-30
chore: make artificially shortened executions count as successes

joonazan:
Currently, successful validations are reported as reverts, which is super confusing.

Why this is safe:
- `TracerExecutionStopReason::Finished` is only emitted on success, Errors always emit Abort instead
- Some of the removed code was just dead code. `self.result` was checked twice.

I'd rather fix the strange behaviour than emulate it to make ShadowVm line up.

### DIFF
--- a/core/lib/multivm/src/versions/vm_latest/tracers/default_tracers.rs
+++ b/core/lib/multivm/src/versions/vm_latest/tracers/default_tracers.rs
@@ -230,7 +230,7 @@ impl<S: WriteStorage, H: HistoryMode> Tracer for DefaultExecutionTracer<S, H> {
         match hook {
             VmHook::TxHasEnded if matches!(self.execution_mode, VmExecutionMode::OneTx) => {
                 self.result_tracer.tx_finished_in_one_tx_mode = true;
-                self.tx_has_been_processed = true
+                self.tx_has_been_processed = true;
             }
             VmHook::NoValidationEntered => self.in_account_validation = false,
             VmHook::AccountValidationEntered => self.in_account_validation = true,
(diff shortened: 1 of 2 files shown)


joonazan:
Turns out that `TracerExecutionStopReason::Finished` is produced when a transaction ends even though the tracer can't know at that point if the bootloader ran out of gas before finishing the whole thing. In that case the old code that reports "maybe out of gas" is correct but we need to run it only in that special case.

I'm not sure what the best way to accomplish that is. We could maybe change `TracerExecutionStopReason` to have a success, failure and tx ended variant. I think I prefer giving `DefaultExecutionTracer` enough info to be able to detect failure.

Also, it would be nice if the bootloader had "did not finish" as the default value of the result instead of failure.

joonazan:
I ended up with a pretty hacky fix. I would have preferred to make the validation tracer tell the result tracer that there has been no out of gas but that isn't done as easily.

perekopskiy (approved):

core/lib/multivm/src/versions/vm_latest/tracers/default_tracers.rs
```diff
@@ -228,7 +228,10 @@ impl<S: WriteStorage, H: HistoryMode> Tracer for DefaultExecutionTracer<S, H> {
         );
 
         match hook {
-            VmHook::TxHasEnded => self.tx_has_been_processed = true,
+            VmHook::TxHasEnded if matches!(self.execution_mode, VmExecutionMode::OneTx) => {
+                self.result_tracer.tx_finished_in_one_tx_mode = true;
+                self.tx_has_been_processed = true
```
slowli, line 233: Nit: Missing a semicolon.

slowli (approved):
> Also, it would be nice if the bootloader had "did not finish" as the default value of the result instead of failure.

Wouldn't help with historical bootloader versions (IIUC, they may be used in the oneshot executor).

slowli (approved):

PR #3210, opened 2024-10-31
feat(zk_toolbox): Make chain create independent from ecosystem

matias-gonz:
## Checklist


zkstack_cli/crates/zkstack/src/commands/chain/create.rs
```diff
@@ -13,77 +14,111 @@ use crate::{
...
     let (chain_id, legacy_bridge) = if args.legacy_bridge {
         // Legacy bridge is distinguished by using the same chain id as ecosystem
-        (ecosystem_config.era_chain_id, Some(true))
+        (args.era_chain_id, Some(true))
     } else {
         (L2ChainId::from(args.chain_id), None)
     };
-    let internal_id = ecosystem_config.list_of_chains().len() as u32;
+
+    let internal_id = args.number_of_chains + 1;
```
Deniallugo, line 100: it's always zero in a case we don't have ecosystem config

Deniallugo (commented):
generally looks good, please have a look closely to method visibility. 
Looks like we have a lot of unnecessary public methods

zkstack_cli/crates/zkstack/src/commands/chain/args/create.rs
```diff
@@ -106,8 +108,14 @@ impl ChainCreateArgs {
                     .ask()
             });
 
+        let l1_network = l1_network.unwrap_or_else(|| {
+            self.l1_network.unwrap_or_else(|| {
+                PromptSelect::new(MSG_L1_NETWORK_PROMPT, L1Network::iter()).ask()
+            })
+        });
```
Deniallugo, line 115: Looks a bit odd. Why do we have two places for l1 network?

matias-gonz: `l1_network` as a parameter of the function is the one coming from ecosystem config. If exists we use that one. `l1_network` as attribute (`self.l1_network`) comes from the command arguments: provided by the user. If there is no ecosystem and the user didn't provide `l1_network` then there is a prompt

PR #3225, opened 2024-11-04
feat(contract-verifier): Support Solidity contracts with EVM bytecode in contract verifier

slowli:
## What ❔

Supports verifying Solidity contracts deployed with EVM bytecode in the contract verifier.

## Why ❔

Part of supporting EVM emulation throughout the Era codebase.

## Checklist


core/lib/types/src/contract_verification_api.rs
```diff
@@ -157,7 +157,7 @@ pub enum CompilerType {
 pub enum CompilerVersions {
     #[serde(rename_all = "camelCase")]
     Solc {
-        compiler_zksolc_version: String,
+        compiler_zksolc_version: String, // FIXME: optional?
```
slowli, line 160: Logically, it would make sense to not specify `compiler_zksolc_version` for contracts with EVM bytecode. This requires a DB migration etc., so I'd prefer to implement it in a follow-up PR.

core/lib/contract_verifier/src/compilers/solc.rs
```diff
@@ -0,0 +1,168 @@
...
+                let source = Source {
+                    content: source_code,
+                };
+                let sources = HashMap::from([(file_name.clone(), source)]);
+                let settings = Settings {
+                    output_selection: Some(default_output_selection),
+                    other: serde_json::json!({
+                        "optimizer": {
+                            "enabled": req.optimization_used,
+                        },
```
slowli, line 82: AFAIU, contract verifiers like Etherscan commonly allow specifying the intended number of runs for the optimizer (200 by default), but this setting doesn't seem to be present in the verification DTOs rn. Again, I'd prefer to add its support (if necessary) separately since this PR is large as-is.

popzxc: It is certainly required, but indeed it makes sense to handle it separately.

popzxc (approved):

core/lib/contract_verifier/src/compilers/solc.rs
perekopskiy: I'm not sure if we should keep `SolSingleFile` type of endpoint. Probably it's better to fully deprecate it, and let wrapping request for single-file happen on client side (hardhat plugin already doesn't use `SolSingleFile` for a long time, explorer uses it but it should be easy to switch, I'm not sure if it's used in foundry and etherscan-like API

perekopskiy (approved):

PR #3232, opened 2024-11-06
test(vm): Test EVM emulator in `multivm` crate

slowli:
## What ❔

Adds basic low-level tests for the real EVM emulator.

## Why ❔

EVM emulator functionality should be covered by tests.

## Checklist


etc/contracts-test-data/evm-contracts/evm.sol
```diff
@@ -0,0 +1,328 @@
...
+        require(counter.get() == _expectedInitialValue + 3, "counter.get() after revert");
+    }
+
+    uint private constant GAS_TO_SEND = 100000;
+
+    function testFarCalls(IGasTester _tester, bool _isEvmTester) external view validEvmCall {
+        uint gasMultiplier = _isEvmTester ? 1 : 5;
+        uint currentGas = gasleft();
+        if (_isEvmTester) {
+            // FIXME: doesn't work for EraVM tester (~115k less gas is passed)
```
0xVolosnikov, line 263: It should be decommit cost

slowli: I've thought about it as well. Is this cost emulated by the EVM emulator? My confusion with this explanation is that it doesn't match how decommits are priced in EraVM; the amount of the subtracted gas doesn't depend on whether the called contact is "cold" or "hot" (i.e., was previously decommitted; IIUC, in the latter case, the decommit cost [should be 0](https://github.com/matter-labs/vm2/blob/457d8a7eea9093af9440662e33e598c13ba41633/crates/vm2/src/decommit.rs#L77-L83)). Note that *both* calls with default gas passed must be switched off; the second one calls a hot contract.

0xVolosnikov: Yes, this was implemented to somehow charge for decommits when we have EVM stipend (charge from callee frame). So it was supposed to charge even if we are calling already "hot" contract, because EVM doesn't know about hot/cold accounts in EraVM context.

But if we remove EVM stipend, we can remove this as well, I guess. Decommit cost will be charged from the caller frame

perekopskiy (approved):

PR #3236, opened 2024-11-07
feat(vm): add gateway changes to fast vm

perekopskiy:
## What ❔

Ports VM changes to vm_fast that are needed for gateway version. Code changes are mostly copied from vm_latest

## Why ❔

vm_fast should support new protocol version

## Checklist


core/lib/multivm/src/versions/vm_fast/version.rs
```diff
@@ -0,0 +1,28 @@
+use crate::{vm_latest::MultiVMSubversion, VmVersion};
+
+#[derive(Debug, Copy, Clone)]
+pub(crate) enum FastVMVersion {
```
slowli, line 4: Bikeshedding: `FastVmVersion` seems more in line with Rust naming conventions.

core/lib/multivm/src/versions/vm_fast/vm.rs
```diff
@@ -353,15 +355,20 @@ impl<S: ReadStorage, Tr: Tracer + Default> Vm<S, Tr> {
...
+                        &mut memory_to_apply,
+                        pubdata_builder.expect("`pubdata_builder` is required to finish batch"),
+                        &pubdata_input,
+                        self.system_env.version,
+                    );
                     self.write_to_bootloader_heap(memory_to_apply);
+
+                    // Save the pubdata for the future initial bootloader memory building
+                    self.bootloader_state
+                        .set_pubdata_input(pubdata_input.clone());
```
slowli, line 371: Nit: `clone()` isn't necessary here.

Also, since `pubdata_input` is almost only read from the VM, so it could be relatively straightforward to return `pubdata_input` from this function (e.g., by supplying `pubdata: Option<(&dyn PubdataBuilder, &mut PubdataInput)>` to it), and then use it in `finish_batch()`. I'm not sure a similar transform would work for `vm_latest` though, and we'd probably want to keep `BootloaderState` equivalent for both (ideally, it'd make sense to share it).

slowli (approved):
Would it be possible to unit-test pubdata logic in `multivm`? (I'd imagine by setting `ProtocolVersionId` to the gateway version.) AFAICT, pubdata-related fields are covered by shadow VM checks, so these tests could help checking that the relevant logic works identically for `vm_latest` and `vm_fast`. Can be done separately, of course.

core/lib/multivm/src/versions/vm_fast/version.rs
perekopskiy: I've also renamed other types

core/lib/multivm/src/versions/vm_fast/vm.rs
perekopskiy: Removed clone.
Not sure I understood the point for returning `pubdata_input`, can you elaborate on why is it better than storing and reading it from `BootloaderState`?

perekopskiy:
> Would it be possible to unit-test pubdata logic in multivm?

Do you mean something particular? Rollup/Validium pubdata builders' logic is covered with tests in `pubdata_builders` module. We also have a vm test in sync-layer-stable that checks pubdata_inputs more thoroughly, it will be merged to main soonish.

PR #3251, opened 2024-11-11
feat(contract-verifier): Support Vyper toolchain for EVM bytecodes

slowli:
## What ❔

- Supports the Vyper toolchain for EVM bytecodes in the contract verifier.
- Adds some more unit tests for the verifier (e.g., testing multi-file inputs, Yul contracts, abstract contract errors etc.).

## Why ❔

Part of preparations to support EVM bytecodes throughout the codebase.

## Checklist

### DIFF
--- a/core/lib/contract_verifier/src/compilers/mod.rs
+++ b/core/lib/contract_verifier/src/compilers/mod.rs
@@ -117,8 +117,12 @@ fn parse_standard_json_output(
         None
     };
 
-    let abi = contract["abi"].clone();
-    if !abi.is_array() {
+    let mut abi = contract["abi"].clone();
+    if abi.is_null() {
+        // ABI is undefined for Yul contracts when compiled with standalone `solc`. For uniformity with `zksolc`,
+        // replace it with an empty array.
+        abi = serde_json::json!([]);
+    } else if !abi.is_array() {
         let err = anyhow::anyhow!(
             "unexpected value for ABI: {}",
             serde_json::to_string_pretty(&abi).unwrap()
--- a/core/lib/contract_verifier/src/compilers/mod.rs
+++ b/core/lib/contract_verifier/src/compilers/mod.rs
@@ -1,17 +1,39 @@
+use std::collections::HashMap;
+
 use anyhow::Context as _;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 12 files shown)


core/lib/contract_verifier/src/compilers/mod.rs
```diff
@@ -31,11 +75,17 @@ fn parse_standard_json_output(
         let errors = errors.as_array().unwrap().clone();
         if errors
             .iter()
-            .any(|err| err["severity"].as_str().unwrap() == "error")
+            .any(|err| err["severity"].as_str() == Some("error"))
         {
+            // FIXME: collected messages contain warnings as well; is this intentional?
```
slowli, line 80: Subj. On the one hand, it sort of makes sense to include warnings from the DevEx perspective, but the fact that they are not distinguished from errors (both are stored as message strings) is somewhat confusing.

core/lib/contract_verifier/src/compilers/vyper.rs
```diff
@@ -0,0 +1,115 @@
...
+        let sources = match req.source_code_data {
+            SourceCodeData::VyperMultiFile(s) => s,
+            other => unreachable!("unexpected `SourceCodeData` variant: {other:?}"),
+        };
+        Ok(Self {
+            contract_name,
+            file_name,
+            sources,
+            optimizer_mode: if req.optimization_used {
+                // FIXME: values have differing semantics compared to other compilers; is this OK?
```
slowli, line 34: Subj. It probably would make sense to make the optimizer mode dependent on the specified compilers on the type level, or at least check correspondence in the API.

core/lib/contract_verifier/src/compilers/mod.rs
popzxc: I think it's OK for now to not spend extra time on it. I don't think it was intentional, but I also don't think it's a big issue.

core/lib/contract_verifier/src/compilers/vyper.rs
popzxc: Probably; but the same as with above issue -- it works today, so, it's probably not that big of an issue.

popzxc (dismissed):

perekopskiy (approved):

PR #3253, opened 2024-11-11
feat(en): remove `user_facing_*` addresses

perekopskiy:
## What ❔

for en: 
- renames prefixes: `user_facing_` -> `l1_`
- `ExternalNodeConfig::l1_diamond_proxy_address` is used instead of remote.diamon_proxy

for main node:
- remove `user_facing_diamond_proxy` and `user_facing_bridgehub` as duplicates for corresponding fields from l1 contracts config

## Why ❔

Renaming for clarity (IMO l1_ prefix providers better understanding of what is put under variable)
Removing for config simplicity and avoiding duplication

## Checklist

### DIFF
--- a/core/bin/external_node/src/config/mod.rs
+++ b/core/bin/external_node/src/config/mod.rs
@@ -101,11 +101,11 @@ impl ConfigurationSource for Environment {
 /// This part of the external node config is fetched directly from the main node.
 #[derive(Debug, Deserialize)]
 pub(crate) struct RemoteENConfig {
-    pub bridgehub_proxy_addr: Option<Address>,
-    pub state_transition_proxy_addr: Option<Address>,
-    pub transparent_proxy_admin_addr: Option<Address>,
-    /// Should not be accessed directly. Use [`ExternalNodeConfig::diamond_proxy_address`] instead.
-    pub user_facing_diamond_proxy: Address,
+    pub l1_bridgehub_proxy_addr: Option<Address>,
+    pub l1_state_transition_proxy_addr: Option<Address>,
+    pub l1_transparent_proxy_admin_addr: Option<Address>,
+    /// Should not be accessed directly. Use [`ExternalNodeConfig::l1_diamond_proxy_address`] instead.
+    pub l1_diamond_proxy_addr: Address,
     // While on L1 shared bridge and legacy bridge are different contracts with different addresses,
     // the `l2_erc20_bridge_addr` and `l2_shared_bridge_addr` are basically the same contract, but with
     // a different name, with names adapted only for consistency.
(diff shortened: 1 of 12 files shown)


core/bin/external_node/src/config/mod.rs
```diff
@@ -101,11 +101,11 @@ impl ConfigurationSource for Environment {
 /// This part of the external node config is fetched directly from the main node.
 #[derive(Debug, Deserialize)]
 pub(crate) struct RemoteENConfig {
-    pub bridgehub_proxy_addr: Option<Address>,
-    pub state_transition_proxy_addr: Option<Address>,
-    pub transparent_proxy_admin_addr: Option<Address>,
-    /// Should not be accessed directly. Use [`ExternalNodeConfig::diamond_proxy_address`] instead.
-    pub user_facing_diamond_proxy: Address,
+    pub l1_bridgehub_proxy_addr: Option<Address>,
+    pub l1_state_transition_proxy_addr: Option<Address>,
```
StanislavBreadless, line 105: IIRC this contract is used to read the preimage for the upgrade. The GW one should be used

core/bin/external_node/src/config/mod.rs
```diff
@@ -101,11 +101,11 @@ impl ConfigurationSource for Environment {
...
+    pub l1_transparent_proxy_admin_addr: Option<Address>,
```
StanislavBreadless, line 106: since CTM is on GW, maybe this one should be on GW too

core/bin/external_node/src/config/mod.rs
StanislavBreadless: On a second thought, the L2 parts of both should be same, so reading just from L1 one is fine, it is just that only the SL one will contain the addresses of the new contracts (such as mailbox, verifier, etc), but these are not used anyway

perekopskiy: You're right, but it's a bug in eth watch, we should use different address there. I'll fix it separately

core/bin/external_node/src/config/mod.rs
perekopskiy: EN needs these 3 addresses only to expose them in `en_getEcosystemContracts` and `zks_getBridgehubContract`. This wasn't discussed explicitly but I think we should keep them returning L1 addresses. GW ecosystem addresses can be queried from gateway (CTM, DP) or are constant (BH)

StanislavBreadless (approved):

core/bin/external_node/src/config/mod.rs
perekopskiy: > On a second thought, the L2 parts of both should be same, so reading just from L1 one is fine.

Eth watch is still broken, it calls L1 CTM with GW client.

PR #3276, opened 2024-11-14
feat: add more metrics for the tee_prover

haraldh:
## What ❔

Add a `sealed_at` column to the `l1_batches` table and extend the tee_prover metrics.

## Why ❔

The duration metric of sealed -> TEE proven is of special interest for Interop. It should not regress and should be optimized in future development.

## Checklist


core/lib/dal/src/blocks_dal.rs
```diff
@@ -2394,6 +2396,29 @@ impl BlocksDal<'_, '_> {
         .flatten())
     }
 
+    pub async fn get_sealed_at(
```
slowli, line 2399: Bikeshedding: Despite the argument making it clear what this method is about, I'd still name it more specifically, like `get_batch_sealed_at()`; other methods in this module usually specify entities they concern.

core/lib/dal/src/blocks_dal.rs
```diff
@@ -2394,6 +2396,29 @@ impl BlocksDal<'_, '_> {
...
+        l1_batch_number: L1BatchNumber,
+    ) -> DalResult<Option<NaiveDateTime>> {
+        Ok(sqlx::query!(
+            r#"
+            SELECT
+                sealed_at
+            FROM
+                l1_batches
+            WHERE
+                is_sealed
```
slowli, line 2410: Isn't this condition redundant? If `!is_sealed`, `sealed_at` should be null. OTOH, it could still be null for old sealed batches; I think this is accounted for, right?

core/node/proof_data_handler/src/tee_request_processor.rs
```diff
@@ -208,6 +204,28 @@ impl TeeRequestProcessor {
...
+        let duration_secs_f64 = if let Some(duration) = duration {
+            METRICS.tee_proof_roundtrip_time[&proof.0.tee_type.into()].observe(duration);
+            duration.as_secs_f64()
+        } else {
+            f64::NAN
+        };
+
+        tracing::info!(
+            l1_batch_number = %l1_batch_number,
+            sealed_to_proven_in_secs = %duration_secs_f64,
```
slowli, line 224: Nit: `f64` values can be supplied to `tracing` macros directly, w/o a `Display` / `%` qualifier.

core/bin/zksync_tee_prover/src/tee_prover.rs
```diff
@@ -80,7 +80,13 @@ impl TeeProver {
                 let msg_to_sign = Message::from_slice(root_hash_bytes)
                     .map_err(|e| TeeProverError::Verification(e.into()))?;
                 let signature = self.config.signing_key.sign_ecdsa(msg_to_sign);
-                observer.observe();
+                let duration = observer.observe();
+                tracing::info!(
+                    proof_generation_time = %duration.as_secs_f64(),
+                    l1_batch_number = %batch_number,
+                    l1_root_hash = %verification_result.value_hash,
```
slowli, line 87: To record the full hash, use `?` (= `Debug`) specifier. AFAIU, the `Display` impl for `H256` doesn't display all hex digits, but rather only starting / ending ones.

core/lib/dal/src/blocks_dal.rs
```diff
@@ -2394,6 +2396,29 @@ impl BlocksDal<'_, '_> {
         .flatten())
     }
 
+    pub async fn get_sealed_at(
+        &mut self,
+        l1_batch_number: L1BatchNumber,
+    ) -> DalResult<Option<NaiveDateTime>> {
```
slowli, line 2402: Bikeshedding: Most existing DAL methods convert time to `DateTime<Utc>`.

haraldh: fixed

core/bin/zksync_tee_prover/src/tee_prover.rs
haraldh: fixed

core/node/proof_data_handler/src/tee_request_processor.rs
haraldh: done

core/lib/dal/src/blocks_dal.rs
haraldh: removed

core/lib/dal/src/blocks_dal.rs
haraldh: fixed

PR #3280, opened 2024-11-14
docs(consensus): instructions for enabling consensus on the main node

pompon0:


brunoffranca (dismissed):
I think I have what I need. Thanks!

brunoffranca (approved):

PR #3285, opened 2024-11-14
docs: update short guide of dev setup

brunoffranca:
## What ❔

- Removes the HTTPS configuration for git. Instead use HTTPS by default since that's Github recommendation.
- Install git along with the other necessary packages.
- Adds `--locked` flag to cargo-nextest installation since that's the recommended way.
- Automatically reload the terminal after installing yarn, instead of telling the user to do so.
- Documented that the usermod change is optional (only to run Docker without sudo).
- Automatically reload bashrc after non-cuda setup, instead of telling the user to do so.
- Changed installation steps order to make more sense (ex: we were installing curl after using it).

## Why ❔

Simplifies the guide and fixes some minor issues.

### DIFF
--- a/docs/guides/setup-dev.md
+++ b/docs/guides/setup-dev.md
@@ -8,54 +8,49 @@ initializing the workspace, it's recommended that you read the whole guide below
 If you run on 'clean' Ubuntu on GCP:
 
 ```bash
-# For VMs only! They don't have SSH keys, so we override SSH with HTTPS
-git config --global url."https://github.com/".insteadOf git@github.com:
-git config --global url."https://".insteadOf git://
+# All necessary stuff
+sudo apt update
+sudo apt install --yes git build-essential pkg-config cmake clang lldb lld libssl-dev libpq-dev apt-transport-https ca-certificates curl software-properties-common
 
 # Rust
 curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
-
-# NVM
-curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.5/install.sh | bash
-
-# All necessary stuff
-sudo apt-get update
-sudo apt-get install -y build-essential pkg-config cmake clang lldb lld libssl-dev libpq-dev apt-transport-https ca-certificates curl software-properties-common
-
-# Install docker
-curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
-sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu focal stable"
-sudo apt install docker-ce
-sudo usermod -aG docker ${USER}
-
-# Start docker.
-sudo systemctl start docker
-
-## You might need to re-connect (due to usermod change).
-
-# Node & yarn
-nvm install 20
-# Important: there will be a note in the output to load
-# new paths in your local session, either run it or reload the terminal.
-npm install -g yarn
-yarn set version 1.22.19
-
 # For running unit tests
-cargo install cargo-nextest
+cargo install cargo-nextest --locked
 # SQL tools
 cargo install sqlx-cli --version 0.8.1
 
-# Foundry ZKsync
+# Install Docker
+curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu focal stable"
+sudo apt install docker-ce
+
+# Optional: If you wish to run Docker commands without sudo, you can run this command.
+# Otherwise, just ignore this line and run all Docker commands with sudo.
+# After this command, you might need to re-connect (due to usermod change).
+sudo usermod -aG docker ${USER}
+
+# Start Docker
+sudo systemctl start docker
+
+# NVM
+curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.5/install.sh | bash
+nvm install 20
+
+# Yarn
+npm install -g yarn | bash
+yarn set version 1.22.19
+
+# Foundry ZK sync
 curl -L https://raw.githubusercontent.com/matter-labs/foundry-zksync/main/install-foundry-zksync | bash
 foundryup-zksync
 
 # Non CUDA (GPU) setup, can be skipped if the machine has a CUDA installed for provers
 # Don't do that if you intend to run provers on your machine. Check the prover docs for a setup instead.
 echo "export ZKSYNC_USE_CUDA_STUBS=true" >> ~/.bashrc
-# You will need to reload your `*rc` file here
+source ~/.bashrc
 
 # Clone the repo to the desired location
-git clone git@github.com:matter-labs/zksync-era.git
+git clone https://github.com/matter-labs/zksync-era.git
 cd zksync-era
 git submodule update --init --recursive
 ```


brunoffranca:
On the Foundry-zksync install [script](https://raw.githubusercontent.com/matter-labs/foundry-zksync/main/install-foundry-zksync), we probably need to add `. "$SHELL_CONFIG_FILE"`  after `echo "Installation completed successfully!"` since the verification afterwards fails (running exec bash solved it on my machine).

brunoffranca:
Weird. When running `zkstack dev fmt` or `zkstack dev lint` after this installation, it complains that Prettier is not installed. But  I can check that it is.

docs/src/guides/setup-dev.md
```diff
@@ -1,68 +1,21 @@
...
+# Installing development setup for ZKsync
 
 ## TL;DR
 
-This is a shorter version of setup guide to make it easier subsequent initializations. If it's the first time you're
+This is a shorter version of setup guide to make it easier for initializations. If it's the first time you're
 initializing the workspace, it's recommended that you read the whole guide below, as it provides more context and tips.
 
-If you run on 'clean' Ubuntu on GCP:
+Just run the following command on Linux:
```
popzxc, line 8: The script doesn't support e.g. arch/fedora.

```suggestion
Just run the following command on Debian-based Linux:
```

docs/src/guides/setup-dev.md
```diff
@@ -77,84 +30,27 @@ Rosetta may cause problems that are hard to spot and debug, so make sure to chec
...
-
-## OpenSSL
-
-Install OpenSSL:
-
 On mac:
 
 ```bash
-brew install openssl
+brew install git build-essential pkg-config cmake clang lldb lld libssl-dev libpq-dev apt-transport-https ca-certificates curl software-properties-common
```
popzxc, line 46: Are these needed? Have you tested it?

docs/src/guides/setup-dev.sh
```diff
@@ -0,0 +1,71 @@
...
+
+   # Install Docker
+   say "Installing Docker..."
+   curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+   sudo add-apt-repository --yes "deb [arch=amd64] https://download.docker.com/linux/ubuntu focal stable"
+   sudo apt install --yes docker-ce
+
+   # Start Docker
+   say "Starting Docker..."
+   sudo systemctl start docker
```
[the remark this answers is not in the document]
brunoffranca: AFAIU the usermod change is just to be able to run docker commands without sudo. In fact, in my test machine, docker is running after this command.

docs/src/guides/setup-dev.sh
```diff
@@ -0,0 +1,71 @@
...
+   # Don't do that if you intend to run provers on your machine. Check the prover docs for a setup instead.
+   say "Setting up the non-CUDA setup..."
+   echo "export ZKSYNC_USE_CUDA_STUBS=true" >> "$HOME/.bashrc"
+
+   # Clone the repo
+   say "Cloning the ZKsync repository..."
+   git clone --recurse-submodules https://github.com/matter-labs/zksync-era.git
+
+   say "Installation of the development environment for ZKsync complete!"
+   say "Please reload your shell configuration by running 'source ~/.bashrc'"
```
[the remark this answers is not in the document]
brunoffranca: See above, there's no need for usermod change I believe.

docs/src/guides/setup-dev.md
brunoffranca: These are the same we had before, I only added curl.
I don't know if they are all necessary, and it would be hard to check that. For example, libssl-dev is only used to compile RocksDB, so I would only figure that out by compiling the node. Unless I run all commands on ZK Stack CLI, I don't know how I would check if the dependencies are all used.

brunoffranca:
@Deniallugo Am I supposed to install prettier for zksync-era after installing ZK Stack CLI?

Deniallugo:
@brunoffranca no you shouldn't, it just must exist in your system

docs/src/guides/setup-dev.sh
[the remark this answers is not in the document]
popzxc: Probably user on your test machine has root access, and ability to run docker commands without sudo is super important, so it must be a part of initialization.

docs/src/guides/setup-dev.md
popzxc: It's a command for `brew` (MacOS). Current `setup-dev` only has two commands for homebrew: `openssl` and `libpq` (the latter is actually not required.

docs/src/guides/setup-dev.md
```diff
@@ -1,68 +1,21 @@
...
+# Installing development setup for ZKsync
 
 ## TL;DR
 
-This is a shorter version of setup guide to make it easier subsequent initializations. If it's the first time you're
+This is a shorter version of setup guide to make it easier for initializations. If it's the first time you're
 initializing the workspace, it's recommended that you read the whole guide below, as it provides more context and tips.
 
-If you run on 'clean' Ubuntu on GCP:
+Just run the following command on Debian-based Linux:
```
popzxc, line 8: I am also hesitatnt to suggest blindly running it. If the script will be run on a machine where, e.g., docker is already installed (_especially_ if it was installed in some other way), it can mess up the configuration.

Do we really need to have a script here? If so, I'd probably want to make it fail-proof, e.g. that it won't do stuff that is not needed, but it's a much bigger effort.

The benefit of a code block is that you see the code and can choose what lines you need and what lines you don't need.

docs/src/guides/setup-dev.md
```diff
@@ -1,68 +1,21 @@
...
 
 ```bash
-# For VMs only! They don't have SSH keys, so we override SSH with HTTPS
-git config --global url."https://github.com/".insteadOf git@github.com:
-git config --global url."https://".insteadOf git://
```
popzxc, line 13: These lines are important for VMs, because submodule fetching may break if running on a fresh machine without valid ssh keys; however they're missing in the script.

docs/src/guides/setup-dev.md
popzxc: If we are to keep this, I'd add some big red warning that you should first read the script code and make sure that it does what you need (which though would render it much less helpful).

docs/src/guides/setup-dev.sh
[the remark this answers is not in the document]
brunoffranca: Why is it important?

docs/src/guides/setup-dev.md
brunoffranca: Fair enough, but the original setup guide specifically says that the code is to be run on a fresh GCP instance. So it seems reasonable that we have a fast setup option for those cases at least.
I can add a code block below with what's in the script, so that people can inspect it though.

docs/src/guides/setup-dev.md
brunoffranca: The standard now really is to declare submodules with https, like we [do](https://github.com/matter-labs/zksync-era/blob/main/.gitmodules). I did follow the tree of submodules down and it seems all of them use https.
Do you think it's worth it to keep this, since it's such a small edge case?

docs/src/guides/setup-dev.md
popzxc: Fairly so.

docs/src/guides/setup-dev.sh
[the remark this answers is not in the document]
popzxc: Because it's a major security prerequisite. Giving docker root access is a bad idea (and in general, you should never give sudo access to anything unless there is a strong request for it). And, well, it's just not convenient to have always run commands with sudo. Plus it will break e.g. `zkstack` CLI because it will (correctly) run docker commands without sudo.

docs/src/guides/setup-dev.md
popzxc: Original -- yes, the new one removed mention of GCP.
Regarding code block -- it would mean duplication, and script will likely get out of sync with the code block.

brunoffranca: Then the only possible solution is for me to close this PR.

docs/src/guides/setup-dev.sh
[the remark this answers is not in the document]
brunoffranca: Regarding security, I'd argue exactly the reverse. With sudo you have to input a password for any docker command, thus requiring human approval. If you add the user to the docker group you effectively allow any app with non-root privileges access to the docker commands.
https://www.reddit.com/r/docker/comments/syngw7/to_sudo_or_not_to_sudo_that_is_the_question/

docs/src/guides/setup-dev.md
popzxc: > I don't really think it's possible to create a fail-proof version of this script (at least without a disproportionate amount of work).

Yup, I agree, and that's exactly the reason why it wasn't done before.

> And if we don't want duplication then the only possible solution is for me to close this PR.

Was there any reason why we need a script except for making usage simpler?

brunoffranca: No, just to make usage simpler. I created one for myself, since I have to run fresh dev nodes with some frequency, and thought it would be useful for more people.

PR #3297, opened 2024-11-16
feat(config): update config with correct root and commitment and compiled with correct compiler

koloz193:
## Checklist


sanekmelnikov:
Thanks for cherrypicking the changes!
As discussed, let's not introduce hardhat instead of forge back again in `build_system_contracts`.
Localhost deployments will have different hashes in `genesis.yaml` than what's actually on chain until the next protocol upgrade
@Deniallugo

sanekmelnikov (changes requested):
Thanks for cherrypicking the changes!
As discussed, let's not introduce hardhat instead of forge back again in build_system_contracts.
Localhost deployments will have different hashes in genesis.yaml than what's actually on chain until the next protocol upgrade
@Deniallugo

deniallugo-ml:

PR #3299, opened 2024-11-18
fix(alerts): Alert if there was no successful eth-watcher iterations for 5 minutes

RomanBrodetski:
We have [this](https://github.com/matter-labs/gitops-kubernetes/blob/main/apps/environments/raas-abstract-testnet/server-v2/server-v2-core.yaml#L185) alert:

```
      no_eth_watch_iterations:
        promql: max(rate(server_eth_watch_eth_poll[30s]))
        compare: "=="
        thresholds:
          critical: 0
        duration: 5m
        annotations:
          description: "No ETH watch iterations for 5 minutes"
          summary: "No ETH watch iterations for 5 minutes"
          runbook_url: "https://www.notion.so/matterlabs/On-Call-1d597d29c9b64c08919bf53c952cd03e?pvs=4#78da3253ab3e41aa8175e72caf456c86"
```

But we reported this `server_eth_watch_eth_poll` metric on every iteration - even if it wasn't successful. This PR changes the logic so that it's only reported when successful - meaning that we will be alerted if there was no successful iterations for five minutes. 

We may want to increase the threshold to something like 10 minutes to avoid being notified at transient L1 provider issues.


Deniallugo (commented):
I think it should be two independent metrics. 
In this particular case, we actually polling ethereum, nothing has stuck and service is working. 

But you want to check that we poll it successfully, and i'd add a new metric for it.

RomanBrodetski:
> In this particular case, we actually polling ethereum, nothing has stuck and service is working.

I think there is much more value in knowing whether there were **successful** operations, not **some** operations. What's the usecase for the metric that currently exist? IMO we should replace it.

Deniallugo (approved):

PR #3315, opened 2024-11-21
feat: add da client can be initialized with the gcs seed

JayT106:
## What ❔

Implement DA client can init with a encrypted seed from google cloud storage. The encrypted seed has been encrypted by a HSM key stored in google KMS.

There are 3 environment variables need to be setup for enabling this feature.
DA_SECRETS_FROM_GCS="true"
DA_SECRETS_GCS_BUCKET_NAME=[gcs-bucket-name]
DA_SECRETS_KMS_DECRYPT_KEY_NAME=[google-kms-key-path]

## Why ❔

Can use the seed encrypted by KMS key and storing in gcs and able to use it in DA client without exposing the key details in the deploy node.  

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1493,7 +1493,7 @@ dependencies = [
  "celestia-tendermint-proto",
  "prost 0.12.6",
  "prost-build",
- "prost-types",
+ "prost-types 0.12.6",
  "protox 0.6.1",
  "serde",
 ]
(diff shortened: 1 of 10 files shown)


dimazhornyk:
IIRC, we discussed this before and agreed that this approach doesn't fit our current secret management flow. To do something like this, we would have to revamp the way we store and use secrets fully and use KMS all over the codebase (using it only for DA secrets doesn't make much sense).

I'll close it for now to not add an unused code upstream but will keep in mind that there is such implementation in case it is required in the future.

PR #3320, opened 2024-11-22
feat: change seal criteria for gateway

perekopskiy:
## What ❔

- replaces sk l1 gas criterion with l1 l2 txs criterion
- removes commit aggregation in eth sender
- enables execute aggregation only for chains that settle on L1
- adds sk criterion for l2 l1 logs

## Why ❔

- make criteria work with gateway

## Checklist


core/node/eth_sender/src/aggregator.rs
```diff
@@ -44,65 +44,63 @@ impl Aggregator {
...
                 }),
-            ],
+                Box::from(ExecuteGasCriterion::new(config.max_aggregated_tx_gas)),
+            ]
+        };
+
+        if config.max_aggregated_blocks_to_commit > 1 {
+            tracing::warn!(
+                "config.max_aggregated_blocks_to_commit is set to {} but \
+                aggregator does not support aggregating commit operations anymore",
```
StanislavBreadless, line 89: Discussed in private, it may make sense to support these for L1 especially in the short term

StanislavBreadless (dismissed):

core/node/state_keeper/src/seal_criteria/criteria/l1_l2_txs.rs
```diff
@@ -0,0 +1,96 @@
...
+        config: &StateKeeperConfig,
+        _block_open_timestamp_ms: u128,
+        _tx_count: usize,
+        l1_tx_count: usize,
+        _block_data: &SealData,
+        _tx_data: &SealData,
+        _protocol_version_id: ProtocolVersionId,
+    ) -> SealResolution {
+        let block_l1_gas_bound =
+            (config.max_single_tx_gas as f64 * config.close_block_at_gas_percentage).round() as u32;
```
StanislavBreadless, line 23: it is actually probably better to change this to actually just check for the number of txs. For chains that settle on top of gateway we will have to set `max_single_tx_gas` to something big (billions of gas) and it may lead to having a large number of priority transactions. Not to say that gateway wont be able to handle a larger amount, but I am not sure how this discrepancy will play out (maybe the amount will become too large accidentally)

StanislavBreadless (changes requested):

StanislavBreadless (approved):

StanislavBreadless (dismissed):

PR #3322, opened 2024-11-22
refactor(zkstack): Preload chain and ecosystem config before each chain subcommand

matias-gonz:
## What ❔
Preload chain and ecosystem config before each chain subcommand


zkstack_cli/crates/zkstack/src/commands/chain/mod.rs
```diff
@@ -95,15 +95,19 @@ pub(crate) async fn run(shell: &Shell, cmd: ChainCommands) -> anyhow::Result<()>
 
     match cmd {
         ChainCommands::Create(args) => create::run(args, shell),
-        ChainCommands::Init(args) => init::run(*args, shell).await,
-        ChainCommands::BuildTransactions(args) => build_transactions::run(args, shell).await,
+        ChainCommands::Init(args) => init::run(*args, shell, chain, ecosystem).await,
```
Deniallugo, line 98: Does it work with create function? 
based on what i see it should fail for create or you have to unwrap chain

matias-gonz: It does because of this:
https://github.com/matter-labs/zksync-era/blob/matias/pre-load-ecosystem-and-chain/zkstack_cli/crates/zkstack/src/commands/chain/mod.rs#L89

I could remove create from the match clause to make it more clear that it will execute before

matias-gonz: [30181f3](https://github.com/matter-labs/zksync-era/pull/3322/commits/30181f30cbf67da643de21bb2e205d722475e43a)

Deniallugo (approved):

PR #3342, opened 2024-11-28
fix(tee): fix race condition in batch locking

pbeza:
## What ❔

After [scaling][1] [`zksync-tee-prover`][2] to two instances/replicas on Azure for `azure-stage2`, `azure-testnet2`, and `azure-mainnet2`, we started experiencing [duplicated proving for some batches][3].
logs
While this is not an erroneous situation, it is wasteful from a resource perspective. This was due to a race condition in batch locking. This PR fixes the issue by adding atomic batch locking.

[1]: https://github.com/matter-labs/gitops-kubernetes/pull/7033/files
[2]: https://github.com/matter-labs/zksync-era/blob/aaca32b6ab411d5cdc1234c20af8b5c1092195d7/core/bin/zksync_tee_prover/src/main.rs
[3]: https://grafana.matterlabs.dev/goto/M1I_Bq7HR?orgId=1

## Why ❔

To fix the bug that only activates after running `zksync-tee-prover` on multiple instances.

## Checklist


core/lib/dal/src/models/storage_tee_proof.rs
```diff
@@ -13,6 +13,11 @@ pub struct StorageTeeProof {
     pub attestation: Option<Vec<u8>>,
 }
 
+#[derive(Debug, Clone, sqlx::FromRow)]
+pub struct StorageBatch {
```
slowli, line 17: Bikeshedding: IMO, creating a new struct for a single field is redundant; you can use an anonymous struct produced by `sqlx::query!` macro.

core/lib/dal/src/tee_proof_generation_dal.rs
```diff
@@ -64,72 +64,93 @@ impl TeeProofGenerationDal<'_, '_> {
...
+        .with_arg("tee_type", &tee_type)
+        .with_arg("processing_timeout", &processing_timeout)
+        .with_arg("min_batch_number", &min_batch_number)
+        .fetch_optional(&mut transaction)
+        .await?;
+
+        let batch_number = match batch_number {
+            Some(batch) => batch.l1_batch_number,
+            None => {
+                transaction.commit().await?;
```
slowli, line 111: There are seemingly no changes in the transaction at this point. Is the commit here just to not clean up DB resources faster?

slowli (commented):
Dumb question: How is the locking made atomic in this PR? AFAIU, the first `SELECT` statement, if queried concurrently, can still return the same L1 batch number unless some kind of row-level locking is implemented (cf. `SELECT FOR UPDATE SKIP LOCKED` in [this contract verifier query](https://github.com/matter-labs/zksync-era/blob/668ca51f5d52646e64b19b973acec05daa1c6f09/core/lib/dal/src/contract_verification_dal.rs#L139-L175)). I'm not even sure the `UPDATE` query will fail for the transaction committed last in case of a race (maybe it would with serialization isolation level, but I'd argue that erroring is not the best cause of action here; row-level locks seem to work better).

pbeza:
Sorry, I misinterpreted what SQL transactions are capable of in this context. I had to educate myself on the subtleties of SQL locking. I found these two links very useful:
- https://stackoverflow.com/q/6998372/1321680
- https://www.postgresql.org/docs/current/sql-select.html#SQL-FOR-UPDATE-SHARE

core/lib/dal/src/models/storage_tee_proof.rs
pbeza: Makes sense. I [reverted it](https://github.com/matter-labs/zksync-era/pull/3342/commits/464b5d4f1a189304fc442eebe847fee2a5d714d5#diff-0bfab201a463b9e47da3be98236bdc07e5f6695b23261dfd69d0dcfa85ae5a44). ✅

core/lib/dal/src/tee_proof_generation_dal.rs
pbeza: To be honest, I thought every `.start_transaction()` needed to either end with `.commit()` or `.rollback()`. I'm not sure how you determine that it's OK to skip it altogether, but I trust your judgment here. I this line [here](https://github.com/matter-labs/zksync-era/pull/3342/commits/464b5d4f1a189304fc442eebe847fee2a5d714d5#diff-46d00b49fb68d6da13c9ecdc5ab3dab6adf502ec216349721586e96553652444L111). ✅

core/lib/dal/src/tee_proof_generation_dal.rs
```diff
@@ -64,72 +64,93 @@ impl TeeProofGenerationDal<'_, '_> {
...
+                AND (
+                    tee.l1_batch_number IS NULL
+                    OR (
+                        (tee.status = $2 OR tee.status = $3)
+                        AND tee.prover_taken_at < NOW() - $4::INTERVAL
+                    )
+                )
+            LIMIT 1
+            FOR UPDATE OF p
+            SKIP LOCKED
```
pbeza, line 94: I had to explicitly use `OF p` here. If you leave it out or switch it to `OF tee`, you will run into this error:
```
error: error returned from database: FOR UPDATE cannot be applied to the nullable side of an outer join
```

pbeza:
@slowli, I’ve addressed your code review comments. Take a look when you get a chance.

It’s kinda hard to test properly without deploying it to stage and letting it run for a while. Specifically, let me know if locking rows in the `proof_generation_details` table is okay (instead of just locking `tee_proof_generation_details rows`).

core/lib/dal/src/tee_proof_generation_dal.rs
slowli: The error is understandable. Just to clarify, since it's the `tee_proof_generation_details` table that is updated below: Do I understand correctly that this new query can actually have `tee_proof_generation_details` side of the join absent (I guess if the query is invoked for a certain batch for the first time?). If so, I'd suggest to add a comment explaining why rows are locked on a table that's not modified in the transaction.

core/lib/dal/src/tee_proof_generation_dal.rs
slowli: A transaction that is dropped w/o being committed is rolled back implicitly. You can add an explicit call to be more precise (but IMO it should be `rollback` so that it's clear that no modifying actions have been performed).

slowli (approved):

core/lib/dal/src/tee_proof_generation_dal.rs
pbeza: I'm fine with rolling back the transaction implicitly then. 👍🏼

core/lib/dal/src/tee_proof_generation_dal.rs
pbeza: > (...) Do I understand correctly that this new query can actually have tee_proof_generation_details side of the join absent (I guess if the query is invoked for a certain batch for the first time?). (...)

Yup, exactly.

> (...) If so, I'd suggest to add a comment explaining why rows are locked on a table that's not modified in the transaction.

Fixed here:
https://github.com/matter-labs/zksync-era/blob/ecd8c335272f6d80ca017f521c5282c469490168/core/lib/dal/src/tee_proof_generation_dal.rs#L69-L72

Lemme know how you would rephrase it if you feel it’s not accurate enough.

slowli (dismissed):

pbeza:
@slowli, @haraldh suggested locking the entire `tee_proof_generation_details` table to keep things simpler. He also raised a concern that if one TEE prover locks the batch, a second TEE prover instance will just get a `no job` response instead of waiting for new batches to become available.

Let me know if this more fine-grained locking approach still works for you!

haraldh (approved):

slowli (approved):

PR #3363, opened 2024-12-06
fix: return min refund when gas_price=0

dimazhornyk:
## What ❔

Return the minimal possible refund when `effective_gas_price=0` to avoid division by zero

## Why ❔

When we set the gas fees to zero, the execution breaks at this point due to division by zero

## Checklist

### DIFF
--- a/core/lib/multivm/src/versions/vm_latest/tracers/refunds.rs
+++ b/core/lib/multivm/src/versions/vm_latest/tracers/refunds.rs
@@ -116,6 +116,9 @@ impl<S> RefundsTracer<S> {
 
         // For now, bootloader charges only for base fee.
         let effective_gas_price = get_batch_base_fee(&self.l1_batch);
+        if effective_gas_price == 0 {
+            return bootloader_refund;
+        }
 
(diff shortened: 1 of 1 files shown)


core/lib/multivm/src/versions/vm_latest/tracers/refunds.rs
```diff
@@ -116,6 +116,9 @@ impl<S> RefundsTracer<S> {
 
         // For now, bootloader charges only for base fee.
         let effective_gas_price = get_batch_base_fee(&self.l1_batch);
+        if effective_gas_price == 0 {
```
StanislavBreadless, line 119: Talked in private: it is a breaking change, but most likely it is fine. Still, it is better to be able to somehow check that the partners (and Era itself) are not affected

ischasny: Hey @StanislavBreadless 
* What if we safeguard it behind a feature toggle? 
* What is the scenario that we need to check for?

deniallugo-ml:

PR #3372, opened 2024-12-10
fix: Fixed cargo deny

brunoffranca:
## What ❔

Fixes cargo deny CI fail.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -97,15 +97,15 @@ dependencies = [
 
 [[package]]
 name = "allocator-api2"
-version = "0.2.18"
+version = "0.2.21"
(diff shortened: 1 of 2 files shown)


Cargo.lock
popzxc, line 1: Why did the cargo.lock change? Normally we try not to mix `cargo update` with any other PRs unless there is a strong reason for it.

brunoffranca: Cargo deny was failing because of some security advisories. Cargo update fixed that.
Then it was failing because of the missing license, and I had to add it.

popzxc: Typically it's better to update just the affected crate in that case (e.g. `cargo update this-one-particular-faulty-crate`), but OK

PR #3383, opened 2024-12-12
feat(consensus): Added view_timeout to consensus config.

brunoffranca:
## Checklist

### DIFF
--- a/core/lib/config/src/configs/consensus.rs
+++ b/core/lib/config/src/configs/consensus.rs
@@ -126,6 +126,9 @@ pub struct ConsensusConfig {
     /// Maximal allowed size of the payload in bytes.
     pub max_payload_size: usize,
 
+    /// View timeout duration in milliseconds.
+    pub view_timeout: usize,
+
     /// Maximal allowed size of the sync-batch payloads in bytes.
     ///
     /// The batch consists of block payloads and a Merkle proof of inclusion on L1 (~1kB),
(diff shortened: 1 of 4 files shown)


core/lib/config/src/configs/consensus.rs
```diff
@@ -126,6 +126,9 @@ pub struct ConsensusConfig {
     /// Maximal allowed size of the payload in bytes.
     pub max_payload_size: usize,
 
+    /// View timeout duration in milliseconds.
+    pub view_timeout: usize,
```
pompon0, line 130: time::Duration

pompon0: also make it an Option<std::time::Duration>, and implement a getter with a default value

core/lib/protobuf_config/src/proto/core/consensus.proto
```diff
@@ -87,6 +87,9 @@ message Config {
   // Maximal allowed size of the sync batches.
   optional uint64 max_batch_size = 10; // required; bytes
 
+  // View timeout for the consensus protocol.
+  optional uint64 view_timeout = 13; // required; milliseconds
```
pompon0, line 91: std.Duration

core/lib/config/src/configs/consensus.rs
pompon0: also make it an Option\<std::time::Duration\>, and implement a getter with a default value

core/lib/protobuf_config/src/proto/core/consensus.proto
brunoffranca: I did try it but then got into an issue where Rust couldn't convert between proto::std::Duration and time::Duration and ended up going back. I'll give it a try again.

pompon0 (approved):

PR #3391, opened 2024-12-16
zk os integration: use nonces from storage, not from transactions table

RomanBrodetski:


deniallugo-ml:
Closing this PR as the experimental work it belongs to has either moved to a dedicated repository or been deprioritized in favour of a different approach. Keeping it open here creates confusion about the current state of this feature.

If this work is continuing, it will be tracked in the appropriate repo or branch.

PR #3398, opened 2024-12-18
fix: block.timestamp is not accurate

thomas-nguy:
## What ❔

Related to https://github.com/zkSync-Community-Hub/zksync-developers/discussions/820

Change the l2 block creation logic to start a new l2 block only when a transaction is ready to be executed.

## Why ❔

Current logic starts a new l2 block as soon as the previous one is sealed.

A contract that relies on `block.timestamp` would be able to predict the time correctly because if the l2 block goes stale (no transaction), then it will be open indefinitely and the timestamp will not be accurate anymore

Solution has been tested locally but any feedbacks would be appreciated

## Checklist


core/node/state_keeper/src/keeper.rs
```diff
@@ -89,6 +90,7 @@ impl ZkSyncStateKeeper {
             sealer,
             storage_factory,
             health_updater: ReactiveHealthCheck::new("state_keeper").1,
+            should_create_l2_block: false,
```
thomas-nguy, line 93: not sure if we should persist this in rocksdb? it looks fine during testing

popzxc:
@slowli PTAL

core/node/state_keeper/src/keeper.rs
```diff
@@ -585,14 +591,30 @@ impl ZkSyncStateKeeper {
...
 
-            if self.io.should_seal_l2_block(updates_manager) {
+            if !self.should_create_l2_block && self.io.should_seal_l2_block(updates_manager) {
                 tracing::debug!(
                     "L2 block #{} (L1 batch #{}) should be sealed as per sealing rules",
                     updates_manager.l2_block.number,
                     updates_manager.l1_batch.number
                 );
                 self.seal_l2_block(updates_manager).await?;
+                self.should_create_l2_block = true;
```
slowli, line 601: I don't quite understand the purpose of this variable. AFAIU, the logic here conceptually should change as follows:

- After sealing the block, do not start the next block immediately; instead, set a local var whether to start it.
- Wait for the next transaction.
- After receiving a transaction, if the flag is set, start a new block and unset the flag.

The logic here *almost* follows this flow, but the flag is non-local, which complicates reasoning.

core/node/state_keeper/src/keeper.rs
```diff
@@ -187,7 +189,10 @@ impl ZkSyncStateKeeper {
 
             // Finish current batch.
             if !updates_manager.l2_block.executed_transactions.is_empty() {
-                self.seal_l2_block(&updates_manager).await?;
+                if !self.should_create_l2_block {
+                    // l2 block has been already sealed
+                    self.seal_l2_block(&updates_manager).await?;
+                }
```
slowli, line 195: This place is confusing with the proposed changes. The check above checks whether the latest block contains any transactions. AFAIU, if `should_create_l2_block` is true, then the check doesn't concern the *latest* block, but rather the previous one; the latest block isn't really started yet. So, a fictive block must be started in any case. IIUC, the current approach technically works because the previous block exists and is non-empty, but it looks hacky.

slowli (commented):
I think it could be worth exploring a slightly differing approach: decoupling setting a new block in the `UpdatesManager` and in `BatchExecutor`. Namely, as soon as a new block is added in the current workflow, it is still added in `UpdatesManager`, but is not sent to `BatchExecutor`. Instead, it is only sent to `BatchExecutor` after receiving the first transaction in the block with the updated timestamp (obviously, the timestamp needs to be updated in `UpdatesManager` as well). IMO, this would make it slightly easier to reason about correctness.

core/node/state_keeper/src/keeper.rs
thomas-nguy: You are right, this could totally be the flag local!

The reason I have it global is because I need to know the state of the last fictive block when closing the batch in the parent loop (parent function).

- Case 1 : The last block has not been sealed. This is the original behavior before the PR change because we are always creating new unsealed block right after sealing one wether we receive new transaction or not. This is why we are "always" sealing the last block in the parent loop before closing the batch.

- Case 2: The last block has been sealed, but no transaction has been received for some period of time and ultimately we are "forced" closing the batch. In that case we should not seal the last block again in the parent loop

It seems a bit hacky indeed but it was the best way I found in order to not introduce too much change in this PR. 

Perhaps I can change this to a local flag and create another global flag for the parent loop if it can make it easier to understand?

core/node/state_keeper/src/keeper.rs
thomas-nguy: see comment below.

core/node/state_keeper/src/keeper.rs
thomas-nguy: @slowli see if it is easier to read with this version?

core/node/state_keeper/src/keeper.rs
```diff
@@ -582,34 +593,51 @@ impl ZkSyncStateKeeper {
                     "L1 batch #{} should be sealed unconditionally as per sealing rules",
                     updates_manager.l1_batch.number
                 );
+
+                // Push the current block if it has not been done yet
+                if is_last_block_sealed {
+                    self.io
+                        .update_next_l2_block_timestamp(updates_manager.next_l2_block_timestamp());
```
slowli, line 600: AFAIU, if `is_last_block_sealed`, we have next block params in `updates_manager`, so this will create a fictive block (= one w/o transactions), right? If so, I'd change the comment above and the log below to mention it explicitly. Also, it could be worth asserting at the callsite that after `process_l1_batch()` returns, `updates_manager` doesn't have next block params; this is necessary for correctness.

core/node/state_keeper/src/keeper.rs
```diff
@@ -582,34 +593,51 @@ impl ZkSyncStateKeeper {
...
+                        updates_manager.l1_batch.number,
+                        display_timestamp(updates_manager.next_l2_block_param.timestamp)
+                    );
+                    Self::start_next_l2_block(updates_manager, batch_executor).await?;
+                }
                 return Ok(());
             }
 
-            if self.io.should_seal_l2_block(updates_manager) {
+            if self.io.should_seal_l2_block(updates_manager) && !is_last_block_sealed {
```
slowli, line 612: Does ordering of the clauses matter here (I guess from the perspective of `should_seal_l2_block` side-effects)? Logically, it makes sense to reorder the clauses, so that the cheaper one (`!is_last_block_sealed`) is checked first.

core/node/state_keeper/src/keeper.rs
```diff
@@ -582,34 +593,51 @@ impl ZkSyncStateKeeper {
...
+                // Push the current block if it has not been done yet
+                if is_last_block_sealed {
+                    self.io
+                        .update_next_l2_block_timestamp(updates_manager.next_l2_block_timestamp());
+                    tracing::debug!(
+                        "Initialized new L2 block #{} (L1 batch #{}) with timestamp {}",
+                        updates_manager.l2_block.number + 1,
+                        updates_manager.l1_batch.number,
+                        display_timestamp(updates_manager.next_l2_block_param.timestamp)
+                    );
```
slowli, line 606: This can be moved inside `Self::start_next_l2_block` to avoid code duplication.

core/node/state_keeper/src/updates/mod.rs
```diff
@@ -39,6 +39,7 @@ pub struct UpdatesManager {
     pub l2_block: L2BlockUpdates,
     pub storage_writes_deduplicator: StorageWritesDeduplicator,
     pubdata_params: PubdataParams,
+    pub next_l2_block_param: L2BlockParams,
```
slowli, line 42: - Please make this field an `Option` to align with semantics; we don't always have the next block params, only sometimes. These params would be set by `set_next_l2_block_parameters()` (asserting that they are `None`, i.e. we never redefine next block params from scratch) and `take()`n by `push_l2_block()`.
- AFAIU, there's no need to make this field public. Some state keeper logic reads a timestamp, but this can be done via `next_l2_block_timestamp()`. Making the field private ensures that its intended lifecycle invariants above are upheld.

core/node/state_keeper/src/updates/mod.rs
```diff
@@ -167,19 +176,23 @@ impl UpdatesManager {
...
-            l2_block_params.virtual_blocks,
+            self.next_l2_block_param.virtual_blocks,
             self.protocol_version,
         );
         let old_l2_block_updates = std::mem::replace(&mut self.l2_block, new_l2_block_updates);
         self.l1_batch
             .extend_from_sealed_l2_block(old_l2_block_updates);
     }
 
+    pub fn set_next_l2_block_parameters(&mut self, l2_block_param: L2BlockParams) {
```
slowli, line 192: Bikeshedding: I'd rename this method to `set_next_l2_block_params()` to align with the field / type naming.

core/node/state_keeper/src/keeper.rs
```diff
@@ -564,8 +572,11 @@ impl ZkSyncStateKeeper {
         batch_executor: &mut dyn BatchExecutor<OwnedStorage>,
         updates_manager: &mut UpdatesManager,
         protocol_upgrade_tx: Option<ProtocolUpgradeTx>,
+        state_restored: bool,
         stop_receiver: &watch::Receiver<bool>,
     ) -> Result<(), Error> {
+        let mut is_last_block_sealed = state_restored;
```
slowli, line 578: IIUC, this var is actually redundant because it corresponds to whether `updates_manager` has next block params; see the other line comment about making those `Option`al. So it can be replaced by the `has_next_block_params(&self) -> bool` getter in `UpdatesManager`.

core/node/state_keeper/src/keeper.rs
```diff
@@ -582,34 +592,55 @@ impl ZkSyncStateKeeper {
                     "L1 batch #{} should be sealed unconditionally as per sealing rules",
                     updates_manager.l1_batch.number
                 );
+
+                // Push the current block if it has not been done yet and this will effectively create a fictive l2 block
+                if updates_manager.has_next_block_params() {
+                    self.io.update_next_l2_block_timestamp(
+                        &mut updates_manager
+                            .get_next_l2_block_params_or_batch_params()
+                            .timestamp,
```
slowli, line 601: This doesn't update the timestamp inside `UpdatesManager` because `get_next_l2_block_params_or_batch_params()` returns a copy of params; see its signature. For it to work properly, you should revert to providing a mutable ref to the timestamp from `UpdatesManager`.

core/node/state_keeper/src/keeper.rs
```diff
@@ -416,12 +418,21 @@ impl ZkSyncStateKeeper {
...
+        let timestamp = updates_manager
+            .get_next_l2_block_params_or_batch_params()
+            .timestamp;
+        tracing::debug!(
+            "Initialized new L2 block #{} (L1 batch #{}) with timestamp {}",
+            updates_manager.l2_block.number + 1,
+            updates_manager.l1_batch.number,
+            display_timestamp(timestamp)
+        );
+        updates_manager.set_next_l2_block_params(None);
```
slowli, line 435: Please use `Self:: set_l2_block_params()` here and move the log message there as discussed previously.

core/node/state_keeper/src/updates/mod.rs
```diff
@@ -77,6 +79,20 @@ impl UpdatesManager {
         self.base_system_contract_hashes
     }
 
+    pub(crate) fn get_next_l2_block_params_or_batch_params(&mut self) -> L2BlockParams {
```
slowli, line 82: AFAICT, this method should be replaced by providing a mutable ref to the timestamp from the next block params. It must not have a fallback; the caller is responsible for calling this method only when the next block params exist (as the state keeper would).

core/node/state_keeper/src/keeper.rs
thomas-nguy: I am not sure which log you are talking about but I added one in Self:: set_l2_block_params()

core/node/state_keeper/src/updates/mod.rs
```diff
@@ -167,19 +183,24 @@ impl UpdatesManager {
 
     /// Pushes a new L2 block with the specified timestamp into this manager. The previously
     /// held L2 block is considered sealed and is used to extend the L1 batch data.
-    pub fn push_l2_block(&mut self, l2_block_params: L2BlockParams) {
+    pub fn push_l2_block(&mut self) {
```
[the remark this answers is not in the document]
thomas-nguy: done

core/node/state_keeper/src/updates/mod.rs
slowli: You mean the first transaction in a batch, right? AFAIU, the current implementation won't set next L2 block params on `UpdatesManager` initialization (= when opening a new batch), meaning that the timestamp of the first block in a batch may still be outdated. Is this the case? If so, it can be solved incrementally, but I wouldn't consider a fix complete without covering this scenario. I *think* there's no choice here other than delaying creating `UpdatesManager` / `BatchExecutor` until we have a first transaction to execute (because the first block in the batch is immediately initialized in the VM and it cannot be modified afterwards), and this would require some refactoring of the state keeper logic.

core/node/state_keeper/src/updates/mod.rs
```diff
@@ -77,6 +79,27 @@ impl UpdatesManager {
         self.base_system_contract_hashes
     }
 
+    pub(crate) fn get_next_l2_block_timestamp_as_mut(&mut self) -> &mut u64 {
+        if let Some(l2_block_params) = self.next_l2_block_params.as_mut() {
+            return &mut l2_block_params.timestamp;
+        }
+        &mut self.l2_block.timestamp
```
slowli, line 86: AFAIU, this fallback is logically incorrect, but it's never used. It's incorrect because changing a timestamp of the 1st block in a batch must be done together with changing the timestamp of the batch (they should always be equal), and neither can really change because a VM instance for the batch is already initialized at this point. It seems that there's a neat opportunity to combine this method with `has_next_block_params()`:

```rust
pub(crate) fn next_l2_block_timestamp_mut(&mut self) -> Option<&mut u64> {
    self.next_l2_block_params.as_mut().map(|params| &mut params.timestamp)
}

// usage in state keeper logic like `process_l1_batch()`
if let Some(timestamp) = updates_manager.next_l2_block_timestamp_mut() {
    self.io.update_next_l2_block_timestamp(timestamp);
}
```

slowli (commented):
Other than `get_next_l2_block_timestamp_as_mut()`, looks good, although as I mentioned in a previous reply, I think the fix as-is is incomplete (could be improved in a separate PR).

core/node/state_keeper/src/updates/mod.rs
thomas-nguy: Make totally sense! Fixed

core/node/state_keeper/src/keeper.rs
```diff
@@ -408,6 +405,18 @@ impl ZkSyncStateKeeper {
         Err(Error::Canceled)
     }
 
+    #[tracing::instrument(skip(updates_manager))]
```
slowli, line 408: Nit: Please use `level = "debug"` here. Or you can remove the instrumentation altogether given that info about L2 block params is logged below anyway.

slowli (dismissed):
Good job 👍

- Ideally, it'd make sense to unit-test the changes using a mock `StateKeeperIO` / batch executor implementations to check that the L2 block timestamp is indeed updated if there are no transactions, and that the timestamp passed to a VM instance is the updated one. Could wait for a follow-up PR.
- Suggest for another review from the platform team (e.g., @perekopskiy or @Deniallugo ), but this isn't absolutely necessary.

core/node/state_keeper/src/keeper.rs
thomas-nguy: Thanks, I'll just remove it

core/node/state_keeper/src/keeper.rs
```diff
@@ -582,34 +594,52 @@ impl ZkSyncStateKeeper {
...
+                    Self::start_next_l2_block(updates_manager, batch_executor).await?;
+                }
+
                 return Ok(());
             }
 
-            if self.io.should_seal_l2_block(updates_manager) {
+            if !updates_manager.has_next_block_params()
+                && self.io.should_seal_l2_block(updates_manager)
+            {
```
perekopskiy, line 611: Do I understand it correctly that adding `!updates_manager.has_next_block_params()` is not strictly necessary here? I.e. `has_next_block_params()` can be true only if `updates_manager.l2_block.executed_transactions.is_empty()`. Just in case I'm not against adding it, just want to understand better

thomas-nguy: if you don't have `!updates_manager.has_next_block_params()` then you might end up sealing the same block twice (=error) in case `wait_for_next_tx` timed out and the function loop over. 

(`start_next_l2_block` is called only when a transaction can be found in the mempool0

perekopskiy: I see, makes sense

perekopskiy (approved):

PR #3400, opened 2024-12-18
chore: Reduction of diff between main and sync-layer-stable

StanislavBreadless:
## What ❔

This PR includes:
- Changes to VM. Note, that since they are only activated during usage of the new ProtocolVersion, they should not affect the current behavior.
- Changes to the VM introduce the max new factory deps as equal to 64 (but only to the last subversion)
- The block reverted made to work on top of gateway, but most of the diff is due to the removal of the support for env-based instead of file based inputs
- Gateway config ported
- Some various changes to gateway-specific logic that should not be active before the upgrade

## Checklist


core/node/api_server/src/web3/namespaces/zks.rs
```diff
@@ -237,6 +238,14 @@ impl ZksNamespace {
         msg: H256,
         l2_log_position: Option<usize>,
     ) -> Result<Option<L2ToL1LogProof>, Web3Error> {
+        if let Some(handler) = &self.state.l2_l1_log_proof_handler {
```
slowli, line 241: Dumb question just for my understanding: How is `l2_l1_log_proof_handler` going to be initialized? (AFAIU, `with_l2_l1_log_proof_handler()` method that sets it isn't currently used anywhere.) Will it proxy to the main node on ENs?

core/node/node_framework/src/implementations/layers/eth_sender/aggregator.rs
```diff
@@ -108,22 +153,29 @@ impl WiringLayer for EthTxAggregatorLayer {
             .map(BoundEthInterface::sender_account);
 
         let config = self.eth_sender_config.sender.context("sender")?;
+        let mut connection = replica_pool
+            .connection_tagged("eth_sender")
+            .await
+            .map_err(DalError::generalize)?;
```
slowli, line 159: This mixes business logic into wiring logic; so e.g. errors related to DB will be reported as wiring errors, and delays in obtaining connection / executing SQL queries will block wiring. Is it possible to move a connection *pool* into the aggregator and read its Merkle tree asynchronously (i.e., after wiring is finished)?

core/node/node_framework/src/implementations/layers/eth_watch.rs
```diff
@@ -66,21 +73,54 @@ impl WiringLayer for EthWatchLayer {
         let main_pool = input.master_pool.get().await?;
         let client = input.eth_client.0;
 
-        let eth_client = EthHttpQueryClient::new(
+        let sl_diamond_proxy_addr = if self.settlement_mode.is_gateway() {
+            self.gateway_contracts_config
+                .clone()
+                .unwrap()
```
slowli, line 79: Is `unwrap()` safe here by construction? Prefer propagating errors if there's even a remote chance that it can be triggered.

core/node/node_framework/src/implementations/layers/pk_signing_eth_client.rs
```diff
@@ -86,15 +92,41 @@ impl WiringLayer for PKSigningEthClientLayer {
...
-                self.sl_chain_id,
+                l1_chain_id,
                 query_client,
             );
             BoundEthInterfaceForBlobsResource(Box::new(signing_client_for_blobs))
         });
+        let signing_client_for_gateway = if input.gateway_client.is_some() {
+            if self
+                .gateway_contracts_config
+                .clone()
```
slowli, line 103: Nit: Cloning isn't necessary here (and in a couple of other similar places); it's possible to use `as_ref()` instead. Also, given that `self.gateway_contracts_config.unwrap()` is used below, it would be more idiomatic to have an `if let`:

```rust
if let Some(gateway_contracts) = &self.gateway_contracts_config {
    if gateway_contracts. gateway_chain_id != SLChainId(0) {
        // ...
    }
}
```

core/node/node_sync/src/validate_chain_ids_task.rs
```diff
@@ -15,49 +15,58 @@ use zksync_web3_decl::{
...
-        eth_client: Box<DynClient<L1>>,
-        expected: SLChainId,
+    async fn check_client(
+        l1_client: Option<Box<DynClient<L1>>>,
+        expected: Option<SLChainId>,
     ) -> anyhow::Result<()> {
+        let (l1_client, expected) = match (l1_client, expected) {
+            (Some(l1_client), Some(expected)) => (l1_client, expected),
+            _ => return Ok(()),
+        };
```
slowli, line 54: Nit: Can rewrite this using let-else:

```rust
let (Some(l1_client), Some(expected)) = (l1_client, expected) else {
    return Ok(());
};
```

core/bin/block_reverter/src/main.rs
```diff
@@ -126,111 +132,112 @@ async fn main() -> anyhow::Result<()> {
...
+        .settlement_mode;
+
+    let (sl_rpc_url, sl_diamond_proxy, sl_validator_timelock) = if settlement_mode.is_gateway() {
+        let gateway_chain_config: GatewayChainConfig =
+            read_yaml_repr::<zksync_protobuf_config::proto::gateway::GatewayChainConfig>(
+                &opts
+                    .gateway_chain_path
+                    .context("Genesis config path not provided")?,
+            )
+            .context("failed decoding genesis YAML config")?;
```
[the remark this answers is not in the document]
StanislavBreadless: By how much would it help you? I can ofc redo to make the thing a part of the `ContractsConfig`, but it will be migration of a "gateway config file" into a "gateway config struct" that is a part of `ContractsConfig`

core/bin/external_node/src/config/mod.rs
```diff
@@ -983,11 +988,11 @@ impl OptionalENConfig {
...
 pub(crate) struct RequiredENConfig {
-    /// The chain ID of the L1 network (e.g., 1 for Ethereum mainnet). In the future, it may be different from the settlement layer.
+    /// The chain ID of the L1 network (e.g., 1 for Ethereum mainnet).
     pub l1_chain_id: L1ChainId,
-    /// The chain ID of the settlement layer (e.g., 1 for Ethereum mainnet). This ID will be checked against the `eth_client_url` RPC provider on initialization
-    /// to ensure that there's no mismatch between the expected and actual settlement layer network.
-    pub sl_chain_id: Option<SLChainId>,
+    /// The chain ID of the gateway. This ID will be checked against the `gateway_rpc_url` RPC provider on initialization
+    /// to ensure that there's no mismatch between the expected and actual gateway network.
+    pub gateway_chain_id: Option<SLChainId>,
```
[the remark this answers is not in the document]
StanislavBreadless: It should be okay (there was no reason to supply the var in the current setting), but better to double check

core/lib/config/src/configs/gateway.rs
```diff
@@ -0,0 +1,73 @@
+use zksync_basic_types::{web3::Bytes, Address, SLChainId};
+
+use super::ContractsConfig;
+
+/// Config that is only stored for the gateway chain.
+#[derive(serde::Serialize, serde::Deserialize, Debug, Clone, PartialEq)]
+pub struct GatewayConfig {
```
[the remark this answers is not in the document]
StanislavBreadless: Just to reduce the diff with the main PR where it will be required

core/lib/config/src/configs/contracts.rs
```diff
@@ -46,6 +48,10 @@ pub struct ContractsConfig {
     pub ecosystem_contracts: Option<EcosystemContracts>,
     // Used by the RPC API and by the node builder in wiring the BaseTokenRatioProvider layer.
     pub base_token_addr: Option<Address>,
+    pub base_token_asset_id: Option<H256>,
+
+    pub predeployed_l2_wrapped_base_token_address: Option<Address>,
```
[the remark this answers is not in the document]
StanislavBreadless: They will be needed during migration. I added those here to reduce the diff

core/lib/config/src/configs/da_client/avail.rs
```diff
@@ -1,31 +1,31 @@
-use serde::Deserialize;
+use serde::{Deserialize, Serialize};
 use zksync_basic_types::secrets::{APIKey, SeedPhrase};
 
 pub const AVAIL_GAS_RELAY_CLIENT_NAME: &str = "GasRelay";
 pub const AVAIL_FULL_CLIENT_NAME: &str = "FullClient";
 
-#[derive(Clone, Debug, PartialEq, Deserialize)]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
```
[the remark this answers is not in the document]
StanislavBreadless: To reduce the diff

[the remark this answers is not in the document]
StanislavBreadless: To reduce the diff

core/node/api_server/src/web3/metrics.rs
```diff
@@ -394,7 +394,9 @@ impl ApiMetrics {
             Web3Error::ProxyError(err) => {
                 tracing::warn!("Error proxying call to main node in method `{method}`: {err}");
             }
-            _ => { /* do nothing */ }
+            _ => {
+                tracing::debug!("Error in method `{method}`: {err:#}");
```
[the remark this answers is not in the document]
StanislavBreadless: okay, fine, I can even revert this change completely if it is better this way. @perekopskiy added it most likely during 
 local debugging and it might be helpful there

core/node/api_server/src/web3/namespaces/zks.rs
StanislavBreadless: yes, it will proxy to the main node on ENs (that's the only usecase)

core/node/node_framework/src/implementations/layers/eth_sender/aggregator.rs
StanislavBreadless: I can do make the merkle tree `OnceCell` and then set the value somewhere down the line if this is better

core/node/node_framework/src/implementations/layers/eth_watch.rs
StanislavBreadless: I think it can be triggered, it is more like invalid configuration state, I will propagate

core/lib/types/Cargo.toml
```diff
@@ -19,6 +19,7 @@ zksync_mini_merkle_tree.workspace = true
 zksync_protobuf.workspace = true
 zksync_crypto_primitives.workspace = true
 
+async-trait.workspace = true
```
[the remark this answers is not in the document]
StanislavBreadless: I did it to reduce the diff, but considering that it is an external crate, okay I wont include it in this PR

core/lib/multivm/src/versions/testonly/refunds.rs
```diff
@@ -140,7 +140,7 @@ pub(crate) fn test_predetermined_refunded_gas<VM: TestedVm>() {
         current_state_without_predefined_refunds.user_l2_to_l1_logs
     );
 
-    assert_ne!(
+    assert_eq!(
```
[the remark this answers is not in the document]
StanislavBreadless: Not sure... The was fixed by @perekopskiy and using `asset_eq` is more logically correct. I will just revert this change from this PR

core/bin/block_reverter/src/main.rs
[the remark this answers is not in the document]
StanislavBreadless: Discussed in private. It was decided that merge + fix later is fine

core/bin/external_node/src/config/mod.rs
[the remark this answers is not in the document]
StanislavBreadless: Double checked, no issue

core/lib/env_config/src/lib.rs
```diff
@@ -39,6 +39,10 @@ pub trait FromEnv: Sized {
     fn from_env() -> anyhow::Result<Self>;
 }
 
+pub trait FromEnvVariant: Sized {
```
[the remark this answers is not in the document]
StanislavBreadless: I'll just remove it. For Gateway I will always use file-based config. For other things it is not needed

core/node/eth_sender/src/aggregator.rs
```diff
@@ -133,7 +141,9 @@ impl Aggregator {
             operate_4844_mode,
             pubdata_da,
             commitment_mode,
-        }
+            priority_merkle_tree: OnceCell::new(),
```
slowli, line 144: Nit: Since initialization in `get_or_init_tree` has exclusive access to `self`, won't `Option<_>` suffice here instead of `OnceCell`?

core/node/eth_sender/src/aggregator.rs
```diff
@@ -178,6 +188,38 @@ impl Aggregator {
...
+            let priority_merkle_tree = MiniMerkleTree::<L1Tx>::from_hashes(
+                KeccakHasher,
+                priority_op_hashes.into_iter(),
+                None,
+            );
+
+            // The error can only be thrown if this tree has been populated before.
+            // It is expected that this tree is populated deterministically based on config and
+            // so if some other thread populated it, the stored value should've been the same.
+            let _ = self.priority_merkle_tree.set(priority_merkle_tree);
```
slowli, line 216: IIUC, other threads cannot have populated the tree because of `&mut self` receiver; if multiple threads could acquire such a receiver in parallel, that'd violate fundamental Rust invariants. (For the same reason, it looks like `Option<_>` would be sufficient for the tree, as I've noted in a comment above.)

core/node/eth_sender/src/aggregator.rs
```diff
@@ -199,11 +241,68 @@ impl Aggregator {
...
             last_sealed_l1_batch,
         )
-        .await;
+        .await?;
+
+        let start_index = self.config.priority_tree_start_index;
+        let priority_merkle_tree = self.get_or_init_tree().await;
+
+        let mut priority_ops_proofs = vec![];
+        for batch in l1_batches.iter() {
```
slowli, line 250: Nit: `for batch in &l1_batches` is more idiomatic.

core/node/eth_sender/src/aggregator.rs
```diff
@@ -199,11 +241,68 @@ impl Aggregator {
...
+            let priority_tree_start_index = if let Some(start_index) = start_index {
+                start_index
+            } else {
+                // Start index not present, we push empty proof.
+                priority_ops_proofs.push(Default::default());
+                continue;
```
slowli, line 256: Not saying this is necessarily incorrect, but looks sus (if I read the code correctly): Since `start_index` isn't changed in the loop, why can't it be refactored out of it, so that the loop isn't executed at all if `start_index == None`, and instead `priority_ops_proofs` is filled with default data? If this is intended, IMO, such refactoring would make the intent more explicit.

core/node/eth_sender/src/aggregator.rs
```diff
@@ -199,11 +241,68 @@ impl Aggregator {
...
+                    .get_l1_transactions_hashes(
+                        priority_tree_start_index + priority_merkle_tree.length(),
+                    )
+                    .await
+                    .unwrap();
+                for hash in new_l1_tx_hashes {
+                    priority_merkle_tree.push_hash(hash);
+                }
+
+                priority_merkle_tree.trim_start(
```
slowli, line 285: Nit: Could make sense to add a comment explaining trimming here. AFAIU, this trims the tree to retain just hashes from the newly processed L1 batch.

core/node/eth_sender/src/aggregator.rs
```diff
@@ -199,11 +241,68 @@ impl Aggregator {
...
+                .blocks_dal()
+                .get_batch_first_priority_op_id(batch.header.number)
+                .await
+                .unwrap()
+            {
+                // Batch has no priority ops, no proofs to send
+                None => None,
+                // We haven't started to use the priority tree in the contracts yet
+                Some(id) if id < priority_tree_start_index => None,
+                Some(id) => Some(id),
```
slowli, line 269: Nit: It's possible to use `Option::filter()` here to be more concise.

core/node/node_framework/src/implementations/layers/eth_sender/aggregator.rs
```diff
@@ -71,13 +76,15 @@ impl EthTxAggregatorLayer {
     pub fn new(
         eth_sender_config: EthConfig,
         contracts_config: ContractsConfig,
+        gateway_contracts_config: Option<GatewayChainConfig>,
```
slowli, line 79: Nit: Why not `gateway_chain_config`? 🙃

core/node/node_framework/src/implementations/layers/eth_sender/aggregator.rs
```diff
@@ -95,7 +102,44 @@ impl WiringLayer for EthTxAggregatorLayer {
...
+        );
+        tracing::info!("Contracts: {:?}", self.contracts_config);
+        tracing::info!("Gateway contracts: {:?}", self.gateway_contracts_config);
         // Get resources.
+
+        let (validator_timelock_addr, multicall3_addr, diamond_proxy_addr) =
+            if self.settlement_mode.is_gateway() {
+                (
+                    self.gateway_contracts_config
+                        .clone()
```
slowli, line 118: `clone()`s here and below are redundant as we've discussed. I'd advise to define a local var as `self.gateway_contracts_config.as_ref().unwrap()` and maybe add error handling.

core/node/node_framework/src/implementations/layers/eth_sender/aggregator.rs
```diff
@@ -95,7 +102,44 @@ impl WiringLayer for EthTxAggregatorLayer {
...
+            } else {
+                (
+                    self.contracts_config.validator_timelock_addr,
+                    self.contracts_config.l1_multicall3_addr,
+                    self.contracts_config.diamond_proxy_addr,
+                )
+            };
+
+        let eth_client = if self.settlement_mode.is_gateway() {
+            input.eth_client_gateway.unwrap().0
```
slowli, line 139: Should also have error handling; it's not that hard to implement, just use `anyhow::Context`.

core/node/node_framework/src/implementations/layers/pk_signing_eth_client.rs
```diff
@@ -86,15 +92,40 @@ impl WiringLayer for PKSigningEthClientLayer {
...
             );
             BoundEthInterfaceForBlobsResource(Box::new(signing_client_for_blobs))
         });
 
+        let signing_client_for_gateway = if let (Some(client), Some(gateway_contracts)) = (
+            &input.gateway_client,
+            self.gateway_contracts_config.as_ref(),
+        ) {
+            if gateway_contracts.gateway_chain_id.0 != 0u64 {
+                let private_key = self.wallets.operator.private_key();
```
slowli, line 106: Dumb question: Is it intentional that the `operator` wallet is used here, not the new `gateway` wallet?

slowli (commented):
Mostly looks good.

core/node/eth_sender/src/aggregator.rs
StanislavBreadless: It would, I thought using OnceCell would be more representative of the meaning, but okay, I will update

slowli: `OnceCell` has interior mutability / synchronization because in the intended use case, it can be attempted to be initialized concurrently from multiple threads. `Option` is thus cheaper to use (although this is obviously not a large concern in this case), and has slightly simpler API.

core/node/node_framework/src/implementations/layers/pk_signing_eth_client.rs
StanislavBreadless: I looked through the code, the `gateway` wallet is not used at all atm. Since we do not need the operator addresses to differ and it is already how the server is implemented, I'll delete it from the wallet

core/lib/config/src/configs/gateway.rs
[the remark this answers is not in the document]
StanislavBreadless: The difference is that `GatewayChainConfig` is for chains that are based on top of gateway, but `GatewayConfig` is for gateway itself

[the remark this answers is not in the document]
StanislavBreadless: But okay let's address later

core/bin/block_reverter/src/main.rs
```diff
@@ -46,6 +48,10 @@ struct Cli {
     /// Path to yaml genesis config. If set, it will be used instead of env vars
     #[arg(long, global = true)]
     genesis_path: Option<PathBuf>,
+    /// Path to yaml config of the chain on top of gateway.
+    /// It may be `None` in case a chain is not settling on top of Gateway.
```
slowli, line 52: Nit: For external users (not sure whether the block reverter has them), speaking in terms of Rust types is unnecessarily low-level. Something like "Should be skipped in case..." is more understandable.

core/bin/zksync_server/src/main.rs
```diff
@@ -56,6 +56,9 @@ struct Cli {
     /// Path to the yaml with contracts. If set, it will be used instead of env vars.
     #[arg(long)]
     contracts_config_path: Option<std::path::PathBuf>,
+    /// Path to the yaml with contracts. If set, it will be used instead of env vars.
```
slowli, line 59: Nit: AFAIU, this description is incorrect; gateway-related params are never parsed from env vars. Also, could be worth specifying "with *gateway* contracts".

core/lib/basic_types/src/lib.rs
```diff
@@ -122,7 +122,7 @@ impl TryFrom<U256> for AccountTreeId {
 
 /// ChainId in the ZKsync network.
 #[derive(Copy, Clone, Debug, Serialize, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct L2ChainId(u64);
+pub struct L2ChainId(pub u64);
```
slowli, line 125: What's the cause of this change? The wrapped value was intentionally made private because not all `u64` represent a valid L2 chain ID; see `L2ChainId::new`.

core/node/eth_sender/src/tester.rs
```diff
@@ -406,15 +411,19 @@ impl EthSenderTester {
...
-                    .await,
-            ]
-            .into_iter()
-            .map(l1_batch_with_metadata)
-            .collect(),
-            priority_ops_proofs: Vec::new(),
+            priority_ops_proofs: l1_batch_headers
+                .iter()
+                .map(|_| Default::default())
+                .collect(),
```
slowli, line 422: Nit: Can't this be expressed as `vec![Default::default(); l1_batch_headers.len()]`?

core/node/node_framework/src/implementations/layers/eth_watch.rs
```diff
@@ -66,21 +74,54 @@ impl WiringLayer for EthWatchLayer {
...
+        let sl_diamond_proxy_addr = if self.settlement_mode.is_gateway() {
+            self.gateway_chain_config
+                .clone()
+                .context("Lacking `gateway_contracts_config`")?
+                .diamond_proxy_addr
+        } else {
+            self.contracts_config.diamond_proxy_addr
+        };
+        tracing::info!(
+            "Diamond proxy address ethereum: {}",
```
slowli, line 86: Nit: IIUC, `Display` representation for `Address` and other fixed hashes like `H256` doesn't output all hex digits, but rather only starting 2 and ending 2 digits. Use the `Debug` repr (`{:?}`) to output the full address.

slowli (dismissed):

kelemeno (approved):

PR #3419, opened 2025-01-02
chore: Reduce diff with sync layer stable (ts/rs)

StanislavBreadless:
## What ❔

- `priority_tree_start_index` is now fetched dynamically.
- added encoding for relayed l2 calldata
- ported some backwards compatible changes for tests

## Checklist


core/lib/protobuf_config/src/proto/config/eth_sender.proto
```diff
@@ -54,7 +54,6 @@ message Sender {
   optional bool tx_aggregation_paused = 20; // required
   optional bool tx_aggregation_only_prove_and_execute = 21; // required
   optional uint32 time_in_mempool_in_l1_blocks_cap = 22; // optional
-  optional uint64 priority_op_start_index = 23; // optional
```
StanislavBreadless, line 57: protobuf is not compatible, but since this field has never been used, I think it should be fine to delete it since we never serialize these config into binary format

perekopskiy: let's better mark it as reserved just in case. Especially since field with `= 24` was just merged into main, lack of 23 will look weird.

perekopskiy (approved):

perekopskiy (approved):

PR #3422, opened 2025-01-03
feat: Features for an easier upgrade

StanislavBreadless:
## What ❔

Sending `execute` operations is stopped when gateway upgrade is ongoing
Also, validator timelock is fetched from state transition manager
Also, l1 shared bridge is updated from bridgehub

## Checklist


core/node/eth_sender/src/eth_tx_aggregator.rs
```diff
@@ -382,6 +464,34 @@ impl EthTxAggregator {
...
+        storage: &mut Connection<'_, Core>,
+        chain_protocol_version: ProtocolVersionId,
+    ) -> bool {
+        // If the gateway protocol version is present in the DB, and its timestamp is larger than `now`, it means that
+        // the upgrade process on the server has begun.
+        // However, if the protocol version on the contract is lower than the `gateway_upgrade`, it means that the upgrade has
+        // not yet completed.
+        let Some(gateway_protocol_version) = storage
+            .protocol_versions_dal()
+            .get_protocol_version_with_latest_patch(ProtocolVersionId::gateway_upgrade())
```
perekopskiy, line 481: Maybe let's better use `pending_protocol_version` from `blocks_dal()` that reads protocol version of the last sealed l2 block? L2 block with new protocol version will be sealed 1 second (l2 block deadline) after upgrade timestamp is reached and previous batch is sealed.

core/node/eth_sender/src/eth_tx_aggregator.rs
```diff
@@ -471,8 +583,32 @@ impl EthTxAggregator {
                 );
                 return Ok(());
             }
+
+            if Self::is_pending_gateway_upgrade(storage, chain_protocol_version_id).await
+                && agg_op.is_execute()
+            {
```
perekopskiy, line 589: Current implementation not only stops aggregating execute operations but also all operations that come after executes in the queue. There can be proof operations for old batches so it won't work. Let's move this to the place where `get_next_ready_operation` is called

core/node/node_framework/src/implementations/layers/web3_api/server/mod.rs
```diff
@@ -201,15 +206,28 @@ impl WiringLayer for Web3ServerLayer {
...
-                .map(|main_node_client| BridgeAddressesUpdaterTask {
-                    bridge_address_updater: bridge_addresses_handle.clone(),
-                    main_node_client: main_node_client.0,
-                    update_interval: self.optional_config.bridge_addresses_refresh_interval,
-                });
+
+        // In case it is an EN, the bridge addresses should be updated by fetching values from the main node.
+        // It is the main node, the bridge addresses need to be updated by querying the L1.
+
+        let bridge_addresses_updater_task = if let Some(main_node_client) = input.main_node_client {
```
perekopskiy, line 213: Let's also do the first `loop_iteration` right after task is created so we init API with correct bridge address.

core/node/eth_sender/src/eth_tx_aggregator.rs
```diff
@@ -52,9 +59,15 @@ pub struct EthTxAggregator {
     aggregator: Aggregator,
     eth_client: Box<dyn BoundEthInterface>,
     config: SenderConfig,
-    timelock_contract_address: Address,
+    // The validator timelock address provided in the config.
+    // If the contracts have the same protocol version as the state transition manager, the validator timelock
+    // from the state transition manager will be used.
+    // The address provided from the config is only used when there is a discrepancy between the two.
+    // TODO(EVM-932): always fetch the validator timelock from L1, but it requires a protocol change.
```
perekopskiy, line 66: Just so I understand better: we could resolve TODO by always using chain's validator timelock but there is no easy way to query it from L1; so the task is mostly about changing contracts to support the query, right?

StanislavBreadless: right

core/node/node_framework/src/implementations/layers/web3_api/server/mod.rs
StanislavBreadless: discussed in private, not needed

core/node/eth_sender/src/aggregator.rs
```diff
@@ -51,6 +51,57 @@ pub struct Aggregator {
...
+/// is stored to be logged.
+#[derive(Debug, Default)]
+pub(crate) struct OperationSkippingRestrictions {
+    pub(crate) commit_restriction: Option<&'static str>,
+    pub(crate) prove_restriction: Option<&'static str>,
+    pub(crate) execute_restriction: Option<&'static str>,
+}
+
+impl OperationSkippingRestrictions {
+    fn skip_and_log(&self, agg_op: &AggregatedOperation, reason: Option<&'static str>) -> bool {
```
perekopskiy, line 65: function name implies it returns true when op should be skipped. Let's either swap true<->false or change name

perekopskiy (approved):

PR #3434, opened 2025-01-07
chore: bump zksync-ethers

kelemeno:
## Checklist


core/tests/ts-integration/src/matchers/transaction.ts
```diff
@@ -71,6 +71,10 @@ export async function toBeReverted(
 
         return fail(message);
     } catch (error: any) {
+        if (error.toString().includes('reverted')) {
```
StanislavBreadless, line 74: this fix hides a very non-standard behavior of the SDK: it measures gasLimit even when it was already provided, it breaks some of the tests (they may pass, but they do not check the intended behavior)

StanislavBreadless (changes requested):

deniallugo-ml:

PR #3445, opened 2025-01-08
feat: add custom DA support in external node

dimazhornyk:
## What ❔

Enable DA info fetching in External Node.

## Why ❔

It is important because DA info is used to construct `commitBatches` transactions, which means for any custom L1DAValidator the consistency checker will fail to produce the same payload without it.

## Checklist


core/bin/external_node/src/config/mod.rs
```diff
@@ -1328,6 +1330,10 @@ impl ExternalNodeConfig<()> {
                 .context("could not load external node config (tree component params)")?,
             consensus_secrets: read_consensus_secrets()
                 .context("config::read_consensus_secrets()")?,
+            data_availability: (
+                DAClientConfig::from_env().ok(),
+                DataAvailabilitySecrets::from_env().ok(),
+            ),
```
slowli, line 1336: JFYI, DA configs present a somewhat major pain point migrating to a new config system (because their env-based and file-based representations significantly differ). So I'd prefer to not introduce new public (= difficult to change) config parameters before the migration. OTOH, I understand that this may be unrealistic depending on the priority of this change (the config system requires at least a couple more weeks to *start* migration).

core/lib/web3_decl/src/namespaces/zks.rs
```diff
@@ -142,4 +143,10 @@ pub trait ZksNamespace {
         &self,
         tx_bytes: Bytes,
     ) -> RpcResult<TransactionDetailedResult>;
+
+    #[method(name = "getDataAvailabilityDetails")]
```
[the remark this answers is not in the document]
dimazhornyk: moved to `unstable`, though I assume it will mature soon, and will be not only used by EN but also by Explorer, should it be moved back to `zks` then?

core/node/api_server/src/web3/namespaces/zks.rs
```diff
@@ -704,4 +706,32 @@ impl ZksNamespace {
...
+            ));
+        }
+
+        let mut connection = self.state.acquire_connection().await?;
+        let db_details = connection
+            .data_availability_dal()
+            .get_da_details_by_batch_number(batch)
+            .await
+            .map_err(DalError::generalize)?
+            .ok_or(Web3Error::NoBlock)?;
```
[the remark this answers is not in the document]
dimazhornyk: yeah, that makes sense, used None there

core/node/node_sync/src/data_availability_fetcher/mod.rs
```diff
@@ -0,0 +1,284 @@
...
+            )));
+        }
+
+        let Some(expected_inclusion_data) = da_details.inclusion_data else {
+            return Ok(StepOutcome::NoInclusionDataFromMainNode);
+        };
+
+        let inclusion_data = self
+            .da_client
+            .get_inclusion_data(da_details.blob_id.as_str())
```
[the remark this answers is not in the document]
dimazhornyk: the real purpose of this component is to populate the `data_availability` table to allow the consistency checker to generate the same commitment

this line is just a sanity check to make sure that the inclusion proof (part of the commitment we are most interested in) is the same both in the main node and the DA layer

theoretically, we could've also added a check that a blob from the DA layer has the same content as `pubdata_input` field of the batch, but this is checked by the L1DAValidator during `commitBatches` tx

do you think it makes sense to add that check here as well?

core/lib/da_client/src/lib.rs
```diff
@@ -23,6 +23,11 @@ pub trait DataAvailabilityClient: Sync + Send + fmt::Debug {
 
     /// Returns the maximum size of the blob (in bytes) that can be dispatched. None means no limit.
     fn blob_size_limit(&self) -> Option<usize>;
+
+    /// Returns the name of the client. It is expected that names are `.to_string()` of
+    /// PubdataType enum, but String is used here to keep the interface more generic and make this
+    /// crate independent of the workspace.
+    fn name(&self) -> String;
```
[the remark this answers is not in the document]
dimazhornyk: done the second option

core/bin/external_node/src/config/mod.rs
dimazhornyk: I'd be happy to postpone this and use new configs
but IIUC, this is required for v26 rollout, so most likely we can't afford to wait here, I will clarify that

core/node/da_dispatcher/src/da_dispatcher.rs
```diff
@@ -127,6 +127,9 @@ impl DataAvailabilityDispatcher {
                     batch.l1_batch_number,
                     dispatch_response.blob_id.as_str(),
                     sent_at.naive_utc(),
+                    // safe to unwrap because this is provided by the client
+                    client_type_to_pubdata_type(self.client.client_type()),
```
perekopskiy, line 131: nothing is unwrapped, isn't it?

core/node/api_server/src/web3/namespaces/unstable/mod.rs
```diff
@@ -151,4 +152,32 @@ impl UnstableNamespace {
...
+
+    pub async fn get_data_availability_details_impl(
+        &self,
+        batch: L1BatchNumber,
+    ) -> Result<Option<DataAvailabilityDetails>, Web3Error> {
+        if self.state.api_config.l1_batch_commit_data_generator_mode
+            == L1BatchCommitmentMode::Rollup
+        {
+            return Err(Web3Error::MethodNotImplemented);
+        }
```
perekopskiy, line 164: it's allowed for rollup to become validium, so it seems this should be changed. Also, I have no idea why we even have `l1_batch_commit_data_generator_mode` in API, its usage is most likely incorrect

core/lib/dal/migrations/20241212121618_add_type_column_to_data_availability.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE data_availability ADD COLUMN client_type TEXT NOT NULL DEFAULT 'NoDA';
```
perekopskiy, line 1: How is it going to work for existing envs with client_type != NoDA?

core/node/da_dispatcher/src/da_dispatcher.rs
dimazhornyk: yep, old comment, removed it

core/node/api_server/src/web3/namespaces/unstable/mod.rs
dimazhornyk: we most likely have `l1_batch_commit_data_generator_mode` in the API to provide it as a part of a genesis config to ENs

but the point makes sense, so I removed that check, now if the data is missing - we will return Ok(None)

core/lib/dal/migrations/20241212121618_add_type_column_to_data_availability.up.sql
dimazhornyk: we would have to manually fix it for them after applying the migration
unfortunately, I don't see any other way around it

dimazhornyk: we can also make the mismatch of the client in EN config and from the server a non-fatal error on the EN side, specifically for this case to give us time to update the values after running migrations

core/bin/external_node/src/config/mod.rs
dimazhornyk: both DA secrets and client configs use `env::var` because nested structs are not correctly processed by `envy`
adding a prefix there would be quite a challenge, we would have to change `FromEnv` trait for that, so I'm not sure it would be worth it

core/lib/dal/migrations/20241212121618_add_type_column_to_data_availability.up.sql
dimazhornyk: That makes sense, I made the `pubdata_type` optional, and added a following check in DA fetcher in EN:
```rust
    let Some(pubdata_type) = da_details.pubdata_type else {
        tracing::warn!(
            "No pubdata type for L1 batch #{}; waiting for the main node to provide it",
            l1_batch_to_fetch
        );
        return Ok(StepOutcome::NoProgress);
    };
```

core/node/node_sync/Cargo.toml
```diff
@@ -26,6 +26,7 @@ zksync_concurrency.workspace = true
 zksync_consensus_roles.workspace = true
 vise.workspace = true
 zksync_vm_executor.workspace = true
+zksync_da_client.workspace = true
```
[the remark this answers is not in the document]
dimazhornyk: The client implementations might be moved out, but the client trait and error types are intended to be used like they are here

core/lib/types/src/api/mod.rs
```diff
@@ -967,6 +968,16 @@ pub struct FeeHistory {
     pub l2_pubdata_price: Vec<U256>,
 }
 
+/// The data availability details type. Used exclusively in Validiums.
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct DataAvailabilityDetails {
```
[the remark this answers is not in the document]
dimazhornyk: Mostly domain separation, and things that might not belong to the `basic_types` like `#[serde(rename_all = "camelCase")]`. Not sure whether they will necessarily evolve in different directions though

I also checked and didn't see us using `basic_types` in API, so kept it as is for now. LMK if you think that only one has to be used

core/lib/types/Cargo.toml
```diff
@@ -18,6 +18,7 @@ zksync_contracts.workspace = true
 zksync_mini_merkle_tree.workspace = true
 zksync_protobuf = { workspace = true, optional = true }
 zksync_crypto_primitives.workspace = true
+zksync_da_client.workspace = true
```
perekopskiy, line 21: I think we can avoid this dependency. You can move `client_type_to_pubdata_type` to `zksync_da_client` which already depends on `zksync_types`.

dimazhornyk: `zksync_da_client` doesn't depend on any packages from the workspace (including `zksync_types`) as it is supposed to be imported by the external projects, so that won't be possible

perekopskiy (dismissed):

slowli (commented):
Other than env-based config param naming, looks good.

slowli (approved):

alexandrst88 (approved):

PR #3448, opened 2025-01-09
feat: Added tx hashes to upgrade yaml, disabled no_governance_prepare_gateway

mm-zk:
## What ❔

* Updated contracts to most recent one (including PR 1187 from contracts)
* Added transaction hashes  to the upgrade yaml file - so that tools can verify bytecodes
* fix to governance_upgrade_timer_initial_delay -  it was missing before.


zkstack_cli/crates/config/src/forge_interface/gateway_ecosystem_upgrade/input.rs
```diff
@@ -33,6 +34,8 @@ impl GatewayEcosystemUpgradeInput {
             era_chain_id,
             testnet_verifier,
             owner_address: current_contracts_config.l1.governance_addr,
+            // FIXME - where should it be loaded from?
+            governance_upgrade_timer_initial_delay: 15,
```
StanislavBreadless, line 38: in local testing providing even 0 is fine

StanislavBreadless: In prod I would expect us manually compiling the input files for foundry scripts

zkstack_cli/crates/zkstack/src/commands/ecosystem/gateway_upgrade.rs
```diff
@@ -145,10 +155,34 @@ async fn no_governance_prepare(
...
+                .join(format!(
+                    "broadcast/EcosystemUpgrade.s.sol/{}/run-latest.json",
+                    l1_chain_id
+                )),
+        )
+        .context("Failed to read broadcast file")?;
+        serde_json::from_str(&file_content).context("Failed to parse broadcast file")?
+    };
+
+    dbg!(&broadcast_file);
```
StanislavBreadless, line 174: it will pollute the screen when not needed, better to remove 
```suggestion
```

PR #3449, opened 2025-01-09
feat: Additional gateway upgrade testing

StanislavBreadless:
## What ❔

- Add server feature to automatically switch l2 shared bridge as well. Also, l2 wrapped base token is also fetched from L1 upon upgrade. This reduces the risk of misconfiguration.
- Added WIP tool for zkstack to generate the calldata for the admin. It will have to be prettified before going public, but it is in `dev` so it is okay

## Checklist


core/lib/dal/src/eth_sender_dal.rs
```diff
@@ -86,6 +86,25 @@ impl EthSenderDal<'_, '_> {
...
+        let count = sqlx::query!(
+            r#"
+            SELECT
+                COUNT(*)
+            FROM
+                eth_txs
+            WHERE
+                confirmed_eth_tx_history_id IS NULL
+            "#
+        )
```
perekopskiy, line 99: This query can be confusing as we use "inflight" for txs that are not yet confirmed and have at least one entry in eth_txs_history. Maybe let's use `get_unconfirmed_txs_count`?

core/lib/web3_decl/src/namespaces/zks.rs
```diff
@@ -142,4 +142,11 @@ pub trait ZksNamespace {
         &self,
         tx_bytes: Bytes,
     ) -> RpcResult<TransactionDetailedResult>;
+
+    // This is an unstable method that may get deleted in the future,
+    // however, it is part of the `zks` namespace in order to have it enabled by
+    // default on all networks. It is needed for the upgrade functionality and may be
+    // stabilized later with a different interface.
+    #[method(name = "unstableInflightTxsCount")]
+    async fn get_inflight_txs_count(&self) -> RpcResult<usize>;
```
perekopskiy, line 151: please move to unstable namespace

perekopskiy (approved):

PR #3451, opened 2025-01-10
ci: finalize ci for sync layer stable

StanislavBreadless:
## Checklist


core/tests/ts-integration/tests/base-token.test.ts
```diff
@@ -78,7 +78,7 @@ describe('base ERC20 contract checks', () => {
         // TODO: should all the following tests use strict equality?
 
         const finalEthBalance = await alice.getBalanceL1();
-        expect(initialEthBalance).toBeGreaterThan(finalEthBalance + fee); // Fee should be taken from the ETH balance on L1.
+        expect(initialEthBalance).toBeGreaterThanOrEqual(finalEthBalance + fee); // Fee should be taken from the ETH balance on L1.
```
perekopskiy, line 81: This test was flaky. tbh I'm not sure how it was working before. For chains with custom base token `initialEthBalance` should be equal to `finalEthBalance + fee` i.e. ether should be spent only for l1 tx fee. @StanislavBreadless I'm not sure if it's worth investigating because now it seems to be working as expected, agree?

StanislavBreadless: okay agree

PR #3452, opened 2025-01-10
fix(api): Change `contractAddress` assignment for transaction receipts

slowli:
## What ❔

Changes how `contractAddress` is assigned for transaction receipts. With these changes, it is assigned for EVM deployment transactions (ones with `to == None`) and EraVM deployments (i.e., calls to `ContractDeployer.{create, create2, createAccount, create2Account}`) regardless of whether the transaction succeeds.

## Why ❔

Improves EVM compatibility.

## Checklist


core/node/api_server/src/utils.rs
```diff
@@ -66,3 +77,383 @@ macro_rules! report_filter {
...
+        assert_eq!(
+            statuses_and_contract_addresses,
+            expected_statuses_and_contract_addresses
+        );
+    }
+
+    /// Because of AA support, determining deployment nonces is non-trivial. E.g., it would be incorrect
+    /// to define a deployment nonce increment as a count of successful deployment txs from an EOA because
+    /// the account might have a non-default AA that can be called as a contract (so it can deploy contracts
+    /// outside of deployment txs).
```
slowli, line 381: Just to understand better: Does this logic apply in practice (i.e., there are no implied limitations of what AA contracts can do)?

core/node/api_server/src/utils.rs
```diff
@@ -66,3 +77,383 @@ macro_rules! report_filter {
...
+
+    let mut filled_receipts = Vec::with_capacity(receipts.len());
+    let mut receipt_indexes_with_unknown_nonce = HashSet::new();
+    for (i, mut receipt) in receipts.into_iter().enumerate() {
+        receipt.inner.contract_address = if receipt.inner.to.is_none() {
+            // This is an EVM deployment transaction.
+            Some(deployed_address_evm_create(
+                receipt.inner.from,
+                receipt.nonce,
+            ))
```
slowli, line 102: To highlight: Is this the correct derivation scheme? I.e., is the nonce indeed taken from the transaction (not the deployment nonce etc.)?

core/node/api_server/src/utils.rs
```diff
@@ -66,3 +77,383 @@ macro_rules! report_filter {
...
+            .filter_map(|(i, receipt)| {
+                receipt_indexes_with_unknown_nonce
+                    .contains(&i)
+                    .then_some(receipt)
+            })
+            .collect();
+        fill_receipts_with_unknown_nonce(storage, block_number, unknown_receipts).await?;
+    }
+
+    Ok(filled_receipts)
```
slowli, line 142: Would it make sense to check whether `contractAddress` for successful deployment transactions correspond to `ContractDeployed` events? These events are loaded anyway in `fill_receipts_with_unknown_nonce` (and could be cached as per another comment).

slowli (commented):
This is a breaking change in terms of API. `contractAddress` will no longer be set for some of transactions (ones that we don't consider EraVM deployment txs). OTOH, it *will* be set for failed deployment txs.

perekopskiy (approved):
Looks good. I haven't checked protocol/nonce questions, so an approval from @0xVolosnikov would be nice to have

core/node/api_server/src/utils.rs
0xVolosnikov: Yep, for EOA transaction nonce is used (EVM doesn't have separate deployment nonce concept)

core/node/api_server/src/utils.rs
```diff
@@ -66,3 +77,383 @@ macro_rules! report_filter {
...
+        let sender = receipt.from;
+        let tx_index = receipt.transaction_index.as_u64();
+        let initial_nonce = nonces_at_block_start
+            .get(&nonce_key)
+            .copied()
+            .flatten()
+            .unwrap_or_default();
+        let (_, initial_deploy_nonce) = decompose_full_nonce(h256_to_u256(initial_nonce));
+
+        let nonce_increment = deployment_events
```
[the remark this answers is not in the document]
0xVolosnikov: Yes, for **EraVM** deployments. Except for an upgrade transaction where it may not be the case.

core/node/api_server/src/utils.rs
0xVolosnikov: You are right.

core/node/api_server/src/utils.rs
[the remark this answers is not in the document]
0xVolosnikov: Also AA can break this assumption

core/node/api_server/src/utils.rs
0xVolosnikov: Sounds reasonable. 
Note: in theory, AA can deploy multiple contracts in one transaction (an exotic case, but generally looks real)

slowli: IIUC, a custom AA can treat deployments (or more generally, calldata / `to`) in an arbitrary way provided that some basic invariants are preserved. So, maybe more realistically than deploying multiple contracts, an AA can use another nonce increment scheme, thus making address derivation logic from this PR incorrect. So would it make sense to to use the logic implemented in this PR for custom AAs? There are a couple of alternatives:

- Always return `contractAddress: null` for custom AAs.
- Base `contractAddress` on the emitted `ContractDeployed` events, as previously. In this case, there may indeed be multiple contracts deployed in a single transaction, and I don't think Web3 clients generally expect `contractAddress` to be an array.

I'd prefer one of the 2 simpler approaches (i.e., either not treating AAs specially, or returning `contractAddress: null`), depending on whether we can have false non-null `contractAddress` values. If we really want to support `contractAddress` for custom AAs, we can add it incrementally. WDYT?

core/node/api_server/src/utils.rs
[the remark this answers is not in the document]
slowli: AFAICT [from this logic](https://github.com/matter-labs/era-contracts/blob/release-candidate-v27/system-contracts/contracts/ContractDeployer.sol):

- The deployment nonce [is always incremented](https://github.com/matter-labs/era-contracts/blob/5731979de1e1a81483503f4f580d2ae96bdaeb95/system-contracts/contracts/ContractDeployer.sol#L475-L476) for the *deployed contract*.
- The deployment nonce [is incremented](https://github.com/matter-labs/era-contracts/blob/5731979de1e1a81483503f4f580d2ae96bdaeb95/system-contracts/contracts/ContractDeployer.sol#L182-L185) for `msg.sender` if it is distinct from `tx.origin` (also for `create2` deploys).

In summary, it is incremented for all EraVM deployments (can be detected by the bytecode hash mentioned in the `ContractDeployed` events, right?), and never for EVM deployments. *Unless* we have custom AA support for EVM contracts, in which case everything falls apart 🙃 Is it correct that the initial release won't support these? But generally, if (when?) we do, it'll become a mess to support (unless we'll ignore AAs for `contractAddress` purposes, as discussed in another comment.)

core/node/api_server/src/utils.rs
0xVolosnikov: @slowli Rn AA can't arbitrary change **deployment** nonce, only "transaction" nonce

core/node/api_server/src/utils.rs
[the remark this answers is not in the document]
0xVolosnikov: Yep, you are right. We won't have EVM AA in first version. It's likely that it'll be done later (ZKOS). 
Generally it doesn't feel right that we try to treat AA transactions in ~same way, wdyt?

core/node/api_server/src/utils.rs
0xVolosnikov: > Always return contractAddress: null for custom AAs

It _may_ break integrated Block explorers a bit I think. So it will became their problem to handle deployment address.

0xVolosnikov: fyi: at least ERC4337 specifies special endpoints for user operations (AA): https://eips.ethereum.org/EIPS/eip-4337#-eth_getuseroperationreceipt

So, as I understand, it should be explicitly separated from usual transactions _(a bundle of operations is a transaction)_.

0xVolosnikov: Strictly speaking, according to Eth JSON-RPC API, we should set `contractAddress` to false both for all EraVM deployments (they are just simple transactions from EVM point of view).

Imo it should be ok to return false at least for custom AAs, but it should be communicated wi BE

slowli:
@0xVolosnikov I've disabled `contractAddress` computations for custom AAs; i.e., it's now always set to `null` for them.

0xVolosnikov (approved):

PR #3475, opened 2025-01-15
fix(api): Improve estimation for gas_per_pubdata_limit

popzxc:
Fixes the `gas_per_pubdata_limit` field calculation. Right now, we provide the value we get from deriving the existing batch input, which is most likely lower than limit provided in fee estimation request.
Given that we scale the overall fee, it should be safe to scale the limit as well, as long as we don't go beyond the limit supplied in the tx.

Providing the limit as-is without scaling can be very flaky: transaction risks to be left in the mempool over minor `gas_per_pubdata` changes.

### DIFF
--- a/core/node/api_server/src/tx_sender/gas_estimation.rs
+++ b/core/node/api_server/src/tx_sender/gas_estimation.rs
@@ -530,11 +530,18 @@ impl<'a> GasEstimator<'a> {
             gas_per_pubdata_byte = self.gas_per_pubdata_byte
         );
 
+        // Given that we scale overall fee, we should also scale the limit for gas per pubdata price that user agrees to.
+        // However, we should not exceed the limit that was provided by the user in the initial request.
+        let gas_per_pubdata_limit = std::cmp::min(
+            ((self.gas_per_pubdata_byte as f64 * estimated_fee_scale_factor) as u64).into(),
+            self.transaction.gas_per_pubdata_byte_limit(),
+        );
+
         Ok(Fee {
             max_fee_per_gas: self.base_fee.into(),
             max_priority_fee_per_gas: 0u32.into(),
             gas_limit: full_gas_limit.into(),
-            gas_per_pubdata_limit: self.gas_per_pubdata_byte.into(),
+            gas_per_pubdata_limit,
         })
     }
 }


core/node/api_server/src/tx_sender/gas_estimation.rs
```diff
@@ -530,11 +530,18 @@ impl<'a> GasEstimator<'a> {
             gas_per_pubdata_byte = self.gas_per_pubdata_byte
         );
 
+        // Given that we scale overall fee, we should also scale the limit for gas per pubdata price that user agrees to.
+        // However, we should not exceed the limit that was provided by the user in the initial request.
+        let gas_per_pubdata_limit = std::cmp::min(
+            ((self.gas_per_pubdata_byte as f64 * estimated_fee_scale_factor) as u64).into(),
+            self.transaction.gas_per_pubdata_byte_limit(),
+        );
```
perekopskiy, line 538: While it should work as a quick fix I think the problem is more deep.
`estimated_fee_scale_factor` is set from config var `estimate_gas_scale_factor` /// The factor by which to scale the gasLimit
we also have `gas_price_scale_factor` /// The multiplier to use when suggesting gas price

Latter is used in `ApiFeeInputProvider` but the problem is that provider returns and uses that factor for {fair_l2_gas_price, fair_pubdata_price} while `gas_per_pubdata = ceil(fair_pubdata_price / fair_l2_gas_price)` so factor is not used for `gas_per_pubdata`. IMO we should rework fee provider so it's responsible for providing `gas_per_pubdata` estimate.
Also, not sure about it but it seems if compute_overhead_part=1.0 then `gas_price_scale_factor` has more effect on `fair_l2_gas_price` than `fair_pubdata_price` effectively decreasing gas_per_pubdata

slowli (commented):
Dumb question: Do I understand correctly that the changed logic influences `zks_estimateFee` output only? (There are 2 other uses of gas estimation, `eth_estimateGas` and `zks_estimateL1ToL2Gas`, but they both seem to output only `Fee.gas_limit`.) If so, `gas_per_pubdata_limit` is overwritten before estimation to a hard-coded value (50,000); is this fine?

https://github.com/matter-labs/zksync-era/blob/c156e791c1d8bbdea32873f43e976d35b228d38e/core/node/api_server/src/web3/namespaces/zks.rs#L81-L84

It sort of contradicts "we should not exceed the limit that was provided by the user in the initial request", since the value provided by the user is ignored.

core/node/api_server/src/tx_sender/gas_estimation.rs
popzxc: Agree, but not sure who/when will be able to tackle this in depth.

popzxc:
@slowli 

> Do I understand correctly that the changed logic influences zks_estimateFee output only?

Yes.

> If so, gas_per_pubdata_limit is overwritten before estimation to a hard-coded value (50,000); is this fine?

Huh, that's both fine and not fine. I would expect `zks_estimateFee` to respect the value if provided, but the reality we're in is that probably nobody actively monitors this value (there is no good tooling and/or documentation for that). So my comment was about the correctness of the _implementation_, but if we are to discuss the real state of affairs -- it's OK to exceed, since I'm fairly sure that nobody relies on this logic.

popzxc:
@perekopskiy @slowli do we want to merge this PR, or do you think a bigger rework is mandatory?

slowli:
I think this PR is fine to merge (AFAIU, it shouldn't break anything), but I won't pretend I'm an expert on the domain, so it's difficult for me to estimate how complete this fix is.

slowli (approved):

PR #3478, opened 2025-01-15
feat: chunk storage logs in dal

ischasny:
In the case of custom genesis state quite a few storage logs can be associated with a single L1 batch (10s of millions depending on how big the initial state was). For such scenarios - chunk storage logs-related queries to avoid overloading on the database / driver sides.


deniallugo-ml:

PR #3487, opened 2025-01-16
fix(en): more correct fees algorithm for external-node

tomg10:


core/node/fee_model/src/l1_gas_price/main_node_fetcher.rs
```diff
@@ -73,6 +99,29 @@ impl MainNodeFeeParamsFetcher {
...
+        match fee_params {
+            FeeParams::V1(..) => FeeParams::V1(FeeParamsV1 {
+                config: FeeModelConfigV1 {
+                    minimal_l2_gas_price: batch_fee_input.fair_l2_gas_price(),
+                },
+                l1_gas_price: batch_fee_input.l1_gas_price(),
+            }),
+            FeeParams::V2(params) => {
+                let mut config = params.config();
+                config.minimal_l2_gas_price = batch_fee_input.fair_l2_gas_price();
```
perekopskiy, line 117: this is incorrect: `minimal_l2_gas_price` and `fair_l2_gas_price` are different. There is no easy way to deduct `minimal_l2_gas_price` from `batch_fee_input`

core/lib/types/src/fee_model.rs
```diff
@@ -284,6 +284,9 @@ impl FeeParamsV2 {
         self.convert_to_base_token(self.l1_pubdata_price)
     }
 
+    pub fn conversion_ratio(&self) -> BaseTokenConversionRatio {
+        self.conversion_ratio
+    }
     /// Converts the fee param to the base token.
```
perekopskiy, line 290: nit: new line

core/node/fee_model/src/l1_gas_price/main_node_fetcher.rs
```diff
@@ -73,6 +99,29 @@ impl MainNodeFeeParamsFetcher {
 
 impl BatchFeeModelInputProvider for MainNodeFeeParamsFetcher {
     fn get_fee_model_params(&self) -> FeeParams {
-        *self.main_node_fee_params.read().unwrap()
+        let fee_params = *self.main_node_fee_params.read().unwrap();
+        let batch_fee_input = *self.main_node_batch_fee_input.read().unwrap();
+        if batch_fee_input.is_none() {
+            return fee_params;
+        }
+        let batch_fee_input = batch_fee_input.unwrap();
```
perekopskiy, line 107: nit: you can use `let Some() ... else ...` and avoid `unwrap`

perekopskiy (commented):
Please add PR description

core/node/fee_model/src/l1_gas_price/main_node_fetcher.rs
tomg10: I agree, but at the cost of extra indentation, so I'd prefer to keep it that way. It's one of very few rust-specific syntactic sugars that Imo do more harm than good in terms of readability.

itegulov: What extra indentation? I think what @perekopskiy is suggesting is doing this:
```suggestion
        let Some(batch_fee_input) = batch_fee_input else {
            return fee_params;
        }
```
Which is IMO more idiomatic

core/node/fee_model/src/l1_gas_price/main_node_fetcher.rs
```diff
@@ -73,6 +99,27 @@ impl MainNodeFeeParamsFetcher {
...
+            }),
+            FeeParams::V2(params) => {
+                return FeeParams::V2(FeeParamsV2::new(
+                    params.config(),
+                    batch_fee_input.l1_gas_price(),
+                    batch_fee_input.fair_pubdata_price(),
+                    params.conversion_ratio(),
+                ));
+            }
+        }
```
itegulov, line 123: Why do we need all of this logic? Can't we just override `get_batch_fee_input_scaled` instead and return `*self.main_node_batch_fee_input.read().unwrap()` there? EN's scaling factor will be ignored but arguably this logic is not designed well in general (i.e. is there a reason to scale gas price differently on EN?). Also, `fn get_fee_model_params(&self) -> FeeParams` being a part of the base trait is strange too, it's more like an implementation detail for half of the implementations.

core/node/fee_model/src/l1_gas_price/main_node_fetcher.rs
tomg10: nvm, my brain malfunctioned, thanks!

PR #3508, opened 2025-01-20
ci: add unified release please and crates.io publishing

antonbaliasnikov:
## What ❔

CI:
* [x] Add unified release-please workflow with automated crates.io publishing
* [x] Add separate workflow to publish crates manually if required

Rust:
* [x] `release-please` is now updating workspace version in Cargo.toml and all crates are re-using the workspace version

## Checklist


.github/workflows/release-please.yml
```diff
@@ -1,29 +1,32 @@
...
+  # Prepare the release PR with changelog updates and create github releases
+  release-please:
+    uses: matter-labs/zksync-ci-common/.github/workflows/release-please.yaml@v1
+    secrets:
+      slack_webhook: ${{ secrets.SLACK_WEBHOOK_RELEASES }}  # Slack webhook for notifications
+      gh_token: ${{ secrets.GITHUB_TOKEN }}                 # GitHub token for release-please
+    with:
+      config: '.github/release-please/config.json'          # Path to the configuration file
+      manifest: '.github/release-please/manifest.json'      # Path to the manifest file
+      update-cargo-lock: true                               # Update Cargo.lock file in the release PR
```
slowli, line 29: Dumb question: Will this update other `Cargo.lock`s in the repo (i.e., core deps for the prover / zkstack workspaces)? Or how are they supposed to be updated?

core/Cargo.toml
```diff
@@ -257,70 +257,70 @@ zksync_protobuf = "=0.7.0"
...
-zksync_storage = { version = "0.1.0", path = "lib/storage" }
-zksync_system_constants = { version = "0.1.0", path = "lib/constants" }
-zksync_tee_verifier = { version = "0.1.0", path = "lib/tee_verifier" }
-zksync_test_contracts = { version = "0.1.0", path = "lib/test_contracts" }
-zksync_types = { version = "0.1.0", path = "lib/types" }
-zksync_utils = { version = "0.1.0", path = "lib/utils" }
-zksync_web3_decl = { version = "0.1.0", path = "lib/web3_decl" }
-zksync_crypto_primitives = { version = "0.1.0", path = "lib/crypto_primitives" }
-zksync_external_price_api = { version = "0.1.0", path = "lib/external_price_api" }
+zksync_multivm = { version = "26.1.0-non-semver-compat", path = "lib/multivm" }
```
slowli, line 260: Dumb question: Since releases are not semver-compatible, would it make sense to have exact / pinned versions specified for deps? Or would it mess with the release-please logic?

.github/workflows/release-please.yml
antonbaliasnikov: I think we need to discuss the way we would like it to be.

Before, we were specifying dependencies like that:
```
zksync_vlog = { path = "../core/lib/vlog" }
```

If we publish it to crates.io, we can switch to version + path:
```
zksync_vlog = { version = "=26.1.0-non-semver-compat", path = "../core/lib/vlog" }
```

But if we try to run `cargo update --workspace` in the `prover` after `core` packages version change, we will get the error:
```
error: failed to select a version for the requirement `zksync_config = "=26.1.0-non-semver-compat"`
candidate versions found which didn't match: 26.1.1-non-semver-compat
location searched: /Users/abalias/projects/matter-labs/zksync-era/core/lib/config
required by package `zksync_circuit_prover v17.1.1 (/Users/abalias/projects/matter-labs/zksync-era/prover/crates/bin/circuit_prover)`
if you are looking for the prerelease package it needs to be specified explicitly
    zksync_config = { version = "26.1.1-non-semver-compat" }
```

If we use `path` only, then everything would be OK and we can do that.

But if we would like to use `version` from crates.io as well, then we do not run `cargo update` for other workspaces in the release PR, but instead we have [an additional helper job](https://github.com/matter-labs/zksync-ci-common/blob/main/.github/workflows/release-please.yaml#L243-L283) that after each component release will upgrade Cargo.toml and Cargo.lock dependencies from crates.io for all dependent crates.

@slowli , @popzxc it would be nice to have a quick call probably to make sure we're aligned.

core/Cargo.toml
antonbaliasnikov: Yeah, sure. I've changed to pinned versions.

.github/workflows/release-please.yml
antonbaliasnikov: I also thought we're moving to the point where we remove `path` cross-workspace, then Cargo.lock in others should not be updated by the time we release one component, right?

slowli (approved):

alexandrst88 (approved):

PR #3527, opened 2025-01-24
feat(contract-verifier): Partial matching & automatic verification

### LINKED ISSUE #3190: Support partial matches in contract verifier
#### 📝 Description

Please first read about full vs partial matches for contract verification [here](https://docs.sourcify.dev/docs/full-vs-partial-match/). Currently contract verifier support only full matches. This request is for supporting partial matches. Verification logic should be updated so that if a full match wasn't found but there is a partial one then the request is persisted and API exposes its data. Also, if a contract is partially verified it should be possible to replace its verification info if some full match request comes in, but not vice versa.

#### 🤔 Rationale

Partial match is useful when the original source code was lost after deployment but developer managed to replicate it apart from metadata. Also, etherscan performs only partial match so devs get used to it. There are cases when only partial match verification is possible without adding a new feature to contract verifier, e.g. when contract's libraries are linked dynamically.

#### 📋 Additional Context

After it's done on backend side it would be nice to display verification type on frontend

### LINKED ISSUE #3309: feat(contract-verifier): Auto-verify the same bytecode
Depends on https://github.com/matter-labs/zksync-era/issues/3190

We need to automatically verify the contracts with the same bytecode:
- When a new contract is deployed, we should check if the verification data for this bytecode exists. If so, it shouldn't require additional verification.
- When a new contract is verified, we should check if there are other contracts with the same bytecode. If so, they should become verified.
- This should interact with partial matches: if there are differences in metadata, the contract should still be considered verified with partial match, but it should still be possible to do the "full" verification later.

⚠️  Don't treat the acceptance criteria as an implementation proposal: _likely_ we don't actually need to scan the whole contracts table each time we deploy a contract or verify it. Most likely, we may just change the way we query verification data.

popzxc:
Fixes #3190
Fixes #3309

I still have to test the migration logic (want to make sure that Era mainnet data can be migrated correctly & quickly), but the PR is reviewable otherwise.

This one is much bigger than I indended it to be, sorry 🥲 

- Changes `zksolc` used in contract verifier tests to 1.5.10 (old one didn't have `ipfs` metadata hash support).
- Fixes problems that occured with newer compilers (e.g. `Yul` bytecode extraction).
- Makes it easier to run contract verifier tests locally (pins compiler versions, so that if you have a lot of compilers locally, tests don't crash).
- Introduces partial matching for contracts
- Reworks the schema so that contracts with matching bytecode can be "automatically" verified.
- Adds a migration to the new table.


core/bin/contract-verifier/src/main.rs
```diff
@@ -25,6 +25,32 @@ struct Opt {
     secrets_path: Option<PathBuf>,
 }
 
+async fn perform_storage_migration(pool: &ConnectionPool<Core>) -> anyhow::Result<()> {
```
popzxc, line 28: OK, this is very slow (6sec per 1000 rows on a high-end CPU). Working on fixing the migration speed.

core/lib/contract_verifier/src/compilers/solc.rs
```diff
@@ -128,7 +132,11 @@ impl Compiler<SolcInput> for Solc {
         if output.status.success() {
             let output = serde_json::from_slice(&output.stdout)
                 .context("zksolc output is not valid JSON")?;
-            parse_standard_json_output(&output, input.contract_name, input.file_name, true)
+            let output =
+                parse_standard_json_output(&output, input.contract_name, input.file_name, true)?;
+            let id =
+                ContractIdentifier::from_bytecode(BytecodeMarker::Evm, output.deployed_bytecode());
```
slowli, line 138: It seems that in all `Compiler` impls, the ID is constructed from `output.deployed_bytecode()` + bytecode type. If this is the case, can't ID construction be moved up the call chain? This would allow retaining a simpler `Compiler` interface.

core/lib/types/src/contract_verification/contract_identifier.rs
```diff
@@ -0,0 +1,388 @@
...
+/// [zksolc_keccak]: https://matter-labs.github.io/era-compiler-solidity/latest/02-command-line-interface.html#--metadata-hash
+// Note: there are missing opportunities here, e.g. Etherscan is able to detect the contracts
+// that differ in creation bytecode and/or constructor arguments (for partial match). This is
+// less relevant for ZKsync, since there is no concept of creation bytecode there; although
+// this may become needed if we will extend the EVM support.
+#[derive(Debug, Clone, Copy)]
+pub struct ContractIdentifier {
+    /// Marker of the bytecode of the contract.
+    pub bytecode_marker: BytecodeMarker,
+    /// SHA3 (keccak256) hash of the full contract bytecode.
```
slowli, line 22: AFAIU, SHA3-256 and keccak256 differ by the finalization padding (in other words, they are 2 different hash functions using the same general construction). I.e., to avoid confusion, it makes sense to mention keccak256 only.

core/lib/types/src/contract_verification/contract_identifier.rs
```diff
@@ -0,0 +1,388 @@
...
+    /// Metadata is different.
+    Partial,
+    /// No match.
+    None,
+}
+
+/// Metadata detected in the contract bytecode.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum DetectedMetadata {
+    /// Keccek256 hash of the metadata detected (only for EraVM).
```
slowli, line 46: Typo: Kecc**e**k256 -> keccak256.

core/lib/types/src/contract_verification/contract_identifier.rs
```diff
@@ -0,0 +1,388 @@
...
+            BytecodeMarker::Evm => {
+                // On EVM, there is no padding.
+                self_.metadata_size = full_metadata_length;
+                &bytecode[..bytecode.len() - full_metadata_length]
+            }
+            BytecodeMarker::EraVm => {
+                // On EraVM, there is padding:
+                // 1. We must align the metadata length to 32 bytes.
+                // 2. We may need to add 32 bytes of padding.
+                let aligned_metadata_length = (metadata_length + 31) / 32 * 32;
```
slowli, line 147: Nit: It's possible to use `metadata_length.div_ceil(32) * 32`.

core/lib/types/src/contract_verification/contract_identifier.rs
```diff
@@ -0,0 +1,388 @@
...
+                // This could be overridden if CBOR metadata is detected.
+                self_.bytecode_without_metadata_sha3 = Some(DetectedMetadata::Keccak256(hash));
+            }
+        }
+
+        // Try to detect CBOR metadata.
+
+        // Last two bytes is the length of the metadata in big endian.
+        if bytecode.len() < 2 {
+            return self_;
```
slowli, line 117: Nit: I'd refactor this part into a separate method returning a `Result` with a simple error type (maybe `()` or `&'static str`), so that shortcuts during extraction are more idiomatic.

core/lib/types/src/contract_verification/contract_identifier.rs
```diff
@@ -0,0 +1,388 @@
...
+                    self_.metadata_size = full_aligned_metadata_length;
+                    &bytecode[..bytecode.len() - full_aligned_metadata_length]
+                } else {
+                    // Padding wasn't added, strip metadata only.
+                    self_.metadata_size = aligned_metadata_length;
+                    &bytecode[..bytecode.len() - aligned_metadata_length]
+                }
+            }
+        };
+        let hash = H256::from_slice(&keccak256(bytecode_without_metadata));
```
slowli, line 171: Nit: `H256(keccak256(_))` should work.

core/lib/contract_verifier/src/tests/real.rs
```diff
@@ -19,38 +26,77 @@ impl Toolchain {
...
-            zkvyper: versions.zkvyper.into_iter().next()?,
+    fn new(versions: SupportedCompilerVersions) -> anyhow::Result<Self> {
+        // Stored compilers for our fork are prefixed with `zkVM-`.
+        let eravm_solc = format!("zkVM-{ERA_VM_SOLC_VERSION}");
+        // Stored compilers for vyper do not have `v` prefix.
+        let vyper = VYPER_VERSION.strip_prefix("v").unwrap().to_owned();
+        anyhow::ensure!(
+            versions.solc.contains(SOLC_VERSION),
+            "Expected solc version {} to be installed, but it is not",
+            SOLC_VERSION
```
slowli, line 55: Nit: The variable can be inlined into the formatted string. Ditto below.

core/lib/contract_verifier/src/tests/real.rs
```diff
@@ -505,7 +598,162 @@ async fn using_real_compiler_in_verifier(bytecode_kind: BytecodeMarker, toolchai
...
+                },
+            },"settings": {
+                "outputSelection": {
+                "*": {
+                    "": [
+                    "abi"
+                    ],
+                    "*": [
+                    "abi"
+                    ]
```
slowli, line 633: Nit: Weird indentation here (I'd format `"": ["abi"]` etc.) and below in `"optimizer"`.

core/lib/dal/src/contract_verification_dal.rs
```diff
@@ -560,6 +583,285 @@ impl ContractVerificationDal<'_, '_> {
...
+            // Mark the contract as partial match (regardless of other issues).
+            info.verification_problems = vec![VerificationProblem::IncorrectMetadata];
+        }
+        Ok(Some(info))
+    }
+
+    /// Checks if migration from `contracts_verification_info` to `contract_verification_info_v2` is performed
+    /// by checking if the latter has more or equal number of rows.
+    pub async fn is_verification_info_migration_performed(&mut self) -> DalResult<bool> {
+        let count_v1 = sqlx::query!(
```
slowli, line 706: Nit: Not likely an issue in practice, but it looks possible to use a single DB query here to make the query transactional (alternatively, open a transaction explicitly, but it looks like an overkill).

core/lib/dal/src/contract_verification_dal.rs
```diff
@@ -560,6 +583,285 @@ impl ContractVerificationDal<'_, '_> {
...
+        .instrument("is_verification_info_migration_performed#count_v2")
+        .fetch_one(self.storage)
+        .await?
+        .count
+        .unwrap() as usize;
+
+        Ok(count_v2 >= count_v1)
+    }
+
+    pub async fn perform_verification_info_migration(
```
slowli, line 737: Bikeshedding: Subjectively, this mixes DB queries with high-level logic. I'd prefer the latter to be moved to the contract verifier library.

popzxc: The logic here was that it's a one-time thing that should be removed after the migration is done everywhere, so it would be easier to remove it if it exists in a single place. Otherwise, we can get some dead leftovers. WDYT?

slowli: IMO, it would be manageable to isolate the temporary DB queries into a separate `impl` block with a `deprecated` attribute. This worked for some previous migrations.

core/lib/types/src/contract_verification/contract_identifier.rs
popzxc: I started refactoring it, and took a slightly different approach (that also doesn't require mutating the structure). Hope it's better now.

core/lib/dal/src/contract_verification_dal.rs
popzxc: I've moved some of the logic out of migration, but otherwise I think it's OK to keep as-is. Now it's _mostly_ SQL, and moreover everything is performed in a single long-running transaction.

core/bin/contract-verifier/src/main.rs
popzxc: Reworked the migration logic:
- Got rid of `OFFSET` in select query
- Got rid of `serde_json::Value` roundtrip
- Added rayon for parallel calculation of identifiers
- Replaced `INSERT` with `COPY FROM STDIN`

All of that combined reduced the migration time on mainnet data from ~19 minutes to ~3.5 minutes.

core/lib/dal/src/contract_verification_dal.rs
```diff
@@ -560,6 +587,285 @@ impl ContractVerificationDal<'_, '_> {
...
+        Ok(row.count_v2 >= row.count_v1)
+    }
+
+    pub async fn perform_verification_info_migration(
+        &mut self,
+        batch_size: usize,
+    ) -> anyhow::Result<()> {
+        // We use a long-running transaction, since the migration is one-time and during it
+        // no writes are expected to the tables, so locked rows are not a problem.
+        let mut transaction = self.storage.start_transaction().await?;
```
popzxc, line 732: I've decided to use a long-running transaction to make sure that we're fully idempotent (since we don't have IDs in the table, sorted addresses do not guarantee the correctness of resumed migration). Since no writes to either of tables are expected during migration, this should be OK.

core/lib/dal/migrations/20250122102800_contract-verifier-new-schema.up.sql
```diff
@@ -0,0 +1,13 @@
...
+    bytecode_without_metadata_keccak256 BYTEA,
+    verification_info JSONB NOT NULL,
+
+    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+    updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
+);
+
+-- Add hash indexes for hash columns
+CREATE INDEX IF NOT EXISTS contract_verification_info_v2_bytecode_keccak256_idx ON contract_verification_info_v2 USING HASH (bytecode_keccak256);
+CREATE INDEX IF NOT EXISTS contract_verification_info_v2_bytecode_without_metadata_keccak256_idx ON contract_verification_info_v2 USING HASH (bytecode_without_metadata_keccak256);
```
perekopskiy, line 13: nit: In practice hash index speeds `select` up a bit but slows down `inserts` significantly so we generally use btree in our codebase. Up to you as for this table insert performance doesn't matter much

core/lib/dal/src/contract_verification_dal.rs
```diff
@@ -560,6 +587,285 @@ impl ContractVerificationDal<'_, '_> {
...
+                let bytecode_without_metadata_keccak256 = row
+                    .bytecode_without_metadata_keccak256
+                    .map(|h| H256::from_slice(&h));
+                Ok((
+                    info,
+                    bytecode_keccak256,
+                    bytecode_without_metadata_keccak256,
+                ))
+            })
+            .instrument("get_contract_verification_info_v2#perfect_match")
```
perekopskiy, line 678: Nit: previous query is instrumented with the same tag and I think it's misleading as it's not always "perfect match". Let's rename first one to "#direct" and this one to "#similar" or smth else.

core/lib/dal/src/contract_verification_dal.rs
```diff
@@ -560,6 +587,285 @@ impl ContractVerificationDal<'_, '_> {
...
+
+        let identifier = ContractIdentifier::from_bytecode(bytecode_marker, deployed_bytecode);
+        // `unwrap_or_default` is safe, since we're checking that `bytecode_without_metadata_keccak256` is not null.
+        let bytecode_without_metadata_keccak256 = identifier
+            .bytecode_without_metadata_keccak256
+            .as_ref()
+            .map(|h| h.hash())
+            .unwrap_or_default();
+
+        let Some((mut info, bytecode_keccak256, bytecode_without_metadata_keccak256)) =
```
perekopskiy, line 645: If similar bytecode is found you can insert info in the DB for this contract. To be honest not sure it will be helpful, it would allow us to easily say how many contracts are verified but we don't really need it. If implemented naively with duplicating verification_info it also will bloat db size, on the other hand splitting the table into two (one with verification_info, bytecode hashes; and another one with verified contract address) will make schema more complex.
So I think the current approach is the right way for now but we can probably consider keeping track of all verified contract addresses explicitly in the future

perekopskiy (commented):
Approach with indexing hashes is nit!

core/lib/dal/src/contract_verification_dal.rs
popzxc: Could you elaborate a bit on how inserting the contract would be helpful?
It won't be accurate, since we will only be doing that if someone queries the contract (so e.g. if nobody checks the contract info, it won't be inserted).
We will save one query at some cases, but at cost of data duplication in the DB.

core/lib/dal/migrations/20250122102800_contract-verifier-new-schema.up.sql
popzxc: Huh, that's a really nice catch, I didn't know that. Will change

perekopskiy (dismissed):

core/lib/dal/src/contract_verification_dal.rs
popzxc: Yes; the migration is blocking and contract verifier is only run as a single instance, so in fact it's guaranteed that it won't happen.
I added a migration for the whole duration just in case something like this happens:
1. We deploy new contract verifier.
2. Something wrong happens during migration.
3. We rollback
4. New contracts are verified. 
5. We deploy new verifier again, and migration skips some rows because new entries were added _between_ runs.

core/bin/contract-verifier/src/main.rs
popzxc: I completely agree with `rayon` being completely unnecessary, but it's already a transitive dependency through `zksync_l1_contract_interface` (together with crates such as `circuit_definitions`, `fflonk`, `zksync_kzg`, and many others that also feel unnecessary), so we're not making anything worse here `¯\_(ツ)_/¯`

This being blocked is not an issue at all, because the migration is blocking and nothing will be executed in parallel until it's complete.

core/lib/types/src/contract_verification/api.rs
```diff
@@ -232,12 +232,28 @@ impl CompilationArtifacts {
...
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct VerificationInfo {
     pub request: VerificationRequest,
     pub artifacts: CompilationArtifacts,
     pub verified_at: DateTime<Utc>,
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub verification_problems: Vec<VerificationProblem>,
```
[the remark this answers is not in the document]
popzxc: This is indeed a nice approach in general, however, I don't think it's necessary this time, since the only thing that gets _potentially_ (and very unlikely with how we will be rolling this feature out) affected is whether we will show an info box stating that it's a partially verified contract.

The behavior we're introducing _should_ be a default, so changing how the method is invoked brings inconvenience with little benefit.

core/lib/dal/src/contract_verification_dal.rs
```diff
@@ -560,6 +587,285 @@ impl ContractVerificationDal<'_, '_> {
...
+
+            copy.send(buffer.as_bytes()).await?;
+
+            offset += batch_size;
+            cursor = addresses.last().unwrap().clone();
+        }
+
+        // Sanity check.
+        tracing::info!("All the rows are migrated, verifying the migration");
+        let count_inequal = sqlx::query!(
```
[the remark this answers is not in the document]
popzxc: [English doesn't cease to amaze me](https://www.merriam-webster.com/grammar/unequal-vs-inequality-negative-prefix-usage)

core/lib/dal/src/contract_verification_dal.rs
```diff
@@ -560,6 +587,285 @@ impl ContractVerificationDal<'_, '_> {
...
+            r#"
+            SELECT
+                COUNT(*)
+            FROM
+                contract_verification_info_v2 v2
+            JOIN contracts_verification_info v1 ON initial_contract_addr = address
+            WHERE v1.verification_info::text != v2.verification_info::text
```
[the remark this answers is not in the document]
popzxc: I have run the migration locally with data dumped with mainnet, and this check passed successfully.
However, I have no theoretical guarantee of correctness -- could you clarify?

core/lib/dal/src/contract_verification_dal.rs
```diff
@@ -560,6 +587,285 @@ impl ContractVerificationDal<'_, '_> {
...
+            for idx in 0..addresses.len() {
+                let address = hex::encode(&addresses[idx]);
+                let bytecode_keccak256 = hex::encode(ids[idx].bytecode_keccak256);
+                let bytecode_without_metadata_keccak256 = ids[idx]
+                    .bytecode_without_metadata_keccak256
+                    .map(|h| format!(r#"\\x{}"#, hex::encode(h.hash())))
+                    .unwrap_or_else(|| "null".to_owned());
+                let verification_info = verification_infos[idx].replace('"', r#""""#);
+
+                let row = format!(
```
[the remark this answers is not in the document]
popzxc: On 1 -- it doesn't really affect the performance (I've checked).
On 2 -- I find `hex::encode` to be much more readable, so will err to my preference 😅 
On 3 -- it would!

core/lib/dal/src/contract_verification_dal.rs
```diff
@@ -560,6 +587,285 @@ impl ContractVerificationDal<'_, '_> {
...
+
+        let identifier = ContractIdentifier::from_bytecode(bytecode_marker, deployed_bytecode);
+        // `unwrap_or_default` is safe, since we're checking that `bytecode_without_metadata_keccak256` is not null.
+        let bytecode_without_metadata_keccak256 = identifier
+            .bytecode_without_metadata_keccak256
+            .as_ref()
+            .map(|h| h.hash())
+            .unwrap_or_default();
+
+        let Some((mut info, bytecode_keccak256, bytecode_without_metadata_keccak256)) =
```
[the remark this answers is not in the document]
popzxc: The address is in fact a part of the request (which is a part of verification info), so this information is already there.

core/lib/dal/src/contract_verification_dal.rs
```diff
@@ -560,6 +587,285 @@ impl ContractVerificationDal<'_, '_> {
...
+        .instrument("get_contract_verification_info_v2#direct")
+        .with_arg("address", &address)
+        .fetch_optional(self.storage)
+        .await?
+        .flatten()
+        {
+            return Ok(Some(info));
+        }
+
+        // Try partial match
```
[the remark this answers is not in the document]
popzxc: This is fair -- will change.

core/lib/dal/src/contract_verification_dal.rs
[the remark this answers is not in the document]
slowli: I think this sort of check is fine for now.

core/lib/types/src/contract_verification/contract_identifier.rs
```diff
@@ -0,0 +1,374 @@
...
+    Full,
+    /// Metadata is different.
+    Partial,
+    /// No match.
+    None,
+}
+
+/// Metadata detected in the contract bytecode.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum DetectedMetadata {
```
slowli, line 43: This structuring somewhat irks me. `DetectedMetadata` is used in `ContractIdentifier.bytecode_without_metadata_keccak256`, which is not about metadata (or its hash), but rather about the hash of the *remaining* part of the bytecode. Maybe, it'd make sense to define this type as a simple C-style enum, and restructure `ContractIdentifier`?

```rust
// or `DetectedMetadataKind`
pub enum DetectedMetadata {
    Keccak256,
    Cbor,
}

pub struct ContractIdentifier {
    // ...
    pub metadata: Option<DetectedMetadata>,
    // equal to `bytecode_keccak256` if there's no metadata
    pub bytecode_without_metadata_keccak256: H256,
}
```

Does such a redefinition of `bytecode_without_metadata_keccak256` make sense? Subjectively, it would be somewhat simpler to reason about (if there's no metadata, "bytecode w/o metadata" is just the entire bytecode).

core/node/contract_verification_server/src/api_impl.rs
```diff
@@ -220,15 +224,80 @@ impl RestApi {
...
+        return Ok(None);
+    };
+
+    if identifier.bytecode_keccak256 != fetched_keccak256 {
+        // Sanity check
+        let has_hash_without_metadata = fetched_keccak256_without_metadata.is_some();
+        let hashes_without_metadata_match = identifier
+            .bytecode_without_metadata_keccak256
+            .map(|h| h.hash())
+            == bytecode_without_metadata_keccak256;
```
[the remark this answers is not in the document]
popzxc: That's a nice catch!

slowli (approved):

PR #3533, opened 2025-01-27
fix: block.timestamp is not accurate

thomas-nguy:
## What ❔

Related to https://github.com/zkSync-Community-Hub/zksync-developers/discussions/820

Change the l2 block creation logic to start a new l2 block only when a transaction is ready to be executed.

## Why ❔

Current logic start a new l2 block as soon as the previous one is sealed.

A contract that relies on `block.timestamp` would be able to predict the time correctly because if the l2 block goes stale (no transaction), then it will be open indefinitely and the timestamp will not be accurate anymore

Solution has been tested locally but any feedbacks would be appreciated

## Checklist

### DIFF
--- a/core/node/node_sync/src/tests.rs
+++ b/core/node/node_sync/src/tests.rs
@@ -222,7 +222,7 @@ async fn external_io_basics(snapshot_recovery: bool) {
     let tx = create_l2_transaction(10, 100);
     let tx_hash = tx.hash();
     let tx = FetchedTransaction::new(tx.into());
-    let actions = vec![open_l1_batch, tx.into(), SyncAction::SealL2Block];
+    let actions = vec![open_l1_batch, tx.into(), SyncAction::SealBatch];
 
     let (actions_sender, action_queue) = ActionQueue::new();
     let client = MockMainNodeClient::default();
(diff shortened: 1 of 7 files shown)


core/node/node_sync/src/tests.rs
```diff
@@ -222,7 +222,7 @@ async fn external_io_basics(snapshot_recovery: bool) {
     let tx = create_l2_transaction(10, 100);
     let tx_hash = tx.hash();
     let tx = FetchedTransaction::new(tx.into());
-    let actions = vec![open_l1_batch, tx.into(), SyncAction::SealL2Block];
+    let actions = vec![open_l1_batch, tx.into(), SyncAction::SealBatch];
```
thomas-nguy, line 225: With the new changes, action that precede an `OpenBatch` has to be `SealBatch`

core/node/state_keeper/src/keeper.rs
```diff
@@ -582,47 +651,70 @@ impl ZkSyncStateKeeper {
                     "L1 batch #{} should be sealed unconditionally as per sealing rules",
                     updates_manager.l1_batch.number
                 );
+
+                // Push the current block if it has not been done yet
+                if is_last_block_sealed {
+                    let new_l2_block_params = self
+                        .wait_for_closing_l2_block_params_when_closing_batch(
```
thomas-nguy, line 658: Trick to make pass the unit tests.

In case of [OpenBlock, SealBatch] , if we replace this line with `wait_for_new_l2_block_params`, unit tests will fail because `wait_for_new_l2_block_params` is called twice

PR #3543, opened 2025-01-28
refactor(zk_toolbox): Minimize zkstack dependency on `zksync_config` (pt. 2)

slowli:
## What ❔

- Completely removes `zksync_*config` deps in zkstack.
- Introduces strongly-typed wrappers for configs in zkstack.

## Why ❔

Part of preparations for integrating a new config system.

## Checklist


core/lib/zksync_core_leftovers/src/temp_config_store/mod.rs
```diff
@@ -5,8 +5,8 @@ use zksync_config::{
     configs::{
         api::{HealthCheckConfig, MerkleTreeApiConfig, Web3JsonRpcConfig},
         chain::{
-            CircuitBreakerConfig, MempoolConfig, NetworkConfig, OperationsManagerConfig,
-            StateKeeperConfig, TimestampAsserterConfig,
+            CircuitBreakerConfig, MempoolConfig, OperationsManagerConfig, StateKeeperConfig,
```
popzxc, line 8: BTW -- do we need `zksync_core_leftovers` at all now? IIRC we only needed it because `zkstack_cli` used to depend on `TempConfigStore`, and the rest of the stuff is easily removable.

popzxc (approved):
Nice. As usual, suggest getting extra review from @sanekmelnikov

core/lib/zksync_core_leftovers/src/temp_config_store/mod.rs
slowli: `TempConfigStore` [is also used](https://github.com/matter-labs/zksync-era/blob/a075b229f054aa21415ef8462b8ad07e4985e678/core/bin/zksync_server/src/main.rs#L182) in the `zksync_server` binary, and [`load_general_config` / `load_database_secrets`](https://github.com/matter-labs/zksync-era/blob/a075b229f054aa21415ef8462b8ad07e4985e678/core/lib/zksync_core_leftovers/src/temp_config_store/mod.rs#L212-L233) are used by most other binaries (e.g., [contract verifier](https://github.com/matter-labs/zksync-era/blob/a075b229f054aa21415ef8462b8ad07e4985e678/core/bin/contract-verifier/src/main.rs#L32)). Naturally, all of these would be removed after the new config system integration, so I'd suggest to wait for it.

core/lib/protobuf_config/src/proto/config/api.proto
```diff
@@ -11,9 +11,7 @@ message MaxResponseSizeOverride {
 
 message Web3JsonRpc {
   optional uint32 http_port = 1; // required; u16
-  optional string http_url = 2; // required
   optional uint32 ws_port = 3; // required; u16
-  optional string ws_url = 4; // required
```
Deniallugo, line 16: That's a bit of a controversial change; on the one hand, it's not required for running the server. But it does help for running other tools against the server

core/node/node_sync/src/client.rs
```diff
@@ -105,7 +105,25 @@ impl MainNodeClient for Box<DynClient<L2>> {
     }
 
     async fn fetch_genesis_config(&self) -> EnrichedClientResult<GenesisConfig> {
-        self.genesis_config().rpc_context("genesis_config").await
+        let dto = self.genesis_config().rpc_context("genesis_config").await?;
+        Ok(GenesisConfig {
```
Deniallugo, line 109: Why do we need both DTO and non dto?

slowli: Single responsibility principle; it's *weird* to have a single type represent both a configuration and something used by API. More practically, it makes `zksync_web3_decl` depend on `zksync_config`, which is unnecessary and makes zkstack transitively depend on `zksync_config` as well (since it depends on `zksync_web3_decl`), even if the direct dependency is removed.

Deniallugo (approved):

Deniallugo (dismissed):

core/lib/config/src/configs/contracts/gateway.rs
```diff
@@ -1,27 +1,4 @@
...
-    pub default_upgrade_addr: Address,
-    pub multicall3_addr: Address,
-    pub relayed_sl_da_validator: Address,
-    pub validium_da_validator: Address,
-    pub diamond_cut_data: Bytes,
-}
+use zksync_basic_types::{Address, SLChainId};
 
 #[derive(serde::Serialize, serde::Deserialize, Debug, Clone, PartialEq)]
 pub struct GatewayChainConfig {
```
Deniallugo, line 4: this struct needs only for zkstack

Deniallugo (approved):

Deniallugo (approved):

artmakh (approved):

PR #3556, opened 2025-01-30
fix: Support newer versions of foundry-zksync

popzxc:
`zksync_contracts` is some spilled spaghetti.


core/lib/contract_verifier/src/contract_identifier.rs
slowli, line 1: Artifact?

Deniallugo (dismissed):

itegulov (approved):

PR #3559, opened 2025-01-31
perf(db): Remove `events.tx_initiator_address` writes and index

slowli:
## What ❔

Removes writes to the `events.tx_initiator_address` field in Postgres and the corresponding index.

## Why ❔

This data isn't used in DB queries, so it bloats the Postgres storage. It has an associated index, so it may negatively influence insertion latency as well.

## Checklist


core/lib/dal/migrations/20250131090112_remove_tx_initiator_address_from_events.up.sql
```diff
@@ -0,0 +1,3 @@
+-- `events.tx_initiator_address` was only used in Block Explorer queries that were removed from the codebase.
+DROP INDEX IF EXISTS events_tx_initiator_address_idx;
```
slowli, line 2: There [was an attempt](https://github.com/matter-labs/zksync-era/blob/ddc953d138456b6cbe2a9bb2f4518bc7800ba0ce/core/lib/dal/migrations/20240306155535_remove_obsolete_indexes.up.sql#L2) to remove this index in an earlier migration, but its name was misspelled unfortunately.

popzxc (approved):

Deniallugo (dismissed):

Deniallugo (changes requested):
we have to split it to two prs. 
One with changing the behaviour and next one with changing the schema, that will actually be merged only after the next protocol upgrade

perekopskiy (commented):
Index on `tx_initiator_address` can be dropped right away, otherwise looks good

perekopskiy (approved):

Deniallugo (approved):

PR #3564, opened 2025-02-03
refactor(api): Refactor VM-invoking API server methods to be less VM-specific

slowli:
## What ❔

- Refactors VM-invoking methods in the API server (e.g., gas estimation) to use less VM-specific logic.
- Simplifies some VM interfaces along the way. E.g., unifies batch and oneshot VM executor results.

## Why ❔

Should somewhat simplify ZK OS integration into the API server.

## Checklist


core/node/api_server/src/execution_sandbox/vm_metrics.rs
```diff
@@ -156,24 +152,25 @@ pub(super) fn collect_tx_execution_metrics(
...
+        writes,
+        vm: VmExecutionMetrics {
+            gas_used: result.statistics.gas_used as usize,
+            published_bytecode_bytes,
+            l2_l1_long_messages,
+            l2_to_l1_logs: result.logs.total_l2_to_l1_logs_count(),
+            user_l2_to_l1_logs: result.logs.user_l2_to_l1_logs.len(),
+            contracts_used: result.statistics.contracts_used,
+            // FIXME: incorrectly defined? (a number of factory deps, not number of contracts deployed)
+            contracts_deployed,
```
slowli, line 164: Subj. I'm not 100% sure this is a mistake, but if we indeed want to have the number of factory deps as a stat, (a) the name is misleading; (b) the stats can be extracted from transaction data, it's not really *VM execution* metrics.

popzxc: IDK, I'd check if this data is used anywhere at all. If not, probably it's just in case we want to look at stats -- and in this case we can remove it.

popzxc (approved):

core/node/api_server/src/execution_sandbox/vm_metrics.rs
slowli: It's not used in the Era codebase, and AFAIU isn't used in the test node / zksync-foundry either. So is it safe to remove? This would allow to simplify metrics construction.

popzxc: It is

core/node/api_server/src/execution_sandbox/vm_metrics.rs
```diff
@@ -132,16 +134,11 @@ pub(super) struct ExecutionMetrics {
...
-    let event_topics = result
-        .logs
-        .events
-        .iter()
-        .map(|event| event.indexed_topics.len() as u16)
-        .sum();
+    let writes = StorageWritesDeduplicator::apply_on_empty_state(&result.logs.storage_logs);
+
+    // FIXME: these computations differ from ones in `VmExecutionResultAndLogs::get_execution_metrics()`; why?
     let l2_l1_long_messages = VmEvent::extract_long_l2_to_l1_messages(&result.logs.events)
```
perekopskiy, line 142: `VmExecutionResultAndLogs::get_execution_metrics` seems to be more precise. It shouldn't matter much since it's only used for debug/metrics

slowli: I understand that it doesn't matter much, but the divergence itself looks weird. If it's possible to replace code here with `VmExecutionResultAndLogs::get_execution_metrics()`, will gladly do it since it'll make code simpler.

perekopskiy (approved):

PR #3567, opened 2025-02-04
fix(vm): Fix VM divergences related to validation

slowli:
## What ❔

Fixes divergences in VM execution during account validation, or reconsiders what is considered a divergence.

## Why ❔

Fixes some of divergences currently observed during oneshot VM execution in the API server in CI (specifically, during the "custom account" integration test).

## Checklist


slowli (commented):
- Locally, I've observed little (~2–3%) degradation in some wall-clock VM benchmarks w.r.t. the "default" / fast validation tracer changes; on most benchmarks, there's no observable degradation. Looks expected given that the tracer only tracks `Ret(Panic)` operations.
- The PR changes fast VM semantics with "default" validation tracing; it now unconditionally fails tx execution if an out-of-gas error is encountered during validation. This is somewhat close to the legacy VM and fixes VM divergences on integration unit tests (e.g. [this one](https://github.com/matter-labs/zksync-era/blob/a85d1f91e914a633fe7eca6946696f9a5f8b539c/core/tests/ts-integration/tests/custom-account.test.ts#L354)). However, it's not a 1-to-1 match because the legacy VM deals with computational gas calculations, meaning that it doesn't flag *some* out-of-gas errors during validation, like with deep recursion.

joonazan (approved):

joonazan (approved):

PR #3568, opened 2025-02-05
refactor: make zksync_utils thinner

itegulov:
## What ❔

This PR removes heavy transitive dependencies from `zksync_utils` by:
* Moving `ManagedTasks` into a separate crate (ideally needs to be removed in favor of node_framework but unfortunately still used by prover and contract-verifier)
* Deleting custom-written retry logic for reqwest. The only place that was using was `RegionFetcher` from prover utils. I migrated it to `reqwest-retry` instead (Cargo.lock diff looks bigger than it actually is - almost every new crate is just some wasm-specific stuff that doesn't matter).

## Why ❔

Depending on `zksync_contracts`, `zksync_multivm` and other seemingly core crates pulls in a lot of transitive dependencies right now (`reqwest`, `sentry`, `opentelemetry`, `openssl` etc)

## Checklist


zkstack_cli/crates/zkstack/Cargo.toml
```diff
@@ -48,7 +48,7 @@ zksync_system_constants.workspace = true
 zksync_eth_client.workspace = true
 zksync_contracts.workspace = true
 prost.workspace = true
-reqwest = "0.12.8"
+reqwest = { version = "0.12.8", features = ["blocking"] }
```
Deniallugo, line 51: why not workspace?

itegulov: the version wasn't coming from the workspace before my changes so I missed it, just pushed a commit that ensures there are no non-workspace versions anymore

popzxc (approved):

PR #3574, opened 2025-02-06
feat: make `zksync_types` thinner

itegulov:
## What ❔

This PR removes a few (effectively) unused dependencies from `zksync_types` and hides a few more behind features flags:
* `protobuf` enables `prost`, `zksync_protobuf`, `zksync_protobuf_build` (only needed because of `ProtoFmt` impls for couple of snapshot types)
* `contract-verification` enables `ciborium`

Alternatively to features I am open to just moving those types to a separate crate as I would consider them to be "internal". Maybe something like `zksync_internal_types`.

One open question: do we need `bigdecimal`? Its only usage is [this](https://github.com/matter-labs/zksync-era/blob/main/core/lib/types/src/fee_model.rs#L289-L294) but the calculation doesn't seem to require bigdecimal at all. I would get rid of it too if there are no objections.

## Why ❔

zksync_types is widely used outside of zksync-era and should ideally be a lot thinner than it is right now

## Checklist


popzxc:
> Alternatively to features I am open to just moving those types to a separate crate as I would consider them to be "internal". Maybe something like zksync_internal_types.

I don't think it will solve the problem, we already have "basic types" and "types" 😓

popzxc:
> One open question: do we need bigdecimal? Its only usage is [this](https://github.com/matter-labs/zksync-era/blob/main/core/lib/types/src/fee_model.rs#L289-L294) but the calculation doesn't seem to require bigdecimal at all. I would get rid of it too if there are no objections.

I guess it can be removed.

popzxc (approved):

PR #3578, opened 2025-02-07
feat(contract-verifier): Do not allow verification requests for verified contracts

popzxc:
## What ❔

- Disallow sending verification request if the contract is already (fully) verified.
- Disallow sending verification request if there is already an ongoing (queued/in progress) request for this address.
- Refactor tests for contract verification API, introduce helpers to make tests more declarative
- Add new checks to tests to check added functionality.

## Why ❔

Just for lulz


core/node/contract_verification_server/src/api_impl.rs
```diff
@@ -132,6 +141,27 @@ impl RestApi {
             .master_connection_pool
             .connection_tagged("api")
             .await?;
+
+        // Verification is only allowed if the contract is either wasn't verified yet
+        // or the verification is partial.
```
perekopskiy, line 146: There is an issue with force deployments. System contracts are forced deployed on protocol upgrade and need to be re-verified. We should either allow re-verification or invalidate successful requests on force deployments

popzxc: Oh, that's interesting. I wonder whether it's supported on any explorer aside from our own.
I'm thinking to allow re-verification for system contracts (address below `0x00..8000` or whatever address it is), or do you know a better heuristic?

popzxc: I've allowed the re-verification in case `is_system` flag was present in the previous verification request.
Do you think it's a good enough heuristic?

perekopskiy: I think it's reasonable to assume `is_system` setting is enabled for system contracts. The field you check is set only for `solidity-single-file` though, for `solidity-standard-json-input` it's hidden in json->"settings"

popzxc: Good catch, fixed.

perekopskiy: it's camelCase in JSON, i.e. field is "isSystem". And also it's called "enableEraVMExtensions" in newer versions of compiler

perekopskiy (approved):

PR #3592, opened 2025-02-11
feat(contract-verifier): Support missing options for EVM in API

popzxc:
## What ❔

Adds a few EVM-specific options to add EVM contract verification to the block explorer front-end (based on Etherscan UI).

## Why ❔

To enable contract verification in the block explorer UI.

## Is this a breaking change?
- [ ] Yes
- [ ] No
- [x] DB migration required

## Operational changes

None


core/lib/types/src/contract_verification/api.rs
```diff
@@ -147,6 +147,19 @@ pub struct VerificationIncomingRequest {
...
     pub force_evmla: bool,
+    #[serde(flatten)]
+    pub evm_specific: VerificationEvmSettings,
+}
+
+/// Settings for EVM verification, used only if
+/// `SourceCodeData` is `SolSingleFile`.
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct VerificationEvmSettings {
```
popzxc, line 158: The verification request is pretty messy tbh, I think if we will have enough resources, we need to redesign the API and rely on the `standard-json` input only.

core/lib/types/src/contract_verification/api.rs
```diff
@@ -147,6 +147,19 @@ pub struct VerificationIncomingRequest {
...
+    #[serde(default)]
+    pub evm_version: Option<String>,
```
slowli, line 160: Isn't EVM version [supported by `zksolc`](https://github.com/matter-labs/era-compiler-solidity/blob/85bb573aadcf3f0772d6f898ceaa7b9487f12e0c/docs/src/03-standard-json.md#input-json)? Or by `VerificationEvmSettings` you don't necessarily mean settings related to EVM bytecode toolchains?

core/lib/types/src/contract_verification/api.rs
```diff
@@ -147,6 +147,19 @@ pub struct VerificationIncomingRequest {
...
+    #[serde(default)]
+    pub optimizer_runs: Option<u16>,
```
slowli, line 162: I wonder whether it would be possible to put this data in the existing `optimizer_mode` column (it's already toolchain-specific, after all). Looks somewhat like a hack, so probably not worth it.

core/lib/types/src/contract_verification/api.rs
```diff
@@ -147,6 +147,19 @@ pub struct VerificationIncomingRequest {
...
+    #[serde(flatten)]
+    pub evm_specific: VerificationEvmSettings,
+}
+
+/// Settings for EVM verification, used only if
+/// `SourceCodeData` is `SolSingleFile`.
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct VerificationEvmSettings {
+    #[serde(default)]
```
slowli, line 159: Don't know whether it matters much, but since this is serialized back to JSON to store in Postgres, it could make sense to have `skip_serializing_if = "Option::is_none"` (or even put a `null` altogether if neither of fields is specified) in order to not bloat the table size.

core/lib/contract_verifier/src/tests/real.rs
```diff
@@ -224,6 +224,53 @@ async fn using_standalone_solc(specify_contract_file: bool) {
     assert_eq!(output.abi, counter_contract_abi());
 }
 
+#[test_casing(3, [(Some(100), None), (None, Some("shanghai")), (Some(200), Some("paris"))])]
+#[tokio::test]
+async fn using_standalone_solc_with_custom_settings(
+    (runs, evm_version): (Option<u16>, Option<&str>),
```
slowli, line 230: Nit: `test_casing` supports multiple args, so it's possible to define `runs` and `evm_version` as separate args here.

slowli (commented):
Is it expected that the new request fields are ignored with the standard JSON input? Looks somewhat messy (neither ignoring them nor attempting to merge them in some way looks obvious), but as you've said, probably the best option would be to rely on standard JSON input only.

core/lib/types/src/contract_verification/api.rs
popzxc: It is supported, but AFAIU it's rather meaningless, since we're not compiling to any particular bytecode. Not sure if it will make sense to try using it outside of the EVM context (especially given that it's not used in the block explorer FE).

core/lib/types/src/contract_verification/api.rs
popzxc: I like `skip_serializing_if` more, it sounds both more flexible and doesn't add custom handling to DAL.

popzxc:
> Is it expected that the new request fields are ignored with the standard JSON input? Looks somewhat messy (neither ignoring them nor attempting to merge them in some way looks obvious), but as you've said, probably the best option would be to rely on standard JSON input only.

@slowli my understanding is that:
- sol-single-file is used by the block explorer front end, e.g. you pass _everything_ there
- standard-json is used by the tooling (e.g. hardhat/foundry) where we pass the same info that we pass to the compiler.

From what I've seen, we don't mix these options between each other, and frankly I'd be afraid to do it.

core/lib/contract_verifier/src/tests/real.rs
popzxc: Nice!

slowli (approved):

core/lib/types/src/contract_verification/api.rs
0xVolosnikov: Can we mark it as resolved to merge this PR? :)

perekopskiy (approved):

PR #3593, opened 2025-02-11
fix: EVM emulation integration – upgrade workflow fixes

slowli:
## What ❔

Fixes the upgrade workflow with new L1 contracts.

## Why ❔

The existing workflow doesn't work because of changes in `ProposedUpgrade` (+ some stubs in the codebase).

## Checklist

### DIFF
--- a/core/lib/basic_types/src/protocol_version.rs
+++ b/core/lib/basic_types/src/protocol_version.rs
@@ -75,11 +75,11 @@ pub enum ProtocolVersionId {
 
 impl ProtocolVersionId {
     pub const fn latest() -> Self {
-        Self::Version26 // FIXME: change to `Version27`
+        Self::Version27
     }
 
     pub const fn next() -> Self {
-        Self::Version27
+        Self::Version28
     }
 
     pub fn try_from_packed_semver(packed_semver: U256) -> Result<Self, String> {
--- a/core/lib/basic_types/src/protocol_version.rs
+++ b/core/lib/basic_types/src/protocol_version.rs
@@ -75,7 +75,7 @@ pub enum ProtocolVersionId {
 
 impl ProtocolVersionId {
     pub const fn latest() -> Self {
-        Self::Version26
+        Self::Version26 // FIXME: change to `Version27`
     }
 
     pub const fn next() -> Self {
@@ -125,7 +125,10 @@ impl ProtocolVersionId {
             ProtocolVersionId::Version25 => VmVersion::Vm1_5_0IncreasedBootloaderMemory,
             ProtocolVersionId::Version26 => VmVersion::VmGateway,
             ProtocolVersionId::Version27 => VmVersion::VmGateway,
-            ProtocolVersionId::Version28 => unreachable!("Version 28 is not yet supported"),
+
+            // Speculative VM version for the next protocol version to be used in the upgrade integration test etc.
+            // TODO: Must be changed when the protocol version is actually implemented!
+            ProtocolVersionId::Version28 => VmVersion::VmGateway,
         }
     }
 
@@ -304,7 +307,10 @@ impl From<ProtocolVersionId> for VmVersion {
             ProtocolVersionId::Version25 => VmVersion::Vm1_5_0IncreasedBootloaderMemory,
             ProtocolVersionId::Version26 => VmVersion::VmGateway,
             ProtocolVersionId::Version27 => VmVersion::VmGateway,
-            ProtocolVersionId::Version28 => unreachable!("Version 28 is not yet supported"),
+
+            // Speculative VM version for the next protocol version to be used in the upgrade integration test etc.
+            // TODO: Must be changed when the protocol version is actually implemented!
+            ProtocolVersionId::Version28 => VmVersion::VmGateway,
         }
     }
 }
--- a/core/lib/basic_types/src/protocol_version.rs
+++ b/core/lib/basic_types/src/protocol_version.rs
@@ -155,6 +155,10 @@ impl ProtocolVersionId {
         self < &Self::Version27
     }
 
+    pub fn is_post_fflonk(&self) -> bool {
+        self >= &Self::Version27
+    }
+
     pub fn is_1_4_0(&self) -> bool {
         self >= &ProtocolVersionId::Version18 && self < &ProtocolVersionId::Version20
     }
--- a/core/lib/types/src/abi.rs
+++ b/core/lib/types/src/abi.rs
@@ -1,5 +1,5 @@
 use anyhow::Context as _;
-use zksync_basic_types::protocol_version::ProtocolSemanticVersion;
+use zksync_basic_types::protocol_version::{ProtocolSemanticVersion, ProtocolVersionId};
 
 use crate::{
     bytecode::BytecodeHash,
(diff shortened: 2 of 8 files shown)


core/lib/basic_types/src/protocol_version.rs
```diff
@@ -125,7 +125,10 @@ impl ProtocolVersionId {
             ProtocolVersionId::Version25 => VmVersion::Vm1_5_0IncreasedBootloaderMemory,
             ProtocolVersionId::Version26 => VmVersion::VmGateway,
             ProtocolVersionId::Version27 => VmVersion::VmGateway,
-            ProtocolVersionId::Version28 => unreachable!("Version 28 is not yet supported"),
+
+            // Speculative VM version for the next protocol version to be used in the upgrade integration test etc.
+            // TODO: Must be changed when the protocol version is actually implemented!
+            ProtocolVersionId::Version28 => VmVersion::VmGateway,
```
slowli, line 131: Is this the correct way to do things? I don't think the upgrade integration test would work otherwise, since it's supposed to run with the next protocol version after the upgrade.

core/lib/types/src/abi.rs
```diff
@@ -315,43 +366,59 @@ impl ProposedUpgrade {
...
-    /// Decodes `ProposedUpgrade` from a RLP token.
-    /// Returns an error if token doesn't match the `schema()`.
-    pub fn decode(token: Token) -> anyhow::Result<Self> {
+    pub(crate) fn decode(data: &[u8]) -> anyhow::Result<Self> {
+        let mut upgrade = None;
+        for (schema, token) in ProposedUpgradeSchema::decode_to_token(data) {
+            match Self::decode_from_token(schema, token) {
+                Ok(decoded) => {
+                    if let Some(upgrade) = &upgrade {
+                        anyhow::bail!("Ambiguous upgrade: can be decoded as either {decoded:?} or {upgrade:?}");
```
slowli, line 375: I don't like this approach since there's no guarantee that this error won't trigger. In practice, the new upgrade payloads decode successfully using either the post-gateway or the post-EVM schema, and the only thing stopping the post-gateway decoding from being correct is a bogus protocol version (0). Ideally, I'd imagine that the protocol version (or maybe the schema identifier itself?) would be encoded at the known offset of the payload.

perekopskiy: This is indeed unfortunate. Most likely changing encoding schema in L1 contracts is not possible at this point without re-audit. As a more stable workaround you can pass `DecentralizedUpgradesEventProcessor::last_seen_protocol_version + 1` down to `decode` method as a hint to choose the schema

core/lib/basic_types/src/protocol_version.rs
```diff
@@ -75,11 +75,11 @@ pub enum ProtocolVersionId {
...
 impl ProtocolVersionId {
     pub const fn latest() -> Self {
-        Self::Version26
+        Self::Version27
     }
 
     pub const fn next() -> Self {
-        Self::Version27
+        Self::Version28
     }
```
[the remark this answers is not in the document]
slowli: Does this only concern `ProtocolVersionId::{latest, next}`? Or do you mean changing the protocol version in `genesis.yaml` so that it's still v26, and removing support of `Version28` in `VmVersion` mapping etc.?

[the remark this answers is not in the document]
perekopskiy: `genesis.yaml` wasn't changed and I mean keeping it unchanged. I don't mind keeping V28 if it's helpful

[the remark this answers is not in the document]
0xVolosnikov: `genesis.yaml` was changed in the [base branch](https://github.com/matter-labs/zksync-era/tree/evm-emulator/integration). So it is supposed to be final V27 support

core/lib/basic_types/src/protocol_version.rs
0xVolosnikov: Shouldn't we use the new subversion (smth like VmEvmEmulator) here?

core/lib/basic_types/src/protocol_version.rs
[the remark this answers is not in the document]
perekopskiy: ah, ok, then let's keep `latest`,`next` updated

core/lib/types/src/abi.rs
slowli: It could prevent some ambiguous decodings, but not all of them, IIUC: the protocol version [may stay the same with an upgrade](https://github.com/matter-labs/zksync-era/blob/298abd2f988229a7fe5ee77de1015940d1d9ad41/core/node/eth_watch/src/event_processors/decentralized_upgrades.rs#L159), which means that an ambiguous decoding I've described above would still occur if `last_seen_protocol_version.minor == ProtocolVersionId::Version26` (since the upgrade may be either a patch bump or a v27 upgrade). Also, just for my understanding: Is it possible for an upgrade to skip a `ProtocolVersionId` (e.g., a v25 -> v27 upgrade)?

core/lib/basic_types/src/protocol_version.rs
slowli: Once we determine how to deal with the new VM version in `multivm`, I'd imagine that yes we'd have `VmVersion:: VmEvmEmulator` for protocol versions v27 and v28.

core/lib/types/src/abi.rs
perekopskiy: Agree, it's not perfect either but no upgrade between v26 and v27 is planned. It's possible in theory, on practice it should only happen if we upgraded some devnet, found an issue, and want to keep the devnet live, then devnet will be upgraded this way x->x+1->x+2, and others x->x+2

slowli: I've tried this approach with passing the min expected version, but it doesn't work straightforwardly because IIUC the watcher component may receive old upgrade events; [here's the corresponding filter](https://github.com/matter-labs/zksync-era/blob/9a5147a304cc9a5dfb705d0f17b14808b1bde052/core/node/eth_watch/src/event_processors/decentralized_upgrades.rs#L117). We most probably want to distinguish between valid old upgrades that can be skipped and upgrades that the node doesn't understand, which makes your proposal not quite straightforward. So if you don't mind, I'd leave the logic as-is.

0xVolosnikov: > no upgrade between v26 and v27 is planned
We had some discussions about v26.1

0xVolosnikov (approved):

core/tests/upgrade-test/tests/upgrade.test.ts
```diff
@@ -265,12 +266,18 @@ describe('Upgrade test', function () {
             'contracts/system-contracts/artifacts-zk/contracts-preprocessed/DefaultAccount.sol/DefaultAccount.json'
         );
 
+        const evmEmulatorCode = readCode(
+            'contracts/system-contracts/zkout/EvmEmulator.yul/contracts-preprocessed/EvmEmulator.yul.json'
```
0xVolosnikov, line 270: Output path from compiler was changed several times recently, so we have some ugly functions to read bytecode. Iiuc, it isn't guaranteed that this hardcoded path will remain valid in the future (but I hope so)

PR #3619, opened 2025-02-18
fix: block.timestamp is not accurate for the first transaction

thomas-nguy:
## What ❔

Create a batch only when a transaction can be found in mempool
This solve the issue of having an outdated `block.timestamp` when executing the transaction

**Context :** 
The first transaction in a batch uses the `batch.timestamp` that is set at the creation of the batch and it can easily be outdated in case the network has low activity and the first transaction in the batch come much later after the batch creation.

The following PR https://github.com/matter-labs/zksync-era/pull/3398 proposes to update the timestamp for l2 block but it does not apply for the first transaction which relies on the l1 batch parameter.

With this PR, the l1 batch will only start when the first transaction can be found in the mempool (with a priority to the upgrade transaction).

## Why ❔

The first transaction in a batch uses the batch timestamp for `block.timestamp` during the execution. This value can be outdated in case the first transaction does not come immediately after the batch creation.

This PR solves this issue

## Is this a breaking change?
- [] Yes
- [x] No

## Operational changes
No

## Checklist


core/node/state_keeper/src/keeper.rs
```diff
@@ -55,6 +54,14 @@ impl Error {
     }
 }
 
+#[derive(Debug)]
+struct ZkSyncStateKeeperBatchEnv {
```
thomas-nguy, line 58: Need this structure to avoid linter complain

perekopskiy (commented):
What does exactly this PR solve? Is it TOCTOU for the first batch transaction?
If so I think it's mostly useless as the situation "tx is dropped from mempool in between batch_params are created and first transaction is requested" is super rare, and there is another situation that leads to inaccurate block.timestamp: "first tx of a block is rejected" and it's not patched.
If there is a strong reason to improve it, then I would prefer a solution that covers all cases: timestamp is updated in the loop until the very first tx was executed and wasn't rejected. However, this would complicate the code significantly we'll have to rollback `start_next_l2_block`.
But tbh I wouldn't patch it as the impact is really low IMO

thomas-nguy:
Hi @perekopskiy 

Sorry for the lack of context,

In continuity to this https://github.com/matter-labs/zksync-era/pull/3398

The first transaction in a batch uses the `batch.timestamp` and it can easily be outdated in case the network has low activity  so it is not really a rare occurrence situation.

The following PR https://github.com/matter-labs/zksync-era/pull/3398 proposes to update the timestamp for l2 block but it does not apply for the first transaction which relies on the l1 batch parameter.

With this PR, the l1 batch will only start when the first transaction can be found in the mempool (with a priority to the upgrade transaction).

Beside fixing the first tx timestamp, I think it also clean up some logic by refactoring the upgrade tx execution in a batch logic


`there is another situation that leads to inaccurate block.timestamp: "first tx of a block is rejected" and it's not patched.`

Yes this is correct. I also found this corner case and it seems odds to me, in which situation the seal resolution return `SealResolution::Unexecutable`?  Is it related to buggy transactions? and what is the odds that it happens on chain?

thomas-nguy:
> This is already the case (if we disregard toctou which is almost impossible to happen) as self.mempool.has_next is checked in wait_for_new_batch_params, isn't it?


Okay you are correct. I have missread the code and I thought that new batch params could be retrieve even though no transaction in the mempool. 
In that case this PR is an overkill I agree.

>Rejecting transactions is a normal situation and it happens regularly. In most cases tx is rejected if its validation is failinge.g. initiator doesn't have enough eth to pay for gas_price*gas_limit

If its already in the execution stage, shouldnt the initiator pay at least for the intrinsic gas even though he doesnt have enough eth and have his transaction rejected instead of rollback?

perekopskiy:
if validation stage passes and execution starts then tx is not rejected (except some corner cases, not sure if any of them is realistic in current protocol version) and fee is charged

PR #3625, opened 2025-02-19
feat(zkos): Implement ZK OS Merkle tree

slowli:
## What ❔

Implements a persistent Merkle tree for ZK OS using RocksDB and amortized design similar to Jellyfish Merkle tree used for Era.

Some stuff in the PR is preliminary:

- Batch insertion proofs don't support read operations and are constructed slightly differently compared to the in-memory tree.
- There are no metrics (but there are logs with latencies / stats for key operations).
- No reader APIs.
- No pruning.

## Why ❔

ZK OS will need a persistent Merkle tree sooner or later. Amortized versioned design looks like a good choice (although not as good as for sparse Merkle trees, since internal nodes are almost guaranteed to be full in this case).

## Is this a breaking change?

- [ ] Yes
- [x] No

## Operational changes

None.

## Checklist


core/lib/zk_os_merkle_tree/src/hasher/proofs.rs
```diff
@@ -0,0 +1,393 @@
...
+    fn empty() -> Self {
+        Self {
+            operations: vec![],
+            sorted_leaves: BTreeMap::new(),
+            hashes: vec![],
+        }
+    }
+
+    /// Returns the new root hash of the tree on success.
+    pub fn verify(
```
slowli, line 77: AFAIU, the verification procedure implemented in the ZK OS repo uses separate Merkle paths for each of `sorted_leaves`, so the proof there is longer, but IIUC, the amount of hashing is the same (e.g., Merkle paths are not folded separately for each leaf; they are used almost in the same way as `hashes` here).

core/lib/zk_os_merkle_tree/README.md
```diff
@@ -0,0 +1,69 @@
...
+2025-02-19T11:06:24.936756Z DEBUG zk_os_merkle_tree: created batch proof, elapsed: 27.751333ms, proof.leaves.len: 23967, proof.hashes.len: 156210
+2025-02-19T11:06:24.944054Z DEBUG zk_os_merkle_tree: updated tree structure, elapsed: 7.285209ms
+2025-02-19T11:06:24.954820Z DEBUG zk_os_merkle_tree: hashed tree, elapsed: 10.747417ms
+2025-02-19T11:06:25.017817Z DEBUG zk_os_merkle_tree: persisted tree, elapsed: 62.967083ms
+2025-02-19T11:06:25.018655Z  INFO loadtest: Processed block #999 in 281.765541ms, root hash = 0x12fa11d7742d67509c9a980e0fb62a1b64a478c9ff4d7596555e1f0d5cb2043f
+2025-02-19T11:06:25.018669Z  INFO loadtest: Verifying tree consistency...
+2025-02-19T11:07:06.144174Z  INFO loadtest: Verified tree consistency in 41.126574667s
+```
+
+I.e., latency is dominated by I/O (~30% for key–index lookup, ~30% for loading tree nodes, and ~20% for tree
```
slowli, line 66: Intuitively, there should be a way to optimize key–index lookup, but I haven't come up with anything functional yet. Stuff like reusing RocksDB iterators doesn't seem to provide efficiency boost.

core/lib/zk_os_merkle_tree/src/hasher/nodes.rs
```diff
@@ -0,0 +1,183 @@
...
+use zksync_basic_types::H256;
+
+use crate::{
+    max_nibbles_for_internal_node,
+    types::{InternalNode, Root},
+    HashTree, TreeParams,
+};
+
+/// Internal hashes for a single `InternalNode`. Ordered by ascending depth `1..internal_node_depth`
+/// where `depth == 1` is just above child refs. I.e., the last entry contains 2 hashes (unless the internal node is incomplete),
```
perekopskiy, line 14: what do you mean by incomplete?

core/lib/zk_os_merkle_tree/src/hasher/nodes.rs
```diff
@@ -0,0 +1,183 @@
...
+        Self {
+            nodes,
+            internal_hashes,
+            internal_node_depth: P::INTERNAL_NODE_DEPTH,
+            level_offsets,
+        }
+    }
+
+    pub(crate) fn get(&self, depth_in_node: u8, index_on_level: u64) -> H256 {
+        let bit_shift = self.internal_node_depth - depth_in_node;
```
perekopskiy, line 126: Does it make sense to validate `depth_in_node` and `index_on_level`? i.e add asserts
`assert(depth_in_node < self.internal_node_depth)`
`assert(index_on_level < self.nodes.len() * (1 << bit_shift))`

core/lib/zk_os_merkle_tree/src/storage/patch.rs
```diff
@@ -0,0 +1,605 @@
...
+                next_index,
+            });
+        }
+
+        Ok(Self {
+            version: 0,
+            sorted_new_leaves,
+            updates: vec![],
+            inserts,
+            operations: vec![],
```
perekopskiy, line 99: why is it empty?

perekopskiy (commented):
I would find some diagrams explaining terms and types used very helpful for reviewing and understanding. Can be done out of the scope for this PR of course

core/lib/zk_os_merkle_tree/src/hasher/nodes.rs
slowli: It's possible, but I'm reasonably sure that both assertions hold by construction due to how `get()` is used in `collect_hashes()`. E.g., `depth_in_node` is defined as `depth % P::INTERNAL_NODE_DEPTH`, i.e., it's always valid.

core/lib/zk_os_merkle_tree/src/storage/patch.rs
slowli: Explained it in the added comment.

core/lib/zk_os_merkle_tree/src/hasher/nodes.rs
slowli: Containing less child refs than the maximum possible amount (e.g., <16 for radix-16 amortization). I've tried to explain amortization in the readme.

popzxc (approved):
Can't say I did a thorough review and fully understood the implementation, but I've read through all the code and each particular snippet made sense to me.

PR #3627, opened 2025-02-19
chore: make memory model version selectable

joonazan:


0xVolosnikov:
https://github.com/matter-labs/zksync-era/pull/3627/files#diff-f3cb1217e8d624491f151b2c1996272ffb8ebb71cdb2d121e61524c1063d1d83L76

zksync-crypto-gpu also has deps from zksync-protocol

joonazan:
@0xVolosnikov I guess I need to add a patch to Cargo.toml then because they are not direct dependencies.

The core test failures are in shadow mode, so vm2 is lacking something but not sure what since the new memory behaviour is not enabled.

0xVolosnikov:
I'll release new zksync-protocol and zksync-crypto-gpu deps

0xVolosnikov:
Imo would be better to rename `VmVersion::VmV27` to `VmVersion::VmEvmEmulator`, looking on previous versions

slowli (commented):
Other than version naming, looks fine. I'm somewhat strongly for version renaming though for consistency.

PR #3637, opened 2025-02-21
fix: Limit number of connections open for GCS interactions

EmilLuta:
## What?

This PR adds an in-house implementation for connections limiting, via semaphore. Ideally, the connection pool would expose a limit to configure (for instance in [python](https://requests.readthedocs.io/en/latest/api/#requests.adapters.HTTPAdapter)), but Sean [prefers it at application level](https://github.com/seanmonstar/reqwest/issues/386#issuecomment-446674871). Whilst this is not ideal, it's the preferred solution.

Longer term, we may contribute the fix (assuming Sean is willing to accept such changes).

## Why?

These changes were prompted by LWG failure (exact line of failure [here](https://github.com/matter-labs/zksync-era/blob/main/prover/crates/bin/witness_generator/src/utils.rs#L173)). There we download # of circuits per run. At some point in January, the machine started to fail. AFAIR, the local interface gets overwhelmed by the sheer amount of requests. Whilst there are multiple fixes possible (batch requests, add timeouts between requests, change the format to an entire folder to download, etc.), this fix will benefit the entire code base. Any time one part of our codebase will reach high number of requests per second, instead of erroring it will be gently throttled.

## Why 500 as throttle?

I've tested locally (GCP VM) with a big amount of requests and threads. GCP seems to be fine with ~3000, but can be affected by noisy neighbours. 500 is a conservative value, but in my testing barely adds any latency. Most requests in our codebase will never reach the 500. Those that do, might be prone to erroring and retring which in turn will end up taking more time (or failing).

## Why not add a config value for it?

Picking a good value means plenty of testing to understand what's going on. If anyone needs to change it (assuming google infra will change drastically), releasing an image will be a small amount of their work. I'd be much happier for this change to be a deliberate effort, rather than "oh, time improvements, maybe we can speed some things up by tweaking the value" or "oh, some timeouts, maybe we can move this number down and it'll be good".


yorik (approved):

core/lib/object_store/src/gcs.rs
```diff
@@ -123,6 +138,17 @@ fn has_transient_io_source(err: &(dyn StdError + 'static)) -> bool {
     get_source::<io::Error>(err).is_some()
 }
 
+/// This is necessary due to the fact that Acquiring a semaphore can fail if the semaphore is closed.
+/// Whilst current code does not expect such scenario, nor "should it be possible", it is better to have a graceful catch case instead of unwrapping.
+impl From<AcquireError> for ObjectStoreError {
+    fn from(err: AcquireError) -> Self {
+        ObjectStoreError::Other {
+            source: err.into(),
+            is_retriable: true,
```
slowli, line 147: I understand the reasoning behind not unwrapping, but why is this error retriable? If the semaphore is closed, it'll remain closed, so all subsequent requests to the store will fail.

core/lib/object_store/src/gcs.rs
```diff
@@ -159,6 +185,11 @@ impl From<HttpError> for ObjectStoreError {
 #[async_trait]
 impl ObjectStore for GoogleCloudStore {
     async fn get_raw(&self, bucket: Bucket, key: &str) -> Result<Vec<u8>, ObjectStoreError> {
+        let _permit = self
+            .semaphore
+            .acquire()
+            .await
+            .map_err(Into::<ObjectStoreError>::into)?;
```
slowli, line 192: Nit: `?` converts error types automatically, provided such a conversion exists.

core/lib/object_store/src/gcs.rs
```diff
@@ -17,13 +17,18 @@ use google_cloud_storage::{
     },
 };
 use http::StatusCode;
+use tokio::sync::{AcquireError, Semaphore};
 
 use crate::raw::{Bucket, ObjectStore, ObjectStoreError};
 
+const DEFAULT_MAX_CONCURRENT_REQUESTS: usize = 500;
```
perekopskiy, line 24: maybe put a shortened explaining from the PR description as a doc-comment here?

perekopskiy (approved):

perekopskiy (approved):

slowli (approved):

PR #3645, opened 2025-02-25
feat(gateway): dont allow v26 deposits without migration

StanislavBreadless:
## What ❔

Adds the ability to stop accepting all L1->L2 if there is at least one unsafe deposit for v26 migration.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


core/node/state_keeper/src/mempool_actor.rs
```diff
@@ -131,6 +348,14 @@ impl MempoolFetcher {
                 .await
                 .context("failed syncing mempool")?;
 
+            let (transactions_with_constraints, transactions_to_return) =
+                filter_unsafe_deposits(transactions_with_constraints, &mut storage).await?;
```
perekopskiy, line 352: The approach looks good! Just a couple of comments
- let's log tx hashes in `transactions_to_return`
- let's maybe add a config var `skip_filtering_unsafe_deposits` with default=false so it's possible to disable filtering in case it doesn't work as expected and filters some safe deposits

core/node/state_keeper/src/mempool_actor.rs
```diff
@@ -131,6 +348,14 @@ impl MempoolFetcher {
...
+            storage
+                .transactions_dal()
+                .return_to_mempool(&transactions_to_return)
```
slowli, line 355: Does transactionality with a `sync_mempool` above matter here? I.e., would anything bad happen if a node crashes after a `sync_mempool()`, but before this call?

core/lib/types/src/storage/mod.rs
```diff
@@ -110,6 +110,38 @@ pub fn get_is_account_key(account: &Address) -> StorageKey {
...
+
+    StorageKey::new(immutable_simulator, key)
+}
+
+#[test]
+fn test_get_immutable_simulator_key() {
+    println!(
+        "{:#?}",
+        get_immutable_simulator_key(&L2_ASSET_ROUTER_ADDRESS, u256_to_h256(96u32.into()))
+    )
```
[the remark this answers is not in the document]
StanislavBreadless: yeah, this is just an artifact, the PR is still WUP

core/node/state_keeper/src/mempool_actor.rs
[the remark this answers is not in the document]
StanislavBreadless: Correct

core/node/state_keeper/src/mempool_actor.rs
StanislavBreadless: Agree, good point.

core/node/state_keeper/src/mempool_actor.rs
```diff
@@ -81,7 +83,8 @@ impl MempoolFetcher {
                 break;
             }
             let latency = KEEPER_METRICS.mempool_sync.start();
-            let mut storage = self.pool.connection_tagged("state_keeper").await?;
+            let mut connection = self.pool.connection_tagged("state_keeper").await?;
+            let mut storage = connection.start_transaction().await?;
```
perekopskiy, line 87: nit: please change name of this var to `transaction` so it will be more clear that executed queries are transactional in code

core/node/state_keeper/src/v26_utils.rs
```diff
@@ -0,0 +1,443 @@
...
+    let stored_l1_address = h256_to_address(&stored_l1_address);
+
+    // No address is stored, it means that the token has never been bridged before
+    // and thus, it is not legacy
+    if stored_l1_address == Address::zero() {
+        return Ok(false);
+    }
+
+    // Just for cross check
+    assert_eq!(expected_l1_address, stored_l1_address);
```
perekopskiy, line 170: If it fails then state keeper will panic which we don't want IMO. Should it be error log + return true as it's unexpected and we don't want to process such deposit anyway until it's investigated? Or maybe `is_l2_token_legacy` can return enum "legacy"/"not-legacy"/"unexpected-l1-address" but anyway we should treat "legacy" and "unexpected-l1-address" the same way later

core/node/state_keeper/src/mempool_actor.rs
StanislavBreadless: I will use `storage_transaction`, it feels like the word `transaction` will collide too much in this context with actual L2 transactions

core/node/state_keeper/src/v26_utils.rs
StanislavBreadless: We should never hit this error in practice as it would mean a new potentially bad bug, but considering that a panic would mean downtime for everyone (and not just L1 txs) it may be better to return an error there, okay will do

core/lib/multivm/src/versions/vm_latest/utils/v26_upgrade.rs
```diff
@@ -0,0 +1,132 @@
...
+        l2_legacy_shared_bridge_address: Address::from_str(
+            "d316600b6009f94ab776e97f1fc985bc18b4e535",
+        )
+        .unwrap(),
+        l1_aliased_shared_bridge: Address::from_str("bcbdabac00000000000000000000000000001112")
+            .unwrap(),
+    }
+}
+
+fn get_test_v26_path(name: &str) -> PathBuf {
```
slowli, line 43: Assuming retaining JSON files is necessary (see the other line comment; after reduction, the *may* be replaceable by inline definitions): The approach with locating a workspace looks fragile / unnecessarily complex. A simpler approach would be to use `include_str!` and copy JSON files to the state keeper crate via symlinks. In this case, there would be some duplication deserializing files, but it should be fairly short. Further, in this case JSON files could be encapsulated in unit tests for each crate (i.e., not compiled as a part of the `multivm` crate API).

core/lib/multivm/src/versions/vm_latest/utils/v26_upgrade.rs
```diff
@@ -0,0 +1,132 @@
...
+
+fn empty_erc20_metadata() -> Vec<u8> {
+    ethabi::encode(&[
+        Token::Bytes(vec![]),
+        Token::Bytes(vec![]),
+        Token::Bytes(vec![]),
+    ])
+}
+
+pub fn encode_new_finalize_deposit(l1_chain_id: U256, l1_token_address: Address) -> Vec<u8> {
```
slowli, line 82: Maybe a dumb suggestion regarding placing this test code (maybe, the entire module, or at least `encode_*deposit` methods). Maybe it makes sense to place it in `zksync_test_contracts` (e.g., as transactions produced by `TestAccount`) given that both `multivm` and `state_keeper` crates use test contracts, and the domain seems roughly right?

core/node/state_keeper/src/v26_utils.rs
```diff
@@ -0,0 +1,481 @@
...
+) -> anyhow::Result<LegacyTokenStatus> {
+    // 1. Read l1 token address from L2 shared bridge (must not be 0)
+
+    let stored_l1_address_key = StorageKey::new(
+        AccountTreeId::new(l2_legacy_shared_bridge_address),
+        get_address_mapping_key(&l2_token_address, L2_LEGACY_SHARED_BRIDGE_L1_ADDRESSES_KEY),
+    );
+    let stored_l1_address = storage
+        .storage_web3_dal()
+        .get_value(&stored_l1_address_key)
```
slowli, line 173: Here and in other `get_value()` calls: Does it matter if new storage logs are inserted in the middle of processing? In this case, later `get_value()` calls may read newer data. For a consistent storage view, it could be possible to use `get_historical_value_unchecked()` instead with a block number obtained at the start of processing. AFAIU, Postgres transactions with the default ["read committed" isolation level](https://www.postgresql.org/docs/14/transaction-iso.html#XACT-READ-COMMITTED) would still see new storage logs, and using stronger isolation could harm performance, so having such app-level consistency looks more appropriate.

core/lib/multivm/src/versions/vm_latest/utils/v26_upgrade.rs
StanislavBreadless: I considered it, but these contracts rely on several imports from the `era-contracts` repo. If I place them fully into the `zksync_test_contracts`, it will require a giant amount of copy-paste from `era-contracts`

core/node/state_keeper/src/v26_utils.rs
StanislavBreadless: I could use `get_historical_value_unchecked`, but in the end of the day it does not seem like it matters too much. If we did not read a newer value, they will be read eventually during the next iterations, the current approach seems to incur less code

StanislavBreadless: Okay, let's use the `get_historical_value_unchecked` just in case

core/lib/multivm/src/versions/vm_latest/utils/v26_upgrade.rs
StanislavBreadless: FYI: I actually moved to this approach because of other reasons, but it appeared a lot easier than I thought initially should've done it from the start

core/lib/multivm/src/versions/testonly/v26_upgrade_utils.rs
```diff
@@ -0,0 +1,289 @@
...
+
+pub(crate) fn test_trivial_test_storage_logs<VM: TestedVm>() {
+    let (vm, test_data) = setup_v26_unsafe_deposits_detection::<VM>();
+    assert_eq!(test_data, get_test_data());
+
+    let storage_ptr = vm.storage.clone();
+    let borrowed = storage_ptr.borrow();
+
+    let expected = remove_irrelevant_keys(borrowed.modified_storage_keys().clone());
+    let found = remove_irrelevant_keys(trivial_test_storage_logs());
```
perekopskiy, line 217: nit: shouldn't it be vice verse i.e. `expected` logs come from file and `found` from the actual execution?
Same for other tests

StanislavBreadless: I would say that no, the `expected` is the source of truth (i.e. how the execution really happens), while `found` is the JSON file that tracks the data.

perekopskiy (approved):

core/lib/dal/src/transactions_dal.rs
```diff
@@ -1752,6 +1752,36 @@ impl TransactionsDal<'_, '_> {
...
+            r#"
+            UPDATE transactions
+            SET
+                in_mempool = FALSE
+            WHERE
+                hash = ANY($1)
+            "#,
+            &hashes as &[&[u8]]
+        )
+        .instrument("return_to_mempool#update_to_false")
```
slowli, line 1771: Nit: The name is probably an artifact; `reset_mempool_status` would work better.

core/lib/multivm/src/versions/testonly/v26_upgrade_utils.rs
```diff
@@ -0,0 +1,289 @@
...
+        test_address,
+        test_contract,
+    );
+
+    vm.vm.push_transaction(system_tx);
+    let result = vm.vm.execute(InspectExecutionMode::OneTx);
+
+    assert!(!result.result.is_failed());
+
+    let l2_token_address = vm.storage.borrow_mut().read_value(&StorageKey::new(
```
slowli, line 153: Nit: `vm.vm.read_storage(_)` would be more idiomatic / general since the new VM implementation doesn't write to `vm.storage`, but rather maintains an internal in-memory write buffer.

core/lib/dal/src/transactions_dal.rs
```diff
@@ -1752,6 +1752,36 @@ impl TransactionsDal<'_, '_> {
...
+        .with_arg("transaction_hashes.len", &hashes.len())
+        .execute(self.storage)
+        .await?;
+
+        // Log debug information about how many rows were affected
```
slowli, line 1776: Bikeshedding: IMO, it'd make more sense to return the number of affected rows from the method and do logging on the caller side. This could e.g. make logging targets more consistent.

core/lib/multivm/src/versions/testonly/v26_upgrade_utils.rs
```diff
@@ -0,0 +1,289 @@
...
+    contract
+        .function("setLegacyTokenAssetId")
+        .unwrap()
+        .encode_input(&[Token::Address(l2_token_address)])
+        .unwrap()
+}
+
+// Returns a list of irrelevant keys that are not required to be the same.
+// For some reason they do not consistently reproduce between test runs, but they are not relevant to the
+// essence of the test.
```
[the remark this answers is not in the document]
StanislavBreadless: Yes, I think it is still better to remove irrelevant keys than to keep only relevant ones

core/lib/dal/src/transactions_dal.rs
StanislavBreadless: Looking at the file it seems like there are a lof of places where `rows_affected` is printed in place, so I guess it makes more sense to do it the same way here

core/lib/multivm/src/versions/testonly/v26_upgrade_utils.rs
```diff
@@ -0,0 +1,288 @@
...
+
+fn setup_v26_unsafe_deposits_detection<VM: TestedVm>() -> (VmTester<VM>, V26TestData) {
+    // In this test, we compare the execution of the bootloader with the predefined
+    // refunded gas and without them
+
+    let l1_chain_id = L1ChainId(1);
+
+    let test_contract = zksync_test_contracts::TestContract::bridge_test();
+    // Any random (but big) address is fine
+    let test_address = Address::from_str("abacabac00000000000000000000000000000000").unwrap();
```
slowli, line 122: Nit: This can use `.parse().unwrap()` as well.

core/lib/test_contracts/build.rs
```diff
@@ -41,22 +41,75 @@ impl ContractEntry {
...
+    manifest_dir: &Path,
+    test_contracts_dir: &Path,
+    factory_deps_to_include: &HashSet<String>,
+    id: &ArtifactId,
+) -> Option<String> {
+    let path_in_dir = id.source.strip_prefix(manifest_dir).ok()?;
+    let next_dir = path_in_dir.iter().next().expect("no dir");
+
+    let module_name = if next_dir.to_str() == test_contracts_dir.to_str() {
+        // We will use the test name folder and not the `contracts` folder for the module
```
slowli, line 54: Nit: "folder" mentions here and in the comment below can be replaced to "directory" as well.

core/lib/types/src/storage/mod.rs
```diff
@@ -110,6 +110,30 @@ pub fn get_is_account_key(account: &Address) -> StorageKey {
...
+pub fn get_immutable_simulator_key(account: &Address, immutable_key: H256) -> StorageKey {
+    // All the immutable variables are stored as inside the ImmutableSimulator system contracts
+    // in a mapping of the following form:
+    // mapping(uint256 contractAddress => mapping(uint256 index => bytes32 value)) internal immutableDataStorage;
+
+    let immutable_simulator = AccountTreeId::new(IMMUTABLE_SIMULATOR_STORAGE_ADDRESS);
+
+    let padded_address = address_to_h256(account);
+
+    let inner_mapping_slot = keccak256(
```
slowli, line 122: Nit: There's `web3::keccak256_concat` function that implements this functionality.

core/node/state_keeper/src/v26_utils/mod.rs
```diff
@@ -0,0 +1,523 @@
...
+        Ok(LegacyTokenStatus::Legacy)
+    } else {
+        Ok(LegacyTokenStatus::NotLegacyOrRegistered)
+    }
+}
+
+/// Accepts a list of transactions to be included into the mempool and filters
+/// the unsafe deposits and returns two vectors:
+/// - Transactions to include into the mempool
+/// - Transactions to return to the mempool
```
slowli, line 227: Nit: The description of the return value is outdated.

slowli (approved):
Suggest to get another review since I feel I have insufficient domain knowledge to review unsafe deposit detection in depth.

core/lib/constants/src/system_context.rs
```diff
@@ -87,3 +87,39 @@ pub const CURRENT_VIRTUAL_BLOCK_INFO_POSITION: H256 = H256([
...
+
+// The `key` for the immutable variable that stores the address of the legacy bridge inside the L2 Asset Router.
+// It is equal to `96`.
+pub const L2_ASSET_ROUTER_LEGACY_SHARED_BRIDGE_IMMUTABLE_KEY: H256 = H256([
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x60,
+]);
+
+// The `key` for the immutable variable that stores the L1 chain id inside the L2 Asset Router.
+// It is equal to `96`.
```
MexicanAce, line 101: Copy/paste error?

core/lib/constants/src/system_context.rs
```diff
@@ -87,3 +87,39 @@ pub const CURRENT_VIRTUAL_BLOCK_INFO_POSITION: H256 = H256([
...
+
+pub const L2_LEGACY_SHARED_BRIDGE_L1_ADDRESSES_KEY: H256 = H256([
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x03,
+]);
+
+pub const L2_NATIVE_TOKEN_VAULT_ASSET_ID_MAPPING_INDEX: H256 = H256([
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xCC,
+]);
```
nikitastupin-matterlabs, line 125: Is it actually `0xCC`? `NativeTokenVault` inherits from `Ownable2StepUpgradeable` and `PausableUpgradeable` both of which have their own and the `__gap` storage variables.

core/node/state_keeper/src/v26_utils/mod.rs
```diff
@@ -0,0 +1,522 @@
...
+            }
+            LegacyTokenStatus::UnexpectedDifferentL1Address(expected, stored) => {
+                tracing::error!(
+                    "Unexpected stored L1 token for L2 token {:#?}. Expected: {:#?}, Stored: {:#?}",
+                    l2_token_address,
+                    expected,
+                    stored
+                );
+                // We return this transaction to ensure that it is not processed
+                return Ok(Some(tx.hash()));
```
nikitastupin-matterlabs, line 326: Why do we mark this case as `Ok`? I'm thinking about the case when some of the above computations have a bug. Then the addresses mismatch and `Ok` is returned while it might not be `Ok`. What bad happens if we returned an error?

StanislavBreadless: Here `Ok`/`Err` is mainly about DB error

StanislavBreadless: In case `Some` is returned, the addition of new transactions will be blocked

nikitastupin-matterlabs (approved):

PR #3650, opened 2025-02-26
feat(eigenda): EigenDA M0

gianbelinche:
## What ❔
This PR extracts the eigenda client into a separate crate and adds new metrics for the grafana dashboard. Completing M0 milestone.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


core/node/da_dispatcher/src/da_dispatcher.rs
```diff
@@ -122,6 +122,7 @@ impl DataAvailabilityDispatcher {
 
         for batch in &batches {
             let dispatch_latency = METRICS.blob_dispatch_latency.start();
+            METRICS.blobs_pending_dispatch.inc_by(1);
```
samlaf, line 125: Is this metric necessary? It will only ever change between 0 and 1 given that batches are submitted sequentially right?

Also what happens if an error is caught and a `?` returns from this function? Won't you be left with a wrong pending count? Does rust have something like golang's `defer METRICS.blobs_pending_dispatch.dec_by(1)`?

core/node/da_clients/src/eigen/client.rs
```diff
@@ -1,46 +1,70 @@
...
+impl EigenDAClient {
     pub async fn new(
         config: EigenConfig,
         secrets: EigenSecrets,
-        get_blob_data: Arc<dyn GetBlobData>,
+        blob_provider: Arc<dyn BlobProvider>,
     ) -> anyhow::Result<Self> {
-        let private_key = SecretKey::from_str(secrets.private_key.0.expose_secret())
-            .map_err(|e| anyhow::anyhow!("Failed to parse private key: {}", e))?;
+        let eth_rpc_url = match config.eigenda_eth_rpc {
```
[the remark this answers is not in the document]
gianbelinche: We cannot implement a trait from a different crate (`DataAvailabilityClient` from `zksync_da_client`) for a struct of an outside crate (`EigenClient` from `eigenda-client-rs`)

```
only traits defined in the current crate can be implemented for types defined outside of the crate
define and implement a trait or new type instead
`EigenClient` is not defined in the current crate
```

core/node/da_dispatcher/src/da_dispatcher.rs
gianbelinche: Yes you are right, I'll remove it.
This used to make sense before, since we had an implementation for a concurrent dispatcher, but we didn't understand exactly how the sequential dispatcher worked (which does have some concurrency due to the separation of dispatch and get inclusion data), and it ended up outperforming our concurrent one.

core/node/da_clients/src/eigen/client.rs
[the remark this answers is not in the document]
samlaf: I know about the orphan rule, but I was saying that the constructor (`new`) is not part of the DataAvailabilityClient trait, so feels weird that it's being defined here in zksync. Why isn't it defined in https://github.com/Layr-Labs/eigenda-client-rs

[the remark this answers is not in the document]
gianbelinche: Since we needed to create a new struct that implements the `DataAvailabilityClient` Trait, it needs to have its own constructor. That constructor does the conversion from zksync types to eigenda types.

core/node/da_clients/Cargo.toml
```diff
@@ -56,18 +56,5 @@ tonic = { workspace = true, features = ["tls-roots", "prost", "codegen"] }
...
-num-bigint.workspace = true
-zksync_web3_decl.workspace = true
-zksync_eth_client.workspace = true
-thiserror.workspace = true
-tempfile.workspace = true
-
-[dev-dependencies]
-serial_test.workspace = true
+rust-eigenda-client.workspace = true
+zksync_dal.workspace = true
```
dimazhornyk, line 60: this import seems unused

core/node/da_clients/src/eigen/client.rs
```diff
@@ -1,46 +1,70 @@
...
+            Some(url) => {
+                let url = Url::from_str(url.expose_str())
+                    .map_err(|_| anyhow::anyhow!("Invalid eth rpc url"))?;
+                Some(rust_eigenda_client::config::SecretUrl::new(url))
+            }
+            None => None,
```
dimazhornyk, line 39: it is always set to some value in `node_builder.rs` and I'm not sure whether your client implementation will work properly with `None` value there

I'd suggest using `ok_or(...)?` here instead of match and removing `Option` in eth_rpc_url in `rust_eigenda_client::config::EigenConfig` (unless there is some way to gracefully handle `None` in `rust_eigenda_client`)

core/node/da_dispatcher/src/metrics.rs
```diff
@@ -32,6 +32,10 @@ pub(super) struct DataAvailabilityDispatcherMetrics {
     pub sealed_to_dispatched_lag: Histogram<Duration>,
     /// The balance of the operator wallet on DA network.
     pub operator_balance: Gauge<u64>,
+    /// Total number of blobs dispatched.
+    pub blobs_dispatched: Gauge<usize>,
```
dimazhornyk, line 36: is there any case where these values will be different from `last_dispatched_l1_batch` and `last_included_l1_batch`?

core/node/da_clients/Cargo.toml
gianbelinche: [Done](https://github.com/matter-labs/zksync-era/pull/3650/commits/1c758caf7a054ea0ae7a07280c805b7edac1227a)

core/node/da_clients/src/eigen/client.rs
gianbelinche: [Done](https://github.com/matter-labs/zksync-era/pull/3650/commits/1d24fbea71045683d745b5c8e71c706bc4e29ec2)

core/node/da_dispatcher/src/metrics.rs
gianbelinche: Yes you are right, thanks for pointing it out!
[Done](https://github.com/matter-labs/zksync-era/pull/3650/commits/672f7c3bfac0871c55de71bee9de7643787cb6fd)

etc/env/file_based/genesis.yaml
```diff
@@ -1,6 +1,6 @@
-genesis_root: 0x7bdb3d822ad837a3611c436d3be457363a08d06d83b74469831482353a7d8277
```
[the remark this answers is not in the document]
gianbelinche: [Done](https://github.com/matter-labs/zksync-era/pull/3650/commits/39078b050200640cb65f9649c077bf5684b919ee)

dimazhornyk (approved):

ischasny (approved):

PR #3651, opened 2025-02-27
Fix typo in test function name: "baxtch" to "batch"

DeVikingMark:
Fixed a typo in the test function name in core/lib/types/src/fee_model.rs where "batch" was incorrectly spelled as "baxtch" in the function name test_compute_baxtch_fee_model_input_v2_only_compute_overhead.


popzxc:
Thanks for the effort, but currently we don't accept minor non-functional contributions.

PR #3652, opened 2025-02-27
fix EVM-578: Forbid both L2 bridge addresses being None

crStiv:
## Description
Implements the check to prevent both L2 bridge addresses (l2_erc20_default_bridge and l2_shared_default_bridge) from being None simultaneously. At least one bridge address must be specified for proper operation.
### Changes
- Added validation in RemoteENConfig::fetch to ensure at least one bridge address is specified
- Added descriptive error message when both addresses are None
- Added unit tests to verify the behavior

### DIFF
--- a/core/bin/external_node/src/config/mod.rs
+++ b/core/bin/external_node/src/config/mod.rs
@@ -169,7 +169,6 @@ impl RemoteENConfig {
         .await?;
 
         // These two config variables should always have the same value.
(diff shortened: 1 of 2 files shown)


popzxc:
Thank you for the effort, but we don't really accept AI-generated contributions; the change mentioned in the TODO requires research on safety of such a check, not blind implementation.

PR #3654, opened 2025-02-27
feat(gateway): Migration to Gateway 

Deniallugo:
## What ❔

Support migration to Gateway. That will be done through restart. 

There are few major changes in server

1. Contracats config now splitted to 3 configs: 
  a. L1SpecificConfig - Stuff that exists only on l1 and will never be migrated to gateway
  b. L2Contracts - Contracts that were deployed on l2
  c. ChainSpecificContracts - Contracts that deployed on l1 and on gateway
2. L1 specific contracts and L2 contracts, for now, should be specified in contracts config, manually. In a future we will be able to load them as well
3. ChainSpecificContracts, are loading from the Settlement Layer, no need to specify
4. There are no configuration for settlement layer. The settlement layer known during the start of the node and the source of truth is contracts both on l1 and settlement layer.
5. Server constantly checks that settlement layer has not changed. If it chaged, server will wait until all processed transaction are sent and confirmed and the necessary contracts are deployed to l1. After that server will be killed. We assume that kubernetes will restart the server
6. During the start server checks the settlement layer and download the necessary contracts config, later on almost all components of the system are manipulating only with settlement layer client and config. Only eth watcher knows about both layers.

External Node:
1. During the sync with the server External node verifies the transaction on l1 for consistency. Once consistency checker is panicking, external node crashes
2. During the start External node checks the latest eth txs in db and depends on it setup the system to work either on Gateway or on L1 

Migration process
1. Operator call the server notifier contract and this contract will send the message to server, that all commit operations, should be stopped
2. Server stops the commit operations, but continue to work in all other parts
3. Operator call the migrate to gateway function on contracts
4. Once the migration is executed both on l1 and gateway, server will crash
5. During the next start server will load all necessary contracts and continue to work 

## Why ❔

The migration to settlement layer should be smooth. Without necessary of manual restarts and setting the contracts, it will drastically increase the mistakes, especially with multiple chains in operation.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


core/bin/zksync_server/src/main.rs
```diff
@@ -189,10 +176,28 @@ fn main() -> anyhow::Result<()> {
         return Ok(());
     }
 
-    node.run(observability_guard)?;
+    let result = node.run(observability_guard);
+    if find_gateway_update_error(&result) {
+        println!("Gateway has changed")
+    };
```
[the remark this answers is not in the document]
Deniallugo: During the development, it was better to see this reason, otherwise it's jut got lost in in logs

core/lib/config/src/configs/contracts/chain.rs
```diff
@@ -0,0 +1,129 @@
...
+#[derive(Debug, Clone)]
+pub struct ChainContracts {
+    pub diamond_proxy_addr: Address,
+    pub chain_admin: Option<Address>,
+}
+
+// Contracts deployed to the l2
+#[derive(Debug, Clone)]
+pub struct L2Contracts {
+    pub erc20_default_bridge: Option<Address>,
```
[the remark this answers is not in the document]
Deniallugo: we still have them in api

core/lib/config/src/configs/contracts/chain.rs
```diff
@@ -0,0 +1,129 @@
+use serde::Deserialize;
+use zksync_basic_types::{Address, H256};
+
+use crate::configs::contracts::{
+    ecosystem::{EcosystemCommonContracts, EcosystemContracts, L1SpecificContracts},
+    ChainSpecificContracts,
+};
+
+/// Data about deployed contracts unified l1/l2 contracts and bridges. To Be Deleted
```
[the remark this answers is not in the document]
Deniallugo: removed to be deleted, unfortunately, it's impossible at this stage

core/lib/config/src/configs/contracts/gateway.rs
```diff
@@ -24,15 +24,13 @@ pub struct GatewayConfig {
...
+    pub validator_timelock_addr: Option<Address>,
     pub multicall3_addr: Address,
     pub diamond_proxy_addr: Address,
-    // TODO(EVM-921): there is no "governace" for a chain, only an admin, we
-    // need to figure out what we mean here
-    pub chain_admin_addr: Option<Address>,
-    pub governance_addr: Address,
+    pub chain_admin_addr: Address,
     pub gateway_chain_id: SLChainId,
+    pub server_notifier: Option<Address>,
```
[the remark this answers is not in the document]
Deniallugo: actually, this struct is used only in zkstack, now it's redundant in this struct. probably the whole struct should be moved to zkstack

core/lib/contracts_loader/src/lib.rs
```diff
@@ -0,0 +1,135 @@
...
+        .for_contract(ctm_address, &state_transition_manager_contract())
+        .call(sl_client)
+        .await
+        .map(|a: Address| if a.is_zero() { None } else { Some(a) })?;
+
+    let validator_timelock_addr = CallFunctionArgs::new("validatorTimelock", ())
+        .for_contract(ctm_address, &state_transition_manager_contract())
+        .call(sl_client)
+        .await
+        .map(|a: Address| if a.is_zero() { None } else { Some(a) })?;
```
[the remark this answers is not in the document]
Deniallugo: server notifier yes, others most likely not

core/lib/contracts_loader/src/lib.rs
```diff
@@ -0,0 +1,135 @@
...
+        base_token_address: base_token,
+        erc_20_bridge: erc20_bridge,
+        // TODO add support for loading bytecodes_supplier_addr and
+        // wrapped_base_token_store from bridgehub
+        bytecodes_supplier_addr: None,
+        wrapped_base_token_store: None,
+    })
+}
+
+pub async fn get_settlement_layer_for_l1_call(
```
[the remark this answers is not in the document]
Deniallugo: I'll add a description. 
Unfortunately, we have slightly non obvious behaviour and the assumption that settlement layer is zero == `SettlementMode::SettlesToL1` is true only for l1. for gateway it'd be vice versa behaviour. 

so we need to use this function only for l1.

core/lib/gateway_migrator/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+            let gateway_client = gateway_client.unwrap();
+
+            let sl_contracts = load_settlement_layer_contracts(
+                gateway_client,
+                L2_BRIDGEHUB_ADDRESS,
+                l2chain_id,
+                None,
+            )
+            .await?;
+            // Wait until the contracts are deployed on l2
```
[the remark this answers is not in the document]
Deniallugo: this check waits for l1->l2 transcation on gateway to be processed. 
we have exactly the same check during the start (even the same function)

core/node/eth_watch/src/event_processors/gateway_migration.rs
```diff
@@ -0,0 +1,75 @@
...
+pub struct GatewayMigrationProcessor {
+    gateway_migration_topic: H256,
+    l2chain_id: L2ChainId,
+}
+
+impl GatewayMigrationProcessor {
+    pub fn new(l2chain_id: L2ChainId) -> Self {
+        Self {
+            gateway_migration_topic: server_notifier_contract()
+                .event("MigrateToGateway")
```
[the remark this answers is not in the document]
Deniallugo: I have to try it one more time, but shoudn't be a problem

core/bin/external_node/src/config/mod.rs
```diff
@@ -1530,17 +1553,57 @@ impl From<&ExternalNodeConfig> for TxSenderConfig {
...
+            timestamp_asserter_params: config.remote.l2_timestamp_asserter_addr.map(|addr| {
                 TimestampAsserterParams {
-                    address,
+                    address: addr,
                     min_time_till_end: Duration::from_secs(
-                        config
-                            .optional
-                            .timestamp_asserter_min_time_till_end_sec
-                            .into(),
+                        config.optional.timestamp_asserter_min_time_till_end_sec as u64,
```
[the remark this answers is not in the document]
Deniallugo: just an accident. fixed back

core/lib/dal/src/eth_sender_dal.rs
```diff
@@ -67,6 +67,7 @@ impl EthSenderDal<'_, '_> {
 
     pub async fn get_non_gateway_inflight_txs_count_for_gateway_migration(
         &mut self,
+        is_gateway: bool,
```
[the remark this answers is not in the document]
Deniallugo: renamed

core/lib/contracts_loader/src/lib.rs
```diff
@@ -0,0 +1,135 @@
...
+pub async fn load_l1_specific_contracts(
+    eth_client: &dyn EthInterface,
+    bridgehub_address: Address,
+    l2_chain_id: L2ChainId,
+) -> anyhow::Result<L1SpecificContracts> {
+    let base_token = CallFunctionArgs::new("baseToken", Token::Uint(l2_chain_id.as_u64().into()))
+        .for_contract(bridgehub_address, &bridgehub_contract())
+        .call(eth_client)
+        .await
+        .map(|a: Address| if a.is_zero() { None } else { Some(a) })?;
```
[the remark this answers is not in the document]
Deniallugo: Made it required

core/lib/contracts_loader/src/lib.rs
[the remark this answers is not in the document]
Deniallugo: made it non optional

core/lib/config/src/configs/contracts/gateway.rs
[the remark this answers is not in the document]
Deniallugo: removed server_notifier

core/lib/contracts_loader/src/lib.rs
```diff
@@ -0,0 +1,135 @@
...
+            .for_contract(bridgehub_address, &bridgehub_contract())
+            .call(sl_client)
+            .await?;
+
+    if gateway_diamond_proxy.is_zero() {
+        return Ok(None);
+    }
+
+    let ctm_address =
+        CallFunctionArgs::new("chainTypeManager", Token::Uint(l2_chain_id.as_u64().into()))
```
StanislavBreadless, line 32: this code will work only for chains that upgraded to v26. This makes the release of this thing blocked until all chains are upgraded to v26

core/lib/contracts_loader/src/lib.rs
```diff
@@ -0,0 +1,135 @@
...
+            .for_contract(bridgehub_address, &bridgehub_contract())
+            .call(sl_client)
+            .await?;
+
+    let server_notifier_addr = CallFunctionArgs::new("serverNotifierAddress", ())
```
StanislavBreadless, line 37: the same goes here, is it okay that it will return an error for the system before v27?

core/lib/contracts_loader/src/lib.rs
```diff
@@ -0,0 +1,135 @@
...
+        .for_contract(ctm_address, &state_transition_manager_contract())
+        .call(sl_client)
+        .await
+        .map(|a: Address| if a.is_zero() { None } else { Some(a) })?;
+
+    let validator_timelock_addr = CallFunctionArgs::new("validatorTimelock", ())
+        .for_contract(ctm_address, &state_transition_manager_contract())
```
StanislavBreadless, line 44: this will also not work fine for chains that have not yet upgraded to v26. For v25->v26 upgrade we used a trick where:
- if the chain has the same version as on CTM, we use the CTM's timelock
- we take it from config otherwise

not sure if the code in this PR is compatible with it

core/lib/contracts_loader/src/lib.rs
```diff
@@ -0,0 +1,135 @@
...
+        .await
+        .map(|a: Address| if a.is_zero() { None } else { Some(a) })?;
+
+    let shared_bridge = CallFunctionArgs::new("sharedBridge", ())
+        .for_contract(bridgehub_address, &bridgehub_contract())
+        .call(eth_client)
+        .await
+        .map(|a: Address| if a.is_zero() { None } else { Some(a) })?;
+
+    let erc20_bridge = CallFunctionArgs::new("legacyBridge", ())
```
StanislavBreadless, line 86: there is no such function in bridgehub

core/lib/contracts_loader/src/lib.rs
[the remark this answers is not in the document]
StanislavBreadless: maybe better to rename to get_settlement_layer_from_l1? It would say where it was taken from instead that it is for l1 call as it is the settlement layer that should be used for l1 calls, which is not correct

core/lib/contracts_loader/src/lib.rs
Deniallugo: since it required only for gateway migration, i think we can wait a bit

core/lib/types/src/server_notification.rs
```diff
@@ -0,0 +1,17 @@
+use zksync_basic_types::L1BlockNumber;
+
+use crate::H256;
+
+#[derive(Debug, PartialEq)]
+pub enum GatewayMigrationState {
```
StanislavBreadless, line 6: is there any difference logic-wise between "NotStarted" and "Finalized"? Maybe it makes to change it to `InProgress` and `NotInProgress`

core/node/consistency_checker/src/lib.rs
```diff
@@ -380,52 +376,36 @@ impl ConsistencyChecker {
     const DEFAULT_SLEEP_INTERVAL: Duration = Duration::from_secs(5);
 
     pub async fn new(
-        l1_client: Box<DynClient<L1>>,
-        gateway_client: Option<Box<DynClient<L1>>>,
+        gateway_client: Box<dyn EthInterface>,
```
StanislavBreadless, line 379: is it only gateway client or settlement layer client?

StanislavBreadless, line 379: is it only gateway client or settlement layer client?

core/lib/contracts_loader/src/lib.rs
StanislavBreadless: probably fallback is better since otherwise you will get a long living branch which modifies multiple components (and so there will be lots of conflicts to resolve)

core/lib/contracts_loader/src/lib.rs
Deniallugo: removed this function, will add somewhere in a future

core/bin/external_node/src/node_builder.rs
```diff
@@ -204,27 +225,12 @@ impl ExternalNodeBuilder {
         // compression.
         const OPTIONAL_BYTECODE_COMPRESSION: bool = true;
 
-        let l2_shared_bridge_addr = self
-            .config
-            .remote
-            .l2_shared_bridge_addr
-            .context("Missing `l2_shared_bridge_addr`")?;
-        let l2_legacy_shared_bridge_addr = if l2_shared_bridge_addr == L2_ASSET_ROUTER_ADDRESS {
```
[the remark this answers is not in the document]
Deniallugo: Thanks for noticing! Reimplemented

core/lib/contracts_loader/src/lib.rs
Deniallugo: do we have any chains, where i can verify it?

core/lib/types/src/server_notification.rs
Deniallugo: fixed

core/node/consistency_checker/src/lib.rs
Deniallugo: fixed

core/lib/dal/migrations/20250218134333_server_notification.up.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE IF NOT EXISTS server_notifications
+(
+    id                    serial        primary key,
```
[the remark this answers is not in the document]
Deniallugo: i don't see any advantages. less space and that's all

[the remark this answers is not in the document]
perekopskiy: yeah, I wasn't even thinking about space but rather number of columns. You can express the same without id and it won't be more complex or less complex IMO. Up to you

core/node/api_server/src/web3/namespaces/en.rs
```diff
@@ -147,27 +147,28 @@ impl EnNamespace {
...
+                .api_config
+                .l1_ecosystem_contracts
+                .bridgehub_proxy_addr
+                .unwrap(),
+            state_transition_proxy_addr: self
+                .state
+                .api_config
+                .l1_ecosystem_contracts
+                .state_transition_proxy_addr,
+            transparent_proxy_admin_addr: None,
```
perekopskiy, line 163: btw this looks like a breaking change. I see that you removed `transparent_proxy_admin_addr` from EN but we can't be sure that someone external doesn't uses this field. I don't think you can just drop it.

Deniallugo: it's completely useless field for everyone

core/bin/zksync_server/src/node_builder.rs
```diff
@@ -185,15 +187,22 @@ impl MainNodeBuilder {
         let query_eth_client_layer = QueryEthClientLayer::new(
             genesis.l1_chain_id,
             eth_config.l1_rpc_url,
-            self.gateway_chain_config
-                .as_ref()
-                .map(|c| c.gateway_chain_id),
+            // TODO query from the server
+            None,
```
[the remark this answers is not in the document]
Deniallugo: solved

core/node/node_framework/src/implementations/layers/settlement_layer_data.rs
```diff
@@ -0,0 +1,193 @@
...
+#[async_trait::async_trait]
+impl WiringLayer for SettlementLayerData {
+    type Input = Input;
+    type Output = Output;
+
+    fn layer_name(&self) -> &'static str {
+        "settlement_layer_data"
+    }
+
+    async fn wire(self, input: Self::Input) -> Result<Self::Output, WiringError> {
```
perekopskiy, line 88: I don't quite like that we put so much logic in `wire`, it is supposed to be simple, ideally without side effects, as simple as constructing tasks and resources and outputting them. But if we stick to it then we can't really get the resources we want here, like contract addresses etc. @popzxc @slowli Curious what you think

perekopskiy (commented):
Overall the approach looks good! I'm a bit concerned that it's pretty complex but don't have suggestions how it can be improved.

core/lib/config/src/configs/contracts/ecosystem.rs
```diff
@@ -0,0 +1,54 @@
+// External uses
+use serde::{Deserialize, Serialize};
+use zksync_basic_types::Address;
+
+// Unified ecosystem contracts. To be deleted, after contracts config migration
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+pub struct EcosystemContracts {
```
StanislavBreadless, line 7: These contracts refer to L1 only, why do we need a separate `L1SpecificContracts`?

core/node/consistency_checker/src/lib.rs
```diff
@@ -777,7 +737,6 @@ impl ConsistencyChecker {
                     self.event_handler
                         .report_inconsistent_batch(batch_number, &err);
                     match &self.l1_data_mismatch_behavior {
-                        #[cfg(test)]
                         L1DataMismatchBehavior::Bail => {
```
StanislavBreadless, line 740: do we need to handle this bail somehow similarly to how you handled the error from gateway migrator?

core/node/node_framework/src/implementations/layers/settlement_layer_client.rs
```diff
@@ -0,0 +1,74 @@
...
+#[context(crate = crate)]
+pub struct Input {
+    initial_settlement_mode: SettlementModeResource,
+    sl_chain_id_resource: SlChainIdResource,
+}
+
+#[derive(Debug, IntoContext)]
+#[context(crate = crate)]
+pub struct Output {
+    query_client_settlement_layer: UniversalClientResource,
```
StanislavBreadless, line 41:
```suggestion
    settlement_layer_client: UniversalClientResource,
```

core/node/node_framework/src/implementations/layers/settlement_layer_client.rs
```diff
@@ -0,0 +1,74 @@
...
+    type Input = Input;
+    type Output = Output;
+
+    fn layer_name(&self) -> &'static str {
+        "settlement_layer_client"
+    }
+
+    async fn wire(self, input: Self::Input) -> Result<Output, WiringError> {
+        Ok(Output {
+            query_client_settlement_layer: match input.initial_settlement_mode.0 {
```
StanislavBreadless, line 55: do you think it makes sense to couple `query_client_settlement_layer: UniversalClientResource` and the resource for the intiial settlement mode? This way it would be less code, and also it would be obvious for everyone that `query_client_settlement_layer` is just a client for whatever is set in `initial_settlement_mode` without any additional processing

core/lib/gateway_migrator/src/lib.rs
```diff
@@ -0,0 +1,137 @@
...
+    // Load chain contracts from sl
+    let sl_contracts =
+        load_settlement_layer_contracts(sl_client, bridge_hub_address, l2chain_id, None).await?;
+    // Deploying contracts on gateway are going through l1->l2 communication,
+    // even though the settlement layer has changed on l1.
+    // Gateway should process l1->l2 transaction.
+    // Even though when we switched from gateway to l1,
+    // we don't need to wait for contracts deployment,
+    // we have to wait for l2->l1 communication to be finalized
+    let res = if let Some(contracts) = sl_contracts {
```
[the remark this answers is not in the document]
Deniallugo: But it actually handles the situation, when we have gateway as a seetlement layer, but the l1->l2 transaction has not passed yet. 

So the contracts are still none and we can't switch it

core/node/consistency_checker/src/lib.rs
Deniallugo: So, we don't handle it ;) 
we have a find function for convenience. I Don't think we have to handle it somehow else

core/node/node_framework/src/implementations/layers/settlement_layer_data_en.rs
```diff
@@ -0,0 +1,150 @@
...
+            .await?
+        };
+
+        let l2_eth_client = self
+            .gateway_rpc_url
+            .map(|url| Client::http(url).context("Client::new()"))
+            .transpose()?
+            .map(|mut builder| {
+                if initial_sl_mode.is_gateway() {
+                    builder = builder.for_network(L2ChainId::new(chain_id.0).unwrap().into());
```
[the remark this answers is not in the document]
Deniallugo: i don't think it's worth it..

core/bin/zksync_server/src/main.rs
```diff
@@ -175,9 +175,9 @@ fn main() -> anyhow::Result<()> {
 
     let result = node.run(observability_guard);
     if find_gateway_update_error(&result) {
-        tracing::warn!("Gateway has changed")
+        tracing::warn!("Settlement layer has changed")
     };
-    Ok(())
+    Ok(result?)
```
StanislavBreadless, line 179: O think you can just return `result`

Deniallugo: we need conversion from `ZkStackServiceError` to anyhow::Error

StanislavBreadless (dismissed):
Would it be possible to get back into some v25 commit, initialize an ecosystem there and then launch this version of the server + run int tests? Just to double check that all the interfaces are compatible with v25 chains

RomanBrodetski (commented):
Epic work and well done overall!

I left a couple of comments - some of them suggest direct improvements (please apply), others are optional suggestions - but please consider them.

Overall of course the configuration is not straigtforward - with some values taken from config, others from L1, and some of the config values overwritten with hardcoded values. I feel like there may exist an abstraction that would make this logic more explicit, but I cannot come up with one right now. I'll keep thinking and suggest if I come up with one. Maybe after writing this @Deniallugo - do you feel like there are other approaches, or would you go the same way if you started over?

But regardless, good work overall and let's see if we can improve

core/bin/block_reverter/src/main.rs
```diff
@@ -235,33 +232,51 @@ async fn main() -> anyhow::Result<()> {
...
-    } else {
-        (
-            l1_secrets.l1_rpc_url,
-            contracts.diamond_proxy_addr,
-            contracts.validator_timelock_addr,
-        )
+    let (client, contracts) = match settlement_mode {
+        SettlementMode::SettlesToL1 => (eth_client, sl_l1_contracts),
+        SettlementMode::Gateway => {
+            let gateway_client: Client<L1> = Client::http(l1_secrets.gateway_rpc_url.unwrap())
```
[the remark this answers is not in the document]
Deniallugo: it's gateway here. 

I didn't want to touch block reverter a lot. probably can do it later and get rid of <L1> and switch to Box<dyn EthInterface>

core/lib/contracts_loader/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+        sl_client,
+        bridgehub_address,
+        l2_chain_id,
+        multicall3,
+    )
+    .await?;
+    Ok(result)
+}
+
+async fn load_settlement_layer_contracts_pure_error(
```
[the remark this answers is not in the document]
Deniallugo: Not anymore thanks.

core/bin/block_reverter/src/main.rs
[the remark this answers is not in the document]
Deniallugo: it's not working. and it wasn't before. 
Scope is already really big. 
changing block reverter to work through the gateway changes will be hard

core/lib/dal/src/server_notifier.rs
```diff
@@ -0,0 +1,76 @@
...
+pub struct ServerNotificationsDal<'a, 'c> {
+    pub storage: &'a mut Connection<'c, Core>,
+}
+
+impl ServerNotificationsDal<'_, '_> {
+    pub async fn save_notification(
+        &mut self,
+        main_topic: H256,
+        l1_block_number: L1BlockNumber,
+        value: serde_json::Value,
```
slowli, line 16: - Nit: Why doesn't this use `ServerNotification`?
- What does `ServerNotification.value` supposed to mean? AFAIU, it's always set to `null` in the `eth_watch` code.

core/node/eth_watch/src/event_processors/gateway_migration.rs
```diff
@@ -0,0 +1,67 @@
...
+                    .as_bytes(),
+            );
+
+            if L2ChainId::from(chain_id.as_u32()) != self.l2chain_id {
+                continue;
+            }
+
+            storage
+                .server_notifications_dal()
+                .save_notification(
```
slowli, line 41: Do we care about atomicity here? I.e., would saving all events in a transaction / single DB query make a difference? (Off-topic: Currently it doesn't look like `EthWatch::loop_iteration()` calls `process_events()` for processors and updates the next block to process for the processor in a transaction, which may lead to a process the same events multiple times; is this OK?)

core/node/eth_watch/src/client.rs
```diff
@@ -65,6 +65,8 @@ pub trait EthClient: 'static + fmt::Debug + Send + Sync {
         hashes: Vec<H256>,
     ) -> EnrichedClientResult<Vec<Option<Vec<u8>>>>;
 
+    async fn get_settlement_layer(&self) -> Result<Address, ContractCallError>;
```
slowli, line 68: Dumb question: It doesn't look like this method is used; the only use is in proxying `ZkSyncExtentionEthClientW` implementation.

core/node/eth_watch/src/client.rs
```diff
@@ -616,19 +652,19 @@ impl L2EthClient for EthHttpQueryClient<L2> {
...
 
-/// Wrapper for L2 client object.
-/// It is used for L2EthClient -> EthClient dyn upcasting coercion:
-///     Arc<dyn L2EthClient> -> L2EthClientW -> Arc<dyn EthClient>
+/// Wrapper for ZkSyncExtention client object.
+/// It is used for ZkSyncExtentionEthClient -> EthClient dyn upcasting coercion:
+///     Arc<dyn ZkSyncExtentionEthClient> -> ZkSyncExtentionEthClientW -> Arc<dyn EthClient>
 #[derive(Debug, Clone)]
-pub struct L2EthClientW(pub Arc<dyn L2EthClient>);
+pub struct ZkSyncExtentionEthClientW(pub Arc<dyn ZkSyncExtentionEthClient>);
```
slowli, line 659: This entire construction is unnecessary. You can add the upcasting method directly into the `ZkSyncExtentionEthClient` trait:

```rust
#[async_trait::async_trait]
pub trait ZkSyncExtentionEthClient: EthClient {
    fn into_base(self: Arc<Self>) -> Arc<dyn EthClient>;
    // other stuff...
}

#[async_trait::async_trait]
impl ZkSyncExtentionEthClient for EthHttpQueryClient<L1> {
    fn into_base(self: Arc<Self>) -> Arc<dyn EthClient> {
        self
    }
    // ...
}
```

Unfortunately, upcasting cannot have a default implementation, but it's trivial to add it to the 3 impls.

slowli (commented):
Some comments (mostly nits / bikeshedding, but there are a couple of serious ones); will try to take another look later.

core/lib/config/src/configs/contracts/chain.rs
```diff
@@ -0,0 +1,141 @@
...
+            bytecodes_supplier_addr: self.l1_bytecode_supplier_addr,
+            wrapped_base_token_store: self.l1_wrapped_base_token_store_addr,
+            bridge_hub: Some(self.bridgehub_proxy_addr),
+            shared_bridge: self.l1_shared_bridge_proxy_addr,
+            erc_20_bridge: self.l1_erc20_bridge_proxy_addr,
+            base_token_address: self.base_token_addr,
+        }
+    }
+
+    pub fn l2_contracts(&self) -> L2Contracts {
```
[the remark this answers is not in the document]
Deniallugo: meanwhile, i really like the idea. 
but the amount of refactoring, that requires to support config management is humongous. i'm not sure, that i want to introduce it here as well. (p.s. i tried)

core/node/node_framework/src/implementations/layers/settlement_layer_data.rs
```diff
@@ -0,0 +1,353 @@
...
+}
+
+// Gateway has different rules for pubdata and gas space.
+// We need to adjust it accordingly.
+fn adjust_eth_sender_config(
+    mut config: SenderConfig,
+    settlement_mode: SettlementMode,
+) -> SenderConfig {
+    if settlement_mode.is_gateway() {
+        config.max_aggregated_tx_gas = 4294967295;
```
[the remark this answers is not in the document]
Deniallugo: yep, i'm going to get rid of this config. and calcualate max_aggregated_tx_gas properly per tx in the next pr

core/lib/dal/src/eth_sender_dal.rs
```diff
@@ -65,8 +65,9 @@ impl EthSenderDal<'_, '_> {
         Ok(txs.into_iter().map(|tx| tx.into()).collect())
     }
 
-    pub async fn get_non_gateway_inflight_txs_count_for_gateway_migration(
+    pub async fn get_inflight_txs_count_for_gateway_migration(
         &mut self,
+        is_gateway: bool,
```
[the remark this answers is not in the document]
Deniallugo: yes, but in db, it'd require a major refactoring and not compatible with current schema

core/bin/zksync_server/src/main.rs
```diff
@@ -187,8 +172,26 @@ fn main() -> anyhow::Result<()> {
...
+            if let TaskError::TaskFailed(task, err) = task {
+                if task.contains("gateway_migrator")
+                    && err.to_string().contains("Settlement layer changed")
+                {
+                    return true;
+                }
+            }
+        }
+    }
+    false
```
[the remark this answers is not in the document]
Deniallugo: Yes, we can, and we can get rid of this function, but it just helps to read the logs.

core/node/consistency_checker/src/lib.rs
```diff
@@ -855,3 +804,24 @@ async fn wait_for_l1_batch_with_metadata(
...
+// Get settlement layer based on ETH tx, all eth txs should be presented on settlement layer. what is the best place for this function?
+pub async fn get_db_settlement_mode(
+    db_pool: ConnectionPool<Core>,
+    l1chain_id: SLChainId,
+) -> anyhow::Result<Option<SettlementMode>> {
+    let db_chain_id = db_pool
+        .connection()
+        .await?
+        .eth_sender_dal()
+        .get_chain_id_of_last_eth_tx()
```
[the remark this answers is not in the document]
Deniallugo: It's not a big deal. In the worst case scenario we will. have +1 restart for en

core/node/eth_sender/src/eth_tx_aggregator.rs
```diff
@@ -852,3 +873,102 @@ impl EthTxAggregator {
...
+async fn gateway_status(
+    storage: &mut Connection<'_, Core>,
+    sl_layer: &SettlementMode,
+) -> GatewayMigrationState {
+    let to_gateway = server_notifier_contract()
+        .event("MigrateToGateway")
+        .unwrap()
+        .signature();
+    let from_gateway = server_notifier_contract()
+        .event("MigrateFromGateway")
```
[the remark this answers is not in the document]
Deniallugo: Unfortunately, it'd require some more effort and change the library for working with eth

core/lib/web3_decl/src/namespaces/zks.rs
```diff
@@ -46,7 +46,10 @@ pub trait ZksNamespace {
     async fn get_bridgehub_contract(&self) -> RpcResult<Option<Address>>;
 
     #[method(name = "getMainContract")]
-    async fn get_main_contract(&self) -> RpcResult<Address>;
+    async fn get_main_l1_contract(&self) -> RpcResult<Address>;
+
+    #[method(name = "getL2Multicall3")]
+    async fn get_l2_multicall3(&self) -> RpcResult<Option<Address>>;
```
popzxc, line 52: Why do we need this as an RPC method? Any new method is a PITA to live with. Can we live without it?

core/lib/web3_decl/src/namespaces/zks.rs
```diff
@@ -46,7 +46,10 @@ pub trait ZksNamespace {
     async fn get_bridgehub_contract(&self) -> RpcResult<Option<Address>>;
 
     #[method(name = "getMainContract")]
-    async fn get_main_contract(&self) -> RpcResult<Address>;
+    async fn get_main_l1_contract(&self) -> RpcResult<Address>;
```
popzxc, line 49: We now have disconnect between actual RPC method name and method name in core. IMHO it adds more confusion.

core/node/node_storage_init/src/main_node/genesis.rs
```diff
@@ -13,7 +13,7 @@ use crate::traits::InitializeStorage;
 #[derive(Debug)]
 pub struct MainNodeGenesis {
     pub genesis: GenesisConfig,
-    pub contracts: ContractsConfig,
+    pub contracts: SettlementLayerSpecificContracts,
```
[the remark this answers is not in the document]
Deniallugo: In this particular case, i'd prefer to keep it like this, for having lesser chance to pass the wrong value

core/node/genesis/src/lib.rs
```diff
@@ -106,11 +106,7 @@ impl GenesisParams {
             default_aa: config
                 .default_aa_hash
                 .ok_or(GenesisError::MalformedConfig("default_aa_hash"))?,
-            evm_emulator: Some(
-                config
-                    .evm_emulator_hash
-                    .ok_or(GenesisError::MalformedConfig("evm_emulator_hash"))?,
-            ),
+            evm_emulator: config.evm_emulator_hash,
```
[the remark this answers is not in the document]
Deniallugo: yes, we don't have v27 upgrade everywhere.

core/node/node_framework/src/implementations/layers/settlement_layer_data.rs
popzxc: You already have logic for migrating the settlement layer in gateway migrator, no?
I would prefer to simply configure the node for the settlement layer we have no, and then gateway migrator can reboot the node when needed.

Then this method would truly be doing wiring only.

popzxc (commented):
Did a partial review; hopefully will do a round 2 tomorrow

core/lib/dal/src/server_notifier.rs
Deniallugo: It was an idea to migrate all server notification to this code. but looks like a premature optimization

core/node/eth_watch/src/event_processors/gateway_migration.rs
Deniallugo: It's ok, until the order is correct.

core/lib/web3_decl/src/namespaces/zks.rs
Deniallugo: Yes, but considered the fact, we have more than one main_contract it's useful to have it like this for not messing up. changing public interface is a bad idea

core/lib/web3_decl/src/namespaces/zks.rs
Deniallugo: We need it for getting l2 multical address from the gateway node. We can't predict this value

core/lib/config/src/configs/contracts/chain.rs
```diff
@@ -0,0 +1,141 @@
...
+// Contracts specific for the chain. Should be deployed to all Settlement Layers
+#[derive(Debug, Clone)]
+pub struct ChainContracts {
+    pub diamond_proxy_addr: Address,
+    pub chain_admin: Option<Address>,
+}
+
+// Contracts deployed to the l2
+#[derive(Debug, Clone)]
+pub struct L2Contracts {
```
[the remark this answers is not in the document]
Deniallugo: Seems, they all independently could be None. 

`erc20_default_bridge` and `shared_bridge_addr` should be always the same and we even have a linear task for making them not none 
`EVM-578`
Same for `legacy_shared_bridge_addr` it's now not none, but will be soon. 

`timestamp_asserter_addr` is optional. Not required to run the server. 
`da_validator_addr` is non optional only for Validium chains 

`testnet_paymaster_addr` is not none only for testnets
 `multicall3` is not none only for gateway

core/lib/contracts_loader/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+use zksync_eth_client::{CallFunctionArgs, ContractCallError, EthInterface};
+use zksync_types::{
+    ethabi::{Contract, Token},
+    protocol_version::ProtocolSemanticVersion,
+    settlement::SettlementMode,
+    Address, L2ChainId, SLChainId, U256,
+};
+
+/// Load contacts specific for each settlement layer, using bridgehub contract
+pub async fn load_settlement_layer_contracts(
```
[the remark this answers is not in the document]
Deniallugo: I tried it a bit, but it lead to some unsatisfied boundary requirements. Looks like from async_trait. Decided not to continue wasting time on it

core/lib/gateway_migrator/src/lib.rs
```diff
@@ -0,0 +1,137 @@
...
+                .await?
+            {
+                bail!("Settlement layer changed")
+            }
+            tokio::time::sleep(Duration::from_secs(1)).await;
+        }
+    }
+}
+
+pub async fn switch_to_current_settlement_mode(
```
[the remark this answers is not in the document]
Deniallugo: It'd require to instantiate `GatewayMigrator`. Which is not necessary for wiring layer

core/node/eth_sender/src/eth_tx_aggregator.rs
```diff
@@ -852,3 +873,102 @@ impl EthTxAggregator {
...
+        .into_uint()
+        .unwrap();
+
+    Ok(Some(priority_tree_start_index.as_usize()))
+}
+
+async fn gateway_status(
+    storage: &mut Connection<'_, Core>,
+    sl_layer: &SettlementMode,
+) -> GatewayMigrationState {
```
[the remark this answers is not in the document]
Deniallugo: It's the same file. I do it in a private function. It's very possible that the logic will change and we won't only apply the current restriction, but do something else. I'd like to split this two things. Getting the current state of the system and make a decision based on it.

core/node/node_framework/src/implementations/layers/settlement_layer_data.rs
```diff
@@ -0,0 +1,353 @@
...
+
+        let switch = switch_to_current_settlement_mode(
+            initial_sl_mode,
+            sl_client,
+            self.config.l2_chain_id,
+            &mut pool.connection().await.context("Can't connect")?,
+            bridgehub,
+            &getters_facet_contract(),
+        )
+        .await?;
```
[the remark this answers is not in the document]
Deniallugo: Changed this method to what you suggested

core/bin/block_reverter/src/main.rs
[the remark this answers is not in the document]
Deniallugo: It looks like it will take a bit more time. 
In the most common scenario it works. (reverting blocks on one SL) 
I'd add the rest in the additional pr

core/bin/zksync_server/src/node_builder.rs
```diff
@@ -92,28 +97,37 @@ pub struct MainNodeBuilder {
...
     pub fn new(
         configs: GeneralConfig,
         wallets: Wallets,
         genesis_config: GenesisConfig,
-        contracts_config: ContractsConfig,
-        gateway_chain_config: Option<GatewayChainConfig>,
         secrets: Secrets,
+        l1_specific_contracts: L1SpecificContracts,
+        l2_contracts: L2Contracts,
+        l1_sl_contracts: Option<SettlementLayerSpecificContracts>,
```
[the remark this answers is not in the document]
Deniallugo: added comments to the struct it self

core/lib/config/src/configs/contracts/chain.rs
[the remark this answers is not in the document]
Deniallugo: as an experiment set bridges to not null

core/node/node_framework/src/implementations/layers/settlement_layer_data.rs
Deniallugo: i've tried to make it as simple as possible. 
It's still contains some code. especially for en.

RomanBrodetski (dismissed):
Thanks for taking care of my comments. I don't have any more at the moment. 

I cannot say I'm happy with the current state, as already convoluted config system became more complex, but I don't have concrete guidelines on how to improve. Perhaps we cannot do much better because of the protocol requirements.

Please wait for @slowli 's approval as well.

core/node/consistency_checker/src/lib.rs
```diff
@@ -855,3 +804,24 @@ async fn wait_for_l1_batch_with_metadata(
             .ok();
     }
 }
+
+// Get settlement layer based on ETH tx in the database.
+pub async fn get_db_settlement_mode(
+    db_pool: ConnectionPool<Core>,
```
slowli, line 810: Please use a `Connection` instead of a pool for more flexibility.

core/node/consistency_checker/src/lib.rs
```diff
@@ -855,3 +804,24 @@ async fn wait_for_l1_batch_with_metadata(
             .ok();
     }
 }
+
+// Get settlement layer based on ETH tx in the database.
+pub async fn get_db_settlement_mode(
```
slowli, line 809: Why is this a part of the consistency checker crate? It's not even used in it. Would it make sense to move to the call site (`node_framework`) and make private?

core/node/consistency_checker/src/lib.rs
```diff
@@ -380,52 +376,36 @@ impl ConsistencyChecker {
...
             contract: zksync_contracts::hyperchain_contract(),
             max_batches_to_recheck,
             sleep_interval: Self::DEFAULT_SLEEP_INTERVAL,
-            l1_chain_data,
-            gateway_chain_data,
+            chain_data,
+            settlement_layer,
             event_handler: Box::new(health_updater),
-            l1_data_mismatch_behavior: L1DataMismatchBehavior::Log,
+            l1_data_mismatch_behavior: L1DataMismatchBehavior::Bail,
```
slowli, line 400: With this change, is the `Log` variant (= the enum as a whole) still necessary? Previously, `Bail` was a test-only variant used to ensure that consistency failures lead to failing unit tests.

core/node/eth_watch/src/event_processors/gateway_migration.rs
slowli: If you say so 🙃

core/node/eth_watch/src/event_processors/gateway_migration.rs
```diff
@@ -0,0 +1,66 @@
...
+        Ok(events.len())
+    }
+
+    fn topic1(&self) -> Option<H256> {
+        None
+    }
+
+    fn topic2(&self) -> Option<H256> {
+        Some(H256::from_low_u64_be(self.l2chain_id.as_u64()))
+    }
```
slowli, line 57: Dumb question: What makes it impossible to fetch unrelated events here? AFAIU, we fetch events from the predetermined list of addresses (diamond proxy, STM, chain admin, etc.); are we sure that all these events will never have a matching topic2 (i.e., some smallish integer)? If there's no way to specify topic1 here (IIUC, there are 2 possible topics, right?), does it make sense to filter events by address and topic1 in `process_events`?

slowli (commented):
I feel I'm still nowhere close to covering the entire PR changeset, so here's partial feedback.

core/node/consistency_checker/src/lib.rs
Deniallugo: yes, it was one of the questions to you (as an author), why it was like this and could we remove it?

core/node/api_server/src/web3/tests/ws.rs
```diff
@@ -168,8 +168,14 @@ async fn test_ws_server(test: impl WsTest) {
     let contracts_config = ContractsConfig::for_tests();
     let web3_config = Web3JsonRpcConfig::for_tests();
     let genesis_config = GenesisConfig::for_tests();
-    let api_config =
-        InternalApiConfig::new(&web3_config, &contracts_config, &genesis_config, false);
+    let api_config = InternalApiConfig::new(
+        &web3_config,
+        &contracts_config.settlement_layer_specific_contracts(),
+        &contracts_config.l1_specific_contracts(),
+        &contracts_config.l2_contracts(),
```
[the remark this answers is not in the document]
Deniallugo: Yes, the main reason is an ability to remove `settlement_layer_specific_contracts` right after v27 deployed everywhere. Hopefully in v29 we can go further and remove even more contracts

core/lib/dal/src/server_notifications.rs
```diff
@@ -0,0 +1,69 @@
...
+        .await?;
+
+        Ok(())
+    }
+
+    pub async fn get_last_notification_by_topics(
+        &mut self,
+        topics: &[H256],
+    ) -> DalResult<Option<ServerNotification>> {
+        let topics: Vec<Vec<u8>> = topics.iter().map(|a| a.as_bytes().to_vec()).collect();
```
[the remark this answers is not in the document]
Deniallugo: query wants bytes as `Vec<u8>`
expected `&[Vec<u8>]`, found `&Vec<&[u8]>`

core/node/eth_watch/src/event_processors/gateway_migration.rs
Deniallugo: Actually i agree with you and i'm not sure i like the whole processing in eth watcher. but making a big change there seems inappropriate now.

Deniallugo: Made some refactoring for avoiding the situation you are describing. but seems we need to revisit the whole system related to events

core/bin/external_node/src/config/mod.rs
```diff
@@ -119,18 +122,20 @@ pub(crate) struct RemoteENConfig {
     pub l1_shared_bridge_proxy_addr: Option<Address>,
     /// Contract address that serves as a shared bridge on L2.
     /// It is expected that `L2SharedBridge` is used before gateway upgrade, and `L2AssetRouter` is used after.
-    pub l2_shared_bridge_addr: Option<Address>,
+    pub l2_shared_bridge_addr: Address,
```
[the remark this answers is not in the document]
Deniallugo: Actually deserialization here makes no sense. 
We can get it only from api. Removed deserialize

core/node/consistency_checker/src/lib.rs
slowli: Ah OK. I've explained why (for unit tests), so it should be safe to remove now.

core/node/genesis/src/lib.rs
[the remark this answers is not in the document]
slowli: Do you mean that the previous code was incorrect? If so, when should `evm_emulator_hash` become mandatory?

core/node/node_framework/src/implementations/layers/base_token/base_token_ratio_persister.rs
```diff
@@ -93,7 +87,11 @@ impl WiringLayer for BaseTokenRatioPersisterLayer {
 
                 let signing_client = PKSigningClient::new_raw(
                     tms_private_key.clone(),
-                    self.contracts_config.diamond_proxy_addr,
+                    input
+                        .l1_contracts_resource
+                        .0
+                        .chain_contracts_config
```
slowli, line 93: Nit: You can extract a reference to this config to a local var to not repeat access to it in 3 places.

slowli (approved):
Haven't found anything major, but I still feel that I haven't covered the entire changeset.

core/lib/eth_client/src/contracts_loader.rs
```diff
@@ -0,0 +1,156 @@
...
+    ethabi::{Contract, Token},
+    protocol_version::ProtocolSemanticVersion,
+    settlement::SettlementLayer,
+    Address, L2ChainId, SLChainId, U256,
+};
+
+use crate::{CallFunctionArgs, ContractCallError, EthInterface};
+
+pub async fn get_diamond_proxy_contract(
+    sl_client: &dyn EthInterface,
```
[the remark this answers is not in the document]
Deniallugo: I tried it only as an additional trait and seems because of async_trait i had a lot of problem. 
Not sure that adding it directly to `EthInterface` is a good idea. It will  be really complex trait

core/lib/web3_decl/src/namespaces/zks.rs
Deniallugo: I agree, that we can improve, this particular part. 
But probably we need to slightly restructure the initial set of methods and make the one method with all necessary contracts, that will deprecate all this small methods. 

Let's plan it and implement as the next step

core/node/genesis/src/lib.rs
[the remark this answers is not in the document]
Deniallugo: I'd rather make it necessary after v27 is everywhere, because now we can't make a proper genesis.

core/lib/web3_decl/src/namespaces/zks.rs
Deniallugo: PLA-1105

slowli (approved):

PR #3664, opened 2025-03-03
feat: Add S3 implementation for object_store

yorik:
## What ❔

Add S3 implementation for object_store.

## Why ❔

To remove hard dependency on GCS, add possibility to use any S3-like storage for snapshots and provers.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist

ref ZKD-2415


core/lib/object_store/src/s3.rs
```diff
@@ -0,0 +1,225 @@
...
+    bucket_prefix: String,
+    client: Client,
+}
+
+impl fmt::Debug for S3Store {
+    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+        formatter
+            .debug_struct("S3Store")
+            .field("bucket_prefix", &self.bucket_prefix)
+            // Skip `client` as its representation may contain sensitive info
```
slowli, line 26: Nit: Is `endpoint` potentially sensitive as well? Otherwise, it could contain useful debug info.

core/lib/object_store/src/s3.rs
```diff
@@ -0,0 +1,225 @@
...
+        bucket_prefix: String,
+        endpoint: Option<String>,
+        region: Option<String>,
+    ) -> Result<Self, ObjectStoreError> {
+        let region_provider = RegionProviderChain::first_try(region.map(Region::new))
+            .or_default_provider()
+            .or_else(Region::new("auto"));
+        let mut sdk_config = Self::get_client_config(auth_mode).region(region_provider);
+        if let Some(endpoint) = endpoint.clone() {
+            tracing::info!(endpoint=%endpoint, "using S3 endpoint defined in storage config");
```
slowli, line 54: Nit: `endpoint=%endpoint` can be shortened to just `%endpoint`.

core/lib/object_store/src/s3.rs
```diff
@@ -0,0 +1,225 @@
...
+                    }
+                } else {
+                    ObjectStoreError::Other {
+                        source: err.into(),
+                        is_retriable: true,
+                    }
+                }
+            }
+            SdkError::ServiceError(http_response) => {
+                let code = http_response.raw().status().as_u16();
```
slowli, line 135: Nit: It looks like handling of this and the previous variant can be extracted to a function in order to avoid code duplication.

core/lib/object_store/src/s3.rs
```diff
@@ -0,0 +1,225 @@
...
+        tracing::trace!(
+            "Storing data to S3 for key {filename} from bucket {}",
+            self.bucket_prefix
+        );
+
+        let length = i64::try_from(value.len()).context("Object is way too big")?;
+        self.client
+            .put_object()
+            .bucket(self.bucket_prefix.clone())
+            .key(&filename)
```
slowli, line 193: Nit: Since the API requires `impl Into<String>` here and `filename` is not used below, it would be more efficient to pass it by value: `key(filename)`.

core/lib/object_store/examples/object_store.rs
```diff
@@ -0,0 +1,73 @@
...
+        };
+        let snapshot_chunk = SnapshotStorageLogsChunk::<H256> {
+            storage_logs: vec![],
+        };
+
+        object_store
+            .put::<SnapshotStorageLogsChunk>(key, &snapshot_chunk)
+            .await?;
+        let result = object_store
+            .get::<SnapshotStorageLogsChunk>(key.into())
```
slowli, line 61: Nit: Useless conversion for `key`. Ditto below.

core/Cargo.toml
```diff
@@ -108,6 +108,9 @@ anyhow = "1"
 assert_matches = "1.5"
 async-trait = "0.1"
 async-recursion = "1"
+aws-config = { version = "1.1.7", features = ["behavior-version-latest"] }
+aws-runtime = "1.5.5"
+aws-sdk-s3 = "1.76.0"
```
slowli, line 113: Could you please check whether all default features are required for these crates to work? E.g., `aws-config` has the `sso` feature enabled by default bringing additional deps, which looks unused; at least the code compiles with

```toml
aws-config = { version = "1.1.7", default-features = false, features = ["behavior-version-latest"] }
```

slowli (commented):
Overall, looks very nice 👍 I guess we have no real alternatives if we want to embed S3 support directly into the config / every binary using `ObjectStoreFactory`. (Otherwise, it could be implemented as a a custom `ObjectStore` in a new crate.)

core/Cargo.toml
yorik: Perfect idea, removed 2 dependencies.

core/lib/object_store/examples/object_store.rs
yorik: Done

core/lib/object_store/src/s3.rs
yorik: Done

core/lib/object_store/src/s3.rs
yorik: I tried to do that, but unfortunately `http_response.raw()` returns private Response type, so I can't create a function which accept it. And `http_response` type is different in both cases.

core/lib/object_store/src/s3.rs
yorik: Done

core/lib/object_store/src/s3.rs
yorik: Added `endpoint` here. It isn't sensitive.

yorik:
> Overall, looks very nice 👍 I guess we have no real alternatives if we want to embed S3 support directly into the config / every binary using `ObjectStoreFactory`. (Otherwise, it could be implemented as a a custom `ObjectStore` in a new crate.)

Thank you so much for the review! I've addressed all the comments, except of the error handling, see the explanation there.

I think we'd like to have direct support of S3 for easier migration to it were needed: the first will be snapshot-creator and EN. All the prover stack could follow later.

Could you please take another look on the PR?

PR #3666, opened 2025-03-03
feat(zksync_server): adding --only-verify-config flag

tomg10:
## What ❔

A new flag that when set will stop server immediately after ensuring that all components can be built correctly with provided config.

## Why ❔

To make deploying config changes and new server versions 

## Is this a breaking change?
- [ ] Yes
- [X] No

## Operational changes
New experimental flag --only-verify-config for zksync_server binary

## Checklist


core/bin/zksync_server/src/main.rs
```diff
@@ -177,7 +180,13 @@ fn main() -> anyhow::Result<()> {
...
     }
 
-    node.build(opt.components.0)?.run(observability_guard)?;
+    let built_node = node.build(opt.components.0)?;
+
+    if opt.only_verify_config {
+        return Ok(());
+    }
+
+    built_node.run(observability_guard)?;
```
perekopskiy, line 189: Component wiring with some checks happens in `node::run`. In most cases it should be ok to wire to just check the config but there are layers that read from DB during wiring, e.g. `OutputHandlerLayer`. We can either leave it as is (do not wire), or refactor so wiring doesn't require DB access. I'm not sure what's better in this case, we want to validate configs as much as possible, but OTOH we want as little code as possible to be executed.

core/bin/zksync_server/src/main.rs
```diff
@@ -177,7 +180,13 @@ fn main() -> anyhow::Result<()> {
         return Ok(());
     }
 
-    node.build(opt.components.0)?.run(observability_guard)?;
+    let built_node = node.build(opt.components.0)?;
```
slowli, line 183: The task description says "after ensuring that **all** components can be built". If this is what we want, it would make sense to ignore `opt.components` and instead provide the full list of components (maybe, make the command-line arg [conflict with](https://docs.rs/clap/latest/clap/struct.Arg.html#method.conflicts_with) `components`.

core/bin/zksync_server/src/main.rs
slowli: Some layers [also access L1 during wiring](https://github.com/matter-labs/zksync-era/blob/8e24403190c4f7e306c88ccdb3c2f4a86e0e56fa/core/node/node_framework/src/implementations/layers/pk_signing_eth_client.rs#L76-L79) 🙃 IMO, it could be worth clarifying how we define a correct config; it not trivial whether it should include or exclude stuff like DBs and L1 client being accessible. FWIW, [config validation](https://github.com/matter-labs/zksync-era/blob/88e4ed3bf660ca565fc6f8a556b6965b542bb393/core/bin/zksync_server/src/main.rs#L130-L141) built into [the new config system PoC](https://github.com/matter-labs/zksync-era/compare/aov-pla-1068-configuration-system-poc) just checks that a config is deserializable.

core/bin/zksync_server/src/main.rs
```diff
@@ -177,7 +180,13 @@ fn main() -> anyhow::Result<()> {
...
+
+    if opt.only_verify_config {
+        return Ok(());
+    }
```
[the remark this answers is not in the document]
slowli: The "config-only" mode should probably be exclusive with the genesis mode (via `conflicts_with`), so their ordering shouldn't matter much. Regarding observability, if we do want to include wiring into config validation, then having logging enabled could make debugging errors easier.

core/bin/zksync_server/src/main.rs
tomg10: I don't think it's what we want, I incorrectly wrote "all", each env can have slightly different components, my CI script will first gather all components existing on given env and run this validator only for those components, imo it's much less error-prone

core/bin/zksync_server/src/main.rs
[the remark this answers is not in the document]
tomg10: +1, I'm adding conflicts with and changing so that observability guard is not installed as at least for now I don't think including wiring into config validation is a good idea

core/bin/zksync_server/src/main.rs
```diff
@@ -71,6 +71,9 @@ struct Cli {
     /// Now the node framework is used by default and this argument is left for backward compatibility.
     #[arg(long)]
     use_node_framework: bool,
+
+    #[arg(long)]
+    only_verify_config: bool,
```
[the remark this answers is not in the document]
tomg10: I don't think I yet fully understand the comment, specifically why it is misleading. what is checked apart from config?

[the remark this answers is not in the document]
popzxc: I guess it's because the application can exit for reasons different to configuration issues (e.g. if something goes wrong during wiring, for example; or if the some non-configuration files like contracts are missing).

Additionally, it doesn't really catch _all_ the configuration errors; since the configuration is loaded lazily, we will only catch the issues with components being passed.

I would call it `no_run` tbh, and have something like this as a doc-string:

```rust
/// Only compose the node with the provided list of the components and then exit.
/// Can be used to catch issues with configuration.
```

core/bin/zksync_server/src/main.rs
```diff
@@ -165,6 +170,12 @@ fn main() -> anyhow::Result<()> {
         secrets,
     )?;
 
+    if opt.only_verify_config {
+        node.build(opt.components.0)?;
+        tracing::info!("Config verification is successful!");
```
popzxc, line 175: Note that you use `tracing` before installing observability, so it has no effect.

core/bin/zksync_server/src/main.rs
```diff
@@ -165,6 +170,12 @@ fn main() -> anyhow::Result<()> {
...
+        return Ok(());
+    }
```
popzxc, line 177: Would it make sense to move it below `if opt.genesis` in a following manner?

```rust
let node = node.build(opt.components.0)?;

if opt.no_run {
  tracing::info!("Node composed successfully; exiting due to --no-run flag");
  return Ok(())
}

node.run(observability_guard)
```

core/bin/zksync_server/src/main.rs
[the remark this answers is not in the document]
tomg10: okay, now I see how it can be confusing, changing it to no_run

core/bin/zksync_server/src/main.rs
tomg10: makes sense for having observability installed

perekopskiy (approved):

PR #3682, opened 2025-03-06
feat(migration): Make universal client

Deniallugo:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


core/node/node_framework/src/implementations/layers/eth_watch.rs
```diff
@@ -114,40 +116,53 @@ impl WiringLayer for EthWatchLayer {
...
-                .diamond_proxy_addr,
-            input
-                .contracts_resource
-                .0
-                .ecosystem_contracts
-                .server_notifier_addr,
-            self.eth_watch_config.confirmations_for_eth_event,
-            self.chain_id,
-        ));
+        macro_rules! create_client {
```
slowli, line 119: This is unnecessary. You can extract a function

```rust
impl EthWatchLayer {
    fn create_client<Net: Network>(
        &self,
        client: Box<DynClient<Net>>,
        l1_contracts: &L1ChainContractsResource,
        l1ecosystem_contracts_resource: &L1EcosystemContractsResource,
    ) -> EthHttpQueryClient<Net> {
        // ...
    }
}
```

...and use it instead of this macro and above to create `l1_client`.

core/node/node_framework/src/implementations/resources/eth_interface.rs
```diff
@@ -14,11 +14,26 @@ impl Resource for EthInterfaceResource {
...
+    fn from(value: UniversalClient) -> Self {
+        match value {
+            UniversalClient::L1(client) => Box::new(client),
+            UniversalClient::L2(client) => Box::new(client),
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct GatewayEthInterfaceResourceUniversalClient(pub UniversalClient);
```
slowli, line 32: Bikeshedding: Why not just `UniversalClientResource`?

core/node/node_framework/src/implementations/layers/pk_signing_eth_client.rs
```diff
@@ -99,31 +100,31 @@ impl WiringLayer for PKSigningEthClientLayer {
...
-                gateway_client.clone(),
-            );
-            Some(BoundEthInterfaceForL2Resource(Box::new(
-                signing_client_for_blobs,
-            )))
-        } else {
-            None
-        };
+        let signing_client_for_l2_gateway =
+            if let UniversalClient::L2(gateway_client) = input.gateway_client.0 {
```
slowli, line 104: Nit: it's safer to use `match` for cases like this in case further variants are added to the enumeration in the future; it'd enumerate all variants explicitly.

core/node/node_sync/src/validate_chain_ids_task.rs
```diff
@@ -32,23 +32,23 @@ impl ValidateChainIdsTask {
...
         }
     }
 
     async fn check_client(
-        l1_client: Option<Box<DynClient<L1>>>,
+        gateway_client: Option<Box<dyn EthInterface>>,
         expected: Option<SLChainId>,
     ) -> anyhow::Result<()> {
-        let (Some(l1_client), Some(expected)) = (l1_client, expected) else {
+        let (Some(l1_client), Some(expected)) = (gateway_client, expected) else {
```
slowli, line 51: Bikeshedding: The old arg / variable naming was outdated already, but I don't understand the new one either. I'd name both the arg and the binding simply `client` given that `check_client()` is called both with L1 client and gateway clients.

slowli (commented):
The approach looks OK-ish, but IMO the overall client logic in `eth_watch` is overly complicated bordering unmaintainable. E.g., AFAIU, the only reason [`GetLogsClient`](https://github.com/matter-labs/zksync-era/blob/7a2ccacb86efe21e14bd07106c1a9f44a70ef069/core/node/eth_watch/src/client.rs#L539-L544) is defined is that [`EthInterface::logs()`](https://github.com/matter-labs/zksync-era/blob/7a2ccacb86efe21e14bd07106c1a9f44a70ef069/core/lib/eth_client/src/lib.rs#L142-L143) returns L1 logs (i.e., ones w/o `l1_batch_number`), while `eth_watch` requires `l1_batch_number` [for one of the processors](https://github.com/matter-labs/zksync-era/blob/7a2ccacb86efe21e14bd07106c1a9f44a70ef069/core/node/eth_watch/src/event_processors/appended_chain_batch_root.rs#L73-L77). I feel it would be easier to make `EthInterface` / `eth_client` crate aware of L2 by returning L2 logs from `EthInterface::logs()`; especially given that `eth_client` [is already aware](https://github.com/matter-labs/zksync-era/blob/7a2ccacb86efe21e14bd07106c1a9f44a70ef069/core/lib/eth_client/src/clients/http/query.rs#L386) of *some* distinctions between L1 and L2 clients. This in turn would allow using `Box<dyn EthInterface>` in `EthHttpQueryClient` etc.

core/node/node_framework/src/implementations/layers/eth_watch.rs
```diff
@@ -114,40 +116,53 @@ impl WiringLayer for EthWatchLayer {
...
+            ($client: expr) => {
+                Box::new(EthHttpQueryClient::new(
```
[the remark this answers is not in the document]
Deniallugo: i'm slightly afraid of the number of changes.

PR #3688, opened 2025-03-07
fix(contract-verifier): Correctly process partial verification for EVM contracts

popzxc:
I've discovered that partial verification doesn't work correctly for EVM contracts, because we expect creation bytecode to be equal for compiled and deployed contracts. This is too strict, because it's OK for them to be different in metadata hash only.

I didn't add any tests for this (it was somewhat tricky), but checked manually that the issue was resolved after that.

### DIFF
--- a/core/lib/contract_verifier/src/lib.rs
+++ b/core/lib/contract_verifier/src/lib.rs
@@ -259,18 +259,8 @@ impl ContractVerifier {
         let bytecode_marker = BytecodeMarker::new(deployed_contract.bytecode_hash)
             .context("unknown bytecode kind")?;
         let artifacts = self.compile(request.req.clone(), bytecode_marker).await?;
-        let identifier =
+        let compiled_identifier =
             ContractIdentifier::from_bytecode(bytecode_marker, artifacts.deployed_bytecode());
-        let constructor_args = match bytecode_marker {
-            BytecodeMarker::EraVm => self
-                .decode_era_vm_constructor_args(&deployed_contract, request.req.contract_address)?,
-            BytecodeMarker::Evm => Self::decode_evm_constructor_args(
-                request.id,
-                &deployed_contract,
-                &artifacts.bytecode,
-                &identifier,
-            )?,
-        };
 
         let deployed_bytecode = match bytecode_marker {
             BytecodeMarker::EraVm => deployed_contract.bytecode.as_slice(),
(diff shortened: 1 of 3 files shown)


core/lib/contract_verifier/src/lib.rs
```diff
@@ -556,7 +557,33 @@ impl ContractVerifier {
...
+            // Note that metadata hash may still be different; this is checked by other part of the code.
+            let constructor_args_with_metadata = calldata
+                .strip_prefix(creation_bytecode_without_metadata)
+                .ok_or("Creation bytecode is different")?;
+
+            // Skip metadata to get to the constructor arguments.
+            if constructor_args_with_metadata.len() < metadata_len {
+                return Err("Provided bytecode has no metadata");
+            }
+            Ok(&constructor_args_with_metadata[metadata_len..])
```
slowli, line 584: I'm not sure I understand how this works. If metadata can be different from one in `creation_bytecode`, what's the guarantee that it has the same length (if it's CBOR metadata)?

core/lib/contract_verifier/src/lib.rs
```diff
@@ -556,7 +557,33 @@ impl ContractVerifier {
         request_id: usize,
         contract: &DeployedContractData,
         creation_bytecode: &[u8],
+        contract_identifier: &ContractIdentifier,
     ) -> Result<ConstructorArgs, ContractVerifierError> {
+        fn extract_arguments<'a>(
```
slowli, line 562: Nit: IMO, it'd be more readable to place this function on the `decode_evm_constructor_args()` level.

core/lib/contract_verifier/src/lib.rs
popzxc: The probability of lengths being different is negligibly low, IMHO (for that we need either different contract hash format or pre-release solc version -- I've never seen either of those in practice).

But sure, just in case we can protect the implementation against that.

core/lib/contract_verifier/src/lib.rs
popzxc: I would prefer to keep it here -- it feels too low-level (including `&'static str` as an `Err` type) to exist outside single scope.

perekopskiy (dismissed):

perekopskiy (approved):

PR #3690, opened 2025-03-10
feat(zkstack-cli): Adding support for v27 upgrade

mm-zk:
## What ❔

Adding the ability do execute v27 upgrade using zkstack cli.

## Is this a breaking change?
- [ ] Yes
- [X] No

## Operational changes
To run:

```
zkstack e v27-upgrade --ecosystem-upgrade-stage no-governance-prepare
```


deniallugo-ml:

PR #3693, opened 2025-03-10
feat(gateway): Support gateway migration on en

Deniallugo:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


core/node/consistency_checker/src/lib.rs
```diff
@@ -829,3 +806,34 @@ async fn wait_for_l1_batch_with_metadata(
             .ok();
     }
 }
+
+// Get settlement layer based on ETH tx, all eth txs should be presented on settlement layer. what is the best place for this function?
+pub async fn get_settlement_mode(
```
perekopskiy, line 811: we save chain id for eth txs (or if chain_id is missing means it's old server and thus it's l1), I believe it should be used to determine settlement mode.

core/node/consistency_checker/src/lib.rs
```diff
@@ -829,3 +806,34 @@ async fn wait_for_l1_batch_with_metadata(
...
+        .eth_sender_dal()
+        .get_last_eth_tx()
+        .await?;
+
+    if let Some(tx_hash) = tx_hash {
+        let tx_status = eth_client.get_tx_status(tx_hash).await?;
+        if tx_status.is_some() {
+            return Ok(SettlementMode::SettlesToL1);
+        } else {
+            // The tx has not been found on l1, we have to check on l2
```
perekopskiy, line 828: nit: while techinically gateway is L2, I'd prefer to not call it this way and say "gateway" explicitly, otherwise it will be confusing to distinguish between and normal chains that we also call L2s

core/node/consistency_checker/src/lib.rs
Deniallugo: oh, nice thank you! didn't know about it

PR #3696, opened 2025-03-11
chore: updated contracts

mm-zk:
## What ❔

* updated contracts to the newest version of v27

## Why ❔

* small changes related to upgrade scripts

## Is this a breaking change?
- [ ] Yes
- [X] No


deniallugo-ml:

PR #3697, opened 2025-03-11
feat(zkstack): Allow skipping build for contracts and server during ecosystem init

popzxc:
- Fix `skip-contract-compilation-override`: it wasn't possible to set it, because it's a boolean flag for which argument is _always_ missing, so it wasn't working.
- Don't run `yarn install` if `skip-contract-compilation-override` is passed.
- Add `--server-command` param to pass custom command instead of `cargo run` to run server.

Sample command:

```
zkstack e init --dev --ignore-prerequisites --skip-contract-compilation-override --server-command /path/to/zksync-era/core/target/release/zksync_server -v
```

(`-v` is optional, obviously)


zkstack_cli/crates/zkstack/src/commands/ecosystem/init.rs
```diff
@@ -111,8 +111,9 @@ async fn init_ecosystem(
     initial_deployment_config: &InitialDeploymentConfig,
 ) -> anyhow::Result<ContractsConfig> {
     let spinner = Spinner::new(MSG_INTALLING_DEPS_SPINNER);
-    install_yarn_dependencies(shell, &ecosystem_config.link_to_code)?;
+    println!("init args: {init_args:?}");
```
Deniallugo, line 114: println?

zkstack_cli/crates/common/src/server.rs
```diff
@@ -4,9 +4,16 @@ use xshell::{cmd, Shell};
...
+/// Default command to run the server; will use `cargo` to build it.
+const DEFAULT_SERVER_COMMAND: &str =
+    "cargo run --manifest-path ./core/Cargo.toml --release --bin zksync_server";
+
 /// Allows to perform server operations.
 #[derive(Debug)]
 pub struct Server {
+    /// Command to run the server, could be a path to the binary.
+    /// If not set, the server will be built using cargo.
+    server_command: Option<String>,
```
Deniallugo, line 16: TBH, i'm a bit of afraid when we have such flexibility. doesn't look useful anywhere except probably ci

zkstack_cli/crates/zkstack/src/commands/ecosystem/init.rs
popzxc: Forgot to remove

zkstack_cli/crates/zkstack/src/commands/ecosystem/gateway_upgrade.rs
```diff
@@ -640,6 +640,7 @@ async fn no_governance_stage_3(
 
     let DBNames { server_name, .. } = generate_db_names(&chain_config);
     let args = GenesisArgsFinal {
+        server_command: None,
```
Raid5594, line 643: Do we want to actually pass the command here?

Deniallugo (approved):

artmakh (approved):

PR #3698, opened 2025-03-11
feat(storage): rocksdb storage extension for zkos

perekopskiy:
## What ❔

- `KeepUpdatedTask` to catch up with postgres indefinitely
- added `rocksdb_with_memory` which inits `CommonStorage::RocksdbWithMemory`

## Why ❔

The functionality will be needed for zk os state keeper.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Operational changes

No

## Checklist

### DIFF
--- a/core/lib/state/src/catchup.rs
+++ b/core/lib/state/src/catchup.rs
@@ -186,7 +186,7 @@ impl AsyncCatchupTask {
             .context("Failed to catch up RocksDB to Postgres")?;
         drop(connection);
 
-        let rocksdb = rocksdb.map(|db| db.into_rocksdb());
+        let rocksdb = rocksdb.map(RocksdbStorage::into_rocksdb);
         if let Some(rocksdb) = &rocksdb {
             self.db_sender.send_replace(Some(rocksdb.clone()));
         } else {
--- a/core/lib/state/src/catchup.rs
+++ b/core/lib/state/src/catchup.rs
@@ -266,7 +266,12 @@ mod tests {
         let initial_state = rocksdb_cell.ensure_initialized().await.unwrap();
         assert_eq!(initial_state.l1_batch_number, None);
 
+        let started_at = Instant::now();
         loop {
+            if started_at.elapsed() > Duration::from_secs(10) {
+                panic!("Timeout waiting for init");
+            }
+
             if let Some(db) = rocksdb_cell.get() {
                 let l1_batch_number = RocksdbStorageBuilder::from_rocksdb(db)
                     .l1_batch_number()
@@ -311,7 +316,12 @@ mod tests {
 
             let rocksdb = rocksdb_cell.get().unwrap();
             let builder = RocksdbStorageBuilder::from_rocksdb(rocksdb);
+            let started_at = Instant::now();
             loop {
+                if started_at.elapsed() > Duration::from_secs(10) {
+                    panic!("Timeout waiting for catch up");
+                }
+
                 if builder.l1_batch_number().await == Some(L1BatchNumber(3)) {
                     break;
                 } else {
--- a/core/lib/state/src/catchup.rs
+++ b/core/lib/state/src/catchup.rs
@@ -210,14 +210,15 @@ impl KeepUpdatedTask {
         const SLEEP_INTERVAL: Duration = Duration::from_millis(100);
 
         let db: RocksDB<StateKeeperColumnFamily> = tokio::select! {
-            res = self.db.wait_for(Option::is_some) => match res {
+            res = self.db.wait_for(Option::is_some) => if let Ok(db) = res {
                 // `unwrap()` is safe by construction
-                Ok(db) => db.clone().unwrap(),
-                // Interrupted
-                Err(_) => return Ok(()),
+                db.clone().unwrap()
+            } else {
+                tracing::info!("Stop signal received, shutting down rocksdb updater task");
+                return Ok(());
             },
             _ = stop_receiver.changed() => {
-                // Interrupted
+                tracing::info!("Stop signal received, shutting down rocksdb updater task");
                 return Ok(());
             }
         };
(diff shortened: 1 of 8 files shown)


core/lib/state/src/storage_factory/mod.rs
```diff
@@ -119,6 +119,58 @@ impl CommonStorage<'static> {
...
+        l1_batch_number: L1BatchNumber,
+    ) -> anyhow::Result<(Self, L1BatchNumber)> {
+        tracing::debug!("Catching up RocksDB synchronously");
+        let rocksdb_builder = RocksdbStorageBuilder::from_rocksdb(rocksdb);
+
+        let rocksdb = rocksdb_builder.skip_synchronize();
+        let rocksdb_l1_batch_number = rocksdb
+            .l1_batch_number()
+            .await
+            .ok_or_else(|| anyhow::anyhow!("No L1 batches available in Postgres"))?;
```
slowli, line 144: - Bikeshedding: It could make sense at this point to accept `RocksdbStorage` as an argument and allow to create it from `RocksDB<_>` (e.g., via `From<_>`). Initially, the builder pattern was used to prevent uninitialized `RocksdbStorage`, but it's evidently no longer working with these changes. This would allow to not return `L1BatchNumber` and not to add `skip_synchronize()`.
- Nit: It's possible to use `.context(_)` instead of `.ok_or_else()` here.

core/lib/state/src/rocksdb/mod.rs
```diff
@@ -224,6 +224,10 @@ impl RocksdbStorageBuilder {
         }
     }
 
+    pub(crate) fn skip_synchronize(self) -> RocksdbStorage {
```
slowli, line 227: There's already a `doc(hidden) build_unchecked()` method below with the same semantics; you can reuse it. Weirdly enough, it's not used anywhere in the codebase; maybe, it was used in some tests that were removed.

core/lib/state/src/storage_factory/rocksdb_with_memory.rs
```diff
@@ -73,3 +73,103 @@ impl ReadStorage for RocksdbWithMemory {
...
+        assert!((to_l1_batch.0 as usize) < (first_diff_number.0 as usize) + self.diffs.len());
+
+        let to_skip = from_l1_batch.0 - first_diff_number.0;
+        let to_take = to_l1_batch.0 - from_l1_batch.0 + 1;
+
+        self.diffs
+            .iter()
+            .skip(to_skip as usize)
+            .take(to_take as usize)
+            .cloned()
```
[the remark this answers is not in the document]
perekopskiy: It's not, but the only use case, i.e. constructing `RocksDbWithMemory`, needs them cloned for current implementation as we work with `OwnedStorage` in state keeper. Do you think it makes sense to return refs here and do cloning when constructing?
Wrt to using non-static storage for state keeper, the only reason that prevents it is zk_ee's `ReadStorage` trait that requires `'static` and I'm not sure why. Overall this part of the system (`ReadStorageTree`, `ReadStorage` trait) will be reworked so that `run_batch` or its analogue will require only `read` and maybe `tree_index` methods and I think we can get back to static requirement after it's done.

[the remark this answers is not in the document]
slowli: > Do you think it makes sense to return refs here and do cloning when constructing?

No, it would look like over-engineering in this case. If the only use is when constructing `RocksdbWithMemory`, can't the method visibility be reduced to something like `pub(crate)`? In this case, gearing the output towards a specific use case is warranted.

Regarding the `'static` boundary: It could theoretically be dealt with by wrapping `BatchDiff` in `Rc` / `Arc`, so that its cloning in `range()` is cheap. Doesn't matter much for a PoC.

core/lib/state/src/catchup.rs
```diff
@@ -174,15 +188,60 @@ impl AsyncCatchupTask {
...
+            _ = stop_receiver.changed() => {
+                // Interrupted
+                return Ok(());
+            }
+        };
+
+        let mut rocksdb_builder = RocksdbStorageBuilder::from_rocksdb(db);
+
+        loop {
+            let mut connection = self.pool.connection_tagged("rocksdb_updater_task").await?;
```
slowli, line 228: Nit: Please drop `connection` before the sleep, just so it's not forgotten if the update interval needs to be increased for whatever reason.

core/lib/state/src/catchup.rs
```diff
@@ -174,15 +188,60 @@ impl AsyncCatchupTask {
...
+
+impl KeepUpdatedTask {
+    pub async fn run(mut self, mut stop_receiver: watch::Receiver<bool>) -> anyhow::Result<()> {
+        const SLEEP_INTERVAL: Duration = Duration::from_millis(100);
+
+        let db: RocksDB<StateKeeperColumnFamily> = tokio::select! {
+            res = self.db.wait_for(Option::is_some) => match res {
+                // `unwrap()` is safe by construction
+                Ok(db) => db.clone().unwrap(),
+                // Interrupted
```
slowli, line 216: Nit: May make sense to log interruption here and below (in another `select!` clause and when processing `RocksdbSyncError::Interrupted`).

core/lib/state/src/storage_factory/rocksdb_with_memory.rs
```diff
@@ -73,3 +76,110 @@ impl ReadStorage for RocksdbWithMemory {
...
+        assert!(from_l1_batch >= first_diff_number);
+        assert!((to_l1_batch.0 as usize) < (first_diff_number.0 as usize) + self.diffs.len());
+
+        let to_skip = from_l1_batch.0 - first_diff_number.0;
+        let to_take = to_l1_batch.0 - from_l1_batch.0 + 1;
+
+        self.diffs
+            .iter()
+            .skip(to_skip as usize)
+            .take(to_take as usize)
```
slowli, line 145: Nit: You can use [`VecDeque::range()`](https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.range) to be more idiomatic.

slowli (approved):

PR #3706, opened 2025-03-12
feat(eth-watch): split heavy get logs requests if 503

perekopskiy:
## What ❔

split heavy get logs requests if got error 503

## Why ❔

We use proxyd that proxies mostly vanila reth nodes in our setup, some get_logs requests with big block range take a lot of time so they can rejected in proxyd by timeout, in this case 503 is returned. We should split such requests into smaller so they can be processed.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Operational changes
No

## Checklist

### DIFF
--- a/core/node/eth_watch/src/client.rs
+++ b/core/node/eth_watch/src/client.rs
@@ -90,7 +90,7 @@ const TOO_MANY_RESULTS_ALCHEMY: &str = "response size exceeded";
 const TOO_MANY_RESULTS_RETH: &str = "length limit exceeded";
 const TOO_BIG_RANGE_RETH: &str = "query exceeds max block range";
 const TOO_MANY_RESULTS_CHAINSTACK: &str = "range limit exceeded";
-const REQUEST_REJECTED_503: &str = "Request rejected `503`";
+const REQUEST_REJECTED_503: &str = "503";
 
 /// Implementation of [`EthClient`] based on HTTP JSON-RPC.
 #[derive(Debug, Clone)]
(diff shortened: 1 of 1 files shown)


core/node/eth_watch/src/client.rs
```diff
@@ -90,6 +90,7 @@ const TOO_MANY_RESULTS_ALCHEMY: &str = "response size exceeded";
 const TOO_MANY_RESULTS_RETH: &str = "length limit exceeded";
 const TOO_BIG_RANGE_RETH: &str = "query exceeds max block range";
 const TOO_MANY_RESULTS_CHAINSTACK: &str = "range limit exceeded";
+const REQUEST_REJECTED_503: &str = "Request rejected `503`";
```
Deniallugo, line 93: probably we could keep only 503 here

core/node/eth_watch/src/client.rs
```diff
@@ -222,6 +223,7 @@ where
                 || err_message.contains(TOO_MANY_RESULTS_RETH)
                 || err_message.contains(TOO_BIG_RANGE_RETH)
                 || err_message.contains(TOO_MANY_RESULTS_CHAINSTACK)
+                || err_message.contains(REQUEST_REJECTED_503)
```
sanekmelnikov, line 226: Can be mistaken for something like "gas fee is too low 100000**503** vs ..." or something else that just contains "503" as a part of integer in message?
Just wondering whether it would be better to check `err_code == Some(503)` instead?

perekopskiy: Valid, unfortunately we don't have err_code returned, maybe it's possible to change client impl to have it but unlikely, I think it makes sense to return "Request rejected `503`"

perekopskiy: yes, jsonprsee (lib we use) doesn't expose error code, I will revert to "Request rejected 503"

sanekmelnikov (approved):

core/node/eth_watch/src/client.rs
sanekmelnikov: Oh, didn't see Danil's comment :)
And what would happen if it's rejected because of infra issue?
It will just split the range until eventually hits `from_number >= mid` check and returns an error, right?

perekopskiy: yes

PR #3715, opened 2025-03-14
fix: make secret optionnal when da client object store is selected

thomas-nguy:
## What ❔

Currently if `ObjectStore`  is selected as DA client, it requires secret to be non empty, however there is no secret when using object store.

To bypass this bug, we can add a dummy secret in secrets.yaml otherwise the server panic such as :

```
da:
  celestia:
    private_key: 0xaa.....

```

This PR solves the issue

## Why ❔

Bug on DA client which can be only fixed with a hack in configuration. This is not natural and need to be fixed

## Is this a breaking change?
- [ ] Yes
- [x] No

## Operational changes

No

## Checklist


dimazhornyk (approved):
Looks good, but we can also do the same thing in EN's `node_builder`

thomas-nguy:
sure. fixed the EN as well

dimazhornyk (approved):

PR #3719, opened 2025-03-17
feat(main): Eigenda add custom quorum params

juan518munoz:
## What ❔

Add `custom_quorum_numbers` parameter for the EigenDA client and explain the da client configuration file structure for ease of setup.

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


core/lib/env_config/src/da_client.rs
```diff
@@ -71,6 +71,13 @@ pub fn da_client_config_from_env(prefix: &str) -> anyhow::Result<DAClientConfig>
                 )),
                 _ => anyhow::bail!("Unknown Eigen points type"),
             },
+            custom_quorum_numbers: match env::var(format!("{}CUSTOM_QUORUM_NUMBERS", prefix)) {
+                Ok(numbers) => numbers
+                    .split(',')
+                    .map(|s| s.parse().map_err(|e: ParseIntError| anyhow::anyhow!(e)))
+                    .collect::<anyhow::Result<Vec<_>>>()?,
+                Err(_) => vec![],
```
samlaf, line 79: You could also check that the numbers are <= 255, since quorums are uint8s onchain: https://github.com/Layr-Labs/eigenlayer-middleware/blob/cd43f0a408e0edb8d87d8c964e4cc8f1a39fe058/src/RegistryCoordinator.sol#L248

samlaf (commented):
Looks good from my side.

core/lib/env_config/src/da_client.rs
juanbono: Ohh, we will add that check to the client lib, to handle it there.

juan518munoz: It may be better if we run this checks inside of the client crate instead of leaving it to the user to verify correct usage, we've made [a PR](https://github.com/Layr-Labs/eigenda-client-rs/pull/23) making the necessary changes.

samlaf: nice thanks; sorry should have flagged this earlier in the clients PR

core/node/da_clients/src/eigen/README.md
```diff
@@ -3,3 +3,56 @@
...
+First you need to set the `use_dummy_inclusion_data` field in the file `etc/env/file_based/general.yaml` to `true`. This
+is a pending solution until the necessary contract changes are done (M1 milestone).
+
+```yaml
+da_dispatcher:
+  use_dummy_inclusion_data: true
+```
+
+The client can be set up by modifying the field `da_client` of the file `etc/env/file_based/overrides/validium.yaml`.
+This are the fields that can be modified:
```
samlaf, line 18:
```suggestion
These are the fields that can be modified:
```

core/node/da_clients/src/eigen/README.md
```diff
@@ -3,3 +3,56 @@
...
+
+So, for example, a client setup that uses the holesky EigenDA client would look like this:
+
+```yaml
+eigen:
+  disperser_rpc: https://disperser-holesky.eigenda.xyz:443
+  settlement_layer_confirmation_depth: 0
+  eigenda_eth_rpc: https://ethereum-holesky-rpc.publicnode.com
+  eigenda_svc_manager_address: 0xD4A7E1Bd8015057293f0D0A557088c286942e84b
+  wait_for_finalization: false
```
samlaf, line 42: can we put `settlement_layer_confirmation_depth` and `wait_for_finalization` next to each other? Also I forget, how do those 2 interact in your implementation? In proxy we merged them into a single flag `confirmation_depth` that accepts either a number or the `finalized` string, like most ethereum json-rpc methods accept.

samlaf: Also can we use safe defaults here, like `confirmation_depth` set to 6?

core/node/da_clients/src/eigen/README.md
```diff
@@ -3,3 +3,56 @@
...
+  authenticated: false
```
samlaf, line 43:
```suggestion
  authenticated: true
```

samlaf: better to set defaults that most people should use.

core/lib/protobuf_config/src/da_client.rs
```diff
@@ -160,6 +165,12 @@ impl ProtoRepr for proto::DataAvailabilityClient {
                         g2_url: Some(g2_url.clone()),
                     }),
                 }),
+                // We need to cast as u32 because proto doesn't support u8
+                custom_quorum_numbers: config
+                    .custom_quorum_numbers
+                    .iter()
+                    .map(|x| *x as u32)
```
samlaf, line 172: not sure what proto means here, not familiar with zksync config, but could you use `bytes` instead, where each byte is interpreted as a (uint8) quorum number?

core/Cargo.toml
```diff
@@ -230,7 +230,7 @@ tonic = { version = "0.11.0", default-features = false }
 pbjson-types = "0.6.0"
 
 # Eigen
-rust-eigenda-client = "0.1.1"
+rust-eigenda-client = "0.1.2"
```
samlaf, line 233: This will need to be updated with the new client version which has custom quorums right? after https://github.com/Layr-Labs/eigenda-client-rs/pull/23 is merged

core/node/da_clients/src/eigen/README.md
juan518munoz: In our case one does not imply the other, settlement_layer_confirmation_depth is used when verifying a batch, while wait_for_finalization is used when trying to retrieve inclusion data in our client. It's true tho that if settlement_layer_confirmation_depth is set, by the time the wait_for_finalization check is reached, the blob will be already finalized.

core/Cargo.toml
juan518munoz: Correct, it's now set to `"0.1.3"`

core/lib/protobuf_config/src/da_client.rs
juan518munoz: It would be possible to make such change, but we'd lose human readability when setting up the config file (`validium.yaml`).

Instead of having it like this:
```yaml
    custom_quorum_numbers: 2,3
```

It would need to be in base64 encoding in order to be parseable as `bytes`:
```yaml
    custom_quorum_numbers: "AgM="
```

core/node/da_clients/src/eigen/README.md
```diff
@@ -3,3 +3,56 @@
...
+If using `authenticated` dispersal, you also need to modify `etc/env/file_based/secrets.yaml` to include the private key
+of the account that will be used. You need to add the following field:
+
+```yaml
+da:
+  eigen:
+    private_key: <PRIVATE_KEY>
+```
+
+Note: the private key should be in hex format, without the `0x` prefix.
```
dimazhornyk, line 58: nice to have: maybe not it the scope of this PR, but would be great to be "0x"-agnostic, and automatically remove it if it is provided

core/node/da_clients/src/eigen/README.md
```diff
@@ -3,3 +3,56 @@
...
+eigen:
+  disperser_rpc: https://disperser-holesky.eigenda.xyz:443
+  settlement_layer_confirmation_depth: 0
+  eigenda_eth_rpc: https://ethereum-holesky-rpc.publicnode.com
+  eigenda_svc_manager_address: 0xD4A7E1Bd8015057293f0D0A557088c286942e84b
+  wait_for_finalization: false
+  authenticated: false
+  points_source_url:
+    g1_url: https://github.com/Layr-Labs/eigenda-proxy/raw/2fd70b99ef5bf137d7bbca3461cf9e1f2c899451/resources/g1.point
+    g2_url: https://github.com/Layr-Labs/eigenda-proxy/raw/2fd70b99ef5bf137d7bbca3461cf9e1f2c899451/resources/g2.point.powerOf2
```
dimazhornyk, line 46: is the example intentionally missing `custom_quorum_numbers`?

dimazhornyk (approved):

core/node/da_clients/src/eigen/README.md
juan518munoz: Yes, it is [following samlaf's advice on using most common settings as defaults](https://github.com/matter-labs/zksync-era/pull/3719#discussion_r2000158387), nonetheless, I have added the field (commented out) to show its input format.

core/lib/protobuf_config/src/da_client.rs
samlaf: Ah right, prob not great. Feel free to resolve

tomg10 (approved):

PR #3720, opened 2025-03-17
feat(consensus): Add consensus protocol versioning

brunoffranca:
## What ❔

It makes some necessary changes to allow supporting multiple consensus protocol versions. Most of the changes are on the era-consensus side (which this PR updates to v0.9). On the zksync-era side, I changed the DB to store versioned block certificates.
This PR also removes the genesis column on the consensus_replica_state table since that was deprecated a long time ago.
Also updates the Rust toolchain a bit (needed for a stable feature used in era-consensus). This caused a bunch of lints to need to also be fixed in this PR (mostly `too_long_first_doc_paragraph`).

Part of BFT-483

## Why ❔

We need to be able to support at least 2 consensus protocol versions concurrently in order to be able to do consensus protocol upgrades.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Operational changes

None.

## Checklist


core/lib/basic_types/src/macros.rs
```diff
@@ -1,6 +1,7 @@
 macro_rules! basic_type {
     ($(#[$attr:meta])* $name:ident, $type:ty) => {
-        $(#[$attr])*
+        $(#[$attr]///
+)*
```
perekopskiy, line 4: was this also changed for linter?

core/lib/dal/src/consensus_dal/mod.rs
```diff
@@ -393,11 +406,11 @@ impl ConsensusDal<'_, '_> {
     pub async fn block_certificate(
         &mut self,
         block_number: validator::BlockNumber,
-    ) -> anyhow::Result<Option<validator::CommitQC>> {
+    ) -> anyhow::Result<Option<BlockCertificate>> {
         let Some(row) = sqlx::query!(
             r#"
             SELECT
-                certificate
+                versioned_certificate
```
perekopskiy, line 413: is it ok to only use `versioned_certificate` here, if so maybe it's worth a comment?

core/node/eth_sender/src/aggregated_operations.rs
```diff
@@ -58,11 +58,13 @@ impl AggregatedOperation {
...
 
+    #[allow(dead_code)]
     pub fn is_prove_or_execute(&self) -> bool {
         self.get_action_type() == AggregatedActionType::PublishProofOnchain
             || self.get_action_type() == AggregatedActionType::Execute
     }
 
+    #[allow(dead_code)]
     pub fn is_execute(&self) -> bool {
         self.get_action_type() == AggregatedActionType::Execute
```
perekopskiy, line 69: nit: let's just remove those

core/lib/dal/migrations/20250314155250_consensus_versioning.up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE consensus_replica_state DROP COLUMN genesis;
```
RomanBrodetski, line 1: migrations are run **before** the server version is upgraded. I think it will break the old server version? I suggest first stopping using the field and remove the field in one of next versions.

This quarter we will work on a more robust process for catching breaking server changes.

core/lib/dal/src/consensus_dal/mod.rs
```diff
@@ -361,15 +346,42 @@ impl ConsensusDal<'_, '_> {
...
+                .as_ref()
+                .map(|cert| d.proto_fmt(cert))
+                .transpose()?;
+
+            if let Some(qc) = qc {
+                return Ok(BlockStoreState {
+                    first,
+                    last: Some(Last::FinalV1(qc)),
+                });
+            }
```
RomanBrodetski, line 384: Would it make sense to add logs for this for easier debug? Feel free to ignore if not

core/lib/dal/src/consensus_dal/mod.rs
```diff
@@ -518,25 +538,33 @@ impl ConsensusDal<'_, '_> {
...
+        cert: &BlockCertificate,
     ) -> Result<(), InsertCertificateError> {
         use InsertCertificateError as E;
-        let header = &cert.message.proposal;
+
+        // Extract block number and payload hash based on certificate variant
+        let (block_number, payload_hash) = match cert {
+            BlockCertificate::V1(qc) => (qc.message.proposal.number, qc.message.proposal.payload),
+            BlockCertificate::V2(qc) => (qc.message.proposal.number, qc.message.proposal.payload),
+        };
```
RomanBrodetski, line 549: Why for some things we have methods in BlockCertificate like (like `number`), whereas for other we `match` on the caller side (like here or above in `block_store_state`)? 

I think it's a good idea to encapsulate versioning logic in the  enum itself, so that you don't have to modify it everywhere when you add a new version. Also it's easier to follow the caller logic (you have to deal with less concepts there)

core/lib/basic_types/src/macros.rs
brunoffranca: I didn't change it manually but was done on one of the lint commits. Probably clippy malfunctioned here. It doesn't change the logic but I'll remove the empty comment.

core/lib/dal/src/consensus_dal/mod.rs
brunoffranca: No, it's not, I missed that one. I'll fix it.

core/lib/basic_types/src/macros.rs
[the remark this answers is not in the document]
brunoffranca: I didn't change it manually but was done on one of the lint commits. Probably clippy malfunctioned here. It doesn't change the logic but I'll remove the empty comment.

core/lib/dal/migrations/20250314155250_consensus_versioning.up.sql
brunoffranca: Should not break the server. Genesis column has been deprecated for months now. https://github.com/matter-labs/zksync-era/blob/main/core/lib/dal/src/consensus_dal/mod.rs#L83-L86

brunoffranca: I mean, I can still just stop using the field completely. But right now is already only used as fallback.

core/lib/dal/src/consensus_dal/mod.rs
brunoffranca: Maybe, I'll think about it.

core/lib/dal/src/consensus_dal/mod.rs
brunoffranca: No particular reason. I'll encapsulate all the methods then.

core/lib/dal/migrations/20250314155250_consensus_versioning.up.sql
RomanBrodetski: > I mean, I can still just stop using the field completely. But right now is already only used as fallback.

Ah I didin't catch this from the code. If you are sure this will not break on user-facing envs, feel free to leave as is

brunoffranca: Got it.

core/lib/basic_types/src/macros.rs
brunoffranca: Yup, clippy freaks out a bit here.
```
error: first doc comment paragraph is too long
   --> lib/basic_types/src/macros.rs:3:11
    |
3   |           $(#[$attr])*
    |             ^
    |
   ::: lib/basic_types/src/lib.rs:256:1
    |
256 | / basic_type!(
257 | |     /// ChainId in the Ethereum network.
258 | |     /// IMPORTANT: Please, use this method when exactly the L1 chain id is required.
259 | |     /// Note, that typically this is not the case and the majority of methods need to work
...   |
262 | |     u64
263 | | );
    | |_- in this macro invocation
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#too_long_first_doc_paragraph
    = note: `-D clippy::too-long-first-doc-paragraph` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::too_long_first_doc_paragraph)]`
    = note: this error originates in the macro `basic_type` (in Nightly builds, run with -Z macro-backtrace for more info)
help: add an empty line
    |
3   |         $(#[$attr]///
    |           ~~~~~~~~~~~

```
I'll try another fix.

perekopskiy (approved):

perekopskiy (approved):

hatemosphere (approved):
approving only changes in CI

PR #3724, opened 2025-03-18
fix(da):  use correct pubdata params on existing batch after switching da

thomas-nguy:
## What ❔

Currently, the pubdata params is defined when we are opening a new batch (it is returned by `wait_for_new_batch_env`) and tied to the `FinishBatch` command. 

This can create concurrent issue in case we are switching da with an existing opened batch.

To solve this, we are updating the logic to fetch the pubdata params ONLY before closing the batch introducing a new method `get_pubdata_params` in `StateKeeperIO` and pass it to the `FinishBatch` command.

With this new flow, the concurrent problem does not exist.

## Why ❔

In case there is an existing non-empty batch when switching DA, the batch will still be using old pubdata parameters and won't be able to be committed to L1.

The system will break

## Is this a breaking change?
- [ ] Yes
- [x] No

## Operational changes
None

## Checklist

### DIFF
--- a/core/lib/vm_executor/src/batch/executor.rs
+++ b/core/lib/vm_executor/src/batch/executor.rs
@@ -11,7 +11,7 @@ use zksync_multivm::interface::{
     storage::{ReadStorage, StorageView},
     BatchTransactionExecutionResult, FinishedL1Batch, L2BlockEnv,
 };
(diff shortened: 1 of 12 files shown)


deniallugo-ml:

PR #3726, opened 2025-03-18
feat(api): Support permissioned deployments

RomanHiden:
## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


core/lib/protobuf_config/src/proto/config/tx_sink.proto
```diff
@@ -0,0 +1,7 @@
+syntax = "proto3";
+
+package zksync.config.tx_sink;
+
+message TxSink {
+  optional bool deployment_allowlist_sink = 1;
```
dimazhornyk, line 6: maybe it's worth naming it `use_deployment_allowlist_sink`

core/lib/vm_interface/src/types/outputs/execution_result.rs
```diff
@@ -204,6 +220,7 @@ impl VmExecutionResultAndLogs {
             computational_gas_used: self.statistics.computational_gas_used,
             pubdata_published: self.statistics.pubdata_published,
             circuit_statistic: self.statistics.circuit_statistic,
+            contract_deployment_count: contract_deployment_count,
```
dimazhornyk, line 223:
```suggestion
            contract_deployment_count,
```

core/lib/vm_interface/src/types/outputs/statistic.rs
```diff
@@ -217,6 +217,8 @@ pub struct VmExecutionMetrics {
     pub computational_gas_used: u32,
     pub pubdata_published: u32,
     pub circuit_statistic: CircuitStatistic,
+    #[serde(default)]
```
dimazhornyk, line 220: is the ` #[serde(default)]` necessary here?

core/node/api_server/src/tx_sender/result.rs
```diff
@@ -99,6 +101,7 @@ impl SubmitTxError {
             Self::ProxyError(_) => "proxy-error",
             Self::Internal(_) => "internal",
             Self::FailedBlockTimestampAssertion => "failed-block-timestamp-assertion",
+            Self::SenderNotInAllowList(_) => "sender-not-allow-list",
```
dimazhornyk, line 104:
```suggestion
            Self::SenderNotInAllowList(_) => "sender-not-in-allow-list",
```

core/node/api_server/src/tx_sender/whitelisted_deploy_pool_sink.rs
```diff
@@ -0,0 +1,67 @@
...
+    ) -> Result<L2TxSubmissionResult, SubmitTxError> {
+        let initiator = tx.initiator_account();
+
+        // Only enforce the allow-list check if VM actually deployed a contract.
+        if execution_metrics.vm.contract_deployment_count > 0 {
+            let mut connection = self
+                .master_pool
+                .connection_tagged("api")
+                .await
+                .map_err(DalError::generalize)?; // Removed `mut`, not needed
```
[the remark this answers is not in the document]
RomanHiden: ahh I tried to refactor something  but that didn't work.

core/lib/protobuf_config/src/proto/config/tx_sink.proto
RomanHiden: definitely

core/lib/vm_interface/src/types/outputs/execution_result.rs
```diff
@@ -95,6 +95,19 @@ impl VmEvent {
...
+
+    /// Returns the number of `ContractDeployed` events in the given list of events.
+    pub fn count_contract_deployments(events: &[Self]) -> usize {
+        events
+            .iter()
+            .filter(|event| {
+                event
+                    .indexed_topics
+                    .get(0)
+                    .map_or(false, |&topic| topic == VmEvent::DEPLOY_EVENT_SIGNATURE)
```
slowli, line 107: Shouldn't this check the `address` that emitted the event as well?

core/node/api_server/src/tx_sender/whitelisted_deploy_pool_sink.rs
```diff
@@ -0,0 +1,67 @@
...
+
+            let is_allowed = connection
+                .contracts_deploy_allow_list_dal()
+                .is_address_allowed(&initiator)
+                .await
+                .map_err(DalError::generalize)?;
```
slowli, line 46: Just to understand better:

- Can the `contract_deploy_allow_list` table be updated at any moment? I.e., it would be unsound to read the allow list from the config at the node start or cache it in general, right?
- Is Postgres the only possible storage for the allow list? Adding a separate table for it will affect the Postgres schema for all nodes on all envs, where it'll be just deadweight. As potential alternatives, it could be read from a file or a server (both assuming that the list must be re-read on each request; see the question above).

core/lib/config/src/configs/tx_sink.rs
```diff
@@ -0,0 +1,15 @@
+use serde::Deserialize;
+
+#[derive(Debug, Deserialize, Clone, PartialEq)]
+pub struct TxSinkConfig {
```
slowli, line 4: Bikeshedding: I don't think this merits a separate config; the only config param could be added to `Web3JsonRpcConfig`.

slowli (commented):
Is filtering supposed to include deployments made not only by the tx initiator, but by contracts inside a transaction? (The current approach will count these as deployments.)

core/lib/vm_interface/src/types/outputs/statistic.rs
RomanHiden: not needed

core/lib/config/src/configs/tx_sink.rs
RomanHiden: will move it there

core/lib/vm_interface/src/types/outputs/execution_result.rs
RomanHiden: it definitely should

core/node/api_server/src/tx_sender/whitelisted_deploy_pool_sink.rs
RomanHiden: We want a permissioned chain from the start, and the allow list hopefully will grow large.

1. File-based approach  
   - Storing in memory is easy; 5MB RAM holds ~150K wallets.  
   - Periodic re-reads are fine; NVMe can load 5MB in ~0.002 sec.  
   - Not unsound if it's chain-specific.

2. HTTP endpoint  
   - Adds unnecessary delays to the sequencer.

3. Postgres + RocksDB  
   - I was thinking of having a RocksDB on top of Postgres - not worth it I think.  
 the table doesn't necessarily have to be present on all schema for all nodes. I understand the maintenance concern here too. 
 
 If you suggest file is better I will rework this to be a file based.

slowli: Subjectively, a file is indeed better, provided that it satisfies all requirements. The path to it may be encoded in the config (i.e., if the path is present, allowlist filtering is enabled). Would love to hear other opinions though; cc @RomanBrodetski @Deniallugo

Deniallugo: I agree. 
RocksDB looks like an overkill, but simple json/yaml file is good enough as an addon

CryptoMan0x: Keep in mind that Sophon needs to be the admin of this whitelist file and needs a way to update it without invoking Matter Labs.

githubdoramon: @slowli @Deniallugo 

If we store this as a json file on a CDN, it gives Sophon full control of it, so we don't need to bother you guys every time we want to change it. Can the sequencer just periodically fetch this file and store in memory?


Also, one important caveat, most likely we can just leave this file empty most of the time, because the idea is just to allow some developers to be able to deploy from L!, but in 99.99999999% of cases, they don't really need this, they can just deploy directly from L2 (FYI @CryptoMan0x).

RomanHiden: @slowli  Alex, any further feedback from you? I have a wip file based approach.

core/node/api_server/src/tx_sender/whitelisted_deploy_pool_sink.rs
```diff
@@ -0,0 +1,64 @@
...
+    async fn submit_tx(
+        &self,
+        tx: &L2Tx,
+        execution_metrics: TransactionExecutionMetrics,
+        validation_traces: ValidationTraces,
+    ) -> Result<L2TxSubmissionResult, SubmitTxError> {
+        let initiator = tx.initiator_account();
+
+        // Only enforce the allow-list check if VM actually deployed a contract.
+        if execution_metrics.vm.contract_deployment_count > 0 {
```
slowli, line 35: Highlighting my previous question:

> Is filtering supposed to include deployments made not only by the tx initiator, but by contracts inside a transaction? (The current approach will count these as deployments.)

To give an example, where this difference matters: suppose we have a contract factory

```solidity
contract Counter {
    // ...
}

contract CounterFactory {
    function createCounter() public {
        Counter counter = new Counter();
        // Do something with the counter...
    }
}
```

As currently implemented a call to `CounterFactory::createCounter()` (even an indirect one, if some other contract calls `CounterFactory`) will be counted in `contract_deployment_count`. This *may* be intended, but looks like spooky action at a distance.

core/node/api_server/src/tx_sender/whitelisted_deploy_pool_sink.rs
```diff
@@ -0,0 +1,64 @@
...
+            let initiator_allowed = self.allowlist_service.is_address_allowed(&initiator).await;
+
+            let contract_address_allowed = self
+                .allowlist_service
+                .is_address_allowed(&tx.execute.contract_address.unwrap_or_default())
```
slowli, line 40: `unwrap_or_default()` looks hacky here; `contract_address == None` corresponds to EVM deployments. It'd be more straightforward to use `if let` here:

```rust
let contract_address_allowed = if let Some(addr) = &tx.execute.contract_address {
    self.allowlist_service.is_address_allowed(addr).await
} else {
    false // EVM deployments require the initiator to be whitelisted
};
```

core/node/node_framework/src/implementations/layers/web3_api/tx_sink/master_pool_sink.rs
```diff
@@ -35,8 +43,18 @@ impl WiringLayer for MasterPoolSinkLayer {
...
     async fn wire(self, input: Self::Input) -> Result<Self::Output, WiringError> {
         let pool = input.master_pool.get().await?;
-        Ok(Output {
-            tx_sink: MasterPoolSink::new(pool).into(),
-        })
+        let master_pool_sink = MasterPoolSink::new(pool.clone());
+
+        let tx_sink = if self.rpc_config.deployment_allowlist_sink {
+            let allowlist_service = AllowListService::new(
+                self.rpc_config.http_file_url.unwrap_or_default(),
```
slowli, line 50: `unwrap_or_default` doesn't look reasonable here and below; "" is not a valid URL, and `Duration::ZERO` is not a reasonable refresh interval.

- `http_file_url` can be used to signal whether allowlisting is enabled (if it's `Some(_)`, it is, if it's `None`, it isn't).
- `refresh_interval_secs` field should be private and have a public getter, as [some other config fields](https://github.com/matter-labs/zksync-era/blob/ed4926f4cd0705ad6f1ed57ff5ee2c8c79af4987/core/lib/config/src/configs/api.rs#L290).

core/lib/config/src/configs/api.rs
```diff
@@ -224,6 +224,10 @@ pub struct Web3JsonRpcConfig {
     /// (hundreds or thousands RPS).
     #[serde(default)]
     pub extended_api_tracing: bool,
+    #[serde(default)]
+    pub deployment_allowlist_sink: bool,
+    pub http_file_url: Option<String>,
+    pub refresh_interval_secs: Option<u64>,
```
slowli, line 230: These field names are too vague; e.g., nothing signals that `refresh_interval_secs` concerns the allowlist service specifically. I'd suggest to create a subconfig struct inside `Web3JsonRpcConfig` so that the allowlist params are well-structured:

```yaml
api:
  web3_json_rpc:
    deployment_allowlist:
      file_url: "http://example.com"
      refresh_interval_secs: 10
```

core/node/api_server/src/tx_sender/result.rs
```diff
@@ -69,6 +69,8 @@ pub enum SubmitTxError {
     Internal(#[from] anyhow::Error),
     #[error("transaction failed block.timestamp assertion")]
     FailedBlockTimestampAssertion,
+    #[error("sender address {0} is not in allow list")]
+    SenderNotInAllowList(Address),
```
slowli, line 73: Bikeshedding: Wouldn't the caller have the initiator address readily available as a part of the submitted transaction? So this info looks redundant. If it's retained, it's weird that the called contract address isn't included as well.

core/lib/dal/src/contract_allow_list_dal.rs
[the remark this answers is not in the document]
RomanHiden: Yes

core/node/api_server/src/tx_sender/whitelisted_deploy_pool_sink.rs
RomanHiden: I have considered the following scenarios
1) tx.initiator -> CONTRACT_DEPLOYER
2) tx.initiator -> factory1 -> CONTRACT_DEPLOYER


We want to be able to:
1) whitelists specific users to allow contract deployment - case 1. 
2) allow specific factories to allow contract deployments for any users. e.g `UniswapV2Factory` like - case 2. 


there is one more case:
3) tx.initiator -> factory2 -> factory1 -> CONTRACT_DEPLOYER
when we would allow any contract to call factory1 to deploy contracts. e.g contract calling `UniswapV2Factory` 
this is not a hard requirement for us. and its also harder to implement in current design for me.

slowli: If you think that the current approach works reasonably well, OK. Strictly speaking though, it doesn't satisfy this req:

> allow specific factories to allow contract deployments for any users. e.g UniswapV2Factory like - case 2.

...because if the factory is called through indirection, it'll be incorrectly blacklisted. (A more esoteric violation is if the whitelisted contract can call other contracts; in this case, deployments from *these* contracts will go through even though they are not whitelisted.) For 100% correspondence to requirements, you'd need to analyze the deployer address in `ContractDeployed` events (which would also eliminate initiator / called contract distinction). Which looks easily doable if you change `TxSink::submit_tx()` signature to accept `&SandboxExecutionOutput` instead of `TransactionExecutionMetrics` (the former includes the latter, but also includes `events: Vec<VmEvent>` – all events emitted during execution).

core/node/node_framework/Cargo.toml
```diff
@@ -67,6 +67,7 @@ serde.workspace = true
 tokio = { workspace = true, features = ["rt"] }
 ctrlc.workspace = true
 semver.workspace = true
+reqwest = { version = "0.11", features = ["json", "rustls-tls"] }
```
RomanHiden, line 70: @slowli  I am not sure this is the right place?

RomanHiden:
I have addressed all review comments. tested and did self review. 

Please check again. @slowli

RomanHiden:
anyone willing to test
```    deployment_allowlist:
      http_file_url: "http://127.0.0.1:8000/whitelist.json"
      refresh_interval_secs: 10
```


whitelist.json 
```
{
    "addresses": [
      "0x618263CE921F7dd5F4f40C29f6c524Aaf97b9bbd",
      "0x0000000000000000000000000000000000000000",
      "0xE0486B081dAcC8e6af7f204abF508Ecc044B48A6"
    ]
  }
```
      
python3 -m http.server 8000

core/node/node_framework/src/implementations/layers/web3_api/tx_sink/whitelisted_master_pool_sink_layer.rs
```diff
@@ -0,0 +1,183 @@
...
+        Self {
+            url,
+            refresh_interval,
+            allowlist,
+            client: Client::new(),
+            etag: tokio::sync::Mutex::new(None),
+        }
+    }
+
+    async fn fetch(&self) -> anyhow::Result<Option<HashSet<Address>>> {
```
slowli, line 104: You can use `&mut self` receiver here to avoid wrapping the `ETag` field into a `Mutex`. Alternatively, return a tag together with data, so that it's updated on the caller side (IMO, that would be a preferable option).

slowli, line 104: You can use `&mut self` receiver here to avoid wrapping the `ETag` field into a `Mutex`. Alternatively, return a tag together with data, so that it's updated on the caller side (IMO, that would be a preferable option).

docs/src/guides/advanced/18_allowlist.md
```diff
@@ -0,0 +1,114 @@
...
+```
+{
+  "addresses": [
+    "0x1234...abcd",
+    "0xbeef...dead"
+  ]
+}
+```
+
+- Addresses must be checksummed.
```
slowli, line 50: What does this mean? AFAICT, no checksum verification is performed on the app level.

docs/src/guides/advanced/18_allowlist.md
```diff
@@ -0,0 +1,114 @@
...
+
+```api:
+  web3_json_rpc:
+    deployment_allowlist:
+      http_file_url: "https://example.com/allowlist.json"
+      refresh_interval_secs: 60
+```
+
+- `http_file_url` (required): HTTP(S) URL to the allowlist JSON.
+- `refresh_interval_secs` (optional): Refresh interval in seconds (default: 60).
```
slowli, line 25: Nit: The default is 300s, AFAICT.

docs/src/guides/advanced/18_allowlist.md
```diff
@@ -0,0 +1,114 @@
...
+- `AllowListTask`: periodically fetches the list from a remote URL.
+- `WhitelistedDeployPoolSink`: enforces the list during L2 transaction submission.
+
+---
+
+## Configuration
+
+Configure via `api.web3_json_rpc.deployment_allowlist` in your node's YAML:
+
+```api:
```
slowli, line 17: The language tag should be `yaml`; `api` should be placed on a new line.

docs/src/guides/advanced/18_allowlist.md
```diff
@@ -0,0 +1,114 @@
+# Contract Deployment Allowlist
+
+This module restricts contract deployments on zkSync Era to a predefined list of approved addresses.
+
+It consists of:
+
+- `SharedAllowList`: in-memory, async-safe address set.
+- `AllowListTask`: periodically fetches the list from a remote URL.
+- `WhitelistedDeployPoolSink`: enforces the list during L2 transaction submission.
```
slowli, line 9: Bikeshedding: IMO, this is unnecessary implementation detail; node operators could be more interested in how to configure this feature rather than implementation details. Ditto for the "How Allowlist Enforcement Works" section below.

core/node/node_framework/src/implementations/layers/web3_api/tx_sink/whitelisted_master_pool_sink_layer.rs
```diff
@@ -0,0 +1,183 @@
...
+                        tracing::info!("AllowListTask received shutdown signal");
+                        break;
+                    }
+
+                    match self.fetch().await {
+                        Ok(Some(new_list)) => {
+                            let writer = self.allowlist.writer();
+                            let mut lock = writer.write().await;
+
+                            if *lock != new_list {
```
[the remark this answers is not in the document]
RomanHiden: If the server doesn’t implement ETag, and the allowlist content hasn’t changed, the same list of wallets will be re-read and redundantly assigned to the in-memory allowlist.
It avoids pointless updates, keeps the logs clear, and doesn't mess with anything reading the allowlist.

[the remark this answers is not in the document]
RomanHiden: if server doesn't implement ETag downloading takes more time than saving this nanoseconds. so I will remove `if`

core/lib/protobuf_config/src/api.rs
```diff
@@ -151,6 +152,8 @@ impl ProtoRepr for proto::Web3JsonRpc {
                 .context("whitelisted_tokens_for_aa")?,
             extended_api_tracing: self.extended_api_tracing.unwrap_or_default(),
             api_namespaces,
+            deployment_allowlist: read_required_repr(&self.deployment_allowlist)
+                .context("deployment_allowlist")?,
```
slowli, line 156: The `deployment_allowlist` shouldn't be required; if it's missing, `DeploymentAllowlist::default()` should be used.

core/lib/config/src/configs/api.rs
```diff
@@ -408,6 +412,50 @@ impl MerkleTreeApiConfig {
...
+}
+
+impl Default for DeploymentAllowlist {
+    fn default() -> Self {
+        Self {
+            http_file_url: None,
+            refresh_interval_secs: None,
+        }
+    }
+}
```
slowli, line 457: Nit: This can be `#[derive]`d.

core/node/node_framework/src/implementations/layers/web3_api/tx_sink/whitelist.rs
```diff
@@ -0,0 +1,98 @@
...
+impl Task for AllowListTask {
+    fn id(&self) -> TaskId {
+        "api_allowlist_task".into()
+    }
+
+    fn kind(&self) -> TaskKind {
+        TaskKind::UnconstrainedTask
+    }
+
+    async fn run(self: Box<Self>, mut stop_receiver: StopReceiver) -> anyhow::Result<()> {
```
slowli, line 73: Please move `run()` to the `api_server` method as well as an inherent method taking `self` receiver. Here, it should just delegate to it [as is the case for other components](https://github.com/matter-labs/zksync-era/blob/d1bac66e339e46d2ecac183db8d33d9ee3e85940/core/node/node_framework/src/implementations/layers/consistency_checker.rs#L103-L105). This will allow for better encapsulation, like making most of `AllowListTask` methods private.

docs/src/guides/advanced/18_allowlist.md
```diff
@@ -0,0 +1,106 @@
...
+
+This module restricts contract deployments on zkSync Era to a predefined list of approved addresses.
+
+## Configuration
+
+Configure via `api.web3_json_rpc.deployment_allowlist` in your node's YAML:
+
+```YAML
+api:
+  web3_json_rpc:fit
```
slowli, line 11: What does "fit" mean? 🙃 It looks unnecessary.

slowli (commented):
Looks better 👍 There's still a couple of architectural concers to be addressed.

RomanHiden:
@slowli  please re-re-review 🧟

core/lib/protobuf_config/src/api.rs
```diff
@@ -151,6 +152,8 @@ impl ProtoRepr for proto::Web3JsonRpc {
                 .context("whitelisted_tokens_for_aa")?,
             extended_api_tracing: self.extended_api_tracing.unwrap_or_default(),
             api_namespaces,
+            deployment_allowlist: read_required_repr(&self.deployment_allowlist)
```
slowli, line 155: This is incorrect at least semantically; `read_required_repr` implies that whatever it reads *is* required. (It may also be incorrect in practice if field conversions inside `DeploymentAllowlist` are fallible; in this case, a conversion error would be incorrectly swallowed.) Please use `read_optional_repr` instead.

core/node/api_server/src/tx_sender/master_pool_sink.rs
```diff
@@ -64,7 +64,7 @@ impl TxSink for MasterPoolSink {
     async fn submit_tx(
         &self,
         tx: &L2Tx,
-        execution_metrics: TransactionExecutionMetrics,
+        execution_outputs: &SandboxExecutionOutput,
```
slowli, line 67: Micro-nit: Since the type is named "..Output", it would make sense to name the arg `execution_output` (singular) for consistency.

core/node/api_server/src/tx_sender/whitelist.rs
```diff
@@ -0,0 +1,188 @@
...
+pub struct SharedAllowList {
+    inner: Arc<RwLock<HashSet<Address>>>,
+}
+
+impl SharedAllowList {
+    pub fn writer(&self) -> Arc<RwLock<HashSet<Address>>> {
+        Arc::clone(&self.inner)
+    }
+
+    pub async fn is_address_allowed(&self, address: &Address) -> bool {
```
slowli, line 93: As I've said in the previous review, the benefit of allowlist logic locality is that methods like this can be made private. Ditto for `AllowListTask::{refresh_interval, fetch}`.

core/node/api_server/src/tx_sender/whitelist.rs
```diff
@@ -0,0 +1,188 @@
...
+
+        let new_etag = match response.headers().get("ETag") {
+            Some(value) => match value.to_str() {
+                Ok(s) => Some(s.to_string()),
+                Err(_) => None,
+            },
+            None => None,
+        };
+        let list = response.json::<WhitelistResponse>().await?;
+        let addresses: HashSet<H160> = list.addresses.into_iter().collect();
```
slowli, line 156: Nit: `H160` and `Address` are the same type; there's no need to use both (we mostly use `Address` in the codebase). Here specifically, it's possible to use a wildcard: `HashSet<_>`, since the compiler can infer the type automatically.

core/node/node_framework/src/implementations/layers/web3_api/tx_sink/whitelist.rs
```diff
@@ -0,0 +1,76 @@
...
+}
+
+#[async_trait]
+impl Task for AllowListTask {
+    fn id(&self) -> TaskId {
+        "api_allowlist_task".into()
+    }
+
+    fn kind(&self) -> TaskKind {
+        TaskKind::UnconstrainedTask
```
slowli, line 70: Why not `TaskKind::Task`? Unconstrained tasks are lower-level and can break more easily; [see the docs](https://github.com/matter-labs/zksync-era/blob/0d2bb67af441758690c195380d6f6d137967f19a/core/node/node_framework/src/task/mod.rs#L20).

core/node/api_server/src/tx_sender/whitelist.rs
```diff
@@ -0,0 +1,188 @@
...
+
+#[derive(Debug, Clone, Default)]
+pub struct SharedAllowList {
+    inner: Arc<RwLock<HashSet<Address>>>,
+}
+
+impl SharedAllowList {
+    pub fn writer(&self) -> Arc<RwLock<HashSet<Address>>> {
+        Arc::clone(&self.inner)
+    }
```
slowli, line 91: Nit: The only use of this method is local, so it can be made private. Also, why is cloning required? It seems that the only use would do fine without it.

core/lib/protobuf_config/src/api.rs
RomanHiden: oh I see. I haven't realized this such function exists.

core/node/api_server/src/execution_sandbox/execute.rs
```diff
@@ -69,8 +69,8 @@ impl SandboxAction {
 }
 
 /// Output of [`SandboxExecutor::execute_in_sandbox()`].
-#[derive(Debug, Clone)]
-pub(crate) struct SandboxExecutionOutput {
+#[derive(Debug, Clone, Default)]
```
slowli, line 72: Deriving `Default` is semantically incorrect here; for at least `are_published_bytecodes_ok`, the default value (`false`) means unsuccessful execution. Please use an inherent constructor instead (e.g. named `mock_success`). Would also allow to not derive `Default` for `ExecutionResult`, which I feel is semantically ambiguous (an empty return value is the default successful value for transactions, but not necessarily for calls).

slowli (dismissed):
Other than questionable `Default` derivation, looks good 👍 @StanislavBreadless, @perekopskiy, @Deniallugo, @RomanBrodetski AFAIU, we're OK with this transaction filtering logic making into the server codebase, right?

RomanHiden:
all CI fixed

PR #3727, opened 2025-03-18
feat(gateway): Test for gateway migration

Deniallugo:
## What ❔

Test for gateway migration. 

## Why ❔

We added a migration, that allows to migrate server back and force without losing the data. This pr adds a test with intermediate transactions, for checking the correctness 

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


core/node/node_framework/src/implementations/layers/settlement_layer_data.rs
```diff
@@ -74,10 +93,11 @@ pub struct Output {
     sl_chain_id: SlChainIdResource,
     l2_contracts: L2ContractsResource,
     l2_eth_client: Option<L2InterfaceResource>,
+    eth_sender_config: Option<SenderConfig>,
```
perekopskiy, line 96: IIUC you rely that every other component will not use SenderConfig as is but rather will accept modified version as input. This seems to be error-prone, there are high chances we will forget about it if new component needs this config. I think we should try to hide `config.eth.sender` and make it available only in some special method, like "get_eth_sender_config_for_sender_layer_data_layer" and write a good doc-comment

core/node/node_framework/src/implementations/layers/settlement_layer_data.rs
```diff
@@ -175,12 +186,150 @@ impl WiringLayer for SettlementLayerData {
...
+) -> SenderConfig {
+    if settlement_mode.is_gateway() {
+        config.max_aggregated_tx_gas = 4294967295;
+        config.max_eth_tx_data_size = 550_000;
+        if config.pubdata_sending_mode == PubdataSendingMode::Blobs
+            || config.pubdata_sending_mode == PubdataSendingMode::Calldata
+        {
+            config.pubdata_sending_mode = PubdataSendingMode::RelayedL2Calldata
+        }
+    }
```
perekopskiy, line 333: let's also add comment on why this is needed and issue a info or warn log that these values were modified

core/tests/gateway-migration-test/tests/migration.test.ts
```diff
@@ -0,0 +1,289 @@
...
+        });
+        await secondDepositHandle.wait();
+        while ((await tester.web3Provider.getL1BatchNumber()) <= initialL1BatchNumber + 1) {
+            await utils.sleep(1);
+        }
+
+        const balance = await alice.getBalance();
+        expect(balance === depositAmount * 2n, 'Incorrect balance after deposits').to.be.true;
+
+        if (process.env.CHECK_EN_URL) {
```
perekopskiy, line 140: I think `if`s with CHECK_EN_URL should be removed from this test

.github/workflows/ci-core-reusable.yml
```diff
@@ -415,27 +415,33 @@ jobs:
...
+      - name: Build test dependencies
+        run: |
+          ci_run zkstack dev test build
+
       - name: Migrate chains to gateway
         if: matrix.use_gateway_chain == 'WITH_GATEWAY'
         run: |
-          ci_run zkstack chain migrate-to-gateway --chain era --gateway-chain-name gateway
+          
+          ci_run zkstack dev test gateway-migration --direction to  --chain era --no-deps
```
perekopskiy, line 426: nit: double space after `to`

perekopskiy (commented):
Please also add description for non-test changes

core/node/node_framework/src/implementations/layers/settlement_layer_data.rs
Deniallugo: yep, i thought about something like this.

core/tests/gateway-migration-test/tests/migration.test.ts
```diff
@@ -0,0 +1,281 @@
...
+            let slMainContract = new ethers.Contract(
+                gatewayInfo?.l2DiamondProxyAddress!,
+                ZK_CHAIN_INTERFACE,
+                gatewayInfo?.gatewayProvider
+            );
+            let slAddressSl = await slMainContract.getSettlementLayer();
+            expect(slAddressl1 != ZeroAddress);
+            expect(slAddressSl == ZeroAddress);
+        } else {
+            let slAddressl1 = await l1MainContract.getSettlementLayer();
```
StanislavBreadless, line 186: can we verify that the `slAddressSl` is not zero here?

it should be equal to `L1_SETTLEMENT_LAYER_VIRTUAL_ADDRESS` constant

core/tests/gateway-migration-test/tests/migration.test.ts
```diff
@@ -0,0 +1,281 @@
...
+    step('Check the settlement layer on l1 and gateway', async () => {
+        if (direction == 'TO') {
+            let gatewayInfo = getGatewayInfo(pathToHome, fileConfig.chain!);
+            let slAddressl1 = await l1MainContract.getSettlementLayer();
+            let slMainContract = new ethers.Contract(
+                gatewayInfo?.l2DiamondProxyAddress!,
+                ZK_CHAIN_INTERFACE,
+                gatewayInfo?.gatewayProvider
+            );
+            let slAddressSl = await slMainContract.getSettlementLayer();
```
StanislavBreadless, line 182: this a bit more correct

```suggestion
            let slAddressGW = await slMainContract.getSettlementLayer();
```

core/tests/recovery-test/tests/genesis-recovery.test.ts
```diff
@@ -135,8 +135,21 @@ describe('genesis recovery', () => {
 
         while (!treeFetcherSucceeded || !reorgDetectorSucceeded || !consistencyCheckerSucceeded) {
             await sleep(1000);
+            console.log(treeFetcherSucceeded, reorgDetectorSucceeded, consistencyCheckerSucceeded);
```
StanislavBreadless, line 138: do you want to keep it?

core/tests/gateway-migration-test/tests/migration.test.ts
```diff
@@ -0,0 +1,281 @@
...
+            await utils.spawn(`zkstack chain notify-about-to-gateway-update --chain ${fileConfig.chain}`);
+        } else {
+            await utils.spawn(`zkstack chain notify-about-from-gateway-update --chain ${fileConfig.chain}`);
+        }
+        // Trying to send a transaction from the same address again
+        await checkedRandomTransfer(alice, 1n);
+
+        if (direction == 'TO') {
+            await utils.spawn(
+                `zkstack chain migrate-to-gateway --chain ${fileConfig.chain} --gateway-chain-name ${gatewayChain}`
```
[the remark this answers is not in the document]
Deniallugo: we have `await mainNodeSpawner.mainNode?.waitForShutdown();`
without killing the server. so the server waits it self

core/tests/gateway-migration-test/tests/migration.test.ts
Deniallugo: unfortunately no, it could fail, because not all parts are exists yet on gateway

StanislavBreadless (approved):

PR #3728, opened 2025-03-19
DRAFT: precompiles & v28

mm-zk:
## What ❔
Preparation for v28 and precompiles


core/lib/vm_executor/src/oneshot/contracts.rs
```diff
@@ -107,9 +107,10 @@ impl<C: ContractsKind> MultiVmBaseSystemContracts<C> {
             ProtocolVersionId::Version25 => &self.vm_protocol_defense,
             ProtocolVersionId::Version26 => &self.gateway,
             ProtocolVersionId::Version27 => &self.vm_evm_emulator,
-
-            // Speculative base system contracts for the next protocol version to be used in the upgrade integration test etc.
+            // TODO: use v28 contracts
```
perekopskiy, line 110: TODO: should be done when contracts commit is finalized

PR #3729, opened 2025-03-19
feat!: Remove old prover stack

EmilLuta:
This PR removes the entire old prover stack. What it means for you: Firstly, if you're using the deprecated witness_vector_generator & prover_fri, the binaries are gone. Secondly, you will need to migrate to circuit_prover (which is faster & comes with extra bells & whistles).

This PR does not attempt to refactor the codebase post removal. It's big enough as it is, a follow-up cleanup PR will be needed.

### DIFF
--- a/etc/env/file_based/general.yaml
+++ b/etc/env/file_based/general.yaml
@@ -135,26 +135,11 @@ prover:
   prometheus_port: 3315
   max_attempts: 10
   generation_timeout_in_secs: 600
-  setup_load_mode: FROM_DISK
-  specialized_group_id: 100
-  queue_capacity: 10
-  witness_vector_receiver_port: 3316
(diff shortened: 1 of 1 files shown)


popzxc (approved):
That's super cool. I do think we should document the migration from the "old stack" to the "new stack" though. It doesn't have to be a part of this PR, but it's something we should do and announce quickly after this is merged, to give people time to actually migrate.

popzxc:
@EmilLuta removed from the queue due to the comment above -- please decide whether you want to merge now or prepare docs first.

yorik (approved):
Thank you for cleaning up! Changes to metrics are fine.

Nit: I'd also remove `docker/prover-gpu-fri*` and `docker/witness-vector-generator`.

PR #3730, opened 2025-03-19
feat(zkos): Implement ZK OS tree manager

slowli:
## What ❔

Implements a tree manager component responsible for updating the ZK OS Merkle tree based on Postgres data and providing REST API for it (to be used for witness input generation).

## Why ❔

Necessary to keep the ZK OS Merkle tree synced.

## Is this a breaking change?

- [ ] Yes
- [x] No

## Operational changes

No operational changes; the new component is not tied to the node binaries.

## Checklist


core/node/zk_os_tree_manager/src/batch.rs
```diff
@@ -0,0 +1,168 @@
...
+        indices_latency.observe_with_count(l1_batches_for_initial_writes.len());
+
+        // Sort tree logs by the enumeration index to get the correct order for inserts.
+        let mut tree_logs = BTreeMap::new();
+        for (storage_key, value) in touched_slots {
+            if let Some(&(initial_write_batch_for_key, leaf_index)) =
+                l1_batches_for_initial_writes.get(&storage_key.hashed_key())
+            {
+                // Filter out logs that correspond to deduplicated writes.
+                if initial_write_batch_for_key <= l1_batch_number {
```
perekopskiy, line 138: I think it can be simplified further: there will be no duplicated writes in `storage_logs` so filtering by `l1_batch_number` is needless. 
Off-topic: probably we won't even save `l1_batch_number` in `initital_writes` as it can be queried from `storage_logs` -- get the first miniblock number for given `hashed_key`. We still need the table to store hashed_key => leaf_index though. I'm not certain because we need to migrate old chains to zk os so we might still want to save `l1_batch_number`.

core/node/zk_os_tree_manager/README.md
```diff
@@ -0,0 +1,12 @@
+# ZK OS Tree Manager
+
+Node component responsible for managing the ZK OS Merkle tree. Functionally equivalent to the metadata calculator for
+the Era tree, except that it does not participate in witness generation (i.e., only supports the lightweight tree mode
+in the metadata calculator terms).
```
perekopskiy, line 5: Just to understand better, full mode is not yet implemented because witness generation process is tbd or do you think we will not need it at all? Querying proofs in-flight is suboptimal so IMO we can't get rid of full-mode

core/node/zk_os_tree_manager/src/batch.rs
slowli: If this doesn't produce much overhead, saving `l1_batch_number` could be beneficial at least for consistency.

core/node/zk_os_tree_manager/README.md
slowli: WDYM by querying proofs in-flight? What I have in mind is the following workflow (I've described it in Notion as well):

1. We start executing the batch recording witness input as in the test node (cf. `run_batch_generate_witness` in zk_ee).
2. Once ZK OS queries the first tree proof, we have a full information about read / written keys (i.e., all storage queries made during execution). So we get a batch proof for all these keys using [a tree client](https://github.com/matter-labs/zksync-era/blob/ee63fc50fcb98840fa874fc83a3e5505edf15c17/core/node/zk_os_tree_manager/src/api/client.rs#L47) (which can rely on an in-process or a remote tree).
3. Then, for each ZK OS tree query, we convert a proof into Merkle paths consumed by ZK OS. This is mostly trivial, esp. if [level / index on level info](https://github.com/matter-labs/zksync-era/blob/f05fffda72393fd86c752e88b7192cc8e0c30b68/core/lib/zk_os_merkle_tree/src/hasher/proofs.rs#L27-L29) is retained in the proof (but it can be restored anyways). We know that ZK OS will only query keys read / written during execution (it's theoretically possible to provide a slow fallback in case it queries something else, but it looks redundant).
4. At the end of the VM run, we get our witness input.

So, there's a single batched call to the tree in the entire VM run. Note that if we want to execute multiple blocks in the entire VM run (i.e., batch proving), this approach still works – even if the tree is already updated with some or all of these blocks! I guess we'd want to prove storage once per run still (to reduce proving costs). In this case, we'd still query the tree once for a batch proof for all keys read / written in the entire run, with the version set to correspond to the initial block in the run.

In other words, [`MerkleTree::prove()`](https://github.com/matter-labs/zksync-era/blob/f05fffda72393fd86c752e88b7192cc8e0c30b68/core/lib/zk_os_merkle_tree/src/lib.rs#L156-L162) is the main way to get proofs, and [`extend_with_proof`](https://github.com/matter-labs/zksync-era/blob/f05fffda72393fd86c752e88b7192cc8e0c30b68/core/lib/zk_os_merkle_tree/src/lib.rs#L272-L277) is there mostly for testing purposes.

perekopskiy: Not sure I understood how we would know all read/written keys in advance so we can do batch call, can you please elaborate?

slowli: Sorry, I wasn't clear. By "in advance" I mean "before tree-related queries are made by ZK OS". We can record distinct storage queries made by ZK OS via `ReadStorage::read()`. I've made a mistake here: we don't know whether it's a read or a write, but if I'm reading ZK OS code correctly, previous values *are* necessarily read for all writes. So, at the end of the VM execution when ZK OS starts verifying storage, we'll have all read / written keys on the host side, and we can supply them to `MerkleTree::prove()` / `TreeApiClient::get_proof()`. Does this make sense, or am I missing something?

perekopskiy: Agree

perekopskiy (dismissed):

perekopskiy (approved):

PR #3747, opened 2025-03-21
fix(contract-verifier): Ignore suppressable errors during compilation

popzxc:
Needed for https://github.com/zkSync-Community-Hub/zksync-developers/discussions/834

For old compiler versions (e.g. zksolc 1.5.4), `suppressedErrors` wasn't a part of compiler settings, so we cannot easily pass it to the contract verifier without changing a lot of components. Moreover, it's not a "real" error but a "heavy warning". If the contract was compiled and deployed with this warning, we should still allow verifying it.


perekopskiy (approved):

hedgar2017 (commented):
Personally, I would avoid relying on any markers or message substrings.
I'm not sure if versions older than v1.5.4 include these instructions in their error messages.

hedgar2017 (approved):

vasyl-ivanchuk (approved):

PR #3751, opened 2025-03-21
fix(contract-verifier): handle errors when the contract code is unavailable in Etherscan

vasyl-ivanchuk:
## What ❔

When a contract is not yet indexed by Etherscan and we try to verify it, the following error is returned:
`Unable to locate ContractCode at <address>`. This PR adds proper processing for such errors to give Etherscan time to index the contract.

## Why ❔

Currently, `Unable to locate ContractCode` errors are not properly processed, and as a result, the verification requests fail.

## Is this a breaking change?
- [ ] Yes
- [X] No

## Checklist

### DIFF
--- a/core/lib/contract_verifier/src/etherscan/errors.rs
+++ b/core/lib/contract_verifier/src/etherscan/errors.rs
@@ -2,13 +2,15 @@ use zksync_dal::DalError;
 
 #[derive(Debug, thiserror::Error)]
 pub(super) enum EtherscanError {
-    #[error(transparent)]
+    #[error("Error while sending an HTTP request")]
     Reqwest(#[from] reqwest::Error),
     #[error("Failed to deserialize content: {error}\n{content}")]
     Serde {
         error: serde_json::Error,
         content: String,
     },
+    #[error("Contract bytecode is not available")]
+    ContractBytecodeNotAvailable,
     #[error("Contract source code not verified")]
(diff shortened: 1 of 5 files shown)


core/lib/contract_verifier/src/etherscan/errors.rs
```diff
@@ -2,13 +2,15 @@ use zksync_dal::DalError;
 
 #[derive(Debug, thiserror::Error)]
 pub(super) enum EtherscanError {
-    #[error(transparent)]
+    #[error("Error while sending an HTTP request")]
```
vasyl-ivanchuk, line 5: I noticed locally that the URL with the API key can be logged in some cases. That's why I've changed the error to a generic one, which should be enough for us.

core/lib/contract_verifier/src/etherscan/types.rs
```diff
@@ -462,7 +465,7 @@ mod tests {
             (
                 RawEtherscanResponse {
                     status: "0".to_string(),
-                    message: "Error".to_string(),
+                    message: "NOTOK".to_string(),
```
vasyl-ivanchuk, line 468: Changing `Error` to `NOTOK` to match the actual Etherscan response.

popzxc (approved):

perekopskiy (approved):

PR #3761, opened 2025-03-25
feat(api): Allow EVM bytecode overrides

slowli:
## What ❔

Allows to accept EVM bytecode overrides in the relevant API server methods like `eth_call` and `eth_estimateGas`.

## Why ❔

With EVM emulator, we want to support EVM bytecodes everywhere, including in state overrides.

## Is this a breaking change?

- [ ] Yes
- [x] No

Technically, a highly unusual EraVM bytecode can be incorrectly detected as an EVM bytecode, but the possibility of this occurring in practice is vanishingly small. E.g., there doesn't seem to be a single functional "unusual" bytecode deployed on the Era mainnet.

## Operational changes

No operational changes.

## Checklist


perekopskiy (dismissed):

itegulov (dismissed):
Just want to highlight that IMHO we should not extend auto-detection any further and any additional bytecode kind we might accept in the future should be supplied with a marker. It's not ideal that we are baking this extra modification into an existing ETH method but I understand the reasoning behind it.

One extra thing I would love to see is metrics for bytecode kinds: specifically if the request contained a marker vs we had to detect one automatically (and what it resulted into). Assuming ZKsync tooling starts using markers as soon as this gets rolled out to the vast majority of envs, we will start seeing the gradual decline of auto-detections resulting into EraVM kind. And, at some point, we will (potentially) have an avenue of enforcing purely EVM bytecode in this endpoint (when no marker is supplied ofc).

But no strong opinion - leaving the decision on metrics up to you

PR #3763, opened 2025-03-25
test(api): Store VM dumps for diverging executions

slowli:
## What ❔

- Stores VM dumps in the core object store at least for local and CI runs.
- Fails integration tests if VM divergences are observed in API.

## Why ❔

Ensures that there are no unexpected VM divergences (previously, there were some because they were only logged).

## Is this a breaking change?

- [ ] Yes
- [x] No

## Operational changes

VM dumps are stored in the object store specified in `core_object_store` in the general config.

## Checklist


core/node/api_server/src/execution_sandbox/execute.rs
```diff
@@ -171,7 +204,53 @@ impl SandboxExecutor {
...
+        self.vm_divergence_counter.clone()
+    }
+
+    async fn dump_vm_state(
+        last_dump_timestamp: &Mutex<Option<Instant>>,
+        store: &dyn ObjectStore,
+        vm_dump: VmDump,
+    ) -> anyhow::Result<()> {
+        /// Minimum interval between VM dumps.
+        const DUMP_INTERVAL: Duration = Duration::from_secs(10);
```
slowli, line 225: Is this policy (saving no more than one VM dump per 10s OK? Maybe it makes sense to make the interval configurable or be able to disable saving VM dumps altogether? Also, do our API server pods in K8s have the necessary GCS authentication data? (If this is a security concern, it could be possible to use a local filesystem object store for VM dumps, but this has obvious downsides, like dumps disappearing after a pod restart and being inaccessible by devs for the Era mainnet etc.)

slowli:
Tested manually by introducing an error to the fast VM wrapper in `multivm`. Integration tests failed in CI, and VM dumps were added to the generated file artifacts as expected.

core/node/api_server/src/execution_sandbox/execute.rs
perekopskiy: Most likely APIs don't have access to GCS currently. IMO, 10s is too frequent, 1-10 minutes is more reasonable. I wouldn't make it configurable to keep it simpler. I'm worried about number of dumps growing in the storage, e.g. if we introduce a regression then we will have a lot dumps, probably we some info alert or another process to monitor it

perekopskiy (dismissed):

artmakh (approved):

PR #3764, opened 2025-03-25
feat: make PRIORITY_EXPIRATION configurable

dimazhornyk:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/core/lib/config/src/configs/eth_watch.rs
+++ b/core/lib/config/src/configs/eth_watch.rs
@@ -11,11 +11,18 @@ pub struct EthWatchConfig {
     /// How often we want to poll the Ethereum node.
     /// Value in milliseconds.
     pub eth_node_poll_interval: u64,
+    /// How many L1 blocks to look back for the priority operations.
+    #[serde(default = "EthWatchConfig::default_state_keeper_db_block_cache_capacity_mb")]
+    pub priotity_tx_expiration_blocks: u64,
 }
 
 impl EthWatchConfig {
     /// Converts `self.eth_node_poll_interval` into `Duration`.
     pub fn poll_interval(&self) -> Duration {
         Duration::from_millis(self.eth_node_poll_interval)
+    }
+
+    pub fn default_priotity_tx_expiration_blocks() -> u64 {
+        zksync_system_constants::PRIORITY_EXPIRATION
     }
 }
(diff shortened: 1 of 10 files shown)


sanekmelnikov:
I'm fine to approve it, although ideally we would have something similar to the logic inside eth_watch here: going back until we find upgrade_tx - that wouldn't require config changes at all.

core/lib/config/src/configs/eth_watch.rs
```diff
@@ -11,11 +11,18 @@ pub struct EthWatchConfig {
     /// How often we want to poll the Ethereum node.
     /// Value in milliseconds.
     pub eth_node_poll_interval: u64,
+    /// How many L1 blocks to look back for the priority operations.
+    #[serde(default = "EthWatchConfig::default_priority_tx_expiration_blocks")]
+    pub priority_tx_expiration_blocks: u64,
```
perekopskiy, line 16: `priority_tx` in name is a bit misleading, let's maybe call it `event_expiration_blocks` so it's clear it's applied to all events?

core/lib/config/src/configs/eth_watch.rs
```diff
@@ -11,11 +11,18 @@ pub struct EthWatchConfig {
...
 
 impl EthWatchConfig {
     /// Converts `self.eth_node_poll_interval` into `Duration`.
     pub fn poll_interval(&self) -> Duration {
         Duration::from_millis(self.eth_node_poll_interval)
     }
+
+    pub fn default_priority_tx_expiration_blocks() -> u64 {
+        zksync_system_constants::PRIORITY_EXPIRATION
+    }
```
perekopskiy, line 27: I think we should remove this constant. Looks like it's still used in `save_set_chain_id_tx`, you can either add constant directly to this method or pass `event_expiration_blocks` from config to this place, I'm ok with both options

perekopskiy (commented):
Look

core/lib/config/src/configs/eth_watch.rs
dimazhornyk: renamed to `event_expiration_blocks`

core/lib/config/src/configs/eth_watch.rs
dimazhornyk: I removed the constant; the only place the value is defined is the default for the config
and passed `event_expiration_blocks` to the `save_set_chain_id_tx`

perekopskiy (commented):
Looks good, please resolve conflicts

perekopskiy (approved):

PR #3772, opened 2025-03-27
docs(): interop docs

kelemeno:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
```diff
@@ -0,0 +1,169 @@
...
+
+### Contract based AssetTracker
+
+Asset tracker should enable interop, as interop does not go through L1, so the simple chain balance tracking is not enough.
+
+To enable interop, besides updating based on L1<>L2 transactions, we will have to [parse all interop transactions](https://github.com/matter-labs/era-contracts/blob/6a53169ec3be11b33e27c1f11902a78bc6c31de1/l1-contracts/contracts/bridge/asset-tracker/AssetTracker.sol#L231). This means we have to parse all `L2toL1` logs, if the sender is the `interopCenter` parse the message, and update the balance on the SL. 
+
+On GW we process all incoming and outgoing messages to chains. L1->L2 messages are processed as they are sent through the Gateway, and L2->L1 messages are processed together with L2->L2 messages in the `processLogsAndMessages` function.
+
+When executing each batch, each chain would call the following function from the Executor facet:
```
StanislavBreadless, line 169: We have `assetTracker.processLogsAndMessages` called inside Executor.sol this means that we trust each chain to call this function. This is okay while the CTM is controlled by the dec governance, but not okay if we plan to add more CTMs with less secure governance

Better to rephrase the description here / move the logic into a different place, so that if a chain wants to include any interop txs, they must provable go through the asset tracker and there is no other way to add an interop tx

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
```diff
@@ -0,0 +1,169 @@
...
+        bytes32 messageRoot;
+    }
+}
+```
+
+#### Restrictions
+
+The `AssetTracker` contract will be deployed on the GW and the L1, but interop and `processLogsAndMessages` is only enabled on GW, since using them on L1 is expensive. On L1 balances will be checked when the depsosit and withdrawals are processed.
+
+For now the `AssetTracker` will only support tokens in the native token vault. These are assumed to have a single origin chain id, i.e. single chain where the supply of the token can change.
```
StanislavBreadless: > in the native token vault

why? we index things by asset id, so it should not matter

> These are assumed to have a single origin chain id, i.e. single chain where the supply of the token can change.

If you mean "L1", then it is bad since ZK token is L2-native. If you mean just that each token has exactly one origin chain (but it does not have to be the same), then okay, makes sense

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
```diff
@@ -0,0 +1,169 @@
...
+
+The chain operator is responsible for migration of all token balances. However as each token migration costs L1 gas, this can be a spam attack vector. The chain can choose to ignore tokens, in which case the users can migrate the balances themselves. 
+
+Withdrawals on the chain do not need to be paused when migrating to L1, since the chain can settle without processing logs. However if the token balance is not yet migrated, the withdrawal will fail on L1. It's not bad, since it doesn't break chain's settlement, in this case the withdrawal should be retried as usual when the withdrawal fails.
+
+#### Migration number
+
+To make the chain balance migration seamless, we expose the current settlement layer in the `Bootloader`. It then saves the settlement layer in the `SystemContext.sol` contract, and the `migrationNumber` (i.e. how many times the chain has migrated) gets saved in the `ChainAssetHandler.sol` contract. We also store the migration numbers for chains on the settlement layers.
```
StanislavBreadless: note: please make this data private (i.e. dont expose any getters for it), hopefully it will be only a temporary solution

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
```diff
@@ -0,0 +1,169 @@
...
+
+We will also save the `assetMigrationNumber` in the `AssetTracker` contract. This show the last migration number where the token's balance was migrated. By comparing the `assetMigrationNumber` with the `migrationNumber`, we can determine if the token's balance was migrated or not, and if it is bridgeable or not.
```
StanislavBreadless: in a similar fashion I would only expose something extremely simple like (`canInteropToken()` flag) or something. This is more of an impl detail ofc, just leaving it here as I believe it is an important one

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
```diff
@@ -0,0 +1,169 @@
...
+
+#### Migrating to Gateway
+
+When migrating each tokens balance to Gateway from L1, we increase the `chainBalance` of the GW, and decrease the balance of the chain. There are outstanding withdrawals on L1 from the L2, so we cannot migrate the whole balance of the chain. We can only migrate the amount that is not withdrawn, i.e. the amount that the token contract on the chain has as its `totalSupply`. This might change as there can be incoming deposits. We save the `totalSupply` at the first such deposit.
+
+In order for information to flow smoothly, migrating to the Gateway has the following steps:
+
+- Chain's `AssetTracker` gets the `totalSupply` of the token (from the token or from storage) and sends a message to L1.
```
StanislavBreadless: We'll have to be careful about deposits. What can happen:

- Someone starts migrating the asset (at this point `totalSupply = 1`)
- Then a deposit is processed (`totalSupply = 2`)
- Still, on L1/GW when calculating the chainBalance the total supply that will be counted is `1` and this other 1 token will be lost.

It is solvable by always tracking new deposits separately (and adding those on top of the `totalSupply` transferred).

StanislavBreadless: At any step of the above the chain may start its migration back to L1. I am not saying I see any explicit attacks, we'll just have to think through/consider all possible edge cases

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
kelemeno: I changes the logic, previously it was Executor -> MessageRoot. Now it is Executor -> AssetTracker -> MessageRoot, precisely via this processLogs

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
kelemeno: 1. custom assetHandler might have different encodings, and they also might have multiple originChainIds. 
2. yes, I meant that.

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
kelemeno: but the underlying structure is the cleanest with this assetMigrationNumber right? I.e. you don't see easier one

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
kelemeno: > 1/GW when calculating the chainBalance the total supply that will be counted is 1

What do you mean? If it happens via L1, then the chainBalance will be automatically added to GW's. And if it happens via interop, it will be counted on GW.

kelemeno: The only attack I see at the moment is iterated L1-> GW migration, without migrating back to L1. This might be a problem. 

We need a formal method of looking at it.

maybe: the chainBalance on GW and the chainBalance on L1 always stores the latest totalSupply in the chain + the outstanding L1 withdrawals.

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
StanislavBreadless: ah okay then good

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
StanislavBreadless: I am not saying it is something unsolvable, but e.g. if logic for attributing balance to chain looked the following way:

```
L1:
to_migrate = totalSupply 

GW:
balance[chain] = to_migrate (as opposed to `+=`)
```

then it would be wrong as this additional `1` would be skipped.

kelemeno: yes, code has to be reviewed carefully

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
```diff
@@ -0,0 +1,173 @@
...
+
+The `AssetTracker` contract will be deployed on the GW and the L1, but interop and `processLogsAndMessages` is only enabled on GW, since using them on L1 is expensive. On L1 balances will be checked when the depsosit and withdrawals are processed.
+
+For now the `AssetTracker` will only support tokens in the native token vault. These are assumed to have a single origin chain id, i.e. single chain where the supply of the token can change.
+
+### Gateway
+
+The Gateway component as an intermediate settlement layer complicates the setup. L1->GW->L2 deposits, failed deposits, and L2->GW->L1 withdrawals have to be handled.
+
+First of all for chains settling on Gateway, the balance of the chain in the L1 `AssetTracker` will be moved to the Gateway's `chainBalance`. This is a complicated migration process, see [below](#migrating-and-settling-on-gateway). Similarly, when a chain migrates back to L1, its balance will be separate from the Gateway's `chainBalance`. On Gateway the `chainBalance`s are tracked for each chain separately in the GW's `AssetTracker`.
```
nikitastupin-matterlabs: > First of all for chains settling on Gateway, the balance of the chain in the L1 `AssetTracker` will be moved to the Gateway's `chainBalance`

Does that mean that if the GW is hacked (whatever that might be) then it can withdraw funds locked in `L1NativeTokenVault` of all the chains  that settle on top of it?

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
```diff
@@ -0,0 +1,173 @@
...
+
+#### Restrictions
+
+The `AssetTracker` contract will be deployed on the GW and the L1, but interop and `processLogsAndMessages` is only enabled on GW, since using them on L1 is expensive. On L1 balances will be checked when the depsosit and withdrawals are processed.
+
+For now the `AssetTracker` will only support tokens in the native token vault. These are assumed to have a single origin chain id, i.e. single chain where the supply of the token can change.
+
+### Gateway
+
+The Gateway component as an intermediate settlement layer complicates the setup. L1->GW->L2 deposits, failed deposits, and L2->GW->L1 withdrawals have to be handled.
```
nikitastupin-matterlabs: What happens in case of:

- GW->L2 deposits
- L2->GW withdrawals
- failed GW-L2 deposits
- L1->GW deposits, GW->L1 withdrawals, failed L1->GW deposits

?

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
```diff
@@ -0,0 +1,173 @@
...
+
+First of all for chains settling on Gateway, the balance of the chain in the L1 `AssetTracker` will be moved to the Gateway's `chainBalance`. This is a complicated migration process, see [below](#migrating-and-settling-on-gateway). Similarly, when a chain migrates back to L1, its balance will be separate from the Gateway's `chainBalance`. On Gateway the `chainBalance`s are tracked for each chain separately in the GW's `AssetTracker`.
+
+#### Deposits
+
+When a user deposits to a chain settling on GW, an L1->GW transaction is sent, which add the transaction to the chains Priority Queue in the Mailbox facet on Gateway. On L1 the Gateway's `chainBalance` is updated. We will include the balance change in the in the L1->GW [forwardTransactionOnGatewayWithBalanceChange](https://github.com/matter-labs/era-contracts/blob/ba7b99eee1111fb7b87df7d6cc371aa21ea864b9/l1-contracts/contracts/interop/InteropCenter.sol#L392) function call, so that the Gateway `AssetTracker` can be updated.
```
nikitastupin-matterlabs: "updated" meaning increased?

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
```diff
@@ -0,0 +1,173 @@
...
+- On GW the chain's `chainBalance` is decreased, when the chain executes its batch, and calls the `processLogsAndMessages` function. Each failed L1->L2 transaction produces a L2->L1 message, which we can verify in `processLogsAndMessages`. To know the balance change for the specific L1->L2 transaction, we save the balance change in the `forwardTransactionOnGatewayWithBalanceChange` function call. If the chain migrates to L1 before this step then it migrates with the increased chainBalance, and the failedDeposit will be verified on the L1.
+
+#### Withdrawals
+
+When a user withdraws from a chain settling on GW, we process the log on Gateway in `processLogsAndMessages`. This means that the balance of the chain is decreased on GW.
+
+Similarly when finalizing the withdrawal on L1, we decrease the Gateway's `chainBalance`. We determine whether a withdrawal is processed on L1 or on GW by checking the `L1Nullifier` on L1, where we [save](https://github.com/matter-labs/era-contracts/blob/6a53169ec3be11b33e27c1f11902a78bc6c31de1/l1-contracts/contracts/bridge/L1Nullifier.sol#L537) the settlement layer.
```
nikitastupin-matterlabs: > We determine whether a withdrawal is processed on L1 or on GW by checking the `L1Nullifier` on L1 where we save the settlement layer

What does it mean? What's the point of determining this? What are the consequences of it?

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
```diff
@@ -0,0 +1,173 @@
...
+
+### Migrating and settling on Gateway
+
+When a chain migrates from L1 to GW or from GW to L1, the `chainBalance` mapping for each token does not get transferred automatically. They can be trustlessly moved by anyone from GW to L1 and vice versa. But our standard tooling will migrate all token balances for best UX.
+
+When a chain is settling on Gateway it calls the `processLogsAndMessages` on settlement. This means that if the token is not yet migrated, and a withdrawal could be processed, the settlement would fail, since the chain's balance would be zero. In order to prevent this, withdrawals and interop will be paused on the chain until the token's balance is migrated.
```
nikitastupin-matterlabs: This paragraph is very confusing.

> This means that if the token is not yet migrated

Why tokens are migrating? The first sentence says that the chain already settles on the GW.

> This means that if the token is not yet migrated, and a withdrawal could be processed, the settlement would fail, since the chain's balance would be zero

Why would the balance be 0?

> In order to prevent this, withdrawals and interop will be paused on the chain until the token's balance is migrated

Does it pause for all tokens at once or each token one by one?

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
kelemeno: yes

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
kelemeno: - we don't have GW-L2  interactions
- GW-L1 interactions work like normal L1-L2 interactions

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
nikitastupin-matterlabs: > withdrawals and interop will be paused on the chain until the token's balance is migrated.

This seems like a blocker for becoming a Stage 1 rollup. I do support pausing for migration but we have to keep this in mind.

deniallugo-ml:

PR #3777, opened 2025-03-27
feat(gateway): Handle server shutting down error

Deniallugo:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/core/lib/web3_decl/src/error.rs
+++ b/core/lib/web3_decl/src/error.rs
@@ -49,6 +49,7 @@ pub enum Web3Error {
     #[error("Internal error")]
     InternalError(#[from] anyhow::Error),
 }
+const SERVER_IS_SHUTTING_DOWN_ERROR: i32 = 3;
 
 /// Client RPC error with additional details: the method name and arguments of the called method.
 ///
@@ -66,8 +67,10 @@ pub fn is_retriable(err: &ClientError) -> bool {
         ClientError::Transport(_) | ClientError::RequestTimeout => true,
         ClientError::Call(err) => {
             // At least some RPC providers use "internal error" in case of the server being overloaded
+
             err.code() == ErrorCode::ServerIsBusy.code()
                 || err.code() == ErrorCode::InternalError.code()
+                || err.code() == SERVER_IS_SHUTTING_DOWN_ERROR
         }
         _ => false,
     }
--- a/core/node/eth_sender/src/eth_tx_manager.rs
+++ b/core/node/eth_sender/src/eth_tx_manager.rs
@@ -542,6 +542,7 @@ impl EthTxManager {
         let pool = self.pool.clone();
 
         loop {
+            tokio::time::sleep(self.config.tx_poll_period()).await;
             let mut storage = pool.connection_tagged("eth_sender").await.unwrap();
 
             if *stop_receiver.borrow() {
@@ -552,11 +553,21 @@ impl EthTxManager {
             let l1_block_numbers = self
                 .l1_interface
                 .get_l1_block_numbers(operator_to_track)
-                .await?;
-            METRICS.track_block_numbers(&l1_block_numbers);
+                .await;
+
+            if let Err(ref error) = l1_block_numbers {
+                // Web3 API request failures can cause this,
+                // and anything more important is already properly reported.
+                tracing::warn!("eth_sender error {:?}", error);
+                if error.is_retriable() {
+                    METRICS.l1_transient_errors.inc();
+                    continue;
+                }
+            }
+
+            METRICS.track_block_numbers(&l1_block_numbers?);
 
             self.loop_iteration(&mut storage).await;
-            tokio::time::sleep(self.config.tx_poll_period()).await;
         }
         Ok(())
     }
--- a/core/node/gateway_migrator/src/lib.rs
+++ b/core/node/gateway_migrator/src/lib.rs
@@ -1,6 +1,6 @@
 use std::time::Duration;
 
-use anyhow::{bail, Context};
+use anyhow::bail;
 use tokio::sync::watch;
 use zksync_basic_types::{ethabi::Contract, settlement::SettlementLayer, L2ChainId};
 use zksync_config::configs::contracts::SettlementLayerSpecificContracts;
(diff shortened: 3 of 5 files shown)


core/lib/web3_decl/src/error.rs
```diff
@@ -66,8 +67,10 @@ pub fn is_retriable(err: &ClientError) -> bool {
         ClientError::Transport(_) | ClientError::RequestTimeout => true,
         ClientError::Call(err) => {
             // At least some RPC providers use "internal error" in case of the server being overloaded
+
             err.code() == ErrorCode::ServerIsBusy.code()
                 || err.code() == ErrorCode::InternalError.code()
+                || err.code() == SERVER_IS_SHUTTING_DOWN_ERROR
```
slowli, line 73: 3 is used as an error code for *all* `SubmitTransactionError`s, not just `ServerShuttingDown`, and some other error variants (`SerializationError`, `ProxyError`). Is this OK? If we want to distinguish `ServerShuttingDown` specifically, it'd be relatively easy to refactor the `Web3Error::SubmitTransactionError` variant (e.g., to make it include the JSON-RPC code / equivalent enum), so that `ServerShuttingDown` has a differing variant (standard `ErrorCode::ServerIsBusy` seems to match by semantics, and it's already handled above).

core/node/gateway_migrator/src/lib.rs
```diff
@@ -56,19 +56,30 @@ impl GatewayMigrator {
...
+            match current_settlement_layer {
+                Ok(current_settlement_layer) => {
+                    if self.settlement_layer != current_settlement_layer {
+                        bail!("Settlement layer changed")
+                    }
+                }
+                Err(ContractCallError::EthereumGateway(err)) => {
+                    if !err.is_retriable() {
+                        tracing::error!("Failed to fetch data from sl {}", &err);
+                        return Err(err)?;
```
slowli, line 77: `?` doesn't make sense here; you return an error anyway. Also, I'd refactor matching here to be more obvious:

```rust
match current_settlement_layer {
    Ok(current_settlement_layer) => { /* ... */ }
    Err(ContractCallError::EthereumGateway(err)) if err.is_retriable() => {
        tracing::info!("Transient error fetching data from SL: {err}");
    }
    Err(err) => {
        tracing::error!("Failed to fetch data from SL: {err}");
        return Err(err.into());
    }
}
```

core/node/gateway_migrator/src/lib.rs
```diff
@@ -96,7 +107,8 @@ pub async fn current_settlement_layer(
     let inflight_count = storage
         .eth_sender_dal()
         .get_inflight_txs_count_for_gateway_migration(!settlement_mode_from_l1.is_gateway())
-        .await?;
+        .await
+        .unwrap();
```
slowli, line 111: I don't think it's a particularly good idea to add an `unwrap()` here just because we want to inspect returned errors. It'd be better to define an error enum:

```rust
enum GatewayMigratorError {
    ContractCall(#[from] ContractCallError),
}
```

...and use the latter variant in this case.

core/node/eth_sender/src/eth_tx_manager.rs
```diff
@@ -542,6 +542,7 @@ impl EthTxManager {
         let pool = self.pool.clone();
 
         loop {
+            tokio::time::sleep(self.config.tx_poll_period()).await;
```
perekopskiy, line 545: why is this needed?

core/lib/web3_decl/src/error.rs
Deniallugo: that is the error i've found when disabled the server. I also haven't found why the number is 3

core/node/eth_sender/src/eth_tx_manager.rs
```diff
@@ -552,11 +553,21 @@ impl EthTxManager {
...
-            METRICS.track_block_numbers(&l1_block_numbers);
+                .await;
+
+            if let Err(ref error) = l1_block_numbers {
+                // Web3 API request failures can cause this,
+                // and anything more important is already properly reported.
+                tracing::warn!("eth_sender error {:?}", error);
+                if error.is_retriable() {
+                    METRICS.l1_transient_errors.inc();
+                    continue;
```
perekopskiy, line 564: I'm not sure if it is a good idea to continue eth sender if SL's API is down. I think it's worth panicking if the loop doesn't make any progress for some time.

perekopskiy (commented):
Please fill in "why" section

core/node/eth_sender/src/eth_tx_manager.rs
Deniallugo: We don't want to shut down the server if the sl api is down. 
especially given that our reliability is lower than eth

core/node/eth_sender/src/eth_tx_manager.rs
Deniallugo: just a small refactoring, for simpler code with continue

core/lib/web3_decl/src/error.rs
Deniallugo: Seems it'd be really better to distinguish these errors.

core/node/eth_sender/src/eth_tx_manager.rs
perekopskiy: Well, this way we'll see that eth sender is in error loop in 1 hour (default no-batch-commited alert interval). I think we need some compromise here and panicking/alerting if there are no successful loop iterations for 5-10 minutes is a good option IMO. This applies only for eth sender, I don't suggest doing this for gas-adjuster>state-keeper

Deniallugo: can we use l1_transient_errors ? for it

perekopskiy: I don't think so, at least not in a reliable manner. Anyway, looks like it has a bigger scope so let's maybe address it in a separate PR

core/lib/web3_decl/src/error.rs
Deniallugo: Did something like this. Should be good

perekopskiy (approved):

PR #3791, opened 2025-04-02
refactor: mark unused setup_sigint_handler function as deprecated

GarmashAlex:
The function has been replaced by SigintHandlerLayer from the 
zksync_node_framework, which provides a more integrated approach to handling
interrupt signals within the node framework.

Instead of immediately removing the function, I've marked it as deprecated
to allow for a smoother transition and maintain backward compatibility for any
potential external users of this library. The deprecation notice provides clear
guidance to use SigintHandlerLayer instead.


deniallugo-ml:

PR #3792, opened 2025-04-03
fix(en): Fix EVM bytecode handling during snapshot recovery

slowli:
## What ❔

Includes a bytecode hash in factory deps stored in a node snapshot.

## Why ❔

Snapshot recovery currently doesn't handle EVM bytecodes correctly. They are misinterpreted as EraVM bytecodes, resulting in incorrect bytecode hashes and correspondingly incorrect restored node state.

## Is this a breaking change?

- [ ] Yes
- [x] No

## Operational changes

Should be none in practice, since EVM emulator is not enabled on real envs.

## Checklist


.github/workflows/ci-core-reusable.yml
```diff
@@ -510,6 +508,11 @@ jobs:
           --db-name=zksync_en_localhost_era_consensus --l1-rpc-url=http://localhost:8545 $GATEWAY_RPC_URL --chain consensus
           ci_run zkstack external-node init --ignore-prerequisites --chain consensus
 
+      - name: Enable EVM emulator for some chains
+        run: |
+          ci_run zkstack chain enable-evm-emulator --chain era
```
itegulov, line 513: Is there no way to just enable it by default through configs? I am just not a big fan of diverging CI flow from `zkstack dev test integration` (run locally) any further than it already is...

itegulov (dismissed):

.github/workflows/ci-core-reusable.yml
slowli: It's possible to specify `--evm-emulator` option in a `zkstack chain create` command, which will enable it once `zkstack chain init` is run, e.g. [here](https://github.com/matter-labs/zksync-era/blob/a3f9f0bf41672224586d284e4e06b8ae11c41b49/.github/workflows/ci-core-reusable.yml#L275-L294). Would this work better than the current approach?

itegulov: IMHO yes specifiying `--evm-emulator` for `validium` on creation is better.

But I was actually talking about `era` specifically. It's the only VCS-tracked chain so having it represent the "norm" is desirable in my mind. I assume we can re-generate its config files in the same fashion (or maybe running `zkstack chain enable-evm-emulator --chain era` locally once would even change the config retroactively, not sure?)

slowli: Ah, I see what you mean. It's possible to change `evm_emulator: true` in [`ZkStack.yaml` for the chain](https://github.com/matter-labs/zksync-era/blob/96919454a11c672cfe5f5f34284a126dd9084cd8/chains/era/ZkStack.yaml#L15), and it seems to work, at least locally. Since I'm going to have a follow-up PR with more integration tests for EVM contracts, do you mind if I change the config there?

itegulov: Yep not a hard blocker by any means

itegulov (dismissed):

itegulov (approved):

PR #3817, opened 2025-04-07
feat(eigenda): EigenDA M1

gianbelinche:
## What ❔
This PR adds

Necessary changes to zkstack for eigenDA M1 contracts
Check on the eigenDA client to make sure blob is verified before returning inclusion data
Related to: 

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


core/node/da_clients/src/eigen/client.rs
```diff
@@ -60,7 +68,76 @@ impl EigenDAClient {
...
+            eigenda_registry_addr: config.eigenda_registry_addr,
+        })
+    }
+}
+
+impl EigenDAClient {
+    /// This function checks a mapping with form
+    /// `mapping(bytes) -> bool` in the EigenDA registry contract.
+    /// The name of the mapping is passed as a parameter.
+    async fn check_mapping(&self, inclusion_data: &[u8], mapping: &str) -> Result<bool, DAError> {
```
anupsv, line 92: whats the max size of the inclusion_data?

anupsv: could we use a hash of the data or something?

samlaf: inclusion_data is the eigenda cert (blob_info for v1), see https://github.com/Layr-Labs/eigenda-client-rs/blob/1fd8e3c2ce19b72b0b0df01e40ac288324b4ff1e/crates/rust-eigenda-client/src/sdk.rs#L217, so its max data size depends on # of blobs in the batch, as well as number of non-signers.

I feel like going forward (at least for the v2 client) it might make more sense to keep the eigenda-client zksync agnostic (so not use terms like get_inclusion_data in the client, but instead use get_eigenda_cert), and then just add a small shim client that would map external zksync terms (like inclusion_data) to eigenda internal terms (like cert).

We should also maintain a glossary of these kinds of mapping (another one is for example zksync dispatch === eigenda dispersal)

core/node/da_clients/src/eigen/client.rs
```diff
@@ -15,13 +17,18 @@ use zksync_da_client::{
...
+use zksync_web3_decl::client::{Client, DynClient, L1};
 
-use crate::utils::to_retriable_da_error;
+use crate::utils::{to_non_retriable_da_error, to_retriable_da_error};
 
 // We can't implement DataAvailabilityClient for an outside struct, so it is needed to defined this intermediate struct
 #[derive(Debug, Clone)]
 pub struct EigenDAClient {
     client: EigenClient,
+    query_client: Box<DynClient<L1>>,
```
samlaf, line 30: why not just call this an eth_client? or eth_call_client if you want to make it more obvious that its only making eth_calls and not sending txs

core/node/da_clients/src/eigen/client.rs
```diff
@@ -60,7 +68,76 @@ impl EigenDAClient {
...
+
+    async fn check_finished_batches(&self, inclusion_data: &[u8]) -> Result<bool, DAError> {
+        self.check_mapping(inclusion_data, "finishedBatches").await
+    }
+
+    async fn check_verified_batches(&self, inclusion_data: &[u8]) -> Result<bool, DAError> {
+        self.check_mapping(inclusion_data, "verifiedBatches").await
+    }
+
+    async fn check_inclusion_data_verification(
```
samlaf, line 131: add comment explaining what this is doing. Otherwise have to read all the other functions and understand what these "finishedBatches" and "verifiedBatches" mappings are on the contract, etc.

zkstack_cli/crates/config/src/contracts.rs
```diff
@@ -114,6 +114,13 @@ impl ContractsConfig {
                 .deployed_addresses
                 .avail_l1_da_validator_addr,
         );
+        self.l1.eigenda_l1_validator_addr = Some(
```
samlaf, line 117:
```suggestion
        self.l1.eigenda_l1_da_validator_addr = Some(
```

core/lib/protobuf_config/src/proto/config/da_client.proto
```diff
@@ -54,6 +54,7 @@ message EigenConfig {
     Url points_source_url = 10;
   }
   repeated uint32 custom_quorum_numbers = 11;
+  optional string eigenda_registry_addr = 12;
```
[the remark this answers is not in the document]
gianbelinche: In `proto3` all fields are optional by default, there is no "required" option (like in `proto2`).
That does not mean you can omit the field, there are checks to make sure they are all present.
About the dummy registry, I agree it is not a good idea to have it, we deleted it.

core/node/da_clients/src/eigen/client.rs
gianbelinche: I added the glossary to the Notion docs

core/lib/config/src/configs/da_client/eigen.rs
```diff
@@ -28,6 +28,8 @@ pub struct EigenConfig {
     pub points_source: PointsSource,
     /// Custom quorum numbers
     pub custom_quorum_numbers: Vec<u8>,
+    /// Address of the eigenDA registry contract
+    pub eigenda_registry_addr: Address,
```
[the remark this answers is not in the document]
gianbelinche: I changed the name to `CertAndBlobVerifier`

core/node/da_clients/src/eigen/client.rs
gianbelinche: Done

core/node/da_clients/src/eigen/client.rs
gianbelinche: Done

zkstack_cli/crates/config/src/contracts.rs
gianbelinche: Done

zkstack_cli/crates/config/src/contracts.rs
```diff
@@ -259,9 +269,15 @@ pub struct L1Contracts {
...
     pub no_da_validium_l1_validator_addr: Option<Address>,
+    // `Option` to be able to parse configs from previous protocol version
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub eigenda_l1_validator_addr: Option<Address>,
     // `Option` to be able to parse configs from pre-gateway protocol version.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub transaction_filterer_addr: Option<Address>,
+    // `Option` to be able to parse configs from previous protocol version
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub eigenda_cert_and_blob_verifier_addr: Option<Address>,
```
dimazhornyk, line 280: seems like it is a part of the da_client config, does it belong to this config?

zkstack_cli/crates/types/src/l1_network.rs
```diff
@@ -48,4 +48,28 @@ impl L1Network {
...
+                //TODO: add real address
+            }
+            L1Network::Mainnet => None, // TODO: add mainnet address after it is known
+        }
+    }
+
+    pub fn eigenda_cert_and_blob_verifier_addr(&self) -> Option<Address> {
+        match self {
+            L1Network::Localhost => {
+                Some(Address::from_str("0xEaF1790229C16Af929f515d2B74Dc0E39820160C").unwrap())
```
dimazhornyk, line 66: was this left by accident, or you expect it to be deterministically deployed every time?

core/node/da_clients/src/eigen/client.rs
```diff
@@ -60,7 +68,79 @@ impl EigenDAClient {
...
+            .await
+            .map_err(to_retriable_da_error)?;
+        let res = self
+            .eth_call_client
+            .as_ref()
+            .call_contract_function(call_request, Some(block_id.into()))
+            .await
+            .map_err(to_retriable_da_error)?;
+        match hex::encode(res.0).as_str() {
+            "0000000000000000000000000000000000000000000000000000000000000000" => Ok(false),
```
dimazhornyk, line 116: is there a more optimal way? can `Vec<u8>` be converted to integer/bool directly?

core/node/da_clients/src/eigen/client.rs
```diff
@@ -60,7 +68,79 @@ impl EigenDAClient {
         let client = EigenClient::new(eigen_config, eigen_secrets, blob_provider)
             .await
             .map_err(|e| anyhow::anyhow!("Eigen client Error: {:?}", e))?;
-        Ok(Self { client })
+
+        let eth_call_client: Client<L1> = Client::http(
```
dimazhornyk, line 72: if we need an L1 client both here and in `rust_eigenda_client` above - would it be possible to create the client once and share it?

core/node/da_clients/src/eigen/client.rs
```diff
@@ -15,13 +17,18 @@ use zksync_da_client::{
...
+use zksync_web3_decl::client::{Client, DynClient, L1};
 
-use crate::utils::to_retriable_da_error;
+use crate::utils::{to_non_retriable_da_error, to_retriable_da_error};
 
 // We can't implement DataAvailabilityClient for an outside struct, so it is needed to defined this intermediate struct
 #[derive(Debug, Clone)]
 pub struct EigenDAClient {
     client: EigenClient,
+    eth_call_client: Box<DynClient<L1>>,
```
dimazhornyk, line 30: nit: for consistency, it would be better to call it a `settlement_layer_client`

dimazhornyk (changes requested):

gianbelinche:
This is an old PR, the new one is https://github.com/matter-labs/zksync-era/pull/4232

PR #3822, opened 2025-04-08
chore: update fast vm with precompiles

koloz193:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/core/lib/multivm/src/versions/vm_fast/tracers/circuits.rs
+++ b/core/lib/multivm/src/versions/vm_fast/tracers/circuits.rs
@@ -23,6 +23,10 @@ pub(super) struct CircuitsTracer {
     sha256_cycles: u32,
     secp256r1_verify_cycles: u32,
     transient_storage_checker_cycles: u32,
+    modexp_cycles: u32,
+    ecadd_cycles: u32,
+    ecmul_cycles: u32,
+    ecpairing_cycles: u32,
 }
 
 impl Tracer for CircuitsTracer {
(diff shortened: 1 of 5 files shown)


core/lib/multivm/src/versions/vm_fast/tracers/circuits.rs
```diff
@@ -160,7 +168,11 @@ impl CircuitsTracer {
                 / GEOMETRY_CONFIG.cycles_per_secp256r1_verify_circuit as f32,
             transient_storage_checker: self.transient_storage_checker_cycles as f32
                 / GEOMETRY_CONFIG.cycles_per_transient_storage_sorter as f32,
-            ..Default::default()
+            modexp: self.modexp_cycles as f32 / GEOMETRY_CONFIG.cycles_per_modexp_circuit as f32,
+            ecadd: self.ecadd_cycles as f32 / GEOMETRY_CONFIG.cycles_per_ecadd_circuit as f32,
+            ecmul: self.ecmul_cycles as f32 / GEOMETRY_CONFIG.cycles_per_ecmul_circuit as f32,
+            ecpairing: self.ecpairing_cycles as f32
+                / GEOMETRY_CONFIG.cycles_per_ecpairing_circuit as f32,
```
joonazan, line 175: Not necessarily a thing for this PR but it seems to me that the use of floating point here is slightly suboptimal. What we're really interested in is `ceil(cycles / cycles_per_circuit)`, right? If so, we could avoid going into floating point at all.

It is nicer for the tests to compare the overly accurate value, though.

joonazan (approved):

core/lib/multivm/src/versions/vm_fast/tracers/circuits.rs
joonazan: This could actually be a problem! If we collect more than 16,777,217 of some type of cycle, the f32 can no longer accurately represent the number.

`u32` is fine, though, because we certainly will not reach 2^32 cycles with our limited gas.

PR #3826, opened 2025-04-09
feat(en): remove dependency on pubdata commitment mode

dimazhornyk:
## What ❔

**EN:** 
- Update ConsistencyChecker to not rely on EN config, but rather determine the pubdata commitment mode based on the batch data
- Add a check into ConsistencyChecker to verify that if Validium mode is used, `data_availability` must have an entry for this batch
- Update the DA fetcher to handle gaps in `data_availability`, if the chain was running as a rollup for some time - there will be a range of batches that don't have the `data_availability` entry

**Core:**
- Update `eth_tx_aggregator` to be able to override the currently configured `PubdataSendingMode` and `L1BatchCommitmentMode` with the ones stored in the database.

## Why ❔

To let the node process batches both for Rollup and Validium pubdata commitment modes to enable Rollup<->Validium transitions.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


core/node/consistency_checker/src/tests/mod.rs
```diff
@@ -11,12 +11,16 @@ use zksync_eth_client::{clients::MockSettlementLayer, EthInterface, Options};
...
+    create_l1_batch, create_l1_batch_metadata, create_l2_block,
+    l1_batch_metadata_to_commitment_artifacts,
 };
 use zksync_types::{
-    aggregated_operations::AggregatedActionType, commitment::L1BatchWithMetadata,
-    protocol_version::ProtocolSemanticVersion, web3::Log, ProtocolVersion, ProtocolVersionId, H256,
-    L2_BRIDGEHUB_ADDRESS,
+    aggregated_operations::AggregatedActionType,
+    block::L2BlockHeader,
+    commitment::{L1BatchCommitmentMode::Validium, L1BatchWithMetadata, PubdataType},
```
slowli, line 20: Bikeshedding: Subjectively, importing variant names is confusing and should be avoided.

docs/src/guides/external-node/11_setup_for_other_chains.md
```diff
@@ -45,6 +45,11 @@ snapshots are stored and update the value of `EN_SNAPSHOTS_OBJECT_STORE_BUCKET_B
 Chains other than ZKsync Era aren't currently running consensus(as of December 2024). You need to disable it by removing
 `--enable-consensus` flag from `entrypoint.sh` invocation in docker-compose
 
-## 5. (Validium chains only) Set `EN_L1_BATCH_COMMIT_DATA_GENERATOR_MODE`
+## 5. (Validium chains only) Enable and configure DA fetcher
 
-For validium chains, you need to set `EN_L1_BATCH_COMMIT_DATA_GENERATOR_MODE: "Validium"`
+For Validium ENs to function properly, the Data Availability fetcher must be enabled and configured. It has to be added
+to the components list, e.g. for a standard list of components and DA fetcher it would be `--components=all,da_fetcher`.
```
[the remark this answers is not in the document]
dimazhornyk: We can fetch the genesis config from the main node, and detect a Validium EN, but that doesn't work with Rollup<->Validium migrations.

If it is not started - the `consistency_checker` won't progress and emit warnings that `da_fetcher` might be missing.

Another option would be to run the DA fetcher by default, it will just be doing nothing for rollups, but that is additional load on the DB, so I'm not sure if it is a right thing to do

core/node/consistency_checker/src/tests/mod.rs
dimazhornyk: if you meant `L1BatchCommitmentMode::Validium` - then totally agree, already removed

core/lib/dal/src/blocks_dal.rs
```diff
@@ -193,6 +193,41 @@ impl BlocksDal<'_, '_> {
...
+    ) -> DalResult<Option<L1BatchNumber>> {
+        let row = sqlx::query!(
+            r#"
+            SELECT
+                MIN(number) AS "number"
+            FROM
+                l1_batches
+            WHERE
+                is_sealed
+                AND number > $1
```
slowli, line 212: Is the strict inequality intentional? Just from the arg naming, I'd assume that `min_number` is inclusive.

core/lib/dal/src/blocks_dal.rs
```diff
@@ -193,6 +193,41 @@ impl BlocksDal<'_, '_> {
...
+                    FROM miniblocks
+                    WHERE l1_batch_number = l1_batches.number
+                    ORDER BY miniblocks.number
+                    LIMIT 1
+                ) != 'Rollup'
+            "#,
+            min_number.0 as i32
+        )
+        .instrument("get_earliest_l1_batch_number")
+        .report_latency()
```
slowli, line 224: Nit: You can add `.with_arg("min_number", &min_number)` for better reporting.

core/lib/dal/src/models/storage_block.rs
```diff
@@ -625,3 +625,17 @@ impl ResolvedL1BatchForL2Block {
         self.block_l1_batch.unwrap_or(self.pending_l1_batch)
     }
 }
+
+pub struct StoragePubdataParams {
```
slowli, line 629: Nit: Please restrict visibility to `pub(crate)` to make unused fields (should they occur in the future) more visible.

core/node/node_sync/src/data_availability_fetcher/mod.rs
```diff
@@ -115,20 +114,17 @@ impl DataAvailabilityFetcher {
         let last_l1_batch_with_da_info = storage
             .data_availability_dal()
             .get_latest_batch_with_inclusion_data()
+            .await?
+            .unwrap_or(L1BatchNumber::default());
```
slowli, line 118: Bikeshedding: `L1BatchNumber(0)` is more transparent.

core/node/node_sync/src/data_availability_fetcher/mod.rs
```diff
@@ -115,20 +114,17 @@ impl DataAvailabilityFetcher {
...
-                .max(L1BatchNumber(1)); // if only the genesis batch is in the storage, we should start from the first one, skipping the genesis
-
-            tracing::debug!("No L1 batches with DA info present in the storage; will fetch the earliest batch #{earliest_l1_batch}");
-            earliest_l1_batch
+        let Some(l1_batch_to_fetch) = l1_batch_to_fetch else {
+            tracing::debug!("No L1 batches to fetch DA info for");
+            return Ok(None);
         };
 
         Ok(if l1_batch_to_fetch <= last_l1_batch {
```
slowli, line 130: AFAIU, this check no longer makes sense because `l1_batch_to_fetch` cannot exceed `last_l1_batch`.

docs/src/guides/external-node/11_setup_for_other_chains.md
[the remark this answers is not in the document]
slowli: > but that is additional load on the DB

Do I understand correctly that the load will grow with time because of how `get_first_validium_l1_batch_number()` is implemented (i.e., it'll scan more and more batches with time, and `last_l1_batch_with_da_info` won't move)? I wonder how much effort is needed to avoid that. AFAIU, the flag whether a batch is a validium one never changes, so conceptually we could add an index on `l1_batches` for efficient queries. Even without that, we don't need to rescan batches we've already scanned, so even a primitive in-process cache could make performance tolerable (only the initial scan would be expensive).

core/node/consistency_checker/src/tests/mod.rs
```diff
@@ -554,14 +609,25 @@ async fn checker_processes_pre_boojum_batches(
...
+            if commitment_mode == L1BatchCommitmentMode::Validium {
+                l2_block.pubdata_params.pubdata_type = PubdataType::NoDA
+            }
+            (batch.to_owned(), l2_block)
+        })
+        .collect::<Vec<_>>();
+
     // Add new batches to the storage.
-    for save_action in save_actions_mapper(&l1_batches) {
+    for save_action in save_actions_mapper(batches_with_blocks.as_slice()) {
```
slowli, line 630: Nit: Why not `&batches_with_blocks`?

core/lib/dal/src/blocks_dal.rs
dimazhornyk: yeah, it is intentional because the value passed in the arguments is the last batch that was already processed
but I agree that the naming was confused, renamed it to `last_processed_batch`

core/node/consistency_checker/src/tests/mod.rs
```diff
@@ -554,14 +609,25 @@ async fn checker_processes_pre_boojum_batches(
...
 
     let (stop_sender, stop_receiver) = watch::channel(false);
     let checker_task = tokio::spawn(checker.run(stop_receiver));
 
+    let batches_with_blocks = l1_batches
+        .iter()
+        .map(|batch| {
+            let mut l2_block = create_l2_block(batch_to_block_number(batch));
+            if commitment_mode == L1BatchCommitmentMode::Validium {
+                l2_block.pubdata_params.pubdata_type = PubdataType::NoDA
```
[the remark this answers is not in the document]
dimazhornyk: added

core/node/eth_sender/src/aggregator.rs
```diff
@@ -418,9 +418,52 @@ impl Aggregator {
         )
         .await;
 
-        batches.map(|batches| {
-            AggregatedOperation::Commit(last_committed_l1_batch, batches, self.pubdata_da)
-        })
+        let batches = batches?;
+
+        // Note: the line below only works correctly during rollup <-> validium transitions
+        // if the limit of commit operation is set to 1.
```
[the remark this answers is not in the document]
dimazhornyk: I added a code block at the beginning of the function that checks that the limit is 1
It doesn't look pretty, but you are right that it would be at least easy to understand what is wrong

docs/src/guides/external-node/11_setup_for_other_chains.md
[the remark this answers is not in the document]
dimazhornyk: yes, that was my initial thinking. also, adding to the original answer - we need the right config for `da_fetcher`, so it won't be easy to leave it in a list of default components unless we want to provide a `da_fetcher` config even for rollups

> add an index on `l1_batches`

Pubdata mode is derived from the `l2_blocks`' `pubdata_type` field, so we can't create an index on `l1_batches` to cover that

but I added the "primitive in-process cache" to make all queries after the first one more efficient, lmk if that doesn't look good enough

core/tests/recovery-test/src/index.ts
```diff
@@ -123,6 +123,10 @@ export enum NodeComponents {
     WITH_TREE_FETCHER_AND_NO_TREE = 'core,api,tree_fetcher'
 }
 
+export function withDAFetcher(components: NodeComponents): string {
+    return components.toString().concat(',da_fetcher');
```
slowli, line 127: Nit: `components` is already a string (enums are syntactic sugar in TypeScript), so `toString()` is redundant. Also, `concat()` can be replaced with `+`.

core/tests/recovery-test/src/index.ts
```diff
@@ -168,11 +172,14 @@ export class NodeProcess {
         logsFile: FileHandle | string,
         pathToHome: string,
         components: NodeComponents = NodeComponents.STANDARD,
-        chain: string
+        chain: string,
+        deploymentMode?: string
     ) {
         const logs = typeof logsFile === 'string' ? await fs.open(logsFile, 'a') : logsFile;
+        let componentsArr = deploymentMode === 'Validium' ? [withDAFetcher(components)] : [components];
```
slowli, line 179: Dumb question: Would it harm the test to always run the node with the `da_fetcher` component here (or more extremely, enable it by default in the node binary)? Do we want to test that the node can function w/o this component?

slowli (approved):

core/tests/recovery-test/src/index.ts
dimazhornyk: if we enable it by default in the node - it will crash when the DA client config is not present. and we don't want people running rollup nodes coming up with dummy DA client configs

same with running it for all the tests - rollup configs don't have the necessary values, so the tests will crash as well

PR #3820, opened 2025-04-08
test: Add integration tests for EVM contracts (pt. 2)

slowli:
## What ❔

- Adds more integration tests for EVM contracts.
- Allows to run separate integration test suites (e.g., `zkstack dev test integration --verbose evm`).

## Why ❔

- More test coverage is better.
- Ability to run separate test suites improves DevEx.

## Is this a breaking change?

- [ ] Yes
- [x] No

## Operational changes

No operational changes.

## Checklist

- [x] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [x] Tests for the changes have been added / updated.
- [x] Code has been formatted via `zkstack dev fmt` and `zkstack dev lint`.

feat(zkstack): Allow to run separate integration test suites

PR #3833, opened 2025-04-11
fix(api): Return deployment nonce for contracts in `eth_getTransactionCount`

slowli:
## What ❔

Returns the deployment nonce for contracts in `eth_getTransactionCount`.

## Why ❔

Previously, `eth_getTransactionCount` always returned 0 for contract addresses. This doesn't match Ethereum behavior.

## Is this a breaking change?

- [x] Yes (technically; assuming the current incorrect behavior is relied upon)
- [ ] No

## Operational changes

No operational changes.

## Checklist


core/node/api_server/src/utils.rs
```diff
@@ -261,6 +266,103 @@ async fn fill_receipts_with_unknown_nonce(
...
+        } else {
+            Self::External(ExternalAccountType::Custom)
+        };
+        Ok((ty, h256_to_u256(full_nonce)))
+    }
+}
+
+#[derive(Debug, Clone)]
+pub(crate) struct AccountTypesCache {
+    cache: LruCache<Address, AccountType>,
```
slowli, line 319: Maybe, having a cache is overcomplicating things? I'm reasonably sure it works correctly, and I expect it to be useful. OTOH, if we can afford having 3 storage accesses (within a single Postgres query) on every `eth_getTransactionCount` call, the additional complexity can be avoided.

slowli (commented):
JFYI, the change intersects with the new EVM contract tests from https://github.com/matter-labs/zksync-era/pull/3820, but it'd be trivial to fix.

core/node/api_server/src/utils.rs
perekopskiy: I wouldn't add a cache but having is also ok with me

perekopskiy (dismissed):
> Is this a breaking change: No (assuming the current incorrect behavior isn't relied upon)

I think we should put "Yes" actually, it will put more attention to this PR so if anyone relies on it, then there is bigger chance it will be noticed in time

perekopskiy (approved):

PR #3840, opened 2025-04-14
feat: add `criterion_capacity_filled` metric

perekopskiy:
## What ❔

Add metric that tracks filled capacities of conditional sealers

## Why ❔

Better observability

## Is this a breaking change?
- [ ] Yes
- [x] No

## Operational changes

None

## Checklist


core/node/state_keeper/src/seal_criteria/mod.rs
```diff
@@ -186,14 +186,26 @@ pub(super) trait SealCriterion: fmt::Debug + Send + Sync + 'static {
...
         tx_count: usize,
         l1_tx_count: usize,
         block_data: &SealData,
         tx_data: &SealData,
         protocol_version: ProtocolVersionId,
     ) -> SealResolution;
 
+    /// Returns fraction of the criterion's capacity filled in the batch.
+    /// If it can't be calculated for the criterion, then it should return `None`.
+    fn capacity_filled(
```
slowli, line 198: Bikeshedding: IMO, it would look more intuitive to return this as the extended output from `should_seal` (whether via a tuple, or a dedicated struct), so it's easier to keep calculations in sync. The only thing making this approach awkward is that in some cases (primarily `ExcludeAndSeal` resolution, I think), it would be necessary to pass the amended block data (with the transaction excluded) to the sealer again to get the new fill stats.

core/node/state_keeper/src/metrics.rs
```diff
@@ -242,6 +244,10 @@ impl TxAggregationMetrics {
     pub fn l2_block_reason_inc(&self, reason: &L2BlockSealReason) {
         self.l2_block_reason[reason].inc();
     }
+
+    pub fn record_criterion_capacity(&self, criterion: &'static str, value: f64) {
+        self.criterion_capacity_filled[&criterion].observe(value)
+    }
```
slowli, line 250: Could it make sense to log these values as well in case we need more accurate data? Persisted metrics data for histograms is imprecise because of bucketing.

core/node/state_keeper/src/seal_criteria/mod.rs
perekopskiy: The downside is that we will mix up the critical path logic with metrics which is not good. I'd rather keep it as is.

core/node/state_keeper/src/metrics.rs
perekopskiy: I think 10% is precise enough and we don't need to log it

slowli: OK. Suggested because the additional volume of logs wouldn't be that high (the logs would be produced per batch).

core/node/state_keeper/src/seal_criteria/mod.rs
slowli: IMO, "mixing" logic is sort of the point in this case. As it's implemented right now, there's a risk that the returned metrics will become useless because there's nothing tying the metrics to the core business logic of the seal criteria. Probably not a huge deal given that the seal criteria would be overhauled by BoojumOS anyway 🤷

slowli (approved):

slowli (approved):

PR #3866, opened 2025-04-16
fix: address issue with evm verification with immutable ref 

dutterbutter:
## What ❔

- addresses issue with evm verification with immutable ref 

## Why ❔

- Could not verify contracts with immutable refs as it resulted in bytecode mismatch error 

## Is this a breaking change?
- [ ] Yes
- [x] No

## Operational changes

- No

## Checklist

### DIFF
--- a/core/lib/contract_verifier/src/lib.rs
+++ b/core/lib/contract_verifier/src/lib.rs
@@ -259,8 +259,7 @@ impl ContractVerifier {
         let bytecode_marker = BytecodeMarker::new(deployed_contract.bytecode_hash)
             .context("unknown bytecode kind")?;
         let artifacts = self.compile(request.req.clone(), bytecode_marker).await?;
-        let compiled_identifier =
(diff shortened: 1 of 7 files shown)


core/lib/contract_verifier/src/lib.rs
```diff
@@ -271,8 +270,16 @@ impl ContractVerifier {
             )
             .context("invalid stored EVM bytecode")?,
         };
+        let mut deployed_code = deployed_bytecode.to_vec();
+
+        // patch immutable references if any
```
popzxc, line 275:
```suggestion
        // If contract contains immutable references (e.g. places to be filled during constructor execution),
        // rewrite them with zeroes, as we can't know the values just yet.
        // We're checking the constructor arguments as well, so assuming tha constructor arguments
        // are the same, the immutable values should also be the same.
```

popzxc (approved):

perekopskiy (approved):

PR #3880, opened 2025-04-18
(feat): add eth_sender transaction's base fee max cap configuration

thomas-nguy:
## What ❔

Add a configuration to prevent eth_sender to send transactions on the settlement layer if the base fee exceed a limit

## Why ❔

Avoid commit/prove/exec batches when network fee become absurds.

Could be usefull for low tps chains 

## Is this a breaking change?
- [ ] Yes
- [x] No

## Operational changes
new config `max_acceptable_base_fee_in_wei`

## Checklist


core/node/eth_sender/src/eth_fees_oracle.rs
```diff
@@ -8,7 +8,9 @@ use zksync_eth_client::{ClientError, EnrichedClientError};
 use zksync_node_fee_model::l1_gas_price::TxParamsProvider;
 use zksync_types::eth_sender::TxHistory;
 
-use crate::{abstract_l1_interface::OperatorType, EthSenderError};
+use crate::{
+    abstract_l1_interface::OperatorType, EthSenderError, EthSenderError::ExceedMaxBaseFee,
```
slowli, line 12: Bikeshedding: Please don't import enum variants into the scope; it makes the code more difficult to read.

core/lib/protobuf_config/src/proto/config/eth_sender.proto
```diff
@@ -58,6 +58,7 @@ message Sender {
   reserved 23; reserved "priority_op_start_index";
   optional bool is_verifier_pre_fflonk = 24; // optional
   optional GasLimitMode gas_limit_mode = 25; // optional
+  optional uint64 max_acceptable_base_fee_in_wei = 26; // required; wei
```
slowli, line 61: AFAIU, this parameter has a reasonable default (maybe `u64::MAX` to replicate the previous behavior), so please don't make it required. Generally, introducing required config params is heavily discouraged because it'll break any node with a non-updated config.

core/node/eth_sender/src/eth_fees_oracle.rs
```diff
@@ -54,6 +57,16 @@ impl GasAdjusterFeesOracle {
         self.assert_fee_is_not_zero(blob_base_fee_per_gas, "blob");
         let blob_base_fee_per_gas = Some(blob_base_fee_per_gas);
 
+        if base_fee_per_gas > self.max_acceptable_base_fee_in_wei {
```
slowli, line 60: AFAIU, the `EthFees.base_fee_per_gas` returned from the method can still be greater than `self.max_acceptable_base_fee_in_wei` if there's a previously sent tx. Is this intended? Ditto for `calculate_fees_no_blob_sidecar()`.

slowli (changes requested):

core/node/eth_sender/src/eth_fees_oracle.rs
thomas-nguy: Can you clarify the alternative?

core/node/eth_sender/src/eth_fees_oracle.rs
thomas-nguy: Yeah this is intended. 

The goal is just to avoid transactions when the network base fee exceeds some threshold (i.e when someone is spamming L1)

It is okay to exceed `max_acceptable_base_fee_in_wei` for a few transactions if it is to solve the `replacement transaction underpriced issue`

thomas-nguy: On this topic, we sometime facing issue of transactions being stucked because previous sent tx exceed the rpc provider limit and the current logic does not allow new transactions to have lower fee to avoid `replacement transaction underpriced issue`

In that case, we need to manually clean the db which is a risky operation.

We probably to design a tool or have an auto-cleaning mechanism in place, but this is a completely different problem unrelated to this PR

tomg10: This is a very clever way to implement such cap, I really like it

core/node/eth_sender/src/eth_fees_oracle.rs
```diff
@@ -43,12 +46,28 @@ impl GasAdjusterFeesOracle {
             );
         }
     }
+
+    fn is_base_fee_exceeding_limit(&self, value: u64) -> bool {
+        if value > self.max_acceptable_base_fee_in_wei {
+            tracing::info!(
```
tomg10, line 52: nit: I think it makes sense to change this log level to warn

tomg10 (dismissed):

core/node/eth_sender/src/eth_fees_oracle.rs
slowli: Use the qualified variant name (`EthSenderError::ExceedMaxBaseFee` in this case):

```rust
return Err(EthSenderError::ExceedMaxBaseFee);
```

thomas-nguy: thanks 👍

slowli (approved):

PR #3885, opened 2025-04-22
feat(eth-sender): limit fees on resend

perekopskiy:
## What ❔

Use proper formula for calculating max acceptable fees in eth sender for every tx type
 
## Why ❔

To optimize eth-sender spendings

## Is this a breaking change?
- [ ] Yes
- [x] No

## Operational changes

None

## Checklist


core/node/eth_sender/src/eth_tx_manager.rs
```diff
@@ -275,7 +275,7 @@ impl EthTxManager {
             );
             return Err(error);
         }
-        Ok(signed_tx.hash)
+        Ok((signed_tx.hash, send_result.is_ok()))
```
Deniallugo, line 278: i'm super confused by this part. 
1. other methods of the code doesn't return any errors. 

Also the previous if checks this result again and return the error, so there are no chances to have `signed_tx.hash, false`

perekopskiy: it's an artifact from another PR; removed

core/node/base_token_adjuster/src/base_token_l1_behaviour.rs
```diff
@@ -270,9 +270,8 @@ impl BaseTokenL1Behaviour {
         prev_base_fee_per_gas: Option<u64>,
         prev_priority_fee_per_gas: Option<u64>,
     ) -> (u64, u64) {
-        // Use get_blob_tx_base_fee here instead of get_base_fee to optimise for fast inclusion.
-        // get_base_fee might cause the transaction to be stuck in the mempool for 10+ minutes.
-        let mut base_fee_per_gas = l1_params.gas_adjuster.as_ref().get_blob_tx_base_fee();
+        // Multiply by 2 to optimise for fast inclusion.
+        let mut base_fee_per_gas = l1_params.gas_adjuster.as_ref().get_base_fee(0) * 2;
```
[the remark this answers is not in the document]
perekopskiy: We can, I just left it in the state it's currently. I'm not going to work on base_token_adjuster as a part of this PR

core/node/eth_sender/src/eth_fees_oracle.rs
```diff
@@ -99,40 +115,37 @@ impl GasAdjusterFeesOracle {
         previous_sent_tx: &Option<TxHistory>,
         time_in_mempool_in_l1_blocks: u32,
     ) -> Result<EthFees, EthSenderError> {
+        const MIN_PRICE_BUMP_MULTIPLIER: f64 = 1.10 + 0.01;
```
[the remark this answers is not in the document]
perekopskiy: I initially added it as a safety margin. It shouldn't be needed since I added ceil() to calculations in the end. Removed

core/node/eth_sender/src/eth_fees_oracle.rs
```diff
@@ -99,40 +115,37 @@ impl GasAdjusterFeesOracle {
         previous_sent_tx: &Option<TxHistory>,
         time_in_mempool_in_l1_blocks: u32,
     ) -> Result<EthFees, EthSenderError> {
+        const MIN_PRICE_BUMP_MULTIPLIER: f64 = 1.10;
```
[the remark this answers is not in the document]
perekopskiy: yes, tbh I don't know why we were bumping it by 20% previously, nodes allow 10%

core/node/eth_sender/src/eth_fees_oracle.rs
```diff
@@ -99,40 +115,37 @@ impl GasAdjusterFeesOracle {
...
+
         // we cap it to not allow nearly infinite values when a tx is stuck for a long time
         let capped_time_in_mempool_in_l1_blocks = min(
             time_in_mempool_in_l1_blocks,
             self.time_in_mempool_in_l1_blocks_cap,
```
tomg10, line 123: The time_in_mempool_cap will no longer work as intended after those changes, it has way too high default now + Without changes, I think it only makes sense for non-blob txs. The cap serves a very important purpose though, without it, it the eth-sender is stuck for a long time, it will start sending almost infinite values

perekopskiy: I think I can add a new config that will cap `param_b ^ time_in_mempool`. I'd use something like 10 as a default. I'll also check how it impacts the simulation I did before. IMO for blob txs it will still make sense (I'll have to add, currently it's not used there), can you elaborate why do you think it isn't?

perekopskiy: Resends are not limited. The formula is `a * b^t * median`, so it's 3 resends if median doesn't change and I think that's ok. If there is a spike median will increase

core/node/fee_model/src/l1_gas_price/gas_adjuster/mod.rs
```diff
@@ -330,6 +330,21 @@ impl TxParamsProvider for GasAdjuster {
         last_block_base_fee * 875 / 1000
     }
 
+    fn get_next_block_minimal_blob_base_fee(&self) -> u64 {
+        let last_block_blob_base_fee = self.blob_base_fee_statistics.last_added_value();
+        let multiplied = last_block_blob_base_fee * U256::from(875);
+        if multiplied > U256::from(u64::MAX) {
```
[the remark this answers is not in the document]
perekopskiy: I will fix in some next PR

PR #3886, opened 2025-04-22
chore(deployment-allowlist): return full address in error message

dimazhornyk:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/core/node/api_server/src/tx_sender/result.rs
+++ b/core/node/api_server/src/tx_sender/result.rs
@@ -1,6 +1,6 @@
 use thiserror::Error;
 use zksync_multivm::interface::ExecutionResult;
-use zksync_types::{l2::error::TxCheckError, Address, U256};
+use zksync_types::{l2::error::TxCheckError, U256};
 use zksync_web3_decl::error::EnrichedClientError;
 
 use crate::execution_sandbox::{SandboxExecutionError, ValidationError};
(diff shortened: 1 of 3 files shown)


core/node/api_server/src/tx_sender/result.rs
```diff
@@ -70,7 +70,7 @@ pub enum SubmitTxError {
     #[error("internal error")]
     Internal(#[from] anyhow::Error),
     #[error("contract deployer address {0} is not in the allow list")]
-    DeployerNotInAllowList(Address),
+    DeployerNotInAllowList(String),
```
slowli, line 73: This change doesn't make sense from the type safety perspective. The issue is with the format specifier in the `error` attribute above (sorry I haven't caught it in the initial code review); it should be `{0:?}` (i.e., the `Debug` impl) or `{0:#x}` (i.e., the `LowerHex` impl, alternate presentation) instead of the `Display` impl. Both return a full, "0x"-prefixed address.

slowli (changes requested):

core/node/api_server/src/tx_sender/result.rs
dimazhornyk: makes sense, thanks for the explanation!
I changed it to `{0:?}`

deniallugo-ml:

PR #3895, opened 2025-04-23
feat: Adding 'rich_accounts' command to zkstack

mm-zk:
## What ❔

* Adding rich_accounts command to zkstack

## Why ❔

* Allows sending ETH from L1 to L2 in a single command - mostly useful during local testing.

By default - the command will send 1ETH from L1 rich account to L2.
```shell
zkstack dev rich-account
```

You can also pass an argument, to specify the destination account:

```shell
zkstack dev rich-account 0x7fc38899307B0C7C4f0A653418dbD3CB48085BFf
```

and use other flags (like --value) to specify the amounts etc.


zkstack_cli/Cargo.toml
```diff
@@ -37,6 +37,7 @@ zksync_consensus_crypto = "=0.9.0"
 
 # External dependencies
 anyhow = "1.0.82"
+alloy = "0.14.0"
```
Deniallugo, line 40: could you use ethers? 
I know it's deprecated, but adding new ethereum dependencies is not a good idea

Deniallugo: Could we then allocate some resources to remove ethers?

mm-zk: yes, we should.

Deniallugo (approved):

Deniallugo (approved):

PR #3898, opened 2025-04-24
feat(eth-sender): set `from_addr` for non-blob txs

perekopskiy:
## What ❔

Set `from_addr` for non-blob txs

## Why ❔

Makes data in DB less awkward, now it will be possible to get `from_addr` for any eth tx. Also helps with operator key rotation. I have tested locally that all you need to do to change operator (blob or non-blob) is to a) add new validator on L1; b) stop eth manager and wait for all txs to be confirmed; c) change wallet address and pk in config. No manual changes in DB.
Also, I've tested locally that code works with in-flight/unsent txs for which from_addr=null. I initialized server from `main`, created some bacthes so that there were some inflight and some unsent eth txs with from_addr=null. Then launched server from this branch and in-flight txs were confirmed; unsent txs were sent and confirmed.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/core/lib/dal/src/eth_sender_dal.rs
+++ b/core/lib/dal/src/eth_sender_dal.rs
@@ -26,7 +26,7 @@ impl EthSenderDal<'_, '_> {
     pub async fn get_inflight_txs(
         &mut self,
         operator_address: Address,
-        consider_null_operator_address: bool,
+        consider_null_operator_address: bool, // TODO (PLA-1118): remove this parameter
         is_gateway: bool,
     ) -> sqlx::Result<Vec<EthTx>> {
         let txs = sqlx::query_as!(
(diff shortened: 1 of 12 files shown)


core/lib/dal/src/eth_sender_dal.rs
```diff
@@ -25,7 +25,8 @@ pub struct EthSenderDal<'a, 'c> {
 impl EthSenderDal<'_, '_> {
     pub async fn get_inflight_txs(
         &mut self,
-        operator_address: Option<Address>,
+        operator_address: Address,
+        consider_null_operator_address: bool,
```
Deniallugo, line 29: let's add todo for removing this

Deniallugo (dismissed):

Deniallugo (approved):

Deniallugo (approved):

PR #3902, opened 2025-04-24
fix(api): Allow any deployment if initiator is in the allowlist

fedealconada:
## What ❔
* [x] Fixes issue in allowlist behaviour to allow any deployment if the initiator address is allowlisted.

## Why ❔
To prevent allowlisted addresses that deploy factories which deploy contracts in the same tx to fail.
E.g 
- Whitelisted address -> deploys Factory A -> deploys contract B (all in the same tx) should be allowed.
- Whitelisted address -> deploys Factory A -> deploys Factory B -> deploys contract C (all in the same tx) should be allowed.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


core/node/api_server/src/tx_sender/whitelist.rs
```diff
@@ -38,6 +38,16 @@ impl TxSink for WhitelistedDeployPoolSink {
         execution_output: &SandboxExecutionOutput,
         validation_traces: ValidationTraces,
     ) -> Result<L2TxSubmissionResult, SubmitTxError> {
+        // Check if the transaction initiator is allowlisted
+        let initiator = tx.initiator_account();
+        if self.shared_allow_list.is_address_allowed(&initiator).await {
```
slowli, line 43: Bikeshedding: IMO, it could make sense to extract a method returning a boolean flag whether to proceed with delegation to `master_pool_sink` to a separate private method, so that the delegation doesn't need to be duplicated.

slowli (dismissed):

core/node/api_server/src/tx_sender/whitelist.rs
fedealconada: makes sense! i've just added that, let me know your thoughts :)

slowli (approved):

dimazhornyk (approved):

PR #3916, opened 2025-04-29
[not for merge] (WIP) feat: EIP4337 & EIP7702

popzxc:


popzxc:
We decided to rely on zksync os instead.

PR #3919, opened 2025-04-29
feat(private-rpc): option to run private proxy in ZKStack CLI + moving Private Proxy files inside zksync-era repo

tomg10:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


zkstack_cli/crates/config/src/private_proxy_compose.rs
```diff
@@ -0,0 +1,58 @@
...
+        ports: Some(vec![format!("{}:{}", port, port)]),
+        volumes: None,
+        depends_on: None,
+        restart: None,
+        environment: Some(HashMap::from([
+            ("DATABASE_URL".to_string(), database_url.to_string()),
+            ("PORT".to_string(), port.to_string()),
+            (
+                "PERMISSIONS_YAML_PATH".to_string(),
+                "./example-permissions.yaml".to_string(),
```
tomg10, line 43: In the future those permissions will be managed using separate service

docker-compose.yml
```diff
@@ -19,6 +19,7 @@ services:
     command: postgres -c 'max_connections=1000'
     ports:
       - 127.0.0.1:5432:5432
+      - 172.17.0.1:5432:5432
```
tomg10, line 22: this is the ip of docker internal host

zkstack_cli/crates/zkstack/src/commands/private_rpc/mod.rs
```diff
@@ -0,0 +1,162 @@
...
+
+    let chain_config = ecosystem_config
+        .load_current_chain()
+        .context(MSG_CHAIN_NOT_INITIALIZED)?;
+    let chain_name = chain_config.name.clone();
+
+    let _dir_guard = shell.push_dir(chain_config.link_to_code.join("private-rpc"));
+
+    logger::info(msg_private_rpc_docker_image_being_built());
+    Cmd::new(cmd!(shell, "docker build -t private-rpc .")).run()?;
```
Deniallugo, line 108: are you sure it's convenient to rebuild docker during the development process? how this process should look like? 

and where is the part with database migrations?

tomg10: It will need to be built exactly once, next times docker will reuse cache, it builds in less than a minute on my machine with cold cache; compared to time needed to compile server it's imo suprisingly convenient. Alternative would be obviously to use already built image, but it would be confusing as we built everything else from source in zksync-era in ZK Stack. The migrations are run during "init" command using sqlx

private-rpc/create_token_example.sh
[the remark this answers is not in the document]
tomg10: It serves mostly as an "interactive" documentation. I want to move it into the readme in later PRs, when I start modifying the proxy code

docker-compose.yml
tomg10: I haven't found any different way to expose that docker internal ip :/ And without exposing it, docker isn't able to connect to the database.

Deniallugo: It has an alias. 
`host.docker.internal`

you can have a look how it is done for explorer

private-rpc/create_token_example.sh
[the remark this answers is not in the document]
Deniallugo: Could you do it straight away. 
don't want to keep it here.

zkstack_cli/crates/zkstack/src/commands/private_rpc/mod.rs
Deniallugo: What i'm saying here: 
1. It's a common way how we do it. Probably it'd make sense to have a few alternative way how to run it
2. It's not a common way for js development either.
3. If we need to do migrations during the init and it's the only way how to build new version, would it mean it will erase all data there? if no how we would erase the data?

tomg10: Having a separate DB for this service is very temporary, it will be removed in the coming weeks, so I prefer leaving it how it is. I will build a better UX around the "second iteration" where DB will be only in different service (some sort of admin panel)

private-rpc/create_token_example.sh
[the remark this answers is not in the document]
tomg10: corrected

zkstack_cli/crates/zkstack/src/commands/private_rpc/mod.rs
tomg10: as for way to build it, it's also likely to change, think of it as a PoC rather than a final version

RomanBrodetski (approved):
As agreed with @tomg10  - approving/merging it now to show to stakeholders early, but @tomg10 will follow up with @Deniallugo on potentially outstanding comments

RomanBrodetski (approved):

RomanBrodetski (approved):

PR #3921, opened 2025-04-30
Kl/number of roots support

kelemeno:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


core/node/state_keeper/src/updates/mod.rs
```diff
@@ -208,11 +208,19 @@ impl UpdatesManager {
...
             self.next_l2_block_params.is_none(),
             "next_l2_block_params cannot be set twice"
         );
+        let mut interop_roots = vec![];
+        for interop_root in l2_block_params.interop_roots {
+            if !self.l1_batch.interop_roots.contains(&interop_root) {
+                interop_roots.push(interop_root.clone());
+                self.l1_batch.interop_roots.push(interop_root);
+            }
+        }
```
Deniallugo, line 222: let's make it slightly cleaner, but the direction is correct, we have to save used roots in memory, until all l2 blocks are sealed

PR #3925, opened 2025-04-30
feat(state-keeper): [DRAFT] allow sub-second block time

perekopskiy:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


core/node/state_keeper/src/io/mempool.rs
```diff
@@ -206,9 +206,10 @@ impl StateKeeperIO for MempoolIO {
         // Block until at least one transaction in the mempool can match the filter (or timeout happens).
         // This is needed to ensure that block timestamp is not too old.
         for _ in 0..poll_iters(self.delay_interval, max_wait) {
-            // We cannot create two L1 batches or L2 blocks with the same timestamp (forbidden by the bootloader).
+            // We cannot create two L1 batches with the same timestamp (forbidden by the bootloader).
             // Hence, we wait until the current timestamp is larger than the timestamp of the previous L2 block.
             // We can use `timeout_at` since `sleep_past` is cancel-safe; it only uses `sleep()` async calls.
+            // TODO: sleep past not previous block timestamp but rather batch timestamp.
```
RomanBrodetski, line 212: (from slack):
>  if they set config.block_interval >= 1s then there will be different timestamps

@perekopskiy we may still need some miniblock-level logic

core/node/state_keeper/src/io/mempool.rs
```diff
@@ -206,28 +208,45 @@ impl StateKeeperIO for MempoolIO {
...
+                    deadline.into(),
+                    sleep_past(cursor.prev_l2_block_timestamp, cursor.next_l2_block),
+                )
+            } else {
+                tokio::time::timeout_at(
+                    deadline.into(),
+                    // TODO: understand if we can sleep past `prev_l1_batch_timestamp` in v29 before merging.
+                    // If not then revert whole `prev_l1_batch_timestamp` thing.
+                    // sleep_past(cursor.prev_l1_batch_timestamp, cursor.next_l2_block),
+                    sleep_past(cursor.prev_l2_block_timestamp, cursor.next_l2_block),
```
perekopskiy, line 237: Depends on whether https://github.com/matter-labs/era-contracts/pull/1493 will be included in v29 (should be decided by tomorrow). When reviewing please assume it will be merged and I will simply remove todo and use `sleep_past(cursor.prev_l1_batch_timestamp, cursor.next_l2_block)`.

core/node/state_keeper/src/updates/mod.rs
```diff
@@ -201,7 +200,7 @@ impl UpdatesManager {
             .take()
             .expect("next l2 block params cannot be empty");
         let new_l2_block_updates = L2BlockUpdates::new(
-            next_l2_block_params.timestamp,
+            next_l2_block_params.timestamp_ms,
```
Deniallugo: can we have duration?

core/node/state_keeper/src/updates/l2_block_updates.rs
```diff
@@ -174,11 +174,15 @@ impl L2BlockUpdates {
...
-            timestamp: self.timestamp,
+            timestamp: self.timestamp(),
             prev_block_hash: self.prev_block_hash,
             max_virtual_blocks_to_create: self.virtual_blocks,
         }
     }
+
+    pub fn timestamp(&self) -> u64 {
+        self.timestamp_ms / 1000
+    }
```
Deniallugo, line 185: is it ok to lose precision?

Deniallugo: If yes, i'd probably move it closer to VM, where we could make the decision. 
Or even better use Duration, instead of timestamp and use the necessary value in vm

core/lib/multivm/src/versions/vm_latest/utils/l2_blocks.rs
```diff
@@ -22,13 +22,19 @@ pub(crate) fn get_l2_block_hash_key(block_number: u32) -> StorageKey {
...
+    next_block: &L2BlockEnv,
+    protocol_version: ProtocolVersionId,
+) {
     if prev_block.number == 0 {
         // Special case for the first block it can have the same timestamp as the previous block.
         assert!(prev_block.timestamp <= next_block.timestamp);
     } else {
         assert_eq!(prev_block.number + 1, next_block.number);
-        assert!(prev_block.timestamp < next_block.timestamp);
+        if protocol_version.is_pre_fast_blocks() {
```
slowli, line 35: Dumb question: Will timestamps be no longer monotonic in the new protocol version? Or would they be non-strictly monotonic (`prev_block.timestamp <= next_block.timestamp`)?

core/node/node_sync/src/fetcher.rs
```diff
@@ -174,7 +174,8 @@ impl IoCursorExt for IoCursor {
                         block.l1_gas_price,
                     ),
                     first_l2_block: L2BlockParams {
-                        timestamp: block.timestamp,
+                        // It's ok that we lose info about millis since it's only used for sealing criteria.
+                        timestamp_ms: block.timestamp * 1000,
```
slowli: - I find it weird that fetched block info loses information about the timestamp *per se*. Are we 100% sure we won't need it?
- Just to be sure: Do I understand correctly that the VM is still fed the timestamp in seconds? So it may have several sequential blocks with the same (seconds) timestamp?

core/node/state_keeper/src/seal_criteria/io_criteria.rs
```diff
@@ -72,7 +76,8 @@ impl IoSealCriteria for TimeoutSealer {
 
     fn should_seal_l2_block(&mut self, manager: &UpdatesManager) -> bool {
         !manager.l2_block.executed_transactions.is_empty()
-            && millis_since(manager.l2_block.timestamp) > self.l2_block_commit_deadline_ms
+            && (millis_since_epoch() - manager.l2_block.timestamp_ms)
```
slowli: If we 100% sure we don't want to save the precise block timestamp, would it make sense to encapsulate it entirely within state keeper logic? I think all it would take is to make `L2BlockParams` fields crate-private, add a getter for timestamp and a constructor (based on the timestamp in seconds).

core/node/state_keeper/src/updates/l2_block_updates.rs
slowli: - If the intention is to not persist precise timestamps, I disagree; in this case, it makes sense to encapsulate sub-second precision as tightly in the state keeper as possible. *If* we might want to use the precise timestamps in the VM, then not persisting them doesn't make sense, so the PR has bigger issues.
- Also, it'll probably be `SystemTime`, not `Duration` 🙃

perekopskiy: The intention is to use precise timestamps in ms for block sealing purposes, and keep using seconds for VM. So I store milliseconds in UpdatesManager, and use .timestamp() that returns seconds when passing L2BlockEnv to VM.

core/lib/multivm/src/versions/vm_latest/utils/l2_blocks.rs
perekopskiy: They would be non-strictly monotonic. I see, it makes sense to assert it as well

core/node/node_sync/src/fetcher.rs
perekopskiy: - Currently milliseconds are only used for sealing criteria and I'm sure that we don't need them in fetched block for this purpose. One potential use-case for millis would be to return them in API. Overall, I think it's pointless, we stick to compatibility with ethereum that uses seconds in VM and block header, so even if we persist millis to DB and include them in fetched blocks they won't be validated by protocol nor consensus.
- yes

core/node/state_keeper/src/seal_criteria/io_criteria.rs
perekopskiy: makes sense

core/node/state_keeper/src/updates/mod.rs
perekopskiy: We can but how it will be better? In the end we will only use timestamp in seconds in some places and timestamp in milliseconds in others.

Deniallugo: yes, this the place will decide what it needs to be. Otherwise it's confusing and could lost precision accidentially

perekopskiy: but how will it be different? I already encapsulate field, so you choose to call timestamp() or timestamp_ms() to get needed precision. WIth SystemTime it will be the same

core/node/state_keeper/src/io/mod.rs
```diff
@@ -56,7 +56,42 @@ pub struct L2BlockParams {
     /// Note that it is the *maximal* number of virtual blocks that can be created within this L2 block since
     /// once the virtual blocks' number reaches the L2 block number, they will never be allowed to exceed those, i.e.
     /// any "excess" created blocks will be ignored.
-    pub virtual_blocks: u32,
+    virtual_blocks: u32,
+}
+
+impl L2BlockParams {
+    pub fn new_with_default_virtual_blocks(timestamp_ms: u64) -> Self {
```
slowli: Bikeshedding: IIUC, we want to use this constructor as much as possible, so it'd make sense to use a shorter name for it (e.g., `new`) and use the longer / more cumbersome name for the other constructor (e.g., `with_custom_virtual_block_count`).

core/node/state_keeper/src/io/mempool.rs
```diff
@@ -293,35 +307,45 @@ impl StateKeeperIO for MempoolIO {
         &mut self,
         cursor: &IoCursor,
         max_wait: Duration,
+        protocol_version: ProtocolVersionId,
```
slowli: Bikeshedding: This looks weird from the API perspective; I/O implementations should retain `ProtocolVersionId` they've returned as a part of L1 batch params, if they need it.

core/node/state_keeper/src/io/mempool.rs
```diff
@@ -293,35 +307,45 @@ impl StateKeeperIO for MempoolIO {
...
     ) -> anyhow::Result<Option<L2BlockParams>> {
-        // We must provide different timestamps for each L2 block.
-        // If L2 block sealing interval is greater than 1 second then `sleep_past` won't actually sleep.
+        // For versions <= v28 timestamps should be increasing for each L2 block.
+        // For versions >  v28 timestamps should be non-decreasing for each L2 block.
+        // - We sleep past `prev_l2_block_timestamp` for <= v28.
+        // - We sleep past `prev_l2_block_timestamp` for > v28 if `block_commit_deadline_ms >= 1000`,
+        //      so that timestamps for blocks are different.
```
slowli: Dumb question, and sorry for not raising it previously: Why does `block_commit_deadline_ms` also control the minimum spacing between blocks, and in such a roundabout way? It looks weird / unobvious. As a hypothetical example: assume that batches are sealed very fast (after a single transaction) by one of of non-timestamp criteria, and there's a steady stream of incoming transactions. IIUC, the time sealing criterion controlled by `block_commit_deadline_ms` is irrelevant in this case (blocks aren't sealed by a timeout), yet `block_commit_deadline_ms` still controls block spacing for some reason.

If we do need to have the control min spacing between blocks, I'd suggest a separate param for that (roughly speaking, something like `min_block_timestamp_spacing: 300ms`, but may be more coarse like `distinct_block_timestamps_in_seconds: true`).

perekopskiy: Hmm, on the one hand `min_block_timestamp_spacing`/`distinct_block_timestamps_in_seconds` gives more control but on the other it makes configuring sub-second blocks a bit more complex. Tbh I don't think there is a use-case for `min_block_timestamp_spacing`; I wanted to preserve `distinct_block_timestamps` property by default just in case, now I think it's not worth it. So I think I will remove "if block_commit_deadline_ms >= 1000" logic and will not introduce new config variable

core/lib/basic_types/src/protocol_version.rs
```diff
@@ -202,6 +202,17 @@ impl ProtocolVersionId {
...
         ProtocolVersionId::Version26
     }
+
+    pub fn is_pre_fast_blocks(&self) -> bool {
+        // Kludge for upgrade test. TODO: remove when v29 contracts are used.
+        if let Ok(v) = std::env::var("UPGRADE_TEST_INITIAL_VERSION") {
+            if v.parse::<usize>().is_ok_and(|v| v <= 28) {
+                return true;
+            }
+        }
```
slowli: - Don't like this approach at all, looks really easy to mess up / forget about. However, cannot think of a good alternative on a spot.
- Dumb question: What's the upgrade test error caused by? Looking at [logs](https://github.com/matter-labs/zksync-era/actions/runs/15579367959/job/43870837219), the immediate error is "Empty pubdata information" during batch finalization, but why is pubdata not set?

core/node/state_keeper/src/io/mempool.rs
```diff
@@ -175,153 +177,54 @@ impl StateKeeperIO for MempoolIO {
...
-        // We must provide different timestamps for each L2 block.
-        // If L2 block sealing interval is greater than 1 second then `sleep_past` won't actually sleep.
+        let protocol_version = self
+            .last_batch_protocol_version
+            .context("`last_batch_protocol_version` is missing")?;
+        // For versions <= v28 timestamps should be increasing for each L2 block.
+        // For versions >  v28 timestamps should be non-decreasing for each L2 block.
+        // - We sleep past `prev_l2_block_timestamp` for <= v28.
+        // - Otherwise, we do sanity sleep past `prev_l2_block_timestamp - 1`,
+        //      if clock returns consistent time then it shouldn't actually sleep.
```
slowli: Micro-nit: Unnecessary indentation.

core/node/state_keeper/src/io/mempool.rs
```diff
@@ -566,6 +471,139 @@ impl MempoolIO {
...
+            } else {
+                cursor
+                    .prev_l1_batch_timestamp
+                    .max(cursor.prev_l2_block_timestamp.saturating_sub(1))
+            };
+            let timestamp_ms = tokio::time::timeout_at(
+                deadline.into(),
+                sleep_past(timestamp_to_sleep_past, cursor.next_l2_block),
+            );
+            let Some(timestamp_ms) = timestamp_ms.await.ok() else {
```
slowli, line 544: Postgres connection is retained across this sleep point; could be better to drop and re-acquire it afterwards (esp. given that the use of `storage` below is conditional with a somewhat rare condition).

core/node/state_keeper/src/io/mod.rs
```diff
@@ -56,7 +56,42 @@ pub struct L2BlockParams {
...
+        self.timestamp_ms / 1000
+    }
+
+    /// The timestamp of the L2 block in milliseconds.
+    pub fn timestamp_ms(&self) -> u64 {
+        self.timestamp_ms
+    }
+
+    /// Mutable reference for the timestamp of the L2 block in milliseconds.
+    pub fn timestamp_ms_ref_mut(&mut self) -> &mut u64 {
```
slowli: Bikeshedding: `_ref` infix looks redundant; `_mut` is ordinarily used for mutable references.

core/node/state_keeper/src/keeper.rs
```diff
@@ -205,14 +214,15 @@ impl ZkSyncStateKeeper {
 
             // Start the new batch.
             next_cursor.l1_batch += 1;
-            (system_env, l1_batch_env, pubdata_params) = self
+            (system_env, l1_batch_env, pubdata_params, timestamp_ms) = self
```
slowli: Nit: Working with 4-element tuples looks cumbersome; it could make sense defining a struct to contain these 4 fields.

core/node/state_keeper/src/seal_criteria/io_criteria.rs
```diff
@@ -72,7 +72,8 @@ impl IoSealCriteria for TimeoutSealer {
 
     fn should_seal_l2_block(&mut self, manager: &UpdatesManager) -> bool {
         !manager.l2_block.executed_transactions.is_empty()
-            && millis_since(manager.l2_block.timestamp) > self.l2_block_commit_deadline_ms
+            && (millis_since_epoch() - manager.l2_block.timestamp_ms())
```
slowli: Nit: Saturating subtraction just in case of clock skew etc.?

core/lib/basic_types/src/protocol_version.rs
perekopskiy: - don't like it either, as an alternative, I can temporary change is_pre_fast_blocks to `self < ProtocolVersionId::Version30` and update to Version29 in final v29 PR where contract are updated. Regardless of the chosen approach it will be clean up fully with final v29 PR. Another alternative is to merge this PR into draft-v29 branch but I'd like to avoid it, mostly to avoid unnecessary conflicts
- Bootloader panics because fictive block has the same timestamp as the previous block and it happens before pubdata is set. Pubdata is unwrapped before vm result is checked

perekopskiy: Decided to change is_pre_fast_blocks=true for now

core/node/state_keeper/src/io/mempool.rs
```diff
@@ -566,6 +471,140 @@ impl MempoolIO {
...
+            let protocol_version = conn
+                .protocol_versions_dal()
+                .protocol_version_id_by_timestamp(curr_timestamp)
+                .await
+                .context("Failed loading protocol version")?;
+            let previous_protocol_version = conn
+                .blocks_dal()
+                .pending_protocol_version()
+                .await
+                .context("Failed loading previous protocol version")?;
```
slowli: Nit: Does this need to be queried on each loop iteration? (I understand that it was queried like that previously, but still.)

core/node/state_keeper/src/io/tests/mod.rs
```diff
@@ -710,7 +717,32 @@ async fn continue_unsealed_batch_on_restart(commitment_mode: L1BatchCommitmentMo
         .unwrap()
         .expect("no batch params generated");
 
-    assert_eq!(old_l1_batch_params, new_l1_batch_params);
```
slowli, line 713: Nit: You can reassign the batch for one set of params after checking it:

```rust
assert_eq!(
    old_l1_batch_params.first_l2_block.timestamp(),
    new_l1_batch_params.first_l2_block.timestamp()
);
*new_l1_batch_params.first_l2_block.timestamp_ms_mut() = 
    old_l1_batch_params.first_l2_block.timestamp_ms();
assert_eq!(new_l1_batch_params, old_l1_batch_params);
```

core/node/state_keeper/src/keeper.rs
```diff
@@ -363,7 +365,14 @@ impl ZkSyncStateKeeper {
         tokio::select! {
             hash_result = self.io.load_batch_state_hash(cursor.l1_batch - 1) => {
                 let previous_batch_hash = hash_result.context("cannot load state hash for previous L1 batch")?;
-                Ok(params.into_env(self.io.chain_id(), contracts, cursor, previous_batch_hash))
+                let timestamp_ms = params.first_l2_block.timestamp_ms();
+                let (system_env, vm_batch_env, pubdata_params) = params.into_env(self.io.chain_id(), contracts, cursor, previous_batch_hash);
```
slowli: Nit: Since `timestamp_ms` can be obtained from `params`, it's possible to return `BatchEnv` from `into_env()` directly.

core/node/state_keeper/src/keeper.rs
```diff
@@ -38,6 +38,14 @@ use crate::{
 /// we only need it to not block on waiting indefinitely and be able to process cancellation requests.
 pub(super) const POLL_WAIT_DURATION: Duration = Duration::from_secs(1);
 
+#[derive(Debug)]
+pub(super) struct BatchEnv {
```
slowli: Bikeshedding: IMO, this name can be confused with `L1BatchEnv`. I'd name this something like `VmInitParams` to differentiate.

slowli (dismissed):

core/node/state_keeper/src/keeper.rs
perekopskiy: I've chosen `BatchInitParams`, don't like `Vm*` prefix since it's not only used for vm but also for updates manager

PR #3928, opened 2025-04-30
fix: Fix issue with EVM bytecodes not being stored in DB

0xVolosnikov:
## What ❔

During deployment of EVM contracts bytecode is created on-chain. It is possible that contract will be deployed and called, but tx will revert in the end, so bytecode won't be published.

## Is this a breaking change?
- [ ] Yes
- [X] No

## Checklist


0xVolosnikov:
Requires https://github.com/matter-labs/zksync-protocol/pull/162

perekopskiy (approved):

PR #3941, opened 2025-05-02
perf(state-keeper): Parallelize loading storage log chunks when recovering SK cache

slowli:
## What ❔

- Parallelizes loading storage log chunks when recovering SK cache.
- Refactors `RocksdbStorage` API to be clearer.

## Why ❔

At least with some infra settings, loading storage log chunks is a bottleneck; loading a single chunk of 200k logs may take ~90s. Currently, the SK cache always uses a single DB connection, as per the general SK policy. Judging by the tree recovery, using multiple connections and loading chunks in parallel could speed up recovery ~5x.

## Is this a breaking change?

- [ ] Yes
- [x] No

## Operational changes

No operational changes.

## Checklist


core/lib/state/src/rocksdb/mod.rs
```diff
@@ -338,33 +290,38 @@ impl RocksdbStorage {
...
+    pub async fn synchronize(
+        mut self,
         storage: &mut Connection<'_, Core>,
         stop_receiver: &watch::Receiver<bool>,
         to_l1_batch_number: Option<L1BatchNumber>,
-    ) -> Result<(), RocksdbSyncError> {
-        let (_, mut current_l1_batch_number) = self
-            .ensure_ready(storage, Self::DESIRED_LOG_CHUNK_SIZE, stop_receiver)
-            .await?;
+    ) -> anyhow::Result<Option<Self>> {
```
itegulov, line 308: Bikeshedding: what's the reasoning behind moving away from `Result<T, Interrupted>` in favor of `Option<T>`? I always thought that former is preferred due to the following:
1. `Option` unlike `Result` is not `must_use` so it is easy to forget to handle it, especially when a function returns `Option<()>` or similar
2. `Option` is a very broad type that can easily be entangled with other meanings. Imagine a function that returns `Option<T>` (where `None` means value is missing) invoke interruptible function that returns `Option<U>` (where `None` means task was interrupted). It is easy to do `?` and collapse latter into former without meaning to do so.
3. This creates `Result<Option<T>, E>` which is unergonomic to work with

I think something like this would be fairly ergonomic at the lower logical levels:
```rust
enum OrInterrupted<E> {
    Internal(E),
    Interrupted,
}
```
And then higher up it all gets folded into `anyhow` regardless. This approach has its quirks of course and would be a lot more ergonomic if we had [try-trait-v2](https://github.com/rust-lang/rfcs/blob/master/text/3058-try-trait-v2.md)

To finish this (already excessively bikeshedding) comment, I am not too opposed to `Option` but in this case shouldn't we get rid of `RocksdbSyncError` entirely? AFAIR there are only a couple of usages left that haven't already been refactored out by this PR.

itegulov (dismissed):

core/lib/state/src/rocksdb/mod.rs
slowli: This idea makes sense. I'd prefer to implement it in a separate PR though, so we can bikeshed name, semantics etc.

perekopskiy (approved):

perekopskiy (approved):

PR #3947, opened 2025-05-02
perf(db): Break loading storage logs for recovery into smaller chunks

slowli:
## What ❔

Uses smaller chunks (10,000 entries vs 200,000 used currently) to load storage logs during Merkle tree and state keeper cache recovery.

## Why ❔

Benchmarking indicates that breaking loading into smaller chunks makes it faster.

## Is this a breaking change?

- [ ] Yes
- [x] No

## Operational changes

No operational changes.

## Checklist


core/lib/dal/src/storage_logs_dal.rs
```diff
@@ -722,44 +723,62 @@ impl StorageLogsDal<'_, '_> {
...
+                key: H256::from_slice(&row.hashed_key),
+                value: H256::from_slice(&row.value),
+                leaf_index: row.index as u64,
+            }));
+
+            if fetched_count < QUERY_LIMIT {
+                break;
+            }
+            // `unwrap()` is safe: `entries` contains >= QUERY_LIMIT items.
+            let next_key = u256_to_h256(h256_to_u256(entries.last().unwrap().key) + 1);
```
perekopskiy, line 777: nit: it can overflow if `entries.last().unwrap().key` is `U256::MAX`. Cannot happen in practice but I would still add smth like `if entries.last().unwrap().key == U256::MAX { break; }

perekopskiy (dismissed):

perekopskiy (approved):

PR #3968, opened 2025-05-07
refactor: Introduce cancelation errors

slowli:
## What ❔

Introduces cancelation errors (i.e., a very lightweight form of structured concurrency). Uses these errors to improve control flow for some components (e.g., snapshot recovery).

## Why ❔

Improves DevEx.

## Is this a breaking change?

- [ ] Yes
- [x] No

## Operational changes

No operational changes.

## Checklist

### DIFF
--- a/core/lib/basic_types/src/errors.rs
+++ b/core/lib/basic_types/src/errors.rs
@@ -1,7 +1,7 @@
 use std::fmt;
 
 /// Represents an error in a stoppable task: either an internal / fatal error, or a task getting stopped
-/// after receiving a stop signal.
+/// after receiving a stop request.
 ///
 /// The [`Error`](std::error::Error) trait is intentionally not implemented for this enum to force users
 /// to treat the `Stopped` variant with care. Depending on the application, it may not be an error.
@@ -10,7 +10,7 @@ use std::fmt;
 pub enum OrStopped<E = anyhow::Error> {
     /// Internal error.
     Internal(E),
-    /// Stop after receiving a signal.
+    /// Graceful stop after receiving the corresponding request.
     Stopped,
 }
 
@@ -18,7 +18,7 @@ impl fmt::Display for OrStopped {
     fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             Self::Internal(err) => write!(formatter, "{err:#}"),
-            Self::Stopped => formatter.write_str("stopped by signal"),
+            Self::Stopped => formatter.write_str("gracefully stopped"),
         }
     }
 }
--- a/core/lib/basic_types/src/errors.rs
+++ b/core/lib/basic_types/src/errors.rs
@@ -1,0 +1,11 @@
+#[derive(Debug)]
+pub enum OrStopped<E = anyhow::Error> {
+    Internal(E),
+    Stopped,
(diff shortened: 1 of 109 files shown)


core/lib/snapshots_applier/src/lib.rs
```diff
@@ -685,9 +691,9 @@ impl<'a> SnapshotsApplier<'a> {
         .await?;
         tracing::info!("Chosen snapshot recovery strategy: {strategy:?} with status: {applied_snapshot_status:?}");
         let (created_from_scratch, snapshot_version) = match strategy {
-            SnapshotRecoveryStrategy::Completed => return Ok((strategy, applied_snapshot_status)),
             SnapshotRecoveryStrategy::New(version) => (true, version),
             SnapshotRecoveryStrategy::Resumed(version) => (false, version),
+            SnapshotRecoveryStrategy::Completed => (false, SnapshotVersion::Version1),
```
itegulov, line 696: Did you mean to hard code `Version1` here instead of `applied_snapshot_status`? Feels like a potential pitfall if we ever introduce a new version.

core/lib/basic_types/src/errors.rs
```diff
@@ -0,0 +1,62 @@
+use std::fmt;
+
+/// Represents an error in a stoppable task: either an internal / fatal error, or a task getting stopped
+/// after receiving a stop signal.
+///
+/// The [`Error`](std::error::Error) trait is intentionally not implemented for this enum to force users
+/// to treat the `Stopped` variant with care. Depending on the application, it may not be an error.
+/// Use the [`try_stoppable!` macro](crate::try_stoppable) to handle this variant by early-returning `Ok(())`.
+#[derive(Debug)]
+pub enum OrStopped<E = anyhow::Error> {
```
itegulov, line 10: Nite: I like that we move away from calling this "interruption", I would also not use "signal" in the docs as that is easy to confuse with POSIX signals

itegulov (dismissed):
Thanks, even better than what I had in mind :heart:

core/lib/basic_types/src/errors.rs
slowli: Re POSIX signals, I've thought about the same thing 🙃 As a counterpoint, rn in node binaries, stopping *is* initiated as a result of receiving a SIGINT / SIGTERM. What would be the preferred terms then? Something like "stop request"?

itegulov: > stopping is initiated as a result of receiving a SIGINT / SIGTERM

Also if one of the tasks exits prematurely (i.e. panics), no?

> What would be the preferred terms then? Something like "stop request"?

Yeah I think "stop request" is okay since we are going with the "stopped" language. For reference tokio's [`CancellationToken`](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html) calls it "signal a cancellation request" so we could maybe rephrase to "or a task getting stopped after being signaled to do so". In any case no strong opinion here.

core/lib/snapshots_applier/src/lib.rs
slowli: Refactored this place to avoid relying on such hacks.

itegulov (approved):

PR #3970, opened 2025-05-07
feat(v29): Remove custom DA

0xVolosnikov:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


bin/build_and_init_ecosystem
```diff
@@ -30,6 +30,9 @@ initialize_ecosystem() {
 ################################################################################
 # Run all jobs in parallel and capture their PIDs
 ################################################################################
+# TODO Remove generate genesis. IT should be done only once, not every time
+zkstack dev contracts
+zkstack dev generate-genesis -v
```
Deniallugo: this is very temporary while this PR is in a state where I can't merge it into either the mainline or the v30 branch

core/node/state_keeper/src/executor/tests/tester.rs
```diff
@@ -270,7 +269,15 @@ impl Tester {
             self.config.validation_computational_gas_limit;
         let mut batch_params = default_l1_batch_env(l1_batch_number.0, timestamp, self.fee_account);
         batch_params.previous_batch_hash = Some(H256::zero()); // Not important in this context.
-        (batch_params, system_params, PubdataParams::default())
+        (
+            batch_params,
+            system_params,
+            PubdataParams {
+                l2_da_validator_address: None,
```
0xVolosnikov: nit: This could be done as an enum with two options, wdyt? Not important

zkstack_cli/crates/zkstack/src/admin_functions.rs
```diff
@@ -513,7 +513,7 @@ pub async fn set_da_validator_pair(
     chain_id: u64,
     bridgehub: Address,
     l1_da_validator_address: Address,
-    l2_da_validator_address: Address,
+    l2_da_commitment_chema: L2DACommitmentScheme,
```
0xVolosnikov: `l2_da_commitment_chema` -> `l2_da_commitment_scheme`

core/node/state_keeper/src/executor/tests/tester.rs
Deniallugo: Generally, I like the idea, but will see about the implementation, because we need to have it as an optional fields in database and we need the correct representation in bootloader. 

As a result we have some hussle to convert it from 2 options from fatbase to enum and then from enum to two options when we are actually placing it to the memory.

I have to think about it more. Not ready to implement it straightaway. 

It looks better, but it complicate the thing

core/lib/basic_types/src/commitment.rs
```diff
@@ -98,8 +98,104 @@ impl FromStr for PubdataType {
     }
 }
 
-#[derive(Default, Copy, Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default, Serialize, Deserialize, Display)]
+#[repr(u8)]
+pub enum L2DACommitmentScheme {
+    #[default]
+    None = 0,
```
perekopskiy, line 104: From contracts: `/// @param NONE Invalid option.`. Do we need None for server?

etc/env/file_based/genesis.yaml
```diff
@@ -1,10 +1,10 @@
-genesis_protocol_semantic_version: 0.29.1
+genesis_protocol_semantic_version: 0.29.0
```
perekopskiy: nit: 0.30.0 I guess

core/lib/basic_types/src/commitment.rs
Deniallugo: I'd say we need it, because it's a signal, that DA has not been set on contracts. and we need to set DA

perekopskiy: Should it be handled in `get_zk_chain_on_chain_params` then maybe?

core/bin/zksync_server/src/node_builder.rs
```diff
@@ -129,6 +129,20 @@ impl MainNodeBuilder {
         }
     }
 
+    pub fn l2_da_commitment_scheme(&self) -> L2DACommitmentScheme {
+        match self.configs.da_client_config {
+            Some(DAClientConfig::NoDA) => L2DACommitmentScheme::EmptyNoDA,
+            Some(DAClientConfig::Avail(_)) => L2DACommitmentScheme::PubdataKeccak256,
+            Some(DAClientConfig::Celestia(_)) => L2DACommitmentScheme::PubdataKeccak256,
+            Some(DAClientConfig::Eigen(_)) => L2DACommitmentScheme::PubdataKeccak256,
+            Some(DAClientConfig::ObjectStore(_)) => L2DACommitmentScheme::BlobsAndPubdataKeccak256,
```
dimazhornyk: ObjectStore can only use NoDA validator, so we should probably do
`Some(DAClientConfig::ObjectStore(_)) => L2DACommitmentScheme::EmptyNoDA `

core/lib/dal/src/consensus/tests.rs
```diff
@@ -56,25 +56,23 @@ fn payload(rng: &mut impl Rng, protocol_version: ProtocolVersionId) -> Payload {
...
-                ),
-                pubdata_type: match rng.gen_range(0..2) {
+            PubdataParams::new(
+                L2PubdataValidator::CommitmentScheme(match rng.gen_range(1..3) {
+                    1 => L2DACommitmentScheme::BlobsAndPubdataKeccak256,
+                    2 => L2DACommitmentScheme::EmptyNoDA,
+                    3 => L2DACommitmentScheme::PubdataKeccak256,
+                    _ => unreachable!("Invalid L2DACommitmentScheme value"),
+                }),
+                match rng.gen_range(0..2) {
```
dimazhornyk: `match rng.gen_range(0..4)` if we want to cover all of them

core/bin/zksync_server/src/node_builder.rs
```diff
@@ -129,6 +129,20 @@ impl MainNodeBuilder {
         }
     }
 
+    pub fn l2_da_commitment_scheme(&self) -> L2DACommitmentScheme {
```
dimazhornyk, line 132: the logic of this function doesn't always stand: some chain might use Celestia but doesn't verify it onchain, so it will have a `L2DACommitmentScheme::EmptyNoDA`

Deniallugo: that's very bad news...

core/bin/zksync_server/src/node_builder.rs
```diff
@@ -129,6 +129,20 @@ impl MainNodeBuilder {
...
+            Some(DAClientConfig::Avail(_)) => L2DACommitmentScheme::PubdataKeccak256,
+            Some(DAClientConfig::Celestia(_)) => L2DACommitmentScheme::PubdataKeccak256,
+            Some(DAClientConfig::Eigen(_)) => L2DACommitmentScheme::PubdataKeccak256,
+            Some(DAClientConfig::ObjectStore(_)) => L2DACommitmentScheme::BlobsAndPubdataKeccak256,
+            None => {
+                tracing::info!("DAClientConfig is not specified, setting L2DACommitmentScheme to BlobsAndPubdataKeccak256");
+                L2DACommitmentScheme::BlobsAndPubdataKeccak256
+            }
+        }
+    }
```
dimazhornyk, line 161: this is much less readable, this way we can determine if the NoDA validator is used
```suggestion
    pub fn l2_da_commitment_scheme(&self) -> anyhow::Result<L2DACommitmentScheme> {
        let da_dispatcher_config =
            self.configs
                .da_dispatcher_config
                .as_ref()
                .ok_or(anyhow::anyhow!(
                    "DA dispatcher config is required to determine L2 DA commitment scheme"
                ))?;

        Ok(
            match (
                self.configs.da_client_config.clone(),
                da_dispatcher_config.use_dummy_inclusion_data,
            ) {
                (Some(DAClientConfig::NoDA), _) => L2DACommitmentScheme::EmptyNoDA,
                (_, true) => L2DACommitmentScheme::EmptyNoDA,
                (Some(DAClientConfig::Avail(_)), false) => L2DACommitmentScheme::PubdataKeccak256,
                (Some(DAClientConfig::Celestia(_)), false) => {
                    L2DACommitmentScheme::PubdataKeccak256
                }
                (Some(DAClientConfig::Eigen(_)), false) => L2DACommitmentScheme::PubdataKeccak256,
                (Some(DAClientConfig::ObjectStore(_)), _) => L2DACommitmentScheme::EmptyNoDA,
                (None, _) => {
                    tracing::info!("DAClientConfig is not specified, setting L2DACommitmentScheme to BlobsAndPubdataKeccak256");
                    L2DACommitmentScheme::BlobsAndPubdataKeccak256
                }
            },
        )
    }
```

core/bin/zksync_server/src/node_builder.rs
```diff
@@ -326,6 +340,7 @@ impl MainNodeBuilder {
                     .genesis_config
                     .l1_batch_commit_data_generator_mode,
                 dummy_verifier: self.genesis_config.dummy_verifier,
+                config_l2_da_commitment_scheme: self.l2_da_commitment_scheme(),
```
dimazhornyk, line 360:
```suggestion
                config_l2_da_commitment_scheme: self.l2_da_commitment_scheme()?,
```

etc/env/file_based/genesis.yaml
Deniallugo: for making it 0.30.0 i need to add new protocol version properly and it's not an aim of this or

core/bin/zksync_server/src/node_builder.rs
Deniallugo: based on the conversation below i updated the code

core/lib/basic_types/src/commitment.rs
Deniallugo: Based on conversation with @dimazhornyk and @0xVolosnikov we end up, that we have to have None. But we can solve it on the level above and now it's not allowed to pass `None` to state keeper, but allowed to handle it inside eth sender

core/lib/multivm/src/versions/vm_latest/bootloader/utils.rs
```diff
@@ -220,13 +220,29 @@ fn bootloader_memory_input(
...
     protocol_version: ProtocolVersionId,
 ) -> Vec<u8> {
-    let l2_da_validator_address = pubdata_builder.l2_da_validator();
     let operator_input = pubdata_builder.l1_messenger_operator_input(input, protocol_version);
 
-    ethabi::encode(&[
-        ethabi::Token::Address(l2_da_validator_address),
-        ethabi::Token::Bytes(operator_input),
-    ])
+    if protocol_version.is_pre_interop_fast_blocks() {
```
perekopskiy: IIUC it should be a new separate method for this. Ditto for all usages of `is_pre_interop_fast_blocks` in this PR

Deniallugo: sure, but i don't want to introduce it here

perekopskiy: maybe mark with some comment/todo to not forget

perekopskiy (approved):

zkstack_cli/crates/zkstack/src/commands/chain/deploy_l2_contracts.rs
```diff
@@ -236,7 +236,7 @@ pub async fn deploy_l2_da_validator(
     shell: &Shell,
```
dimazhornyk, line 236: can we completely remove this function? it is not needed in v30 if I understand correctly

dimazhornyk (commented):
looks good! usually the most problematic part with DA is ENs, but everything seems to be fine here

dimazhornyk (approved):

zkstack_cli/crates/zkstack/src/commands/chain/deploy_l2_contracts.rs
Deniallugo: No, we still need to deploy it...

PR #3974, opened 2025-05-08
feat: EN contract verifier api

alfarkas:
## What ❔

Adds support to the external node for the contract verifier api.

## Why ❔

This api currently only runs on the main node. This makes the main node the only one that can perform contract verifications. In case of not exposing to the public or not running the main node rpc/http api, is not possible to use the contract verification feature of the block explorer.
More about this in [this discussion](https://github.com/zkSync-Community-Hub/zksync-developers/discussions/1009).

## Is this a breaking change?
- [ ] Yes
- [x] No

## Operational changes

Contract verification api is run by default when running the EN.
A new component can be run providing its name to the components flag: `--components=contract_verification_api`

## Checklist


cytadela8 (dismissed):
LGTM, but needs codeowner review

core/bin/external_node/src/main.rs
```diff
@@ -92,6 +94,7 @@ impl Component {
                 Component::WsApi,
                 Component::Tree,
                 Component::Core,
+                Component::ContractVerificationApi,
```
Deniallugo, line 97: I don't think contract verifier should be in `all`, normally it's not required

alfarkas: Updated.

mauricio-b-r:
Hey @Deniallugo , did you had a chance to review this? We want to know if we should wait for this or we should proceed with the other option that @alfarkas commented in the previous message

cytadela8:
@mauricio-b-r Danil happens to be out this week. I will ping him on Monday :)

cytadela8:
Hi, providing here the result of our internal discussion:

Our issue with this PR is that the contract_verifier API should not be started as part of the EN binary. Although this is similar to what goes on the main node, we want to fix it rather than add to the problem by adding this API server to the EN binary. 

We will accept this/new PR if you would rather add the possiblity of launching contract verifier API to the contract_verifier binary.

cytadela8:
Closing for now. Fill free to reopen if you address the above point

PR #3975, opened 2025-05-08
fix: fix zkstack cli upgrade features + remove redundant CI workflow

kelemeno:
## What ❔

1. The zkstack did not compile with v27, v28 features flags, this PR fixes that.
2. We had a redundant CI workflow shared between prover and core parts (the one which is deleted in this PR), the workflow that was being done in it was present in [core](https://github.com/matter-labs/zksync-era/blob/b7aeab64ce5c915233a773542ef64e79bf3893ee/.github/workflows/ci-core-lint-reusable.yml#L6-L56) part of CI, and is now present in prover part as well (Lints section was added). This way we have to spin up one less container on each CI run.

## Why ❔

The zkstack did not compile with v27, v28 features flags.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


.github/workflows/ci-core-reusable.yml
```diff
@@ -94,6 +94,45 @@ jobs:
...
+          echo "RUSTC_WRAPPER=sccache" >> .env
+          echo RUN_CONTRACT_VERIFICATION_TEST=true >> .env
+
+      - name: Start services
+        run: |
+          ci_localnet_up
+
+      - name: Init
+        run: |
+          ci_run run_retried rustup show
```
Deniallugo, line 124: seems redundant

.github/workflows/ci-core-reusable.yml
```diff
@@ -94,6 +94,45 @@ jobs:
...
+      - name: Install zkstack
+        run: |
+          ci_run ./zkstack_cli/zkstackup/install -g --path ./zkstack_cli/zkstackup/zkstackup
+          ci_run zkstackup -g --local --cargo-features v28_precompiles
+
+      - name: Install v28
+        run: |
+          ci_run ./zkstack_cli/zkstackup/install -g --path ./zkstack_cli/zkstackup/zkstackup
+          ci_run zkstackup -g --local --cargo-features v28_precompiles
+
```
Deniallugo, line 135: they are the same. 
i assume you wanted to build v27 and v28 separately. 
better to unify it to only one with both flags

.github/workflows/ci-core-reusable.yml
```diff
@@ -94,6 +94,45 @@ jobs:
...
+          echo IN_DOCKER=1 >> .env
+          echo "SCCACHE_GCS_BUCKET=matterlabs-infra-sccache-storage" >> .env
+          echo "SCCACHE_GCS_SERVICE_ACCOUNT=gha-ci-runners@matterlabs-infra.iam.gserviceaccount.com" >> .env
+          echo "SCCACHE_GCS_RW_MODE=READ_WRITE" >> .env
+          echo "RUSTC_WRAPPER=sccache" >> .env
+          echo RUN_CONTRACT_VERIFICATION_TEST=true >> .env
+
+      - name: Start services
+        run: |
+          ci_localnet_up
```
Deniallugo, line 120: if we build it from scratch. 
i'd rather try to use just rust container without our custom containers.

Deniallugo (approved):

PR #3977, opened 2025-05-08
fix: fix integration

zkzoomer:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


core/node/api_server/src/web3/namespaces/zks.rs
```diff
@@ -413,47 +412,31 @@ impl ZksNamespace {
...
-                // Modify the last log proof metadata accordingly: its length gets decreased by 1
-                let metadata_index = batch_chain_proof.batch_proof_len as usize + 3;
-                let mut metadata = batch_chain_proof.proof[metadata_index].0;
-                let log_tree_height = metadata[1];
-                metadata[1] = log_tree_height.saturating_sub(1);
-                batch_chain_proof.proof[metadata_index] = H256::from_slice(&metadata);
-
-                // The returned proof specifies the GW batch number on L1 where this L2 tx was included and a mask for the log proof
-                // We need to instead use the batch number to be the GW block number where the block containing this L2 tx was executed
-                let block_number = self.get_gw_block_number(l1_batch_number).await?;
```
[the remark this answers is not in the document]
kelemeno: ah yes, you got rid of it below. But then we can get rid of confirmet_at_block as well.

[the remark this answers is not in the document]
zkzoomer: It is used somewhere else as you still need to get the GW block number gor when a L2 seals its batch

core/node/eth_watch/src/event_processors/appended_chain_batch_root.rs
```diff
@@ -216,13 +234,38 @@ impl BatchRootProcessor {
...
+            .get_last_sent_and_confirmed_eth_storage_tx(eth_tx_id.unwrap())
+            .await
+            .map_err(|err| anyhow::anyhow!("Execute tx not found: {}", err))?;
+
+        let Some(tx) = tx else {
+            return Err(EventProcessorError::Internal(anyhow::anyhow!(
+                "Execute tx not found"
+            )));
+        };
+        tx.confirmed_at_block
```
zkzoomer, line 257: @kelemeno here

core/node/api_server/src/web3/namespaces/zks.rs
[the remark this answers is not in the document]
kelemeno: ah I see that you still use it, just in a different place, makes sense

core/node/api_server/src/web3/namespaces/unstable/mod.rs
```diff
@@ -142,6 +142,53 @@ impl UnstableNamespace {
         }))
     }
 
+    pub async fn get_chain_log_proof_for_block_impl(
```
kelemeno, line 145: since the namespace is unstable, do you think it might make sense to unify the two (this and get_chain_log_proof_impl) functions, with an Option<BatchNumber>, Option< BlockNumber> input, and output the corresponding result? 

The inner logic is also quite similar iiuc

kelemeno: Also maybe add a LogProofTarget here as well, because the difference is not the block/batchnumber, but whether we append the L1 extension or not.

zkzoomer: I can look into it again. Originally I decided against it as you'd need to define a new type, an enum or something, just for this. That's because you're dealing with both an L1BatchNumber and L2BlockNumber.

kelemeno: I meant to allow to give both, i.e. Option<BlockNumber>, Option<BatchNumber>, and only one of them could be defined. With an enum how would you know which one the number represents?

zkzoomer: With an enum you could later match to either type, L1BatchNumber or L2BlockNumber, and you wouldn't need to add another input

core/lib/web3_decl/src/namespaces/unstable.rs
```diff
@@ -37,7 +38,7 @@ pub trait UnstableNamespace {
     #[method(name = "getChainLogProof")]
     async fn get_chain_log_proof(
         &self,
-        l1_batch_number: L1BatchNumber,
+        batch_or_block_number: BatchOrBlockNumber,
```
kelemeno, line 41: Can we change the interface like this? I think it changes the input type, no? Before it was a u64, now its an enum, which I'm not sure how it is represented. I think we should just change the variable name, and add an optional flag, that makes the interface backwards compatible

zkzoomer: I implemented custom serialization for the BatchOrBlockNumber enum, so you can call the rpc via:
`"method": "unstable_getChainLogProof",  "params": [1, 271]` - which maintains the old behavior, ie calling for an L1BatchNumber
`"method": "unstable_getChainLogProof",  "params": [{ "block": 15 }, 271]` - which uses new behavior, ie calling for an L2BlockNumber

kelemeno (approved):

PR #3983, opened 2025-05-08
feat(eigenda): EigenDA V2 M0

gianbelinche:
## What ❔
This PR adds the client for EigenDA V2M0.
It maintains the old EigenDA integration (aka EigenDA V1) renaming it to EigenDAV1M0.

## Why ❔
To support the latest version of EigenDA (V2 M0), while preserving compatibility with the existing client (V1 M0).

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


core/node/da_clients/src/eigenv2m0/client.rs
```diff
@@ -0,0 +1,135 @@
...
+        let blob_key = BlobKey::from_bytes(
+            bytes
+                .try_into()
+                .map_err(|_| anyhow::anyhow!("Failed to convert bytes to a 32-byte array"))
+                .map_err(to_non_retriable_da_error)?,
+        );
+        let eigenda_cert = self
+            .client
+            .get_inclusion_data(&blob_key)
+            .await
```
samlaf, line 105: I personally would have prefered if the eigenda client remains agnostic to zksync terminology, such that it can be used for other stacks.

Can we rename `client.get_inclusion_data` to `client.get_cert` or something like that? the mapping from cert=inclusion_data should only live in the zksync codebase imo

core/node/da_clients/src/eigenv2m0/client.rs
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    async fn ensure_finality(
+        &self,
+        dispatch_request_id: String,
+    ) -> Result<Option<FinalityResponse>, DAError> {
+        // TODO: return a quick confirmation in `dispatch_blob` and await here
+        Ok(Some(FinalityResponse {
+            blob_id: dispatch_request_id,
+        }))
```
samlaf, line 89: seems like the implementation is not finished?
@dimazhornyk I forget if you prefer to wait for the full cert (~10sec) in `dispatch_blob` and make `ensure_finality` a noop for v2, or we can have `ensure_finality` actually do the 10s polling.

samlaf: Also @dimazhornyk https://docs.zksync.io/zk-stack/running/validium needs an update, it doesn't contain the `ensute_finality` function

samlaf (commented):
Most of this PR looks like zksync-specific glue code so will let Dima make a broad-stroke review before nit-picking on the details (would def like to have better comments for the configurations, since I don't think a zksync engineer right now would be able to know how to populate an eigenda client config properly)

Otherwise we discussed in yesterday's meeting with @juanbono and @dimazhornyk and I think overall it would be best if there could be a single eigenda client, at the expense of having a larger config (encompassing both v1/v2, and eventually secure v2 fields).

Also don't think there should be mentions of m0/m1 in this codebase. Don't think zksync uses the m0/m1 terminology so it would just be confusing I think.

dimazhornyk (commented):
looks great, special thanks for the amazing README
let's just decide on whether we need to keep the v1 and structure the config accordingly

dimazhornyk:
The code looks good, but I thought we agreed to remove V1 completely, both the config and an implementation.
I don't see any benefits of having the combined config if it is a breaking change anyway.

To make it smooth on the operations side - we can have an ENV/file-based configs in k8s that have fields for both, but the node will only support one of them (older versions would use V1 and newer versions would use V2) without dividing them on V1 and V2

dimazhornyk (approved):
looks good, I'm only concerned about the growing `Cargo.lock`
would be great to isolate some dependencies via `features` in rust-eigenda, but I'm ok with merging this as is

core/deny.toml
```diff
@@ -49,6 +49,39 @@ crate = "ring"
 expression = "ISC AND MIT AND OpenSSL"
 license-files = [{ path = "LICENSE", hash = 0xbd0eed23 }]
 
+[[licenses.exceptions]]
```
dimazhornyk: Would adding the "BUSL-1.1" to `allow` list above work as well? that would look much cleaner

dimazhornyk (commented):
looks good to me, I would only look into the `deny.toml` so we don't have to maintain the list of all eigen-rust crates

core/deny.toml
gianbelinche: Yes that would work, but that would imply that any crate with that license will be accepted, if you are ok with that I'll make the change

dimazhornyk: I will get a confirmation from another MatterLabs reviewer, but I think it is fine

core/node/da_clients/src/node/eigen.rs
```diff
@@ -39,28 +31,8 @@ impl WiringLayer for EigenWiringLayer {
...
-    pool: ConnectionPool<Core>,
-}
-
-#[async_trait::async_trait]
-impl BlobProvider for GetBlobFromDB {
-    async fn get_blob(&self, input: &str) -> Result<Option<Vec<u8>>, Box<dyn Error + Send + Sync>> {
-        let mut conn = self.pool.connection_tagged("eigen_client").await?;
-        let batch = conn
-            .data_availability_dal()
-            .get_blob_data_by_blob_id(input)
```
perekopskiy, line 62: nit: looks like `get_blob_data_by_blob_id` can be fully removed

core/node/da_clients/src/node/eigen.rs
```diff
@@ -26,9 +20,7 @@ impl EigenWiringLayer {
 }
 
 #[derive(Debug, FromContext)]
-pub struct Input {
-    master_pool: PoolResource<MasterPool>,
-}
+pub struct Input {}
```
perekopskiy: nit: remove `Input`, and replace `type Input = Input;` with `type Input = ();`

core/node/da_clients/src/eigen/EIGENDAV2_MIGRATION.md
```diff
@@ -0,0 +1,46 @@
+# Migration from V1 to V2 client
+
+If you updated your version of zksync-era, and your EigenDA config looks something like this:
```
perekopskiy: I think this guide should be changed. Instead of "If you updated your version of zksync-era", I would say "before you update your version of zksync-era you should put values needed for both version of the config; after you update you can remove old config values". This way there will be no downtime. Does it make sense?

gianbelinche: Done https://github.com/matter-labs/zksync-era/pull/3983/commits/87280460636faba5e1635710a16b4a33818516ad

core/node/da_clients/src/node/eigen.rs
gianbelinche: Done https://github.com/matter-labs/zksync-era/pull/3983/commits/87280460636faba5e1635710a16b4a33818516ad

core/node/da_clients/src/node/eigen.rs
gianbelinche: Done https://github.com/matter-labs/zksync-era/pull/3983/commits/87280460636faba5e1635710a16b4a33818516ad

perekopskiy (approved):

PR #3988, opened 2025-05-09
feat: add `en_getProtocolVersionInfo`, fix `eth_protocolVersion`

itegulov:
## What ❔

* Adds `en_getProtocolVersionInfo` that replaces `zks_getProtocolVersion` (to be removed after all ENs upgrade)
* Makes `eth_protocolVersion` return correct protocol semantic version

## Why ❔

Stop leaking internal structures in `zks_getProtocolVersion`

## Is this a breaking change?
- [x] Yes
- [ ] No

## Operational changes
Whitelist `en_getProtocolVersionInfo`

## Checklist


core/lib/dal/src/protocol_versions_web3_dal.rs
```diff
@@ -49,4 +52,45 @@ impl ProtocolVersionsWeb3Dal<'_, '_> {
...
+        Ok(storage_protocol_version.map(ProtocolVersionInfo::from))
+    }
+
+    pub async fn get_latest_protocol_version_info(&mut self) -> DalResult<ProtocolVersionInfo> {
+        let latest_version = self
+            .storage
+            .protocol_versions_dal()
+            .latest_semantic_version()
+            .await?
+            .unwrap();
```
slowli, line 91: Bikeshedding: I'd rather return `DalResult<Option<_>>` here and turn the `Ok(None)` variant into an error / `unreachable!` on the caller side. This would allow for more "localized" correctness reasoning.

core/node/api_server/src/web3/namespaces/en.rs
```diff
@@ -191,4 +193,27 @@ impl EnNamespace {
...
+            storage
+                .protocol_versions_web3_dal()
+                .get_protocol_version_info_by_id(id)
+                .await
+                .map_err(DalError::generalize)?
+        } else {
+            Some(
+                storage
+                    .protocol_versions_web3_dal()
+                    .get_latest_protocol_version_info()
```
slowli, line 212: Bikeshedding: Since `latest_semantic_version()` is used in the API server handlers, it may make sense to use it here as well (followed by `get_protocol_version_info_by_id`) instead of introducing a specialized method.

core/node/node_sync/src/client.rs
```diff
@@ -97,8 +97,8 @@ impl MainNodeClient for Box<DynClient<L2>> {
     async fn fetch_protocol_version(
         &self,
         protocol_version: ProtocolVersionId,
-    ) -> EnrichedClientResult<Option<api::ProtocolVersion>> {
-        self.get_protocol_version(Some(protocol_version as u16))
+    ) -> EnrichedClientResult<Option<api::ProtocolVersionInfo>> {
+        self.get_protocol_version_info(Some(protocol_version as u16))
```
slowli, line 101: Is it impossible to have an updated EN for a certain env with the main node not upgraded yet? If it *is* possible, `get_protocol_version_info()` will error. AFAIU, it's possible to use `get_protocol_version` as a fallback (`api:: ProtocolVersionInfo` is a subset of `api::ProtocolVersion`).

core/tests/ts-integration/tests/api/web3.test.ts
```diff
@@ -183,7 +183,7 @@ describe('web3 API compatibility tests', () => {
         ['net_peerCount', [], '0x0'],
         ['net_listening', [], false],
         ['web3_clientVersion', [], 'zkSync/v2.0'],
-        ['eth_protocolVersion', [], 'zks/1'],
+        ['eth_protocolVersion', [], 'zks/0.28.0'],
```
slowli, line 186: This will require frequent updates, right? Maybe, it makes sense to extract to a separate test and testing matching a regex like `^zks\/0\.\d+\.\d+$`?

core/node/node_sync/src/client.rs
itegulov: > Is it impossible to have an updated EN for a certain env with the main node not upgraded yet?

I was operating under that assumption, yes. @perekopskiy do you think that's a fair assumption?

It would probably be a bit awkward to try and use jsonrpsee in the way you propose but I can use `get_protocol_version` as a fallback and add `impl From<api::ProtocolVersion> for api:: ProtocolVersionInfo`

slowli: This assumption could be reasonable for outside EN operators, but IMO it's too shaky in the general case. E.g., in our infra, we have one of our ENs on the Era testnet and mainnet auto-updated from `main`.

itegulov: @slowli I have added a fallback just in case but I still suspect it is very unlikely to ever trigger since that can only happen during protocol upgrade. In any case PTAL

core/lib/types/src/api/mod.rs
```diff
@@ -773,6 +772,48 @@ impl ProtocolVersion {
...
+}
+
+impl TryFrom<ProtocolVersion> for ProtocolVersionInfo {
+    type Error = anyhow::Error;
+
+    fn try_from(value: ProtocolVersion) -> Result<Self, Self::Error> {
+        Ok(ProtocolVersionInfo {
+            minor_version: value
+                .minor_version
+                .ok_or(anyhow::anyhow!("missing minor protocol version"))?,
```
slowli, line 803: Nit: `.context("...")` is more idiomatic.

core/node/node_sync/src/client.rs
```diff
@@ -96,11 +96,36 @@ impl MainNodeClient for Box<DynClient<L2>> {
...
-        self.get_protocol_version(Some(protocol_version as u16))
+    ) -> EnrichedClientResult<Option<api::ProtocolVersionInfo>> {
+        match self
+            .get_protocol_version_info(Some(protocol_version as u16))
             .rpc_context("fetch_protocol_version")
             .with_arg("protocol_version", &protocol_version)
             .await
+        {
+            Ok(result) => Ok(result),
+            Err(err) => {
```
slowli, line 107: Can probably filter errors here; a vanilla node should return a JSON-RPC error with code `-32601` (method not found). OTOH, I'm not sure whether this error will be retained by proxies, so maybe it does make sense to be more general.

slowli (approved):

core/node/node_sync/src/client.rs
itegulov: Yeah, I considered that but wanted to stay on the safe side just in case. Also, I am logging the error so it's not completely lost.

slowli (approved):

Issue #3834, opened 2025-04-11
feat(contract-verifier): support new IPFS metadata


hedgar2017:
### 🌟 Feature Request

Starting from zksolc v1.5.13 and zkvyper v1.5.10, CBOR-encoded bytecode metadata contains an additional field `solc` with full information on compiler versions the bytecode was produced with.

#### 📝 Description

CBOR metadata JSON and binary representations are documented in these corresponding sections of the compiler docs:

[zksolc](https://matter-labs.github.io/era-compiler-solidity/latest/02-command-line-interface.html#--metadata-hash)
[zkvyper](https://matter-labs.github.io/era-compiler-vyper/latest/02-command-line-interface.html#--metadata-hash)

In order to make verification more reliable, the verifier must read compiler versions from the CBOR payload instead of verification request.

Please keep in mind that:

1. The `solc` field in CBOR metadata can be missing if the bytecode was produced with a compiler older than zksolc v1.5.13 and zkvyper v1.5.10.
2. The CBOR metadata can be missing altogether, as we still have the `keccak256` hash type which is not CBOR-encoded and doesn't include compiler version data.

In both cases, the verifier must revert to its old behavior.

#### 🤔 Rationale

Verification process will be more robust and less dependent on additional data that must be provided by user.

PR #4002, opened 2025-05-13
feat(contract_verifier): read compiler versions from cbor metadata if available

vasyl-ivanchuk:
## What ❔

Read compiler versions for contract verification from CBOR-encoded bytecode metadata, if available. PR for [this](https://github.com/matter-labs/zksync-era/issues/3834) issue.

## Why ❔

In order to make verification more reliable, the verifier must read compiler versions from the CBOR payload instead of the verification request. Verification process will be more robust and less dependent on additional data that must be provided by user.

## Is this a breaking change?
- [ ] Yes
- [X] No

## Checklist


core/lib/contract_verifier/src/lib.rs
```diff
@@ -353,6 +360,73 @@ impl ContractVerifier {
...
     }
 
+    /// Checks whether the compiler versions in the request match those in the CBOR metadata (if available).
+    /// If they do not match, the versions in the request are updated to match the metadata.
+    async fn update_request_compiler_versions(
+        &self,
+        request: &mut VerificationRequest,
+        deployed_metadata: &Option<DetectedMetadata>,
+    ) -> Result<(), ContractVerifierError> {
+        if let Some(DetectedMetadata::Cbor {
```
popzxc, line 370: Nit: this function does nothing if `deployed_metadata` is `None`, so it's better to accept `&DetectedMetadata` here and have `if let Some(..)` on the caller side.

core/lib/contract_verifier/src/lib.rs
```diff
@@ -353,6 +360,73 @@ impl ContractVerifier {
...
+                tracing::warn!(
+                    request_id = request.id,
+                    request_compiler,
+                    request_zk_compiler,
+                    metadata_compiler,
+                    metadata_zk_compiler,
+                    "Compiler versions in the request don't match the ones in the CBOR metadata. Updating the request."
+                );
+
+                let mut storage = self
```
popzxc, line 394: Why do we need to update the request in the database? It can be somewhat unexpected, IMO.
Assuming that bytecode is the same between requests in the database, choosing the compiler versions right before compilation should be idempotent, no?
So, we can check the version before resolving the compiler, e.g.:
- If bytecode has CBOR metadata, use data from the metadata.
- Otherwise, use values from the request.

core/lib/contract_verifier/src/lib.rs
vasyl-ivanchuk: Agreed, refactored!

core/lib/contract_verifier/src/lib.rs
vasyl-ivanchuk: Yeah, I was also concerned about updating the request in the DB. I added a comment [here](https://github.com/matter-labs/zksync-era/pull/4002/files#diff-3a93a498c0982e91b13d6d4a7a1597dd22a90dd045a10f52457a8f8b1fe43982R399):
```
// Update the verification request in the DB because other components depend on the compiler versions,
// such as Etherscan verification and block explorer.
```

There are two things that depend on the compiler versions in the request:
1. Etherscan verification that happens afterward. Right now, it fetches the verification request, maps it to the proper format, and sends it to Etherscan. I think Etherscan will eventually support CBOR metadata, but in the meantime, requests with incorrect compiler versions will fail.
2. Block Explorer UI that shows compiler versions from the request object:
image

An alternative I was thinking about is:
1. Add compiler versions to the `VerificationInfo`, which is conceptually correct.
2. Customize Etherscan verifier to fetch a `VerificationInfo` for the request and get compiler versions from it (one additional call to the DB).
3. Update the UI to read compiler versions from the `VerificationInfo` and if they don't exist, fallback to the request object.

Do you think it's worth doing it this way?

popzxc: Oh, I see. Both approaches seem hacky, but one you've chosen seems _less_ hacky. I guess let's stick to it, but I would suggest expanding the comment here, e.g. something like "Since we know the version that should be used for the contract, we override one provided in the request, so that if this request is interpreted by anyone else, they can get correct compilation results".

Additionally, I think we only need to override the request if verification was _successful_. What if CBOR metadata was provided but incorrect (e.g. placed by hand)?

Maybe a better flow would be:
1. Try compiling using CBOR metadata
2. If it succeeds, override the request and store verification info.
3. If it fails, try one more time using the versions provided in the requests.

This way we will probably avoid the situation where we don't do what user asked us to do.
WDYT?

vasyl-ivanchuk: I think it's unlikely that CBOR metadata will be constructed manually, and if it is - it's fine that the verification won't fallback to the previous behavior. OTOH, it's very easy to add, so I'll include it just in case.

core/lib/contract_verifier/src/lib.rs
```diff
@@ -353,6 +363,98 @@ impl ContractVerifier {
...
+                        // Update the request for downstream use
+                        request.req = updated_request;
+
+                        Ok(artifacts)
+                    }
+                    Err(err) => {
+                        tracing::warn!(
+                    request_id = request.id,
+                    "Failed to compile with the compiler versions from the metadata: {err}. Falling back to the original request."
+                );
```
popzxc, line 421: Nit: Missing formatting

core/lib/contract_verifier/src/lib.rs
```diff
@@ -353,6 +363,98 @@ impl ContractVerifier {
...
+                                request.req.compiler_versions.zk_compiler_version(),
+                            metadata_zk_compiler =
+                                updated_request.compiler_versions.zk_compiler_version(),
+                            "Updating request compiler versions in the DB."
+                        );
+
+                        // Since we know the version that should be used for the contract, we override one provided in
+                        // the request, so that if this request is interpreted by anyone else, they can get correct
+                        // compilation results
+                        self.update_request_compiler_versions(
```
popzxc, line 405: We should still check that the verification passes, right? E.g. it can compile but still not result in a match.

core/lib/types/src/contract_verification/api.rs
```diff
@@ -270,6 +316,21 @@ impl CompilationArtifacts {
             }
         }
     }
+
+    /// Returns true if the provided `deployed_code` matches the compiled code from the artifacts.
+    /// Otherwise, returns false.
+    pub fn match_bytecode(&self, bytecode_marker: &BytecodeMarker, bytecode: &[u8]) -> bool {
```
popzxc, line 322: Bikeshedding: `matches_bytecode`

core/lib/contract_verifier/src/lib.rs
```diff
@@ -353,6 +364,99 @@ impl ContractVerifier {
...
+                let updated_request = request
+                    .req
+                    .clone()
+                    .with_updated_compiler_versions(cbor_metadata);
+
+                match self
+                    .compile(updated_request.clone(), *bytecode_marker)
+                    .await
+                {
+                    Ok(artifacts) if artifacts.match_bytecode(bytecode_marker, deployed_code) => {
```
popzxc, line 391: Under the hood you calculate identifier, but you have it already: it can be passed instead of `detected_metadata`
(and I guess it'll make `match_bytecode` method redundant).

core/lib/contract_verifier/src/lib.rs
```diff
@@ -353,6 +364,99 @@ impl ContractVerifier {
...
+    /// Tries to use the compiler versions from the metadata if they are different from the ones in the request.
+    /// If the metadata is not present, or the versions are the same, uses the ones from the request.
+    async fn get_compilation_artifacts(
+        &self,
+        request: &mut VerificationRequest,
+        detected_metadata: Option<&DetectedMetadata>,
+        bytecode_marker: &BytecodeMarker,
+        deployed_code: &[u8],
+    ) -> Result<CompilationArtifacts, ContractVerifierError> {
+        match detected_metadata {
```
popzxc, line 377: There is a bit too much nesting in this function. I guess it can be simplified to something like this (pseudocode):

```rust
// If compiler versions from the metadata don't match with the request,
// we'll first try to use info from metadata.
let alternative_request = detected_metadata
  .filter_map(|metadata| {
    if !request.compiler_versions_match(&metadata) {
      let updated_request = ...;
      Some(updated_request)
    }  else {
       None
    });
if let Some(alternative_request) = alternative_request {
    // compile
    // check against identifier
    // if matches, save it and return artifacts
}

// We either have no alternative request or it didn't succeed
self.compile(request.req.clone(), *bytecode_marker).await
```

core/lib/contract_verifier/src/lib.rs
```diff
@@ -353,6 +364,99 @@ impl ContractVerifier {
...
+                    .req
+                    .clone()
+                    .with_updated_compiler_versions(cbor_metadata);
+
+                match self
+                    .compile(updated_request.clone(), *bytecode_marker)
+                    .await
+                {
+                    Ok(artifacts) if artifacts.match_bytecode(bytecode_marker, deployed_code) => {
+                        tracing::warn!(
```
popzxc, line 392: I think `info` is sufficient here.

core/lib/types/src/contract_verification/contract_identifier.rs
```diff
@@ -58,28 +64,147 @@ impl DetectedMetadata {
...
+    ZKsync(String),
+}
+
+impl CborCompilerVersion {
+    /// Returns the compiler versions from the metadata in a tuple (compiler_version, zk_compiler_version).
+    pub fn get_compiler_versions(&self) -> (Option<String>, Option<String>) {
+        match self {
+            CborCompilerVersion::Native(compiler_version) => {
+                // For native Solc and Vyper compilers, CBOR is a 3 byte encoding of the compiler version: one byte each
+                // for major, minor and patch version number. For example, 0.8.28 is encoded as [0, 8, 28].
```
popzxc, line 101: It doesn't have to be 3-byte, [from the docs](https://docs.soliditylang.org/en/latest/metadata.html#encoding-of-the-metadata-hash-in-the-bytecode):

> pre-release builds will instead use a complete version string including commit hash and build date

core/lib/types/src/contract_verification/contract_identifier.rs
```diff
@@ -90,19 +215,23 @@ impl ContractIdentifier {
...
-        } else {
-            // Fallback
-            (None, bytecode_keccak256)
-        };
+        let (detected_metadata, bytecode_without_metadata_keccak256) =
+            if let Some((full_length, hash, cbor_metadata)) =
+                Self::detect_cbor_metadata(bytecode_marker, bytecode)
+            {
+                (
+                    Some(DetectedMetadata::Cbor {
```
popzxc, line 223: Nit: nesting makes it a bit too hard to read, I'd suggest assigning `DetectedMetadata` to a variable. Also, you can rename the variable to `metadata` for nicer constructing, e.g.

```rust
let detected_metadata = DetectedMetadata::Cbor {
  full_length,
  metadata
});
(detected_metadata, hash)
```

core/lib/types/src/contract_verification/contract_identifier.rs
vasyl-ivanchuk: yeah, but we [don't support](https://github.com/matter-labs/zksync-era/blob/main/core/lib/contract_verifier/src/resolver/github/gh_api.rs#L216) pre-releases, do we?
```
            if release.prerelease {
                continue;
            }
```

core/lib/contract_verifier/src/lib.rs
vasyl-ivanchuk: refactored `get_compilation_artifacts` fn, please check

core/lib/types/src/contract_verification/contract_identifier.rs
vasyl-ivanchuk: I compiled a contact with a prerelease version and tested, here is what happens in such case:
1. The CBOR data is properly deserialized.
2. When `get_compiler_versions` is called on `CborCompilerVersion`, it returns None for a prerelease version. In such cases, the implementation will fall back to the version from the verification request when compiling the contract.

I added unit tests to cover this use case and added comments to the `get_compiler_versions` fn.

core/lib/types/src/contract_verification/contract_identifier.rs
```diff
@@ -58,28 +64,150 @@ impl DetectedMetadata {
     pub fn length(self) -> usize {
```
popzxc, line 64: Nit: since the type is no longer `Copy`, this method should take `&self` instead of `self`.

popzxc (dismissed):

core/lib/types/src/contract_verification/contract_identifier.rs
vasyl-ivanchuk: Fixed

popzxc (approved):

PR #4008, opened 2025-05-13
feat(prover): Autoscaler detect which pod out of resources and reschedule

yorik:
## What ❔

Autoscaler detect which pod out of resources and reschedule.
For one run mark a pool as full (`max_pool_size` = number of Running + Pending pods) if there is out of resources pod in it.
Add `priority` config override.
Remove unneeded need_to_move_duration config parameter.

## Why ❔

Before all pending pod could be pushed to different cluster if even one of them is out of resources.
Add possibility to quickly switch to backup GPU if the main one is out.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist

ref ZKD-2682

### DIFF
--- a/prover/crates/bin/prover_autoscaler/src/global/scaler.rs
+++ b/prover/crates/bin/prover_autoscaler/src/global/scaler.rs
@@ -6,8 +6,8 @@ use debug_map_sorted::SortedOutputExt;
 use crate::{
     agent::{ScaleDeploymentRequest, ScaleRequest},
     cluster_types::{Cluster, ClusterName, Clusters, DeploymentName, NamespaceName, PodStatus},
-    config::QueueReportFields,
-    key::{Gpu, Key},
+    config::{PriorityConfig, QueueReportFields},
+    key::{Gpu, GpuKey, Key},
     metrics::{JobLabels, AUTOSCALER_METRICS},
 };
 
@@ -48,7 +48,6 @@ pub struct ScalerConfig {
     pub apply_min_to_namespace: Option<NamespaceName>,
     pub long_pending_duration: chrono::Duration,
     pub scale_errors_duration: chrono::Duration,
-    pub need_to_move_duration: chrono::Duration,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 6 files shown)


prover/crates/bin/prover_autoscaler/src/global/scaler.rs
```diff
@@ -182,9 +213,29 @@ impl<K: Key> Scaler<K> {
...
-                .then(b.max_pool_size.cmp(&a.max_pool_size)) // Reverse sort by cluster size.
+                        )
+                }
+            }
+        };
+
+        pools.sort_by(|a, b| {
+            if self.target_priority.is_some() {
+                // Use target_priority for sorting.
+                std::cmp::Ordering::Equal
```
Artemka374, line 224: maybe missing something obvious, but why is this equal here?

Artemka374: I mean why we are not sorting here by something

prover/crates/bin/prover_autoscaler/src/global/scaler.rs
```diff
@@ -160,19 +155,55 @@ impl<K: Key> Scaler<K> {
...
+                tracing::debug!(
+                    "Pool {}:{:?} has NeedToMove pods, max_pool_size adjusted to {}",
+                    pool.name,
+                    pool.key,
+                    pool.max_pool_size
+                );
+            }
+        }
+
+        let priority_cmp = |a: &Pool<K>, b: &Pool<K>| {
```
Artemka374, line 172: why don't we implement Ord explicitly?

prover/crates/bin/prover_autoscaler/src/global/scaler.rs
yorik: Key is GPU or nothing, if there is `target_priority`, it should be used for GPU sorting too, so disabling sorting by GPU.

prover/crates/bin/prover_autoscaler/src/global/scaler.rs
yorik: `priority_cmp` is just a part of ordering. I'll try to implement complete Ord.

yorik: Actually it becomes way too complicated: to sort pools it needs `cluster_priority` and `target_priority`.

Artemka374 (approved):

PR #4012, opened 2025-05-14
fix(zkstack): fixes for private rpc in zkstack cli

tomg10:


zkstack_cli/crates/zkstack/src/commands/private_rpc/mod.rs
```diff
@@ -28,10 +26,15 @@ use crate::{
     },
 };
 
+#[derive(Debug, Clone, Serialize, Deserialize, Parser, Default)]
+pub struct PrivateRpcCommandInitArgs {
+    #[clap(long)]
+    pub use_default: bool,
+}
```
Deniallugo, line 33: nit: new line

zkstack_cli/crates/zkstack/src/commands/private_rpc/mod.rs
```diff
@@ -28,10 +26,15 @@ use crate::{
     },
 };
 
+#[derive(Debug, Clone, Serialize, Deserialize, Parser, Default)]
+pub struct PrivateRpcCommandInitArgs {
+    #[clap(long)]
+    pub use_default: bool,
```
Deniallugo, line 32: let's use `dev` as in other commands

zkstack_cli/crates/zkstack/src/commands/private_rpc/mod.rs
```diff
@@ -127,15 +131,19 @@ pub async fn init(shell: &Shell) -> anyhow::Result<()> {
     logger::info(msg_private_rpc_docker_image_being_built());
     Cmd::new(cmd!(shell, "docker build -t private-rpc .")).run()?;
 
-    let mut db_config = prompt_db_config(&chain_config)?;
+    let db_config = if args.use_default {
+        let private_rpc_db_name: String = generate_private_rpc_db_name(&chain_config);
+        let private_rpc_db_name = slugify!(&private_rpc_db_name, separator = "_");
```
Deniallugo, line 136: why is it 2 functions? i'd assume it has to be the only. is there are any reason not to slugify it?

tomg10: actually, we may not need it at all, it should be already in correct format at this point, I removed the slugify!

zkstack_cli/crates/zkstack/src/commands/private_rpc/mod.rs
tomg10: corrected

zkstack_cli/crates/zkstack/src/commands/private_rpc/mod.rs
Deniallugo: no you don't ;)

Deniallugo (approved):

PR #4041, opened 2025-05-19
fix(gateway_migrator): add gateway migrator config

Deniallugo:
## What ❔

Introduce gateway migrator config 

## Why ❔

For consuming less eth calls

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/etc/env/file_based/general.yaml
+++ b/etc/env/file_based/general.yaml
@@ -280,4 +280,4 @@ consensus:
   gossip_dynamic_inbound_limit: 100
 
 gateway_migrator:
-  eth_node_poll_interval_ms: 300
+  eth_node_poll_interval: 300 ms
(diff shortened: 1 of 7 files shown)


etc/env/file_based/general.yaml
```diff
@@ -278,3 +278,6 @@ consensus:
   public_addr: "127.0.0.1:3054"
   max_payload_size: 2500000
   gossip_dynamic_inbound_limit: 100
+
+gateway_migrator:
+  eth_node_poll_interval: 300
```
RomanBrodetski, line 283: can we make the name so that duration unit is clear? Also, why does it differ from the default config?

Deniallugo: I want to use default config in production. And this config for testing

RomanBrodetski (approved):

PR #4048, opened 2025-05-20
feat: private-rpc blacklist of rpc methods

calvogenerico:
Era rpc exposes a lot of methods that require careful handling in the prividium private rpc. At this stage we are just filtering them out.

## What ❔

Filtering out all methods that might cause to expose more state than desired in the prividium private rpc.

## Why ❔

In the context of prividium each piece of state requires permisions to be read. These rpc methods expose state that is not correctly protected, yet. According to the team discussions we are blacklisting these methods for now.

## Checklist


private-rpc/src/rpc/rpc-method-handlers.ts
```diff
@@ -14,19 +14,57 @@ import {
...
+    forbiddenMethod('en_consensusGlobalConfig'),
+    forbiddenMethod('en_blockMetadata'),
+    forbiddenMethod('en_syncTokens'),
+    forbiddenMethod('en_genesisConfig'),
+    forbiddenMethod('en_whitelistedTokensForAA'),
+    forbiddenMethod('en_getEcosystemContracts'),
+    forbiddenMethod('en_getProtocolVersionInfo'),
+
+    // Filter out snapshot_* methods
+    forbiddenMethod('snapshot_getAllSnapshots'),
```
tomg10, line 34: typo: it's "snapshots", not "snapshot"

tomg10 (changes requested):
other than a typo looks good

private-rpc/src/rpc/rpc-method-handlers.ts
calvogenerico: Fixed!

tomg10 (approved):

yuliyaalexiev (approved):

tomg10 (approved):

PR #4061, opened 2025-05-23
fix(contract-verifier): improve etherscan error handling

vasyl-ivanchuk:
## What ❔

**This PR improves Etherscan verificaation processing and error handling.**

**Contract bytecode is not available error retry policy:**
It turned out that the retry policy for the `Contract bytecode is not available` error doesn't work well, as it sometimes takes more than 10 minutes for Etherscan to index the contract data. This exceeds the total retry timeout for the error.

To address this, I decided to introduce an increasing delay for this specific error - starting at 2 minutes and going up to 1 day for the final attempt.

IMO this approach is gonna work better: it begins with a quick retry to handle the positive scenario, while still giving Etherscan enough time to index the contract by using progressively longer timeouts for later attempts.

Existing verifications that failed due to this error are requeued for processing.

**Requests processing and error handling:**
Previously, in the case of any error, the verifier waited for a timeout and retried the same request without attempting to process other requests. This could be problematic, especially with the progressively increasing timeouts introduced for `Contract bytecode is not available` errors.

With this change, when a retryable error occurs, the verifier updates the `retry_at` timestamp for the request and skips processing it until the timeout has passed. In the meantime, it continues processing other requests that are ready.

For indefinitely retryable errors, the verifier still pauses and does not process other requests. These types of errors are typically not specific to a particular request - they may indicate a service outage, hitting a rate limit, a misconfiguration, etc. To avoid spamming Etherscan, the verifier pauses and retries without moving on to other queued requests.

**Metrics**
We lack visibility into when the Etherscan verification queue is growing. I'm adding a metric for the number of queued requests so we can set up proper alerts based on it.

## Why ❔

- To improve request processing and properly handle the `Contract bytecode is not available` error.
- To improve service monitoring.

## Is this a breaking change?
- [ ] Yes
- [X] No

## Checklist


core/lib/contract_verifier/src/etherscan/metrics.rs
```diff
@@ -0,0 +1,59 @@
+//! Reports etherscan verifier metrics
+use std::time::Duration;
+
+use tokio::sync::watch;
+use zksync_dal::{ConnectionPool, Core, CoreDal};
+
+use crate::metrics::CONTRACT_VERIFIER_METRICS;
+
+pub struct EtherscanVerifierMetrics {
```
popzxc, line 9: Nit: Please always derive `Debug` for your structures, as it can simplify debugging in the future (most common case: when your structure is included in some tester object, and lack of `Debug` derive forbids deriving `Debug` there).

core/lib/contract_verifier/src/etherscan/metrics.rs
```diff
@@ -0,0 +1,59 @@
...
+        let number_of_queued_requests = connection
+            .etherscan_verification_dal()
+            .get_number_of_queued_requests()
+            .await?;
+        CONTRACT_VERIFIER_METRICS.number_of_queued_requests[&"etherscan_verifier"]
+            .set(number_of_queued_requests);
+        Ok(())
+    }
+
+    pub async fn report(mut self) -> anyhow::Result<()> {
```
popzxc, line 43: Nit: I think `run` or `run_reporter` name makes more sense (`report` feels like a single iteration)

popzxc (approved):

core/lib/contract_verifier/src/etherscan/metrics.rs
vasyl-ivanchuk: Agreed, renamed to `run`.

popzxc (approved):

PR #4069, opened 2025-05-23
feat: add some traffic to hit edge case during validium migration

ischasny:
This PR adds submits some transactions before rollup -> validium migration to hit an edgecase when a batch was created as a rollup but committed as a validium


deniallugo-ml:

PR #4073, opened 2025-05-26
fix(eth_aggregator): choose Blobs or Calldata based on the previous batch

dimazhornyk:
## What ❔

Detect if blobs or calldata were used before the migration instead of using blobs by default.

## Why ❔

If the chain is settling on Gateway and there are uncommitted batches - commit will fail if `Blobs` pubdata type is used by default.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


.github/workflows/ci-core-reusable.yml
```diff
@@ -558,13 +558,19 @@ jobs:
 
       - name: DA migration tests
         run: |
-          echo "Reducing L1 batch seal timeout and restarting server"
+          echo "Increasing L1 batch seal timeout and restarting server"
           sudo yq -i '.state_keeper.block_commit_deadline_ms = "10000"' ./chains/da_migration/configs/general.yaml
```
cytadela8, line 562: Are you sure about this 10sec wait? Below, we wait for 60sec after restarting. Maybe at this point we can set something like 30sec and be "safer" regarding existence of an unsealed batch on server start.

Just wanted to flag this, but I don't necessarily understand fully what are the possible interlacing of events here.

cytadela8 (approved):

.github/workflows/ci-core-reusable.yml
dimazhornyk: we actually want an unsealed batch to be there at the moment of restart
on the real environment there will almost certainly be an unsealed batch during it, so we want to be as close to that as possible

alexandrst88 (approved):

hatemosphere (approved):

PR #4076, opened 2025-05-26
feat: improve validation on eth_call for private-rpc

aon:
## What ❔
Improves validation on eth_call for private-rpc, preventing users from passing additional unwanted params.


mikhail-egorov:
@aon, could you take a look at the `onlyUserParamsSchema`? It currently allows overly broad parameters after the first address. https://github.com/matter-labs/zksync-era/blob/main/private-rpc/src/rpc/methods/utils.ts#L47

aon:
@mikhail-egorov thanks for the comment. Fixed in dcb39fc99b3f38db94fc1139099bbe2ba34fc4ab.

cytadela8 (approved):

PR #4082, opened 2025-05-27
feat(en): make `max_batches_to_recheck` configurable

dimazhornyk:
## What ❔

Make the number of batches to fetch by DA fetcher and to check by ConsistencyChecker configurable.
The number can be the same for these two components because we only use the DA information to create a correct commitment in the consistency checker.

## Why ❔

The initial motivation was to avoid scenarios where ConsistencyChecker can't progress for a very long because DA fetcher is fetching all the info from batch 0. But I also noticed that we wanted to make it configurable for ConsistencyChecker as well.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


core/bin/external_node/src/config/mod.rs
```diff
@@ -517,6 +517,9 @@ pub(crate) struct OptionalENConfig {
     /// Minimum time between current block.timestamp and the end of the asserted range for TimestampAsserter
     #[serde(default = "OptionalENConfig::default_timestamp_asserter_min_time_till_end_sec")]
     pub timestamp_asserter_min_time_till_end_sec: u32,
+    /// Maximum number of batches to recheck.
+    #[serde(default = "OptionalENConfig::default_max_batches_to_recheck")]
+    pub max_batches_to_recheck: u32,
```
slowli: JFYI, https://github.com/matter-labs/zksync-era/pull/4104 removes `OptionENConfig` altogether 🙃 I'm completely fine with merging this PR first, but to not immediately require a deprecated alias (look at lots of those [in the config schema definition](https://github.com/matter-labs/zksync-era/blob/a7586c2bf40ba2977593ad4e619545fe6a9ad8f5/core/bin/external_node/src/config/mod.rs#L232)), I'd suggest to define a consistency checker config in the `zksync_config` crate, and have it "mount" (virtually so far) at the `consistency_checker` prefix – i.e. use the env var name like `EN_CONSISTENCY_CHECKER_MAX_BATCHES_TO_RECHECK`.

core/node/node_sync/src/data_availability_fetcher/mod.rs
```diff
@@ -91,9 +93,34 @@ impl DataAvailabilityFetcher {
...
+            .await?
+            .blocks_dal()
+            .get_number_of_last_l1_batch_committed_on_eth()
+            .await?
+            .unwrap_or(L1BatchNumber(0));
+
+        let first_batch_to_check: L1BatchNumber = last_committed_batch
+            .0
+            .saturating_sub(self.max_batches_to_recheck + 1) // +1 to cover any race conditions with consistency checker
+            .into();
```
slowli: I can't say I understand this logic fully, but it looks fragile. Since there's a persistent cursor for the last batch checked by the consistency checker (available via `blocks_dal().get_consistency_checker_last_processed_l1_batch()` from DAL), can't it be used for the same effect here?

slowli (commented):
Not sure whether this should be solved in the scope of this PR, but the consistency checker logic has a flaw. During a revert, the persistent consistency checker cursor is not updated (it should), so batches re-added after a revert are not rechecked.

Another feature I see making sense is allowing to redefine the first batch to check manually, which would override the persistent consistency checker cursor. No clear demand from node operators, so it could wait.

core/node/node_sync/src/data_availability_fetcher/mod.rs
dimazhornyk: yeah, great suggestion
rewrote it to use `get_consistency_checker_last_processed_l1_batch`

core/bin/external_node/src/config/mod.rs
dimazhornyk: added it as a separate config, let me know if you envision it differently

dimazhornyk:
> Not sure whether this should be solved in the scope of this PR, but the consistency checker logic has a flaw. During a revert, the persistent consistency checker cursor is not updated (it should), so batches re-added after a revert are not rechecked.
> 
> Another feature I see making sense is allowing to redefine the first batch to check manually, which would override the persistent consistency checker cursor. No clear demand from node operators, so it could wait.

I added the revert part, but didn't do the second point as this might require a db migration (to avoid consistently overwriting it at restart), so I decided to keep it simple for now

core/lib/config/src/configs/consistency_checker.rs
```diff
@@ -0,0 +1,10 @@
+use serde::Deserialize;
+use smart_config::{DescribeConfig, DeserializeConfig};
+
+#[derive(Debug, Clone, PartialEq, Deserialize, DescribeConfig, DeserializeConfig)]
+#[config(derive(Default))]
+pub struct ConsistencyCheckerConfig {
+    /// Maximum number of batches to recheck.
+    #[config(default_t = 10)]
```
slowli, line 8: You will need to specify `#[serde(default = _)]` as well. (Obviously, wouldn't be necessary if https://github.com/matter-labs/zksync-era/pull/4104 was merged first.) Without it, the param would be required when using env-based config for EN, which is not what we want.

core/lib/config/src/configs/general.rs
```diff
@@ -98,6 +98,8 @@ pub struct GeneralConfig {
     pub timestamp_asserter_config: TimestampAsserterConfig,
     #[config(nest, rename = "gateway_migrator")]
     pub gateway_migrator_config: GatewayMigratorConfig,
+    #[config(nest, rename = "consistency_checker")]
+    pub consistency_checker_config: ConsistencyCheckerConfig,
```
slowli, line 102: Meh; `GeneralConfig` is only used by the main node, and the consistency checker isn't applicable for it. Since it'll get removed in https://github.com/matter-labs/zksync-era/pull/4104 anyway, probably OK to have it temporarily.

core/node/node_sync/src/data_availability_fetcher/mod.rs
```diff
@@ -94,6 +94,26 @@ impl DataAvailabilityFetcher {
         }
     }
 
+    async fn determine_first_batch_to_scan(&self) -> anyhow::Result<L1BatchNumber> {
+        let last_checked_by_consistency_checker = self
```
slowli, line 103: Nit: Could you please add a comment explaining why we use the consistency checker cursor as the baseline here?

core/lib/config/src/configs/consistency_checker.rs
dimazhornyk: done, thanks for pointing out! at some point I will get used to the new configs 😆

core/node/node_sync/src/data_availability_fetcher/mod.rs
dimazhornyk: done

core/lib/config/src/configs/general.rs
dimazhornyk: yeah, I don't like it here as well
but adding it to EN config was also looking weird
if you don't mind - I'd better keep it as is and change it properly in your #4104 PR

slowli (dismissed):

slowli (approved):

PR #4087, opened 2025-05-28
feat(state-keeper): add `process_block` method

perekopskiy:
## What ❔

Reworks state keeper so it's possible to process single L2 block with `process_block` method.

## Why ❔

Prerequisite for leader rotation.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Operational changes

None

## Checklist

### DIFF
--- a/core/lib/vm_interface/src/executor.rs
+++ b/core/lib/vm_interface/src/executor.rs
@@ -44,6 +44,9 @@ pub trait BatchExecutor<S>: 'static + Send + fmt::Debug {
 
     /// Finished the current L1 batch.
     async fn finish_batch(self: Box<Self>) -> anyhow::Result<(FinishedL1Batch, StorageView<S>)>;
+
+    /// Return gas remaining in VM instance.
+    async fn gas_remaining(&mut self) -> anyhow::Result<u32>;
 }
 
 /// VM executor capable of executing isolated transactions / calls (as opposed to [batch execution](BatchExecutor)).
--- a/core/node/consensus/src/testonly.rs
+++ b/core/node/consensus/src/testonly.rs
@@ -33,8 +33,8 @@ use zksync_state_keeper::{
     io::{IoCursor, L1BatchParams, L2BlockParams},
     seal_criteria::NoopSealer,
     testonly::{fee, fund, test_batch_executor::MockReadStorageFactory, MockBatchExecutor},
-    AsyncRocksdbCache, OutputHandler, StateKeeperPersistence, TreeWritesPersistence,
-    ZkSyncStateKeeper,
+    AsyncRocksdbCache, OutputHandler, StateKeeperInner, StateKeeperPersistence,
+    TreeWritesPersistence,
 };
 use zksync_test_contracts::Account;
 use zksync_types::{
@@ -538,6 +538,20 @@ impl StateKeeperRunner {
                 self.rocksdb_dir.path().join("cache"),
                 Default::default(),
             );
+            let executor_factory = MainBatchExecutorFactory::<()>::new(false);
+            let state_keeper = StateKeeperInner::new(
+                Box::new(io),
+                Box::new(executor_factory),
+                OutputHandler::new(Box::new(persistence.with_tx_insertion()))
+                    .with_handler(Box::new(self.sync_state.clone())),
+                Arc::new(NoopSealer),
+                Arc::new(async_cache),
+                None,
+            )
+            .initialize(&stop_recv)
+            .await
+            .unwrap();
+
             s.spawn_bg({
                 let stop_recv = stop_recv.clone();
                 async {
@@ -555,21 +569,12 @@ impl StateKeeperRunner {
             });
 
             s.spawn_bg({
-                let executor_factory = MainBatchExecutorFactory::<()>::new(false);
                 let stop_recv = stop_recv.clone();
                 async {
-                    ZkSyncStateKeeper::new(
-                        Box::new(io),
-                        Box::new(executor_factory),
-                        OutputHandler::new(Box::new(persistence.with_tx_insertion()))
-                            .with_handler(Box::new(self.sync_state.clone())),
-                        Arc::new(NoopSealer),
-                        Arc::new(async_cache),
-                        None,
-                    )
-                    .run(stop_recv)
-                    .await
-                    .context("ZkSyncStateKeeper::run()")?;
+                    state_keeper
+                        .run(stop_recv)
+                        .await
+                        .context("StateKeeper::run()")?;
                     Ok(())
                 }
             });
@@ -626,6 +631,21 @@ impl StateKeeperRunner {
                 Box::<MockMainNodeClient>::default(),
                 L2ChainId::default(),
             )?;
+
+            let state_keeper = StateKeeperInner::new(
+                Box::new(io),
+                Box::new(MockBatchExecutor),
+                OutputHandler::new(Box::new(persistence.with_tx_insertion()))
+                    .with_handler(Box::new(tree_writes_persistence))
+                    .with_handler(Box::new(self.sync_state.clone())),
+                Arc::new(NoopSealer),
+                Arc::new(MockReadStorageFactory),
+                None,
+            )
+            .initialize(&stop_recv)
+            .await
+            .unwrap();
+
             s.spawn_bg(async {
                 Ok(l2_block_sealer
                     .run()
@@ -644,19 +664,10 @@ impl StateKeeperRunner {
             s.spawn_bg({
                 let stop_recv = stop_recv.clone();
                 async {
-                    ZkSyncStateKeeper::new(
-                        Box::new(io),
-                        Box::new(MockBatchExecutor),
-                        OutputHandler::new(Box::new(persistence.with_tx_insertion()))
-                            .with_handler(Box::new(tree_writes_persistence))
-                            .with_handler(Box::new(self.sync_state.clone())),
-                        Arc::new(NoopSealer),
-                        Arc::new(MockReadStorageFactory),
-                        None,
-                    )
-                    .run(stop_recv)
-                    .await
-                    .context("ZkSyncStateKeeper::run()")?;
+                    state_keeper
+                        .run(stop_recv)
+                        .await
+                        .context("StateKeeper::run()")?;
                     Ok(())
                 }
             });
--- a/core/node/state_keeper/src/keeper.rs
+++ b/core/node/state_keeper/src/keeper.rs
@@ -38,6 +38,42 @@ use crate::{
 /// we only need it to not block on waiting indefinitely and be able to process cancellation requests.
 pub(super) const POLL_WAIT_DURATION: Duration = Duration::from_secs(1);
 
+#[derive(Debug)]
+struct L1BatchData {
+    updates_manager: UpdatesManager,
+    batch_executor: Box<dyn BatchExecutor<OwnedStorage>>,
+    protocol_upgrade_tx: Option<ProtocolUpgradeTx>,
+}
+
+#[derive(Debug)]
+enum BatchState {
+    Uninit(IoCursor),
+    Init(Box<L1BatchData>),
+}
+
+impl BatchState {
+    pub fn as_uninit(&self) -> Option<IoCursor> {
+        match self {
+            BatchState::Uninit(cursor) => Some(*cursor),
+            BatchState::Init(_) => None,
+        }
+    }
+
+    pub fn unwrap_init_ref_mut(&mut self) -> &mut L1BatchData {
+        match self {
+            BatchState::Uninit(_) => panic!("Unexpected `BatchState::Uninit`"),
+            BatchState::Init(data) => data.as_mut(),
+        }
+    }
+
+    pub fn unwrap_init(self) -> Box<L1BatchData> {
+        match self {
+            BatchState::Uninit(_) => panic!("Unexpected `BatchState::Uninit`"),
+            BatchState::Init(data) => data,
+        }
+    }
+}
+
 /// State keeper represents a logic layer of L1 batch / L2 block processing flow.
 ///
 /// It's responsible for taking all the data from the `StateKeeperIO`, feeding it into `BatchExecutor` objects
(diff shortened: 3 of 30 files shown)


core/node/consensus/src/testonly.rs
```diff
@@ -538,6 +538,20 @@ impl StateKeeperRunner {
...
+                Box::new(executor_factory),
+                OutputHandler::new(Box::new(persistence.with_tx_insertion()))
+                    .with_handler(Box::new(self.sync_state.clone())),
+                Arc::new(NoopSealer),
+                Arc::new(async_cache),
+                None,
+            )
+            .initialize(&stop_recv)
+            .await
+            .unwrap();
```
brunoffranca: Just out of curiosity, did you have to move this outside the spawn block or was is it just cosmetic?

core/node/state_keeper/src/keeper.rs
```diff
@@ -49,28 +85,37 @@ pub(super) const POLL_WAIT_DURATION: Duration = Duration::from_secs(1);
 /// You can think of it as a state machine that runs over a sequence of incoming transactions, turning them into
 /// a sequence of executed L2 blocks and batches.
 #[derive(Debug)]
-pub struct ZkSyncStateKeeper {
+pub struct StateKeeper {
+    inner: StateKeeperInner,
+    batch_state: BatchState,
+    last_l1_batch_sealed_at: Option<Instant>,
```
brunoffranca: What's the purpose of the `last_l1_batch_sealed_at` field?

core/node/consensus/src/testonly.rs
perekopskiy: It's cosmetic. I wanted to emphasize that initialize happens at the start and is not included in the main consensus loop

core/node/state_keeper/src/keeper.rs
perekopskiy: it's needed to record `seal_delta` metric, that's it

core/node/state_keeper/src/keeper.rs
```diff
@@ -38,6 +38,42 @@ use crate::{
...
+}
+
+#[derive(Debug)]
+enum BatchState {
+    Uninit(IoCursor),
+    Init(Box<L1BatchData>),
+}
+
+impl BatchState {
+    pub fn as_uninit(&self) -> Option<IoCursor> {
```
slowli: - Nit: `pub` visibility doesn't make sense since the type itself is private. 
- I'd made these method more high-level, encapsulating entire "uninit -> init" and "init -> uninit" transitions inside `BatchState` methods. IIUC, it's possible and would allow to get rid of lower-level methods altogether.

core/node/state_keeper/src/keeper.rs
```diff
@@ -38,6 +38,42 @@ use crate::{
 /// we only need it to not block on waiting indefinitely and be able to process cancellation requests.
 pub(super) const POLL_WAIT_DURATION: Duration = Duration::from_secs(1);
 
+#[derive(Debug)]
+struct L1BatchData {
```
slowli: Bikeshedding: The struct fields are not entirely data though; both `updates_manager` and especially `batch_executor` have processing logic. Maybe, something like `InitializedBatchState` would work better?

core/node/state_keeper/src/keeper.rs
```diff
@@ -938,4 +794,314 @@ impl ZkSyncStateKeeper {
...
+                .inner
+                .process_one_tx(batch_executor, updates_manager, tx.clone())
+                .await?;
+
+            #[derive(Debug)]
+            enum Outcome {
+                SealBatch,
+                SealBlock,
+                None,
+            }
```
slowli: Bikeshedding: Defining types in the middle of logic looks dirty.

core/node/state_keeper/src/keeper.rs
```diff
@@ -938,4 +794,314 @@ impl ZkSyncStateKeeper {
...
+                        .with_context(|| format!("cannot reject transaction {tx_hash:?}"))?;
+                    Outcome::None
+                }
+            };
+            latency.observe();
+
+            if matches!(outcome, Outcome::SealBatch | Outcome::SealBlock) {
+                match outcome {
+                    Outcome::SealBatch => self.seal_batch().await?,
+                    Outcome::SealBlock => self.inner.seal_l2_block(updates_manager).await?,
```
slowli: Is it possible to execute `seal_l2_block()` / `seal_batch()` after returning from this method (IDK, maybe by returning `Outcome`)? RN, it isn't *entirely* clear from the method logic that the method necessarily seals a single L2 block / L1 batch and then exits.

core/node/state_keeper/src/keeper.rs
```diff
@@ -938,4 +794,314 @@ impl ZkSyncStateKeeper {
...
+                    _ => unreachable!(),
```
slowli: Nit: Wildcard matching for a single variant is considered an antipattern in the general case because other variants may be added to the enum in the future; not a huge concern here.

core/node/state_keeper/src/keeper.rs
```diff
@@ -938,4 +794,314 @@ impl ZkSyncStateKeeper {
...
+                        .reject(&tx, reason.clone())
+                        .await
+                        .with_context(|| format!("cannot reject transaction {tx_hash:?}"))?;
+                    Outcome::None
+                }
+            };
+            latency.observe();
+
+            if matches!(outcome, Outcome::SealBatch | Outcome::SealBlock) {
+                match outcome {
```
slowli: Bikeshedding: a `matches!` + `match` on the same var looks weird / redundant; it'd be possible to just use `match` and do nothing if `outcome` is `None`. Perhaps, use a flag whether to exit from the method and check it after observing `full_latency`.

core/lib/vm_interface/src/executor.rs
```diff
@@ -45,6 +45,9 @@ pub trait BatchExecutor<S>: 'static + Send + fmt::Debug {
 
     /// Finished the current L1 batch.
     async fn finish_batch(self: Box<Self>) -> anyhow::Result<(FinishedL1Batch, StorageView<S>)>;
+
+    /// Return gas remaining in VM instance.
+    async fn gas_remaining(&mut self) -> anyhow::Result<u32>;
```
slowli: Dumb question: There's `gas_remaining` available from `VmExecutionStatistics` via `VmExecutionResultAndLogs` and `BatchTransactionExecutionResult` returned after executing each transaction. Can't it be used instead of introducing a new method?

core/node/state_keeper/src/keeper.rs
slowli: Nit: Can't this field be moved somewhere to the output handler though? IMO, it'd belong better there than introducing a field in the logic-heavy type just for a metric.

core/node/state_keeper/src/keeper.rs
```diff
@@ -938,4 +794,314 @@ impl ZkSyncStateKeeper {
...
+                }
+
+                self.seal_batch().await?;
+                return Ok(());
+            }
+        }
+
+        while !is_canceled(stop_receiver) {
+            let full_latency = KEEPER_METRICS.process_block_loop_iteration.start();
+            if self
```
slowli: Bikeshedding: IMO, it'd make sense to extract iteration to a separate method. It'd allow to handle `full_latency` better (observe it in a single place), among other things.

core/node/state_keeper/src/keeper.rs
```diff
@@ -49,28 +85,37 @@ pub(super) const POLL_WAIT_DURATION: Duration = Duration::from_secs(1);
...
+pub struct StateKeeper {
+    inner: StateKeeperInner,
+    batch_state: BatchState,
+    last_l1_batch_sealed_at: Option<Instant>,
+}
+
+/// Helper struct that is used for state keeper initialization.
+/// Also encapsulates some private state keeper methods.
+#[derive(Debug)]
+pub struct StateKeeperInner {
```
slowli: Bikeshedding: `_Inner` suffix is commonly used for private structures, so using it publicly (e.g., in DI) looks weird. If the only purpose of this struct is to initialize `StateKeeper`, I'd call it `StateKeeperBuilder` + rename `initialize` to `build()` (i.e., lean into the builder pattern).

slowli (commented):
Mostly bikeshedding, but I hope there are useful suggestions as well 🙃

core/lib/vm_interface/src/executor.rs
perekopskiy: Yes, changed to use it

core/node/state_keeper/src/keeper.rs
perekopskiy: It's not only initialization, but also some inner methods. Code looks much better when having methods for the struct that doesn't own updates_manager/batch_executor. If there is no Inner struct and we move methods to StateKeeper then we can't hold mutable reference to `updates_manager` and call some mutable method which is very inconvenient and bloats the code. 
I've split StateKeeperInner and StateKeeperBuilder, only StateKeeperBuilder is used publicly

core/node/state_keeper/src/keeper.rs
```diff
@@ -38,6 +34,68 @@ use crate::{
...
+impl BatchState {
+    /// Initializes batch if it was not.
+    /// Returns mutable reference of the initialized state.
+    async fn ensure_initialized(
+        &mut self,
+        inner: &mut StateKeeperInner,
+        stop_receiver: &mut watch::Receiver<bool>,
+    ) -> Result<&mut InitializedBatchState, OrStopped> {
+        let state = match self {
+            BatchState::Uninit(cursor) => inner.start_batch(*cursor, stop_receiver).await?,
```
slowli: Nit: It would be more idiomatic to use `Self::` instead of `BatchState::` here and below.

core/node/state_keeper/src/keeper.rs
```diff
@@ -38,6 +34,68 @@ use crate::{
...
+    }
+
+    fn unwrap_init_mut(&mut self) -> &mut InitializedBatchState {
+        match self {
+            BatchState::Uninit(_) => panic!("Unexpected `BatchState::Uninit`"),
+            BatchState::Init(init) => init.as_mut(),
+        }
+    }
+
+    fn unwrap_init_ref(&self) -> &InitializedBatchState {
```
slowli, line 85: Nit: I think it should be possible to get rid of this and `unwrap_init_mut()` by refactoring `process_block` and methods `process_block_iteration` so that they have `StateKeeperInner` as the receiver and `&mut InitializedBatchState` as an additional argument. Not high-priority obviously; the current implementation is OK-ish.

core/node/state_keeper/src/keeper.rs
```diff
@@ -938,4 +843,315 @@ impl ZkSyncStateKeeper {
...
+            }
+
+            return if updates_manager.l2_block.executed_transactions.is_empty() {
+                Ok(Some(ProcessBlockOutcome::SealBatch))
+            } else {
+                tracing::debug!(
+                    "L2 block #{} should be sealed as per L1 batch unconditional sealing rules",
+                    updates_manager.l2_block.number,
+                );
+                Ok(Some(ProcessBlockOutcome::SealBlock))
```
slowli: Dumb question: How does this branch work? Does execute another `process_block` iteration and expects that it will necessarily have `should_seal_l1_batch_unconditionally` return `true` immediately? If so, maybe it makes sense to add a flag to `ProcessBlockOutcome::SealBatch` (e.g., `seal_block: bool`) so that it's not necessary to rely on such implicit logic?

core/node/state_keeper/src/keeper.rs
```diff
@@ -938,4 +843,315 @@ impl ZkSyncStateKeeper {
...
+            .report_seal_criteria_capacity(&state.updates_manager);
+
+        let (finished_batch, _) = state.batch_executor.finish_batch().await?;
+        state.updates_manager.finish_batch(finished_batch);
+        let l1_batch_number = state.updates_manager.l1_batch.number;
+        self.inner
+            .output_handler
+            .handle_l1_batch(Arc::new(state.updates_manager))
+            .await
+            .with_context(|| format!("failed sealing L1 batch #{}", l1_batch_number))?;
```
slowli: Micro-nit: `l1_batch_number` can be inlined into the interpolated string.

core/node/state_keeper/src/keeper.rs
```diff
@@ -938,4 +843,315 @@ impl ZkSyncStateKeeper {
...
+                    );
+                };
+                updates_manager.extend_from_executed_transaction(
+                    tx,
+                    *tx_result,
+                    *tx_execution_metrics,
+                    call_tracer_result,
+                    gas_remaining,
+                );
+                Some(ProcessBlockOutcome::SealBlock)
```
slowli: This isn't obvious as an outcome of `IncludeAndSeal` either; after all, `SealResolution`s are about sealing *batches*. From this perspective, the only place that should return `ProcessBlockOutcome::SealBlock` is one that checks `should_seal_l2_block()`.

perekopskiy: Hmm, looks like it's an issue indeed. If e.g. `close_block_at_geometry_percentage = 0.95` then blocks will be sealed with only 1 tx while the capacity of the criterion is between 95% and 100%. If it wasn't about leader rotation then returning here e.g. `SealBatch { seal_block: true }` like you suggested would work without any issues.
With leader rotation my thoughts were that we allow different `close_block_at_*` params per leader node (But each node will only verify that actual max limits are not exceeded, i.e. if run sealer with close_block_at=reject_tx_at=1 then it doesn't return neither ExcludeAndSeal nor Unexecutable. So with leader rotation it's possible that leader A gets `SealBatch { seal_block: true }` but then it's a turn of another leader B and he seals another block and also gets `SealBatch { seal_block: true }`. If leader is changed every time it will be a problem. On practice I think we will not use configuration where leader is changed every block so the solution with `SealBatch { seal_block: true }` should mostly work.

So I see 2 options:
- in case state keeper wants to seal non-fictive block and then seal the batch right-away, then return special block outcome that will modify batch state so that on the next `process_block` invocation fictive block will be created.
- for all criteria replace code like
```
let include_and_seal_bound = (bootloader_tx_encoding_space as f64
            * config.close_block_at_geometry_percentage)
            .round();
```
with `let exclude_and_seal_bound = ...`. So when leader reaches `close_block_at_*` it will seal the batch by ExcludeAndSeal.

I would prefer first option as it's easier. Given that batches will be removed in ZKsync OS I think it's ok to have it as a temporary solution. WDYT?

perekopskiy: I've implemented the first option in [this](https://github.com/matter-labs/zksync-era/pull/4087/commits/fc2d6675b395e5a2efbbbd21c4bb12aaee35f43c) commit

core/node/state_keeper/src/keeper.rs
```diff
@@ -49,28 +101,60 @@ pub(super) const POLL_WAIT_DURATION: Duration = Duration::from_secs(1);
...
     sealer: Arc<dyn ConditionalSealer>,
     storage_factory: Arc<dyn ReadStorageFactory>,
     health_updater: HealthUpdater,
     deployment_tx_filter: Option<DeploymentTxFilter>,
 }
 
-impl ZkSyncStateKeeper {
+/// Helper struct that encapsulates some private state keeper methods.
+#[derive(Debug)]
+pub(super) struct StateKeeperInner {
```
slowli, line 122: This has the same fields as `StateKeeperBuilder`. Is this intentional? Would it make sense to encapsulate `_Inner` in `_Builder` (if the goal is to separate them by the usage pattern) or use a single type (i.e., separate by method visibility)?

core/node/state_keeper/src/keeper.rs
```diff
@@ -943,3 +843,265 @@ impl ZkSyncStateKeeper {
...
+                    );
+                };
+                updates_manager.extend_from_executed_transaction(
+                    tx,
+                    *tx_result,
+                    *tx_execution_metrics,
+                    call_tracer_result,
+                );
+            }
+            SealResolution::IncludeAndSeal => {
```
slowli: Dumb question: Why was this split off from the `SealResolution::NoSeal` match hand above? AFAICT, the code executed is still identical in both cases.

slowli (approved):
Looks good, just got a couple of questions / suggestions.

core/node/state_keeper/src/keeper.rs
perekopskiy: IMO it's cleaner to keep separate types. It adds a little code and usage pattern is straightforward this way. Also, in future it's possible they will have different set of fields

slowli (approved):

PR #4090, opened 2025-05-28
Precommitment support on top of v29 branch

StanislavBreadless:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


core/lib/contracts/src/lib.rs
```diff
@@ -1549,3 +1549,102 @@ pub static POST_SHARED_BRIDGE_EXECUTE_FUNCTION: Lazy<Function> = Lazy::new(|| {
...
+                            "internalType": "bytes",
+                            "name": "_executeData",
+                            "type": "bytes"
+                        }
+                    ],
+                    "stateMutability": "nonpayable",
+                    "type": "function",
+                    "name": "executeBatchesSharedBridge",
+                    "outputs": []
+                }"#;
```
perekopskiy: nit: please fix tabs for POST_V26_GATEWAY abis

core/lib/dal/src/blocks_dal.rs
```diff
@@ -549,14 +588,139 @@ impl BlocksDal<'_, '_> {
         Ok(storage_oracle_info.and_then(DbStorageOracleInfo::into_optional_batch_oracle_info))
     }
 
+    /// The server is doing precommits based on the txs.
+    /// That means, that the last fictive l2 block will never be included into precommit txs and it's the only way how the miniblock will have no txs.
+    /// So we have to check if the last 2 rolling txs hashes for miniblocks are equal and if yes, we can set the final precommit tx id for the batch,
+    /// and that will mean we can commit the batch.
+    pub async fn set_final_precommit_operation(&mut self) -> DalResult<()> {
```
perekopskiy: This is too complex for dal method. Let's move most of the logic to the caller side.

core/lib/dal/migrations/20250514091339_precommit_tx.up.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE miniblocks ADD COLUMN rolling_txs_hash BYTEA;
```
perekopskiy, line 1: Is it only needed to do `if miniblocks[0].rolling_txs_hash == miniblocks[1].rolling_txs_hash`? If so you can simply check that tx_count for miniblocks[0] is 0; and remove rolling_txs_hash column

core/lib/dal/src/blocks_dal.rs
```diff
@@ -549,14 +588,139 @@ impl BlocksDal<'_, '_> {
...
+                );
+                let result = instrumentation
+                    .clone()
+                    .with(query)
+                    .execute(self.storage)
+                    .await?;
+
+                if result.rows_affected() == 0 {
+                    let err = instrumentation.constraint_error(anyhow::anyhow!(
+                        "Update eth_tx_id that is is not null is not allowed"
```
perekopskiy: nit: eth_tx_id -> eth_precommit_tx_id
also I would check `if result.rows_affected() != number_range.len()` instead of `if result.rows_affected() == 0`

core/lib/dal/src/blocks_dal.rs
```diff
@@ -756,6 +920,52 @@ impl BlocksDal<'_, '_> {
...
+            SELECT
+                transactions.hash, transactions.error,
+                miniblock_number AS "miniblock_number!",
+                miniblocks.timestamp
+            FROM transactions
+            JOIN miniblocks ON transactions.miniblock_number = miniblocks.number
+            WHERE
+                (
+                    transactions.l1_batch_number IS NULL AND $2
+                    OR transactions.l1_batch_number = $1
```
perekopskiy: there is no index on `transactions.l1_batch_number` so this query will do sequential scan

core/lib/dal/src/blocks_web3_dal.rs
```diff
@@ -303,6 +303,36 @@ impl BlocksWeb3Dal<'_, '_> {
...
+                            (
+                                 SELECT number
+                                 FROM l1_batches
+                                          JOIN eth_txs_history ON l1_batches.eth_commit_tx_id = eth_txs_history.eth_tx_id
+                                 WHERE eth_txs_history.finality_status = 'finalized'
+                                 ORDER BY number DESC
+                                 LIMIT 1
+                            )
+                       ),
+                        0
```
perekopskiy, line 333: nit: formatting is off for the whole query

core/lib/dal/src/eth_sender_dal.rs
```diff
@@ -986,9 +993,16 @@ impl EthSenderDal<'_, '_> {
...
+            AggregatedActionType::L1Batch(L1BatchAggregatedActionType::Commit) => {
+                row.eth_commit_tx_id
+            }
+            AggregatedActionType::L1Batch(L1BatchAggregatedActionType::PublishProofOnchain) => {
+                row.eth_prove_tx_id
+            }
+            AggregatedActionType::L1Batch(L1BatchAggregatedActionType::Execute) => {
+                row.eth_execute_tx_id
+            }
+            _ => panic!("Unsupported"),
```
perekopskiy: nit: you can accept `L1BatchAggregatedActionType` as input and do not panic in the method

core/lib/types/src/api/mod.rs
```diff
@@ -91,9 +94,10 @@ impl<'de> Deserialize<'de> for BlockNumber {
                     "l1_committed" => BlockNumber::L1Committed,
                     "earliest" => BlockNumber::Earliest,
                     // For zksync safe is l1 committed. Real chances of revert are very low.
-                    "safe" => BlockNumber::L1Committed,
+                    "safe" => BlockNumber::Precommitted,
```
perekopskiy: change comment

core/node/block_reverter/src/lib.rs
```diff
@@ -531,7 +528,7 @@ impl BlockReverter {
             .context("`revertBatchesSharedBridge` function must be present in contract")?;
         let data = revert_function
             .encode_input(&[
-                Token::Uint(eth_config.hyperchain_id.as_u64().into()),
+                Token::Address(eth_config.sl_diamond_proxy_addr),
```
perekopskiy, line 539: it we merge this, then we won't be able to revert if chain is on <v29, right? Probably it's ok anyway, sending tx manually is always an option

core/node/eth_sender/src/publish_criterion.rs
```diff
@@ -317,11 +329,17 @@ impl GasConsts {
...
-    const L1_BATCH_EXECUTE_BASE_COST: u64 = 30_000;
+    const L1_BATCH_EXECUTE_BASE_COST: u64 = 50_000;
 
     /// Additional gas cost of processing `Execute` operation per L1->L2 tx.
     /// It's applicable iff SL is Ethereum.
-    const L1_OPERATION_EXECUTE_COST: u64 = 12_500;
+    const L1_OPERATION_EXECUTE_COST: u64 = 15_000;
+
+    /// Base gas cost of processing `Precommit` operation.
+    const PRECOMMIT_BASE_COST: u64 = 200_000;
```
perekopskiy, line 363: is it so expensive?

core/lib/dal/migrations/20250514091339_precommit_tx.up.sql
Deniallugo: i'd still keep it, it really helps if something is going wrong

core/lib/dal/src/blocks_dal.rs
Deniallugo: we don't have such method, range is not supposed to know all elements

core/node/block_reverter/src/lib.rs
Deniallugo: I assume, that it's easier to simply use old version of block reverter instead of supporting multiple versions here

core/lib/dal/src/blocks_dal.rs
Deniallugo: added an index, but we have to try to apply it first on stage independently. 
it's likely will take some time

core/node/eth_sender/src/publish_criterion.rs
Deniallugo: That is the numbers from @ly0va . 
I hope it will be dropped a little, i'll check the real consumption, on stage

core/lib/dal/src/blocks_dal.rs
Deniallugo: Moved out and splited to two methods

core/lib/dal/src/blocks_dal.rs
perekopskiy: I didn't mean to add an index, it will affect sequncing performance; it's possible to do it without an index (e.g. join with miniblocks and use miniblocks.l1_batch_number)

core/lib/dal/src/blocks_dal.rs
perekopskiy: you can use range.end() - range.start() + 1

core/lib/dal/src/blocks_dal.rs
Deniallugo: good idea

core/lib/dal/src/blocks_dal.rs
```diff
@@ -345,10 +355,42 @@ impl BlocksDal<'_, '_> {
...
+            r#"
+            SELECT
+                number,
+                l1_tx_count,
+                l2_tx_count,
+                timestamp
+            FROM
+                miniblocks
+            WHERE
+                eth_precommit_tx_id = $1
```
perekopskiy, line 436: miniblocks.eth_precommit_tx_id is not indexed and looks like it's the only place where index is needed. Adding index just for metrics is definitely overkill. TBH I don't know how to deal with it in the best way

core/lib/dal/src/blocks_dal.rs
```diff
@@ -549,14 +591,122 @@ impl BlocksDal<'_, '_> {
...
+                    )
+                ORDER BY number DESC
+                LIMIT 2
+            ) miniblocks ON TRUE
+            WHERE
+                l1_batches.number > 0
+                AND l1_batches.is_sealed
+                AND l1_batches.eth_commit_tx_id IS NULL
+                AND l1_batches.final_precommit_eth_tx_id IS NULL
+                AND miniblocks.rolling_txs_hash IS NOT NULL
```
perekopskiy, line 697: This query is expected to work fast because there is an index on `l1_batches.eth_commit_tx_id` so it will just scan rows where `l1_batches.eth_commit_tx_id IS NULL`. But please double-check it, maybe run this query against stage2 db and check its plan

core/lib/dal/src/blocks_dal.rs
```diff
@@ -756,6 +939,52 @@ impl BlocksDal<'_, '_> {
...
+            JOIN miniblocks ON transactions.miniblock_number = miniblocks.number
+            WHERE
+                (
+                    miniblocks.l1_batch_number IS NULL AND $2
+                    OR miniblocks.l1_batch_number = $1
+                )
+                AND
+                miniblocks.rolling_txs_hash IS NOT NULL
+                AND
+                miniblocks.eth_precommit_tx_id IS NULL
```
perekopskiy, line 1035: ok, looks like in this place the index I was talking about in the first comment is also required, so maybe it's worth just adding it. Its impact on miniblock header insertion time most likely will be negligible.
Also I think it makes sense to invert JOIN, i.e. select from miniblocks and join transactions, this way it will be guaranteed that index will be used.

core/node/eth_sender/src/aggregator.rs
```diff
@@ -275,6 +311,84 @@ impl Aggregator {
...
+                (Some(_), Some(committed_l1_batch_number)) => (
+                    committed_l1_batch_number + 1,
+                    Some(committed_l1_batch_number + 1),
+                ),
+                (Some(_), None) => (1.into(), Some(1.into())),
+            };
+
+        let txs = storage
+            .blocks_dal()
+            .get_ready_for_precommit_txs(desired_batch_number_for_db)
```
perekopskiy: When desired_batch_number_for_db is NULL there may be race condition, e.g. you have last_sealed_l1_batch == last_committed_l1_batch = 10, then pending L1 batch 11 is sealed, then you do get_ready_for_precommit_txs(None) and get results for L1 batch 12 which is now pending. I don't have immediate suggestions how to fix it

core/node/state_keeper/src/updates/l1_batch_updates.rs
```diff
@@ -15,6 +15,7 @@ pub struct L1BatchUpdates {
     pub txs_encoding_size: usize,
     pub l1_tx_count: usize,
     pub finished: Option<FinishedL1Batch>,
+    pub rolling_txs_hash: H256,
```
perekopskiy: is this one (L1BatchUpdates::rolling_txs_hash) even used?

core/node/eth_sender/src/aggregator.rs
Deniallugo: @perekopskiy, i found a solution for this. 
Meanwhile, i noticed, that disabling precommits, doesn't properly work, will address it, but the main  code won't change

core/node/eth_sender/src/eth_tx_aggregator.rs
```diff
@@ -90,6 +90,7 @@ pub struct EthTxAggregator {
     priority_tree_start_index: Option<usize>,
     settlement_layer: Option<SettlementLayer>,
     initial_pending_nonces: HashMap<Address, u64>,
+    needs_to_check_precomit: bool,
```
perekopskiy: typo: needs_to_check_precommit

core/node/eth_sender/src/eth_tx_aggregator.rs
```diff
@@ -706,14 +711,50 @@ impl EthTxAggregator {
...
+
+        let needs_to_precommit = storage
+            .blocks_dal()
+            .any_precommit_txs_after_batch(last_committed)
+            .await?;
+
+        if needs_to_precommit {
+            Ok(Some(PrecommitParams::fast_precommit()))
+        } else {
+            self.needs_to_check_precomit = true;
```
perekopskiy: should be false I suppose

core/node/eth_sender/src/aggregator.rs
```diff
@@ -237,6 +235,7 @@ impl Aggregator {
         l1_verifier_config: L1VerifierConfig,
         restrictions: OperationSkippingRestrictions,
         priority_tree_start_index: Option<usize>,
+        precommit_params: Option<&PrecommitParams>,
```
perekopskiy, line 238: nit: it's more common to use &Option<>

Deniallugo: disagree..

perekopskiy (approved):

core/node/eth_sender/src/aggregator.rs
perekopskiy: oh, actually you can just drop all & and implement Copy, but up to you

PR #4093, opened 2025-05-29
perf: deallocate all heaps that don't need to be kept

joonazan:
Only keep heaps containing decommitted code forever.

(A quick bugfix resulted in us keeping all heaps for the duration of a batch. This wastes memory.)

### DIFF
--- a/core/lib/multivm/src/versions/vm_latest/old_vm/memory.rs
+++ b/core/lib/multivm/src/versions/vm_latest/old_vm/memory.rs
@@ -286,10 +286,12 @@ impl<H: HistoryMode> Memory for SimpleMemory<H> {
         let is_returndata_page_static = last_callstack_this == CODE_ORACLE_ADDRESS;
         let current_observable_pages = self.observable_pages.inner().current_frame();
 
-        // It is possible that the code oracle runs out of gas after decommitting code.
-        // The page with the code needs to be kept but the zero page is returned instead
-        // in that case.
-        let returndata_page = if is_returndata_page_static {
+        let returndata_page = if is_returndata_page_static && returndata_fat_pointer.length == 0 {
+            // This only happens when the code oracle runs out of gas after decommitting code.
+            // The page with the code needs to be kept but the zero page is returned instead
+            // in that case.
+            // If the page has been decommitted in past, this erroneously keeps an empty heap
+            // alive but that is a very rare case.
             heap_page_from_base(base_page).0
         } else {
             returndata_fat_pointer.memory_page
--- a/core/lib/multivm/src/versions/vm_latest/old_vm/memory.rs
+++ b/core/lib/multivm/src/versions/vm_latest/old_vm/memory.rs
@@ -278,19 +278,31 @@ impl<H: HistoryMode> Memory for SimpleMemory<H> {
 
     fn finish_global_frame(
         &mut self,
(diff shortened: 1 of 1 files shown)


core/lib/multivm/src/versions/vm_latest/old_vm/memory.rs
```diff
@@ -278,19 +278,31 @@ impl<H: HistoryMode> Memory for SimpleMemory<H> {
...
-        let is_returndata_page_static =
-            last_callstack_this == CODE_ORACLE_ADDRESS && returndata_fat_pointer.length > 0;
+        let is_returndata_page_static = last_callstack_this == CODE_ORACLE_ADDRESS;
+        let current_observable_pages = self.observable_pages.inner().current_frame();
+
+        // It is possible that the code oracle runs out of gas after decommitting code.
+        // The page with the code needs to be kept but the zero page is returned instead
+        // in that case.
+        let returndata_page = if is_returndata_page_static {
+            heap_page_from_base(base_page).0
```
0xVolosnikov, line 293: A little bit dangerous - if we modify CodeOracle in the future it may be hard to debug

joonazan: This is true.

I know nothing about CodeOracle. Maybe we could change it instead?

Though at least there is no chance of such a mistake getting into production because then the decommitted code would always be zero and tests would fail.

0xVolosnikov: Also, iiuc, if CodeOracle just forwards old memory page with bytecode, it won't be saved in "returndata_page", right?

core/lib/multivm/src/versions/vm_latest/old_vm/memory.rs
```diff
@@ -278,19 +278,31 @@ impl<H: HistoryMode> Memory for SimpleMemory<H> {
...
-        // We must keep this memory page forever for future decommits.
-        let is_returndata_page_static =
-            last_callstack_this == CODE_ORACLE_ADDRESS && returndata_fat_pointer.length > 0;
+        let is_returndata_page_static = last_callstack_this == CODE_ORACLE_ADDRESS;
+        let current_observable_pages = self.observable_pages.inner().current_frame();
+
+        // It is possible that the code oracle runs out of gas after decommitting code.
+        // The page with the code needs to be kept but the zero page is returned instead
+        // in that case.
+        let returndata_page = if is_returndata_page_static {
```
0xVolosnikov, line 292: Check forwarding case

0xVolosnikov (changes requested):

core/lib/multivm/src/versions/vm_latest/old_vm/memory.rs
0xVolosnikov: > Also, iiuc, if CodeOracle just forwards old memory page with bytecode, it won't be saved in "returndata_page", right?

joonazan: Documented and reduced the impact to just cases where already decommitted and OOM.

0xVolosnikov:
You changed the order of `push_to_frame` and `merge_frame` - is it ok?

PR #4102, opened 2025-05-30
feat: [NOT FOR MERGE] state keeper leader rotation integration with consensus

perekopskiy:
## Why ❔

Preparation for leader rotation

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


core/lib/mempool/src/mempool_store.rs
```diff
@@ -246,6 +246,28 @@ impl MempoolStore {
...
+                if let Some(score) = metadata.previous_score {
+                    self.l2_priority_queue.remove(&score);
+                }
+                if let Some(score) = metadata.new_score {
+                    self.l2_priority_queue.insert(score);
+                }
+            }
+        }
+    }
+
```
brunoffranca: IIUC, this already handles the mempool updating itself based on new blocks? So the consensus component would just need to run a state keeper instance and, for each block that it processes as a replica, feed the block to the state keeper and the AdvanceInput to the mempool? Where would I get this AdvanceInput from?

core/lib/vm_interface/src/executor.rs
```diff
@@ -48,6 +48,9 @@ pub trait BatchExecutor<S>: 'static + Send + fmt::Debug {
 
     /// Return gas remaining in VM instance.
     async fn gas_remaining(&mut self) -> anyhow::Result<u32>;
+
+    /// Rolls back the last executed l2 block.
+    async fn rollback_l2_block(&mut self) -> anyhow::Result<()>;
```
brunoffranca: Can this method be called several times? I.e. can I rollback several blocks, one at a time?

core/node/node_sync/src/leader_io.rs
```diff
@@ -0,0 +1,176 @@
...
+/// creates new blocks, decides when to seal block and batch.
+///
+/// Otherwise, it receives a sequence of actions from the fetcher via the action queue and propagates it
+/// into the state keeper.
+#[derive(Debug)]
+pub struct LeaderIO {
+    mempool_io: MempoolIO,
+    external_io: ExternalIO,
+    is_active_leader: bool,
+}
```
brunoffranca: More a nit than anything else. I don't believe we will have any use for MempoolIO, so maybe it would make sense for LeaderIO to replace MempoolIO. ExternalIO I can see being used for full nodes.

core/node/node_sync/src/leader_io.rs
```diff
@@ -0,0 +1,176 @@
...
+    async fn load_batch_state_hash(&self, number: L1BatchNumber) -> anyhow::Result<H256> {
+        // Note: impls are the same for mempool and external.
+        self.mempool_io.load_batch_state_hash(number).await
+    }
+
+    // TODO: what non-leader should do? signal to consensus that block is invalid?
+    async fn reject(&mut self, tx: &Transaction, reason: UnexecutableReason) -> anyhow::Result<()> {
+        let io = self.active_io_ref_mut();
+        io.reject(tx, reason).await
+    }
```
brunoffranca: I assume this is called when a given transaction fails. By whom though? Anyway, if the block is invalid, it should be communicated in someway to consensus so it can move to the next leader. The same way that the block being valid needs to be communicated (right now the consensus just watches DAL and waits for the block to appear, but we might want to use the same signalling mechanism for valid and invalid blocks).

core/node/node_sync/src/leader_io.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    // TODO: what non-leader should do? signal to consensus that block is invalid?
+    async fn rollback(&mut self, tx: Transaction) -> anyhow::Result<()> {
+        let io = self.active_io_ref_mut();
+        io.rollback(tx).await
+    }
```
brunoffranca: I see it's different from `rollback_block` below. When would this be called and by whom?

core/node/state_keeper/src/io/mempool.rs
```diff
@@ -383,6 +385,45 @@ impl StateKeeperIO for MempoolIO {
...
+        {
+            let constraint = self.mempool.rollback(&tx);
+            to_add.push((tx, constraint));
+        }
+
+        to_add.reverse();
+        self.mempool.insert(to_add, HashMap::new());
+
+        Ok(())
+    }
```
brunoffranca: If we are rolling back a block, do we really want to add its transactions back to the mempool? The block failed for some reason.

core/node/state_keeper/src/keeper.rs
```diff
@@ -830,19 +848,52 @@ impl StateKeeperInner {
...
+            if pending_l2_block_header.number.0 != 1 {
+                self.rollback().await?;
+                match mode {
+                    RunMode::Propose => {
+                        self.propose(&mut stop_receiver).await?;
+                    }
+                    RunMode::Verify => self.verify(&mut stop_receiver).await?,
+                    RunMode::Default => self.run_block(&mut stop_receiver).await?,
+                }
+            }
```
brunoffranca: Does this run for a single block or until stopped?

core/lib/mempool/src/mempool_store.rs
perekopskiy: >IIUC, this already handles the mempool updating itself based on new blocks?

Yes. It's called in state keeper after it commits the block (i.e. currently when it starts new block), no need to call it in consensus directly

core/lib/vm_interface/src/executor.rs
perekopskiy: At the moment -- no. BatchExecutor is relatively easy to adapt to be able to rollback several blocks, in other places, e.g. UpdatesManager it will be more difficult to rollback several blocks, but it looks possible

core/node/node_sync/src/leader_io.rs
perekopskiy: Agree, but I'd prefer to do it later

core/node/node_sync/src/leader_io.rs
perekopskiy: It's called in state keeper when either
a) transaction validation fails e.g. initiator doesn't have enough eth to cover fee
b) transaction doesn't fit into some sealing criteria (e.g. it publishes more pubdata than it's allowed per batch)

For leader node it's clear what to do, it simply doesn't include such tx. For non-leader we will have to handle it properly, currently it just panics (see ExternalIO::reject).

core/node/node_sync/src/leader_io.rs
perekopskiy: It's similar to `reject` in some sense. It's called in state keeper when it executes tx and after that sees that the batch doesn't fit into some sealing criterion. So it does `rollback` and seals the batch.
If it's called for non-leader then it should be treated the same way as `reject` -- i.e. assume that the block is invalid.

core/node/state_keeper/src/io/mempool.rs
perekopskiy: We should. Block failed for a reason, not because transactions are invalid but leader did something wrong when building block. Transactions can be re-included in a valid block by some other non-malicious leader

core/node/state_keeper/src/keeper.rs
perekopskiy: Until stopped. Consensus component shouldn't use this function. It's supposed to call `propose`/`verify` one by one. I've left this function to run state keeper in old-fashioned way, until we have consensus integration and for tests

core/node/node_sync/src/leader_io.rs
brunoffranca: Sure, no problem.

core/lib/vm_interface/src/executor.rs
brunoffranca: Not relevant for this PR. But, for us to be able to pipeline execution and consensus we need to be able to rollback up to 2 blocks. Or we need some way of producing a block without executing it (which seems more complex).

core/node/node_sync/src/leader_io.rs
brunoffranca: Got it. Yes, non-leader needs to reject the block.

core/node/state_keeper/src/io/mempool.rs
brunoffranca: Fair enough.

PR #4104, opened 2025-06-02
feat(en): Use config system for env-based EN configuration

slowli:
## What ❔

Uses `smart-config` for the external node configuration in all cases.

## Why ❔

Previously, EN used `smart-config` only with file-based configs, and there was an awkward / hard-to-maintain translation layer for configs.

## Is this a breaking change?

- [x] Yes (technically, for some experimental config params)
- [ ] No

## Operational changes

Will be documented in a separate PR.

## Checklist


core/node/metadata_calculator/src/lib.rs
```diff
@@ -102,26 +103,31 @@ pub struct MetadataCalculatorConfig {
 }
 
 impl MetadataCalculatorConfig {
-    pub fn for_main_node(
+    // FIXME: not using `OperationsManagerConfig` is technically a breaking change, but it seems to have low impact
```
slowli: Is this change OK? We do have multiple `operations_manager.delay_interval` mentions in the env configs, but AFAICT, they all have the default value (100ms).

core/bin/external_node/src/config/mod.rs
```diff
@@ -278,834 +207,180 @@ where
...
+        let mut schema = ConfigSchema::default();
+        schema.insert(&PrometheusConfig::DESCRIPTION, "prometheus")?;
+        schema
+            .insert(&ObservabilityConfig::DESCRIPTION, "observability")?
+            .push_deprecated_alias("")?;
+
+        schema.insert(&ApiConfig::DESCRIPTION, "api")?;
+        schema
+            .single_mut(&Web3JsonRpcConfig::DESCRIPTION)?
+            .push_alias("api")? // FIXME: is this OK (used for single param)?
```
slowli: IMO, it would be OK to have this alias for shorter env var names (e.g., `EN_API_HTTP_PORT` instead of `EN_API_WEB3_JSON_RPC_HTTP_PORT`). Not using any prefix at all, like we're doing rn, looks too ambiguous to me, hence `push_deprecated_alias("")` below.

core/lib/config/src/configs/en_config.rs
```diff
@@ -4,27 +4,52 @@ use smart_config::{
...
+pub struct SharedL1ContractsConfig {
+    /// Address of the L1 diamond proxy. EN fetches most contract addresses from the L2 peer,
+    /// but the diamond proxy address is still available locally as the root of trust.
+    pub diamond_proxy_addr: Option<Address>,
+}
 
 /// Temporary config for initializing external node, will be completely replaced by consensus config later
+// FIXME: rename?
 #[derive(Debug, Clone, PartialEq, DescribeConfig, DeserializeConfig)]
 pub struct ENConfig {
```
slowli: I've named the canonical prefix for this config `networks` (i.e., would make sense to name it `NetworksConfig`), but other suggestions would be appreciated. What I don't really like is positioning the config as a "config collecting all EN-specific params"; configs should be split by the component / domain.

core/bin/zksync_server/src/node_builder.rs
```diff
@@ -424,90 +428,70 @@ impl MainNodeBuilder {
...
-        };
-        if with_debug_namespace {
-            namespaces.push(Namespace::Debug)
+        // FIXME: Why doesn't the node do what I'm telling it to do?
+        let mut namespaces = rpc_config.api_namespaces.clone();
+        if state_keeper_config.shared.save_call_traces {
+            namespaces.insert(Namespace::Debug);
         }
-        namespaces.push(Namespace::Snapshots);
+        namespaces.insert(Namespace::Snapshots);
```
slowli: Subj. I find it weird that we force some namespaces to be enabled.

core/bin/external_node/src/config/tests/config.yaml
```diff
@@ -0,0 +1,129 @@
...
+  pushgateway_url: http://prometheus/
+  push_interval_ms: 150
+observability:
+  log_format: json
+  log_directives: warn,zksync=info
+  sentry_url: https://example.com/new
+  sentry_environment: 'mainnet - mainnet2'
+
+postgres:
+  server_url: postgres://postgres:notsecurepassword@localhost:5432/en
```
Deniallugo, line 13: It looks like a mix between secrets and general config, let's split it

slowli: That's a test configuration though; what purpose would a split serve other than having to manage more files around?

core/node/metadata_calculator/src/lib.rs
Deniallugo: I always find the name `operations_manager` is ultra confusing. 
I think we can introduce the default value for it and that's all

core/bin/external_node/src/node_builder.rs
```diff
@@ -188,35 +173,33 @@ impl<R> ExternalNodeBuilder<R> {
...
-        let save_call_traces = self
-            .config
-            .optional
-            .api_namespaces()
+        // FIXME: this is backwards / unobvious. Can readily use `config.state_keeper.save_call_traces` instead.
+        let save_call_traces = config
+            .api
+            .web3_json_rpc
+            .api_namespaces
             .contains(&Namespace::Debug);
```
perekopskiy: I agree that it should be changed, but we should be careful/communicate breaking change really well. If node is launched and save_call_traces=false, it will be difficult to backfill it (most likely rollback block and reprocess is the easiest option)

core/bin/zksync_server/src/node_builder.rs
perekopskiy: IMO it's ok to do a breaking change here and remove kludges for `debug` and `snapshots`. Contrary to not saving call traces, not enabling namespace is much less desctructive

core/node/metadata_calculator/src/lib.rs
perekopskiy: IMO, it's OK

core/bin/external_node/src/node_builder.rs
slowli: Yep, my thoughts as well. Will add a task to the backlog then.

perekopskiy (dismissed):

PR #4111, opened 2025-06-02
feat(fee_model): scale the batch fee unconditionally

dimazhornyk:
## What ❔

Apply scaling to the batch fee even if there is an open batch.

## Why ❔

When the transaction gas price is estimated at the end of the batch, there is a risk of transaction price being too low if the next batch has a higher gas price

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/core/lib/types/src/fee_model.rs
+++ b/core/lib/types/src/fee_model.rs
@@ -68,6 +68,21 @@ impl BatchFeeInput {
                 })
             })
     }
+
+    pub fn scale_linearly(
+        &self,
+        l1_gas_price_scale_factor: f64,
+        l1_pubdata_price_scale_factor: f64,
+    ) -> BatchFeeInput {
+        match self {
+            BatchFeeInput::L1Pegged(input) => {
+                BatchFeeInput::L1Pegged(input.scale(l1_gas_price_scale_factor))
+            }
+            BatchFeeInput::PubdataIndependent(input) => BatchFeeInput::PubdataIndependent(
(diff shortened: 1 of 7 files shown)


core/lib/types/src/fee_model.rs
```diff
@@ -182,6 +206,25 @@ pub struct PubdataIndependentBatchFeeModelInput {
...
+        l1_pubdata_price_scale_factor: f64,
+    ) -> PubdataIndependentBatchFeeModelInput {
+        clip_batch_fee_model_input_v2(PubdataIndependentBatchFeeModelInput {
+            fair_l2_gas_price: (self.fair_l2_gas_price as f64
+                * (l1_gas_price_scale_factor + l1_pubdata_price_scale_factor)
+                / 2.0) as u64,
+            fair_pubdata_price: (self.fair_pubdata_price as f64 * l1_pubdata_price_scale_factor)
+                as u64,
+            l1_gas_price: (self.l1_gas_price as f64 * l1_gas_price_scale_factor) as u64,
+        })
```
perekopskiy: End users only care if base fee and gas per pubdata are high enough. I would suggest to scale only `fair_l2_gas_price`, this way both base fee and gas per pubdata will be scaled (please mention it in the comment).
Also, I think it's better to use new optional config var for this purpose

dimazhornyk: makes sense, I added an optional config var and changed the code to only scale the `fair_l2_gas_price`

PR #4124, opened 2025-06-04
feat(api): implement `unstable_gasPerPubdata`

itegulov:
## What ❔

Adds `unstable_gasPerPubdata` that is to be stabilized as `zks_gasPerPubdata` in the future.

## Why ❔

Preparation for deprecation of `zks_estimateFee`

## Is this a breaking change?
- [ ] Yes
- [x] No

## Operational changes
Whitelist  `unstable_gasPerPubdata` in proxyd

## Checklist


core/lib/web3_decl/src/namespaces/unstable.rs
```diff
@@ -68,4 +68,7 @@ pub trait UnstableNamespace {
         &self,
         tx_bytes: Bytes,
     ) -> RpcResult<TransactionDetailedResult>;
+
+    #[method(name = "gasPerPubdata")]
+    async fn gas_per_pubdata(&self) -> RpcResult<U256>;
```
popzxc, line 73: Would it help us to implement re-execution of blocks in `anvil-zksync` if we could provide block ID here?
E.g. if we pass `latest`, we get current value, for particular block numbers we get historical data.

itegulov: `zks_getBlockDetails` already exposes everything we need to calculate gas per pubdata on `anvil-zksync`'s side, so IMHO no need for historical blocks here. Technically, we could avoid adding this method entirely and do the same on the client's side. But then we'd be screwed if formula was to ever change.

Also it feels more natural to mimic `eth_gasPrice`'s API here

PR #4125, opened 2025-06-04
feat(private-rpc): invalidate old token when new one is created

tomg10:


private-rpc/src/routes/users-routes.ts
```diff
@@ -23,6 +24,8 @@ export function usersRoutes(app: WebServer) {
             throw new HttpError('forbidden', 403);
         }
 
+        await app.context.db.delete(usersTable).where(eq(usersTable.address, address));
```
iscxyz: Are insert and delete going to happen in the same transaction?

tomg10: nope, but you're right that would be better if they did

PR #4126, opened 2025-06-04
feat(config): Report config params in more ways

slowli:
## What ❔

Reports config params as `INFO` metrics and as the `config` component in the healthcheck (the latter is opt-in).

## Why ❔

More observability; more possibilities for automation. 

## Is this a breaking change?

- [ ] Yes
- [x] No

## Operational changes

No necessary changes. New possibilities will be documented in a separate PR.

## Checklist

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -212,8 +212,8 @@ proc-macro2 = "1.0"
 trybuild = "1.0"
 
 # "Internal" dependencies
-vise = "0.3.0"
-vise-exporter = "0.3.0"
+vise = "0.3.1"
+vise-exporter = "0.3.1"
 smart-config = { version = "0.1.0", git = "https://github.com/matter-labs/smart-config.git", rev = "6bc86aa7833ce08276861951ba72b4426a37a893" }
 smart-config-commands = { version = "0.1.0", git = "https://github.com/matter-labs/smart-config.git", rev = "6bc86aa7833ce08276861951ba72b4426a37a893" }
 foundry-compilers = { version = "0.11.6", git = "https://github.com/Moonsong-Labs/compilers.git", rev = "7c69695e5c75451f158dd2456bf8c94a7492ea0b" }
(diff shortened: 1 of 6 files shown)


core/Cargo.toml
```diff
@@ -345,3 +345,7 @@ zksync_node_api_server = { version = "28.4.0-non-semver-compat", path = "node/ap
 zksync_base_token_adjuster = { version = "28.4.0-non-semver-compat", path = "node/base_token_adjuster" }
 zksync_logs_bloom_backfill = { version = "28.4.0-non-semver-compat", path = "node/logs_bloom_backfill" }
 zksync_gateway_migrator = { version = "28.4.0-non-semver-compat", path = "node/gateway_migrator" }
+
+[patch.crates-io]
+vise = { git = "https://github.com/matter-labs/vise.git", rev = "8c64392753ab8fe563c6fa3db5e4d987952ad011" }
+vise-exporter = { git = "https://github.com/matter-labs/vise.git", rev = "8c64392753ab8fe563c6fa3db5e4d987952ad011" }
```
slowli: Sort of blocks the PR merge if we want to have publishable crates at all times. [The release PR](https://github.com/matter-labs/vise/pull/38) for `vise` is borken because of not bumping crate versions in some places (but it can be fixed manually).

core/lib/config/src/observability_ext/metrics.rs
```diff
@@ -0,0 +1,36 @@
...
+    value: Option<String>,
+    default: bool,
+    secret: bool,
+}
+
+#[derive(Debug, Metrics)]
+#[metrics(prefix = "server_config")]
+pub(super) struct ConfigMetrics {
+    #[metrics(labels = ["path"])]
+    params: LabeledFamily<String, Info<ConfigParamInfo>>,
```
slowli, line 19: Does exposing params as metrics make sense? AFAIU, since metric labels don't change during node operations, metrics shouldn't consume much space in the Prometheus data store, despite there being ~150 params reported for the main node. (This number would be reduced once config is decentralized, and only necessary configs are parsed.)

perekopskiy: IMO it's useful, checking metric is easier than querying health endpoint, also it captures historical data

PR #4129, opened 2025-06-04
docs(en): Improve config docs for node

slowli:
## What ❔

Documents basic features of the new config system.

## Why ❔

## Is this a breaking change?

- [ ] Yes
- [x] No

## Operational changes

No operational changes.

## Checklist


core/lib/config/src/configs/object_store.rs
```diff
@@ -45,28 +46,52 @@ impl ObjectStoreConfig {
 #[config(tag = "mode")]
 #[serde(tag = "mode")]
 pub enum ObjectStoreMode {
+    /// Writable GCS bucket with ambient authentication.
     #[config(alias = "Gcs")]
-    GCS { bucket_base_url: String },
+    GCS {
+        /// Name of the bucket.
+        // FIXME: this isn't a URL; would it make sense to rename it to `bucket_name`?
+        bucket_base_url: String,
```
slowli: Subj. IMO, such renaming would make sense, and the old name could be retained as a (deprecated?) alias.

core/node/gateway_migrator/src/node/settlement_layer_data.rs
```diff
@@ -354,12 +354,10 @@ fn adjust_eth_sender_config(
...
     if settlement_layer.is_gateway() {
         config.max_aggregated_tx_gas = 30000000000;
-        config.max_eth_tx_data_size = 550_000;
         tracing::warn!(
             "Settling to Gateway requires to adjust ETH sender configs: \
-               max_aggregated_tx_gas = {}, max_eth_tx_data_size = {}",
-            config.max_aggregated_tx_gas,
-            config.max_eth_tx_data_size
+               max_aggregated_tx_gas = {}",
+            config.max_aggregated_tx_gas
```
slowli, line 385: @Deniallugo Am I not breaking anything here? AFAICT, `max_eth_tx_data_size` is never read in the repo workspaces (at least not directly), but I guess some very roundabout reads could be missed.

noelportillo (approved):

perekopskiy (approved):

core/lib/config/src/configs/object_store.rs
slowli: Apparently, this can be an URL (at least, judging by [the usage in `zkstack`](https://github.com/matter-labs/zksync-era/blob/db4f036a5c4fd450178d709793a62ddca43f54ec/zkstack_cli/crates/zkstack/src/commands/prover/gcs.rs#L36)), so I'm removing this FIXME for now.

core/node/gateway_migrator/src/node/settlement_layer_data.rs
Deniallugo: TBH, looks like it's abandoned

PR #4131, opened 2025-06-04
fix(gateway): Do not create l2 client if server is settlment layer

Deniallugo:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/core/node/gateway_migrator/src/node/settlement_layer_data.rs
+++ b/core/node/gateway_migrator/src/node/settlement_layer_data.rs
@@ -18,7 +18,8 @@ use zksync_dal::{
 };
 use zksync_eth_client::{
     contracts_loader::{
-        get_server_notifier_addr, get_settlement_layer_from_l1, load_settlement_layer_contracts,
+        get_server_notifier_addr, get_settlement_layer_from_l1, is_settlement_layer,
+        load_settlement_layer_contracts,
     },
     node::SenderConfigResource,
     EthInterface,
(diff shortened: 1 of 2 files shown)


core/node/gateway_migrator/src/node/settlement_layer_data.rs
```diff
@@ -319,6 +325,25 @@ impl WiringLayer for SettlementLayerData<ENConfig> {
...
+    eth_client: &dyn EthInterface,
+    bridgehub_address: Address,
+    l2_chain_id: L2ChainId,
+    gateway_rpc_url: Option<SensitiveUrl>,
+) -> anyhow::Result<Option<Box<DynClient<L2>>>> {
+    // If the server is the settlement layer, the gateway is the server itself,
+    // so we can't point to ourselves.
+    let is_settlement_layer = is_settlement_layer(eth_client, bridgehub_address, l2_chain_id)
+        .await
+        .unwrap_or_default();
```
perekopskiy: any reason to use `unwrap_or_default` here instead of propagating error?

Deniallugo: it just safe to do it this way. 
If somethign wrong happened with contracts, we will do the safest way and create a l2_client

perekopskiy: hmm, won't it panic anyway later if this is a gateway or chain that settles to gateway?

slowli: IMO, it'd make sense to at least log a `WARN`ing in case of an error.

Deniallugo: It would panic only in a situation, if the chain is gateway by it self and it was in a crash loop before

perekopskiy: ok, agree, please leave a comment then

Deniallugo: i just dont want to stop the process if something is bad with this particular method

slowli (commented):
IMO, a better solution to "everything fails on startup if L1 isn't available" problem would be to resolve the relevant gateway-related resources lazily (i.e., not during wiring, but rather using something like `Arc<OnceCell<_>>`). But it'd require a lot more work 🙃

core/node/gateway_migrator/src/node/settlement_layer_data.rs
Deniallugo: tbh. the result is absolutely the same. let's just return an error

Deniallugo:
> IMO, a better solution to "everything fails on startup if L1 isn't available"

That's not a problem i'm solving with this pr. 
The problem is recursive dependency if the server we are running is settlement layer

perekopskiy (approved):

PR #4153, opened 2025-06-06
chore(external-node): Clean up Docker tags for external node docherhub

tomg10:
This PR intends to:
a) stop publishing external-node "SHA tags" images for public dockerhub
b) change so that latest tag is pushed only for explicitly tagged versions that are not alpha, for alpha we push latest-alpha
d) stop pushing latest2.0 version completely for public dockerhub
After those changes the only tags will be latest, latest-alpha and releases versions tags


otani88 (approved):

noelportillo (approved):

deniallugo-ml:

PR #4175, opened 2025-06-12
fix: Changes bootloader hash

EmilLuta:
I can't tell exactly why this works, but without it, sequencer code will not run init at all. After changing it runs. Unclear if this is the right fix. Whomever reviews & approves, please leave a comment on why this is fine.


noelportillo (commented):
Good

noelportillo (commented):
Ok

PR #4190, opened 2025-06-17
feat: Introduce whitelisted logic for prividium mode

Jrigada:
## What ❔
- Added `whitelisted_wallets` to the YAML configuration, allowing for flexible wallet authorization.
- Updated the `Authorizer` class to handle the new `whitelisted_wallets` property, supporting both 'all' and specific wallet addresses.
- Enhanced `YamlParser` to parse and validate the `whitelisted_wallets` field.
- Implemented a new route in `usersRoutes` to check if a user's address is whitelisted based on the updated logic.

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


private-rpc/src/permissions/authorizer.ts
```diff
@@ -50,7 +52,15 @@ export class Authorizer {
         console.log(`loading permissions from ${filePath}`);
         this.permissions = new Map();
         this.postReadFilters = new Map();
-        new YamlParser(filePath).load_rules(this);
+        const parser = new YamlParser(filePath);
+        parser.load_rules(this);
+
+        const wallets = parser.getWhitelistedWallets();
+        if (wallets.includes('all')) {
```
aon: Using an array seems deceiving here. Can we change it to be either an array or a string literal "all"?

private-rpc/src/permissions/authorizer.ts
```diff
@@ -50,7 +52,15 @@ export class Authorizer {
...
+            this.whitelistedWallets = 'all';
+        } else {
+            this.whitelistedWallets = new Set(wallets as Address[]);
```
aon: Validation should be performed here, or in the zod schema to check passed addresses are actually address type.

private-rpc/src/permissions/yaml-parser.ts
```diff
@@ -42,20 +42,25 @@ const methodSchema = z.object({
...
-        })
-    )
+});
+
+type RawContract = z.infer<typeof contractSchema>;
+
+const yamlSchema = z.object({
+    whitelisted_wallets: z
+        .array(z.union([addressSchema, z.literal('all')]))
+        .nonempty({ message: 'whitelisted_wallets cannot be empty. To allow all wallets, use ["all"].' }),
```
aon: I didn't know this existed, very nice.

private-rpc/src/permissions/yaml-parser.ts
```diff
@@ -89,23 +98,26 @@ export class YamlParser {
...
                 throw new Error('Unknown rule type');
         }
     }
 
     load_rules(authorizer: Authorizer) {
-        this.yaml.contracts.forEach((rawContract) => {
-            const readPermissions = rawContract.methods.map((method) => {
+        this.yaml.contracts.forEach((rawContract: RawContract) => {
+            const readPermissions = rawContract.methods.map((method: RawMethod) => {
                 const selector = method.signature === '#BASE_TOKEN_TRANSFER' ? '0x' : this.extractSelector(method);
```
aon: No new changes here, please revert in order to avoid conflicts.

private-rpc/src/routes/users-routes.ts
```diff
@@ -35,4 +35,24 @@ export function usersRoutes(app: WebServer) {
...
+
+    app.get('/:address', getUserSchema, (req, reply) => {
+        const { address } = req.params;
+        const { authorizer } = app.context;
+
+        if (authorizer.whitelistedWallets === 'all') {
+            return reply.send({ authorized: true });
+        }
+
+        const isWhitelisted = authorizer.whitelistedWallets.has(address);
```
aon: This logic should be moved into the authorizer class.

aon (approved):

aon (dismissed):

private-rpc/src/permissions/authorizer.ts
```diff
@@ -45,12 +47,27 @@ export class Authorizer {
         return this.postReadFilters.get(`${address}:${method}`) || null;
     }
 
+    isAddressWhitelisted(address: Address): boolean {
```
calvogenerico, line 50: Does it make sense to call this something like "addressWhitelistedForExplorer"? Because this does not forbides other addresses to use the rest of the system, is just to access the explorer.

calvogenerico (approved):
Looks great!

I left a comment, but no stopper at all.

private-rpc/example-permissions.yaml
```diff
@@ -1,3 +1,9 @@
+# To allow any wallet to connect, use the literal string "all".
+# Example: whitelisted_wallets: "all"
+
+# To restrict access to specific wallets, provide a non-empty list of addresses.
+whitelisted_wallets: "all"
```
jcsec-security: Setting the example permission as `all` might result in unexpectedly exposed services if template is carelessly copy-pasted. I would say it is better to just to add a random address as whitelisting example.

Jrigada: I agree, let me just check the test suite because I think we might be using random addresses. In that case I will change that in order to control wether they are whitelisted or not.

zkstack_cli/crates/zkstack/src/commands/private_rpc/mod.rs
```diff
@@ -34,6 +34,9 @@ pub struct PrivateRpcCommandInitArgs {
     /// Initializes private proxy with network host mode. This is useful for environments that don't support host.docker.internal.
     #[clap(long)]
     pub docker_network_host: bool,
+    /// Use test permissions file for integration tests
+    #[clap(long)]
+    pub integration_test: bool,
```
tomg10: can you instead change integration tests to change whitelisted wallets (see private-rpc-permissions-editor.ts)? I want to avoid adding this logic just for that

zkstack_cli/crates/zkstack/src/commands/private_rpc/mod.rs
```diff
@@ -164,7 +164,7 @@ pub async fn init(shell: &Shell, args: PrivateRpcCommandInitArgs) -> anyhow::Res
 
     initialize_private_rpc_database(shell, &chain_config, &db_config).await?;
 
-    let src_permissions_path = "example-permissions.yaml";
+    let src_permissions_path = ecosystem_path.join("private-rpc/example-permissions.yaml");
```
tomg10: Are you sure about this change? It's a file from zksync-era, ecosystem_path doesn't necessarily contain the repo contents, right?

Jrigada: I will rollback the change just to be safe, I think in reality it should be the same but the less changes the better.

tomg10 (approved):

PR #4197, opened 2025-06-18
feat(state-keeper): implement block commit/rollback

perekopskiy:
## What ❔

Another PR with preparation for leader rotation. It adds block commit/rollback capabilities to state keeper including batch executor, updates manager, IO

## Why ❔

Prerequisite for leader rotation.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Operational changes

None

## Checklist


core/lib/vm_executor/src/batch/factory.rs
```diff
@@ -326,6 +330,10 @@ impl<S: ReadStorage + 'static, Tr: BatchTracer> CommandReceiver<S, Tr> {
         system_env: SystemEnv,
         pubdata_builder: Rc<dyn PubdataBuilder>,
     ) -> anyhow::Result<StorageView<S>> {
+        // Maximal block depth that can be rolled back. 2 is enough for the current consensus implementation.
+        // In simpler words, value equals 2 means that block #X can be rolled back up until block #(X+2) is started.
+        const MAX_BLOCK_ROLLBACK_DEPTH: usize = 2;
+
```
brunoffranca: Can we make it larger if necessary? What's the drawback?

perekopskiy: Yes, we will have to keep more VM snapshots in memory; Also maybe it's worth reworking the approach: add CommitBlock command to batch executor so we can get rid of this constant

perekopskiy: I've replaced the constant with CommitBlock command so now batch executor supports arbitrary rollback depth

core/lib/mempool/src/mempool_store.rs
```diff
@@ -23,7 +23,7 @@ pub struct MempoolStats {
 #[derive(Debug)]
 pub struct MempoolStore {
     /// Pending L1 transactions
-    l1_transactions: HashMap<PriorityOpId, L1Tx>,
+    l1_transactions: BTreeMap<PriorityOpId, L1Tx>,
```
RomanBrodetski, line 26: please write a comment what operations you do on this field so that it's clear why this data structure fits well

core/lib/vm_executor/src/batch/factory.rs
```diff
@@ -356,13 +360,18 @@ impl<S: ReadStorage + 'static, Tr: BatchTracer> CommandReceiver<S, Tr> {
             }
         }
 
+        let mut has_snapshot_before_tx = false;
```
RomanBrodetski, line 363: oh uh ... Is it hard to have maybe two snapshot Queues - one before tx and one before block? If not, we'd need to do it like this but it's so error prone..

core/node/state_keeper/src/io/mempool.rs
```diff
@@ -286,6 +288,45 @@ impl StateKeeperIO for MempoolIO {
...
+            to_add.push((tx, constraint));
+        }
+
+        to_add.reverse();
+        self.mempool.insert(to_add, HashMap::new());
+
+        Ok(())
+    }
+
+    async fn advance_nonces(&mut self, txs: Box<&mut (dyn Iterator<Item = &Transaction> + Send)>) {
```
RomanBrodetski: is it clear that `advance_nonces` also advances the `serial_id`? Maybe it is - you know better - but if not consider commenting/name change

RomanBrodetski (approved):
I understand where the need for this comes from, but the logic is quite complex. I don't see a way to simplify it. Approving to avoid blocking you, but may have another look tomorrow morning.

core/lib/vm_executor/src/batch/factory.rs
perekopskiy: It's possible to have two queues, but it will make VM interface strongly tied to state keeper. IMO it's better to keep generic snapshot/rollback interface for VM. Agree that it's easy to forget about keeping `has_snapshot_before_tx` consistent with VM state but I don't expect any changes to this code in the near future

PR #4223, opened 2025-06-23
fix: add InteropBatchRootProcessor

zkzoomer:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/core/node/eth_watch/src/event_processors/appended_chain_batch_root_interop.rs
+++ b/core/node/eth_watch/src/event_processors/appended_chain_batch_root_interop.rs
@@ -31,7 +31,6 @@ impl BatchRootProcessorInterop {
         l2_chain_id: L2ChainId,
         sl_l2_client: Arc<dyn ZkSyncExtentionEthClient>,
     ) -> Self {
-        println!("BatchRootProcessorInterop::new SNEED");
         Self {
             appended_chain_batch_root_signature: ethabi::long_signature(
                 "AppendedChainBatchRoot",
--- a/core/node/eth_watch/src/event_processors/appended_chain_batch_root_interop.rs
+++ b/core/node/eth_watch/src/event_processors/appended_chain_batch_root_interop.rs
@@ -1,0 +1,175 @@
+use std::sync::Arc;
+
+use anyhow::Context;
+use itertools::Itertools;
+use zksync_dal::{eth_watcher_dal::EventType, Connection, Core, CoreDal, DalError};
+use zksync_mini_merkle_tree::MiniMerkleTree;
+use zksync_types::{
+    api::Log, ethabi, h256_to_u256, l2_to_l1_log::BatchAndChainMerklePath, L1BatchNumber,
+    L2BlockNumber, L2ChainId, H256,
+};
+
+use crate::{
+    client::ZkSyncExtentionEthClient,
+    event_processors::{BatchRootProcessor, EventProcessor, EventProcessorError, EventsSource},
+};
+
+/// Listens to `AppendedChainBatchRoot` events and saves `BatchAndChainMerklePath` for batches.
+/// These events are emitted on SL each time L1 batch is executed. Processor uses them to track which batches are already executed
+/// and group them by SL's block number they are executed in as this data is required to build `BatchAndChainMerklePath` for interop proofs.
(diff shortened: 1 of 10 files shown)


core/node/eth_watch/src/event_processors/interop_chain_batch_root.rs
```diff
@@ -0,0 +1,155 @@
...
+
+    fn event_source(&self) -> EventsSource {
+        EventsSource::SL
+    }
+
+    fn event_type(&self) -> EventType {
+        EventType::ChainBatchRoot
+    }
+
+    fn only_finalized_block(&self) -> bool {
```
kelemeno: we want this to be not finalized, right? I'm unsure atm ( did not think it fully through)

Deniallugo: why do we even need this?

kelemeno: We need to set save two distinct merkle paths for interop vs withdrawals.

core/node/eth_watch/src/event_processors/appended_chain_batch_root_interop.rs
```diff
@@ -0,0 +1,179 @@
...
+                let mut proof = vec![H256::from_low_u64_be(leaf_position as u64)];
+                proof.extend(batch_proof);
+
+                BatchAndChainMerklePath {
+                    batch_proof_len,
+                    proof,
+                }
+            };
+
+            let chain_agg_proof_until_msg_root = self
```
Deniallugo: It'd fail the server,  if there is an error here. 
Is it a guarantee that `get_chain_log_proof_until_msg_root` won't fail? is it a gateway sl client?

core/node/eth_watch/src/event_processors/appended_chain_batch_root_interop.rs
```diff
@@ -0,0 +1,179 @@
...
+            // Update the tree with the new batches first
+            // Else we may store an incorrect batch proof if the chain happens to execute again in the same block
+            for (batch_number, logs_root_hash) in &chain_batches {
+                let root_from_db = transaction
+                    .blocks_dal()
+                    .get_l1_batch_l2_l1_merkle_root(*batch_number)
+                    .await
+                    .map_err(DalError::generalize)?
+                    .context("Missing l2_l1_merkle_root for finalized batch")?;
+                assert_eq!(root_from_db, *logs_root_hash);
```
Deniallugo, line 99: Why do we need to compare it? could it be wrong?

core/node/eth_watch/src/event_processors/appended_chain_batch_root_interop.rs
Deniallugo, line 1: I spoke with @perekopskiy and seems my concerned are not really valid, but. do we really need a new event processor? if it's almost copypasted from the BatchRootProcessor?

core/node/eth_watch/src/event_processors/appended_chain_batch_root_interop.rs
```diff
@@ -0,0 +1,175 @@
...
+    sl_l2_client: Arc<dyn ZkSyncExtentionEthClient>,
+}
+
+impl BatchRootProcessorInterop {
+    pub fn new(
+        merkle_tree: MiniMerkleTree<[u8; 96]>,
+        l2_chain_id: L2ChainId,
+        sl_l2_client: Arc<dyn ZkSyncExtentionEthClient>,
+    ) -> Self {
+        println!("BatchRootProcessorInterop::new SNEED");
```
Deniallugo: seems debug info

zkzoomer: indeed that shouldn't be there

core/node/eth_watch/src/event_processors/appended_chain_batch_root_interop.rs
zkzoomer: Correct, these could potentially be merged but it would lead to a very complex implementation, as the BatchRootProcessor operates through GW batches, while the BatchRootProcessorInterop operates through GW blocks

PR #4215, opened 2025-06-20
test(integration): upgrade zksync-ethers

itegulov:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

- [ ] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [ ] Tests for the changes have been added / updated.
- [ ] Documentation comments have been added / updated.
- [ ] Code has been formatted via `zkstack dev fmt` and `zkstack dev lint`.

PR #4225, opened 2025-06-23
feat(api): stabilize zks_gasPerPubdata

itegulov:
## What ❔

This PR:
* Stabilizes `unstable_gasPerPubdata` as `zks_gasPerPubdata`
* Makes it never return less than `1` gas per pubdata as such transactions are considered invalid
* Upgrades zksync-ethers to use a [WIP branch](https://github.com/zksync-sdk/zksync-ethers/pull/252) that stops calling `zks_estimateFee` in favour of `zks_gasPerPubdata` (if available)

## Why ❔

`zks_estimateFee` is scheduled for deprecation and this is the last missing piece

## Is this a breaking change?
- [x] Yes
- [ ] No

## Operational changes

This is technically a breaking change but realistically `unstable_gasPerPubdata` is unused by the public

## Checklist

### DIFF
--- a/core/tests/gateway-migration-test/package.json
+++ b/core/tests/gateway-migration-test/package.json
@@ -26,7 +26,7 @@
     "node-fetch": "^2.6.1",
     "ts-node": "^10.1.0",
     "typescript": "^4.3.5",
-    "zksync-ethers": "^6.18.0"
+    "zksync-ethers": "https://github.com/zksync-sdk/zksync-ethers#di/avoid-zks-estimate-fee-build"
   },
   "dependencies": {
     "prettier": "^2.3.2"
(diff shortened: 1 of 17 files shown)


core/tests/gateway-migration-test/package.json
```diff
@@ -26,7 +26,7 @@
     "node-fetch": "^2.6.1",
     "ts-node": "^10.1.0",
     "typescript": "^4.3.5",
-    "zksync-ethers": "^6.18.0"
+    "zksync-ethers": "https://github.com/zksync-sdk/zksync-ethers#di/avoid-zks-estimate-fee-build"
```
popzxc, line 29: I guess this should be fixed before PR is merged

itegulov: Well, we were using a random branch before I upgraded zksync-ethers in https://github.com/matter-labs/zksync-era/pull/4215 so I don't think this is a huge deal. That being said I can also move changes in integration tests to another PR and we can merge just stabilization first. WDYT? I would rather get this to all chains as soon as we can.

popzxc: Ugh, I think it's fine -- but let's try not to forget to update it once new zksync-ethers is released.

popzxc (approved):

PR #4226, opened 2025-06-24
feat(api): Support Unix domain sockets for healthcheck server

slowli:
## What ❔

Enables serving healthcheck info via a Unix domain socket (UDS).

## Why ❔

This is beneficial for integration testing to avoid specifying healthcheck URLs / ports via env vars. May have use in production as well (healthcheck data is private, although it's not sensitive).

## Is this a breaking change?

- [ ] Yes
- [x] No

## Operational changes

No mandatory changes. The `api.healthcheck.port` config param now supports full socket addresses (e.g., `127.0.0.1:3070` or `[::]:3070`) and socket paths like `ipc:/path/to/socket.sock`.

## Checklist


core/Cargo.toml
```diff
@@ -156,11 +157,11 @@ num_cpus = "1.13"
...
-opentelemetry = "0.24.0"
-opentelemetry_sdk = "0.24.0"
-opentelemetry-otlp = "0.17.0"
-opentelemetry-semantic-conventions = "0.16.0"
-opentelemetry-appender-tracing = "0.5"
+opentelemetry = "0.30.0"
+opentelemetry_sdk = "0.30.0"
+opentelemetry-otlp = { version = "0.30.0", default-features = false }
+opentelemetry-semantic-conventions = "0.30.0"
+opentelemetry-appender-tracing = "0.30.0"
```
slowli, line 164: Motivated by `axum` upgrade (these crates use `axum` via `tonic`). I've checked the related changes locally, but not 100% sure they work in all cases. So they can be extracted to a separate PR if desired.

core/lib/config/src/configs/api.rs
```diff
@@ -46,6 +46,107 @@ impl ApiConfig {
...
+/// Supports any of 3 formats:
+///
+/// - Just a `u16` port. This will bind a server to all IPv4 interfaces (i.e., `0.0.0.0`) for backward compatibility.
+/// - Full socket address (e.g., `127.0.0.1:8080`).
+/// - (For Unix systems) Path to a Unix domain socket (UDS) prefixed by `ipc:` (e.g., `ipc:chains/era/health.sock` or `ipc:/var/zksync.sock`).
+///   If the path is relative, it will resolve relative to the current working directory.
+#[derive(Debug, Clone, PartialEq)]
+pub enum BindAddress {
+    Tcp(SocketAddr),
+    #[cfg(unix)]
```
slowli, line 60: Does this sort of conditional compilation make sense, or do we not care about non-Unix targets at this point?

core/lib/config/src/configs/api.rs
```diff
@@ -46,6 +46,107 @@ impl ApiConfig {
     }
 }
 
+/// Port binding specification.
+///
+/// Supports any of 3 formats:
+///
+/// - Just a `u16` port. This will bind a server to all IPv4 interfaces (i.e., `0.0.0.0`) for backward compatibility.
+/// - Full socket address (e.g., `127.0.0.1:8080`).
+/// - (For Unix systems) Path to a Unix domain socket (UDS) prefixed by `ipc:` (e.g., `ipc:chains/era/health.sock` or `ipc:/var/zksync.sock`).
```
slowli: Bikeshedding on syntax appreciated. AFAIU, there's no standardized spec, but some other apps use `ipc://` prefix (i.e., `ipc://./chains/era/health.sock` or `ipc:///var/zksync.sock`).

core/Cargo.toml
```diff
@@ -116,7 +116,7 @@ aws-config = { version = "1.1.7", default-features = false, features = [
 ] }
 aws-runtime = "1.5.5"
 aws-sdk-s3 = "1.76.0"
-axum = "0.7.5"
+axum = "0.8.4"
```
slowli, line 119: New `axum` supports UDS out of the box, v0.7 doesn't.

core/node/external_proof_integration_api/src/types.rs
```diff
@@ -85,7 +85,6 @@ impl ExternalProof {
     }
 }
 
-#[async_trait::async_trait]
```
perekopskiy, line 88: Can you please explain why annotation is not required here?

perekopskiy (dismissed):

core/lib/config/src/configs/api.rs
perekopskiy: IMO let's keep it since it's pretty isolated and doesn't add almost any code

core/node/external_proof_integration_api/src/types.rs
slowli: New Rust versions support async traits natively, albeit with significant limitations (e.g., these traits are not dyn-compatible, which we extensively use – so it's not like we should throw away `async_trait` immediately). `axum` v0.8 uses this new feature, and it's not compatible with the code produced by `async_trait` transforms. (In order to keep traits dyn-compatible, `async_trait` boxes futures output by async methods in the trait.)

core/lib/config/src/configs/api.rs
itegulov: JFYI I am pretty sure modern Windows supports Unix sockets (both in WSL and natively). Seems like tokio just [hasn't been updated](https://github.com/tokio-rs/mio/issues/1609) to support that yet. For now it seems like there is [`interprocess`](https://docs.rs/interprocess/latest/interprocess/) crate that provides a unified interface for UNIX and Windows.

Not saying we care that much though. I would be mildly surprised if our stack works on Windows even in its current state.

core/lib/config/src/configs/api.rs
itegulov: `reth` and `anvil` use `/tmp/{reth,anvil}.ipc` by default. That being said they provide full JSON RPC API through that socket. Tangential to this PR but maybe we should do the same for the sake of minimal latency in integration tests? `reth` has a `jsonrpsee` middleware that we can likely adapt for our needs.

As for the name I like the `ipc://` prefix (btw the rustdoc uses `ipc:` instead, not sure if its intentional). I think its reasonable to assume that there is one server per machine just like we assume default port without taking chain into account. So I would use one of these: `ipc:///run/zksync.sock`, `ipc:///var/run/zksync.sock`, `ipc:///tmp/zksync.sock`. Not sure if there is a difference between 1 and 2 in modern Linux, I have seen both conventions being used.

itegulov (approved):

core/lib/config/src/configs/api.rs
slowli: > maybe we should do the same for the sake of minimal latency in integration tests?

I have a separate task for that. I've already sketched an implementation; it seems to work, so will publish it separately. Besides latency, the use of UDS circumvents the issue of assigning separate ports for each chain. Even if some tooling for L2 doesn't support UDS, we can run it sequentially for each chain w/ TCP sockets configured, and then run integration tests etc. in parallel with UDS. The only wrinkles breaking this idyllic picture seems to be node interactions (EN–main node and main node–gateway), but adding client-side UDS support in *our* code is straightforward (I've done this as a part of the aforementioned implementation for tests anyway 🙃).

itegulov (approved):

PR #4228, opened 2025-06-24
feat: add `pubdata_limit` as batch parameter

perekopskiy:
## What ❔

Adds `pubdata_limit` as batch parameter. It's sourced from config and saved in DB for each batch as a part of `UnsealedBatchHeader`. It's added to the consensus payload which is why it's only set for >= v29.

## Why ❔

Prerequisite for leader rotation: it's going to be used in seal criterion instead of `config.max_pubdata_per_batch` so non-leader nodes can verify that the batch fits into the limit. Currently, it's unused; seal criteria will be modified separately.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Operational changes
None

## Checklist


core/lib/dal/src/consensus_dal/mod.rs
```diff
@@ -507,11 +510,34 @@ impl ConsensusDal<'_, '_> {
...
+            .iter()
+            .map(|b| b.l1_batch_number)
+            .unique()
+            .collect::<Vec<_>>();
+        let mut pubdata_limits = HashMap::new();
+        for l1_batch_number in l1_batch_numbers {
+            let pubdata_limit = self
+                .storage
+                .blocks_dal()
+                .get_common_l1_batch_header(l1_batch_number)
```
slowli: This (as opposed to getting just `pubdata_limit`s for a range of L1 batch numbers in a single dedicated query) looks somewhat wasteful. AFAIU, we do always have a range of L1 batch numbers, right?

core/lib/dal/src/sync_dal.rs
```diff
@@ -103,7 +106,18 @@ impl SyncDal<'_, '_> {
         } else {
             None
         };
-        Ok(Some(block.into_api(transactions)))
+
+        let pubdata_limit = self
```
slowli: Dumb question: Why not load the `pubdata_limit` as a part of the query in `sync_blocks_inner()`? Would joining `l1_batches` harm query performance?

core/lib/vm_executor/src/storage.rs
```diff
@@ -263,22 +271,28 @@ impl L1BatchParamsProvider {
...
-        // L1 batch timestamp is set to the timestamp of its first L2 block.
-        let l1_batch_timestamp = first_l2_block_in_batch.header.timestamp;
+
+        let l1_batch_header = conn
+            .blocks_dal()
+            .get_common_l1_batch_header(first_l2_block_in_batch.l1_batch_number)
+            .await
+            .map_err(DalError::generalize)?
+            .context("pending batch is missing in DB")?;
+        let l1_batch_timestamp = l1_batch_header.timestamp;
```
slowli, line 292: Dumb question: Is the old approach ("L1 batch timestamp is set to the timestamp of its first L2 block") no longer true? Without it, `l1_batch_header` seems to only be used for `pubdata_limit`, which is also the case for some internal DAL uses of `get_common_l1_batch_header()`, so it could be refactored to return just it (see the other line comments).

core/node/state_keeper/src/io/mempool.rs
```diff
@@ -164,7 +164,10 @@ impl StateKeeperIO for MempoolIO {
                 pending_batch_data
                     .l1_batch_env
                     .clone()
-                    .into_unsealed_header(Some(pending_batch_data.system_env.version)),
+                    .into_unsealed_header(
+                        Some(pending_batch_data.system_env.version),
+                        pending_batch_data.pubdata_limit,
+                    ),
```
slowli, line 170: Dumb question: Is it fine that pubdata limit is not enforced during VM execution (e.g., in the bootloader), hence it's not a part of `L1BatchEnv`?

core/lib/dal/src/sync_dal.rs
perekopskiy: I had problems with SQL in `sync_blocks_inner`, but it's resolved now. I don't think it will harm performance

core/lib/dal/src/consensus_dal/mod.rs
perekopskiy: Removed queries completely as know pubdata_limit is a part of SyncBlock. Answering the original question, the only caller of `block_payloads` does `.block_payloads(number..number + 1)` so it's actually one L1 batch always. Anyway, doesn't matter now

core/lib/vm_executor/src/storage.rs
perekopskiy: The approach is true. However, I would formulate it a bit differently "First L2 block timestamp is set to the timestamp of the L1 batch". I've changed this place so the code gets more general and doesn't rely on the approach. As for replacing `get_common_l1_batch_header` with `get_pubdata_limit`, I think it's not a big deal to load whole CommonL1BatchHeader, it's just several integers. And in case there are other usages for `get_common_l1_batch_header` in the future we won't have both `get_common_l1_batch_header` and `get_pubdata_limit` in the code.

core/node/state_keeper/src/io/mempool.rs
perekopskiy: IMO it's fine, check in seal criterion is enough (I will change it in a separate PR)

core/lib/dal/src/consensus/proto/mod.proto
```diff
@@ -34,6 +34,7 @@ message Payload {
   repeated TransactionV25 transactions_v25 = 12;
   optional bool last_in_batch = 10; // required
   optional PubdataParams pubdata_params = 13; // optional
+  optional uint64 pubdata_limit = 14; // optional
```
brunoffranca: should say required since protocol version 29.

core/lib/vm_executor/src/storage.rs
```diff
@@ -42,6 +42,14 @@ impl FirstL2BlockInBatch {
     }
 }
 
+#[derive(Debug)]
+pub struct RestoredL1BatchEnv {
```
RomanBrodetski, line 48: please add a detailed Doc Comment

RomanBrodetski (dismissed):

RomanBrodetski (approved):

PR #4232, opened 2025-06-24
feat(eigenda): EigenDA V2 M1

gianbelinche:
## What ❔
[V2M0 PR](https://github.com/matter-labs/zksync-era/pull/3983) should be merged first

This PR adds EigenDA V2 M1 client.

It nows sends the blob key of the dispersed blob to the EigenDA Sidecar for it to verify it.

Then on get inclusion data, we query the sidecar for the proof, if it is still generating it, it will return None.

New contracts are added to verify this proofs. `EigenDAL1DAValidator` and `EigenDAL2DAValidator`

Related to: https://github.com/matter-labs/era-contracts/pull/1603

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


core/lib/basic_types/src/commitment.rs
```diff
@@ -24,6 +24,7 @@ impl From<PubdataType> for L1BatchCommitmentMode {
             | PubdataType::Avail
             | PubdataType::Celestia
             | PubdataType::Eigen
+            | PubdataType::EigenDAV2Secure
```
dimazhornyk: I don't think that we need to introduce a separate PubdataType for it
e.g. in current implementation Avail can be used with and without DA verification and that is only defined by the `use_dummy_inclusion` flag

core/lib/config/src/configs/da_client/eigen.rs
```diff
@@ -1,46 +1,98 @@
...
+}
+
+impl WellKnown for PolynomialForm {
+    type Deserializer = Serde![str];
+    const DE: Self::Deserializer = Serde![str];
+}
+
+/// The EigenDA client has two versions: V2 and V2Secure.
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub enum Version {
```
dimazhornyk: if the only purpose of this is to define if we have to call the sidecar - we can remove it for simplicity
the sidecar url can be optional, and if it is - we don't send any requests there

core/node/da_clients/src/eigen/client.rs
```diff
@@ -1,70 +1,150 @@
...
+    secure: bool,
 }
 
 impl EigenDAClient {
-    pub async fn new(
-        config: EigenConfig,
-        secrets: EigenSecrets,
-        blob_provider: Arc<dyn BlobProvider>,
-    ) -> anyhow::Result<Self> {
+    pub async fn new(config: EigenConfig, secrets: EigenSecrets) -> anyhow::Result<Self> {
```
dimazhornyk: we can do it as a follow-up, but I assume some kind of token needs to be used with a sidecar
if we intend to share it across different networks we need to be able to control the access to the sidecar and track its usage

core/node/da_clients/src/eigen/client.rs
```diff
@@ -104,15 +202,44 @@ impl DataAvailabilityClient for EigenDAClient {
...
-            .get_inclusion_data(blob_id)
+            .get_cert(&blob_key)
             .await
             .map_err(to_retriable_da_error)?;
-        if let Some(inclusion_data) = inclusion_data {
-            Ok(Some(InclusionData {
-                data: inclusion_data,
-            }))
+        if let Some(eigenda_cert) = eigenda_cert {
+            if self.secure {
```
dimazhornyk: this check is unnecessary: if M0 is used - it is meant to be used with the `use_dummy_inclusion_data`
if M1 is used - it is assumed that a sidecar is always used running

I would only suggest to check in `node_builder.rs` that if the sidecar url is None - `use_dummy_inclusion_data` has to be true, otherwise - panic

zkstack_cli/crates/config/src/contracts.rs
```diff
@@ -114,6 +114,16 @@ impl ContractsConfig {
                 .deployed_addresses
                 .avail_l1_da_validator_addr,
         );
+        self.l1.eigenda_l1_validator_addr = Some(
+            deploy_l1_output
+                .deployed_addresses
+                .eigenda_l1_validator_addr,
+        );
+        self.l1.eigenda_risc_zero_verifier_addr = Some(
```
dimazhornyk: seems like risc_zero_verifier_addr is only needed as a constructor argument for the eigenda_l1_validator
why do we need to store it in the configs?

core/node/da_clients/src/eigen/client.rs
```diff
@@ -75,26 +155,44 @@ impl DataAvailabilityClient for EigenDAClient {
...
             .client
-            .dispatch_blob(data)
+            .send_payload(payload)
             .await
             .map_err(to_retriable_da_error)?;
 
-        Ok(DispatchResponse::from(blob_id))
+        if self.secure {
+            // In V2Secure, we need to send the blob key to the sidecar for proof generation
+            self.send_blob_key(blob_key.to_hex())
```
dimazhornyk: you can make this call in `ensure_finality` so that when the blob is not finalized, the sidecar is not overloaded with useless requests (unless you handle it on sidecar side)

core/node/da_clients/src/eigen/client.rs
samlaf: @juanbono can you comment on this, curious what you guys were thinking here

core/lib/config/src/configs/da_client/eigen.rs
```diff
@@ -1,46 +1,73 @@
...
-    /// Custom quorum numbers
-    #[config(default, with = Delimited(","))]
-    pub custom_quorum_numbers: Vec<u8>,
+    /// Address of the EigenDA cert verifier router
+    pub cert_verifier_router_addr: String,
+    /// Blob version
+    pub blob_version: u16,
+    /// Polynomial form to disperse the blobs
+    #[serde(default)]
+    pub polynomial_form: PolynomialForm,
```
samlaf: Do we need this in the config? I'm assuming you guys are using coeff form (same as eigenda disperser expects). We created the PolynomialForm distinction because optimistic rollups need to be able to do point openings for interactive fraud proofs. But afaiu its not needed for zk rollups like zksync, so we should probably just get rid of this configuration and always use CoeffForm?

core/lib/config/src/configs/da_client/eigen.rs
```diff
@@ -1,46 +1,73 @@
...
+    /// Address of the operator state retriever
+    pub operator_state_retriever_addr: String,
+    /// Address of the registry coordinator
+    pub registry_coordinator_addr: String,
+    /// URL of the EigenDA Sidecar RPC server
+    /// This is used for EigenDA V2 Secure integration,
+    /// so if its either `None` or `Some` defines whether we are using EigenDA V2 Secure or not.
+    pub eigenda_sidecar_rpc: Option<String>,
```
samlaf: Is the "sidecar" still a sidecar? Or is it fully shareable now? Should probably not call it a sidecar if its not a sidecar. @dimazhornyk what do celestia and avail call their prover server? Maybe just `eigenda_prover_service_rpc`?

core/node/da_clients/src/eigen/EIGENDAV2_MIGRATION.md
```diff
@@ -0,0 +1,48 @@
...
+
+You need to update the client config to the new format.
+
+`chains/<YOUR_CHAIN>/configs/general.yaml`
+
+```yaml
+da_client:
+  client: Eigen
+  disperser_rpc: https://disperser-testnet-holesky.eigenda.xyz
+  eigenda_eth_rpc: https://ethereum-holesky-rpc.publicnode.com
```
samlaf: can you add docstring pointing to https://docs.eigencloud.xyz/products/eigenda/networks/mainnet so people can find the addresses for mainnet and holesky. Also @dimazhornyk is the zksync testnet on holesky and not sepolia?

core/node/da_clients/src/eigen/EIGENDAV2_MIGRATION.md
```diff
@@ -0,0 +1,48 @@
...
+  authenticated: true
```
samlaf: I don't think this is needed. This is only for v1, and v1 should always be authenticated anyways right?

core/node/da_clients/src/eigen/EIGENDAV2_MIGRATION.md
```diff
@@ -0,0 +1,48 @@
...
+
+> Note: There is a missing field, `eigenda_sidecar_rpc`, this determines wheter the client will use secure mode or not.
+
+Check the [README.md](./README.md) for more details on the new fields.
+
+### Note
+
+- Remember to run `zkstackup --local` before running the new server after this changes
+
+- Be sure that your private key has the needed permissions set on the V2 client.
```
samlaf: Is the private_key shared between v1 and v2? For v2 we should mention that some form of payment needs to have been performed onchain, and link to https://docs.eigencloud.xyz/products/eigenda/core-concepts/payments

core/node/da_clients/src/eigen/README.md
```diff
@@ -6,54 +6,89 @@ For more information you can check the [docs](https://docs.eigenda.xyz/).
 
 ## Client configuration
 
-First you need to set the `use_dummy_inclusion_data` field in the file `etc/env/file_based/general.yaml` to `true`. This
-is a pending solution until the necessary contract changes are done (M1 milestone).
+To set up an eigenda client, you need to modify `etc/env/file_based/overrides/validium.yaml`:
+
+First you need to set the `use_dummy_inclusion_data` field to `true`. This is a pending solution until the necessary
+contract changes are done (M1 milestone).
```
samlaf: this is outdated?

core/node/da_clients/src/eigen/README.md
```diff
@@ -6,54 +6,89 @@ For more information you can check the [docs](https://docs.eigenda.xyz/).
...
 
 - `disperser_rpc` (string): URL of the EigenDA Disperser RPC server.
-- `settlement_layer_confirmation_depth` (unsigned number): Block height needed to reach in order to consider the blob
-  finalized. A value less or equal to 0 means that the disperser will not wait for finalization.
 - `eigenda_eth_rpc` (optional string): URL of the Ethereum RPC server. If the value is not set, the client will use the
   same rpc as the rest of the zk server.
-- `eigenda_svc_manager_address` (string): Address of the service manager contract.
-- `wait_for_finalization` (boolean): Wait for the blob to be finalized before returning the response.
 - `authenticated` (boolean): Authenticated dispersal. If true, the client will use the authentication mechanism, using a
   whitelisted account. Using non authenticated dispersal is not recommended, as to many requests to the EigenDA
```
samlaf: prob want to get rid of this and always set to true. Otherwise mention that this is only needed for v1.

core/lib/config/src/configs/da_client/eigen.rs
samlaf: Oh wait seems like its the opposite? Seems like the code iiuc is hardcoded to use evaluation form? https://github.com/lambdaclass/zksync-eigenda-m1/pull/24/files?short_path=6288e01#diff-bb727b8d363b7cbca6a393c8df2efa93d03cc3de5e3f38e6ce7ad059033c9d83R29

Or does the code actually support either form? How should a user decide which form to use?

gianbelinche: I removed the field from the config, and Coeff form is now always used

core/lib/config/src/configs/da_client/eigen.rs
gianbelinche: Done
About the fully sharable, it could be, but would not be ideal.
Currently the sidecar just receives blobs to prove, and it does not know who send them.
Also due to how risc0 works, the blob proofs are done sequentialy.
So it could happen in an scenario where multiple chains are using it, if one chain sends blobs a lot faster than the others, it could make the other chains wait a lot of time for their proofs.
In the future the sidecar could be modified to better suite multiple chains, like having some kind of scheduling for the different chains.

core/node/da_clients/src/eigen/EIGENDAV2_MIGRATION.md
gianbelinche: Done

core/node/da_clients/src/eigen/EIGENDAV2_MIGRATION.md
gianbelinche: I removed it and set it always to true

core/node/da_clients/src/eigen/EIGENDAV2_MIGRATION.md
gianbelinche: We meant that they should either use the same one or change it for one with payments on V2.
Technically someone could use the same private key for both, altough it probably isn't ideal.
I added the link

core/node/da_clients/src/eigen/README.md
gianbelinche: Yes, fixed

core/node/da_clients/src/eigen/README.md
gianbelinche: Done

zkstack_cli/crates/config/src/chain.rs
```diff
@@ -73,6 +73,8 @@ pub enum DAValidatorType {
     Rollup = 0,
     NoDA = 1,
     Avail = 2,
+    EigenDA = 3,
+    EigenDAV2Secure = 4,
```
dimazhornyk: this probably needs to be deleted

zkstack_cli/crates/config/src/general.rs
```diff
@@ -255,6 +255,11 @@ impl GeneralConfigPatch {
         Ok(())
     }
 
+    pub fn set_eigendav2secure_client(&mut self) -> anyhow::Result<()> {
+        self.0.insert("da_client.client", "EigenDAV2Secure")?;
```
dimazhornyk: does it have to be EigenDAV2Secure?

zkstack_cli/crates/zkstack/src/commands/chain/args/init/da_configs.rs
```diff
@@ -29,6 +29,7 @@ pub enum ValidiumTypeInternal {
     NoDA,
     Avail,
     EigenDA,
+    EigenDAV2Secure,
```
dimazhornyk: same here and 10 lines below

zkstack_cli/crates/zkstack/src/commands/chain/init/mod.rs
```diff
@@ -258,7 +258,8 @@ pub(crate) async fn get_l1_da_validator(chain_config: &ChainConfig) -> anyhow::R
             match general_config.da_client_type().as_deref() {
                 Some("Avail") => contracts_config.l1.avail_l1_da_validator_addr,
                 Some("NoDA") | None => contracts_config.l1.no_da_validium_l1_validator_addr,
-                Some("Eigen") => contracts_config.l1.no_da_validium_l1_validator_addr, // TODO: change for eigenda l1 validator for M1
+                Some("Eigen") => contracts_config.l1.no_da_validium_l1_validator_addr,
+                Some("EigenDAV2Secure") => contracts_config.l1.eigenda_l1_validator_addr,
```
dimazhornyk: and here

dimazhornyk (commented):
could you please pull from main so we only see the relevant diff without M0V2 code

zkstack_cli/crates/config/src/chain.rs
juanbono: should we just use the `EigenDA` for secure (and remove EigenDAV2Secure)?

juanbono: This is the only way we found so that the user can select which DAValidator contracts to use. Is there another way?

dimazhornyk: you can do it in the same way it is implemented for Avail: if Eigen is selected - it deploys the dummy EigenDA validators

if the `use_dummy_inclusion_data`  is going to be used - we expect the chain to be created as NoDA, and then the config to be modified

we would probably refactor it all a bit later, but let's keep it like that for now

core/node/da_clients/src/eigen/client.rs
```diff
@@ -87,6 +173,15 @@ impl DataAvailabilityClient for EigenDAClient {
                 "Failed to convert bytes to a 32-byte array"
             ))
         })?);
+
+        // Prover Service RPC being set means we are using EigenDA V2 Secure
+        if self.eigenda_prover_service_rpc.is_some() {
```
dimazhornyk: if the `.get_cert()` below fails - we would be calling `send_blob_key` multiple times
I don't know if you have any deduplication on the prover service side, but putting `self.send_blob_key` into `if eigenda_cert.is_some() {` will help solve the problem on this side

core/node/da_clients/src/eigen/client.rs
```diff
@@ -120,14 +215,27 @@ impl DataAvailabilityClient for EigenDAClient {
...
-            let inclusion_data = eigenda_cert.to_bytes().map_err(|_| {
-                to_non_retriable_da_error(anyhow::anyhow!(
-                    "Failed to convert eigenda cert to bytes"
-                ))
-            })?;
-            Ok(Some(InclusionData {
-                data: inclusion_data,
-            }))
+            // Prover Service RPC being set means we are using EigenDA V2 Secure
+            if self.eigenda_prover_service_rpc.is_some() {
```
dimazhornyk: nit : if `get_inclusion_data` - it  by default means that `use_dumy_inclusion_data` is false, so you can remove the `if self.eigenda_prover_service_rpc.is_some() {` check

core/node/da_clients/src/eigen/client.rs
```diff
@@ -120,14 +215,27 @@ impl DataAvailabilityClient for EigenDAClient {
...
+                if let Some(proof) = self
+                    .get_proof(blob_id)
+                    .await
+                    .map_err(to_non_retriable_da_error)?
+                {
+                    Ok(Some(InclusionData { data: proof }))
```
dimazhornyk: just to be sure: is `proof` here the abi-encoded `EigenDAInclusionData`?

gianbelinche: Yes it is
```
struct EigenDAInclusionData {
    bytes seal;
    bytes32 imageId;
    bytes journal;
}
```

abi-encoded

core/node/da_clients/src/eigen/client.rs
gianbelinche: [Done](https://github.com/matter-labs/zksync-era/pull/4232/commits/aeffa52e0c0bd0e207dacd0f47211071956ded1b)

core/node/da_clients/src/eigen/client.rs
gianbelinche: [Done](https://github.com/matter-labs/zksync-era/pull/4232/commits/aeffa52e0c0bd0e207dacd0f47211071956ded1b)

deniallugo-ml:

PR #4235, opened 2025-06-25
feat(api): Support combined HTTP / WS server

slowli:
## What ❔

- Supports combined HTTP / WS server bound to a single port.
- Allows configuring RPC namespaces for WS.

## Why ❔

## Is this a breaking change?

- [ ] Yes
- [x] No

## Operational changes

- HTTP / WS server is combined automatically if `api.web3_json_rpc.{http_port, ws_port}` are equal.
- RPC namespaces for WS are represented via new `api.web3_json_rpc.ws_api_namespaces` param. `api.web3_json_rpc.api_namespaces` is still used if the new param is not set.

## Checklist

### DIFF
--- a/core/node/api_server/src/web3/metrics.rs
+++ b/core/node/api_server/src/web3/metrics.rs
@@ -500,6 +500,16 @@ pub(super) struct TxReceiptMetrics {
 #[vise::register]
 pub(super) static TX_RECEIPT_METRICS: vise::Global<TxReceiptMetrics> = vise::Global::new();
 
+#[derive(Debug, Metrics)]
+#[metrics(prefix = "api_jsonrpc_backend_batch")] // FIXME: prefix is bogus; there's no `jsonrpc_backend`
+pub(super) struct LimitMiddlewareMetrics {
+    /// Number of rate-limited requests.
+    pub rate_limited: Family<ApiTransportLabel, Counter>,
+}
+
+#[vise::register]
+pub(super) static LIMIT_METRICS: vise::Global<LimitMiddlewareMetrics> = vise::Global::new();
+
 #[cfg(test)]
 mod tests {
     use assert_matches::assert_matches;
--- a/core/node/api_server/src/web3/metrics.rs
+++ b/core/node/api_server/src/web3/metrics.rs
@@ -10,7 +10,7 @@ use zksync_instrument::filter::{report_filter, ReportFilter};
 use zksync_types::api;
 use zksync_web3_decl::error::Web3Error;
 
-use super::{backend_jsonrpsee::MethodMetadata, ApiTransport, TypedFilter};
+use super::{backend_jsonrpsee::MethodMetadata, TypedFilter};
 use crate::tx_sender::SubmitTxError;
 
 /// Observed version of RPC parameters. Have a bounded upper-limit size (256 bytes), so that we don't over-allocate.
@@ -82,15 +82,6 @@ impl<'a> ObservedRpcParams<'a> {
 pub(crate) enum ApiTransportLabel {
     Http,
     Ws,
-}
-
-impl From<&ApiTransport> for ApiTransportLabel {
-    fn from(transport: &ApiTransport) -> Self {
-        match transport {
-            ApiTransport::Http(_) => Self::Http,
-            ApiTransport::WebSocket(_) => Self::Ws,
-        }
-    }
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, EncodeLabelValue)]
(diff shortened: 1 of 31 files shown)


core/node/api_server/src/web3/mod.rs
```diff
@@ -415,26 +508,6 @@ impl ApiServer {
...
-            (ApiTransport::WebSocket(_), None) => {
-                tracing::warn!(
-                    "`subscriptions_limit` is not set - unlimited subscriptions are allowed"
-                );
-            }
-            (ApiTransport::Http(_), Some(_)) => {
-                tracing::warn!(
-                    "`subscriptions_limit` is ignored for HTTP transport, use WebSocket instead"
-                );
-            }
```
slowli, line 434: There's a change in `subscriptions_limit` semantics: it is now applied to the HTTP server as well. I.e., it is connection limit as supported by `jsonrpsee`. Not that it was limiting subscriptions specifically before; it still was limiting *all* connections for the WS server. If this is OK, I'll rename the param to `connection_limit` (`subscriptions_limit` would be a deprecated alias). IMO, this is the easiest and most transparent approach.

core/node/api_server/src/web3/mod.rs
```diff
@@ -616,39 +683,36 @@ impl ApiServer {
...
-                .context("Failed building WS JSON-RPC server")?;
-            (server.local_addr(), server.start(rpc))
-        };
-        let local_addr = local_addr.with_context(|| {
+        match transport {
+            ApiTransport::Http(_) => {
+                server_builder = server_builder.http_only();
+            }
+            ApiTransport::Ws(_) => {
+                // FIXME: WS server didn't have `ws_only` set; why?
```
slowli, line 701: Checked locally: On the `main` branch, the WS server indeed accepts (non-WS) HTTP requests. Is this intended?

zkstack_cli/crates/zkstack/src/utils/ports.rs
```diff
@@ -156,7 +162,60 @@ impl EcosystemPorts {
             }
             _ => {}
         }
+        Ok(())
+    }
+
+    fn use_new_ports_in_addresses(
```
slowli, line 168: This logic assumes that new ports are defined in the same file. I think this holds for now. More generally, I feel that the entire approach of copying configs is bogus / error-prone, and thankfully would be moderately easy to get rid of with the new config system. Besides the bug that I've fixed, there's an issue with lots of unused params in the EN config, and frustrating UX (say I've changed some params in the main node config to debug, and now all these changes are copied to the EN config – why?). Instead, configs should be initialized from a template (IDK, maybe literally using Handlebars or another template engine).

zkstack_cli/crates/zkstack/src/utils/ports.rs
```diff
@@ -627,6 +689,57 @@ mod tests {
...
+
+    #[test]
+    fn test_traverse_allocate_ports_in_yaml_with_duplicate_source_ports() {
+        let yaml_content = r#"
+            api:
+                web3_json_rpc:
+                    http_port: 3050
+                    http_url: http://127.0.0.1:3050
+                    ws_port: 3050
+                    ws_url: ws://127.0.0.1:3050
```
slowli, line 725: The old logic had a bug when it did assign separate ports for `http_port` / `ws_port` if they were equal in the copied config, but didn't update ports correctly in `_url` params. It looks like the old approach is fundamentally unworkable if there may be port collisions because it's impossible to tell which ports in URLs / socket addresses belong to which port param. The new approach makes these links explicit.

core/node/api_server/src/web3/metrics.rs
```diff
@@ -509,6 +500,16 @@ pub(super) struct TxReceiptMetrics {
 #[vise::register]
 pub(super) static TX_RECEIPT_METRICS: vise::Global<TxReceiptMetrics> = vise::Global::new();
 
+#[derive(Debug, Metrics)]
+#[metrics(prefix = "api_jsonrpc_backend_batch")] // FIXME: prefix is bogus; there's no `jsonrpc_backend`
```
slowli, line 504: Subj. Would it be safe to change naming, e.g. so the metric is called `api_web3_rate_limited_requests`?

popzxc: Dunno. I guess we need to check grafana. If it's used anywere, probably still safe but dashboards will need to be updated (although I would probably approach that separately).

core/node/api_server/src/web3/mod.rs
popzxc: Not sure, probably not.

popzxc (approved):

perekopskiy (approved):

deniallugo-ml:

PR #4241, opened 2025-06-26
feat: Proof manager watcher

Artemka374:
## What ❔

Basic implementation of ETH proof manager watcher part.
This PR implements basic listening(with just logging received events) for event from proof manager contracts
Parts of this PR are stubs, which should be filled in in coming changes.

## Why ❔

This is the part of proving networks integration

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/core/Cargo.lock
+++ b/core/Cargo.lock
@@ -13938,7 +13938,7 @@ dependencies = [
 
 [[package]]
 name = "zksync_eth_proof_manager"
-version = "28.8.0-non-semver-compat"
+version = "28.9.0-non-semver-compat"
 dependencies = [
  "anyhow",
  "async-trait",


core/bin/zksync_server/src/components.rs
```diff
@@ -12,6 +12,8 @@ pub(crate) enum Component {
     Tree,
     /// Merkle tree API.
     TreeApi,
+    /// Eth proof manager.
```
EmilLuta: I'd add a bit more detaisl on what it does (I.E. Component for interacting with proof manager (proof distribution contracts))

zkstack_cli/crates/config/src/contracts.rs
```diff
@@ -24,7 +24,7 @@ pub struct ContractsConfig {
     pub create2_factory_salt: H256,
     pub ecosystem_contracts: EcosystemContracts,
     pub bridges: BridgesContracts,
-    pub proving_network: ProvingNetworkContracts,
+    pub proof_manager_contracts: EthProofManagerContracts,
```
EmilLuta: Dumb question, but why do we call them EthProofManagerContracts? We actually run them in ZKSync, right?

etc/env/file_based/general.yaml
```diff
@@ -281,3 +281,8 @@ consensus:
 
 gateway_migrator:
   eth_node_poll_interval: 300 ms
+
+eth_proof_manager:
+  event_poll_interval: 10000 ms
```
EmilLuta, line 286: Do I understand correctly that we'll look for events once every 10s?

etc/env/file_based/contracts.yaml
```diff
@@ -19,7 +19,7 @@ ecosystem_contracts:
   state_transition_proxy_addr: "0x0000000000000000000000000000000000000000"
   bridgehub_proxy_addr: "0x0000000000000000000000000000000000000000"
   transparent_proxy_admin_addr: "0x0000000000000000000000000000000000000000"
-proving_network:
+proof_manager_contracts:
```
EmilLuta, line 22: Side question, but does contract even admit 0x0 as an address? if not, we may want to make this 0x1 or smth similar.

core/lib/config/src/configs/contracts/chain.rs
```diff
@@ -241,6 +241,10 @@ impl ContractsConfig {
         }
     }
 
+    pub fn eth_proof_manager_contracts(&self) -> ProofManagerContracts {
```
EmilLuta, line 244: Makes me wonder, logically, does it really belong here? ProofManagerContracts is for all chains, not a single chain.

core/node/eth_proof_manager/src/node.rs
```diff
@@ -0,0 +1,112 @@
...
+    service::StopReceiver,
+    task::{Task, TaskId},
+    wiring_layer::{WiringError, WiringLayer},
+    FromContext, IntoContext,
+};
+use zksync_object_store::ObjectStore;
+
+use crate::{client::ProofManagerClient, EthProofManager};
+
+/// Wiring layer for ethereum watcher
```
EmilLuta: Is this just copy paste, maybe? If so, let's rewrite it to something more sensible.

core/node/eth_proof_manager/src/watcher/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+                break;
+            }
+
+            let to_block = self.client.get_finalized_block().await?;
+
+            // todo: this should be changed
+            let from_block = to_block.saturating_sub(self.config.event_expiration_blocks);
+
+            tracing::info!(
+                "Getting events from block {} to block {}",
```
EmilLuta, line 59: Question: Interesting we chose to go from block to block. Shouldn't we actually look into DB to tell what's the last unproven block and go to that, instead of going from finalized block to finalized block - config?

etc/env/file_based/general.yaml
Artemka374: yes

etc/env/file_based/contracts.yaml
Artemka374: it doesn't, but the thing here is that after deployment(at least with zkstack), we are setting the correct addresses

core/node/eth_proof_manager/src/watcher/mod.rs
Artemka374: This is a stub, but this logic will be done properly in the next PR

zkstack_cli/crates/config/src/contracts.rs
Artemka374: will be renamed in next PR

etc/env/file_based/contracts.yaml
EmilLuta: I think we should still make those consistent with contract behavior. That said, can be tackled in next PR.

core/lib/config/src/configs/contracts/chain.rs
EmilLuta: I believe this question still stands, but not a blocker.

core/node/eth_proof_manager/src/watcher/mod.rs
EmilLuta: I believe this has been discussed in the other PR.

PR #4252, opened 2025-06-27
feat(db): Attach miniblock to l1 batch during the initial insertion

Deniallugo:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


core/lib/dal/src/blocks_dal.rs
```diff
@@ -2593,7 +2579,7 @@ impl BlocksDal<'_, '_> {
     /// to any batch yet).
     pub async fn pending_batch_exists(&mut self) -> DalResult<bool> {
         let count = sqlx::query_scalar!(
-            "SELECT COUNT(miniblocks.number) FROM miniblocks WHERE l1_batch_number IS NULL"
+            "SELECT COUNT(miniblocks.number) FROM miniblocks JOIN l1_batches ON miniblocks.l1_batch_number = l1_batches.number WHERE l1_batches.is_sealed = FALSE"
         )
```
perekopskiy: This will scan all rows in miniblocks

core/lib/dal/src/blocks_dal.rs
```diff
@@ -2720,22 +2706,46 @@ impl BlocksDal<'_, '_> {
...
+        .with_arg("l1_batch_number", &l1_batch_number)
+        .execute(self.storage)
+        .await?;
+        Ok(())
+    }
+
     pub async fn set_protocol_version_for_pending_l2_blocks(
         &mut self,
         id: ProtocolVersionId,
+        l1batch_number: L1BatchNumber,
```
perekopskiy: nit: name is a bit confusing now as you can call method with any l1 batch number. `set_protocol_version_for_l2_blocks_in_l1_batch` would be better. Also `l1batch_number` -> `l1_batch_number `

core/lib/dal/src/blocks_web3_dal.rs
```diff
@@ -71,6 +69,15 @@ impl BlocksWeb3Dal<'_, '_> {
 
         let block = rows.into_iter().fold(None, |prev_block, row| {
             let mut block = prev_block.unwrap_or_else(|| {
+                let (l1_batch_number, l1_batch_timestamp) =
+                    if row.l1_batch_is_sealed.unwrap_or(false) {
```
perekopskiy: nit: you can do `LEFT JOIN l1_batches ON l1_batches.number = miniblocks.l1_batch_number AND l1_batches.is_sealed = TRUE` instead of selecting `l1_batch_is_sealed`

core/lib/dal/src/blocks_web3_dal.rs
```diff
@@ -362,74 +369,28 @@ impl BlocksWeb3Dal<'_, '_> {
...
+            FROM
+                miniblocks
+            WHERE
+                l1_batch_number = $1
+            ORDER BY
+                number
+            LIMIT
+                1
+            "#,
+            i64::from(l1_batch_number.0)
```
perekopskiy, line 381: actually, you can just query timestamp from `l1_batches`, it will be equivalent

core/lib/dal/src/models/storage_block.rs
```diff
@@ -640,21 +640,12 @@ impl From<StorageL2BlockHeader> for L2BlockHeader {
 /// Information about L1 batch which a certain L2 block belongs to.
 #[derive(Debug)]
 pub struct ResolvedL1BatchForL2Block {
-    /// L1 batch which the L2 block belongs to. `None` if the L2 block is not explicitly attached
-    /// (i.e., its L1 batch is not sealed).
-    pub block_l1_batch: Option<L1BatchNumber>,
+    /// L1 batch which the L2 block belongs to. `None` if the L2 block doesn't exist yet
+    pub block_l1_batch: L1BatchNumber,
```
perekopskiy: Comment is misleading

core/node/state_keeper/src/io/mempool.rs
```diff
@@ -136,6 +136,12 @@ impl StateKeeperIO for MempoolIO {
         else {
             return Ok((cursor, None));
         };
+        // During the transition period we have to ensure that all pending miniblocks has l1 batch number
```
perekopskiy: nit: has -> have

perekopskiy (commented):
If server is run in distributed mode then code will only work properly if the process that runs state keeper is updated first. (in our terms server-v2-core should be updated first). I don't think we can afford to enforce it. IMO it makes sense to split it in 2 PRs:
- save miniblocks.l1_batch_number for pending blocks, make sure all queries work in both cases (if miniblocks.l1_batch_number is set and if it isn't). Make sure this PR is merged before v29
- update all queries to only support case if miniblocks.l1_batch_number is set after v29

core/lib/dal/src/blocks_web3_dal.rs
Deniallugo: I need l1_batch_is_sealed for returning pending batches in a way it was in api before

perekopskiy: but adding `l1_batches.is_sealed = TRUE` to the JOIN will do the job, no?

Deniallugo: hm, from the second glance yes. let's see in tests

core/lib/dal/src/blocks_dal.rs
```diff
@@ -2642,15 +2628,12 @@ impl BlocksDal<'_, '_> {
...
-            "SELECT COUNT(miniblocks.number) FROM miniblocks WHERE l1_batch_number IS NULL"
-        )
-        .instrument("pending_batch_exists")
-        .fetch_one(self.storage)
-        .await?
-        .unwrap_or(0);
+        let count = sqlx::query_scalar!("SELECT number FROM l1_batches WHERE is_sealed = FALSE")
+            .instrument("pending_batch_exists")
+            .fetch_optional(self.storage)
+            .await?;
```
perekopskiy: - let's update comment
- let's not change sematics i.e. return false if there is a batch with is_sealed = FALSE and no miniblocks
- current query still will scan whole table, this is what I suggest:
```
WITH u AS (
                SELECT number as l1_batch_number, is_sealed
                FROM l1_batches
                ORDER BY number DESC
                LIMIT 1
            )
SELECT COUNT(*) FROM miniblocks
WHERE miniblocks.l1_batch_number = u.l1_batch_number AND u.is_sealed = FALSE
```

perekopskiy (commented):
Not required for this PR, but I would also change `get_l2_blocks_to_reexecute`. We can query unsealed l1 batch number, its miniblocks and simply do
```
SELECT * FROM transactions WHERE miniblock_number BETWEEN $1 AND $2
ORDER BY miniblock_number, index_in_block
```

core/lib/dal/src/blocks_dal.rs
Deniallugo: i need a number anyway. so i'd rather keep it as is

perekopskiy: WDYM by "i need a number anyway"? I said that you changed the semantics of the method, it's not acceptable, callers rely on the fact that method returns true only if `there is one+ stored L2 block in the pending batch`

PR #4255, opened 2025-06-28
feat(ci): fast integration tests framework

tomg10:


core/tests/highlevel-test-tools/.gitignore
Deniallugo, line 1: nit: we don't need additional .gitignore

zkstack_cli/crates/zkstack/src/commands/ecosystem/args/init.rs
```diff
@@ -17,6 +17,50 @@ use crate::{
     },
 };
 
+/// Check if L1 RPC is healthy by calling eth_chainId
+async fn check_l1_rpc_health(l1_rpc_url: &str) -> anyhow::Result<()> {
```
Deniallugo, line 24: you can use client and do not make the queries by your own and i'd move it somewhere to utils.

core/tests/highlevel-test-tools/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "scripts": {
+    "test": "vitest run"
+  },
+  "devDependencies": {
+    "@types/node": "^18.19.15",
+    "@types/uuid": "^9.0.0",
+    "@vitest/coverage-v8": "^1.0.0",
+    "ts-node": "^10.1.0",
+    "typescript": "^4.3.5",
+    "vitest": "^1.0.0"
```
popzxc: Concern: we introduce an _additional_ testing framework to the workspace. It is jest-compatible, sure, but AFAIU vitest is mostly designed to work together with vite (which we don't seem to use).
Why do we need it? Can we just use jest to avoid making the tech stack bigger?

bin/build_and_init_ecosystem
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
```
popzxc, line 1: 2nd concern: in the past, `zk` tool was introduced as a way to get rid of heckton of bash scripts. Then, it was reimplemented as `zkstack`. And now we seem to be adding bash scripts again. This signals a very big issue with `zkstack` to me -- if it's easier to add a shell script rather than a command for seemingly common action, we're doing something wrong.

core/tests/highlevel-test-tools/global-setup.ts
```diff
@@ -0,0 +1,71 @@
+import { exec } from 'child_process';
```
popzxc, line 1: Third concern: we're adding a new undocumented testing tool to the workspace. I expect that it will significantly increase the speed of codebase rotting: people won't know which tool to edit, and will not know how to work with the tool to keep it "alive". There is a good chance than in N months, the combination of this tool and `ts-integration` will work significantly worse than today.

Maybe a better way (though obviously more time consuming) would be to integrate this functionality to `ts-integration` and embed existing high-level tests there, so that we _reduce_ the number of tests rather than increase. Just in case, I don't suggest doing it naively (such an attempt will most likely fail), but rather I see it as an exercise to come up with a set of _tests_ (which can be either parallel or sequential w.r.t single chain) and _test configurations_ (e.g. a set of chains against which the tests must be ran) that can be invoked against a prepared setup.

popzxc (commented):
I absolutely see value in the work that has been done here, and it's cool that it brings significant CI speed improvements. I do not object merging this PR.

That said, I find the need in this PR to be a symptom of significant issues with the repo maintenance on many levels, including CI, tooling, and testing. While it does bring improvements, it does so at cost of more (undocumented) complexity, which will _also_ likely suffer from the lack of maintenance, and bring more issues as repo gets more changes.

I would strongly recommend the team to evaluate the level of support they want to provide, and potentially invest into rethinking our approach to CI and testing in general.

bin/build_and_init_ecosystem
Deniallugo: i'd disagree with it.. 
it's not a common action at all.. building a binary in parallel with ecosystem init is not a common think

core/tests/highlevel-test-tools/.gitignore
Deniallugo, line 1: do we need a gitignore?

Deniallugo (commented):
TBH, i don't feel it's the right move at all. 

Some commands, like genearte load or wait for batches are good, and probably it's a good command to zkstack dev test. 

So, i'd agree to move the additional tools you introduced to zkstack directly. 

But other than that, looks like it's an additional complexity with a little value. 

we could achieve the same behaviour with the current tooling and with less code

core/tests/highlevel-test-tools/.gitignore
tomg10: fixed

zkstack_cli/crates/zkstack/src/commands/ecosystem/args/init.rs
tomg10: fixed

core/tests/highlevel-test-tools/.gitignore
tomg10: fixed

RomanBrodetski (approved):

RomanBrodetski (approved):

artmakh (approved):

PR #4266, opened 2025-07-02
feat: Eth proof manager sender

Artemka374:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


zkstack_cli/crates/zkstack/src/messages.rs
```diff
@@ -75,6 +75,15 @@ pub(super) const MSG_OBSERVABILITY_PROMPT: &str = "Do you want to setup observab
 pub(super) const MSG_DEPLOY_ECOSYSTEM_PROMPT: &str =
     "Do you want to deploy ecosystem contracts? (Not needed if you already have an existing one)";
 pub(super) const MSG_L1_RPC_URL_PROMPT: &str = "What is the RPC URL of the L1 network?";
+pub(super) const MSG_FERMAH_ADDRESS_PROMPT: &str =
+    "What is the address of the Fermah(who will be able to send proofs on Fermah side)?";
+pub(super) const MSG_LAGRANGE_ADDRESS_PROMPT: &str =
+    "What is the address of the Lagrange(who will be able to send proofs on Lagrange side)?";
+pub(super) const MSG_USDC_ADDRESS_PROMPT: &str =
```
EmilLuta: NIT: This message seems broken. Nobody sends proofs from USDC side.

zkstack_cli/crates/zkstack/src/messages.rs
```diff
@@ -75,6 +75,15 @@ pub(super) const MSG_OBSERVABILITY_PROMPT: &str = "Do you want to setup observab
 pub(super) const MSG_DEPLOY_ECOSYSTEM_PROMPT: &str =
     "Do you want to deploy ecosystem contracts? (Not needed if you already have an existing one)";
 pub(super) const MSG_L1_RPC_URL_PROMPT: &str = "What is the RPC URL of the L1 network?";
+pub(super) const MSG_FERMAH_ADDRESS_PROMPT: &str =
```
EmilLuta, line 81: NIT: I would rephrase this to: `What is Fermah's Proving Network address (allows acknowledging/submitting proofs for Fermah)?`

zkstack_cli/crates/zkstack/src/messages.rs
```diff
@@ -75,6 +75,15 @@ pub(super) const MSG_OBSERVABILITY_PROMPT: &str = "Do you want to setup observab
...
+    "What is the address of the Fermah(who will be able to send proofs on Fermah side)?";
+pub(super) const MSG_LAGRANGE_ADDRESS_PROMPT: &str =
+    "What is the address of the Lagrange(who will be able to send proofs on Lagrange side)?";
+pub(super) const MSG_USDC_ADDRESS_PROMPT: &str =
+    "What is the address of the USDC(who will be able to send proofs on USDC side)?";
+pub(super) const MSG_PROOF_MANAGER_OWNER_ADDRESS_PROMPT: &str = "What is the address of the ProofManager owner(who will be able to send proofs on ProofManager side)?";
```
EmilLuta: NIT: Maybe `What address is allowed to submit Proof Requests?`

zkstack_cli/crates/zkstack/src/commands/prover/deploy_proving_network.rs
```diff
@@ -33,7 +37,15 @@ pub(crate) async fn deploy_proving_networks(
         ));
     }
 
-    shell.set_var("RPC_URL", l1_rpc_url);
+    shell.set_var("RPC_URL", args.l1_rpc_url);
```
EmilLuta: Dumb question, but why do we deploy this on L1, instead of the chain we're running on? In production, we'd never deploy it on L1, right?

zkstack_cli/crates/zkstack/src/commands/prover/deploy_proving_network.rs
```diff
@@ -58,13 +70,13 @@ pub(crate) async fn deploy_proving_networks(
     let mut proxy_admin_addr = String::new();
 
     for line in output.lines() {
```
EmilLuta, line 160: This looks like something waiting to be broken/misused. Can we at least add proper docs on how to use the command/function?

etc/env/file_based/general.yaml
```diff
@@ -144,6 +144,7 @@ data_handler:
   http_port: 3320
   proof_generation_timeout_in_secs: 18000
   gateway_api_url: http://127.0.0.1:3322
+  proving_mode: PROVING_NETWORK
```
EmilLuta: Maybe it's worthwhile to have this enum something across the lines of:
```
PROVER_CLUSTER # goes to internal prover cluster
TRIAL_PROVING_NETWORK # small percentage, configurable, goes to proving networks
FULLY_PROVING_NETWORK # big percentage (99.9% or w/e we decide), hard coded, goes to proving networks
ONLY_PROVING_NETWORK # 100% goes to proving networks (I.E. for chains that prefer not to run their own prover cluster or when we have an internal outage)
```

core/lib/basic_types/src/prover_dal.rs
```diff
@@ -10,6 +10,12 @@ use crate::{
     L1BatchNumber, L2ChainId,
 };
 
+#[derive(Debug, Clone, PartialEq)]
+pub enum ProvingMode {
```
EmilLuta: I believe comment from Config applies here. Also, should we have a default implementation that defaults to ProverCluster so we're backwards compatible? (if not, let's be 100% sure we have the backwards compatibility layer used)

core/lib/config/src/configs/eth_proof_manager.rs
```diff
@@ -6,15 +6,58 @@ use smart_config::{DescribeConfig, DeserializeConfig};
...
     #[config(default_t = Duration::from_secs(10))]
     pub event_poll_interval: Duration,
+    #[config(default_t = Duration::from_secs(10))]
+    pub request_sending_interval: Duration,
     #[config(default_t = 1000)]
     pub event_expiration_blocks: u64,
+    #[config(default_t = 1000000)]
+    pub default_priority_fee_per_gas: u64,
+    // todo: should be revisited
+    #[config(default_t = 10)]
```
EmilLuta: Given our current price, should be 4$. Also, make sure that 4 corresponds to 4$ and not 0.004$.

core/lib/config/src/configs/eth_proof_manager.rs
```diff
@@ -6,15 +6,58 @@ use smart_config::{DescribeConfig, DeserializeConfig};
...
+    pub max_reward: u64,
+    #[config(default_t = Duration::from_secs(120))]
```
EmilLuta: I assume this is internal for us, right? If so, maybe it makes sense to make it closer to 150s (in case issues with network & other problems).

core/lib/config/src/configs/eth_proof_manager.rs
```diff
@@ -6,15 +6,58 @@ use smart_config::{DescribeConfig, DeserializeConfig};
...
+    pub acknowledgment_timeout: Duration,
+    #[config(default_t = Duration::from_secs(7200))]
```
EmilLuta: 2h sounds very generous. If they are not generated for proving network, we'll have 1h to generate the entire proof. I'd suggest we keep it closer to 1h (which is what it takes for us internally to get a proof going).

core/lib/config/src/configs/eth_proof_manager.rs
```diff
@@ -6,15 +6,58 @@ use smart_config::{DescribeConfig, DeserializeConfig};
...
+    pub max_tx_gas: u64,
+    #[config(default)]
+    pub path_to_fflonk_verification_key: String,
+    #[config(default)]
+    pub path_to_plonk_verification_key: String,
 }
 
 impl Default for EthProofManagerConfig {
     fn default() -> Self {
         Self {
```
EmilLuta: If we have the `#[config(default)]` above, do we need to duplicate the data here as well? I assume not.

core/lib/config/src/configs/proof_data_handler.rs
```diff
@@ -14,6 +15,32 @@ pub struct ProofDataHandlerConfig {
     pub proof_gen_data_submit_interval: Duration,
     #[config(default_t = true)]
     pub fetch_zero_chain_id_proofs: bool,
+    #[config(default_t = ProvingMode::ProverCluster)]
+    pub proving_mode: ProvingMode,
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub enum ProvingMode {
```
EmilLuta, line 23: Is this a clone of the same struct from `basic_types/src/prover_dal.rs`? If so, let's keep just 1.

core/lib/dal/migrations/20250630105521_add_eth_proof_manager_table.up.sql
```diff
@@ -0,0 +1,16 @@
+CREATE TABLE IF NOT EXISTS eth_proof_manager (
+    l1_batch_number BIGINT NOT NULL PRIMARY KEY REFERENCES l1_batches (number) ON DELETE CASCADE,
```
EmilLuta, line 2: Random thought, but do we care about having chain id here as well? I assume not, just double checking.

core/lib/dal/migrations/20250630105521_add_eth_proof_manager_table.up.sql
```diff
@@ -0,0 +1,16 @@
...
+    assigned_to TEXT,
+    status TEXT NOT NULL,
+    witness_inputs_url TEXT NOT NULL,
```
EmilLuta, line 5: What bucket contains this data? (haven't seen the rest of the implementation, yet)

core/lib/dal/src/proof_generation_dal.rs
```diff
@@ -40,8 +40,16 @@ impl ProofGenerationDal<'_, '_> {
     pub async fn lock_batch_for_proving(
         &mut self,
         processing_timeout: Duration,
+        proving_mode: ProvingMode,
     ) -> DalResult<Option<L1BatchNumber>> {
         let processing_timeout = pg_interval_from_duration(processing_timeout);
+
+        let status = if proving_mode == ProvingMode::ProverCluster {
```
EmilLuta: NIT: Worthwhile to map this to a function on `ProvingMode`? Think `proving_mode.status_for_dal()`.

core/lib/dal/src/proof_generation_dal.rs
```diff
@@ -442,7 +503,7 @@ mod tests {
 
         let picked_l1_batch = conn
             .proof_generation_dal()
-            .lock_batch_for_proving(Duration::MAX)
+            .lock_batch_for_proving(Duration::MAX, ProvingMode::ProverCluster)
```
EmilLuta, line 544: Maybe we should have some separate tests with `ProvingMode::ProvingNetwork` as well?

core/lib/dal/src/proof_generation_dal.rs
```diff
@@ -63,7 +71,7 @@ impl ProofGenerationDal<'_, '_> {
                             AND l1_batches.hash IS NOT NULL
                             AND l1_batches.aux_data_hash IS NOT NULL
                             AND l1_batches.meta_parameters_hash IS NOT NULL
-                            AND status = 'unpicked'
+                            AND status = $2
```
EmilLuta: I believe the way the code is structured, it implies that batches will be stuck if you change ProvingMode mid proving.

For instance: If a batch is picked by Proving Network and then we move to ProverCluster, that batch will be stuck forever (because we look for `unpicked` as status to retry, but it would be on `fallbacked`). That entire interaction looks a bit weird and I believe it might worthwhile to unify it (but also, please document all these state transitions, it was very difficult to make sense of the logic by looking at code only).

core/lib/dal/src/eth_proof_manager_dal.rs
```diff
@@ -0,0 +1,363 @@
...
+}
+
+impl ProvingNetwork {
+    pub fn as_str(&self) -> &str {
+        match self {
+            ProvingNetwork::None => "none",
+            ProvingNetwork::Lagrange => "lagrange",
+            ProvingNetwork::Fermah => "fermah",
+        }
+    }
```
EmilLuta, line 57: Do we really need to map this internally? What's the use case? Also, some docstrings would be nice for this enum.

core/lib/dal/src/eth_proof_manager_dal.rs
```diff
@@ -0,0 +1,363 @@
...
+        )
+        .instrument("update_status")
+        .with_arg("batch_number", &batch_number)
+        .with_arg("status", &status.as_str())
+        .execute(self.storage)
+        .await?;
+
+        Ok(())
+    }
+
```
EmilLuta: We have a bunch of functions that deal with status (like mark as validated, mark as proven, etc.), do we need `update_status`? Is there a world where we can stay consistent with them (either have a specific function for each for code readability, or use `update_status` for reducing the amount of code to maintain)?

core/lib/dal/src/eth_proof_manager_dal.rs
```diff
@@ -0,0 +1,363 @@
...
+                    status = $4
+                    AND validated_proof_request_tx_sent_at < NOW() - $5::INTERVAL
+                )
+                OR (status = $6 AND proof_validation_result IS false)
+            RETURNING l1_batch_number
+            "#,
+            EthProofManagerStatus::Fallbacked.as_str(),
+            EthProofManagerStatus::Sent.as_str(),
+            &acknowledgment_timeout,
+            EthProofManagerStatus::Acknowledged.as_str(),
```
EmilLuta, line 295: :fire: More of this!

core/lib/dal/src/eth_proof_manager_dal.rs
```diff
@@ -0,0 +1,363 @@
...
+        .execute(&mut transaction)
+        .await?;
+
+        sqlx::query!(
+            r#"
+            UPDATE proof_generation_details
+            SET status = 'fallbacked', updated_at = NOW()
+            WHERE l1_batch_number = $1
+            "#,
+            i64::from(batch_number.0),
```
EmilLuta, line 361: NIT: I think `fallback_batch(batch_number: ...)` should be enough for the signature.

core/lib/object_store/src/raw.rs
```diff
@@ -20,6 +20,7 @@ pub enum Bucket {
     StorageSnapshot,
     DataAvailability,
     VmDumps,
+    PublicWitnessInputs,
```
EmilLuta, line 23: Ummm, wouldn't this mean we use the `core` bucket and have a public `"folder"`? Is that possible under GCS? Might be worthwhile to check.

core/node/eth_proof_manager/src/watcher/mod.rs
```diff
@@ -29,14 +28,35 @@ impl EthProofWatcher {
         blob_store: Arc<dyn ObjectStore>,
         config: EthProofManagerConfig,
     ) -> Self {
+        let fflonk_vk = serde_json::from_slice::<FflonkFinalVerificationKey>(
```
EmilLuta, line 31: Should we support both FFLONK and PLONK? Maybe we can go only for FFLONK, WDYT?

core/node/eth_proof_manager/src/watcher/mod.rs
```diff
@@ -50,18 +70,28 @@ impl EthProofWatcher {
...
+
+                let from_block = self
+                    .connection_pool
+                    .connection()
+                    .await?
+                    .eth_watcher_dal()
+                    .get_or_set_next_block_to_process(
+                        event.event_type(),
+                        self.client.chain_id(),
+                        to_block.saturating_sub(self.config.event_expiration_blocks),
```
EmilLuta, line 78: As mentioned in previous PR, this diff seems pretty random. We should only look to the last block unproven, rather than a fixed window.

core/node/eth_proof_manager/src/sender/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+            if *stop_receiver.borrow() {
+                tracing::info!("Stop request received, eth proof sender is shutting down");
+                break;
+            }
+
+            if let Err(e) = proof_request_submitter.loop_iteration().await {
+                tracing::error!("Error submitting proof request: {e}");
+            }
+
+            if let Err(e) = proof_validation_submitter.loop_iteration().await {
```
EmilLuta: Dumb question, but shouldn't these 2 be in different threads? Given they're somewhat sync, if proof_request_submitter takes a lot of time (think GCS being down), we won't do any proof validation until it's fixed. The reverse also applies.

core/lib/object_store/src/raw.rs
Artemka374: No, these are just bucket names. You see there are actually all the buckets in this enum

core/node/eth_proof_manager/src/watcher/mod.rs
Artemka374: Yeah, makes sense I guess

etc/env/file_based/general.yaml
Artemka374: I would prefer to add such functionality in a separate PR... It seems like it will require a proper configuration

core/lib/config/src/configs/proof_data_handler.rs
Artemka374: it's not that simple... I need to have implementation from smart config(which I don't think we want to import to basic types) and I also need to use it in things like prover DAL, which will require importing configs. Overall we'll just make everything heavier if we try to have just 1

core/lib/dal/migrations/20250630105521_add_eth_proof_manager_table.up.sql
Artemka374: No, we can get it from configs

core/lib/dal/src/proof_generation_dal.rs
Artemka374: Actually I think that makes sense. Will add

core/lib/dal/src/eth_proof_manager_dal.rs
Artemka374: This is just for debugging purposes, so we know in DB by which chain batch was picked up

core/lib/dal/src/eth_proof_manager_dal.rs
Artemka374: Actually it's not even used anywhere, I'll just remove it

core/lib/basic_types/src/prover_dal.rs
Artemka374: it's a bit complicated. Overall we indeed are using the default value for this enum, but it is specified in config

core/lib/dal/src/proof_generation_dal.rs
Artemka374: I have reworked this interaction by adding a separate column to `proof_generation_details`

core/lib/dal/migrations/20250630105521_add_eth_proof_manager_table.up.sql
Artemka374: it will be separate public bucket

zkstack_cli/crates/zkstack/src/commands/prover/args/deploy_proving_network.rs
```diff
@@ -0,0 +1,109 @@
...
+
+impl DeployProvingNetworkArgs {
+    pub fn fill_values_with_prompt(self, default_rpc_url: &str) -> DeployProvingNetworkArgsFinal {
+        let rpc_url = self.rpc_url.unwrap_or_else(|| {
+            Prompt::new(MSG_RPC_URL_PROMPT_PROVING_NETWORK)
+                .default(default_rpc_url)
+                .ask()
+        });
+        let l1_rpc_url = self.l1_rpc_url.unwrap_or_else(|| {
+            Prompt::new(MSG_L1_RPC_URL_PROMPT)
```
EmilLuta, line 55: Is there a specific reason this is L1 instead of L2? Or we consider Era "L1" in this case? If so, the URL seems wrong (:8545 is reth, IIRC)

zkstack_cli/crates/zkstack/src/commands/chain/args/init/configs.rs
```diff
@@ -39,7 +39,7 @@ pub struct InitConfigsArgsFinal {
 impl InitConfigsArgs {
     pub fn fill_values_with_prompt(self, config: &ChainConfig) -> InitConfigsArgsFinal {
         let l1_rpc_url = self.l1_rpc_url.unwrap_or_else(|| {
-            let mut prompt = Prompt::new(MSG_L1_RPC_URL_PROMPT);
+            let mut prompt = Prompt::new(MSG_RPC_URL_PROMPT);
```
EmilLuta: Is this on purpose? (question applies to everything under `zkstack_cli/crates/zkstack/stc/commands/`)

etc/env/file_based/general.yaml
```diff
@@ -280,8 +281,26 @@ consensus:
 
 gateway_migrator:
   eth_node_poll_interval: 300 ms
-
+
```
EmilLuta: Nit: Unnecessary spaces here.

core/lib/dal/Cargo.toml
```diff
@@ -29,7 +29,7 @@ zksync_l1_contract_interface.workspace = true
 
 # DI-specific dependencies
 zksync_node_framework = { workspace = true, optional = true }
-zksync_config = { workspace = true, optional = true }
+zksync_config.workspace = true
```
EmilLuta, line 32: Nit: Is this necessary, can't we keep it optional?

core/lib/dal/migrations/20250630105521_add_eth_proof_manager_table.up.sql
```diff
@@ -0,0 +1,19 @@
...
+    validated_proof_request_tx_sent_at TIMESTAMP,
+    created_at TIMESTAMP NOT NULL,
+    updated_at TIMESTAMP NOT NULL
+);
+
+CREATE TYPE proving_mode AS ENUM ('proving_network', 'prover_cluster');
+ALTER TABLE proof_generation_details ADD COLUMN proving_mode proving_mode DEFAULT 'proving_network';
+
+ALTER TYPE event_type ADD VALUE 'ProofRequestAcknowledged';
+ALTER TYPE event_type ADD VALUE 'ProofRequestProven';
```
EmilLuta, line 19: Bikeshedding: Fulfilled or Done would've been better than Proven, but I wouldn't change everything in the code now just for this.

core/lib/dal/src/proof_generation_dal.rs
```diff
@@ -64,6 +71,10 @@ impl ProofGenerationDal<'_, '_> {
                             AND l1_batches.aux_data_hash IS NOT NULL
                             AND l1_batches.meta_parameters_hash IS NOT NULL
                             AND status = 'unpicked'
+                            AND (
+                                $2 = 'prover_cluster'
```
EmilLuta: I'm not sure I understand this line. Are you dynamically injecting a column to check for `prover_cluster`?

core/lib/dal/src/proof_generation_dal.rs
```diff
@@ -88,6 +100,53 @@ impl ProofGenerationDal<'_, '_> {
...
 
+    pub async fn lock_batch_for_proving_network(
+        &mut self,
+        proving_mode: ProvingMode,
+    ) -> DalResult<Option<L1BatchNumber>> {
+        let result: Option<L1BatchNumber> = sqlx::query!(
+            r#"
+            UPDATE proof_generation_details
+            SET
+                status = 'picked_by_prover',
```
EmilLuta: The usage of status, proving_mode and some other column to infer status looks confusing. I wonder, can't we do better here?

core/lib/dal/src/proof_generation_dal.rs
```diff
@@ -480,7 +539,7 @@ mod tests {
 
         let picked_l1_batch = conn
             .proof_generation_dal()
-            .lock_batch_for_proving(Duration::MAX)
+            .lock_batch_for_proving(Duration::MAX, ProvingMode::ProverCluster)
```
EmilLuta, line 582: Should we maybe add some extra tests in here for the new methods?

core/lib/dal/src/eth_sender_dal.rs
```diff
@@ -952,6 +952,7 @@ impl EthSenderDal<'_, '_> {
         )
         .execute(self.storage.conn())
         .await?;
+
```
EmilLuta, line 1081: Okay...?

core/lib/config/src/configs/eth_proof_manager.rs
```diff
@@ -2,19 +2,62 @@ use std::time::Duration;
 
 use smart_config::{DescribeConfig, DeserializeConfig};
 
+use crate::ObjectStoreConfig;
 #[derive(Debug, Clone, PartialEq, DescribeConfig, DeserializeConfig)]
 pub struct EthProofManagerConfig {
+    /// Chain id of settlement layer(where contracts are deployed)
+    #[config(default_t = 271)]
+    pub sl_chain_id: u64,
+    /// HTTP RPC URL of settlement layer(where contracts are deployed)
```
EmilLuta: Don't we deploy on Era? How comes we deploy on SL?

core/lib/config/src/configs/object_store.rs
```diff
@@ -21,6 +21,12 @@ pub struct ObjectStoreConfig {
     pub local_mirror_path: Option<PathBuf>,
 }
 
+impl Default for ObjectStoreConfig {
```
EmilLuta, line 24: Nit: Should the default implementation be the one for tests? I wonder if someone might break something up by mistake with this.

core/node/eth_proof_manager/src/tests/fallbacking.rs
```diff
@@ -0,0 +1,276 @@
...
+// test basic flow of proof manager with fallbacking due to acknowledgment timeout
+#[tokio::test]
+async fn test_fallbacking_acknowledgment_timeout() {
+    let connection_pool = ConnectionPool::<Core>::test_pool().await;
+    let config = test_config();
+    let blob_store = MockObjectStore::arc();
+
+    let mut connection = connection_pool.connection().await.unwrap();
+
+    prepare_database(&mut connection).await;
```
EmilLuta: Nit: These first lines look like some boilerplate, maybe worthwhile to extract in a struct and call it once to initialize everything?
```
let context = Context::new().init();
```

core/node/eth_proof_manager/src/tests/fallbacking.rs
```diff
@@ -0,0 +1,276 @@
...
+        .await
+        .unwrap();
+
+    connection
+        .eth_proof_manager_dal()
+        .mark_batch_as_sent(L1BatchNumber(1), H256::zero())
+        .await
+        .unwrap();
+
+    tokio::time::sleep(Duration::from_secs(2)).await;
```
EmilLuta, line 45: Nit: Is there a way to dodge sleeping here? I'm thinking either to make the duration shorter, or simply hardcode the value in DB.

core/node/eth_proof_manager/src/node.rs
```diff
@@ -42,25 +53,70 @@ pub struct Output {
...
+        let client = Box::new(
+            Client::<L2>::http(SensitiveUrl::from_str(&http_rpc_url).expect("failed to parse url"))
+                .expect("failed to create client")
+                .for_network(L2::from(L2ChainId::from(sl_chain_id.0 as u32)))
+                .build(),
+        ) as Box<DynClient<L2>>;
+
+        let gas_adjuster_client = GasAdjusterClient::from(client.clone_boxed());
+
+        // todo: what is this?
```
EmilLuta: Did we make sense out of this comment?

zkstack_cli/crates/zkstack/src/commands/prover/args/deploy_proving_network.rs
Artemka374: We need L1 here just to transfer tokens to L2(if needed for testing)

core/lib/dal/migrations/20250630105521_add_eth_proof_manager_table.up.sql
Artemka374: This is the name of event itself, so that's why it called that way

core/lib/dal/src/proof_generation_dal.rs
Artemka374: This is a check for a global proving mode which is specified in config. I could've made 2 queries depending on that value, but I don't see much sense in it

core/lib/config/src/configs/eth_proof_manager.rs
Artemka374: I called it as some abstract layer where it is deployed. But maybe it's confusing, I can change that

core/node/eth_proof_manager/src/node.rs
Artemka374: More or less :) removing it

core/lib/dal/Cargo.toml
Artemka374: we are reusing values from configs, so we either import dal to config or the other way around

core/lib/dal/src/proof_generation_dal.rs
Artemka374: Yes, they are covered inside the proof manager tests

core/lib/config/src/configs/object_store.rs
Artemka374: I'm not sure which implementation should be default, there's no easy answer

core/node/eth_proof_manager/src/tests/fallbacking.rs
Artemka374: Not sure actually. It's just the question of how we understand time in tests, if we are at timeout - we need to do something

core/lib/dal/src/proof_generation_dal.rs
EmilLuta: But what does it mean that `$2 = 'prover_cluster'`. What column are we checking?

PR #4273, opened 2025-07-02
fix(consensus): Debug page server port reuse

brunoffranca:
## What ❔

Introduces a new consensus version that contains a fix that solves the debug page server being stopped because another debug page server is already using the same port.
Also adds new config parameters: one that determines how often a  node reads from the ConsensusRegistry contract for validator set updates, another for the maximum size of transactions propagated in the p2p network.

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Operational changes

The new parameters can be left at their default values.

## Checklist

### DIFF
--- a/core/lib/config/src/configs/consensus.rs
+++ b/core/lib/config/src/configs/consensus.rs
@@ -146,9 +146,14 @@ pub struct ConsensusConfig {
     /// that will be advertised to peers, so that they can connect to this
     /// node.
     pub public_addr: Host,
+    /// Local socket address to expose the node debug page.
+    pub debug_page_addr: Option<std::net::SocketAddr>,
     /// Maximal allowed size of the payload in bytes.
     #[config(default_t = ByteSize(2_500_000), with = Fallback(SizeUnit::Bytes))]
     pub max_payload_size: ByteSize,
+    /// Maximal allowed size of transactions propagated through the p2p network, in bytes.
+    #[config(default_t = ByteSize(1_000_000), with = Fallback(SizeUnit::Bytes))]
+    pub max_transaction_size: ByteSize,
     /// View timeout duration.
     #[config(default_t = Duration::from_secs(2), with = Fallback(CustomDurationFormat))]
     pub view_timeout: Duration,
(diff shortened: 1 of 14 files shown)


core/lib/config/src/configs/consensus.rs
```diff
@@ -170,19 +175,20 @@ pub struct ConsensusConfig {
     /// establish and maintain.
     #[config(default, with = Entries::WELL_KNOWN.named("key", "addr"))]
     pub gossip_static_outbound: BTreeMap<NodePublicKey, Host>,
+    /// Rate limiting configuration for the p2p RPCs.
+    #[config(nest)]
+    pub rpc: RpcConfig,
+
+    /// Rate at which the node tries to read from the ConsensusRegistry contract, in milliseconds.
+    /// It should be set to a value that is much less than what we expect a validator epoch to be.
+    #[config(default_t = Duration::from_secs(30), with = Fallback(CustomDurationFormat))]
```
perekopskiy, line 184: Do you need `CustomDurationFormat`? Especially since Duration is rounded to seconds

prover/crates/bin/prover_autoscaler/src/global/scaler.rs
```diff
@@ -1618,7 +1618,7 @@ mod tests {
         };
 
         assert_eq!(
-            scaler.calculate(&"prover".into(), 2 * 1500 + 1 * 3000 - 1500, &clusters),
```
perekopskiy: let's revert change in this file, because they block CI from passing

core/lib/config/src/configs/consensus.rs
brunoffranca: No, but it's also used in another field in this config. I'd rather change both in a future PR (was already on my list).

prover/crates/bin/prover_autoscaler/src/global/scaler.rs
brunoffranca: This was from the lint.

perekopskiy (dismissed):

perekopskiy (approved):

PR #4288, opened 2025-07-04
fix(prover): Propagating errors for CP & Compressor

AlexOnTheStorm:
## What ❔
Fix for error propagating for CP & Compressor. We need run_inner for instant cancelling all the operations upon receiving SIGTERM

## Why ❔
Errors weren't propagating from tokio::select!

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/prover/crates/bin/circuit_prover/src/main.rs
+++ b/prover/crates/bin/circuit_prover/src/main.rs
@@ -75,9 +75,12 @@ async fn main() -> anyhow::Result<()> {
     let (metrics_stop_sender, metrics_stop_receiver) = tokio::sync::watch::channel(false);
 
     tokio::select! {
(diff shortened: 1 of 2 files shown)


prover/crates/bin/circuit_prover/src/main.rs
```diff
@@ -75,7 +75,9 @@ async fn main() -> anyhow::Result<()> {
     let (metrics_stop_sender, metrics_stop_receiver) = tokio::sync::watch::channel(false);
 
     tokio::select! {
-        _ = run_inner(cancellation_token.clone(), metrics_stop_receiver, &mut managed_tasks) => {},
+        res = run_inner(cancellation_token.clone(), metrics_stop_receiver, &mut managed_tasks) => {
+            res?
```
EmilLuta, line 79: NIT: I would've added a `.context(...)` here, but this should be enough.

PR #4299, opened 2025-07-08
feat: add gas cap configuration for eth_call requests

bxpana:
## What ❔

- Add optional eth_call_gas_cap field to Web3JsonRpcConfig
- Modify default_eth_call_gas() to respect gas cap when set
- Update eth and debug API namespaces to pass gas cap parameter
- Add comprehensive unit tests for gas cap logic
- Include documentation and configuration examples

The gas cap uses min(protocol_default,
configured_cap) when set,
maintaining full backward compatibility. Setting
to 0 disables limiting.

## Why ❔

- Implements gas limiting for eth_call similar to geth's --rpc.gascap option to prevent resource exhaustion from computationally expensive calls.
- Compatibility: Matches geth --rpc.gascap behavior
- Request from Abstract for support

## Is this a breaking change?
- [ ] Yes
- [x] No

## Operational changes
- Added optional `eth_call_gas_cap` field to Web3JsonRpcConfig in YAML configuration
  - Example: `api.web3_json_rpc.eth_call_gas_cap: 50000000`
- Default behavior unchanged when not specified (uses protocol version gas limits)

- **Optional**: Configure gas cap to prevent resource exhaustion from expensive eth_call requests
- **Compatible**: Behaves like geth's --rpc.gascap option for familiar migration
- **Flexible**: Set to 0 for unlimited, positive value for cap, or omit for protocol default
- **Location**: Add to existing node configuration YAML under `api.web3_json_rpc` section

- Added commented examples in `etc/env/configs/ext-node.toml`
- Documentation added to External Node configuration guide

## Checklist


core/node/api_server/src/web3/namespaces/eth.rs
```diff
@@ -77,7 +77,11 @@ impl EthNamespace {
                 .diff_with_block_args(&block_args),
         );
         if request.gas.is_none() {
-            request.gas = Some(block_args.default_eth_call_gas(&mut connection).await?);
+            request.gas = Some(
+                block_args
+                    .default_eth_call_gas(&mut connection, self.state.api_config.eth_call_gas_cap)
+                    .await?,
+            );
```
perekopskiy, line 96: In your case, you apply cap only if `request.gas` is not specified, i.e. user can provide limit higher than the value from config and it will be processed. Reth returns error in this case: 
```
        if request.gas > Some(self.call_gas_limit()) {
            // configured gas exceeds limit
            return Err(
                EthApiError::InvalidTransaction(RpcInvalidTransactionError::GasTooHigh).into()
            )
        }
```
I'm sure geth does the same. Please change the code so it has the same behavior

core/node/api_server/src/web3/namespaces/debug.rs
```diff
@@ -279,7 +279,11 @@ impl DebugNamespace {
                 .diff_with_block_args(&block_args),
         );
         if request.gas.is_none() {
-            request.gas = Some(block_args.default_eth_call_gas(&mut connection).await?);
+            request.gas = Some(
```
perekopskiy, line 294: same here

core/node/api_server/src/execution_sandbox/mod.rs
```diff
@@ -388,6 +389,79 @@ impl BlockArgs {
...
+    }
+}
+
+#[cfg(test)]
+mod gas_cap_tests {
+    use zksync_multivm::utils::get_eth_call_gas_limit;
+    use zksync_types::ProtocolVersionId;
+
+    #[test]
+    fn test_gas_cap_logic() {
```
perekopskiy, line 421: Such test is not good because it tests the code in tests itself, not the one that API uses. So if someone changes `default_eth_call_gas` the test will still pass. You can extract
```
let default_gas_limit = get_eth_call_gas_limit(protocol_version.into());

        // Apply gas cap if specified (0 means no cap)
        let effective_gas_limit = if let Some(cap) = gas_cap.filter(|&cap| cap > 0) {
            std::cmp::min(default_gas_limit, cap)
        } else {
            default_gas_limit
        };
```
into a separate method and use it both in `default_eth_call_gas` and the test

PR #4308, opened 2025-07-11
feat: split ecosystem init command

Raid5594:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/zkstack_cli/crates/zkstack/src/commands/chain/gateway/convert_to_gateway.rs
+++ b/zkstack_cli/crates/zkstack/src/commands/chain/gateway/convert_to_gateway.rs
@@ -37,6 +37,7 @@ use crate::{
         governance_execute_calls, grant_gateway_whitelist, revoke_gateway_whitelist,
         set_transaction_filterer, AdminScriptMode,
     },
+    commands::chain::utils::display_admin_script_output,
     messages::MSG_CHAIN_NOT_INITIALIZED,
     utils::forge::{check_the_balance, fill_forge_private_key, WalletOwner},
 };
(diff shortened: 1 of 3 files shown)


zkstack_cli/crates/zkstack/src/commands/ecosystem/init_core_contracts.rs
```diff
@@ -0,0 +1,319 @@
...
+        ecosystem_config,
+        initial_deployment_config,
+        init_args.support_l2_legacy_shared_bridge_test,
+    )
+    .await?;
+    contracts.save_with_base_path(shell, &ecosystem_config.config)?;
+    Ok(contracts)
+}
+
+async fn deploy_erc20(
```
Raid5594: We can probably move this function and init_chains to common folder

zkstack_cli/crates/zkstack/src/commands/chain/gateway/convert_to_gateway.rs
```diff
@@ -113,6 +169,10 @@ pub async fn run(args: ForgeScriptArgs, shell: &Shell) -> anyhow::Result<()> {
     )
     .await?;
 
+    if convert_to_gw_args.only_save_calldata {
+        display_admin_script_output(output);
```
Raid5594, line 173: Do we want to save the grant gateway whitelist calldata as well?

Without this been executed, the gw_vote_preparation will fail.
I can add the check to see if addresses have been whitelisted before continuing

StanislavBreadless: yes, they should be saved, but as a separate thing, since it is to be executed by the GW admin

PR #4319, opened 2025-07-16
fix(prover): Fixing sequence_number assigning

AlexOnTheStorm:
## What ❔
1. `sequence_number` is now not continuous across all the BWG output artifacts but instead reflects a sequence number within the same circuit type. 
2. Also it is guaranteed that urls match (sequence_number, circuit_id) pair in DB.

## Why ❔
Circuit url includes a sequence_number which is a continuous numbering across all the circuits. It turned out that the ordering between different circuit types is not strict. e.g. we have circuit_10 A, B, C and circuit_2 X, Y, Z. They can be pushed to vec in different order like [A,B,C,X,Y,Z], [A,X,B,Y,C,Z], [A,X,Y,B,C,Z] etc. And then sequence_number is assigned like [0,1,2,3,4,5].

That can cause creating two identical jobs with different url and following proving failure. 

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/prover/crates/lib/prover_dal/src/fri_prover_dal.rs
+++ b/prover/crates/lib/prover_dal/src/fri_prover_dal.rs
@@ -70,21 +70,18 @@ impl FriProverDal<'_, '_> {
     pub async fn insert_prover_jobs(
         &mut self,
         batch_id: L1BatchId,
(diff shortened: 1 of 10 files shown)


Artemka374 (approved):
LGTM

PR #4322, opened 2025-07-17
fix(da_fetcher): use the same logic as in consistency_checker to defi…

dimazhornyk:
…ne a first batch

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/core/node/node_sync/src/data_availability_fetcher/mod.rs
+++ b/core/node/node_sync/src/data_availability_fetcher/mod.rs
@@ -7,7 +7,7 @@ use tokio::sync::watch;
 use zksync_da_client::{types::InclusionData, DataAvailabilityClient};
 use zksync_dal::{ConnectionPool, Core, CoreDal};
 use zksync_health_check::{Health, HealthStatus, HealthUpdater, ReactiveHealthCheck};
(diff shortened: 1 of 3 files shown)


core/node/node_sync/src/data_availability_fetcher/mod.rs
```diff
@@ -99,7 +102,49 @@ impl DataAvailabilityFetcher {
     /// to populate the necessary DA info for the consistency checker to verify L1 commitments.
     /// So there is no point in scanning batches that were already checked by the consistency
     /// checker, or batches that will be skipped by it.
-    async fn determine_first_batch_to_scan(&self) -> anyhow::Result<L1BatchNumber> {
+    async fn determine_first_batch_to_scan(
```
perekopskiy: This looks too hacky, code is duplicated, if it's changed in one place most likely we will forget about second place. At the very least you should share this code with consistency checker. Maybe it's also worth refactoring DataAvailabilityFetcher to be a subtask of ConsistencyChecker

dimazhornyk: fair, reused the consistency_checker's implementation here

perekopskiy (approved):

PR #4329, opened 2025-07-21
docs: Added guide for interop

mm-zk:
## What ❔

* Added advanced guide for interop - how to send a message, how to verify and how to do basic debugging.

### DIFF
--- a/docs/src/guides/advanced/19_interop_basics.md
+++ b/docs/src/guides/advanced/19_interop_basics.md
@@ -96,7 +96,8 @@ The command returns a proof, that will look like this:
 }
 ```
 
-getL2toL1LogProof can return different proof types depending on the interop mode specified. Currently, only "proof_based_gw" is supported.
+getL2toL1LogProof can return different proof types depending on the interop mode specified. Currently, only
+"proof_based_gw" is supported.
 
 The second argument (0) is the index of the message - if your transaction created multiple interop messages, you have to
 select the correct one. You can figure out the correct one, by looking at the event logs from the transaction. In our


docs/src/guides/advanced/19_interop_basics.md
```diff
@@ -0,0 +1,197 @@
...
+      "0xf84927dc03d95cc652990ba75874891ccc5a4d79a0e10a2ffdd238a34a39f828",
+      "0x2c21d37d09509f7ccb3d7d93c86ff4d7b29246113b0032dec9688c9e119d6872"
+    ],
+    "id": 0,
+    "root": "0xeb6861fa2f2fd60c93dba1343bf85e2312811fd23549b3df2b8017102430f086"
+  }
+}
+```
+
+getL2toL1LogProof can return different proof types - it is important to pass the "proof_based_gw" as proof type.
```
zkzoomer:
```suggestion
getL2toL1LogProof can return different proof types depending on the interop mode specified. Currently, only "proof_based_gw" is supported.
```
nit: mention interop mode

docs/src/guides/advanced/19_interop_basics.md
```diff
@@ -0,0 +1,197 @@
...
+While messages are usually sent by contracts, you can send one directly using:
+
+```
+cast send -r http://localhost:3050 0x0000000000000000000000000000000000008008 "sendToL1(bytes)" "0x1234" --private-key $PRIVATE_KEY
+```
+
+Here, `0x...8008` is the L1Messenger, a special system contract. Remember the transaction hash, as we'll use it in
+commands below.
+
+Next, retrieve the interop proof from the sequencer (replace the transaction hash with your own):
```
popzxc: It probably makes sense to mention that the proof might not be available right away (it took me a few iterations of method returning `null` until I got the proof).

docs/src/guides/advanced/19_interop_basics.md
```diff
@@ -0,0 +1,197 @@
...
+    --rpc-url http://localhost:3050
+```
+
+A successful verification will return a "true" value (or `0x00...01`), confirming that the InteropMessage was correctly
+received.
```
popzxc, line 123: I'm on `kl/interop-support` branch so some stuff may work differently there, but still.

It looks like this `cast call` returns `true` even if I specify wrong chain ID/batch number/transaction index/address/message. It only fails if the proof itself is incorrect. Is that expected?

mm-zk: That's a great question to @kelemeno .

I've just checked on main and it was working correctly (any modified parameter returned 0x0)

docs/src/guides/advanced/19_interop_basics.md
mm-zk: good point. added

PR #4334, opened 2025-07-22
feat: High Priority L2 transactions

Artemka374:
## What ❔

Add new mempool feature: high priority L2 transactions, that are prioritized before all the other transactions after a certain protocol upgrade is done. These transactions are defined by initiator account and protocol version after which they start to execute.

## Why ❔

These transactions are intended to be executed right after a protocol upgrade is done. Sometimes there are few that should be executed to complete the upgrade and they should be prioritized over all the others.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/core/lib/mempool/src/mempool_store.rs
+++ b/core/lib/mempool/src/mempool_store.rs
@@ -305,14 +305,16 @@ impl MempoolStore {
                     return constraint;
                 }
 
-                if let Some((score, constraint)) = self
-                    .high_priority_l2_transactions_per_account
-                    .get_mut(&tx.initiator_account())
-                    .expect("account is not available in mempool")
-                    .reset(tx)
-                {
-                    self.high_priority_l2_priority_queue.remove(&score);
-                    return constraint;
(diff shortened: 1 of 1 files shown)


core/lib/mempool/src/mempool_store.rs
```diff
@@ -120,7 +137,15 @@ impl MempoolStore {
...
         let account = transaction.initiator_account();
 
-        let metadata = match self.l2_transactions_per_account.entry(account) {
+        let mut txs_per_account = &mut self.l2_transactions_per_account;
+        let mut priority_queue = &mut self.l2_priority_queue;
+
+        if Some(transaction.initiator_account()) == self.high_priority_l2_tx_initiator {
+            priority_queue = &mut self.high_priority_l2_priority_queue;
+            txs_per_account = &mut self.high_priority_l2_transactions_per_account;
+        }
```
perekopskiy: nit: instead of declaring txs_per_account and priority_queue as mutable you can return pair of (txs_per_account, priority_queue) from if-else

perekopskiy (approved):

PR #4349, opened 2025-07-24
feat: read execution delay from ValidatorTimelock contract instead of config

bxpana:
## What ❔

Changes:
- Add executionDelay() call to multicall in EthTxAggregator
- Add execution_delay field to MulticallData struct
- Update Aggregator to accept execution delay as parameter instead of using config
- Remove l1_batch_min_age_before_execute from SenderConfig
- Add ValidatorTimelock contract definition with executionDelay function
- Add necessary Duration imports

Documentation updates:
- Update external node configuration docs to note execution delay is now contract-based
- Update ValidatorTimelock contract documentation to explain dynamic execution delay reading
- Update launch guide to mention execution delay is no longer configurable

## Why ❔

This change removes the dependency on YAML configuration for execution delay and instead reads it dynamically from the ValidatorTimelock contract via multicall. This makes the system more flexible and reduces configuration management overhead.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Operational changes
- Removes execution delay from config (`l1_batch_min_age_before_execute_seconds:`)

## Checklist


Deniallugo (approved):

core/node/eth_sender/src/aggregator.rs
```diff
@@ -402,11 +404,10 @@ impl Aggregator {
...
         last_sealed_l1_batch: L1BatchNumber,
         priority_tree_start_index: Option<usize>,
+        execution_delay: Duration,
     ) -> Result<Option<ExecuteBatches>, EthSenderError> {
-        let max_l1_batch_timestamp_millis = self
-            .config
-            .l1_batch_min_age_before_execute
-            .map(|age| unix_timestamp_ms() - age.as_millis() as u64);
+        let max_l1_batch_timestamp_millis =
+            Some(unix_timestamp_ms() - execution_delay.as_millis() as u64);
```
perekopskiy: Please add some safety margin. On L1 time is discrete and in worst case if you send time at `X`, it will be included in the block with timestamp `X - 12`. So the margin should be greater than 12 sec, I think 30 seconds would be ok.

perekopskiy: And apply margin only if execution_delay > 0

Deniallugo (approved):

perekopskiy (approved):

PR #4357, opened 2025-07-28
chore: Remove legacy prover code

Artemka374:
## What ❔

Remove bincode encoding for prover-related types(ones involved in both prover and core)
Remove API mode config option. 
So, after this PR it will be not possible to:
* Run proof data handler in server mode
* Run prover gateway in client mode
* Save/get files(like witness inputs/final proofs) from Object store which are encoded with bincode

## Why ❔

This code was left in codebase for the sake of backwards compatibility. Now, since these changes were introduced in protocol version 27, it is "safe" to remove them in version 29.

## Is this a breaking change?
- [x] Yes
- [ ] No

## Checklist


core/bin/zksync_server/src/node_builder.rs
```diff
@@ -329,11 +329,9 @@ impl MainNodeBuilder {
     }
 
     fn add_proof_data_handler_layer(mut self) -> anyhow::Result<Self> {
-        let gateway_config = try_load_config!(self.configs.prover_gateway);
         self.node.add_layer(ProofDataHandlerLayer::new(
             try_load_config!(self.configs.proof_data_handler_config),
             self.genesis_config.l2_chain_id,
-            gateway_config.api_mode,
```
EmilLuta, line 345: I think we released this in v28, no? If so, isn't it too early to deprecate in v29?

core/lib/prover_interface/src/inputs.rs
```diff
@@ -158,7 +151,7 @@ impl WitnessInputMerklePaths {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
-pub struct VMRunWitnessInputData<FM: FormatMarker = CBOR> {
+pub struct VMRunWitnessInputData {
```
EmilLuta, line 154: 100% agree to deprecate it, but we should add some sort of public notice somewhere. Should it go in a release tag or maybe in the README?

core/bin/zksync_server/src/node_builder.rs
Artemka374: it was part of prover cluster, so it's v27

core/lib/prover_interface/src/inputs.rs
Artemka374: we can try to add this info when we will be creating release. The main part is not to forget, but I've marked these changes as breaking

PR #4361, opened 2025-07-29
fix(api): Return back fee params backward compatibility for non gateway chains

Deniallugo:
## What ❔

Use default base token conversion params for preserving backward compatibility 

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/core/lib/types/src/fee_model.rs
+++ b/core/lib/types/src/fee_model.rs
@@ -306,7 +306,7 @@ impl FeeParamsV2 {
         FeeModelConfigV2 {
             minimal_l2_gas_price: self.convert_l1_to_base_token(
                 self.config.minimal_l2_gas_price,
-                self.conversion_ratio.l1,
+                self.conversion_ratio.l1_conversion_ratio(),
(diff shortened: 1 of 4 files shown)


core/lib/types/src/fee_model.rs
```diff
@@ -380,8 +386,8 @@ impl Default for ConversionRatio {
 /// The struct that represents the BaseToken<->ETH conversion ratio.
 #[derive(Debug, Clone, Copy, Serialize, Deserialize)]
 pub struct BaseTokenConversionRatio {
-    pub l1: ConversionRatio,
-    pub sl: ConversionRatio,
+    l1: Option<ConversionRatio>,
```
dimazhornyk, line 389: is it ever None? it seems this struct is always created with `Some(l1)`

Deniallugo: yes, except, the cases when we receive this struct from legacy api

dimazhornyk (approved):

PR #4363, opened 2025-07-29
fix(state-keeper): limit to max interop roots per batch

zkzoomer:
## What ❔

Includes two changes to state keeper:
1. Adds a check to limit the number of interop roots per batch to that set by the protocol version. 
2. Includes a batch seal criterion for interop roots to close batches when that limit is reached.

Note that these new checks are unlikely to ever trigger: evey new batch (for any chain) creates a new message root on GW, meaning that if we have a total of $n$ chains settling on gateway, and they seal batches at approximately the same rate, then each chain should have around $n$ interop roots per batch. This approximation does not hold for very low volume chains, as they would each settle much less often and thus would import more interop roots.

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/core/node/state_keeper/src/seal_criteria/criteria/interop_roots.rs
+++ b/core/node/state_keeper/src/seal_criteria/criteria/interop_roots.rs
@@ -1,4 +1,4 @@
-use zksync_multivm::utils::get_bootloader_max_msg_roots_in_batch;
+use zksync_multivm::utils::get_bootloader_max_interop_roots_in_batch;
 use zksync_types::ProtocolVersionId;
 
 use crate::seal_criteria::{SealCriterion, SealData, SealResolution, StateKeeperConfig};
@@ -19,7 +19,7 @@ impl SealCriterion for InteropRootsCriterion {
         protocol_version: ProtocolVersionId,
     ) -> SealResolution {
         let max_interop_roots_in_batch =
-            get_bootloader_max_msg_roots_in_batch(protocol_version.into());
+            get_bootloader_max_interop_roots_in_batch(protocol_version.into());
         assert!(
             interop_roots_count <= max_interop_roots_in_batch,
             "Interop roots count ({}) must be lower than the bootloader constant MAX_MSG_ROOTS_IN_BATCH={} for protocol version {}",
@@ -43,7 +43,7 @@ impl SealCriterion for InteropRootsCriterion {
         protocol_version: ProtocolVersionId,
     ) -> Option<f64> {
         let max_interop_roots_in_batch =
-            get_bootloader_max_msg_roots_in_batch(protocol_version.into());
+            get_bootloader_max_interop_roots_in_batch(protocol_version.into());
         Some((interop_roots_count as f64) / (max_interop_roots_in_batch as f64))
     }
 
@@ -61,7 +61,7 @@ mod tests {
         // Create an empty config and only setup fields relevant for the test.
         let config = StateKeeperConfig::for_tests();
         let max_interop_roots_in_batch =
-            get_bootloader_max_msg_roots_in_batch(ProtocolVersionId::latest().into());
+            get_bootloader_max_interop_roots_in_batch(ProtocolVersionId::latest().into());
 
         let criterion = InteropRootsCriterion;
 
--- a/core/node/state_keeper/src/updates/mod.rs
+++ b/core/node/state_keeper/src/updates/mod.rs
@@ -4,7 +4,7 @@ use zksync_contracts::BaseSystemContractsHashes;
 use zksync_multivm::{
     interface::{Call, FinishedL1Batch, VmExecutionMetrics, VmExecutionResultAndLogs},
     utils::{
-        get_batch_base_fee, get_bootloader_max_msg_roots_in_batch, get_max_batch_gas_limit,
+        get_batch_base_fee, get_bootloader_max_interop_roots_in_batch, get_max_batch_gas_limit,
         get_max_gas_per_pubdata_byte, StorageWritesDeduplicator,
     },
 };
@@ -267,7 +267,7 @@ impl UpdatesManager {
         );
         // We need to filter already applied interop roots, and take up to the batch limit set by the bootloader.
         // Because we seal L2 blocks in async manner, it's possible that database returns already applied interop roots.
-        let limit = get_bootloader_max_msg_roots_in_batch(self.protocol_version.into());
+        let limit = get_bootloader_max_interop_roots_in_batch(self.protocol_version.into());
         let new_interop_roots: Vec<_> = l2_block_params
             .interop_roots()
             .iter()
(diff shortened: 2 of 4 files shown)


kelemeno:
looks ok

core/lib/config/src/configs/chain.rs
```diff
@@ -64,6 +64,9 @@ pub struct StateKeeperConfig {
     /// The max number of slots for txs in a block before it should be sealed by the slots sealer.
     #[config(default_t = 8_192)]
     pub transaction_slots: usize,
+    /// The max number of interop roots in a batch before it should be sealed by the interop roots sealer.
+    #[config(default_t = 95)]
+    pub interop_roots: usize,
```
perekopskiy: is there any reason why we would want to limit number of interop roots further than the limit imposed by protocol? If not then let's remove config value and use `get_bootloader_max_msg_roots_in_batch` directly

core/node/state_keeper/src/keeper.rs
```diff
@@ -846,6 +846,7 @@ impl StateKeeperInner {
                     updates_manager.l1_batch_number().0,
                     updates_manager.pending_executed_transactions_len() + 1,
                     updates_manager.pending_l1_transactions_len() + is_tx_l1,
+                    updates_manager.committed_updates().interop_roots.len(),
```
perekopskiy: Using `committed_updates` is incorrect, you should add `pending_interop_roots` method taking into account pending l2 blocks and use it instead.

core/node/state_keeper/src/seal_criteria/criteria/interop_roots.rs
```diff
@@ -0,0 +1,89 @@
...
+            interop_roots_count <= max_interop_roots_in_batch,
+            "Interop roots count ({}) must be lower than the bootloader constant MAX_MSG_ROOTS_IN_BATCH={} for protocol version {}",
+            interop_roots_count, max_interop_roots_in_batch, protocol_version as u16
+        );
+
+        if interop_roots_count >= config.interop_roots {
+            SealResolution::IncludeAndSeal
+        } else {
+            SealResolution::NoSeal
+        }
```
perekopskiy: getting into the state with `interop_roots_count > config.interop_roots` seem to be unreachable because of the code in `set_next_l2_block_params` that contains `take`. But still let's do "if `interop_roots_count > config.interop_roots` then return `ExcludeAndSeal`" just in case

core/node/state_keeper/src/updates/mod.rs
```diff
@@ -277,6 +278,7 @@ impl UpdatesManager {
                         .iter()
                         .all(|block| !block.interop_roots.contains(root))
             })
+            .take(limit.saturating_sub(self.committed_updates.interop_roots.len()))
```
perekopskiy: Using `committed_updates` here is also incorrect, you should use pending len

core/lib/config/src/configs/chain.rs
zkzoomer: Resolved in 9703c0da15b3d7b1cd562b19203fd8e6ef86242b

core/node/state_keeper/src/keeper.rs
zkzoomer: Resolved in 856666f1ec026df9dc64ff34d9782c6ecbdd2989

core/node/state_keeper/src/seal_criteria/criteria/interop_roots.rs
zkzoomer: Resolved in f203e8411a9f56dc61e7ded3817c3eee3b078899

core/node/state_keeper/src/updates/mod.rs
zkzoomer: Resolved in 856666f1ec026df9dc64ff34d9782c6ecbdd2989

core/node/state_keeper/src/keeper.rs
```diff
@@ -869,6 +870,7 @@ impl StateKeeperInner {
         let capacities = self.sealer.capacity_filled(
             manager.pending_executed_transactions_len(),
             manager.pending_l1_transactions_len(),
+            manager.committed_updates().interop_roots.len(),
```
Deniallugo: why not pending?

core/node/state_keeper/src/seal_criteria/criteria/interop_roots.rs
```diff
@@ -0,0 +1,92 @@
...
+            interop_roots_count, max_interop_roots_in_batch, protocol_version as u16
+        );
+
+        if interop_roots_count > max_interop_roots_in_batch {
+            SealResolution::ExcludeAndSeal
+        } else if interop_roots_count == max_interop_roots_in_batch {
+            SealResolution::IncludeAndSeal
+        } else {
+            SealResolution::NoSeal
+        }
```
Deniallugo: How exclude and seal suppose to work in this circumstances? 
Exclude and seal is about transaction, not about l2 block or root @perekopskiy

core/node/state_keeper/src/keeper.rs
zkzoomer: Fixed in 876ab645d86cd3417e83274115e5492ffd0c5dc3

core/node/state_keeper/src/seal_criteria/criteria/interop_roots.rs
perekopskiy: 1. It should really be unreachable, see my comment https://github.com/matter-labs/zksync-era/pull/4363#discussion_r2241918497
2. Even if it's not unreachable it should work like this: state keeper starts new block with too many interop roots -> first tx of the block triggers ExcludeAndSeal -> batch is being sealed and this block becomes fictive and interop roots are cleared from there

core/node/state_keeper/src/seal_criteria/criteria/interop_roots.rs
```diff
@@ -0,0 +1,90 @@
...
+        _config: &StateKeeperConfig,
+        _tx_count: usize,
+        _l1_tx_count: usize,
+        interop_roots_count: usize,
+        _block_data: &SealData,
+        _tx_data: &SealData,
+        protocol_version: ProtocolVersionId,
+    ) -> SealResolution {
+        let max_interop_roots_in_batch =
+            get_bootloader_max_msg_roots_in_batch(protocol_version.into());
```
StanislavBreadless: does it make sense to rename the function to get_bootloader_max_interop_roots_in_batch?

zkzoomer: Good point, as that makes it more consistent with https://github.com/matter-labs/era-contracts/blob/draft-v29/system-contracts/bootloader/bootloader.yul#L125-L127, renamed in d7b9fa1bd7e71cc65d9d7646096b7ffc7ba6af9b. There are probably other parts in the codebase where it refers to interop roots as "message roots", as the naming changed during development. Should probably make an effort towards consistency.

Deniallugo (approved):
lgtm

PR #4403, opened 2025-08-19
feat(en): Verify priority txs

Deniallugo:
## What ❔

Verify L1txs during the EN sync

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


core/node/node_sync/src/external_io.rs
```diff
@@ -491,6 +534,153 @@ impl StateKeeperIO for ExternalIO {
...
+    ) -> Result<(), PriorityTransactionVerificationError>;
+}
+
+#[derive(Debug)]
+pub struct PriorityTxVerifierL1 {
+    l1_client: Box<DynClient<L1>>,
+    l1_diamond_proxy: Address,
+    new_priority_request_signature: H256,
+
+    pub new_priority_request_id_signature: H256,
```
perekopskiy: nit: does it have to be `pub`?

core/node/node_sync/src/external_io.rs
```diff
@@ -375,11 +389,40 @@ impl StateKeeperIO for ExternalIO {
...
+                let tx = Transaction::from(*tx);
+                let Some(protocol_version) = self.current_protocol_version else {
+                    return Ok(Some(tx));
+                };
+                if let Some(l1_tx) = &tx.l1_tx() {
+                    let max_attempts = 5;
+                    for i in 0..=max_attempts {
+                        if let Err(err) = self
+                            .txs_verifier
+                            .verify_transaction(l1_tx, protocol_version)
```
perekopskiy, line 401: It would be better if you verified all txs at block start in parallel. I'm ok if it's merged this way and done later as a follow-up in case it will be too slow

core/node/node_sync/src/external_io.rs
```diff
@@ -491,6 +534,153 @@ impl StateKeeperIO for ExternalIO {
...
+        let filter = FilterBuilder::default()
+            .address(vec![self.l1_diamond_proxy])
+            .topics(
+                Some(vec![self.new_priority_request_id_signature]),
+                Some(vec![u256_to_h256((*tx.serial_id()).into())]),
+                None,
+                None,
+            )
+            .from_block((tx.common_data.eth_block - 1).into())
+            .to_block((tx.common_data.eth_block + 1).into())
```
perekopskiy: eth_block is deprecated, it's set to 0 for JSON RPC API sync

core/node/node_sync/src/external_io.rs
Deniallugo: In this case there is a chance to process bad transaction

core/node/node_sync/src/external_io.rs
Deniallugo: do you have suggestion, how to make it better? 
We can for sure use just the maximum range, but it's not allowed for all chains

core/node/node_sync/src/external_io.rs
perekopskiy: why? I meant do not starting any block executing before validation passes

core/node/node_sync/src/external_io.rs
perekopskiy: tbh no idea how to make it work in this case

core/node/node_sync/src/external_io.rs
Deniallugo: you mean check all txs in a batch all at once, right?

core/node/node_sync/src/external_io.rs
Deniallugo: could we just keep sending eth_block?

core/node/node_sync/src/external_io.rs
perekopskiy: not batch but block, otherwise correct

core/node/node_sync/src/external_io.rs
perekopskiy: maybe, I don't remember why it was deprecated..

core/node/node_sync/src/external_io.rs
Deniallugo: let's make it as a follow up

core/node/node_sync/src/external_io.rs
Deniallugo: I just returned back the support from consensus

core/lib/dal/src/consensus/proto/mod.proto
```diff
@@ -36,8 +36,10 @@ message Payload {
   optional bytes operator_address = 7; // required; H160
   // Set for protocol_version < 25.
   repeated Transaction transactions = 8;
-  // Set for protocol_version >= 25.
+  // Set for protocol_version >= 25 and < 29.
   repeated TransactionV25 transactions_v25 = 12;
+  // Set for protocol_version >= 29.
+  repeated TransactionV29 transactions_v29 = 16;
```
perekopskiy, line 42: format cannot change mid protocol version so we must ensure all nodes run this code before v29 upgrade

core/node/node_sync/src/external_io.rs
perekopskiy: let's get rid of +-1, it's not needed, and in some cases may break the node e.g. if `tx.common_data.eth_block + 1` will be greater than latest block

perekopskiy (dismissed):

PR #4406, opened 2025-08-20
fix: sending blobs

Artemka374:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/core/node/fee_model/src/l1_gas_price/blob_base_fee_predictor.rs
+++ b/core/node/fee_model/src/l1_gas_price/blob_base_fee_predictor.rs
@@ -1,0 +1,134 @@
+use zksync_dal::{Connection, Core, CoreDal};
+use zksync_types::{
+    aggregated_operations::{AggregatedActionType, L1BatchAggregatedActionType},
+    web3::{BlockId, BlockNumber},
+    L1BatchNumber, U256,
+};
+
+use crate::l1_gas_price::GasAdjusterClient;
+
+pub(crate) async fn predict_blob_base_fee(
+    connection: &mut Connection<'_, Core>,
+    client: &GasAdjusterClient,
+    last_known_l1_gas_price: u64,
+) -> u64 {
+    let last_finalized_block = client
+        .inner
+        .block(BlockId::Number(BlockNumber::Finalized))
+        .await
+        .unwrap()
+        .unwrap();
+
+    let batch_stats = connection
+        .eth_sender_dal()
+        .get_eth_all_blocks_stat()
+        .await
+        .expect("Failed to get batch stats");
+    let mut last_l1_commited_batch: Option<L1BatchNumber> = None;
+
+    for (tx_type, block) in batch_stats.mined {
+        if tx_type == AggregatedActionType::L1Batch(L1BatchAggregatedActionType::Commit) {
+            last_l1_commited_batch = Some(block.into());
+        }
+    }
+
+    let last_l1_commited_batch = last_l1_commited_batch.expect("No committed batch found");
+
+    // todo: implement this
+    let total_blobs_to_send = 0;
+
+    let total_l1_blocks_for_these_blocks =
+        U256::from(chrono::Utc::now().timestamp()) - last_finalized_block.timestamp;
+
+    predict_blob_fee_cap(
+        total_blobs_to_send,
+        total_l1_blocks_for_these_blocks.as_u64(),
+        last_known_l1_gas_price,
+    )
+}
+
+// todo: revisit this
+const MIN_BASE_FEE_PER_BLOB_GAS: u64 = 1;
+const BLOB_BASE_FEE_UPDATE_FRACTION: u64 = 3_338_477;
+const BLOB_GAS_PER_BLOB: u64 = 131_072; // spec value
+const BLOB_GAS_PER_BLOCK_TARGET: u64 = 393_216; // e.g. 3 blobs * 131_072
+const CAPACITY_BLOBS_PER_BLOCK: u64 = 6; // spec max
+
+// todo: revisit this
+const SAFETY_BPS: u32 = 11000; // +10%
+const TIP: u64 = 0; // wei per blob gas
+
+fn predict_blob_fee_cap(
+    blobs_total: u64,
+    l1_blocks_total: u64,
+    last_known_l1_gas_price: u64,
+) -> u64 {
+    // --- invert median_fee -> excess_now ---
+    let mut excess = excess_from_fee(last_known_l1_gas_price);
+    let mut max_fee = last_known_l1_gas_price;
+
+    // --- planned path: spread blobs_total across blocks_total evenly ---
+    let q = blobs_total / l1_blocks_total;
+    let r = blobs_total % l1_blocks_total;
+    for i in 0..l1_blocks_total {
+        let served = q + if i < r { 1 } else { 0 };
+        excess = next_excess(excess, served.min(CAPACITY_BLOBS_PER_BLOCK));
+        let fee = base_fee_from_excess(excess);
+        if fee > max_fee {
+            max_fee = fee;
+        }
+    }
+
+    // --- return a safe cap = max_fee * (1 + safety) + tip ---
+    ((max_fee as u128) * (SAFETY_BPS as u128) / 10_000u128) as u64 + TIP
+}
+
+fn fake_exponential(factor: u64, numerator: u64, denominator: u64) -> u64 {
+    let mut i: u128 = 1;
+    let mut output: u128 = 0;
+    let mut accum: u128 = (factor as u128) * (denominator as u128);
+    let num = numerator as u128;
+    let den = denominator as u128;
+    while accum > 0 {
+        output += accum;
+        accum = (accum * num) / (den * i);
+        i += 1;
+    }
+    (output / den) as u64
+}
+
+fn base_fee_from_excess(excess: u64) -> u64 {
+    MIN_BASE_FEE_PER_BLOB_GAS.max(fake_exponential(
+        BLOB_BASE_FEE_UPDATE_FRACTION,
+        excess,
+        BLOB_GAS_PER_BLOCK_TARGET,
+    ))
+}
+
+fn excess_from_fee(target_fee: u64) -> u64 {
+    if target_fee <= MIN_BASE_FEE_PER_BLOB_GAS {
+        return 0;
+    }
+    // binary search
+    let mut hi: u64 = 1;
+    while base_fee_from_excess(hi) < target_fee && hi < u64::MAX / 2 {
+        hi *= 2;
+    }
+    let mut lo: u64 = 0;
+    while lo < hi {
+        let mid = lo + (hi - lo) / 2;
+        let f = base_fee_from_excess(mid);
+        if f < target_fee {
+            lo = mid + 1;
+        } else {
+            hi = mid;
+        }
+    }
+    lo
+}
+
+fn next_excess(excess: u64, blobs_served: u64) -> u64 {
+    excess
+        .saturating_add(blobs_served.saturating_mul(BLOB_GAS_PER_BLOB))
+        .saturating_sub(BLOB_GAS_PER_BLOCK_TARGET)
+}
--- a/core/Cargo.lock
+++ b/core/Cargo.lock
@@ -14543,6 +14543,7 @@ version = "29.1.0-non-semver-compat"
 dependencies = [
  "anyhow",
  "async-trait",
+ "chrono",
  "test-casing",
  "tokio",
  "tracing",
--- a/core/node/api_server/src/web3/backend_jsonrpsee/namespaces/zks.rs
+++ b/core/node/api_server/src/web3/backend_jsonrpsee/namespaces/zks.rs
@@ -134,7 +134,7 @@ impl ZksNamespaceServer for ZksNamespace {
     }
 
     async fn get_fee_params(&self) -> RpcResult<FeeParams> {
-        Ok(self.get_fee_params_impl())
+        Ok(self.get_fee_params_impl().await)
     }
 
     async fn get_batch_fee_input(&self) -> RpcResult<PubdataIndependentBatchFeeModelInput> {
--- a/core/node/api_server/src/web3/namespaces/zks.rs
+++ b/core/node/api_server/src/web3/namespaces/zks.rs
@@ -447,12 +447,13 @@ impl ZksNamespace {
     }
 
     #[tracing::instrument(skip(self))]
-    pub fn get_fee_params_impl(&self) -> FeeParams {
+    pub async fn get_fee_params_impl(&self) -> FeeParams {
         self.state
             .tx_sender
             .0
             .batch_fee_input_provider
             .get_fee_model_params()
+            .await
     }
 
     #[deprecated]
--- a/core/node/eth_proof_manager/src/node.rs
+++ b/core/node/eth_proof_manager/src/node.rs
@@ -9,7 +9,10 @@ use zksync_config::{
     GasAdjusterConfig,
 };
 use zksync_contracts::proof_manager_contract;
-use zksync_dal::node::{MasterPool, PoolResource};
+use zksync_dal::{
+    node::{MasterPool, PoolResource},
+    ConnectionPool, Core,
+};
 use zksync_eth_client::clients::{Client, DynClient, SigningClient, L2};
 use zksync_eth_signer::PrivateKeySigner;
 use zksync_node_fee_model::l1_gas_price::{GasAdjuster, GasAdjusterClient};
@@ -75,6 +78,7 @@ impl EthProofManagerLayer {
         l2_chain_id: L2ChainId,
         contracts: &ProofManagerContracts,
         owner_wallet: Wallet,
+        connection_pool: ConnectionPool<Core>,
     ) -> ProofManagerClient {
         let operator_private_key = owner_wallet.private_key().clone();
         let operator_address = operator_private_key.address();
@@ -96,6 +100,7 @@ impl EthProofManagerLayer {
                 self.gas_adjuster_config.clone(),
                 PubdataSendingMode::Custom,
                 L1BatchCommitmentMode::Rollup,
+                connection_pool,
             )
             .await
             .unwrap(),
@@ -148,6 +153,7 @@ impl WiringLayer for EthProofManagerLayer {
                     .eth_proof_manager
                     .clone()
                     .expect("Eth proof manager wallet is required"),
+                main_pool.clone(),
             )
             .await;
 
--- a/core/node/eth_sender/src/tester.rs
+++ b/core/node/eth_sender/src/tester.rs
@@ -256,6 +256,7 @@ impl EthSenderTester {
                 },
                 pubdata_sending_mode,
                 commitment_mode,
+                connection_pool.clone(),
             )
             .await
             .unwrap(),
--- a/core/node/fee_model/Cargo.toml
+++ b/core/node/fee_model/Cargo.toml
@@ -24,6 +24,7 @@ tokio = { workspace = true, features = ["time"] }
 anyhow.workspace = true
 async-trait.workspace = true
 tracing.workspace = true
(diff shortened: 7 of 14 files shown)


core/lib/dal/src/blocks_dal.rs
```diff
@@ -1143,6 +1144,17 @@ impl BlocksDal<'_, '_> {
...
+        let blobs_so_far = self
+            .storage
+            .blocks_dal()
+            .get_blobs_so_far(L1BatchNumber(header.number.0.saturating_sub(1)))
+            .await?
+            + pubdata_input
+                .clone()
+                .map(|input| input.len() as u64)
+                .unwrap_or(0)
+                .div_ceil(ZK_SYNC_BYTES_PER_BLOB as u64);
```
Deniallugo: why? I'd assume you need to save only current number of blobs, otherwise you are summarizing all blobs from 0 batch

core/node/fee_model/src/l1_gas_price/blob_base_fee_predictor.rs
```diff
@@ -0,0 +1,166 @@
...
+    let total_blobs_to_send = connection
+        .blocks_dal()
+        .get_blobs_so_far(last_sealed_batch)
+        .await
+        .expect("Failed to get blobs so far for last sealed batch")
+        - connection
+            .blocks_dal()
+            .get_blobs_so_far(last_l1_commited_batch)
+            .await
+            .expect("Failed to get blobs so far for last committed batch");
```
Deniallugo: Let's just use SELECT sum(blobs) FROM l1_batches WHERE number BETWEEN sealed AND comitted

core/lib/dal/src/blocks_dal.rs
Artemka374: this is actually just amount_of_blobs_before + amount of blobs in this batch. Not sure whether I understand your comment

Artemka374: we are not saving amount of blobs explicitly anywhere. We are just splitting the pubdata in chunks when sending, so that's what I'm doing here

Artemka374: ah, I get it now. I wonder about efficiency difference though

Deniallugo: it will be pretty small query, based on summarizing and we have index for numbers. so, the query will be fine and we at least will have a meaningful number in blobs column, because now it just starting to count blobs from a random place

core/node/fee_model/src/l1_gas_price/blob_base_fee_predictor.rs
```diff
@@ -0,0 +1,166 @@
...
+const CAPACITY_BLOBS_PER_BLOCK: u64 = 6; // spec max
+
+// todo: revisit this
+const SAFETY_BPS: u32 = 11000; // +10%
+const TIP: u64 = 0; // wei per blob gas
+
+fn predict_blob_fee_cap(
+    blobs_total: u64,
+    l1_blocks_total: u64,
+    last_known_l1_gas_price: u64,
```
StanislavBreadless: we should use `l1_blob_base_fee` to avoid confusion

core/node/fee_model/src/l1_gas_price/blob_base_fee_predictor.rs
```diff
@@ -0,0 +1,166 @@
...
+) -> u64 {
+    // --- invert median_fee -> excess_now ---
+    let mut excess = excess_from_fee(last_known_l1_gas_price);
+    let mut max_fee = last_known_l1_gas_price;
+
+    // --- planned path: spread blobs_total across blocks_total evenly ---
+    let q = blobs_total / l1_blocks_total;
+    let r = blobs_total % l1_blocks_total;
+    for i in 0..l1_blocks_total {
```
StanislavBreadless: maybe better to do it in O(1)? 

```
excess = max(0, total_blobs_to_publish - (blob_target * l1_blocks_total))
let fee = base_fee_from_excess(excess)
```

I think this will capture even best-case distribution anyway. It will also cover the case when the number of l1 blocks is too low to even accomodate this many blobs

core/node/fee_model/src/l1_gas_price/blob_base_fee_predictor.rs
```diff
@@ -0,0 +1,166 @@
...
+// todo: revisit this
+const MIN_BASE_FEE_PER_BLOB_GAS: u64 = 1;
+const BLOB_BASE_FEE_UPDATE_FRACTION: u64 = 3_338_477;
+const BLOB_GAS_PER_BLOB: u64 = 131_072; // spec value
+const BLOB_GAS_PER_BLOCK_TARGET: u64 = 393_216; // target number of blobs per block(Ethereum uses 3)
+const CAPACITY_BLOBS_PER_BLOCK: u64 = 6; // spec max
+
+// todo: revisit this
+const SAFETY_BPS: u32 = 11000; // +10%
+const TIP: u64 = 0; // wei per blob gas
```
StanislavBreadless: leave a comment on why it is zero

core/node/fee_model/src/l1_gas_price/blob_base_fee_predictor.rs
```diff
@@ -0,0 +1,166 @@
...
+        total_l1_blocks_for_these_blocks,
+        last_known_l1_gas_price,
+    )
+}
+
+// todo: revisit this
+const MIN_BASE_FEE_PER_BLOB_GAS: u64 = 1;
+const BLOB_BASE_FEE_UPDATE_FRACTION: u64 = 3_338_477;
+const BLOB_GAS_PER_BLOB: u64 = 131_072; // spec value
+const BLOB_GAS_PER_BLOCK_TARGET: u64 = 393_216; // target number of blobs per block(Ethereum uses 3)
```
StanislavBreadless: 6 starting from Pectra

core/node/fee_model/src/l1_gas_price/blob_base_fee_predictor.rs
```diff
@@ -0,0 +1,166 @@
...
+const CAPACITY_BLOBS_PER_BLOCK: u64 = 6; // spec max
```
StanislavBreadless: 9 starting from Pectra

core/node/fee_model/src/l1_gas_price/blob_base_fee_predictor.rs
```diff
@@ -0,0 +1,166 @@
...
+    let den = denominator as u128;
+    while accum > 0 {
+        output += accum;
+        accum = (accum * num) / (den * i);
+        i += 1;
+    }
+    (output / den) as u64
+}
+
+fn base_fee_from_excess(excess: u64) -> u64 {
```
StanislavBreadless, line 100: are you sure this is correct function? it differs greatly from what we have in this EIP:  https://github.com/ethereum/EIPs/blob/master/EIPS/eip-4844.md

in the eip we have the following:

```
def get_base_fee_per_blob_gas(header: Header) -> int:
    return fake_exponential(
        MIN_BASE_FEE_PER_BLOB_GAS,
        header.excess_blob_gas,
        BLOB_BASE_FEE_UPDATE_FRACTION
    )
```

core/node/fee_model/src/l1_gas_price/blob_base_fee_predictor.rs
```diff
@@ -0,0 +1,166 @@
...
+const SAFETY_BPS: u32 = 11000; // +10%
+const TIP: u64 = 0; // wei per blob gas
+
+fn predict_blob_fee_cap(
+    blobs_total: u64,
+    l1_blocks_total: u64,
+    last_known_l1_gas_price: u64,
+) -> u64 {
+    // --- invert median_fee -> excess_now ---
+    let mut excess = excess_from_fee(last_known_l1_gas_price);
```
StanislavBreadless: can we query it from the block info somehow? I dont like binary search here... but if JSON RPC for blocks does not expose anything like it, then okay

core/node/fee_model/src/l1_gas_price/blob_base_fee_predictor.rs
Artemka374: oh sorry, it was just some random misusing of constants, will change

core/lib/dal/Cargo.toml
```diff
@@ -60,6 +60,7 @@ strum = { workspace = true, features = ["derive"] }
 tracing.workspace = true
 chrono = { workspace = true, default-features = false, features = ["now"] }
 rayon.workspace = true
+kzg.workspace = true
```
Deniallugo: is it possible to avoid this huge dependency in dal?

core/node/fee_model/src/l1_gas_price/blob_base_fee_predictor.rs
```diff
@@ -0,0 +1,113 @@
...
+    web3::{BlockId, BlockNumber},
+    L1BatchNumber,
+};
+
+use crate::l1_gas_price::GasAdjusterClient;
+
+pub(crate) async fn predict_blob_base_fee(
+    connection: &mut Connection<'_, Core>,
+    client: &GasAdjusterClient,
+    last_known_l1_gas_price: u64,
```
StanislavBreadless: last known l1 blob fee/price probably

core/node/fee_model/src/l1_gas_price/blob_base_fee_predictor.rs
```diff
@@ -0,0 +1,113 @@
...
+    let last_sealed_batch = connection
+        .blocks_dal()
+        .get_sealed_l1_batch_number()
+        .await
+        .expect("Failed to get last sealed batch")
+        .unwrap_or(L1BatchNumber::from(0));
+
+    let total_blobs_to_send = connection
+        .blocks_dal()
+        .get_blobs_amount_for_range(last_l1_commited_batch + 1, last_sealed_batch)
```
StanislavBreadless: just in case, will it work of if last_l1_committed_batch == last_sealed_batch?

core/node/fee_model/src/l1_gas_price/blob_base_fee_predictor.rs
```diff
@@ -0,0 +1,113 @@
...
+        .expect("Latest block number is None");
+    let last_commited_block_number = connection
+        .eth_sender_dal()
+        .get_sent_at_block_for_commited_block(last_l1_commited_batch)
+        .await
+        .expect("Failed to get sent at block for commited block")
+        .unwrap_or(latest_block_number.as_u32());
+
+    let mut total_l1_blocks_for_these_blocks = latest_block_number
+        .saturating_sub(last_commited_block_number.into())
```
StanislavBreadless, line 51: > last_commited_block_number

may belong to either GW or L1, so please be careful here, the formula should be only used for L1

core/node/fee_model/src/l1_gas_price/blob_base_fee_predictor.rs
```diff
@@ -0,0 +1,113 @@
...
+    tracing::debug!("Predicting blob fee cap with params: blobs_total: {total_blobs_to_send}, l1_blocks_total: {total_l1_blocks_for_these_blocks}, last_known_l1_gas_price: {last_known_l1_gas_price}");
+
+    predict_blob_fee_cap(
+        total_blobs_to_send,
+        total_l1_blocks_for_these_blocks,
+        last_known_l1_gas_price,
+    )
+}
+
+// todo: revisit this
```
StanislavBreadless: why todo?

core/node/fee_model/src/l1_gas_price/blob_base_fee_predictor.rs
```diff
@@ -0,0 +1,113 @@
...
+const MIN_BASE_FEE_PER_BLOB_GAS: u64 = 1;
+const BLOB_BASE_FEE_UPDATE_FRACTION: u64 = 3_338_477;
+const BLOB_GAS_PER_BLOCK_TARGET: u64 = 786_432; // target number of blobs per block(Ethereum uses 6)
+
+const SAFETY_BPS: u32 = 11000; // +10%
+const TIP: u64 = 0; // wei per blob gas
+
+fn predict_blob_fee_cap(blobs_total: u64, l1_blocks_total: u64, l1_blob_base_fee: u64) -> u64 {
+    let excess = blobs_total.saturating_sub(BLOB_GAS_PER_BLOCK_TARGET * l1_blocks_total);
+    let max_fee = l1_blob_base_fee.max(base_fee_from_excess(excess));
```
StanislavBreadless: you should not be using max here, as e.g. if the additional costs from excess are medium and the current l1_blob_base_fee is around the same value as well, the effect of these additional blobs will not be applied (i.e. you should assume that the excess blobs are on top of the blobs that have been sent on L1). 

before to derive the "previous excess" you used binary search, I said that I dont like it, but I agreed that if there is no better way then so be it.

on another note, `base_fee_from_excess = e^(excess / blob_fee_fraction)`. If you are really keen on skipping the binary search, you can do:

`base_fee_from_excess_new = e^((excess_old + additional_excess) / blob_fee_fraction) = e^(excess_old/blob_fee_fraction) * e^(additional_excess/blob_fee_fraction)`

so you might as well multiply the two instead of using `max`

(note that the above formula only works as long as the factor is 1, you should modify it a bit to support arbitrary factors

core/node/fee_model/src/l1_gas_price/blob_base_fee_predictor.rs
```diff
@@ -0,0 +1,113 @@
...
+        .expect("Failed to get latest block")
+        .expect("Latest block is None")
+        .number
+        .expect("Latest block number is None");
+    let last_commited_block_number = connection
+        .eth_sender_dal()
+        .get_sent_at_block_for_commited_block(last_l1_commited_batch)
+        .await
+        .expect("Failed to get sent at block for commited block")
+        .unwrap_or(latest_block_number.as_u32());
```
Deniallugo: could we have 1 query for last committed batch and their sent at block?

core/lib/dal/Cargo.toml
Artemka374: I guess we can either create a separate constant(but it can produce conflicts if main one changed) or pass it in function directly

core/node/fee_model/src/l1_gas_price/blob_base_fee_predictor.rs
Artemka374: yes, it will just return 0

core/lib/dal/src/eth_sender_dal.rs
```diff
@@ -291,6 +291,45 @@ impl EthSenderDal<'_, '_> {
...
+    ) -> DalResult<(u32, u32)> {
+        // if no commited batch found, return first batch number and latest block number
+        let result = sqlx::query!(
+            r#"
+            SELECT number, sent_at_block
+            FROM l1_batches
+            INNER JOIN
+                eth_txs_history
+                ON l1_batches.eth_commit_tx_id = eth_txs_history.eth_tx_id
+            WHERE eth_txs_history.confirmed_at IS NOT NULL
```
Deniallugo: you have to check the finality status, not is confirmed

core/lib/dal/src/eth_sender_dal.rs
```diff
@@ -291,6 +291,45 @@ impl EthSenderDal<'_, '_> {
...
+    ) -> DalResult<(u32, u32)> {
+        // if no commited batch found, return first batch number and latest block number
+        let result = sqlx::query!(
+            r#"
+            SELECT number, sent_at_block
+            FROM l1_batches
+            INNER JOIN
+                eth_txs_history
+                ON l1_batches.eth_commit_tx_id = eth_txs_history.eth_tx_id
+            WHERE eth_txs_history.finality_status <> 'pending'
```
Deniallugo: i'd rather use finalized

Deniallugo (dismissed):

core/lib/dal/src/blocks_dal.rs
```diff
@@ -1116,6 +1116,7 @@ impl BlocksDal<'_, '_> {
         storage_refunds: &[u32],
         pubdata_costs: &[i32],
         predicted_circuits_by_type: CircuitStatistic, // predicted number of circuits for each circuit type
+        bytes_per_blob: u64,
```
StanislavBreadless, line 1119: this is always a constant, right?

core/lib/dal/src/eth_sender_dal.rs
```diff
@@ -291,6 +291,45 @@ impl EthSenderDal<'_, '_> {
         Ok(stats)
     }
 
+    /// This query returns the number and sent_at_block of the latest commited batch.
+    /// Query should be primary used for L1, not for gateway settlement layer.
+    ///
+    /// Although, here we are not checking if the batch is from gateway settlement layer,
```
StanislavBreadless: why? The chain might've just migrated to L1 and so the last batch has been committed to GW

core/node/fee_model/src/l1_gas_price/blob_base_fee_predictor.rs
```diff
@@ -0,0 +1,186 @@
...
+    let latest_block_number = client
+        .inner
+        .block(BlockId::Number(BlockNumber::Latest))
+        .await
+        .expect("Failed to get latest block")
+        .expect("Latest block is None")
+        .number
+        .expect("Latest block number is None");
+
+    let (last_l1_commited_batch, last_commited_block_number) = connection
```
StanislavBreadless, line 34: as said above, the returned `last_commited_block_number` may still belong to GW

core/lib/dal/src/blocks_dal.rs
Artemka374: yes, but it's coming from kzg crate, so we don't want dal to depend on it and we're passing it separately

core/lib/dal/src/eth_sender_dal.rs
```diff
@@ -291,6 +292,65 @@ impl EthSenderDal<'_, '_> {
...
+                        if notification == GatewayMigrationNotification::FromGateway {
+                            notification_block_number.0
+                        } else {
+                            latest_block_number
+                        }
+                    } else {
+                        latest_block_number
+                    }
+                };
+
```
Deniallugo: Could you please describe in a comment what is going on here?

Deniallugo (dismissed):

core/lib/dal/src/eth_sender_dal.rs
```diff
@@ -291,6 +292,66 @@ impl EthSenderDal<'_, '_> {
...
+                    if let Some((notification, notification_block_number)) = latest_notification {
+                        if notification == GatewayMigrationNotification::FromGateway {
+                            notification_block_number.0
+                        } else {
+                            panic!("Trying to get the latest commited batch, but there was no gateway migration GW -> L1");
+                        }
+                    } else {
+                        panic!("Trying to get the latest commited batch, but there was no gateway migration GW -> L1");
+                    }
```
Deniallugo:
```suggestion
                    if let Some((GatewayMigrationNotification::FromGateway, notification_block_number)) = latest_notification {
                        notification_block_number.0
                    } else {
                        panic!("Trying to get the latest commited batch, but there was no gateway migration GW -> L1");
                    }
```

core/lib/dal/src/eth_sender_dal.rs
```diff
@@ -291,6 +292,66 @@ impl EthSenderDal<'_, '_> {
...
+                        .storage
+                        .server_notifications_dal()
+                        .get_latest_gateway_migration_notification_and_block_number()
+                        .await?;
+
+                    if let Some((notification, notification_block_number)) = latest_notification {
+                        if notification == GatewayMigrationNotification::FromGateway {
+                            notification_block_number.0
+                        } else {
+                            panic!("Trying to get the latest commited batch, but there was no gateway migration GW -> L1");
```
StanislavBreadless: as discussed in private please remove panics and replace with error logs. even if some miscalculation is present it is still better than state keeper panicking

StanislavBreadless (approved):

PR #4418, opened 2025-08-22
feat: split zkstack cli command

Raid5594:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/zkstack_cli/crates/zkstack/src/admin_functions.rs
+++ b/zkstack_cli/crates/zkstack/src/admin_functions.rs
@@ -154,16 +154,21 @@ pub async fn make_permanent_rollup(
 pub async fn governance_execute_calls(
     shell: &Shell,
     ecosystem_config: &EcosystemConfig,
-    governor: &Wallet,
+    mode: AdminScriptMode,
     encoded_calls: Vec<u8>,
     forge_args: &ForgeScriptArgs,
     l1_rpc_url: String,
-) -> anyhow::Result<()> {
+    governance_address: Address,
+) -> anyhow::Result<AdminScriptOutput> {
     // resume doesn't properly work here.
     let mut forge_args = forge_args.clone();
     forge_args.resume = false;
 
-    let governance_address = ecosystem_config.get_contracts_config()?.l1.governance_addr;
+    let governance_address = if governance_address.is_zero() {
+        ecosystem_config.get_contracts_config()?.l1.governance_addr
+    } else {
+        governance_address
+    };
 
     let calldata = ADMIN_FUNCTIONS
         .encode(
@@ -181,7 +186,24 @@ pub async fn governance_execute_calls(
         .with_rpc_url(l1_rpc_url)
         .with_broadcast()
         .with_calldata(&calldata);
-    accept_ownership(shell, governor, forge).await
+
+    let description = "executing governance calls";
+    let (forge, spinner_text) = match mode {
+        AdminScriptMode::OnlySave => (forge, format!("Preparing calldata for {description}")),
+        AdminScriptMode::Broadcast(wallet) => {
+            let forge = forge.with_broadcast();
+            let forge = fill_forge_private_key(forge, Some(&wallet), WalletOwner::Governor)?;
+            check_the_balance(&forge).await?;
+            (forge, format!("Executing {description}"))
+        }
+    };
+
+    let spinner = Spinner::new(&spinner_text);
+    forge.run(shell)?;
+    spinner.finish();
+
+    let output_path = ACCEPT_GOVERNANCE_SCRIPT_PARAMS.output(&foundry_contracts_path);
+    Ok(AdminScriptOutputInner::read(shell, output_path)?.into())
 }
 
 #[allow(clippy::too_many_arguments)]
@@ -371,6 +393,7 @@ async fn accept_ownership(
     Ok(())
 }
 
+#[derive(Clone)]
 pub enum AdminScriptMode {
     OnlySave,
     Broadcast(Wallet),
--- a/zkstack_cli/crates/zkstack/src/commands/chain/gateway/convert_to_gateway.rs
+++ b/zkstack_cli/crates/zkstack/src/commands/chain/gateway/convert_to_gateway.rs
@@ -1,13 +1,17 @@
 use anyhow::Context;
+use clap::Parser;
 use ethers::{
     abi::{parse_abi, Address},
     contract::BaseContract,
+    types::U256,
     utils::hex,
 };
 use lazy_static::lazy_static;
+use serde::{Deserialize, Serialize};
 use xshell::Shell;
 use zkstack_cli_common::{
     config::global_config,
+    ethereum::get_ethers_provider,
     forge::{Forge, ForgeScriptArgs},
     wallets::Wallet,
 };
@@ -29,10 +33,12 @@ use zkstack_cli_config::{
 use zkstack_cli_types::ProverMode;
 
 use crate::{
+    abi::BridgehubAbi,
     admin_functions::{
         governance_execute_calls, grant_gateway_whitelist, revoke_gateway_whitelist,
         set_transaction_filterer, AdminScriptMode,
     },
+    commands::chain::utils::display_admin_script_output,
     consts::PATH_TO_GATEWAY_OVERRIDE_CONFIG,
     messages::MSG_CHAIN_NOT_INITIALIZED,
     utils::forge::{check_the_balance, fill_forge_private_key, WalletOwner},
@@ -41,11 +47,40 @@ use crate::{
 lazy_static! {
     static ref DEPLOY_GATEWAY_TX_FILTERER_ABI: BaseContract =
         BaseContract::from(parse_abi(&["function runWithInputFromFile() public"]).unwrap(),);
-    static ref GATEWAY_VOTE_PREPARATION_ABI: BaseContract =
-        BaseContract::from(parse_abi(&["function run() public"]).unwrap(),);
+    static ref GATEWAY_VOTE_PREPARATION_ABI: BaseContract = BaseContract::from(
+        parse_abi(&["function prepareForGWVoting(uint256 ctmChainId) public"]).unwrap(),
+    );
 }
 
-pub async fn run(args: ForgeScriptArgs, shell: &Shell) -> anyhow::Result<()> {
+#[derive(Debug, Serialize, Deserialize, Parser)]
+pub struct ConvertToGatewayArgs {
+    /// All ethereum environment related arguments
+    #[clap(flatten)]
+    #[serde(flatten)]
+    pub forge_args: ForgeScriptArgs,
+
+    /// Pass the bridgehub, if existing ecosystem is being used
+    #[clap(long)]
+    pub bridgehub_addr: Option<Address>,
+
+    /// Pass the chain id, for which we want to deploy ctm on GW
+    #[clap(long, value_parser = parse_decimal_u256)]
+    pub ctm_chain_id: Option<U256>,
+
+    #[clap(long, default_value_t = false)]
+    pub only_save_calldata: bool,
+}
+
+fn parse_decimal_u256(s: &str) -> Result<U256, String> {
+    if s.starts_with("0x") || s.starts_with("0X") {
+        return Err("Hexadecimal format not allowed for ctm_chain_id. Use a decimal value.".into());
+    }
+    U256::from_dec_str(s).map_err(|e| format!("Invalid decimal U256: {e}"))
+}
+
+pub async fn run(convert_to_gw_args: ConvertToGatewayArgs, shell: &Shell) -> anyhow::Result<()> {
+    let args = convert_to_gw_args.forge_args;
+    let bridgehub_address = convert_to_gw_args.bridgehub_addr.unwrap_or(Address::zero());
     let chain_name = global_config().chain_name.clone();
     let ecosystem_config = EcosystemConfig::from_file(shell)?;
     let chain_config = ecosystem_config
@@ -68,53 +103,71 @@ pub async fn run(args: ForgeScriptArgs, shell: &Shell) -> anyhow::Result<()> {
         .deployer
         .context("deployer")?;
 
-    let output = deploy_gateway_tx_filterer(
-        shell,
-        args.clone(),
-        &ecosystem_config,
-        &chain_config,
-        &chain_deployer_wallet,
-        GatewayTxFiltererInput::new(
-            &ecosystem_config.get_initial_deployment_config().unwrap(),
-            &chain_contracts_config,
-        )?,
-        l1_url.clone(),
-    )
-    .await?;
+    let grantees;
 
-    set_transaction_filterer(
+    let bridgehub_governance_addr = if bridgehub_address.is_zero() {
+        let output: GatewayTxFiltererOutput = deploy_gateway_tx_filterer(
+            shell,
+            args.clone(),
+            &ecosystem_config,
+            &chain_config,
+            &chain_deployer_wallet,
+            GatewayTxFiltererInput::new(
+                &ecosystem_config.get_initial_deployment_config().unwrap(),
+                &chain_contracts_config,
+            )?,
+            l1_url.clone(),
+        )
+        .await?;
+
+        set_transaction_filterer(
+            shell,
+            &args,
+            &chain_config.path_to_l1_foundry(),
+            AdminScriptMode::Broadcast(chain_config.get_wallets_config()?.governor),
+            chain_config.chain_id.as_u64(),
+            chain_contracts_config
+                .ecosystem_contracts
+                .bridgehub_proxy_addr,
+            output.gateway_tx_filterer_proxy,
+            l1_url.clone(),
+        )
+        .await?;
+
+        chain_contracts_config.set_transaction_filterer(output.gateway_tx_filterer_proxy);
+        chain_contracts_config.save_with_base_path(shell, chain_config.configs.clone())?;
+
+        grantees = vec![
+            ecosystem_config.get_contracts_config()?.l1.governance_addr,
+            chain_deployer_wallet.address,
+            chain_contracts_config
+                .ecosystem_contracts
+                .stm_deployment_tracker_proxy_addr
+                .context("No CTM deployment tracker")?,
+        ];
+
+        Address::zero()
+    } else {
+        let l1_provider = get_ethers_provider(&l1_url)?;
+        let l1_bridgehub = BridgehubAbi::new(bridgehub_address, l1_provider);
+        let bridgehub_governance_addr = l1_bridgehub.owner().await?;
+
+        grantees = vec![bridgehub_governance_addr, chain_deployer_wallet.address];
+
+        bridgehub_governance_addr
+    };
+
+    let mode = if convert_to_gw_args.only_save_calldata {
+        AdminScriptMode::OnlySave
+    } else {
+        AdminScriptMode::Broadcast(chain_config.get_wallets_config()?.governor)
+    };
+
+    let mut output = grant_gateway_whitelist(
         shell,
         &args,
         &chain_config.path_to_l1_foundry(),
-        AdminScriptMode::Broadcast(chain_config.get_wallets_config()?.governor),
-        chain_config.chain_id.as_u64(),
-        chain_contracts_config
-            .ecosystem_contracts
-            .bridgehub_proxy_addr,
-        output.gateway_tx_filterer_proxy,
-        l1_url.clone(),
-    )
-    .await?;
-
-    chain_contracts_config.set_transaction_filterer(output.gateway_tx_filterer_proxy);
-    chain_contracts_config.save_with_base_path(shell, chain_config.configs.clone())?;
-
-    let grantees = vec![
-        ecosystem_config.get_contracts_config()?.l1.governance_addr,
-        chain_deployer_wallet.address,
-        // This is only for local deployment, to allow easier future testing
-        ecosystem_config.get_wallets()?.deployer.unwrap().address,
-        chain_contracts_config
-            .ecosystem_contracts
-            .stm_deployment_tracker_proxy_addr
-            .context("No CTM deployment tracker")?,
-    ];
-
-    grant_gateway_whitelist(
-        shell,
-        &args,
-        &chain_config.path_to_l1_foundry(),
-        AdminScriptMode::Broadcast(chain_config.get_wallets_config()?.governor),
+        mode.clone(),
         chain_config.chain_id.as_u64(),
         chain_contracts_config
             .ecosystem_contracts
@@ -123,6 +176,10 @@ pub async fn run(args: ForgeScriptArgs, shell: &Shell) -> anyhow::Result<()> {
         l1_url.clone(),
     )
     .await?;
+
+    if convert_to_gw_args.only_save_calldata {
+        display_admin_script_output(output);
+    }
 
     let vote_preparation_output = gateway_vote_preparation(
         shell,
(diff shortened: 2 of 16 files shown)


zkstack_cli/README.md
```diff
@@ -96,6 +96,41 @@ zkstack ecosystem init --dev
 If the process gets stuck, resume it with `--resume`. This flag keeps track of already sent transactions and sends new
 ones with provided params.
 
+If the ecosystem has already been initialized, you can create a new ctm and rigester it in the existing bridgehub.
+
+This command create a new ctm and all related contracts:
+
+```bash
+zkstack ecosystem init-new-ecosystem --l1-rpc-url <L1_RPC_ENDPOINT> --bridgehub <BRIDGEHUB_CONTRACT_ADDRESS>
```
StanislavBreadless: let's use `ecosystem` terminology as a synonym for "a single bridgehub with shared bridge (asset trackers)". 

So in this case, we dont init new ecosystem, we just append a new CTM to the existing one

zkstack_cli/README.md
```diff
@@ -96,6 +96,41 @@ zkstack ecosystem init --dev
...
+
+#### Gateway
+
+If the want to convert the newly created gateway chain to become the settlement layer for local testing:
+
+```bash
+zkstack chain gateway convert-to-gateway --chain <GATEWAY_CHAIN_NAME>
+```
+
+If we want to deploy a ctw on gateway for an existing settlement layer, please provide bridgehub address and id of the
```
StanislavBreadless:
```suggestion
If we want to deploy a CTM on gateway for an existing settlement layer, please provide bridgehub address and id of the
```

zkstack_cli/crates/zkstack/src/commands/chain/gateway/convert_to_gateway.rs
```diff
@@ -68,53 +103,71 @@ pub async fn run(args: ForgeScriptArgs, shell: &Shell) -> anyhow::Result<()> {
...
-        AdminScriptMode::Broadcast(chain_config.get_wallets_config()?.governor),
-        chain_config.chain_id.as_u64(),
-        chain_contracts_config
-            .ecosystem_contracts
-            .bridgehub_proxy_addr,
-        output.gateway_tx_filterer_proxy,
-        l1_url.clone(),
-    )
-    .await?;
+    let bridgehub_governance_addr = if bridgehub_address.is_zero() {
```
StanislavBreadless: could you please separate these into two commands? `convert to gateway` is responsible for converting a chain to GW, we should have a separate command responsible for CTM registration only

zkstack_cli/crates/zkstack/src/commands/ecosystem/args/init.rs
```diff
@@ -136,9 +143,39 @@ pub struct EcosystemInitArgs {
     pub skip_contract_compilation_override: bool,
     #[clap(long, help = MSG_SERVER_COMMAND_HELP)]
     pub server_command: Option<String>,
+    #[clap(long, help = MSG_BRIGEHUB)]
+    pub bridgehub: Option<String>,
 }
 
 impl EcosystemInitArgs {
+    pub fn resolve_deploy_erc20(dev: bool, deploy_erc20: Option<bool>, prompt: bool) -> bool {
```
StanislavBreadless: why do we have these changes in this PR?

zkstack_cli/crates/zkstack/src/commands/ecosystem/init_new_ctm.rs
```diff
@@ -0,0 +1,341 @@
...
+            &config.get_wallets()?.governor,
+            contracts_config
+                .ecosystem_contracts
+                .state_transition_proxy_addr,
+            &forge_args,
+            l1_rpc_url.clone(),
+        )
+        .await?;
+
+        println!("One");
```
StanislavBreadless: fyi debug logs

zkstack_cli/crates/zkstack/src/commands/ecosystem/init_new_ctm.rs
```diff
@@ -0,0 +1,341 @@
...
+        &l1_rpc_url,
+        None,
+        true,
+        support_l2_legacy_shared_bridge_test,
+        bridgehub_address,
+    )
+    .await?;
+    spinner.finish();
+
+    if bridgehub_address == H160::zero() {
```
StanislavBreadless: unless it is very hard, please try not mix "ecosystem init" with "CTM init", the command is all about CTM, while having lots of ifs for `bridgehub == 0` (which on its own is also not very explicit about the intention, bridgehub is actually never zero, it is just a way to tell the code that we should init bridgehub/rest of the ecosystem too). 

So:
- Please separate the two functionality 
- If the above is hard, please use explicit terms (like `first_ecosystem_init` instead of using `bridgehub == 0`)

zkstack_cli/crates/zkstack/src/commands/ecosystem/init_new_ctm.rs
```diff
@@ -0,0 +1,341 @@
...
+                .stm_deployment_tracker_proxy_addr
+                .context("stm_deployment_tracker_proxy_addr")?,
+            &forge_args,
+            l1_rpc_url.clone(),
+        )
+        .await?;
+
+        println!("Six");
+    } else {
+        accept_owner(
```
StanislavBreadless: this part of code is repeated, please avoid repetition

zkstack_cli/crates/zkstack/src/admin_functions.rs
```diff
@@ -181,7 +186,24 @@ pub async fn governance_execute_calls(
         .with_rpc_url(l1_rpc_url)
         .with_broadcast()
```
StanislavBreadless: note that we always do broadcast even in the save mode

zkstack_cli/crates/zkstack/src/commands/ecosystem/common.rs
```diff
@@ -80,3 +181,150 @@ pub async fn deploy_l1(
...
+        // It's a kludge for reth, just because it doesn't behave properly with large amount of txs
+        forge = forge.with_slow();
+    }
+
+    if let Some(address) = sender {
+        forge = forge.with_sender(address);
+    } else {
+        // forge =
+        // fill_forge_private_key(forge, Some(&wallets_config.governor), WalletOwner::Governor)?;
+        forge = fill_forge_private_key(
```
StanislavBreadless: could you please add a comment why we need deployer wallet when usually only governor can register new ctms

zkstack_cli/crates/zkstack/src/commands/chain/gateway/convert_to_gateway.rs
```diff
@@ -148,40 +205,46 @@ pub async fn run(args: ForgeScriptArgs, shell: &Shell) -> anyhow::Result<()> {
             Address::zero(),
         ),
         l1_url.clone(),
+        convert_to_gw_args.ctm_chain_id.unwrap_or_default(),
```
StanislavBreadless: is it safe to use 0 here?

zkstack_cli/crates/zkstack/src/commands/ecosystem/args/init.rs
```diff
@@ -136,9 +143,39 @@ pub struct EcosystemInitArgs {
     pub skip_contract_compilation_override: bool,
     #[clap(long, help = MSG_SERVER_COMMAND_HELP)]
     pub server_command: Option<String>,
+    #[clap(long, help = MSG_BRIGEHUB)]
```
StanislavBreadless: typo should be `MSG_BRIDGEHUB`

zkstack_cli/crates/zkstack/src/commands/ecosystem/args/init.rs
```diff
@@ -152,43 +189,52 @@ impl EcosystemInitArgs {
...
+        let ecosystem = ecosystem
+            .fill_values_with_prompt(l1_network, dev || prompt_policy.skip_ecosystem)
             .await?;
-        let observability = if self.dev {
-            true
+
+        let bridgehub_address: H160 = if let Some(ref addr_str) = bridgehub {
+            addr_str
+                .parse::<H160>()
+                .expect("Invalid bridgehub address format")
```
StanislavBreadless: nit: better to return an error and not panic upon invalid input

zkstack cli does not display panic reason nicely

zkstack_cli/README.md
StanislavBreadless: `init-new-ecosystem` command does not exist

Raid5594: It's a readme error, command is already init-new-ctm, apologies

zkstack_cli/crates/zkstack/src/commands/ecosystem/args/init.rs
Raid5594: I don't understand the question really. this is related to the prompt policy.
If you're against the approach, please offer another solution

zkstack_cli/crates/zkstack/src/commands/chain/gateway/convert_to_gateway.rs
Raid5594: So then we want to re-write CI and call two commands consecutively? 
Or we keep the convert-to-gateway that calls internally `set-gw-tx-filterer` and then calls `gateway-vote-preparation`? 
It seems like extra work to separate the set-tx-filterer

zkstack_cli/crates/zkstack/src/commands/ecosystem/init_new_ctm.rs
Raid5594: if it's about ctm init, why would we name it first_ecosystem_init?
also, by saying `hard`, what do you exactly mean? 

The reality is much simpler, a simple comment can explain that passing no bridgehub assumes registration of a new ecosystem, so we have to update ownership for more than 2 contracts.
Otherwise we can set some variable, which would replace the comment and be self-explanatory and call 2 respective self-explanatory functions underneath to avoid the if statement if the current implementation seems too convoluted 

FYI, we're using bh=zero twice in rust part of code, if I didn't miss anything, potentially more in the contracts script, so I can introduce self-explanatory variables there as well

zkstack_cli/README.md
Deniallugo: i won't mix ecosystem and CTM at all, i'd rather create  a new highlevel command ctm for such cases

zkstack_cli/crates/zkstack/src/commands/chain/gateway/convert_to_gateway.rs
```diff
@@ -41,11 +47,40 @@ use crate::{
...
+#[derive(Debug, Serialize, Deserialize, Parser)]
+pub struct ConvertToGatewayArgs {
+    /// All ethereum environment related arguments
+    #[clap(flatten)]
+    #[serde(flatten)]
+    pub forge_args: ForgeScriptArgs,
+
+    /// Pass the bridgehub, if existing ecosystem is being used
+    #[clap(long)]
+    pub bridgehub_addr: Option<Address>,
```
Deniallugo, line 59: we have a bridgehub inside ecosystem. you have to use it from there, if non is provided

zkstack_cli/crates/zkstack/src/commands/chain/gateway/convert_to_gateway.rs
Deniallugo: why deploy tx filter is when bridgehub is zero?

zkstack_cli/crates/zkstack/src/commands/ecosystem/mod.rs
```diff
@@ -27,21 +30,30 @@ pub enum EcosystemCommands {
     /// Initialize ecosystem and chain,
     /// deploying necessary contracts and performing on-chain operations
     Init(EcosystemInitArgs),
+    /// Initialize new ecosystem on existing bridgehub
```
Deniallugo, line 35: how is it possible to have the same args?

zkstack_cli/crates/zkstack/src/commands/ecosystem/init_new_ctm.rs
Deniallugo: No passing an argument means to deploy somthing? 
it's non obvious, just split two processes.

zkstack_cli/crates/zkstack/src/commands/chain/gateway/convert_to_gateway.rs
Raid5594: yes, we'd use gateway_chain_id

StanislavBreadless: maybe it would be better to provide gw chain id then? i.e. instead of zero just use the gw chain id as it is available in this script anyway

zkstack_cli/crates/zkstack/src/commands/ecosystem/init_core_contracts.rs
```diff
@@ -0,0 +1,204 @@
...
+        ecosystem_config,
+        initial_deployment_config,
+        init_args.support_l2_legacy_shared_bridge_test,
+    )
+    .await?;
+    contracts.save_with_base_path(shell, &ecosystem_config.config)?;
+    Ok(contracts)
+}
+
+async fn deploy_ecosystem(
```
StanislavBreadless: we probably dont need the inner function then since this one is so small

zkstack_cli/crates/zkstack/src/admin_functions.rs
```diff
@@ -154,16 +154,21 @@ pub async fn make_permanent_rollup(
 pub async fn governance_execute_calls(
     shell: &Shell,
     ecosystem_config: &EcosystemConfig,
-    governor: &Wallet,
+    mode: AdminScriptMode,
     encoded_calls: Vec<u8>,
     forge_args: &ForgeScriptArgs,
     l1_rpc_url: String,
-) -> anyhow::Result<()> {
+    governance_address: Address,
```
Deniallugo: Please use options, instead of zeros

zkstack_cli/crates/zkstack/src/commands/chain/gateway/convert_to_gateway.rs
```diff
@@ -68,53 +97,40 @@ pub async fn run(args: ForgeScriptArgs, shell: &Shell) -> anyhow::Result<()> {
...
-            .ecosystem_contracts
-            .bridgehub_proxy_addr,
-        output.gateway_tx_filterer_proxy,
-        l1_url.clone(),
-    )
-    .await?;
-
-    chain_contracts_config.set_transaction_filterer(output.gateway_tx_filterer_proxy);
-    chain_contracts_config.save_with_base_path(shell, chain_config.configs.clone())?;
+    let grantees;
```
Deniallugo: you have to use  tuple `let (grantees, bridgehub_governance_addr) = ... `

zkstack_cli/crates/zkstack/src/commands/chain/gateway/create_tx_filterer.rs
```diff
@@ -0,0 +1,124 @@
...
+        .context(MSG_CHAIN_NOT_INITIALIZED)?;
+    let l1_url = chain_config.get_secrets_config().await?.l1_rpc_url()?;
+    let mut chain_contracts_config = chain_config.get_contracts_config()?;
+    override_config(
+        shell,
+        ecosystem_config
+            .link_to_code
+            .join(PATH_TO_GATEWAY_OVERRIDE_CONFIG),
+        &chain_config,
+    )?;
```
Deniallugo: why?

zkstack_cli/crates/zkstack/src/commands/ecosystem/args/init.rs
```diff
@@ -13,13 +14,20 @@ use crate::{
...
         MSG_SERVER_COMMAND_HELP, MSG_SERVER_DB_NAME_HELP, MSG_SERVER_DB_URL_HELP,
     },
 };
 
+#[derive(Clone, Copy, Default)]
+pub struct PromptPolicy {
+    pub deploy_erc20: bool,
+    pub observability: bool,
+    pub skip_ecosystem: bool,
+}
```
Deniallugo: what does it mean?

zkstack_cli/crates/zkstack/src/commands/ecosystem/args/init.rs
```diff
@@ -205,4 +251,219 @@ pub struct EcosystemInitArgsFinal {
...
+#[derive(Debug, Clone, Serialize, Deserialize, Parser)]
+pub struct RegisterCTMArgs {
+    #[clap(flatten)]
+    #[serde(flatten)]
+    pub ecosystem: EcosystemArgs,
+    #[clap(flatten)]
+    #[serde(flatten)]
+    pub forge_args: ForgeScriptArgs,
+    #[clap(long)]
+    pub update_submodules: Option<bool>,
```
Deniallugo, line 245: could you add `default_missing_value = "true"` ?

Deniallugo: and are you sure, that non updating submodules should be a default behaviour?

zkstack_cli/crates/zkstack/src/commands/ecosystem/args/init.rs
```diff
@@ -205,4 +251,219 @@ pub struct EcosystemInitArgsFinal {
...
+
+impl From<&EcosystemInitArgsFinal> for RegisterCTMArgsFinal {
+    fn from(args: &EcosystemInitArgsFinal) -> Self {
+        RegisterCTMArgsFinal {
+            ecosystem: args.ecosystem.clone(),
+            forge_args: args.forge_args.clone(),
+            update_submodules: None,
+        }
+    }
+}
```
Deniallugo: if it's clone inside you shouldn't use reference in `from`

zkstack_cli/crates/zkstack/src/commands/ecosystem/args/init.rs
```diff
@@ -205,4 +251,219 @@ pub struct EcosystemInitArgsFinal {
...
+            .fill_values_with_prompt(l1_network, prompt_policy.skip_ecosystem)
+            .await?;
+
+        // Parse bridgehub address
+        let bridgehub_address = if let Some(ref addr_str) = bridgehub {
+            addr_str
+                .parse::<H160>()
+                .with_context(|| format!("Invalid bridgehub address format: {}", addr_str))?
+        } else {
+            H160::zero()
```
Deniallugo: is it safe to use `zero`?

zkstack_cli/crates/zkstack/src/commands/ecosystem/args/init.rs
```diff
@@ -205,4 +251,219 @@ pub struct EcosystemInitArgsFinal {
...
+pub struct InitNewCTMArgsFinal {
+    pub ecosystem: EcosystemArgsFinal,
+    pub forge_args: ForgeScriptArgs,
+    pub update_submodules: Option<bool>,
+    pub skip_contract_compilation_override: bool,
+    pub support_l2_legacy_shared_bridge_test: bool,
+    pub bridgehub_address: H160,
+}
+
+impl From<&EcosystemInitArgsFinal> for InitNewCTMArgsFinal {
```
Deniallugo: same as above

zkstack_cli/crates/zkstack/src/commands/ecosystem/common.rs
```diff
@@ -80,3 +111,215 @@ pub async fn deploy_l1(
...
+        shell,
+        DEPLOY_ECOSYSTEM_SCRIPT_PARAMS.output(&config.path_to_l1_foundry()),
+    )?;
+    let mut contracts_config = ContractsConfig::default();
+    contracts_config.update_from_l1_output(&script_output);
+
+    Ok(contracts_config)
+}
+
+#[allow(clippy::too_many_arguments)]
```
Deniallugo: it's not too many

zkstack_cli/crates/zkstack/src/commands/ecosystem/init.rs
```diff
@@ -121,64 +119,35 @@ async fn init_ecosystem(
     }
     spinner.finish();
 
-    let contracts = deploy_ecosystem(
+    deploy_ecosystem(
```
Deniallugo: hm, this method returns contracts. and you don't use it? have you saved everything between this call and the next call?

zkstack_cli/crates/zkstack/src/commands/chain/gateway/create_tx_filterer.rs
Raid5594: Not sure, that's how it was in convert-to-gw

zkstack_cli/crates/zkstack/src/commands/ecosystem/args/init.rs
Raid5594: this means that we can choose which prompts to display to the user instead of copy/pasting the code each time

zkstack_cli/crates/zkstack/src/commands/ecosystem/args/init.rs
Raid5594: `and are you sure, that non updating submodules should be a default behaviour?`

If we're calling init with update_submodules it would do it once, do we want to do it another 3 times when calling subcommands?

zkstack_cli/crates/zkstack/src/commands/ecosystem/args/init.rs
Raid5594: yes

zkstack_cli/crates/zkstack/src/commands/ecosystem/args/init.rs
```diff
@@ -152,48 +178,68 @@ impl EcosystemInitArgs {
...
+            support_l2_legacy_shared_bridge_test,
+            bridgehub,
+            ..
+        } = self;
+
+        let deploy_ecosystem = ecosystem.deploy_ecosystem.unwrap_or_else(|| {
+            if dev {
+                true
+            } else {
+                PromptConfirm::new(MSG_DEPLOY_ECOSYSTEM_PROMPT)
```
StanislavBreadless, line 186: here it looks like we disregard the prompt policy struct and decide to ask for prompt anyway, this is an example why this struct may be confusing a bit.

zkstack_cli/crates/zkstack/src/commands/ecosystem/args/init.rs
```diff
@@ -152,48 +178,68 @@ impl EcosystemInitArgs {
...
-            })
-        };
-        let ecosystem = self
-            .ecosystem
-            .fill_values_with_prompt(l1_network, self.dev)
+            }
+        });
+
+        let ecosystem = ecosystem
+            .fill_values_with_prompt(l1_network, dev || prompt_policy.skip_ecosystem)
```
StanislavBreadless: why?

zkstack_cli/crates/zkstack/src/commands/ecosystem/init_new_ctm.rs
```diff
@@ -0,0 +1,167 @@
...
+pub async fn deploy_new_ctm(
+    shell: &Shell,
+    ecosystem: &mut EcosystemArgsFinal,
+    forge_args: ForgeScriptArgs,
+    ecosystem_config: &EcosystemConfig,
+    initial_deployment_config: &InitialDeploymentConfig,
+    support_l2_legacy_shared_bridge_test: bool,
+    bridgehub_address: H160,
+) -> anyhow::Result<ContractsConfig> {
+    return deploy_new_ctm_inner(
```
StanislavBreadless: please inline the inner function

zkstack_cli/crates/zkstack/src/commands/ecosystem/register_ctm.rs
```diff
@@ -0,0 +1,70 @@
...
+    Ok(())
+}
+
+pub async fn register_ctm(
+    init_args: &mut RegisterCTMArgsFinal,
+    shell: &Shell,
+    forge_args: ForgeScriptArgs,
+    ecosystem_config: &EcosystemConfig,
+) -> anyhow::Result<()> {
+    register_ctm_inner_inner(
```
StanislavBreadless: please inline the inner function

zkstack_cli/crates/zkstack/src/commands/ecosystem/args/init.rs
StanislavBreadless: @Raid5594 maybe it is better to use `Option`? To clearly denote that this address may be invalid

StanislavBreadless: If Solidity scripts dont accept Option, I would:
- consider trying to always provide the correct value for them
- do `unwrap_or_default()` in the exact place where we call the Solidity scripts and add a comment on why it is okay

zkstack_cli/crates/zkstack/src/commands/ecosystem/args/init.rs
Raid5594: This is a mistake, I should have added it to resolve function as well

zkstack_cli/crates/zkstack/src/commands/ecosystem/args/init.rs
Raid5594: bug as well

zkstack_cli/crates/zkstack/src/commands/chain/gateway/create_tx_filterer.rs
Deniallugo: but why is it in create tx filter?

zkstack_cli/crates/zkstack/src/commands/ecosystem/args/init.rs
Deniallugo: so, you assume that we have to call update submodules, only for ecosystem init?

Raid5594: No, I don't why?

Raid5594: I'm saying that when you do ecosystem init you would want to update submodules once, in other cases it's up to the operator

zkstack_cli/crates/zkstack/src/commands/chain/gateway/create_tx_filterer.rs
Raid5594: https://github.com/matter-labs/zksync-era/commit/c6b815d038c39782838618059c4a35894ca527ee

based on git blame this line was added by you and perekopskiy 3 months ago
I have no clue why. maybe it's more efficient to point whether we need it or not than playing back and forth on some code you definitely have more context than me

Deniallugo: I mean, it's the correct line. but it doesn't belong into create_tx_filter. 
we already have it inside `convert_to_gateway`

zkstack_cli/crates/zkstack/src/commands/ecosystem/args/init.rs
Deniallugo: @Raid5594 the only place where zero bridgehub is kinda correct (just don't want to go to the contract yet), is `runWithBridgehub` call. 
Please keep the unwrap there. 
It'll allow in the future use this bridgehub correctly

Deniallugo (approved):

iluwaa (approved):

PR #4420, opened 2025-08-22
chore(db): Add index for finalization

Deniallugo:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


core/lib/dal/migrations/20250822160137_indexes-for-finalization.up.sql
```diff
@@ -0,0 +1,3 @@
+CREATE INDEX IF NOT EXISTS eth_txs_finalized_tx
+  ON eth_txs_history (eth_tx_id)
+  WHERE finality_status = 'finalized';
```
dimazhornyk: is there any chance we might need to reuse this index for `finality_status != 'finalized';`? we could remove `WHERE` clause then

core/lib/dal/migrations/20250822160137_indexes-for-finalization.down.sql
```diff
@@ -0,0 +1 @@
+DROP INDEX IF NOT EXISTS eth_txs_finalized_tx;
```
dimazhornyk: it should be `DROP INDEX IF EXISTS`

core/lib/dal/migrations/20250822160137_indexes-for-finalization.up.sql
Deniallugo: For the majority of queries it's with finalized. but the table is small.

dimazhornyk (dismissed):

crimean-celica (approved):

PR #4456, opened 2025-09-03
feat(zkstack): Add zksync_os flag for ecosystem init

Deniallugo:
## What ❔

Add an ability for zkstack to work with zksync os contracts.

IMPORTANT!
It's still doesn't support different contracts for one ecosystem out of the box. 

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/zkstack_cli/crates/zkstack/src/commands/ecosystem/common.rs
+++ b/zkstack_cli/crates/zkstack/src/commands/ecosystem/common.rs
@@ -305,14 +305,8 @@ pub async fn init_chains(
 
         let chain_init_args = chain::args::init::InitArgs {
             forge_args: args.forge_args.clone(),
-            server_db_url: genesis_args
-                .as_ref()
-                .map(|a| a.server_db_url.clone())
-                .flatten(),
-            server_db_name: genesis_args
-                .as_ref()
-                .map(|a| a.server_db_name.clone())
-                .flatten(),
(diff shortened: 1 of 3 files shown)


StanislavBreadless (approved):

zkstack_cli/crates/zkstack/src/commands/chain/create.rs
```diff
@@ -119,6 +119,8 @@ pub(crate) async fn create_chain_inner(
         legacy_bridge,
         args.evm_emulator,
         args.tight_ports,
+        // TODO support creating zksync OS chains
```
mm-zk, line 122: could we make it 'crash' if zkos is set to true?
(otherwise someone might run this command, and then create a non-zkos chain which will be painful to debug later)

zkstack_cli/crates/zkstack/src/commands/ecosystem/common.rs
```diff
@@ -138,6 +140,8 @@ pub async fn deploy_l1_core_contracts(
         config.prover_version == ProverMode::NoProofs,
         config.l1_network,
         support_l2_legacy_shared_bridge_test,
+        // ZKSync OS flag is not used in core contracts deployment
+        false,
```
mm-zk, line 144: this might bite us in the future (someone will see that field exists in DeployL1Config and try to use it).

Maybe it should be 'Option<bool>' then?

PR #4469, opened 2025-09-09
fix: make sed work on macOS and Linux

mdqst:
## What ❔

fix `sed -i` so the script works on both macOS and Linux.

## Why ❔

`sed -i` behaves differently on macOS and Linux.
on macOS it requires a backup suffix, otherwise it breaks.
this change ensures cross-platform compatibility by checking `OSTYPE`.

## Is this a breaking change?

* [ ] Yes
* [x] No

## Operational changes

none, just updates the script to work correctly on both platforms.

## Checklist

* [x] PR title corresponds to the body of PR (we generate changelog entries from PRs).
* [ ] Tests for the changes have been added / updated.
* [ ] Documentation comments have been added / updated.
* [x] Code has been formatted via `zkstack dev fmt` and `zkstack dev lint`.

### DIFF
--- a/etc/nix/update-foundry-zksync.sh
+++ b/etc/nix/update-foundry-zksync.sh
@@ -11,16 +11,32 @@ update_sha() {
 
     case "$platform" in
         "linux_amd64")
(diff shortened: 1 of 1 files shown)


deniallugo-ml:
Closing — we don’t merge standalone cosmetic-only PRs; review overhead outweighs the benefit. (This one is deliberately direct — it discourages re-submission of the same kind.)

PR #4471, opened 2025-09-09
feat: add flag to save calldata

Raid5594:
## What ❔

Adding the ability to save calldata for register-ctm calls (without execution)

## Why ❔

We want to execute it with PUH

## Is this a breaking change?
- [ ] Yes
- [x] No

## Operational changes
Yes, it adds a flag `only_save_calldata` to the `register-ctm` command.

## Checklist

### DIFF
--- a/zkstack_cli/crates/zkstack/src/commands/ecosystem/args/init.rs
+++ b/zkstack_cli/crates/zkstack/src/commands/ecosystem/args/init.rs
@@ -269,6 +269,8 @@ pub struct RegisterCTMArgs {
     pub update_submodules: Option<bool>,
     #[clap(long, help = MSG_DEV_ARG_HELP)]
     pub dev: bool,
+    #[clap(long, default_missing_value = "false", num_args = 0..=1)]
+    pub only_save_calldata: bool,
 }
 
 impl RegisterCTMArgs {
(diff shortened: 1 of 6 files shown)


zkstack_cli/crates/zkstack/src/commands/ecosystem/args/init.rs
```diff
@@ -269,6 +269,8 @@ pub struct RegisterCTMArgs {
     pub update_submodules: Option<bool>,
     #[clap(long, help = MSG_DEV_ARG_HELP)]
     pub dev: bool,
+    #[clap(long, default_missing_value = "false", num_args = 0..=1)]
+    pub only_save_calldata: bool,
```
mm-zk, line 273: add comment (I guess if it runs on only_save_calldata, it will store it somewhere, right? where?)

mm-zk: and btw - does it save the calldata or only display on the screen?

Raid5594: saves the calldata in script-out as well as displays

StanislavBreadless (approved):

PR #4480, opened 2025-09-10
chore: adapt server for zksync os compatible contracts in era

0xValera:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


infrastructure/scripts/interop.sh
```diff
@@ -10,7 +10,7 @@ zkstack ecosystem init --deploy-paymaster --deploy-erc20 \
     --ignore-prerequisites --observability=false \
     --chain era \
     --update-submodules false
-    
+
```
StanislavBreadless, line 13: nit: no need to change this file

StanislavBreadless (approved):

infrastructure/scripts/interop.sh
0xValera: It was the only place where we had tabulation there, I'll leave it as is.
That's true that it could've been done in other PR though, agreed.

PR #4481, opened 2025-09-11
feat: update zkstack-cli for updated scripts

Raid5594:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/zkstack_cli/crates/zkstack/src/commands/ecosystem/common.rs
+++ b/zkstack_cli/crates/zkstack/src/commands/ecosystem/common.rs
@@ -22,7 +22,7 @@ use zkstack_cli_config::{
             DEPLOY_ERC20_SCRIPT_PARAMS, REGISTER_CTM_SCRIPT_PARAMS,
         },
     },
-    traits::{ReadConfig, ReadConfigWithBasePath, SaveConfig, SaveConfigWithBasePath},
+    traits::{ReadConfig, SaveConfig, SaveConfigWithBasePath},
     ContractsConfig, ContractsConfigForDeployERC20, CoreContractsConfig, EcosystemConfig,
     GenesisConfig, ZkStackConfigTrait, GENESIS_FILE,
 };
@@ -113,7 +113,7 @@ pub async fn deploy_ctm(
         shell,
         DEPLOY_CTM_SCRIPT_PARAMS.output(&config.path_to_foundry_scripts()),
     )?;
-    let mut contracts_config = ContractsConfig::read_with_base_path(shell, &config.config)?;
+    let mut contracts_config = ContractsConfig::default();
     contracts_config.update_from_l1_output(&script_output);
 
     Ok(contracts_config)


zkstack_cli/crates/zkstack/src/commands/ecosystem/args/init.rs
```diff
@@ -303,15 +305,19 @@ pub struct RegisterCTMArgsFinal {
     pub forge_args: ForgeScriptArgs,
     pub update_submodules: Option<bool>,
     pub only_save_calldata: bool,
+    pub bridgehub_address: H160,
+    pub ctm_address: H160,
 }
 
-impl From<EcosystemInitArgsFinal> for RegisterCTMArgsFinal {
-    fn from(args: EcosystemInitArgsFinal) -> Self {
+impl From<(EcosystemInitArgsFinal, H160, H160)> for RegisterCTMArgsFinal {
```
StanislavBreadless: a bit nicer would be to create a method in `RegisterCTMArgsFinal` `from_init_args(..., bridgehub, ctm_address)`

as it is a bit awkard that one has to spawn a tuple to get `RegisterCTMArgsFinal`

zkstack_cli/crates/zkstack/src/commands/ecosystem/build_transactions.rs
```diff
@@ -50,7 +51,8 @@ pub async fn run(args: BuildTransactionsArgs, shell: &Shell) -> anyhow::Result<(
         Some(args.sender),
         false,
         false,
-        None,
+        H160::default(), // What's the purpose of build_transactions?
```
StanislavBreadless: Indeed it is not right intuitively, I think we should firstly deploy bridgehub and then provide the value here, but @Deniallugo you know better as I am not sure about the purpose of this file too

zkstack_cli/crates/zkstack/src/commands/ecosystem/common.rs
```diff
@@ -37,10 +38,10 @@ use crate::{
 
 lazy_static! {
     static ref DEPLOY_L1_FUNCTIONS: BaseContract = BaseContract::from(
```
StanislavBreadless: Since we are accepting changes of this kind, maybe rename it to deploy ctm?

StanislavBreadless: and the same for `deploy_l1` function in zkstack

Raid5594: do you wanna rename script as well? I mean contracts files

zkstack_cli/crates/zkstack/src/commands/ecosystem/init.rs
```diff
@@ -113,60 +113,53 @@ async fn init_ecosystem(
...
-        init_args.forge_args.clone(),
-        ecosystem_config,
-        initial_deployment_config,
-        init_args.support_l2_legacy_shared_bridge_test,
-        init_args.deploy_ecosystem,
-    )
-    .await?;
-    contracts.save_with_base_path(shell, &ecosystem_config.config)?;
+    let contracts = if !init_args.deploy_ecosystem {
+        return_ecosystem_contracts(shell, &mut init_args.ecosystem, ecosystem_config).await?
```
Raid5594, line 117: what is the aim here?

shall we return ecosystem contracts instead of all 3 steps?

zkstack_cli/crates/config/src/consts.rs
```diff
@@ -19,6 +19,8 @@ pub(crate) const INITIAL_DEPLOYMENT_FILE: &str = "initial_deployments.yaml";
 /// Name of the erc20 deployments config file
 pub(crate) const ERC20_DEPLOYMENT_FILE: &str = "erc20_deployments.yaml";
 /// Name of the contracts file
+pub const CORE_CONTRACTS_FILE: &str = "core_contracts.yaml";
+/// Name of the contracts file
```
Deniallugo: what is the purpose of the additional file? 
i'd prefer to keep it all in contracts, because it's easier from the server side

zkstack_cli/crates/zkstack/src/commands/ecosystem/args/init.rs
```diff
@@ -331,9 +341,11 @@ pub struct InitNewCTMArgs {
     #[clap(long, default_missing_value = "false", num_args = 0..=1)]
     pub support_l2_legacy_shared_bridge_test: Option<bool>,
     #[clap(long, help = MSG_BRIDGEHUB)]
-    pub bridgehub: Option<String>,
+    pub bridgehub: String,
     #[clap(long, help = MSG_ZKSYNC_OS)]
     pub zksync_os: bool,
+    #[clap(long, default_value_t = false)]
+    pub reuse_gov_and_admin: bool,
```
Deniallugo: please set, `default_missing_value = "true"` it'd allow to pass this argument, without specifying `=true` from CLI

zkstack_cli/crates/zkstack/src/commands/ecosystem/build_transactions.rs
Deniallugo: I'll set a task for my self, to remove this part of code. 
@Raid5594 It'd be great if you set a comment with TODO.

it's a bad attempt to generate calldata, instead of deploying everything

Deniallugo: But if you want to have a working code here (which is required for ci, you have to pass bridgehub in arguments

zkstack_cli/crates/zkstack/src/commands/ecosystem/init.rs
Deniallugo: The idea, is to reuse ecosystem contracts straight away if deployment is not necessary

zkstack_cli/crates/config/src/consts.rs
Raid5594: Just the output for core-contracts script, otherwise we'll overwrite output when running deploy ctm

zkstack_cli/crates/zkstack/src/commands/ecosystem/init.rs
Raid5594: Ok I will move register ctm inside the 2nd branch of if statement, so it's skipped

Deniallugo: kk

zkstack_cli/crates/config/src/consts.rs
Deniallugo: can we overwrite, only things that are necessary? 
e.g. using different structs for out?

zkstack_cli/crates/zkstack/src/commands/ecosystem/common.rs
Deniallugo: Yes, please ;)

zkstack_cli/crates/config/src/consts.rs
Raid5594: I'm fetching in the script all the implementations as well, so nothing is lost

Deniallugo: Actually, we need `core_contracts.yaml`, but for a different reason. 
We have to support multiple CTMs, and we have to write each ctm deployments to their individual file and merge `core_contracts` and potential CTM in flight for each chain and produce just contracts.yaml. 

But, it's not for this PR. 
It's for the next step

zkstack_cli/crates/zkstack/src/commands/ecosystem/init.rs
```diff
@@ -113,60 +115,53 @@ async fn init_ecosystem(
...
+            init_args.forge_args.clone(),
+            ecosystem_config,
+            initial_deployment_config,
+            init_args.support_l2_legacy_shared_bridge_test,
+            core_contracts.core_ecosystem_contracts.bridgehub_proxy_addr,
+            init_args.zksync_os,
+            true,
+        )
+        .await?;
+        contracts.merge_and_save_with_base_path(shell, &ecosystem_config.config)?;
```
Deniallugo: This part is confusing. 

You have `core_contracts` that you have saved a few lines before. 

But in `deploy_new_ctm` instead of reading this file from the disk or passing `mut core_contracts` to the deploy_ctm.
 You are creating empty contracts struct and later merging `core_contracts` that you saved before with `contracts` from `deploy_new_ctm`. 

I'd suggest either pass `core_contracts` as mutable or instead of `Contracts::default()` read it from the disk.

zkstack_cli/crates/zkstack/src/commands/ecosystem/common.rs
```diff
@@ -108,9 +109,9 @@ pub async fn deploy_l1(
 
     forge.run(shell)?;
 
-    let script_output = DeployL1Output::read(
+    let script_output = DeployCTMOutput::read(
         shell,
-        DEPLOY_ECOSYSTEM_SCRIPT_PARAMS.output(&config.path_to_foundry_scripts()),
+        DEPLOY_CTM_SCRIPT_PARAMS.output(&config.path_to_foundry_scripts()),
     )?;
     let mut contracts_config = ContractsConfig::default();
```
Deniallugo, line 116: Please load it from the disc.

PR #4487, opened 2025-09-15
fix(en): allow to disable sanity checks for commitment generation

Deniallugo:
## What ❔

Disabling sanity checks for commitment generation

## Why ❔

Our external node is designed around commitment generation and verification. Now it's possible to be in a situation, that we have incorrect block and we can't revert it, because we don't have a commitment for it. 

EN is not able to produce any commitment => EN can't verify the commitment => EN can't revert incorrect block. 

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/core/node/commitment_generator/src/node/commitment_generator.rs
+++ b/core/node/commitment_generator/src/node/commitment_generator.rs
@@ -17,6 +17,7 @@ use crate::CommitmentGenerator;
 #[derive(Debug, Default)]
 pub struct CommitmentGeneratorLayer {
     max_parallelism: Option<NonZero<u32>>,
+    disable_sanity_checks: bool,
 }
 
 #[derive(Debug, FromContext)]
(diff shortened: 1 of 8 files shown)


core/node/commitment_generator/src/node/commitment_generator.rs
```diff
@@ -37,6 +38,10 @@ impl CommitmentGeneratorLayer {
         self.max_parallelism = max_parallelism;
         self
     }
+    pub fn disable_sanity_checks(mut self) -> Self {
```
perekopskiy, line 42: nit: it's a bit weird to see both `disable_sanity_checks` and `enable_sanity_checks`, I'd choose one and use it for the layer and the component

Deniallugo: agree

perekopskiy (approved):

PR #4488, opened 2025-09-16
fix(EN): properly set chain_id for old batch transactions

cytadela8:
## What ❔

If received batch details contain None/Null chain_id (commit_chain_id, prove_chain_id, execute_chain_id) set it to config provided chain id for L1. This happens for old batches.

## Why ❔

If you try syning EN from scratch with a "old" chain you will encounter following error: 
```
Error: One or more tasks failed: ["Task transaction_finality_updater failed: No batch transactions to process for chain id 1 while there are some for None chain_id.Error is thrown so node can restart and reload SL data. If node doesn't make any progress after restart, then it's bug, please contact developers."]
```

Logic for fetching batch transactions to process their finality status assumes chain_ids are set properly and no transaction should have chain_id = None. However this is the case for old chains. When fetching they return *_chain_id: null in zks_getL1BatchDetails for some old batches.

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/core/bin/external_node/src/node_builder.rs
+++ b/core/bin/external_node/src/node_builder.rs
@@ -278,8 +278,9 @@ impl ExternalNodeBuilder {
     }
 
     fn add_batch_transaction_fetcher_layer(mut self) -> anyhow::Result<Self> {
-        self.node
-            .add_layer(BatchStatusUpdaterLayer::new(self.config.local.networks.l1_chain_id));
+        self.node.add_layer(BatchStatusUpdaterLayer::new(
+            self.config.local.networks.l1_chain_id,
+        ));
         Ok(self)
     }
 
--- a/core/node/node_sync/src/node/batch_transaction_fetcher.rs
+++ b/core/node/node_sync/src/node/batch_transaction_fetcher.rs
@@ -55,7 +55,8 @@ impl WiringLayer for BatchStatusUpdaterLayer {
             app_health,
         } = input;
 
-        let updater = BatchStatusUpdater::new(self.l1_chain_id, main_node_client, pool.get().await?);
+        let updater =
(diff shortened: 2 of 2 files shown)


cytadela8:
Seems to be working after testing sync with Abstract testnet (managed to fully sync) and Sophon testnet (sill syncing)

core/node/node_sync/src/batch_transaction_fetcher/mod.rs
```diff
@@ -103,10 +104,14 @@ struct UpdaterCursor {
...
     last_proven_l1_batch: L1BatchNumber,
     last_committed_l1_batch: L1BatchNumber,
+    default_sl_chain_id: SLChainId,
 }
 
 impl UpdaterCursor {
-    async fn new(storage: &mut Connection<'_, Core>) -> anyhow::Result<Self> {
+    async fn new(
+        storage: &mut Connection<'_, Core>,
+        default_sl_chain_id: SLChainId,
```
dimazhornyk, line 113: not sure if this is intentional, but we use `l1_chain_id` as `sl_chain_id` here, which means if the chain is on GW - they are different

if I understand this correctly - we can't just use L1 or GW chain ID here, and we have to look into what was the settlement layer for every specific batch before the chain_id column was introduced

cytadela8: My (I believe correct) assumption is that chain_id column was introduced before GW, so we can assume its L1. This column in the eth_txs table was introduced in migration from 20240819.

dimazhornyk: makes sense then!

dimazhornyk (approved):

PR #4492, opened 2025-09-17
feat(en): Check commitment allonside with roothashes

Deniallugo:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


core/node/reorg_detector/src/lib.rs
```diff
@@ -165,6 +173,21 @@ impl MainNodeClient for Box<DynClient<L2>> {
         };
         Ok(batch.base.root_hash.ok_or(MissingData::RootHash))
     }
+
+    async fn l1_batch_commitment(
```
cytadela8: I'm somewhat hesitant about this addition of 1 RPC call per batch. It might make debugging easier, but this mismatch would also be found during commit transaction finality status processing. (there would be mismatch between event logs and local state)

Deniallugo: that's right, but it won't be in reorg detector....we just can't find the diverged block

Deniallugo: but, i'll add a small cache, that will avoid 2 rpc requsts

core/node/reorg_detector/src/lib.rs
```diff
@@ -329,7 +329,8 @@ impl ReorgDetector {
                     .update_correct_block(checked_l2_block, checked_l1_batch);
                 None
             } else {
-                let diverged_l1_batch = checked_l1_batch;
+                let diverged_l1_batch =
+                    checked_l1_batch + max(root_hashes_match, commitments_match) as u32;
```
cytadela8, line 333: why not `+ (root_hashes_match || commitments_match) as u32`

Deniallugo: it's the same.

PR #4495, opened 2025-09-22
feat: Bump dependencies

EmilLuta:
Updates dependencies to latest version

PR #4501, opened 2025-09-23
feat: Update dependencies

EmilLuta:
This is a release from prover-v21.3.0 that updates the and codebase in a similar matter to [this PR](https://github.com/matter-labs/zksync-era/pull/4495).


etc/env/file_based/genesis.yaml
```diff
@@ -1,16 +1,16 @@
 genesis_root: 0x85fa0abaa93032259defb7bc3d060f1fcce909600a71841e9a578db386723966
 genesis_rollup_leaf_index: 78
 genesis_batch_commitment: 0xc66918c8860c607e73892cb3e686bfba6185f7a49f8c7ec21fb1cce3c4f353cd
-genesis_protocol_version: 28
+genesis_protocol_version: null
```
Deniallugo: you have to either remove it or set the proper version

EmilLuta: You're absolutely right, but we will not merge this code & CI will not run as we have too many conflicts.

deniallugo-ml:

PR #4504, opened 2025-09-24
fix(upgrades): Read all skipped events 

Deniallugo:
## What ❔

Read upgrades even if we skipped some of them 

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/core/node/eth_watch/src/client.rs
+++ b/core/node/eth_watch/src/client.rs
@@ -11,11 +11,11 @@ use zksync_eth_client::{
     EthInterface,
 };
 use zksync_system_constants::L2_MESSAGE_ROOT_ADDRESS;
-use zksync_types::protocol_version::ProtocolSemanticVersion;
 use zksync_types::{
     abi::ZkChainSpecificUpgradeData,
     api::{ChainAggProof, Log},
     ethabi::{decode, Contract, ParamType},
+    protocol_version::ProtocolSemanticVersion,
     u256_to_h256,
     utils::encode_ntv_asset_id,
     web3::{BlockId, BlockNumber, Filter, FilterBuilder},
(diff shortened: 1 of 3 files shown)


core/node/eth_watch/src/client.rs
```diff
@@ -55,10 +57,10 @@ pub trait EthClient: 'static + fmt::Debug + Send + Sync {
         verifier_address: Address,
     ) -> Result<Option<H256>, ContractCallError>;
     /// Returns upgrade diamond cut by packed protocol version.
```
perekopskiy: update comment, also please mention that `version` is exclusive lower bound

core/node/eth_watch/src/client.rs
```diff
@@ -421,31 +448,43 @@ where
...
+            .ok_or(EnrichedClientError::custom(
+                format!("No diamond cut found for version {from_version}"),
+                "diamond_cuts_since_version",
+            ))
+            .map_err(|e| {
+                tracing::error!("{e}");
+                e
+            })
+            .unwrap_or(to_block.saturating_sub((LOOK_BACK_BLOCK_RANGE - 1).into()))
+            + 1;
```
perekopskiy, line 476: looks like in this case you can return upgrades for version <= from_version and it looks like you don't handle it anywhere. IMO you should propagate error here

Deniallugo: I don't think it's the case.
Only if someone modified the database

Deniallugo: Or if it was a long time ago more than `LOOK_BACK_BLOCK_RANGE` it's also safe, because we won't process it. 

At the same time we need to properly handle the genesis upgrade

perekopskiy: tbh I didn't understand a work for your last comment

perekopskiy (approved):

PR #4522, opened 2025-10-03
docs: sb/extend auditor docs on AssetTracker component

0xValera:
## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
+++ b/docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
@@ -1,28 +1,36 @@
 # Contract-based and Full AssetTracker
 
-## Security assumptions and vocabulary
+## Introduction
 
-- All settlement layers are whitelisted and completely trusted: they are controlled by the decentralized governance and sufficient monitoring and defense in depth mechanisms ensure that their proof system will not be exploited.
+While interop allows for more possibilities for cross chain interaction, it also comes with additional risks: a malicious chain has more ways to impact other chains, since now direct communication between chains is available that does go through L1.
+
+The `AssetTracker` is the main component that is used to ensure that even if a malicious chain is present, it will not affect the broader ecosystem. For now it is implemented in Solidity (and that's why interop is only allowed on top of a ZK Gateway).
+
+## Glossary
+
 - A "completely compromised" chain assumes that not only their proof system, but their facets can change arbitrarily. This is a relevant assumption since zksync os chains CTM will be controlled by a temporary multisig before final transition to the governance. 
 - A "ZK compromised" means that the chain uses a canonical implementation with our Diamond Proxy, facets, etc, but its ZK proving system is compromised.
 
+## Security assumptions
+
+- All settlement layers are whitelisted and completely trusted: they are controlled by the decentralized governance and sufficient monitoring and defense in depth mechanisms ensure that their proof system will not be exploited.
+
+> In the future, we want to allow untrusted settlement layers, however the scope of this release includes only trusted settlement layers. Any places that disallow it for now should be clearly marked in the codebase.
+
 ### Before v30
 
-- Only era-based chains were settling on Gateway, and those are trusted fully (TODO do we trust fully their proofs or just impleentation? We should only assue that they can be ZK compromised).
-- Non-era based chains are possible. They should be assumed to be completely malicious even before the upgrade, but they can not settle on top of ZK based Gateway. Before v30 they only settle on L1.
+- Only era-based chains were settling on Gateway, and those are trusted fully (TODO: do we trust fully their proofs or just implementation? We should only assue that they can be ZK compromised).
+- ZKsync OS based chains are possible. Before their upgrade system is migrated to the decentralized governance, they can be assumed to be potentially completely malicious even before the upgrade, but they can not settle on top of ZK based Gateway. Before v30 they only settle on L1.
 - Only one whitelisted settlement layer exists and it is ZK Gateway. 
 
 ### After v30
 
-- A new zksync os powered settlement layer may be added. The transfer of control to decentralized governance is a prerequisite, i.e. then this settlement layer will also be completely trusted, but also zksync os chains will be able to become "ZK compromised at most".
-- The system should be ready for chains to be able to migrate from one GW to L1, to another Gateway. While it will not be supported in Bridgehub immediately, overall the security model should work with it
+- A new ZKsync OS powered settlement layer may be added. The transfer of control to decentralized governance is a prerequisite before the creation of such a settlement layer, i.e. then this settlement layer will also be completely trusted, but also zksync os chains will be able to become "ZK compromised" at most. Note, that the transfer of the ownership to governance is a prerequisite for ZKsync OS based settlement layer, not v30 upgrade in general.
+- Chains can only migrate between L1 and ZK Gateway that belongs to their CTM (i.e. Era chains wont be able to migrate to ZKsync OS settlement layer and vice versa).
 
-From the above it follows that in the meantime while zksync os chains can be completely malicious, they can only settle on L1, while their settlement on top of a ZK Gateway will be only possible only when they may be "ZK compromised".
+> Future compatibility note 1. In the future we want to allow Era-based chains to migrate on top of ZKsync OS powered Gateway. The codebase should be ready for it in general. Any places that disallow it for now should be clearly marked in the codebase.
 
-
-## Introduction
-
-The `AssetTracker` is a component that provides additional security, by tracking the balance of chains on the chains settlement layer (only Gateway and L1 as of now). This is done by parsing all interop transactions on the SL and updating the balance of the chains.
+> Future compatibility note 2. In the future we would want to support chains from untrusted, potentially completely malicious CTMs settling on top of ZKsync OS powered Gateywa. Any places that disallow it for now should be clearly marked in the codebase. 
 
 ## Before v30 (only L1<>L2 messaging)
 
@@ -30,28 +38,26 @@ Before v30, the chain balance was tracked only inside `L1NativeTokenVault`.
 
 - the balances of the tokens for each chain are stored only on L1.
 - when there is a deposit to the chain we increase it's relevant balance, opposite for withdrawals, we decrease the chain's balance for the token that's being withdrawn.
-- For L2 Native tokens, we store the origin chainId. We don’t track the balance for that token on said chain, since the token can be minted arbitrarily. TODO: do we store it for l1 based tokens???
+- For all tokens, we store the origin chainId. We don’t track the balance for that token on said chain, since the token can be minted arbitrarily.
 
 ```solidity
 mapping(chainId => assetId => balance) chainBalance;
 mapping(bytes32 assetId => uint256 originChainId) public originChainId;
 ```
 
-## After v30 (interop with asset transfers support)
-
-### Contract based AssetTracker
+## Contract based AssetTracker
 
 To enable interop, while keeping the old 2FA mechanism, we would need a contract that would all outcoming messages from a chain (both interop messages and withdrawals) while ensuring that the chain does not spend more funds then it has. This contract is the asset tracker.
 
 There are three types of asset trackers:
 
 - `L1AssetTracker`. It took over the job of tracking chain balances on L1, that was previosuly held by `L1NativeTokenVault`. Note that due to costs, it does not support interop.
-- `GWAssetTracker`. Its job is to ensure that chains that settle on top of ZK Gateway do not send more funds in outgoing messages, then they have.
+- `GWAssetTracker`. Its job is to ensure that chains that settle on top of a ZK Gateway do not send more funds in outgoing messages, then they have.
 - `L2AssetTracker`. It is predeployed on all L2 chains. Its main purpose is to facilitate migration of chain between L1 and ZK Gateway as well as ensuring the secure behavior of tokens, native to this chain. More on it here (TODO).
 
 ### Invariants around `chainBalance`
 
-The topic of migration between settlement layer will be discussed in the later sections. In this section, we'll discuss what the invariants around `chainBalance` and `totalSupplyAcrossAllChains` look like and how they changed compared to the previous version.
+The topic of migration between settlement layer will be discussed in the later sections. In this section, we'll discuss what the invariants around `chainBalance` look like and how they changed compared to the previous version.
 
 #### Motivation
 
@@ -59,8 +65,9 @@ For comparison, in the past, `chainBalance[chainId][assetId]` had the following 
 
 - Every time a chain `chainId` deposits `X` of `assetId` to the L1NativeTokenVault (only possible via L1->L2 communication), the `chainBalance[chainId][assetId]` gets increased by `X`.
 - Every time a chain `chainId` withdraws `X` of `assetId` to the L1NativeTokenVault (either through withdrawals or claiming failed deposits), the `chainBalance[chainId][assetId]` got decreased by `X`.
+- If a chain is the origin chain of the token, its `chainBalance` was not tracked.
 
-Note, that the `chainBalance` for a chain was tracked *lazily*, i.e. the `chainBalance` displayed the difference between deposited and withdrawn funds from the shared bridge, not how much funds the chain actually has internally. This is an important distinction, e.g. a potentially malicious chain (obviously assuming totally compromised ZK) could create "withdraw" messages for more funds than the chain actually has. 
+Note, that the `chainBalance` for a chain was tracked *lazily*, i.e. the `chainBalance` displayed the difference between deposited and withdrawn funds from the shared bridge, not how much funds the chain actually has internally. This is an important distinction, e.g. a potentially malicious chain (obviously assuming totally compromised ZK) could *create*, but *not finalize* "withdraw" messages for more funds than the chain actually has. 
 
 So in case of a compromised ZK, the `L1NativeTokenVault` and its `chainBalance` tracking protected the shared bridge, but it gave no guarantee about the validity of all the messages sent from the chain. The above is very neat for chains that only communicate with L1 as it is very cheap and provides easy isolation between the chains.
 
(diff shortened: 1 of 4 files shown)


docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
```diff
@@ -2,26 +2,167 @@
...
 ```
 
-### Contract based AssetTracker
+## Contract based AssetTracker
+
+To enable interop, while keeping the old 2FA mechanism, we would need a contract that would read all outcoming messages from a chain (both interop messages and withdrawals) while ensuring that the chain does not spend more funds then it has. This contract is the asset tracker.
+
+There are three types of asset trackers:
+
+- `L1AssetTracker`. It took over the job of tracking chain balances on L1, that was previosuly held by `L1NativeTokenVault`. Note that due to costs, it does not support interop.
```
0xValera, line 63: Does it track chain balances for all of the chains, or only the ones who have L1 as their SL?

0xValera, line 63: What about when we'll have L1 interop support (v31?)?

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
```diff
@@ -2,26 +2,167 @@
...
+- `GWAssetTracker`. Its job is to ensure that chains that settle on top of a ZK Gateway do not send more funds in outgoing messages, then they have.
+- `L2AssetTracker`. It is predeployed on all L2 chains. Its main purpose is to facilitate migration of chain between L1 and ZK Gateway as well as ensuring the secure behavior of tokens, native to this chain. More on it here (TODO link).
```
0xValera: > as well as ensuring the secure behavior of tokens, native to this chain

Including activity of these tokens on other L2s? The statement itself sounds a bit weird to me.

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
```diff
@@ -2,26 +2,167 @@
...
+
+### Restrictions
+
+For now the `AssetTracker` will only support tokens in the native token vault. These are assumed to have a single origin chain id, i.e. single chain where the supply of the token can change (via native minting or burning).
```
0xValera, line 69: > will only support tokens in the native token vault

What exactly does it mean? That AT will only support the same set of tokens that NTV supported before v30?

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
```diff
@@ -2,26 +2,167 @@
...
+
+However, it becomes a big issue for interop: what if a compromised malicious chain that only holds 100 USC sends two messages, where each is worth 80 USDC? Only one of those messages could be processed. Putting the responsibility of accounting for each individual message on the recipient is hard and error prone, so it was decided that if a chains are to use interop, *each and every message* has be validated whenever it is added to the shared message tree.
+
+Also, note that the previous way of using `chainBalance` stored on L1 for withdrawas is not an option, since a chain might have never deposited funds to the shared bridge, but only received those from the other chains via interop. Its `chainBalance` on L1 would be zero, but it does have the funds. In such cases, ZK Gateway, who approved the withdrawl via `GWAssetTracker` should be help responsible (and so its `chainBalance` should be reduced).
```
0xValera, line 89: > In such cases, ZK Gateway, who approved the withdrawl via `GWAssetTracker` should be help responsible (and so its `chainBalance` should be reduced).

"such case" is when chain A interop-ed token X to chain B, and then chain B withdraws these tokens to L1, right?

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
```diff
@@ -2,26 +2,167 @@
...
+Overall, the algorithm for determining whether to reduce the chain balance should look the following way:
+
+```
+1. If withdrawal is directly from L2 to L1 (chain didnt settle on GW), the chainBalance of the chain should be reduced.
+2. Alternatively, if the chain did settle on GW, check whether the batch is before the chain has upgraded to v30.
+    2a. If the batch is before the chain upgraded to v30, the chain's balance is reduced.
+    2b. If the batch is after the chain upgraded to v30, the settlement layer's balance is reduced.
+```
+
+Note, that the above scheme relies on a fact that chains never lie about the time they upgraded to v30. This is ensured inside MessageRoot (TODO add link to the upgrade process):
```
0xValera: > chains never lie about the time they upgraded to v30

Including ZKsync OS (i.e. potentially "completely compromised") chains?

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
```diff
@@ -2,26 +2,167 @@
...
+ - Additionally we also assume that chains that have non-zero v30 upgrade batch number, belong to either era CTM or ZKsync OS based CTM (that are controlled by the governance). For all the other CTMs, when a chain will be spawned, its "v30 upgrade batch number" will be started from 0.
```
0xValera: Again, same question as above, "completely compromised" ZKsync OS chain (which resides on a multisig-controlled/malicious CTM) is OOS?

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
```diff
@@ -2,26 +2,167 @@
...
+
+> Note, that to avoid any sort of edge cases related to chain migration, we assume that each deposit is always fully processed on the settlement layer, i.e. a chain can only have a message for success or failure of a deposit when settling on top of Gateway only if this deposit was initiated settled on Gateway too and as a result when through `GWAssetTracker` (more on it later). Note, that we can not assume the contrary: a malicious chain MAY publish an arbitrary message when settling failed depoist/successful deposit on L1, since L1AssetTracker does not check contents of the message.
+
+When chain migrates on top of GW, it needs to "give" some of its balance for internal usage of Gateway. More on the migration process will be described later (TODO), but basically a chain can tell that a certain amount of funds inside the `chainBalance` is still "active", i.e. maintained within the chain and so can be used for future interop/withdrawals etc.
```
0xValera: Why do we need this?
Is it to account for withdrawals initiated before migrating that weren't finalized yet at the moment of migration?
Then, does the "algorithm for determining whether to reduce the chain balance" determines where the chain was settling at the moment of initiating of withdrawal? To decrease the correct chainBalance

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
```diff
@@ -44,81 +185,190 @@ contract AssetTracker {
...
+- On L1, when the user proves the `failedDeposit`, the balance is subtracted from the Gateway's `chainBalance`.
+- On GW the chain's `chainBalance` is decreased, when the chain executes its batch, and calls the `processLogsAndMessages` function. Each failed L1->L2 transaction produces a L2->L1 message, which we can verify in `processLogsAndMessages`. To know the balance change for the specific L1->L2 transaction, we save the balance change in the `forwardTransactionOnGatewayWithBalanceChange` function call.
+
+Note, that to ensure that the logic above is always secure, `GWAssetTracker` is also responsible for maintaining an invariant that all deposits that were submitted when the chain was settling on top of ZK Gateway must be processed *before* the chain tries to leave Gateway.
```
0xValera: Is the system configured in such a way that it's not possible for Gateway to block the chain from leaving it?
Basically that it's impossible to always have an "in-flight" deposit which will prevent the chain from leaving it. I know that we have 3.5 days of full deposit stop, is that 100% enough? I'm not sure how that was chosen.

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
```diff
@@ -44,81 +185,190 @@ contract AssetTracker {
...
 
 ### Migrating and settling on Gateway
 
 When a chain migrates from L1 to GW or from GW to L1, the `chainBalance` mapping for each token does not get transferred automatically. They can be trustlessly moved by anyone from GW to L1 and vice versa. But our standard tooling will migrate all token balances for best UX.
 
-When a chain is settling on Gateway it calls the `processLogsAndMessages` on settlement. This means that if the token is not yet migrated, and a withdrawal could be processed, the settlement would fail, since the chain's balance would be zero. In order to prevent this, withdrawals and interop will be paused on the chain until the token's balance is migrated.
+When a chain is settling on Gateway it calls the `processLogsAndMessages` on settlement. This means that if the token is not yet migrated, and a withdrawal is processed, the settlement would fail, since the chain's balance would be zero. In order to prevent this, withdrawals and interop will be paused on the chain (via `L2AssetTracker`) until the token's balance is migrated.
```
0xValera, line 222: > In order to prevent this, withdrawals and interop will be paused on the chain (via `L2AssetTracker`) until the token's balance is migrated

Responsibility of ChainAdmin?

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
```diff
@@ -44,81 +185,190 @@ contract AssetTracker {
...
 
-The chain operator is responsible for migration of all token balances. However as each token migration costs L1 gas, this can be a spam attack vector. The chain can choose to ignore tokens, in which case the users can migrate the balances themselves. 
+Anyone can migrate token balances permissionless, though typically it would be the chain operator that migrates them. This ensures that even if there are lots of tokens to migrate, the operator may choose to ignore rarely used assets that can be migrated by users at their own pace.
```
0xValera, line 224: Why chain operator and not chain admin?

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
```diff
@@ -44,81 +185,190 @@ contract AssetTracker {
...
+
+TODO: maybe ask Vlad for stage1 alignment + the current process seems too ddosable by service transactions
+
+#### Replay protection and edge cases with messaging
+
+Our L2->L1 (or GW->L1) messages are valid in perpetuity. Typically, for actions like token withdrawals, we used `L1Nullifier` contract to store the nullifier that ensures that the same message can not be replayed twice.
+
+However, for migration-related messages for the ease of implementation we used the `migrationNumber` and `assetMigrationNumber`: An asset-migrating message can be only processed on L1 once, since to process it, the carried `migrationNumber` needs to be greater than the current `assetMigrationNumber`.
+
+TODO: the current system allows for some of the old token withdrawals to never get finalized, this is bad, we'll have to rework it.
```
0xValera: > we'll have to rework it

OOS of v30? They will be finalizeable once we fix, but until then some of the old token withdrawals might be infinitely stuck?

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
```diff
@@ -44,81 +185,190 @@ contract AssetTracker {
...
+
+#### Recovering from missed migrations
+
+What happens if did not migrate a tokens' balance?
```
0xValera: missing word(s)

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
0xValera: while reading I understood as "only the ones who have L1 as their SL"

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
0xValera: as I understand after reading, yes

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
0xValera: as I understand after reading yes, the algorithm should look at the proof, and we can get info about settlement layer there (?).
It's not listed in the "algorithm..." though, I think

docs/src/specs/contracts/interop/message_root.md
```diff
@@ -119,6 +119,24 @@ It is the responsibility of the chain to ensure that each message has a unique i
 
 Another notable example of the redundancy of data, is that we also have total `MessageRoot` on L1, which contains the aggregated root of all chains, while for chains that settle on L1, we still store the `ChainBatchRoot` for the efficiency.
 
+## MessageRoot as global source of batches
+
+Before v30, we used to simply store the chain batch roots for chains inside their DiamondProxy and then whenever someone needed to check whether a message was present, we queried its Mailbox (TODO: link).
+
+However, the logic about relies on the fact that this getter will never be malicious. This is no longer acceptable taking into account that potentially malicious chains could be present. Starting from v30 each chain (whether it settles on L1 or GW) is expected to store its batch roots inside `MessageRoot` (`chainBatchRoots` mapping). To ensure that the chain does append only blocks in the sequential order, we also maintain `currentChainBatchNumber`.
```
0xValera:
```suggestion
However, the logic above relies on the fact that this getter will never be malicious. This is no longer acceptable taking into account that potentially malicious chains could be present. Starting from v30 each chain (whether it settles on L1 or GW) is expected to store its batch roots inside `MessageRoot` (`chainBatchRoots` mapping). To ensure that the chain does append only blocks in the sequential order, we also maintain `currentChainBatchNumber`.
```

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
StanislavBreadless: yes, it is better to read the whole thing first to get a wholistic picture

StanislavBreadless: > only the ones who have L1 as their SL

but also for chains that settle on top of SL it stores the balances it needs to serve the unfinalized withdrawals/failed deposits that the chain is responsible for (i.e. from the times it settled on L1 or before it upgraded to v30)

StanislavBreadless: Unlikely in v31, since the same reasoning about costs applies. Most likely it will require full ZK-based asset tracker for L1 support

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
StanislavBreadless: "native to their respective chains"?

E.g., `L2AssetTracker` deployed on Era protects only Era from its own bad tokens

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
StanislavBreadless: Yes. In the future AT can support arbitrary assetIds, but for now supporting only NTV-based assets is a lot easier

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
StanislavBreadless: yes, more on it in the v30 upgrade process doc

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
StanislavBreadless: I meant there that if something settles on GW and has non zero v30 batch number, it is controlled by governance. Since ZKsync OS chains will get on top of ZK Gateway once they are controlled by governance

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
StanislavBreadless: Yes, it is extended below in the withdrawal flow for L1AssetTracker

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
StanislavBreadless: We'll have to double check on the future stage1 alignment, but as long as L1->GW txs are not censored it is not possible to block

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
StanislavBreadless: No, it is done automatically by `assetMigrationNumber` tracking

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
StanislavBreadless: As stated in this section, anyone can do it. The reason why I said "operator" is because it is a hot wallet. chain admin will definitely not do it since it is a multisig. In reality, a 3rd hot wallet will be used for it

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
StanislavBreadless: It should be inside v30 IMO

docs/src/specs/contracts/bridging/asset_tracker/asset_tracker.md
0xValera: Is that not a blocker for stage 1? If we're not able to "catch" an in-flight interop transfer started on L2-A on L1

StanislavBreadless: Receiving interop may indeed be a blocker for stage1, but it will not be supported in v30 anyway.

0xValera: v30 true, but in v31 we planned to have stage1, so that's why I brought it up initially

docs/src/specs/contracts/interop/message_root.md
```diff
@@ -35,7 +35,7 @@ Note that the `MessageRoot` appears twice in the structure. So the structure is
...
 
-When ZKsync chain's batch gets executed on the settlement layer, the chain calls the `MessageRoot.addChainBatchRoot` function, while providing the `ChainBatchRoot` for the chain. Then, the `BatchRootLeaf` will be calculated and appended to the incremental merkle tree with which the `ChainRoot` & `ChainIdLeaf` is calculated, which will be updated in the merkle tree of `ChainIdLeaf`s.
+When a ZKsync chain's batch gets executed on the settlement layer, the chain calls the `L2AssetRouter.processLogsAndMessages` function, processing all the messages within the batch (see the flow [here](../bridging/asset_tracker/asset_tracker.md#settlement-of-chains-and-interop)). After all the checks are done, the `BatchRootLeaf` will be calculated and appended to the incremental Merkle tree with which the `ChainRoot` & `ChainIdLeaf` is calculated, which will be updated in the Merkle tree of `ChainIdLeaf`s.
```
kelemeno: on GW only, on L1 the addChainBatchRoot is still called

PR #4530, opened 2025-10-08
feat: adds l1_batch_number to zks_get_l2l1_logproof

dutterbutter:
## What ❔
- adds l1_batch_number to zks_get_l2l1_logproof

## Why ❔
- zksync-os-server returns batch_number with `zks_getL2ToL1LogProof`. Adding the field allows the same SDK to be used for EraVM and zkos use. 

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/core/lib/types/src/api/mod.rs
+++ b/core/lib/types/src/api/mod.rs
@@ -258,6 +258,9 @@ pub struct L2ToL1LogProof {
     pub id: u32,
     /// The root of the tree.
     pub root: H256,
+    /// The L1 batch number where the log was included.
+    #[serde(rename = "batch_number")]
+    pub l1_batch_number: L1BatchNumber,
 }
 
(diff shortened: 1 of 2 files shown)


core/lib/types/src/api/mod.rs
```diff
@@ -258,6 +258,9 @@ pub struct L2ToL1LogProof {
     pub id: u32,
     /// The root of the tree.
     pub root: H256,
+    /// The L1 batch number where the log was included.
+    #[serde(rename = "batch_number")]
+    pub l1_batch_number: L1BatchNumber,
```
itegulov: Why do `serde(rename)`? IMHO let's just the field `batch_number`

dutterbutter: I was thinking of keeping it consistent with the other types (e.g. TransactionReceipt). i updated it nonetheless.

PR #4536, opened 2025-10-10
feat(zkstack): support genesis for zksync os

Deniallugo:
## What ❔

ZkSync OS has slightly different genesis file and the chain creation process. This PR reflects this changes  

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/zkstack_cli/crates/zkstack/src/commands/chain/create.rs
+++ b/zkstack_cli/crates/zkstack/src/commands/chain/create.rs
@@ -5,7 +5,7 @@ use xshell::Shell;
 use zkstack_cli_common::{logger, spinner::Spinner};
 use zkstack_cli_config::{
     create_local_configs_dir, create_wallets, traits::SaveConfigWithBasePath, ChainConfig,
-    EcosystemConfig, GenesisConfig, SourceFiles, ZkStackConfig, ERA_VM_GENESIS_FILE,
+    EcosystemConfig, GenesisConfig, SourceFiles, ZkStackConfig,
 };
 use zksync_basic_types::L2ChainId;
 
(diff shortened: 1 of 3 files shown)


zkstack_cli/crates/zkstack/src/commands/update.rs
```diff
@@ -48,7 +48,7 @@ pub async fn run(shell: &Shell, args: UpdateArgs) -> anyhow::Result<()> {
         .join(EN_CONFIG_FILE);
     let genesis_config_path = ecosystem
         .default_configs_path_for_ctm(vm_option)
-        .join(GENESIS_FILE);
+        .join(ERA_VM_GENESIS_FILE);
```
mm-zk, line 51: if this command is run wihth vm_option zksync os -- should we still load era VM genesis file?

zkstack_cli/crates/zkstack/src/commands/chain/create.rs
```diff
@@ -95,7 +95,7 @@ pub(crate) async fn create_chain_inner(
     );
     let genesis_config_path = ecosystem_config
         .default_configs_path_for_ctm(vm_option)
-        .join(GENESIS_FILE);
+        .join(ERA_VM_GENESIS_FILE);
```
mm-zk: shouldn't we load based on Vm option?

Deniallugo: yes, here it should.

zkstack_cli/crates/zkstack/src/commands/update.rs
Deniallugo: It's applied only to era =(

Deniallugo: it's probably one of the functions to be removed soon

PR #4563, opened 2025-10-28
feat: Improve proving networks setup

Artemka374:
## What ❔

* Add `attempts` column for submitting validation results. Initially the setup worked the way that validation results are being sent with retries, but without any measuring of attempts - so they are being sent until accepted. In some cases it gets the system stuck - like change of contracts or some other issues. So this PR adds `attempts` and metrics for reached max attempts(configurable) - for which we can configure alert and react.
* Metrics `proven_batches` and `acknowledged_batches` were reworked, so they emit the latest proven/acknowledged batch instead of total amount.
* `failed_to_send_txs` metric was removed in favor of `reached_max_attempts` metric.
* New metrics added:
    - `submitter_address` - to track the address of the submitter in Grafana, for easier accesibility
    - `submitter_balance` - to track/set alerts on balance of submitter
    - `contract_address` - address of proof manager contract - to be able to track the contract's state from Grafana
* Crypto dependencies bumped - fixes the issue with verifier panicking if receiving wrong proof. We don't want to crash the component in case the submitted proof was incorrect
* Add filtering events by chain ID of current network - so we won't receive proofs/events from other networks.

## Why ❔

This PR improves a few things in proving networks setup as preparation to going in production.

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/core/node/eth_proof_manager/src/lib.rs
+++ b/core/node/eth_proof_manager/src/lib.rs
@@ -35,6 +35,7 @@ impl EthProofManager {
                 client.clone_boxed(),
                 connection_pool.clone(),
                 blob_store.clone(),
+                l2_chain_id,
                 config.clone(),
             ),
             sender: sender::EthProofSender::new(
--- a/core/node/eth_proof_manager/src/watcher/events/proof_request_acknowledged.rs
+++ b/core/node/eth_proof_manager/src/watcher/events/proof_request_acknowledged.rs
@@ -1,7 +1,7 @@
 use anyhow::Context;
 use async_trait::async_trait;
 use zksync_dal::{eth_watcher_dal::EventType, ConnectionPool, Core, CoreDal};
-use zksync_types::{api::Log, ethabi, h256_to_u256, L1BatchNumber, H256, U256};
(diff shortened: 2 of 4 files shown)


core/node/eth_proof_manager/src/metrics.rs
```diff
@@ -19,11 +19,16 @@ pub(super) enum ValidationResult {
...
-    pub failed_to_send_tx: Family<TxType, Counter>,
     pub validated_batches: Family<ValidationResult, Counter>,
-    pub proven_batches: Family<ProvingNetwork, Counter>,
-    pub acknowledged_batches: Family<ProvingNetwork, Counter>,
+    pub proven_batches: Family<ProvingNetwork, Gauge<u64>>,
+    pub acknowledged_batches: Family<ProvingNetwork, Gauge<u64>>,
     pub fallbacked_batches: Counter<u64>,
+    pub reached_max_attempts: Family<TxType, Gauge<u64>>,
+    #[metrics(labels = ["submitter_address"])]
+    pub submitter_address: LabeledFamily<&'static str, Gauge, 1>,
```
EmilLuta, line 28: Not sure that's how I'd deal with emitting this data, but nothing against it.

EmilLuta (approved):

PR #4565, opened 2025-10-31
feat(api): add EIP-7966 eth_sendRawTransactionSync method

coffeexcoin:
## What ❔

Implement `eth_sendRawTransactionSync` API method to add [EIP-7966](https://eips.ethereum.org/EIPS/eip-7966) support to zksync era API.

`sendRawTransactionSync` uses the existing new block pubsub broadcast under the hood to check for transaction inclusion, on new blocks, while racing a user specified timeout (with default and node-configurable max) for inclusion. 

## Why ❔

With the v29 release and fast (200ms) blocks, dapps can benefit greatly by using synchronous calls for transaction submission to receipt, instead of having to poll for the receipt.

EIP-7966 has clear benefits for a number of onchain use cases:

- Trading. DEXs, CLOBs, and other trading apps can double their fee revenue by processing twice as many transactions per second, with no modification to block times or gas limits. The increased speed also allows these apps to offer tighter spreads, reduce stale quotes, and increase order execution probability – all of these improve UX.
- Gaming. Many onchain games translate rapid, repeated in-game actions (e.g. moves, item trades, crafting, etc.) into individual transactions, each of which generates a polling loop under today’s eth_sendRawTransaction paradigm. EIP-7966’s introduction of synchronous receipts reduces the load on RPCs by eliminating those loops and provides smoother gameplay for users.
- Payments. Onchain payments apps with high volumes can process more payments – and therefore generate more fees – with EIP-7966’s eth_sendRawTransactionSync method. This method also allows them to confirm payments faster and eliminate the anxiety users get waiting for a big transfer to be confirmed – a huge UX win.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Operational changes

Two new configuration values:
- `send_raw_tx_sync_default_timeout_ms`: default 2000 (2 seconds) per recommendation in EIP
- `send_raw_tx_sync_max_timeout_ms`: default 10000 (10 seconds) - prevents callers from very-long/infinite duration calls by specifying an unreasonable transaction inclusion timeout

## Checklist

### DIFF
--- a/core/node/api_server/src/web3/state.rs
+++ b/core/node/api_server/src/web3/state.rs
@@ -252,17 +252,13 @@ impl InternalApiConfig {
     }
 
     pub fn new(
-        web3_config: &Web3JsonRpcConfig,
-        state_keeper_config: &StateKeeperConfig,
+        base: InternalApiConfigBase,
         l1_contracts_config: &SettlementLayerSpecificContracts,
         l1_ecosystem_contracts: &L1SpecificContracts,
         l2_contracts: &L2Contracts,
         genesis_config: &GenesisConfig,
-        l1_to_l2_txs_paused: bool,
         settlement_layer: SettlementLayer,
     ) -> Self {
-        let base = InternalApiConfigBase::new(genesis_config, web3_config, state_keeper_config)
-            .with_l1_to_l2_txs_paused(l1_to_l2_txs_paused);
         Self::from_base_and_contracts(
             base,
             l1_contracts_config,
--- a/core/node/api_server/src/web3/tests/mod.rs
+++ b/core/node/api_server/src/web3/tests/mod.rs
@@ -62,7 +62,7 @@ use super::*;
 use crate::{
     testonly::{mock_execute_transaction, store_custom_l2_block},
     tx_sender::SandboxExecutorOptions,
-    web3::testonly::TestServerBuilder,
+    web3::{state::InternalApiConfigBase, testonly::TestServerBuilder},
 };
 
 mod debug;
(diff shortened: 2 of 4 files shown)


core/node/api_server/src/web3/namespaces/eth.rs
```diff
@@ -695,6 +695,70 @@ impl EthNamespace {
...
+                        Err(tokio::sync::broadcast::error::RecvError::Closed) => {
+                            return Err(Web3Error::TransactionUnready(
+                                "Block notification channel closed".to_string()
+                            ));
+                        }
+                    }
+                }
+            }
+        }
+    }
```
Deniallugo, line 742: why do you need `block_rx` ? 
Can we simplify the code and `get_transaction_receipt_impl` every second or better use a config variable `miniblock_commit_deadline` as an interval. 
The effect will be the same. Only if the current transaction is the only transaction and this transaction doesn't meet fee criteria this code will produce less request. but taking into account the pubsub complexity i don't think it worth it

coffee-the-dev: I assumed that a subscription was more efficient/preferred over polling; happy to adjust this if you think just polling every `miniblock_commit_deadline` is a better approach.

Deniallugo: It's much simpler and reduce the complexity

coffee-the-dev: updated this; for EN, `miniblock_commit_deadline` is not available in the config. Are you okay with me moving this into the shared statekeeper config?

core/bin/external_node/src/config/mod.rs
```diff
@@ -216,6 +216,10 @@ impl From<&LocalConfig> for InternalApiConfigBase {
             filters_disabled: web3_rpc.filters_disabled,
             l1_to_l2_txs_paused: false,
             eth_call_gas_cap: web3_rpc.eth_call_gas_cap,
+            send_raw_tx_sync_max_timeout_ms: web3_rpc.send_raw_tx_sync_max_timeout_ms,
+            send_raw_tx_sync_default_timeout_ms: web3_rpc.send_raw_tx_sync_default_timeout_ms,
+            // Default to 1 second polling interval (matches StateKeeperConfig default)
+            send_raw_tx_sync_poll_interval_ms: 1000,
```
Deniallugo: let's add it to the Local Config

Deniallugo:
@coffeexcoin LGTM, but please make sure that lints are passing

PR #4571, opened 2025-11-17
feat(eth_sender): Set fusaka upgrade timestamp

Deniallugo:
## What ❔

Use upgrade timestamp as well as block for switching to fusaka blobs

## Why ❔

Unfortunately ETH foundation doesn't provide the exact block for fusaka, but they have the exact timestamp. So it was decided to use this timestamp and half an hour before and after this as a safety margin 
https://blog.ethereum.org/2025/11/06/fusaka-mainnet-announcement#fusaka-activation

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/core/lib/config/src/configs/eth_sender.rs
+++ b/core/lib/config/src/configs/eth_sender.rs
@@ -56,6 +56,7 @@ impl EthConfig {
                 force_use_validator_timelock: false,
                 fusaka_upgrade_block: Some(0),
                 fusaka_upgrade_safety_margin: 0,
+                fusaka_upgrade_timestamp: Some(1),
             },
             gas_adjuster: GasAdjusterConfig {
                 default_priority_fee_per_gas: 1000000000,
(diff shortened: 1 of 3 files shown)


core/lib/config/src/configs/eth_sender.rs
```diff
@@ -181,8 +182,13 @@ pub struct SenderConfig {
...
     /// Use fusaka blob tx format if  the block has passed.
     pub fusaka_upgrade_block: Option<u64>,
-    #[config(default_t = 100)]
+    /// Half an hour safety margin
+    #[config(default_t = 1800)]
     pub fusaka_upgrade_safety_margin: u64,
+    /// Use fusaka blob tx format if  the timestamp has passed. Default is mainnet upgrade.
+    /// Use this value if block is not set
+    #[config(default_t = Some(1764798551))]
+    pub fusaka_upgrade_timestamp: Option<u64>,
```
perekopskiy, line 191: What will happen to sepolia nodes? (default is applied for every node)

Deniallugo: They all use block_number: 0 (otherwise they won't work already)

PR #4605, opened 2026-01-09
feat(prover): Add aggressive mode for prover_autoscaler

yorik:
## What ❔

In aggresive mode autoscaler will try to schedule missing pods in all the pools at the same time to increase chances to make pods running ASAP.

The mode activates if more than `aggressive_mode_threshold` pools received out of resources errors.

## Why ❔

To get pods running during global resource shortage.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist

ref ZKD-3182

### DIFF
--- a/zkstack_cli/crates/zkstack/src/commands/chain/gateway/gateway_common.rs
+++ b/zkstack_cli/crates/zkstack/src/commands/chain/gateway/gateway_common.rs
@@ -175,7 +175,7 @@ async fn get_batch_execution_status(
     let zk_chain_address = sl_bridgehub.get_zk_chain(U256::from(l2_chain_id)).await?;
     let zk_chain = ZkChainAbi::new(zk_chain_address, provider);
     let total_committed = zk_chain.get_total_batches_committed().await?;
-    let total_executed = zk_chain.get_total_batches_committed().await?;
+    let total_executed = zk_chain.get_total_batches_executed().await?;
 
     Ok((total_committed, total_executed))
 }


prover/crates/bin/prover_autoscaler/README.md
```diff
@@ -159,6 +177,12 @@ agent_config:
   cluster. Default: 10m.
 - `scale_errors_duration` defines the time window for including scale errors in Autoscaler calculations. Clusters will
   be sorted by number of the errors. It should be between 20m and 2h. Default: 1h.
+- `aggressive_mode_threshold` is a percentage (0-100) of pools with "GCE out of resources" errors needed to trigger
```
EmilLuta, line 180: I somehow think of this threshold exactly reverse. If at least X% of machines have issues, then go into aggressive mode. With that reading, 0% would be mean always on (the threshold is crossed at 0, which means if you have 0 or more machines erroring).

Might be worthwhile to rename or change the functionality, but treat it as optional.

prover/crates/bin/prover_autoscaler/README.md
```diff
@@ -159,6 +177,12 @@ agent_config:
...
+  aggressive mode. Set to 0 to disable (default). For example, 50 means aggressive mode activates when 50% of all pools
+  have resource shortage errors. In aggressive mode, the scaler requests pods from ALL pools simultaneously to speed up
+  resource acquisition. Default: 0 (disabled).
+- `aggressive_mode_cooldown` is the duration to stay in aggressive mode after successfully obtaining resources. This
+  cooldown prevents rapid oscillation between normal and aggressive modes. Default: 10m.
```
EmilLuta, line 185: I'm not sure how I'd pick this value. How did you land up on 10m? If I were to pick a random number, I'd say 30m, but would love to hear more of your thoughts.

P.S. No change required, just trying to understand reasoning.

prover/crates/bin/prover_autoscaler/src/global/scaler.rs
```diff
@@ -272,6 +278,256 @@ impl<K: Key> Scaler<K> {
...
 
+    /// Check if we should enter/stay in aggressive mode based on percentage of pools with errors
+    /// Also handles deactivation when cooldown expires
+    fn is_aggressive_mode(&self, pools: &[Pool<K>], total_running: usize, queue: usize) -> bool {
+        // If threshold is 0, aggressive mode is disabled
+        if self.config.aggressive_mode_threshold == 0 {
+            return false;
+        }
+
+        if let Ok(mut guard) = self.aggressive_mode_active.lock() {
```
EmilLuta: Shouldn't we add some sort of failing/panicing/erroring in the scenario where the lock is poisoned? Not sure if we should continue execution at all.

prover/crates/bin/prover_autoscaler/src/global/scaler.rs
```diff
@@ -272,6 +278,256 @@ impl<K: Key> Scaler<K> {
...
+                    // Still in cooldown period with enough resources - stay in aggressive mode
+                    return true;
+                }
+
+                // Don't have enough resources yet - stay in aggressive mode regardless of time
+                return true;
+            }
+        }
+
+        if pools.is_empty() {
```
EmilLuta: When exactly could this happen?

prover/crates/bin/prover_autoscaler/src/global/scaler.rs
```diff
@@ -272,6 +278,256 @@ impl<K: Key> Scaler<K> {
...
+    /// Check if we should enter/stay in aggressive mode based on percentage of pools with errors
+    /// Also handles deactivation when cooldown expires
+    fn is_aggressive_mode(&self, pools: &[Pool<K>], total_running: usize, queue: usize) -> bool {
+        // If threshold is 0, aggressive mode is disabled
+        if self.config.aggressive_mode_threshold == 0 {
+            return false;
+        }
+
+        if let Ok(mut guard) = self.aggressive_mode_active.lock() {
+            if let Some(activated_at) = *guard {
```
EmilLuta: I'd leave a comment here -- this branch deals with "already active" mode.

prover/crates/bin/prover_autoscaler/src/global/scaler.rs
```diff
@@ -272,6 +278,256 @@ impl<K: Key> Scaler<K> {
...
+            }
+        }
+
+        total
+    }
+
+    /// Aggressive mode calculation: add pods to ALL pools simultaneously
+    fn calculate_aggressive(
+        &self,
+        _namespace: &NamespaceName,
```
EmilLuta: Is this needed?

prover/crates/bin/prover_autoscaler/src/global/scaler.rs
```diff
@@ -272,6 +278,256 @@ impl<K: Key> Scaler<K> {
         queue.div_ceil(speed) * speed
     }
 
+    /// Check if we should enter/stay in aggressive mode based on percentage of pools with errors
+    /// Also handles deactivation when cooldown expires
+    fn is_aggressive_mode(&self, pools: &[Pool<K>], total_running: usize, queue: usize) -> bool {
```
EmilLuta, line 300: One thing I struggled to understand for a bit, you have the concept of aggressive_mode but that only triggers after we acquire the nodes. Before that, it's not really clear what state we're in. I reckon I'd introduce some `enum OperationMode` that states what we're doing. As options I imagine: `Regular`, `AggressiveMode`, `CooldownAggressiveMode(DateTime<Utc>)`.

Unclear if this is much better, but might read easier.

prover/crates/bin/prover_autoscaler/src/global/scaler.rs
```diff
@@ -272,6 +278,256 @@ impl<K: Key> Scaler<K> {
...
+        // If threshold is 0, aggressive mode is disabled
+        if self.config.aggressive_mode_threshold == 0 {
+            return false;
+        }
+
+        if let Ok(mut guard) = self.aggressive_mode_active.lock() {
+            if let Some(activated_at) = *guard {
+                // Check if we got enough Running resources and should start cooldown
+                if total_running >= queue {
```
EmilLuta: IIRC, we had some sort of normalizing values (like, # of pods -> # of jobs it can do per unit of time). I see here we use `total_running` which IIUC corresponds to the normalized value. If so, `total_running` might be a bit misleading (as it implies pods, not proving capacity).

prover/crates/bin/prover_autoscaler/src/global/scaler.rs
```diff
@@ -272,6 +278,256 @@ impl<K: Key> Scaler<K> {
...
+    /// Aggressive mode calculation: add pods to ALL pools simultaneously
+    fn calculate_aggressive(
+        &self,
+        _namespace: &NamespaceName,
+        queue: usize,
+        sorted_clusters: Vec<Pool<K>>,
+    ) -> HashMap<PoolKey<K>, usize> {
+        let mut pods: HashMap<PoolKey<K>, usize> = HashMap::new();
+        let mut total_running: i64 = 0;
+        let mut total_capacity: i64 = 0; // Running + Pending
```
EmilLuta: Can't this be usize? i see you convert back and forth between these types.

prover/crates/bin/prover_autoscaler/src/global/scaler.rs
```diff
@@ -272,6 +278,256 @@ impl<K: Key> Scaler<K> {
...
+
+            if total_in_pool > 0 {
+                pods.insert(cluster.to_key(), total_in_pool);
+                total_running += self.pods_to_speed(cluster.key, running) as i64;
+                total_capacity += self.pods_to_speed(cluster.key, total_in_pool) as i64;
+            }
+        }
+
+        // Log current state and handle cooldown timer
+        self.start_aggressive_cooldown(total_running as usize, queue);
```
EmilLuta: Aggressive mode doesn't really trigger here if we don't have enough pods. Can this cause problems down the line? I'm assuming that it will run again and try to retrigger it after waiting for a while (whatever is the speed of the loop). Thought I should mention it as it looks a bit weird from naming (it states that it starts the cooldown, but it doesn't most of the times, really -- maybe a better name is set_cooldown_if_required_is_met or something).

prover/crates/bin/prover_autoscaler/src/global/scaler.rs
```diff
@@ -272,6 +278,256 @@ impl<K: Key> Scaler<K> {
...
+        }
+
+        // Count ALL pools with out_of_resources issues (NeedToMove or scale_errors)
+        let pools_with_errors = pools
+            .iter()
+            .filter(|p| p.sum_by_pod_status(PodStatus::NeedToMove) > 0 || p.scale_errors > 0)
+            .count();
+
+        let total_pools = pools.len();
+        let error_percentage = (pools_with_errors * 100) / total_pools;
```
EmilLuta: Can `total_pools` ever be 0 here?

EmilLuta (commented):
Left a bunch of questions.

The only thing I'm concerned with is the fact that we don't treat potential mutex poison.

prover/crates/bin/prover_autoscaler/src/global/scaler.rs
yorik: When there is no pools in the config or something went completely wrong. This is prevents div by 0 11 lines below. It's better than panic, I think.

prover/crates/bin/prover_autoscaler/src/global/scaler.rs
yorik: Fixed

prover/crates/bin/prover_autoscaler/src/global/scaler.rs
yorik: Done

prover/crates/bin/prover_autoscaler/src/global/scaler.rs
yorik: Removed

prover/crates/bin/prover_autoscaler/src/global/scaler.rs
yorik: Changed to usize. Before they could be negative, not needed anymore.

prover/crates/bin/prover_autoscaler/src/global/scaler.rs
yorik: Nope, pool checked to be non empty ~10 lines above.

prover/crates/bin/prover_autoscaler/README.md
yorik: It's not about machines, it's about pools. 50% means that half of the pools should be touched and have scale errors. Using 0 as magic number which disables it sounds ok for me: if you want aggressive mode being always on, just set it to 1%. Other option would be to set it to > 100% (e.g. 101% to disable it naturally), but it sounds too confusing to me.

prover/crates/bin/prover_autoscaler/README.md
yorik: It's pretty random: we already recovered and waiting for 10 or 30m doesn't change much. To know a good value need to monitor it in prod during aggressive mode.

prover/crates/bin/prover_autoscaler/src/global/scaler.rs
yorik: Good idea. Done

prover/crates/bin/prover_autoscaler/src/global/scaler.rs
yorik: `total_running` is special version of `total` without pending, which is in jobs, not pods. Probably not the best naming, but I would change it everywhere if you'll insist. (see line 607)

prover/crates/bin/prover_autoscaler/src/global/scaler.rs
yorik: I've renamed it to `update_operation_mode`. It should be triggered only from aggressive mode, which is reachable if `is_aggressive_mode` returns true.

prover/crates/bin/prover_autoscaler/README.md
EmilLuta: Alright, let's leave it as is.

prover/crates/bin/prover_autoscaler/README.md
EmilLuta: Agreed, my value is equally random. Let's leave it as is.

PR #4606, opened 2026-01-09
fix: use hashed_keys for extracting storage logs from db

sanekmelnikov:
## What ❔
When recovering lost storage logs using compressed state diffs from L1, only the `hashed_key` is available. As a result, the backfilled postgres data in `storage_logs` won't have the `address` and `key` fields.

The current logic causes tree recovery to fail [here](https://github.com/matter-labs/zksync-era/blob/core-v29.6.0/core/node/metadata_calculator/src/helpers.rs#L804), as the query [expects](https://github.com/matter-labs/zksync-era/blob/main/core/lib/dal/src/storage_logs_dal.rs#L366-L367) these fields to be present.
(Note that `tree_writes` field in `l1_batches` isn't available in this case as well - it can't be recovered because it is decoded using `address` and `key`)

This PR replaces `get_touched_slots_for_executed_l1_batch` with [get_touched_slots_for_l1_batch](https://github.com/matter-labs/zksync-era/blob/main/core/lib/dal/src/storage_logs_dal.rs#L316) to use `hashed_key` instead.
To ensure the order of writes is preserved, we sort values by `leaf_index`.

## Why ❔
To enable L1 recovery

## Is this a breaking change?
- [ ] Yes
- [x] No

## Checklist


perekopskiy (dismissed):
Looks good!

core/node/metadata_calculator/src/helpers.rs
```diff
@@ -801,46 +801,56 @@ impl L1BatchWithLogs {
...
             // in the considered L1 batch. Thus, we can remove them from `touched_slots` in order to simplify
             // their further processing. This is not a required step; the logic below works fine without it.
             // Indeed, extra no-op updates that could be added to `storage_logs` as a consequence of no filtering,
             // are removed on the Merkle tree level (see the tree domain wrapper).
-            let log = TreeInstruction::Read(storage_key.hashed_key_u256());
-            storage_logs.insert(storage_key, log);
+            if let Some(&(initial_write_batch_for_key, leaf_index)) =
+                l1_batches_for_initial_writes.get(&hashed_key)
+            {
+                if initial_write_batch_for_key <= l1_batch_number {
```
perekopskiy: nit: I'd assert! that these 2 ifs are entered for each protective read

deniallugo-ml (approved):

PR #4613, opened 2026-01-16
fix(gateway_migration): Add wait for starting migration

Deniallugo:
## What ❔

Check that we don't have any inflight blocks with interop roots 

## Why ❔

During the gateway migration. 
Every step is important. And currently it was not taken into account that some blocks could have interop roots, that were not sent to the SL. 

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist

### DIFF
--- a/core/node/api_server/src/web3/namespaces/unstable/mod.rs
+++ b/core/node/api_server/src/web3/namespaces/unstable/mod.rs
@@ -250,26 +250,26 @@ impl UnstableNamespace {
             .await
             .map_err(DalError::generalize)?;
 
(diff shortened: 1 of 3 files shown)


core/node/api_server/src/web3/namespaces/unstable/mod.rs
```diff
@@ -248,6 +250,32 @@ impl UnstableNamespace {
             .await
             .map_err(DalError::generalize)?;
 
+        let latest_processed_l1_batch_number = connection
```
dimazhornyk: is this used?

PR #4630, opened 2026-01-28
feat: add support for Base Token Holder contract in server

Raid5594:
## What ❔

Add support for Base Token Holder contract in server.

## Why ❔

Base tokens have to be minted to the contract at genesis. 

## Is this a breaking change?
- [x] Yes
- [ ] No

## Checklist

### DIFF
--- a/core/lib/multivm/src/versions/testonly/l1_tx_execution.rs
+++ b/core/lib/multivm/src/versions/testonly/l1_tx_execution.rs
@@ -137,8 +137,8 @@ pub(crate) fn test_l1_tx_execution<VM: TestedVm>() {
     assert!(result.result.is_failed(), "The transaction should fail");
 
     let res = StorageWritesDeduplicator::apply_on_empty_state(&result.logs.storage_logs);
-    assert_eq!(res.initial_storage_writes, basic_initial_writes + 1);
-    assert_eq!(res.repeated_storage_writes, 1);
+    assert_eq!(res.initial_storage_writes, basic_initial_writes);
+    assert_eq!(res.repeated_storage_writes, 2);
 }
 
 pub(crate) fn test_l1_tx_execution_high_gas_limit<VM: TestedVm>() {
--- a/core/lib/multivm/src/versions/testonly/tester/mod.rs
+++ b/core/lib/multivm/src/versions/testonly/tester/mod.rs
@@ -1,7 +1,7 @@
 use std::{collections::HashSet, fmt, rc::Rc};
 
 use zksync_contracts::BaseSystemContracts;
-use zksync_system_constants::L2_ASSET_TRACKER_ADDRESS;
+use zksync_system_constants::{BASE_TOKEN_HOLDER_ADDRESS, L2_ASSET_TRACKER_ADDRESS};
 use zksync_test_contracts::{Account, TestContract, TxType};
 use zksync_types::{
     get_deployer_key,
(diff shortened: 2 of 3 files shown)


etc/env/file_based/genesis.yaml
```diff
@@ -1,5 +1,17 @@
+genesis_protocol_semantic_version: 0.31.0
```
Raid5594: This file might be irrelevant to PR

core/lib/types/src/storage/mod.rs
```diff
@@ -149,3 +149,32 @@ pub fn get_system_context_init_logs(chain_id: L2ChainId) -> Vec<StorageLog> {
         ),
     ]
 }
+
+/// The initial balance of the BaseTokenHolder contract (2^127 - 1).
+/// This value is used in L2BaseToken.totalSupply() calculation:
+/// totalSupply = INITIAL_BASE_TOKEN_HOLDER_BALANCE - balance[BASE_TOKEN_HOLDER_ADDRESS]
+pub const INITIAL_BASE_TOKEN_HOLDER_BALANCE: U256 = U256([
```
kelemeno: can't you set this in the genesis upgrade? but it might be ok

etc/env/file_based/genesis.yaml
```diff
@@ -1,5 +1,17 @@
...
+genesis_protocol_version: null
+genesis_root: 0x9470ae067ffd2b8661281c0d7d5a6f57697a1d32cd352f41efea06c696c4136a
+genesis_rollup_leaf_index: 107
+genesis_batch_commitment: 0x19df9a1ed1b5ddf93f41c38be7cfca4ab35aba92ba926ec7b03a8e3422729b64
```
kelemeno: these were removed, you need to run zkstackup --local probably and do regensis. Changes should be minimal here

zkstack_cli/crates/config/src/genesis.rs
```diff
@@ -99,7 +99,7 @@ impl GenesisConfigPatch {
                 self.0.insert("evm_emulator_hash", data)?;
             }
             self.0.insert(
-                "prover.recursion_scheduler_level_vk_hash",
+                "prover.snark_wrapper_vk_hash",
```
kelemeno: why is this change needed?

core/lib/types/src/storage/mod.rs
Raid5594: Yes, we could set it during genesis upgrade, but I don't think there will be any practical difference.

zkstack_cli/crates/config/src/genesis.rs
Raid5594: #[derive(Debug, Clone, Default, Eq, PartialEq, Serialize, Deserialize)]
pub struct L1VerifierConfig {
    // Rename is required to not introduce breaking changes in the API for existing clients.
    #[serde(
        alias = "recursion_scheduler_level_vk_hash",
        rename(serialize = "recursion_scheduler_level_vk_hash")
    )]
    pub snark_wrapper_vk_hash: String,
    pub fflonk_snark_wrapper_vk_hash: Option<String>,
}


It is using aliasing for reading, so we'll just write the renamed version straightaway

StanislavBreadless: Does it make sense to remove the alias then?

PR #4723, opened 2026-03-18
fix(eth-proof-manager): three correctness fixes for validation retries, metrics labels, and verifier panics

Artemka374-claude:
Three bug fixes for the `eth_proof_manager` node component.

## Fix 1 — Validation tx retries after first tx was already confirmed

When `send_tx` timed out polling for a receipt, it discarded the broadcast hash. The retry re-queried the nonce, saw it had advanced (original tx was mined), and sent a new tx against the next nonce slot — which reverted on-chain since the validation was already submitted. The batch stayed `Proven` in the DB and the loop kept retrying until `validation_tx_attempts` was exhausted.

`send_tx` now returns the broadcast hash alongside any receipt-timeout error. `send_tx_with_retries` checks all pending hashes for confirmation before sending a replacement.

## Fix 2 — Truncated addresses in Prometheus labels

`Address` (`H160`) `Display` truncates to `0x<2 bytes>…<2 bytes>`. Both address label values were set via `.to_string()`, producing e.g. `0x1a2b…ef01` instead of the full address. Changed to `format!("{:#x}", addr)`.

## Fix 3 — Verifier and event-handler panics crash the component

`fflonk::verify` can panic on malformed proof data (field elements out of range). Several `panic!` calls in both event handlers fired on unexpected ABI token values or unknown `ProvingNetwork` discriminants. All panics inside the task caused a restart loop instead of marking the proof invalid.

- `fflonk::verify` wrapped in `std::panic::catch_unwind(AssertUnwindSafe(...))`.
- `panic!` calls in handlers replaced with `anyhow::bail!`.
- `ProvingNetwork::from_u256` changed to return `anyhow::Result<Self>`.

### DIFF
--- a/core/node/eth_proof_manager/src/client.rs
+++ b/core/node/eth_proof_manager/src/client.rs
@@ -122,7 +122,37 @@ impl ProofManagerClient {
         let mut prev_priority_fee_per_gas: Option<u64> = None;
         let mut last_error = None;
 
+        // Hashes of transactions we broadcast successfully but whose receipt we did not
+        // observe within the polling window. We keep them so that we can detect the case
+        // where one of those transactions was actually mined between our polling and the
+        // next retry attempt. Without this check we would re-query the nonce, see it has
+        // advanced (the original tx was mined), build a new tx with the incremented nonce,
+        // and that tx would revert on-chain because the operation was already completed.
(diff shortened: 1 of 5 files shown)


EmilLuta (approved):
Overall LGTM. I assume you've already tested this path. If so, feel free to merge.

PR #4746, opened 2026-03-29
fix(prover-autoscaler): fix event watcher and running weight cap

hatemosphere:
## Summary
- **Event watcher field selectors**: Use server-side `fieldSelector` on event watchers (`reason=FailedScheduling`, `reason=FailedScaleUp`) to filter at the API server. Previously 18K+ events flooded the initial LIST, overwhelming the watch stream so the agent never processed GPU exhaustion events — aggressive mode and H100 fallback never triggered.
- **GPU_EVENT_MATCHERS**: Refactor event patterns into a single constant used for both watch stream setup and handler logic.
- **Running weight cap**: Three-tier behavior based on global `total_running_weight` across all namespaces:
  - `running < max_running_weight` → no cap, desired uncapped, scale freely
  - `running >= max_running_weight` → cap desired to `max_running_weight` (stop scaling up, only Pending pods removed, Running pods stay)
  - `running >= max_running_weight + burst` → cap desired to `max + burst` (scale down Running pods to that level)
  
  The cap is a shared counter across namespaces (busiest first). On each 60s cycle, `total_running_weight` is recomputed — if running drops below max, desired uncaps and scaling resumes.
- **Fix metrics labels**: `target_max_weight` and `target_max_running_weight` are global caps (per deployment), not per-namespace.

## Context
Follow-up to #4744. On mainnet2, all L4 GPU pools were exhausted but the scaler couldn't fall back to H100 because:
1. The event watcher never delivered FailedScheduling events (18K initial LIST overwhelmed the stream)
2. The weight cap consumed the entire desired budget with Pending L4 pods, trimming H100 to 0

## Test plan
- [x] All 20 tests pass, including:
  - `test_run_no_cap_scales_freely` — no cap passed → desired uncapped
  - `test_run_no_cap_when_under_max` — running < max → no cap; running >= max → cap to max; running >= max+burst → cap to max+burst
  - `test_run_cap_shared_across_namespaces` — cap shared across namespaces, ns1 consumes cap, ns2 gets nothing
- [x] Event field selectors confirmed: 5K+ GPU events processed (was 0 before)
- [x] H100 fallback confirmed: 91 H100 on use4, 16 on usc1 when L4 exhausted

### DIFF
--- a/prover/crates/bin/prover_autoscaler/src/global/manager.rs
+++ b/prover/crates/bin/prover_autoscaler/src/global/manager.rs
@@ -147,24 +147,46 @@ impl Task for Manager {
                     })
                     .collect();
 
+                // Busiest namespace first so it gets priority for the shared cap.
                 namespace_queues.sort_by(|(ns_a, _, q_a), (ns_b, _, q_b)| {
                     q_b.cmp(q_a).then_with(|| ns_a.cmp(ns_b))
                 });
 
-                let mut remaining_desired_weight =
-                    scaler.max_desired_weight().map(|max| max as i64);
+                // Compute total running weight across all namespaces.
+                let total_running_weight: usize = namespace_queues
+                    .iter()
+                    .map(|(ns, _, _)| scaler.current_running_weight(ns, &guard.clusters))
+                    .sum();
 
-                for (ns, ppv, q) in namespace_queues {
+                // Cap desired based on total running weight:
+                // - running < max: no cap, scale freely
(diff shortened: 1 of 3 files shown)


EmilLuta (dismissed):
Disagree that the description & code match, but this would be an issue only during protocol upgrades. In interest of speed, we'll merge as is and if we identify issues in the future, we can revisit.

EmilLuta (dismissed):

EmilLuta (approved):

PR #4750, opened 2026-03-31
fix(prover-autoscaler): reliable GPU fallback via pool capping, aggressive mode refactor, and deployment-level stuck detection

hatemosphere:
## Summary

Fixes GPU exhaustion handling so the autoscaler reliably detects stuck pools (L4 unavailable) and falls back to H100, without killing non-GPU scalers like the compressor.

### 1. Per-pool cap modes (FreezeAtRunning / ScaleDown)

Replaces the old shared desired budget with per-pool cap logic controlled by the manager:

- **No cap** (`running < max_running_weight`): scale freely.
- **`FreezeAtRunning`** (`running >= max`): each pool's `desired = min(desired, running)`. Kills Pending pods, preserves all Running pods and GCE nodes.
- **`ScaleDown`** (`running >= max + burst`): freeze at running first, then trim Running from lowest-priority pools until total ≤ target.

The old budget approach consumed shared desired capacity with higher-priority L4 pools, starving H100 even when L4 was stuck.

### 2. Aggressive mode evaluated globally, not per-namespace

`evaluate_aggressive_mode()` is called once per cycle in the manager with **aggregated** `total_queue` and `total_running_weight` across all namespaces.

Previously, prover-blue (queue=0) triggered "Resources obtained" (`running(0) >= queue(0)`) and entered AggressiveCooldown on the shared mode state, exiting aggressive mode while prover-red still had 300K+ queue.

### 3. Pool capping on stuck pools

Three signals cap a pool's `max_pool_size` to `Running + Pending`, forcing queue overflow to the next-priority pool:

- **NeedToMove** — pod-level, set by FailedScaleUp events (existing)
- **LongPending** — pod pending longer than `long_pending_duration` (new)
- **deployment_stuck** — deployment has `desired > 0 && running < desired` for longer than `long_pending_duration` (new, survives pod recycling from spot preemptions)

`stuck_since` only clears when the pool actually recovers (`running >= desired` with `desired > 0`), NOT when our scaler caps desired to 0.

### 4. Aggressive mode bypasses runtime caps via `configured_max_pool_size`

Pool struct now carries both `max_pool_size` (runtime-capped) and `configured_max_pool_size` (from config). The two allocation paths use them differently:

- **Regular mode**: uses `max_pool_size` — stuck pools are capped, queue overflows to fallback pools.
- **Aggressive mode**: uses `configured_max_pool_size` — bypasses runtime caps, allocates to ALL configured pools at their original limits.

This also fixes the aggressive mode active-pool filter: it now uses `configured_max_pool_size > 0` to determine which pools are "active", so runtime-capped pools still count toward the error threshold. Previously, all pools capped to 0 were excluded as "inactive", preventing aggressive mode from triggering — this caused a deadlock for NoKey scalers (compressor, witness-generators) where all pools were stuck with no recovery path.

### 5. Event watcher: FailedScaleUp only

Removed `FailedScheduling` matcher — its message varies based on GPU node availability ("Insufficient nvidia.com/gpu" only when some GPU nodes exist, "untolerated taints" otherwise). `FailedScaleUp` with "GCE out of resources" is the definitive signal.

### Config (via gitops, not in this PR)
- `max_running_weight` / `max_desired_burst_weight` — FreezeAtRunning/ScaleDown thresholds
- `long_pending_duration` — how long before deployment_stuck triggers (default 10m)
- `aggressive_mode_threshold` — percentage of active pools with errors to enter aggressive mode

## Test plan
- [x] 24 tests pass (19 existing + 5 new)
- [x] `test_deployment_stuck_caps_pool` — stuck_since > long_pending_duration caps pool, overflows to H100
- [x] `test_deployment_not_stuck_when_recent` — stuck_since < threshold does NOT cap
- [x] `test_evaluate_aggressive_mode_transitions` — full Regular → Aggressive → AggressiveCooldown → Regular cycle
- [x] `test_freeze_at_running_preserves_running_pods` — FreezeAtRunning prevents scale-up beyond running
- [x] `test_long_pending_caps_pool` — LongPending pods cap pool, overflow to H100
- [x] Deployed to mainnet2, confirmed deployment_stuck detection, L4 capping, H100 fallback, and compressor recovery via aggressive mode


EmilLuta (approved):
Blind approve over already deployed to production changes.

PR #4756, opened 2026-04-02
feat(core): lay groundwork for v31 interop (additive schema, fee fallback)

zkzoomer:
## What ❔

Adds the v31 interop data-model groundwork:
- Additive `l1_batches` migration for:
  - `settlement_layer_type`
  - `settlement_layer_chain_id`
  - `interop_fee`
- Wires new fields through DAL / storage / sync / state-keeper / VM execution.
- Adds `en_getInteropFee` and EN polling with graceful fallback when method is missing.
- Updates API block arg resolution to use settlement-layer-aware context.

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


core/lib/dal/migrations/20250904115232_settlement_layer_and_interop_fee_to_l1_batches.up.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE l1_batches ADD COLUMN settlement_layer_type TEXT NOT NULL DEFAULT 'L1';
+ALTER TABLE l1_batches ADD COLUMN settlement_layer_chain_id BIGINT NOT NULL DEFAULT 0;
```
Deniallugo: In code you use `L1(SLChainId(1))` as a default. should we use the same?

core/node/state_keeper/src/keeper.rs
```diff
@@ -40,6 +41,7 @@ struct InitializedBatchState {
     batch_executor: Box<dyn BatchExecutor<OwnedStorage>>,
     protocol_upgrade_tx: Option<ProtocolUpgradeTx>,
     next_block_should_be_fictive: bool,
+    _settlement_layer: SettlementLayer,
```
Deniallugo: why is it here?

core/lib/dal/migrations/20250904115232_settlement_layer_and_interop_fee_to_l1_batches.up.sql
```diff
@@ -0,0 +1,4 @@
...
+ALTER TABLE l1_batches ADD COLUMN interop_fee BIGINT NOT NULL DEFAULT 0;
+ALTER TABLE l1_batches ADD CONSTRAINT l1_batches_settlement_layer_type_check CHECK (settlement_layer_type IN ('L1', 'Gateway'));
```
Deniallugo: IMHO: 
I'm not a fan of enums or such checks on the database level. it is always painful to modify it later. 
But if you feel comfortable, that we will never change, i'm not fully against it

core/lib/vm_executor/src/oneshot/env.rs
```diff
@@ -35,9 +39,24 @@ impl<C: ContractsKind> OneshotEnvParameters<C> {
...
         }
     }
 
+    pub fn set_interop_fee_fallback(&mut self, interop_fee: U256) {
+        self.interop_fee_fallback = Some(Arc::new(AtomicU64::new(interop_fee.as_u64())));
+    }
+
+    pub fn set_interop_fee_fallback_provider(&mut self, interop_fee: Arc<AtomicU64>) {
+        self.interop_fee_fallback = Some(interop_fee);
+    }
```
Deniallugo: Looks a bit odd for the interfaces to have 2 setters, and they are not using each other. 

You could probably introduce a small interface for it. 

Now you have a very high chance to update only one of the setters.

core/node/api_server/src/node/tx_sender/main_node_interop_fee_update_task.rs
```diff
@@ -0,0 +1,69 @@
...
+};
+use zksync_web3_decl::{
+    client::{DynClient, L2},
+    jsonrpsee,
+    namespaces::EnNamespaceClient as _,
+};
+
+#[derive(Debug)]
+pub(super) struct MainNodeInteropFeeUpdateTask {
+    interop_fee: Arc<AtomicU64>,
```
Deniallugo: Let's introduce a new type for InteropFee. It'd allow us to not mess with this type. Now, it's not profitable, just because we never use AtomicU64 in the code. But it'd be much cleaner with a proper new type with necessary `From`. See the comment above

core/node/api_server/src/tx_sender/mod.rs
```diff
@@ -157,6 +158,17 @@ impl SandboxExecutorOptions {
         self.fast_vm_mode = fast_vm_mode;
     }
 
+    pub fn set_interop_fee_fallback(&mut self, interop_fee: U256) {
+        self.estimate_gas.set_interop_fee_fallback(interop_fee);
+        self.eth_call.set_interop_fee_fallback(interop_fee);
+    }
+
+    pub fn set_interop_fee_fallback_provider(&mut self, interop_fee: Arc<AtomicU64>) {
+        self.estimate_gas
```
Deniallugo: looks like, i've seen these methods somewhere else ;)

core/node/state_keeper/src/interop_fee.rs
```diff
@@ -0,0 +1,28 @@
...
+
+#[async_trait]
+pub trait InteropFeeInputProvider: fmt::Debug + Send + Sync + 'static {
+    async fn get_interop_fee(&self) -> Result<U256>;
+}
+
+#[derive(Debug, Clone)]
+pub struct ConstantInteropFeeInputProvider {
+    fee: U256,
+}
```
Deniallugo: is it for tests?

core/node/state_keeper/src/node/mempool_io.rs
```diff
@@ -134,6 +135,10 @@ impl WiringLayer for MempoolIOLayer {
             mempool_fetcher_pool,
         );
 
+        let interop_fee_input_provider = Arc::new(ConstantInteropFeeInputProvider::new(
```
Deniallugo: This one is extremely unexpected. So we don't actually calculate the fees.

core/node/api_server/src/node/tx_sender/main_node_interop_fee_update_task.rs
zkzoomer: Made it plain u64 in 1a907470e442b335d1dc8fb179abb25a453f115d

core/lib/vm_executor/src/oneshot/env.rs
zkzoomer: Using just one setter 331488fcb6a1372e77b73205e54ffa03b8b3f47e

core/node/api_server/src/tx_sender/mod.rs
zkzoomer: Simplified in 7d5e599a4f1ade6ac0f34d0d46791ce7a14ede93

core/lib/dal/migrations/20250904115232_settlement_layer_and_interop_fee_to_l1_batches.up.sql
zkzoomer: Resolved in c1b4fd699fd628157b02f907602e35cfc43008ab

core/node/state_keeper/src/keeper.rs
zkzoomer: Removed in f6afa7c9b882c8bfe878fcc8dc2bedefea14a919

core/lib/dal/migrations/20250904115232_settlement_layer_and_interop_fee_to_l1_batches.up.sql
zkzoomer: Fixed in 4071f17e0f0dbf711a12eba3f7aa8909d34e4f40

core/node/api_server/src/web3/state.rs
```diff
@@ -31,7 +31,7 @@ use zksync_shared_resources::{
     tree::TreeApiClient,
 };
 use zksync_types::{
-    api, commitment::L1BatchCommitmentMode, l2::L2Tx, settlement::SettlementLayer,
+    api, commitment::L1BatchCommitmentMode, l2::L2Tx, settlement::WorkingSettlementLayer,
```
Deniallugo, line 34: As i remember. I introduced concepts of different types of settlement layers, when it was possible to work with unknown settlement layer. 
But currently it's impossible and seems we have to keep only settlement layer

core/node/api_server/src/tx_sender/mod.rs
```diff
@@ -101,22 +78,20 @@ pub async fn build_tx_sender(
...
     .await?;
-    executor_options.set_interop_fee_provider(Arc::new(ConstantInteropFeeProvider::new(
-        tx_sender_config.interop_fee,
-    )));
+    executor_options.set_batch_fee_input_provider(batch_fee_input_provider.clone());
     let tx_sender = tx_sender_builder.build(
-        Arc::new(batch_fee_input_provider),
+        batch_fee_input_provider,
         Arc::new(vm_concurrency_limiter),
         executor_options,
```
Deniallugo: why do you need `batch_fee_input_provider` twice?

core/node/api_server/src/node/tx_sender.rs
```diff
@@ -184,6 +184,7 @@ impl WiringLayer for TxSenderLayer {
             config.validation_computational_gas_limit,
         )
         .await?;
+        executor_options.set_batch_fee_input_provider(fee_input.clone());
```
Deniallugo: Why you haven't passed it in new?

Deniallugo (commented):
lgtm

core/node/api_server/src/node/tx_sender.rs
zkzoomer: Added in 8b9eb99cf2433fc1af807c3813a761fb85140bf2

core/node/api_server/src/tx_sender/mod.rs
zkzoomer: It's the same `Arc<ApiFeeInputProvider>` shared across:
- `TxSender` uses it for normal fee reads as before
- Sandbox env building now uses it to provide a fallback for interop fee: https://github.com/matter-labs/zksync-era/blob/8b9eb99cf2433fc1af807c3813a761fb85140bf2/core/node/api_server/src/execution_sandbox/execute.rs#L347
The fallback is used as a safety net (eg no unsealed batch, target batch not sealed):
https://github.com/matter-labs/zksync-era/blob/8b9eb99cf2433fc1af807c3813a761fb85140bf2/core/lib/vm_executor/src/oneshot/block.rs#L240-L260

Deniallugo (approved):

PR #4761, opened 2026-04-07
fix(contract-verifier): allow relative imports

coffeexcoin:
## What ❔
This PR fixes contract verification for Solidity standard JSON inputs that contain legitimate relative imports using `../`, such as OpenZeppelin upgradeable contracts.

The verifier previously rejected any import string containing `../` as a dangerous traversal path, even when the imported file was part of the submitted `sources` set. This PR narrows that validation to reject only absolute imports such as `/...` and `file://...`, while keeping source-key validation and compiler sandboxing in place.

It also adds regression tests for:
- valid OpenZeppelin-style relative imports within submitted sources
- attempted relative-path escape outside the submitted source tree

## Why ❔
Relative imports with `../` are normal Solidity practice, especially in widely used libraries like OpenZeppelin upgradeable. Rejecting them made valid deployed contracts unverifiable through the zkSync contract verifier API.

The security goal of preventing filesystem escape is still valid, but the previous implementation was too broad. The correct boundary is:
- reject submitted source keys that escape the virtual source tree
- reject absolute import paths that target the host filesystem
- allow relative imports that resolve within the submitted source set
- rely on the compiler sandbox to ensure unresolved escape attempts cannot read host files

This keeps verification compatible with standard Solidity projects without weakening the filesystem isolation model.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Operational changes
No config changes, flags, or script behavior changes for verifier operators.

One note from local development: the repository push-format hook in this workspace could not complete because `zkstack dev fmt` expects a `contracts` prettier script that is not available in this checkout. This does not affect runtime behavior of the verifier change itself.

## Checklist


core/lib/contract_verifier/src/compilers/solc.rs
Deniallugo, line 242: Hey @coffeexcoin let's also add
```suggestion
            .current_dir(compile_dir.path())
```

Deniallugo: @coffeexcoin  sorry, stupid github UI made a wrong suggestion, it was suppose to add line not replace it..

Deniallugo (approved):

PR #4772, opened 2026-04-11
fix: add explicit timeouts to all reqwest HTTP clients + add lots of prover logs

tomg10:
Several HTTP clients across the codebase were created with `reqwest::Client::new()` which has no timeout configured. This can cause indefinite hangs when a remote server is slow or unresponsive, with no error logged — the task silently freezes.

This was the root cause of batch 58064 being stuck for 5 hours: the ProofGenDataSubmitter's HTTP POST to the prover gateway hung indefinitely, leaving the batch as `picked_by_prover` until the `proof_generation_timeout` (18000s) reclaim caught it.

Affected clients:
- proof_data_handler HTTP client (60s) — root cause of the incident
- Avail DA bridge API client (60s)
- Contract verifier GitHub resolver (300s — downloads large binaries)
- Metadata calculator tree API client (60s)
- ZK OS tree manager API client (60s)
- Prover autoscaler HTTP client (60s)

## Is this a breaking change?
- [ ] Yes
- [ ] No

## Checklist


core/node/proof_data_handler/src/client/http_client.rs
```diff
@@ -12,13 +14,15 @@ pub(crate) struct HttpClient {
 
 const SUBMIT_REQUEST_FOR_PROOFS_ENDPOINT: &str = "/submit_request_for_proofs";
 const POLL_GENERATED_PROOFS_ENDPOINT: &str = "/poll_generated_proofs";
+const HTTP_TIMEOUT: Duration = Duration::from_secs(60);
```
EmilLuta: Can we reuse this constant everywhere?

tomg10: I'm not a huge fan of centralizing it, those constants feel context specific (some are 1min, some are 5min) and we would need to put it in zksync_utils in a file like constants.rs with only that constant

Deniallugo (commented):
Ideally we have to have an opportunity to overwrite it using config. 

These constants could become a huge pain

Deniallugo (approved):

PR #4782, opened 2026-04-20
fix: Ensure prover jobs are assigned as intended

EmilLuta:
Description

There was an use case where prover jobs would be left hanging on core and sent to prover only after timeout duration (default 5h).

Here's how it would work:
1. eth proof manager will get a job and lock it for proving networks
2. eth proof manager starts loading all required data
3. in the meanwhile eth watcher will fail (i.e. rpc timeout)
4. the tokio select! loop will error out (which causes a panic)
5. the eth proof manager needs to exit, leaving the job locked

As such, the job is locked (looks like it will be proven very soon), but in reality it's just been timed out.

This PR adds some state machine around how locking/unlocking and general state management works. As such, even though we lock it early on, we don't consider it "fully locked" until it's been sent. If it's not been sent in the timeout time (see timeouts in queries), it will simply be fallbacked on prover cluster.


core/lib/dal/src/eth_proof_manager_dal.rs
```diff
@@ -58,6 +62,71 @@ impl ProvingNetwork {
...
+            RETURNING
+            proof_generation_details.l1_batch_number
+            "#
+        )
+        .instrument("lock_batch_for_proving_network_and_prepare#lock_proof")
+        .fetch_optional(&mut transaction)
+        .await?
+        .map(|row| L1BatchNumber(row.l1_batch_number as u32));
+
+        if let Some(batch_number) = batch {
```
EmilLuta, line 108: This is our "probe" that ensures data is not lost. We track current status and if we somehow don't have it, we can always fallback.

core/lib/dal/src/eth_proof_manager_dal.rs
```diff
@@ -319,7 +415,14 @@ impl EthProofManagerDal<'_, '_> {
                     AND submit_proof_request_tx_sent_at < NOW() - $5::INTERVAL
                 )
                 OR (status = $6 AND proof_validation_result IS false)
-                OR (status = $7 AND updated_at < NOW() - $8::INTERVAL)
+                OR (
```
EmilLuta, line 418: This logic is the fallback logic. Perhaps most important part of the PR (logic wise).

core/node/eth_proof_manager/src/sender/submit_proof_request.rs
```diff
@@ -16,6 +16,14 @@ use crate::{
     types::{ProofRequestIdentifier, ProofRequestParams},
 };
 
+#[derive(Debug, thiserror::Error)]
+enum SubmitRequestError {
```
EmilLuta, line 20: Only introduced for nicer logging and correctness. Was on the fence with this one, but yeah. Can be removed if it's too much clutter.

Artemka374 (approved):

Deniallugo (approved):

PR #4784, opened 2026-04-20
fix: Prioritize Era prover jobs

EmilLuta:
Marks Era prover jobs throughout the system as priority 1 (over 0, as rest of chains). This will prioritize proving chain Era on internal prover cluster. This is true across every step of proving.

Note: Also fixes proof compressor to use priority.


Artemka374 (approved):

Deniallugo:
I'd peronally go through the config value and we can configure it in the prover directly.  e.g. what chain should we prioritize now.

EmilLuta:
> I'd peronally go through the config value and we can configure it in the prover directly. e.g. what chain should we prioritize now.

That doesn't make a lot of sense to me. Why prioritize a single chain? The only chain we need for CEXs is Era. I can only imagine us needing to set different priorities for different chains (i.e. chain A at priority 1, chain B & C at 2, chain D at 3).

But this is all speculation and premature optimization IMO. This specific change is to keep Era up & running at max speed.

PR #2174, opened 2024-06-06
chore: Fix links 

jjjike2021:
## What ❔
1.fix links since zksync_core has been moved
2.fix code line eg.`process_l1_batch`

## Checklist

- [ ] PR title corresponds to the body of PR (we generate changelog entries from PRs).
- [ ] Tests for the changes have been added / updated.
- [ ] Documentation comments have been added / updated.
- [ ] Code has been formatted via `zk fmt` and `zk lint`.
- [ ] Spellcheck has been run via `zk spellcheck`.

Issue #2171, opened 2024-06-06
Zksync era must be list to TRUSTWALLET and META MASK an sync the STAKING PLATFORM


iandareen:
When crypto will list in biggest and many users uses the wallet example meta mask and TRUSTWALLET it will avoid dump..better to add zksync era in staking platform too.

PR #4794, opened 2026-04-24
fix(upgrades): support Era v29→v31 accepted upgrade path

valera-grinenko-ai:
## Summary

Keeps the public zksync-era pieces needed for an honest Era v29→v31 upgrade path, after moving the private `era-cacher` harness out of this repository.

### Primary change: `eth_watch` now listens to `ServerNotifier::UpgradeTimestampUpdated`

In v31 the `ServerNotifier::setUpgradeTimestamp` function validates that cut data already exists for the chain's current protocol version before storing the timestamp:

```solidity
function setUpgradeTimestamp(uint256 _chainId, uint256 _upgradeTimestamp) external onlyChainAdmin(_chainId) {
    if (_upgradeTimestamp == 0) revert ZeroUpgradeTimestamp();
    uint256 _oldProtocolVersion = chainTypeManager.getProtocolVersion(_chainId);
    if (chainTypeManager.upgradeCutHash(_oldProtocolVersion) == bytes32(0))
        revert CutDataForProtocolVersionNotAvailable(_oldProtocolVersion);
    protocolVersionToUpgradeTimestamp[_chainId][_oldProtocolVersion] = _upgradeTimestamp;
    emit UpgradeTimestampUpdated(_chainId, _oldProtocolVersion, _upgradeTimestamp);
}
```

`DecentralizedUpgradesEventProcessor` is updated to listen for `ServerNotifier::UpgradeTimestampUpdated` (filtered by `l2_chain_id` via `topic2`) instead of `ChainAdmin::UpdateUpgradeTimestamp`. Because the event now carries the **old** (from) protocol version rather than the new (to) version, `diamond_cut_for_version` is simplified to a single-block point query for that specific old version's cut, using the on-chain `upgradeCutDataBlock` mapping to locate the exact block.

**During the v31 transition upgrade the zkstack CLI still calls `ChainAdmin::setUpgradeTimestamp` as well**, because the old server (running pre-v31 code before the restart) listens to the ChainAdmin event. The new server (post-restart) listens to the ServerNotifier event. Both calls must be emitted once; subsequent upgrades after v31 only need the ServerNotifier call.

### v31 upgrade tx data is chain-specific

The v31 diamond cut stored on the CTM contains placeholder base-token force-deployment data. Before the node can execute the L2 upgrade tx it must rewrite that data with chain-specific values fetched from the L1 `EraSettlementLayerV31Upgrade` helper contract (`getL2UpgradeTxData`). This is done by:

- Adding `rewrite_upgrade_tx_data` to `ProtocolUpgradePreimageOracle` (default no-op; overridden for `&dyn EthClient` where it calls `getL2UpgradeTxData` for `Version31`).
- Calling the rewrite inside `ProtocolUpgrade::try_from_init_calldata`, before `prepare_upgrade_call`, so stored upgrade txs always carry chain-specific calldata.

### Other changes retained here

- **`diamond_cuts_since_version` removed** — replaced by `diamond_cut_for_version` which returns a single `Option<Vec<u8>>` for the requested old version. Returns a hard error if more than one `NewUpgradeCutData` event is found in the same block (contract invariant: `upgradeCutDataBlock` always points to a single block per version).
- **`zkstack` v31 upgrade support**: `ServerNotifier::setUpgradeTimestamp` calldata generation; pre-v31 vs v31 dispatch for `upgradeChainFromVersion` (v31 adds `hyperchain_addr` as first argument); `finishMigrateChainToGatewayWithCutData` replaces the old `finishMigrateChainToGateway`; `governance_execute_calls` returns early after a broadcast (script never writes the output file in that path); `noGovernancePrepare` calldata built inline for V31; new `CoreUpgradeV31Abi` and `IServerNotifierAbi` ABI imports.
- **`upgrade.test.ts`** updated to call `ServerNotifier::setUpgradeTimestamp` (fetching the `serverNotifierAddress` from the CTM) instead of `ChainAdmin::setUpgradeTimestamp`.
- **New DAL migration** adds a partial index `l1_batches_number_idx` on `l1_batches(number) WHERE is_sealed = false` to speed up unsealed-batch lookups.
- **`contracts` submodule** pinned to matter-labs/era-contracts#2174 head, which includes the v31 Era contract fixes from #2171, protocol-ops command support, the v31 ecosystem-upgrade ABI, and a refreshed `AllContractsHashes.json` + genesis configs.
- **Stale `infrastructure/local-upgrade-testing/era-cacher` scripts and README** removed; the harness now lives in the private upgrade-tests repo.

## Dependencies

Both must merge before this PR can land cleanly:

- **matter-labs/era-contracts#2174** — protocol-ops command additions and review-response cleanup.
- **matter-labs/era-contracts#2171** — era-contracts-side fixes for the v31 Era upgrade tx. #2174 is stacked on #2171.

The `contracts` submodule pin points to the current #2174 head. Once #2171/#2174 merge into `draft-v31`, repin to the merged contracts SHA through the normal branch flow.

## Validation

Previously validated as a paired flow against local reth + zkstack Postgres on a fresh v29 ecosystem. All five protocol-ops phases (`upgrade-prepare`, `upgrade-governance`, `set-upgrade-timestamp`, `chain upgrade`, `set-da-validator-pair`) completed cleanly; the post-upgrade L2 server applied the v31 upgrade tx and produced v31 batches. Non-prividium integration tests after upgrade: **16 passed, 1 skipped**; **182 tests passed, 4 skipped (186 total)**. L1 protocol version verified at `0.31.0`.

The command harness now lives in the private upgrade-tests repo, not in this public repo.

## Test plan

- [x] e2e: full v29 → v31 upgrade flow on local reth, exit 0, 16/16 non-prividium integration suites pass
- [x] `eth_watch` unit tests updated and passing: wrong-chain-ID events ignored, later-block diamond cut preferred, upgrade timestamp retry on missing cut
- [x] `cargo fmt --check` in `zkstack_cli`
- [x] `cargo check -p zkstack`
- [x] Contracts submodule repinned to current era-contracts#2174 head
- [x] `cargo clippy -p zksync_contracts -p zksync_eth_watch -p zksync_types --all-targets` clean against the new pin
- [x] `cargo fmt --check` clean on `core/` workspace


zkstack_cli/crates/zkstack/src/commands/dev/commands/upgrades/default_ecosystem_upgrade.rs
```diff
@@ -184,6 +199,94 @@ async fn no_governance_prepare(
     Ok(())
 }
 
+fn build_v31_no_governance_prepare_calldata(
```
valera-grinenko-ai, line 194: It is a bit less v29/prod-like because Rust manually mirrors `EcosystemUpgradeParams` and hardcodes paths. If we want to harden it, maybe adding a v31 `run()` entrypoint like v29, generated ABI bindings, or moving these paths into `ForgeScriptParams` / upgrade config, wdyt?

infrastructure/local-upgrade-testing/era-cacher/do-upgrade.sh
```diff
@@ -9,6 +9,12 @@ upgrade_file_extension="v31"
 # changes the beavior of zkstack ecosystem init.
 [ -d "./chains/gateway" ] && rm -rf "./chains/gateway"
 
+# The old v29 checkout's local reth genesis lacks the deterministic CREATE2 factory
+# predeploy expected by the v31 deployment scripts. Use the current local L1 genesis
+# instead of deploying the factory manually from this harness.
+rm -rf etc/reth/chaindata
```
valera-grinenko-ai: Non-blocking: this is better than manually deploying CREATE2 from the harness, but it still couples the test to the local reth chaindata layout. Longer term, a versioned local-L1 genesis fixture/input would make this less implicit.

infrastructure/local-upgrade-testing/era-cacher/do-upgrade.sh
```diff
@@ -56,17 +55,31 @@ zkstack dev run-ecosystem-upgrade --upgrade-version $upgrade_version --ecosystem
 zkstack dev run-ecosystem-upgrade --upgrade-version $upgrade_version --ecosystem-upgrade-stage governance-stage1
 
 cd contracts/l1-contracts
-UPGRADE_ECOSYSTEM_OUTPUT=script-out/v31-upgrade-ecosystem.toml UPGRADE_ECOSYSTEM_OUTPUT_TRANSACTIONS=broadcast/EcosystemUpgrade_v31.s.sol/9/run-latest.json  YAML_OUTPUT_FILE=script-out/v31-local-output.yaml yarn upgrade-yaml-output-generator
+NO_GOVERNANCE_PREPARE_SELECTOR=$(cast sig \
```
valera-grinenko-ai: Non-blocking: this duplicates the v31 ABI signature in shell only to find Foundry's selector-named broadcast file. A small follow-up could make the YAML generator accept a broadcast directory and fail unless exactly one *-latest.json exists.

infrastructure/local-upgrade-testing/era-cacher/do-upgrade.sh
```diff
@@ -56,17 +55,31 @@ zkstack dev run-ecosystem-upgrade --upgrade-version $upgrade_version --ecosystem
...
 
+# Governance stage 1 updates the canonical contracts config. Refresh the local
+# permanent values and generated runtime config before the chain upgrade.
+../era-cacher/update-permanent-values.sh
+zkstack dev contracts
+
 zkstack dev run-chain-upgrade --upgrade-version $upgrade_version --force-display-finalization-params=true --dangerous-local-default-overrides=true --chain era
 zkstack dev run-ecosystem-upgrade --upgrade-version $upgrade_version --ecosystem-upgrade-stage governance-stage2
 
+L1_DA_VALIDATOR=$(awk '/^  rollup_l1_da_validator_addr:/ { print $2; exit }' chains/era/configs/contracts.yaml)
```
valera-grinenko-ai: Non-blocking: this re-applies required post-v31 DA validator state via the normal zkstack command, which is fine for era-cacher. It would be cleaner if the upgrade output/config made this required post-step explicit.

infrastructure/local-upgrade-testing/era-cacher/update-permanent-values.sh
```diff
@@ -92,9 +94,18 @@ EOF
 echo "$CONTENT" > "$OUTPUT_FILE1"
 echo "$CONTENT" > "$OUTPUT_FILE2"
 
+# The runtime reads the generated per-chain contracts config directly.
+# Keep the key addresses in sync with the canonical config used above.
+if [ -f "$CHAIN_CONTRACTS_FILE" ]; then
+  sed -i -E "s#(^  bridgehub_proxy_addr: ).*#\\1$BRIDGEHUB_ADDR#" "$CHAIN_CONTRACTS_FILE"
```
valera-grinenko-ai: Non-blocking: the canonical source is correct, but syncing generated YAML with awk/sed is fragile if config shape changes. A zkstack config refresh/sync command would be a cleaner long-term home for this harness step.

infrastructure/local-upgrade-testing/era-cacher/do-upgrade.sh
```diff
@@ -56,17 +55,31 @@ zkstack dev run-ecosystem-upgrade --upgrade-version $upgrade_version --ecosystem
...
+zkstack chain set-da-validator-pair "$L1_DA_VALIDATOR" BlobsAndPubdataKeccak256 --chain era
+
 # Stage 3: Migrate token balances from NTV to AssetTracker
 # This can be done with any private key (deployer is used here)
+BRIDGEHUB_PROXY=$(awk '/^  bridgehub_proxy_addr:/ { print $2; exit }' chains/era/configs/contracts.yaml)
 cd contracts/l1-contracts
+[ -f script-config/v31-bridged-tokens.toml ] || printf "[tokens]\nbridged_tokens = []\n" > script-config/v31-bridged-tokens.toml
```
valera-grinenko-ai: Non-blocking: creating an empty bridged-token config is OK for this local era-cacher path, but it is another harness default. If stage3 remains part of the standard flow, the empty/default token list should probably come from script config generation instead of shell.

core/node/eth_watch/src/client.rs
```diff
@@ -582,6 +593,34 @@ where
...
+        &self,
+        init_address: Address,
+        existing_tx_data: Vec<u8>,
+    ) -> Result<Vec<u8>, ContractCallError> {
+        let bridgehub: Address = CallFunctionArgs::new("getBridgehub", ())
+            .for_contract(self.diamond_proxy_addr, &self.getters_facet_contract_abi)
+            .call(&self.client)
+            .await?;
+        let zksync_os: bool = CallFunctionArgs::new("getZKsyncOS", ())
+            .for_contract(self.diamond_proxy_addr, &self.hyperchain_contract_abi)
```
kelemeno: always false here

infrastructure/local-upgrade-testing/era-cacher/do-upgrade.sh
```diff
@@ -21,12 +27,10 @@ zkstack ecosystem init --deploy-paymaster --deploy-erc20 \
 
 zkstack server --chain era &> ../rollup3.log &
 
-zkstack dev rich-account 0x36615Cf349d7F6344891B1e7CA7C72883F5dc049  --chain era 
+zkstack dev rich-account 0x36615Cf349d7F6344891B1e7CA7C72883F5dc049  --chain era
 
 sleep 10
 
-pkill -9 zksync_server
```
kelemeno, line 28: I think we need to kill the old server so we can start the new one. different codebases

infrastructure/local-upgrade-testing/era-cacher/do-upgrade.sh
valera-grinenko-ai: Outdated by the protocol-ops migration. Current `do-upgrade.sh` doesn't read `forge broadcast/*-latest.json` files at all — the upgrade output is consumed via `script-out/v31-upgrade-ecosystem.toml` (parsed by `protocol_ops ecosystem sync-runtime-contracts`). No bash-side ABI signature duplication or selector lookup remains. Resolving.

infrastructure/local-upgrade-testing/era-cacher/do-upgrade.sh
valera-grinenko-ai: Outdated by the protocol-ops migration. Phase 8 of the current `do-upgrade.sh` calls `protocol_ops chain set-da-validator-pair --execute` directly — not `zkstack chain set-da-validator-pair`. The required-ness is explicit at the call site (Phase 8 banner + the comment in the script header listing `set-da-validator-pair` as a protocol-ops-driven step). Resolving.

core/node/eth_watch/src/client.rs
valera-grinenko-ai: Correct that this is always `false` in Era's `eth_watch` instance — Era diamond proxies return `false` from `getZKsyncOS()`, and an eth_watch is bound to a single chain.

The reason the call exists is that `settlementLayerV31Upgrade.getL2UpgradeTxData(bridgehub, chainId, zksync_os, existing_tx_data)` takes the flag as a parameter (the contract is shared between Era and ZKsyncOS), and querying the chain is the chain-agnostic way to populate it — keeps the Rust code from having a hardcoded VM assumption that drifts from the ABI it's encoding.

Two reasonable cleanups, though, depending on intent:

1. **Cache it once at construction.** `eth_watch_client.zksync_os` is a per-chain constant; it doesn't need to be re-fetched on every upgrade event rebuild. One eth_call per server lifetime instead of one per rebuild.
2. **Hardcode `false` here.** Honest about Era's invariant; loses a bit of generality if eth_watch ever needs to run against a ZKsyncOS chain in the same binary, but trivial to revisit.

I'd lean toward (1) — it's the smallest defensible change and keeps the ABI contract right. Let me know which you'd prefer and I'll do the change in this PR.

infrastructure/local-upgrade-testing/era-cacher/do-upgrade.sh
valera-grinenko-ai: Addressed in the current `do-upgrade.sh` (post-rebase).

The v29 → v31 server swap now does the kill explicitly right before starting the v31 binary:

```bash
# Phase 10: restart server with v31 tree, run integration tests
pkill -9 zksync_server || true
zkstack server --ignore-prerequisites --chain era &> "$WORKSPACE_PARENT/rollup2.log" &
```

The v29 server is left running through phases 3–8 on purpose — chain upgrade emits an L2 upgrade event the v29 `eth_watch` can't decode, so it crashes itself and the L2 upgrade tx queues into the L1 priority queue. The v31 server (started after the `pkill`) picks the queued tx up post-swap. There's a longer-form note about this on the `Verifying protocol version on L1` phase comment if you want the full rationale.

Net effect for your concern: by the time the v31 binary starts, no v29 zksync_server process is alive (either crashed itself, or `pkill`'d here as a belt-and-suspenders).

core/node/eth_watch/src/client.rs
0xValera: I'm not sure of the best solution, I think (2) might be not very good code-practice wise. Don't like hardcoding stuff

infrastructure/local-upgrade-testing/era-cacher/do-upgrade.sh
```diff
@@ -1,96 +1,534 @@
-era-cacher/reset.sh
```
0xValera, line 1: For now, reviewers can ignore helpers on lines before 
`# ── Phase 0: clean slate, swap to v29 checkout, deploy v29 ecosystem ──────`
They are unstable, some of them will be removed, most will get polished.

core/node/eth_watch/src/client.rs
kelemeno: its literlayy in the eraVM server, its ok

infrastructure/local-upgrade-testing/era-cacher/do-upgrade.sh
```diff
@@ -1,96 +1,559 @@
...
+# files.
+#
+# zkstack still drives:
+#   - `ecosystem init` (v29 deploy + DB + wallets — protocol-ops doesn't
+#     exist at v29, so we must use zkstack to seed the v29 state)
+#   - `server`, `dev contracts`, `dev database migrate`
+#   - `dev test integration` (suite runner)
+#   - `dev rich-account`, `dev init-test-wallet` (test fixture setup)
+#
+# Direct cast / forge:
```
kelemeno: the diff in the file is huge, can you cut it down, and if new files are needed split it up into multiple small ones. I don't think this amount of changes are necessary

etc/reth/chaindata/reth_config
```diff
@@ -27,7 +27,7 @@
   },
   "nonce": "0x0",
   "timestamp": "0x5ca9158b",
-  "gasLimit": "0x1c9c380",
+  "gasLimit": "0x4c4b400",
```
kelemeno: i'm pretty sure this can't be chantged, we need to execute on real chains after all

kelemeno: if needed split it up into multiple governance calls or something similar

zkstack_cli/crates/zkstack/src/commands/dev/commands/upgrades/default_ecosystem_upgrade.rs
```diff
@@ -184,6 +199,94 @@ async fn no_governance_prepare(
...
+    ecosystem_config: &EcosystemConfig,
+    vm_option: VMOption,
+) -> anyhow::Result<Bytes> {
+    let contracts_config = ecosystem_config.get_contracts_config()?;
+    let ctm = contracts_config.ctm(vm_option);
+    let governance = contracts_config.l1.governance_addr;
+    let zk_token_asset_id = ecosystem_config.l1_network.zk_token_asset_id();
+    #[allow(deprecated)]
+    let ecosystem_upgrade_v31 = Function {
```
kelemeno: we should not declare the function here we should save its abi in zkstackout and just import it

core/lib/types/src/protocol_upgrade.rs
```diff
@@ -206,6 +218,13 @@ impl ProtocolUpgrade {
                     .await?
             };
 
+            if version.minor == ProtocolVersionId::Version31 {
```
kelemeno: looks good to me

core/node/eth_watch/src/event_processors/decentralized_upgrades.rs
```diff
@@ -152,6 +166,7 @@ impl EventProcessor for DecentralizedUpgradesEventProcessor {
                     (upgrade, scheduler_vk_hash, fflonk_scheduler_vk_hash),
                 );
             }
+            processed_events += 1;
```
kelemeno: why is this needed?

infrastructure/local-upgrade-testing/era-cacher/do-upgrade.sh
0xValera: Addressed in [10d2a5de2](https://github.com/valera-grinenko-ai/zksync-era/commit/10d2a5de2): the public `era-cacher` harness was removed from this PR. The large shell flow now lives in the private upgrade-tests repo; this PR keeps only the public runtime / zkstack support.

core/node/eth_watch/src/client.rs
0xValera: Addressed in [6d66256af](https://github.com/valera-grinenko-ai/zksync-era/commit/6d66256af): this now passes `false` directly with a comment. This `eth_watch` runs in the Era server, so the v31 settlement-layer helper should always use the Era path here.

core/node/eth_watch/src/event_processors/decentralized_upgrades.rs
0xValera: Needed because `eth_watch` advances its cursor from the returned processed-event count. If the timestamp event arrives before the diamond cut data and we still return `events.len()`, the watcher skips that event permanently. Returning only fully processed events leaves it at the cursor so it is retried. Added a comment in [e0410e3ae](https://github.com/valera-grinenko-ai/zksync-era/commit/e0410e3ae).

core/lib/types/src/protocol_upgrade.rs
0xValera: Thanks, resolving.

zkstack_cli/crates/zkstack/src/commands/dev/commands/upgrades/default_ecosystem_upgrade.rs
0xValera: Addressed in [e0410e3ae](https://github.com/valera-grinenko-ai/zksync-era/commit/e0410e3ae): `noGovernancePrepare` is now imported from `zkstack-out` through `abigen!`, and the contracts submodule is pinned to the commit that exports the v31 ecosystem-upgrade ABI.

core/lib/types/src/protocol_upgrade.rs
```diff
@@ -90,14 +90,20 @@ impl From<abi::VerifierParams> for VerifierParams {
...
     async fn get_protocol_upgrade_preimages(
         &self,
         hashes: Vec<H256>,
     ) -> anyhow::Result<Vec<Vec<u8>>>;
+
+    async fn rewrite_v31_upgrade_tx_data(
+        &self,
+        init_address: Address,
+        existing_tx_data: Vec<u8>,
+    ) -> anyhow::Result<Vec<u8>>;
```
Deniallugo: it's extremely confusing to have it as a trait

core/lib/types/src/protocol_upgrade.rs
```diff
@@ -206,6 +218,13 @@ impl ProtocolUpgrade {
                     .await?
             };
 
+            if version.minor == ProtocolVersionId::Version31 {
+                let placeholder = std::mem::take(&mut upgrade.l2_protocol_upgrade_tx.data);
+                upgrade.l2_protocol_upgrade_tx.data = preimage_oracle
+                    .rewrite_v31_upgrade_tx_data(init_address, placeholder)
+                    .await?;
+            }
```
Deniallugo: can you explain it?

core/node/eth_watch/src/event_processors/decentralized_upgrades.rs
Deniallugo: how is it possible? could we keep this invariant on the contracts level?

zkstack_cli/crates/zkstack/src/admin_functions.rs
```diff
@@ -199,6 +200,10 @@ pub async fn governance_execute_calls(
     forge.run(shell)?;
     spinner.finish();
 
+    if is_broadcast {
+        return Ok(AdminScriptOutput::default());
+    }
```
Deniallugo, line 207: why?

core/lib/types/src/protocol_upgrade.rs
Deniallugo: the implementation doesn't look v31 specific. 
if it's really necessary to overwrite something in tx calldata. then let's make it more general. 

but overall at current point of time, seems that tx_calldata in log is outlived it self and we probably need to get the whole txdata onchain.

Deniallugo: Discussed offline. 
decided to keep the logic the same. 
but i'd like at least change the name and pass protocol version and return the original value if it's not v31 and do the job if it's v31

zkzoomer: Refactored in f1cb69b8e7c06016909e777fabdc64bcbe2cf33d

zkstack_cli/crates/zkstack/src/admin_functions.rs
zkzoomer: `AdminFunction`'s `governanceExecuteCalls` in the Solidity script is a pure execution helper, it just decodes and forwards calls, unlike every other function here which calls saveAndSendAdminTx to write the output file. So in broadcast mode there's nothing to read back. Added a comment in 7b7715ee3768fdd735ed0698163d45e92888f273 to explain that.  But do note that these functions are effectively dead going forward since ecosystem upgrades are handled by protocol ops now.

core/node/eth_watch/src/client.rs
```diff
@@ -582,6 +590,32 @@ where
             base_token_symbol,
         }))
     }
+
+    async fn get_v31_l2_upgrade_tx_data(
```
Deniallugo: why it's still v31 specific?
it's trait must be generic

core/lib/types/src/protocol_upgrade.rs
```diff
@@ -206,6 +222,11 @@ impl ProtocolUpgrade {
                     .await?
             };
 
+            let tx_data = std::mem::take(&mut upgrade.l2_protocol_upgrade_tx.data);
+            upgrade.l2_protocol_upgrade_tx.data = preimage_oracle
+                .rewrite_upgrade_tx_data(version.minor, init_address, tx_data)
+                .await?;
```
Deniallugo: why do you use mem::take? are you trying to optimize? it?

Deniallugo: shouldn't it be part of `prepare_upgrade_call` looks weird that you are changing the data twice

core/node/eth_watch/src/client.rs
```diff
@@ -582,6 +590,32 @@ where
...
+        &self,
+        init_address: Address,
+        existing_tx_data: Vec<u8>,
+    ) -> Result<Vec<u8>, ContractCallError> {
+        let bridgehub: Address = CallFunctionArgs::new("getBridgehub", ())
+            .for_contract(self.diamond_proxy_addr, &self.getters_facet_contract_abi)
+            .call(&self.client)
+            .await?;
```
Deniallugo: Is it unknown at this point?
we should have it saved at the start of the pod

core/node/eth_watch/src/event_processors/decentralized_upgrades.rs
```diff
@@ -65,6 +71,22 @@ impl ProtocolUpgradePreimageOracle for &dyn EthClient {
...
+        existing_tx_data: Vec<u8>,
+    ) -> anyhow::Result<Vec<u8>> {
+        if protocol_version != ProtocolVersionId::Version31 {
+            return Ok(existing_tx_data);
+        }
+
+        // V31 fills empty per-chain base-token force-deployment data from Bridgehub/NTV.
+        self.get_v31_l2_upgrade_tx_data(init_address, existing_tx_data)
+            .await
+            .map_err(Into::into)
```
Deniallugo: should it be inlined?

zkzoomer: Refactored in 2a4c0df7b39c8be02056e3c1c4a35ae681b27f50

core/node/eth_watch/src/client.rs
zkzoomer: Refactored in 2a4c0df7b39c8be02056e3c1c4a35ae681b27f50

core/lib/types/src/protocol_upgrade.rs
zkzoomer: Refactored in 2a4c0df7b39c8be02056e3c1c4a35ae681b27f50

core/node/eth_watch/src/client.rs
zkzoomer: Refactored in 2a4c0df7b39c8be02056e3c1c4a35ae681b27f50

core/node/eth_watch/src/client.rs
```diff
@@ -582,6 +599,32 @@ where
...
     }
+
+    fn bridgehub_addr(&self) -> Option<Address> {
+        self.bridgehub_proxy_addr
+    }
+
+    async fn get_l2_upgrade_tx_data(
+        &self,
+        bridgehub_addr: Address,
+        init_address: Address,
```
Deniallugo, line 608: Why do you have different bridgehubs? 
could you the one which is already presented in eth_client?

Deniallugo (dismissed):
It's rather nit. 
LGTM

core/node/eth_watch/src/client.rs
zkzoomer: Resolved in 5850b2775e2b30e01adef99f45c211a0ac6a3238

Deniallugo (approved):

PR #4806, opened 2026-05-12
chore(airbender commit): Refactoring

Deniallugo:
## What ❔

Refactors the airbender commitment pipeline so a single `L1BatchCommitment` carries both the Boojum-shape and Airbender-shape `aux_commitments` and exposes the two derived commitments side-by-side. Replaces the v1 path where airbender commitment was computed via a separate helper with duplicated inputs.

Key changes in `zksync_types::commitment`:
- `L1BatchAuxiliaryOutput::PostBoojum` gains `airbender_aux_commitments: Option<AuxCommitments>` so the parallel `aux_data_hash` / commitment can be derived without recomputing the rest of the auxiliary output.
- New `AuxCommitmentsBoth { boojum, airbender }` struct carrying both variants.
- **Breaking:** `CommitmentInput::PostBoojum.aux_commitments` changes type from `AuxCommitments` to `AuxCommitmentsBoth`. The field name is preserved; external pattern-matches/constructors on the variant need the type update.
- New public API: `L1BatchAuxiliaryOutput::airbender_hash` / `airbender_aux_commitments`, `L1BatchCommitment::airbender_artifacts` (returns an `AirbenderBatchCommitment`).
- Serialization of the auxiliary output is shared between Boojum and Airbender via `post_boojum_aux_bytes` — the caller picks which `&AuxCommitments` to feed in, no implicit variant flag.

Downstream:
- `commitment_generator` now computes and persists both commitments per batch (Boojum in `l1_batches`, Airbender in the new `airbender_batch_commitments` table populated by `save_airbender_batch_commitment`).
- `genesis::utils` persists the airbender row for batch 0 (zero aux on both variants → values collapse to the Boojum row).
- `airbender_prover_interface` consumes the new `AirbenderBatchCommitment` shape; `airbender_request_processor` passes commitment + aux_data_hash from the parallel table.

## Why ❔

The v1 design assembled airbender commitment inputs from a separate helper that re-derived data already present in the commitment pipeline; the airbender prover needed a fragile path through duplicated fields. Consolidating both variants on `L1BatchCommitment` means:
- One source of truth for `pass_through_data` and `meta_parameters` — the airbender artifact reuses them and only swaps the auxiliary output hash.
- The airbender row in `airbender_batch_commitments` is computed alongside the existing Boojum row in `commitment_generator`, so the prover reads a stable per-batch commitment from the DB instead of rebuilding inputs.

## Is this a breaking change?
- [x] Yes
- [ ] No

Public API affected: `zksync_types::commitment::CommitmentInput::PostBoojum.aux_commitments` now holds `AuxCommitmentsBoth` (was `AuxCommitments`). Anything in a downstream crate that constructs or destructures this variant must wrap the inner value as `AuxCommitmentsBoth { boojum, airbender }`.

## Operational changes

- New migration `add_airbender_batch_commitments` adds a sibling table to `l1_batches`. Backfill is not required for new chains; existing chains will have the airbender row populated forward from the first batch processed under this version (genesis-row insertion happens on chain init).

## Checklist

- [ ] Code has been formatted via \`zkstack dev fmt\` and \`zkstack dev lint\`.

### DIFF
--- a/core/lib/types/src/commitment/mod.rs
+++ b/core/lib/types/src/commitment/mod.rs
@@ -357,6 +357,15 @@ pub struct PrevBatchAirbenderCommitmentInput {
     pub prev_aux_hash: H256,
 }
 
+/// Selects which `aux_commitments` variant of a `PostBoojum` auxiliary output
+/// to serialize/hash. Used internally to share the serialization path between
+/// the default (Boojum) hash and the Airbender hash.
+#[derive(Debug, Clone, Copy)]
+enum AuxCommitmentsVariant {
+    Boojum,
+    Airbender,
+}
+
 /// Block Output produced by Virtual Machine
 #[derive(Debug, Clone, Eq, PartialEq)]
 #[cfg_attr(test, derive(Serialize, Deserialize))]
@@ -375,6 +384,11 @@ pub enum L1BatchAuxiliaryOutput {
         state_diffs_compressed: Vec<u8>,
         state_diffs_hash: H256,
         aux_commitments: AuxCommitments,
+        /// Airbender-shape `aux_commitments`, kept here to derive a parallel
+        /// `aux_data_hash`/commitment without recomputing the rest of the
+        /// auxiliary output. `None` when only the Boojum variant is available.
+        #[cfg_attr(test, serde(default))]
+        airbender_aux_commitments: Option<AuxCommitments>,
         blob_hashes: Vec<BlobHash>,
         aggregation_root: H256,
         local_root: H256,
@@ -428,10 +442,14 @@ impl L1BatchAuxiliaryOutput {
                 common: common_input,
                 system_logs,
                 state_diffs,
-                aux_commitments,
+                aux_commitments_both,
                 blob_hashes,
                 aggregation_root,
             } => {
+                let AuxCommitmentsBoth {
+                    boojum: aux_commitments,
+                    airbender: airbender_aux_commitments,
+                } = aux_commitments_both;
                 let l2_l1_logs_compressed = serialize_commitments(&common_input.l2_to_l1_logs);
                 let merkle_tree_leaves = l2_l1_logs_compressed
                     .chunks(UserL2ToL1Log::SERIALIZED_SIZE)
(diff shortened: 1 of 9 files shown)


core/lib/types/src/commitment/mod.rs
```diff
@@ -535,6 +565,14 @@ impl L1BatchAuxiliaryOutput {
     }
 
     pub fn to_bytes(&self) -> Vec<u8> {
```
0xVolosnikov, line 567: Should we add an annotation that this is only for Boojum?

core/lib/types/src/commitment/mod.rs
```diff
@@ -580,12 +625,35 @@ impl L1BatchAuxiliaryOutput {
         H256::from_slice(&keccak256(&self.to_bytes()))
     }
 
+    /// Hash of the auxiliary output computed with the stored Airbender-shape
+    /// `aux_commitments`. Returns `None` for pre-Boojum batches or when the
+    /// Airbender variant was not populated.
+    pub fn airbender_hash(&self) -> Option<H256> {
```
0xVolosnikov, line 631: NIt: perhaps it would be more consistent to make `hash_for` in the same way as you did with  `bytes_for`. Feel free to ignore

core/lib/types/src/commitment/mod.rs
Deniallugo: i don't want to break public interface...

core/lib/types/src/commitment/mod.rs
Deniallugo: Same as before.. i don't want to break public interfaces. 
so want to keep hash function with the same name

0xVolosnikov (approved):

core/lib/types/src/commitment/mod.rs
0xVolosnikov: Yes, but we can add a comment to explain this detail

core/lib/types/src/commitment/mod.rs
0xVolosnikov: Yes, but we can add `hash_for` which will be used by `hash`

PR #4820, opened 2026-05-22
test: address four flaky failure modes in check-flaky-tests

deniallugo-claude:
## Summary

Addresses three of the failure clusters I traced in the last 30 `check-flaky-tests` runs on `main`. One commit per cluster; each is independently reviewable and explained in the commit message.

Cluster #1 (BigInt-serialization) is in #4819.

Two earlier commits in this branch — `DepositsAlreadyPaused` tolerance and `WithdrawalAlreadyFinalized` idempotency — were reverted after review (see f4b774098). Per @zkzoomer those workarounds were masking real bugs rather than fixing them; both clusters remain open for separate investigation by someone with the right contract-level context.

| Commit | Cluster | Failing case in CI |
|---|---|---|
| `test(gateway-migration): bump timeout …` (tolerance reverted in f4b77409, **timeout bump retained**) | #2 partial | runs [25761254141](https://github.com/matter-labs/zksync-era/actions/runs/25761254141), [24884805964](https://github.com/matter-labs/zksync-era/actions/runs/24884805964) |
| `test(paymaster): pin gasPrice in "Should pay fees with testnet paymaster"` | #4 | runs [24842518808](https://github.com/matter-labs/zksync-era/actions/runs/24842518808), [24672755909](https://github.com/matter-labs/zksync-era/actions/runs/24672755909) |
| `test(node-spawner): absorb expected RPC rejections during restart` | #5 | run [25743495583](https://github.com/matter-labs/zksync-era/actions/runs/25743495583) |

## Cluster #2 — gateway-migration timeout

`Migrate to/from gateway` runs a 60-retry × ~10s loop (plus node restart + migration command). The suite-level mocha timeout was `240000` ms, only enough for 2–3 retries before mocha kills the test. Bumped to 15 min so the retry budget can be used.

The downstream `DepositsAlreadyPaused()` in `Migrating back to gateway fails` is most likely a *consequence* of the first step timing out before migration finished — once the timeout is generous, the migration completes and unpauses deposits as expected. **No tolerance/try-catch is added.**

## Cluster #4 — paymaster gasPrice drift

The testnet paymaster computes `requiredETH = gasLimit * tx.maxFeePerGas` and rejects with `magic=0` if `paymasterInput.amount < requiredETH`. The test captured `gasPrice` up front, used it to build `paymasterInput.amount = gasPrice * gasLimit`, but did not pin `gasPrice` on the outgoing transaction. ethers re-sampled at submission time, and if L1 fees ticked up between the two reads the tx arrived with `maxFeePerGas > captured gasPrice` and tripped one of two reported flakes:

- `Paymaster validation returned invalid magic value` (amount < requiredETH branch)
- `Transaction did not have the ERC20 fee log (or the amount was incorrect)` — `paidFeeWithPaymaster` asserts ERC20 transfer amount == ETH fee log amount, both of which depend on the actual fee.

Setting `gasPrice` explicitly removes the drift. Matches the pattern already in use by `sendTxWithTestPaymasterParams` for the custom-paymaster tests in the same file.

## Cluster #5 — restart-time socket hang up

The fees suite restarts the main node multiple times. Each restart leaves a few in-flight HTTP requests in zksync-ethers internals (background pollers, subscriptions on already-resolved `tx.wait()` promises) without a user-code catch handler. When the node process dies they reject with `socket hang up` / `ECONNRESET`, jest collects them as unhandled rejections, and the suite is reported as failed even when every individual test passed (5×`Test suite failed to run: socket hang up`, with `Tests: 3 passed`).

`killAndSpawnMainNode` now installs a scoped `unhandledRejection` handler around the restart window that absorbs *only* the network-disconnect families and forwards everything else to the previously-registered listeners. Unrelated unhandled rejections still surface normally.

Also silences the routinely-noisy `kill: No such process` log in `kill.ts` when the parent has already exited via SIGCHLD cascade.

## Test plan

- [ ] CI: re-run `check-flaky-tests` multiple times and confirm clusters #4 and #5 disappear from the failure list, and that #2's timeout-driven failures stop.
- [ ] Spot-check that the 15-min gateway-migration timeout doesn't push the workflow past its job-level timeout.


core/tests/gateway-migration-test/tests/migration.test.ts
```diff
@@ -376,11 +376,22 @@ describe('Migration From/To gateway test', function () {
...
         if (direction == 'TO') return;
-        // Pause deposits before trying migration back to gateway
-        await zkstackExecWithMutex(
-            `zkstack chain pause-deposits --chain ${fileConfig.chain}`,
-            'pausing deposits before migrating back to gateway'
-        );
+        // Pause deposits before trying migration back to gateway.
+        // Deposits may already be paused from the earlier "Pause deposits before initiating
+        // migration" step — the migrate-from-gateway flow does not always re-enable them — so
+        // a `DepositsAlreadyPaused()` revert here is a no-op precondition and not a failure.
```
zkzoomer: The direction for cluster #2 is right, but here this is not expected behavior. After migrating back from GW, deposits get unpaused, so we need to pause them again to test that repeated migrations fail. From what you mention, deposits might still be paused from an unfinished migration from gateway, but they should get unpaused once that migration is finished. We do not expect this pause deposits to ever fail, so this change here should be reverted.

core/tests/highlevel-test-tools/tests/token-balance-migration-tester.ts
```diff
@@ -878,7 +878,15 @@ export class WithdrawalHandler {
 
         await waitForL2ToL1LogProof(l2Wallet, receipt.blockNumber, this.txHash);
 
-        await (await l2Wallet.finalizeWithdrawal(this.txHash)).wait();
+        // `finalize-token-balance-migration --to-gateway false` may finalize pending
+        // withdrawals as a side effect, so by the time we get here the L1 finalize
```
zkzoomer: `finalize-token-balance-migration --to-gateway false` should not handle any withdrawals, what could cause that as a side effect? This would be unexpected behavior. I think the direction for this fix is wrong.

core/tests/gateway-migration-test/tests/migration.test.ts
deniallugo-claude: You're right — the migrate-from-gateway flow is supposed to unpause deposits, so this pause call should never fail in a healthy run. Reverted in f4b774098. My read now is that the DepositsAlreadyPaused symptom was a downstream consequence of the previous step timing out before migration finished — the 15-min timeout bump on its own should fix it. I'll leave the deeper question (why the unpause didn't propagate in that specific run) for someone with the right domain context.

core/tests/highlevel-test-tools/tests/token-balance-migration-tester.ts
deniallugo-claude: Agreed — if that command finalizes withdrawals as a side effect, that's the bug to fix, not paper over. Reverted in f4b774098 so the failure surfaces unmasked. The flake reproducer is run 25729024872 on 'Can finalize pending withdrawals after migrating token balances from gateway' (selector 0xae899454 = WithdrawalAlreadyFinalized via 0x665fD79D…01D9 / L1Nullifier), in case it's useful for the underlying investigation.

deniallugo-claude:
Update from the deeper investigation @zkzoomer asked for:

## Cluster #2 — DepositsAlreadyPaused: root cause located

After `migrate-from-gateway` completes successfully, `Migrator.forwardedBridgeMint` does set `s.settlementLayer = 0` AND calls `_unpauseDeposits()` (Migrator.sol:298 on `draft-v31`). So deposits are auto-unpaused — exactly as you said. The CLI does not need an explicit unpause call.

In the failing run [25761254141](https://github.com/matter-labs/zksync-era/actions/runs/25761254141), what actually happened: the `Migrate to/from gateway` step hit mocha's 240 s timeout before migration finished, so `forwardedBridgeMint` never ran for that chain → deposits stayed paused from step 1 → step 9's `pause-deposits` reverted with `DepositsAlreadyPaused()`.

The reason this propagated all the way to step 9 instead of failing at step "Check the settlement layer" was that those checks were written `expect(slAddressl1 != ZeroAddress)` / `expect(slAddressGW == ZeroAddress)` — chai `expect()` without a chained matcher is a no-op, so the step silently passed even though the chain was still settling on Gateway.

**Action**: I kept the timeout bump (the actual fix) and added a small commit (12f537f5d) converting those four no-op assertions into real `.to.equal(...) / .to.not.equal(...)` calls with messages. Now if migration ever fails to complete again, this step will fail loudly and immediately, pointing at the right place.

## Cluster #3 — WithdrawalAlreadyFinalized: investigation inconclusive

Tracing in the v31 contracts: `isWithdrawalFinalized[chainId][batch][msgIdx]` only gets set in `L1Nullifier._finalizeDeposit`, BEFORE `_verifyWithdrawal` and the asset-router call. Anything that reverts downstream rolls the flag back with the rest of the tx. So step 4's `InsufficientChainBalance` revert does not leave the flag set.

I read through `finalize-token-balance-migration --to-gateway false` end-to-end (`migrate_token_balances.rs::finalize_token_balance_migration` → `L1AssetTracker.receiveGatewayToL1MigrationOnL1` → `_proveMessageInclusion` + `_migrateFunds`). None of those touch `L1Nullifier.isWithdrawalFinalized`, consistent with your read that this command shouldn't be finalizing user withdrawals.

The failing trace targets L1Nullifier directly (`0x665f…01D9` is L1Nullifier in that env, selector `0x74beea82 = finalizeDeposit(FinalizeL1DepositParams)`, message body starts with `finalizeEthWithdrawal` and `l2Sender = 0x800a` = `L2_BASE_TOKEN_SYSTEM_CONTRACT_ADDR`, so the failing message is the *base-token* withdrawal). On chainId `0xb97fd` / batch 192 / msgIdx 1, that tuple was already marked true somewhere — but I couldn't find which call set it from logs + code alone.

I'd appreciate a pointer if you have intuition about what else can write `isWithdrawalFinalized` on the user chain side during the migration flow. Otherwise this would need a reproduction with extra contract-side instrumentation to identify the actual setter.

deniallugo-ml (approved):

deniallugo-ml (approved):

Issue #4822, opened 2026-05-25
Gateway Mainnet RPC endpoint from docs does not resolve


ily33:
Hi team,

The Gateway Mainnet RPC endpoint listed in the official docs does not currently resolve:
```
https://rpc.era-gateway-mainnet.zksync.dev/
```
A full DNS lookup returns NOERROR, but with no answer records. As a result, ZKsync external nodes that require Gateway Mainnet fail during startup when initializing the settlement layer.

Observed error:
```
Wiring layer settlement_layer_en can't be initialized:
JSON-RPC request chain_id() failed: client error (Connect)
dns error
failed to lookup address information: No address associated with hostname
```
Could you please confirm the correct public Gateway Mainnet RPC endpoint for chain ID 9075, or confirm whether the documented endpoint is currently unavailable?

Thanks.

PR #4829, opened 2026-05-30
fix(en): remove built-in mainnet Gateway fallback

peter941221:
## What ❔

1. Remove the built-in Era mainnet Gateway fallback from `zksync_settlement_layer_data`.
2. Keep the default Gateway URL mapping for testnet and stage.
3. Replace the generic Gateway-client `expect(...)` paths with explicit operator-facing errors.
4. Keep the new Gateway error strings centralized in `core/lib/settlement_layer_data/src/lib.rs` so the operator-facing contract stays consistent across settlement and migration paths.

## Why ❔

Issue: https://github.com/matter-labs/zksync-era/issues/4822

The old mainnet fallback still mapped the Era mainnet bridgehub to `https://rpc.era-gateway-mainnet.zksync.dev/`.

That endpoint is no longer reachable. The fallback did not give operators a working default. It selected a dead URL and deferred the failure until the node actually needed Gateway connectivity.

Live checks on 2026-05-30 narrowed the failure mode:

- `en_getInteropFee` against `https://zksync2-mainnet.zksync.io` returned HTTP 403 with JSON-RPC `-32601` and `rpc method is not whitelisted`.
- The same public endpoint still returned HTTP 200 for `zks_getBridgeContracts`, `zks_getBaseTokenL1Address`, and `zks_getTimestampAsserter`, and returned HTTP 200 with `result: null` for `zks_getL2Multicall3`.
- A live fetch to `https://rpc.era-gateway-mainnet.zksync.dev/` failed on 2026-05-30, matching the report in #4822.

This patch makes that failure mode explicit. Era mainnet now requires an intentional Gateway RPC configuration, while testnet and stage keep their existing defaults.

## Is this a breaking change?
- [x] Yes
- [ ] No

## Operational changes

1. If you run EN against Era mainnet and rely on Gateway settlement or Gateway migration flows, you now need to set `gateway_rpc_url` or `EN_GATEWAY_URL` explicitly.
2. Testnet and stage behavior stays the same.
3. The new error messages point operators at the missing Gateway RPC configuration instead of panicking with `expect(...)`.

## Checklist

Validation run in WSL:

- `cargo fmt -p zksync_settlement_layer_data`
- `cargo check -p zksync_settlement_layer_data --lib`
- `cargo test -p zksync_settlement_layer_data --lib`
- `zkstack --ignore-prerequisites dev fmt rustfmt --check`
- `zkstack --ignore-prerequisites dev lint -t rs --check`

### DIFF
--- a/core/lib/settlement_layer_data/src/gateway_urls.rs
+++ b/core/lib/settlement_layer_data/src/gateway_urls.rs
@@ -16,11 +16,11 @@ pub static TESTNET_BRIDGEHUB_ADDR: Lazy<Address> =
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum DefaultGatewayUrl {
     Testnet,
-    Mainnet,
     Stage,
 }
 
-/// While MatterLabs is the only gateway provider, it's safe to use default parameters for the gateway client.
+/// While MatterLabs is the only gateway provider, testnet and stage still have stable default
+/// parameters for the gateway client.
 /// These URLs serve as fallbacks in case the gateway URL is not specified in the secrets.
 /// This is not defined in JSON, as it's easier to configure the value in secrets
 /// rather than modifying a JSON file.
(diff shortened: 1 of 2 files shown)


deniallugo-ml:
We don't plan to introduce new gateways. Please reopen the PR with removing the default gateway functionality

PR #4831, opened 2026-05-30
fix(en): allow missing bridgehub in ecosystem RPC

peter941221:
## What ❔

This PR makes `EcosystemContractsDto.bridgehub_proxy_addr` optional, removes the EN ecosystem-RPC `unwrap()`, and lets remote EN config ingestion consume the optional field directly.

- `en_getEcosystemContracts` now returns `null` for `bridgehub_proxy_addr` when bridgehub is absent.
- The rest of the ecosystem-contracts DTO stays unchanged.
- `remote_en_config` keeps preserving the present-address path and now also accepts the missing-address path.
- Added focused RPC / serde / remote-config regression coverage for both `None` and `Some(...)` cases.

## Why ❔

The EN stack already modeled bridgehub as optional, but one RPC path still treated it as guaranteed.

`RemoteENConfig` stores `l1_bridgehub_proxy_addr` as `Option<Address>`, and `zks_getBridgehubContract` already exposed bridgehub as optional. `en_getEcosystemContracts` still called `unwrap()` on the same underlying field, so the API server could admit "bridgehub may be absent" on one path and panic on another path backed by the same config.

## Is this a breaking change?
- [x] Yes
- [ ] No

Downstream consumers must now accept `bridgehub_proxy_addr = null` in the ecosystem-contracts RPC response, and Rust consumers of `EcosystemContractsDto` must handle `Option<Address>` instead of `Address`.

## Operational changes

No node-operator action is required.

Client-side consumers of the ecosystem-contracts RPC and downstream crates using `EcosystemContractsDto` must accept a missing bridgehub value.

## Checklist

Validation run locally:
- `cargo fmt -p zksync_settlement_layer_data -p zksync_web3_decl -p zksync_node_api_server`
- `cargo check -p zksync_settlement_layer_data --lib --message-format short`
- `cargo check -p zksync_web3_decl --lib --message-format short`
- `cargo check -p zksync_node_api_server --lib --message-format short`
- `cargo test -p zksync_settlement_layer_data remote_en_config::tests::fetch_remote_en_config_allows_missing_bridgehub_addr --lib -- --exact --nocapture`
- `cargo test -p zksync_settlement_layer_data remote_en_config::tests::fetch_remote_en_config_preserves_present_bridgehub_addr --lib -- --exact --nocapture`
- `cargo test -p zksync_web3_decl types::tests::ecosystem_contracts_serde_keeps_optional_bridgehub --lib -- --exact --nocapture`

### DIFF
--- a/core/bin/external_node/src/tests/utils.rs
+++ b/core/bin/external_node/src/tests/utils.rs
@@ -309,7 +309,7 @@ pub(super) fn mock_l2_client(env: &TestEnvironment) -> MockClient<L2> {
         .method("en_whitelistedTokensForAA", || Ok([] as [Address; 0]))
         .method("en_getEcosystemContracts", move || {
             Ok(EcosystemContractsDto {
-                bridgehub_proxy_addr: contracts.ecosystem_contracts.bridgehub_proxy_addr.unwrap(),
+                bridgehub_proxy_addr: contracts.ecosystem_contracts.bridgehub_proxy_addr,
                 state_transition_proxy_addr: contracts
                     .ecosystem_contracts
                     .state_transition_proxy_addr,
--- a/core/lib/settlement_layer_data/src/remote_en_config.rs
+++ b/core/lib/settlement_layer_data/src/remote_en_config.rs
@@ -135,15 +135,14 @@ where
 #[cfg(test)]
 mod tests {
     use zksync_basic_types::{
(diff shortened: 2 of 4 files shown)


deniallugo-ml:
Bridgehub proxy address is required, even though it could be treated in some places option for convenience

PR #4834, opened 2026-05-30
fix(api_server): guard missing post-gateway aggregation roots

peter941221:
## What ❔

This PR replaces the post-gateway `aggregation_root` `expect()` in `zks_getL2ToL1LogProof` with contextual error propagation.

- Pre-gateway proofs stay unchanged.
- Post-gateway proofs with a present `aggregation_root` stay unchanged.
- Only the missing-field path changes, from panic to structured internal failure.
- Added a regression test that clears `aggregation_root` in a sealed batch and asserts the RPC call now fails as an internal error.

## Why ❔

`aggregation_root` is optional in storage, so the API path can legitimately observe batch metadata that exists but still lacks that post-gateway-only field.

The post-gateway proof branch still used `expect()` for `aggregation_root`. That made a DB-representable state crash the RPC path instead of returning a structured internal error. The regression test also needed one fixture correction: the transaction must be marked as executed in the L1 batch before the proof lookup, otherwise the test exits earlier with `Ok(None)` and never reaches the guarded branch.

## Is this a breaking change?
- [ ] Yes
- [x] No

## Operational changes

No operational changes.

## Checklist

Validation run locally:
- `cargo fmt --manifest-path core/Cargo.toml --all`
- `cargo check -p zksync_node_api_server --lib --message-format short`
- `cargo test -p zksync_node_api_server web3::tests::getting_l2_to_l1_log_proof_without_aggregation_root -- --exact --nocapture`

### DIFF
--- a/core/Cargo.lock
+++ b/core/Cargo.lock
@@ -14632,6 +14632,7 @@ dependencies = [
  "reqwest 0.12.22",
  "serde",
  "serde_json",
+ "sqlx",
  "strum 0.26.3",
  "tempfile",
  "test-casing",
--- a/core/node/api_server/Cargo.toml
+++ b/core/node/api_server/Cargo.toml
@@ -64,6 +64,7 @@ zk_evm_1_5_0.workspace = true
 zksync_node_genesis.workspace = true
 zksync_node_test_utils.workspace = true
 zksync_test_contracts.workspace = true
+sqlx.workspace = true
 
 assert_matches.workspace = true
 http-body-util.workspace = true
--- a/core/node/api_server/src/web3/namespaces/zks.rs
+++ b/core/node/api_server/src/web3/namespaces/zks.rs
@@ -1,3 +1,4 @@
+use anyhow::Context as _;
 use zksync_crypto_primitives::hasher::{keccak::KeccakHasher, Hasher};
 use zksync_dal::{Connection, Core, CoreDal, DalError};
 use zksync_mini_merkle_tree::MiniMerkleTree;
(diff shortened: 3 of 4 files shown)


peter941221:
1. Why I treated this as a real crash surface rather than an impossible state
`aggregation_root` is optional all the way through the storage and metadata model:
core/lib/types/src/commitment/mod.rs:141-148
core/lib/dal/src/models/storage_block.rs:324-326

The API path reads batch metadata through `get_l1_batch_metadata()`:
core/lib/dal/src/blocks_dal.rs:3022-3029

That path only drops rows when `StorageL1Batch -> L1BatchMetadata` fails entirely:
core/lib/dal/src/blocks_dal.rs:3240-3242

But `aggregation_root` itself is not part of the required-field gate in `TryFrom<StorageL1Batch> for L1BatchMetadata`:
core/lib/dal/src/models/storage_block.rs:261-327

So `Some(L1BatchWithMetadata)` does not imply `metadata.aggregation_root.is_some()`. The previous `expect()` was therefore an RPC-edge panic on a DB-representable state.

2. Why the test adds `sqlx` in dev-dependencies
The regression test needs to construct the exact state that the fix is guarding: a post-gateway batch with otherwise present metadata but a cleared `aggregation_root`.

The test does that with a single test-only `UPDATE l1_batches SET aggregation_root = NULL ...`:
core/node/api_server/src/web3/tests/mod.rs:1493-1502

That is why `sqlx.workspace = true` was added under `[dev-dependencies]` only:
core/node/api_server/Cargo.toml:62-72

It does not affect runtime dependencies or the production code path.

peter941221:
Validation is green now.

I fixed the regression fixture so the test reaches the post-gateway missing-aggregation_root path instead of exiting early with Ok(None):
- mark the tx as executed in the L1 batch
- call get_l2_to_l1_log_proof(..., None) instead of forcing ProofBasedGateway

Re-run result:
cargo test -p zksync_node_api_server web3::tests::getting_l2_to_l1_log_proof_without_aggregation_root -- --exact --nocapture

deniallugo-ml:
Closing — we don’t merge standalone cosmetic-only PRs; review overhead outweighs the benefit. (This one is deliberately direct — it discourages re-submission of the same kind.)

